PR #6, opened 2019-04-03
Agent in python

### LINKED ISSUE #4: Incorrect agent state returned
## Steps to reproduce

  1. Replace the file [src/esp/gfx/Viewer.cpp](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/Viewer.cpp) with the file [Viewer.txt](https://github.com/facebookresearch/habitat-sim/files/2973550/Viewer.txt) (Note: the file extension should be changed to cpp)
  2. Build the project
  3. Run the "viewer" executable on one of the habitat-test-scenes
  4. Rotate the camera view using the arrow keys

## Observed Results
  * The y-rotation computed from the cameraNode_->getRotation() is always zero while the one computed from cameraNode_->getAbsoluteTransformation() is correct. The link below shows the results. This link shows the result: [video](https://drive.google.com/open?id=1MXZKLwcvPzyQSHeC6aoifHSI2sMdCjFS)
   * Similar thing is observed when using the python binding [habitat_sim/simulator.py](https://github.com/facebookresearch/habitat-sim/blob/master/habitat_sim/simulator.py). Steps to reproduce this is not included for brevity, but can be provided if needed.

## Expected Results

  * The two rotations (default rotation and transformation based rotation) should be the same

## Relevant Code

```
  std::cout
    << std::fixed
    << std::setprecision(2)
    << "default rotation: " << cameraNode_->getRotation().toRotationMatrix().eulerAngles(1, 0, 2).transpose() * 180 / M_PI << std::endl
    <<"matrix based rotation: " << quaternion.toRotationMatrix().eulerAngles(1, 0, 2).transpose() * 180 / M_PI << std::endl
    << std::endl;
```
Connection of controls to navmesh is disabled for pure teleoperation
```
if (false) { // pathfinder_->isLoaded()) {
```

erikwijmans:
## Motivation and Context

Moves implementation of the agent into python.  There is no performance reason why that all needs to be in c++, so python it is!

Changed a lot of the Eigen parts that touch python to `const Eigen::Ref<const T>` instead of `const T&` as that allows pybind11 to do a better job at sharing memory.

Some additional things moving to python allows for:

1.   Allows for the agents action space to be indexed by any hashable type.  The agent's action space is just a dictionary where the action name (previously only strings) points to an `ActionSpec`.  The action names can now be anything you want (including `int`s).
1. @dhruvbatra Allows for easy change of an agent's actuation parameters during runtime (e.g. in the middle of an episode).  It even allows for redefining the function that implements a given action in the middle of an episode.

@abhiskk @mathfac These changes are not backwards compatible with HAPI.  Once things looks good here, I will make a PR there to support these changes.

Resolves #4 

## How Has This Been Tested

Things run/get rendered out.

Wrote unit tests for the controls, creating the agent, and the sensors render outputs.  Tested with `pytest`!  Can everyone run pytest on their machines to make sure I got the tolerance numbers correct?

For the sensor tests, I also added a custom marker called `gfxtext` which lets you easily filter out any tests that render (so we can use travis or something).  If you do `pytest -m "not gfxtest"`, then only the tests that don't require a rendering context will run :)


habitat_sim/simulator.py
```diff
@@ -28,31 +34,53 @@ def reset(self):
...
+
         self._config = config
-        self._default_agent = self.get_agent(agent_id=config.default_agent_id)
-        agent_cfg = config.agents[config.default_agent_id]
+        self.agents = [Agent(cfg) for cfg in agent_cfgs]
+        for i in range(len(self.agents)):
+            self.agents[i].attach(
+                self._sim.get_active_scene_graph().get_root_node().create_child()
+            )
+            self.agents[i].controls.move_filter_fn = self._step_filer
```
msavva: typo: `_step_filter` to `_step_filter`

habitat_sim/simulator.py
```diff
@@ -78,34 +106,43 @@ def last_state(self):
...
         self._default_agent.act(action)
-        self._default_agent.get_state(self._last_state)
+        self._last_state = self._default_agent.get_state()
         observations = self.get_sensor_observations()
         return observations
 
     def make_action_pathfinder(self, agent_id=0):
         return self._sim.make_action_pathfinder(agent_id)
 
+    def _step_filer(self, start_pos, end_pos):
```
msavva: typo: `_step_filter` to `_step_filter`

habitat_sim/utils.py
```diff
@@ -5,6 +5,72 @@
 # LICENSE file in the root directory of this source tree.
 
 import numpy as np
+import quaternion
+
+
```
msavva: General comment about functions in utils: let's add at least some minimal docstrings now that the functions are fresh. :-)

src/esp/bindings/bindings.cpp
```diff
@@ -606,13 +578,14 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
...
       m, "SimulatorConfiguration")
       .def(py::init(&SimulatorConfiguration::create<>))
-      .def_readwrite("agents", &SimulatorConfiguration::agents)
       .def_readwrite("scene", &SimulatorConfiguration::scene)
       .def_readwrite("default_agent_id",
                      &SimulatorConfiguration::defaultAgentId)
       .def_readwrite("default_camera_uuid",
                      &SimulatorConfiguration::defaultCameraUuid)
       .def_readwrite("gpu_device_id", &SimulatorConfiguration::gpuDeviceId)
+      .def_readwrite("width", &SimulatorConfiguration::width)
```
msavva: Do we need to add `width` and `height` as sim-level properties?  Introducing these is a bit counter to supporting multi-sensor setups with differing resolutions and might be confusing longer-term

src/esp/gfx/Simulator.h
```diff
@@ -5,14 +5,14 @@
...
 #include "WindowlessContext.h"
-#include "esp/agent/Agent.h"
-#include "esp/assets/ResourceManager.h"
 #include "esp/core/esp.h"
 #include "esp/core/random.h"
 #include "esp/scene/SceneConfiguration.h"
 #include "esp/scene/SceneManager.h"
 #include "esp/scene/SceneNode.h"
 
+#include "esp/assets/ResourceManager.h"
```
msavva: Is there a reason for re-ordering this header include to be last?

src/esp/gfx/Simulator.h
```diff
@@ -28,11 +28,11 @@ class Renderer;
 
 struct SimulatorConfiguration {
   scene::SceneConfiguration scene;
-  std::vector<agent::AgentConfiguration::ptr> agents;
   int defaultAgentId = 0;
   int gpuDeviceId = 0;
   std::string defaultCameraUuid = "rgba_camera";
   bool compressTextures = false;
+  int width = 256, height = 256;
```
msavva: See comment above about avoiding sim-level resolution parameters if possible

tests/test_agent.py
```diff
@@ -0,0 +1,96 @@
...
+    assert habitat_sim.utils.angle_between_quats(s1.rotation, s2.rotation) < 1e-5
+
+
+def test_bad_state():
+    agent = habitat_sim.Agent()
+    with pytest.raises(habitat_sim.errors.InvalidAttachedObject):
+        _ = agent.state
+
+
+def test_attach_dettach():
```
msavva: typo: `dettach` to `detach`

examples/demo_runner.py
```diff
@@ -75,16 +76,21 @@ def make_cfg(self, settings):
...
+                "turn_left", habitat_sim.agent.ActuationSpec(amount=10.0)
+            ),
+            "turn_right": habitat_sim.agent.ActionSpec(
+                "turn_right", habitat_sim.agent.ActuationSpec(amount=10.0)
+            ),
         }
-        sim_cfg.agents = [agent_cfg]
 
-        return sim_cfg
+        return sim_cfg, [agent_cfg]
```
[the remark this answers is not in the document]
erikwijmans: The difficulty with that is comparison of configs.  If the agent config is part of the sim config, that would imply that two sim configs with different agent configs would be different.  Since we have a custom `==` operator implemented because of shared pointers, this isn't the case as the agent config definition lives in python.

It is also nice as you can now change the agent config without the backend reloading anything.

examples/demo_runner.py
```diff
@@ -226,8 +229,8 @@ def print_semantic_scene(self):
             input("Press Enter to continue...")
 
     def init_common(self):
-        self._sim_cfg = self.make_cfg(self._sim_settings)
-        self._sim = habitat_sim.Simulator(self._sim_cfg)
+        self._sim_cfg, self._agent_cfgs = self.make_cfg(self._sim_settings)
+        self._sim = habitat_sim.Simulator(self._sim_cfg, self._agent_cfgs)
```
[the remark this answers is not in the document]
erikwijmans: Same response as above :)

habitat_sim/utils.py
erikwijmans: On it

src/esp/bindings/bindings.cpp
erikwijmans: Currently these are needed as the renderer owns the global frame buffer.  The current code just looks up the width and height of sensor 0 and uses that for all the rendering.

src/esp/gfx/Simulator.h
erikwijmans: There is a name collisions somewhere if that header is include further up.  Previously it seems that having the header for the agent included dealt with whatever the issue is (the header for the agent was before that header)

examples/demo_runner.py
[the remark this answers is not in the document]
erikwijmans: What would work is having a higher level config that has both a backend config (the current `sim_cfg`) and a list of `agent_cfg`.  Then the comparison would function as expected.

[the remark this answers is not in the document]
msavva: Yes, agreed -- I think a higher level config perhaps general `Configuration` that includes the `SimulatorConfig` and `AgentConfig`s is a good idea.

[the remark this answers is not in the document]
erikwijmans: Added

tests/test_sensors.py
```diff
@@ -0,0 +1,45 @@
...
+import habitat_sim.utils
+import habitat_sim.errors
+import numpy as np
+import quaternion
+import pytest
+import os.path as osp
+
+
+_test_scene = osp.abspath(
+    osp.join(osp.dirname(__file__), "17DRP5sb8fy", "17DRP5sb8fy.glb")
```
mathfac: @erikwijmans, really like sensor_tests. We have public testing scenes (http://dl.fbaipublicfiles.com/habitat/habitat-test-scenes.zip). Can that be used for testing?

erikwijmans: Those don't have semantic labels unfortunately.  Although it is probably worth adding tests for RGB and Depth on public testing scenes

erikwijmans: Added.  That was easier than I thought it would be.  Yay pytest!

Issue #13, opened 2019-04-11
setup.py does not work


danielgordon10:
## Steps to reproduce

  1. git clone habitat-sim blah blah
  2. run `python setup.py install --headless
  3. run `python` then `import habitat_sim`

## Observed Results
```
Traceback (most recent call last):
  File "/data/xkcd/habitat-sim/habitat_sim/bindings/__init__.py", line 10, in <module>
    from habitat_sim._ext.habitat_sim_bindings import Simulator as SimulatorBackend
ModuleNotFoundError: No module named 'habitat_sim._ext'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data/xkcd/habitat-sim/habitat_sim/bindings/dev_bindings.py", line 72, in <module>
    import habitat_sim_bindings
ModuleNotFoundError: No module named 'habitat_sim_bindings'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "examples/example.py", line 9, in <module>
    import demo_runner as dr
  File "/data/xkcd/habitat-sim/examples/demo_runner.py", line 15, in <module>
    import habitat_sim
  File "/data/xkcd/habitat-sim/habitat_sim/__init__.py", line 12, in <module>
    from .simulator import *
  File "/data/xkcd/habitat-sim/habitat_sim/simulator.py", line 7, in <module>
    from habitat_sim.bindings import *
  File "/data/xkcd/habitat-sim/habitat_sim/bindings/__init__.py", line 18, in <module>
    from .dev_bindings import *
  File "/data/xkcd/habitat-sim/habitat_sim/bindings/dev_bindings.py", line 84, in <module>
    raise ImportError(msg)
ImportError: 
Could not import habitat sim bindings
Please follow the building instructions in the README
-----------------------------------------------------------

Could not find the expected bindings build folder: "/data/xkcd/habitat-sim/build/esp/bindings"
Make sure to not delete the build folder after building
```

Seems like those files don't exist. Installing through the "developer instructions" still works fine.
```bash
(habitat-env) xkcd@ava:/data/xkcd/habitat-sim$ ls build
bdist.linux-x86_64  lib.linux-x86_64-3.6  temp.linux-x86_64-3.6
```

## Expected Results

  * habitat_sim imports correctly

erikwijmans:
[...]

PR #14, opened 2019-04-12
Better install with setup.py

erikwijmans:
## Motivation and Context

Issue #13 brought up that the current method of figuring out how to import the bindings leads to confusing error messages when things are installed with `setup.py.   The relevant error, that it couldn't import `habitat_sim._ext.habitat_sim_bindings`, is buried by the error message that it couldn't import the bindings in developer mode.  

This PR uses `setup.py` to modify the value of a variable such that the code knows where it should expect the bindings to be, and only looks in that location.

## How Has This Been Tested

Installed in a clean conda environment.  After deleting the `build` folder and doing `unset PYTHONPATH`, `python examples/example.py` works.   Doing

```
cd ..
python
>>> import habitat_sim
```
also works.  That `cd ..` is still key however.

In a different clean conda environment, I built with `build.sh`.  Doing 
```
python
>>> import habitat_sim
```

succeeds and `python examples/example.py` fails (this is what I would expect to happen).  `python examples/example.py` works if I add `/path/to/habitat-sim` to my `PYTHONPATH`.

### DIFF
--- a/habitat_sim/bindings/__init__.py
+++ b/habitat_sim/bindings/__init__.py
@@ -9,7 +9,6 @@ from habitat_sim.bindings.mode import use_dev_bindings
 
 if use_dev_bindings:
     from .dev_bindings import *
-
 else:
     from habitat_sim._ext.habitat_sim_bindings import Simulator as SimulatorBackend
 
--- a/habitat_sim/bindings/__init__.py
+++ b/habitat_sim/bindings/__init__.py
@@ -5,8 +5,12 @@
 # LICENSE file in the root directory of this source tree.
 
 from habitat_sim.bindings.modules import modules
+from habitat_sim.bindings.mode import use_dev_bindings
 
-try:
+if use_dev_bindings:
+    from .dev_bindings import *
+
+else:
     from habitat_sim._ext.habitat_sim_bindings import Simulator as SimulatorBackend
 
     exec(
@@ -14,7 +18,5 @@ try:
             ", ".join(modules)
         )
     )
-except ImportError:
-    from .dev_bindings import *
 
 __all__ = ["SimulatorBackend"] + modules
--- a/habitat_sim/bindings/dev_bindings.py
+++ b/habitat_sim/bindings/dev_bindings.py
@@ -72,9 +72,8 @@ try:
     import habitat_sim_bindings
 except ImportError:
     msg = """
-Could not import habitat sim bindings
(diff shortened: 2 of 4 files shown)


danielgordon10:
I think this now causes an issue if you install via setup.py but also have habitat-sim in your pythonpath. When I had that combo I got this error:
```
Python 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import habitat_sim
Traceback (most recent call last):
  File "/raid/xkcd/external_code/habitat-sim/habitat_sim/bindings/dev_bindings.py", line 72, in <module>
    import habitat_sim_bindings
ModuleNotFoundError: No module named 'habitat_sim_bindings'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/raid/xkcd/external_code/habitat-sim/habitat_sim/__init__.py", line 12, in <module>
    from .agent import *
  File "/raid/xkcd/external_code/habitat-sim/habitat_sim/agent/__init__.py", line 7, in <module>
    from .agent import *
  File "/raid/xkcd/external_code/habitat-sim/habitat_sim/agent/agent.py", line 9, in <module>
    from .controls import ObjectControls, ActuationSpec
  File "/raid/xkcd/external_code/habitat-sim/habitat_sim/agent/controls.py", line 10, in <module>
    import habitat_sim.bindings as hsim
  File "/raid/xkcd/external_code/habitat-sim/habitat_sim/bindings/__init__.py", line 11, in <module>
    from .dev_bindings import *
  File "/raid/xkcd/external_code/habitat-sim/habitat_sim/bindings/dev_bindings.py", line 83, in <module>
    raise ImportError(msg)
ImportError: 
Failed to to import habitat sim bindings in developer mode
----------------------------------------------------------

Could not find the expected bindings build folder: "/raid/xkcd/external_code/habitat-sim/build/esp/bindings"
Make sure to not delete the build folder after building
        
>>> quit()
```
I was able to fix it by removing habitat-sim from the pythonpath, but now I just feel like I'm going in circles.

erikwijmans:
Yeah. If you install via setup.py, it’ll be weird if you also add habitat sim to your pythonpath. At some point we should switch to just having one build method.

PR #15, opened 2019-04-14
Fixes and improvements to benchmark script

msavva:
## Motivation and Context

The `benchmark.py` script assumed the presence of a `test.glb` file and had unexpected behavior when the file was not present.  Moreover, multi-process benchmarking was not incorporated, making it hard to reproduce the benchmark table out-of-the-box.  @danielgordon10 thanks for bringing up this issue.

This PR removes defaulting of the `--scene` argument from both `benchmark.py` and `example.py` and requires the user to specify the intended scene, avoiding confusing behavior and unexpected errors.  It also incorporates multi-process benchmarking implemented by @erikwijmans .

## How Has This Been Tested

Download the Matterport3D Habitat data and run `examples/benchmark.py --scene /path/to/mp3d/17DRP5sb8fy/17DRP5sb8fy.glb`. The benchmark script should output a table similar to the one below:
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution             rgb           rgbd     depth_only  semantic_only  rgbd_semantic
128 x 128         4288.3         2052.2         4266.9          540.3          429.1
256 x 256         3162.0         1704.4         3245.0          518.6          387.3
512 x 512         1611.6          803.4         1614.5          451.1          280.9
 ==============================================================================
 ================ Performance (FPS) NPROC=3 ===================================
Resolution             rgb           rgbd     depth_only  semantic_only  rgbd_semantic
128 x 128         8629.4         4258.8         8634.9          531.3          468.2
256 x 256         7483.4         3686.6         7462.1          522.5          456.1
512 x 512         4346.8         2228.6         4251.1          540.0          435.3
 ==============================================================================
 ================ Performance (FPS) NPROC=5 ===================================
Resolution             rgb           rgbd     depth_only  semantic_only  rgbd_semantic
128 x 128         8484.4         4178.5         8415.8          528.0          467.8
256 x 256         7444.6         3651.7         7423.0          522.4          455.7
512 x 512         7069.9         3503.1         7187.1          540.0          472.5
 ==============================================================================
```

## Types of changes

## Checklist

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -35,7 +35,6 @@ default_settings["compute_action_shortest_path"] = False
 
 default_settings["max_frames"] = args.max_frames
 
-demo_runner = dr.DemoRunner(default_settings, dr.DemoRunnerType.BENCHMARK)
 
 benchmark_items = {
     "rgb": {},
@@ -45,32 +44,46 @@ benchmark_items = {
     "rgbd_semantic": {"depth_sensor": True, "semantic_sensor": True},
 }
 
-# resolutions = [128] # (debug)
+#  resolutions = [128] # (debug)
 resolutions = [128, 256, 512]
+nprocs_tests = [1, 3, 5]
 
-performance = []
-for resolution in resolutions:
-    default_settings["width"] = default_settings["height"] = resolution
-    perf = {}
-    for key, value in benchmark_items.items():
-        print(" ---------------------- %s ------------------------ " % key)
-        settings = default_settings.copy()
-        settings.update(value)
-        perf[key] = demo_runner.benchmark(settings).get("fps")
-        print(
-            " ====== FPS (%d x %d, %s): %0.1f ======"
-            % (settings["width"], settings["height"], key, perf[key])
+performance_all = {}
+for nprocs in nprocs_tests:
+    default_settings["num_processes"] = nprocs
+    performance = []
+    for resolution in resolutions:
+        default_settings["width"] = default_settings["height"] = resolution
+        perf = {}
+        for key, value in benchmark_items.items():
+            demo_runner = dr.DemoRunner(default_settings, dr.DemoRunnerType.BENCHMARK)
+            print(" ---------------------- %s ------------------------ " % key)
+            settings = default_settings.copy()
+            settings.update(value)
+            perf[key] = demo_runner.benchmark(settings).get("fps")
+            print(
+                " ====== FPS (%d x %d, %s): %0.1f ======"
+                % (settings["width"], settings["height"], key, perf[key])
+            )
+        performance.append(perf)
+
+    performance_all[nprocs] = performance
+
+for nproc, performance in performance_all.items():
+    print(
+        " ================ Performance (FPS) NPROC={} ===================================".format(
+            nproc
         )
-    performance.append(perf)
-
-print(" ================ Performance (FPS) ===========================================")
-title = "Resolution "
-for key, value in perf.items():
-    title += "%15s" % key
-print(title)
-for idx in range(len(performance)):
-    row = "%d x %d" % (resolutions[idx], resolutions[idx])
-    for key, value in performance[idx].items():
-        row += "%15.1f" % value
-    print(row)
-print(" ==============================================================================")
+    )
+    title = "Resolution "
+    for key, value in perf.items():
+        title += "%15s" % key
+    print(title)
+    for idx in range(len(performance)):
+        row = "%d x %d" % (resolutions[idx], resolutions[idx])
+        for key, value in performance[idx].items():
+            row += "%15.1f" % value
+        print(row)
+    print(
+        " =============================================================================="
+    )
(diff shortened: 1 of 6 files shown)


examples/benchmark.py
```diff
@@ -10,7 +10,7 @@
 import numpy as np
 
 parser = argparse.ArgumentParser("Running benchmarks on simulator")
-parser.add_argument("--scene", type=str, default="test.glb")
+parser.add_argument("--scene", type=str, required=True)
```
erikwijmans: Let's keep a default for this arguments and make it one of the habitat-test-scenes, say `data/scene_datasets/habitat-test-scenes/skokloster-castle.glb`

examples/benchmark.py
```diff
@@ -10,7 +10,7 @@
...
 parser.add_argument(
     "--max_frames",
     type=int,
```
erikwijmans: Also, let's add a `--benchmark-semantic` flag so that this script can be run without semantics?

examples/demo_runner.py
```diff
@@ -237,16 +240,47 @@ def init_common(self):
...
+        best_perf = None
+        for _ in range(3):
+
+            if _barrier is not None:
+                _barrier.wait()
+                if _idx == 0:
+                    _barrier.reset()
+
+            perf = self.do_time_steps()
+            if best_perf is None or perf["fps"] > best_perf["fps"]:
```
erikwijmans: This should probably have a comment on why taking the max is a common/legitimate thing in benchmarking.  Since I did this, here is my reasoning:

```
The variance introduced between runs is due to the worker threads being interrupted a different number of times by the kernel, not due to difference in the speed of the code itself.  The most accurate representation of the performance would be a run where the kernel never interrupted the workers, but this isn't feasible,  so we just take the run with the least number of interrupts (the fastest) instead.
```

examples/example.py
```diff
@@ -10,9 +10,9 @@
 import numpy as np
 
 parser = argparse.ArgumentParser()
+parser.add_argument("--scene", type=str, required=True)
```
erikwijmans: Same comment about default arg.

examples/benchmark.py
danielgordon10: Additionally, if none is specified, can we auto download and extract for them?

examples/benchmark.py
```diff
@@ -43,32 +43,46 @@
...
+        title += "%15s" % key
+    print(title)
+    for idx in range(len(performance)):
+        row = "%d x %d" % (resolutions[idx], resolutions[idx])
+        for key, value in performance[idx].items():
+            row += "%15.1f" % value
+        print(row)
+    print(
+        " =============================================================================="
+    )
```
danielgordon10: Can you fix the spacing to have the titles more centered? Right now it looks like 
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution             rgb           rgbd     depth_only  semantic_only  rgbd_semantic
128 x 128         4288.3         2052.2         4266.9          540.3          429.1
256 x 256         3162.0         1704.4         3245.0          518.6          387.3
512 x 512         1611.6          803.4         1614.5          451.1          280.9
```

But should look like 
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution         rgb            rgbd        depth_only    semantic_only  rgbd_semantic
128 x 128         4288.3         2052.2         4266.9          540.3          429.1
256 x 256         3162.0         1704.4         3245.0          518.6          387.3
512 x 512         1611.6          803.4         1614.5          451.1          280.9
 ==============================================================================
 ```
Or
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution           rgb           rgbd     depth_only  semantic_only  rgbd_semantic
128 x 128         4288.3         2052.2         4266.9          540.3          429.1
256 x 256         3162.0         1704.4         3245.0          518.6          387.3
512 x 512         1611.6          803.4         1614.5          451.1          280.9
 ==============================================================================
 ```
Or
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution        rgb            rgbd           depth_only      semantic_only  rgbd_semantic
128 x 128         4288.3         2052.2         4266.9          540.3          429.1
256 x 256         3162.0         1704.4         3245.0          518.6          387.3
512 x 512         1611.6          803.4         1614.5          451.1          280.9
 ==============================================================================
 ```

mathfac:
Just follow up, what is our plan regarding this PR.

msavva:
@mathfac thanks for the ping -- this PR is on my plate still. I will get it updated and ready for review soon.

mathfac:
Maybe we can use it in continues integration loop later.

msavva:
@mathfac @danielgordon10 @erikwijmans the PR is now updated with the following changes and ready for reviewing:
- Fix output column header + fps alignment
- Add parameter to enable semantic sensor (disabled by default)
- Remove hardcoded resolution and nprocs and make into proper defaulted argparse arguments
- Automatically download and extract test scene data if not present locally

PR #21, opened 2019-04-17
Exposes the subclasses used for {Mp3d, Suncg}SemanticScene

erikwijmans:
## Motivation and Context

PyBind11 requires explicit bindings whenever subclasses are used to implement an abstract base class or when a subclass is used to override virtual methods.  Those subclasses weren't bound, which results in memory corruptions.

Also binds a lambda for the OBB rotation as we can't convert an `Eigen::Quaternion` into python

https://github.com/facebookresearch/habitat-api/issues/59

## How Has This Been Tested

With the following re-production script:

```
import habitat_sim

cfg = habitat_sim.SimulatorConfiguration()
cfg.agents = [habitat_sim.AgentConfiguration()]
cfg.scene.id = "data/scene_datasets/mp3d/ur6pFq6Qu1A/ur6pFq6Qu1A.glb"
sim = habitat_sim.Simulator(cfg)

scene = sim.semantic_scene
for obj in scene.objects:
    obj.id
    obj.category.name()
    obj.category.index()

for region in scene.regions:
    region.id
    region.category.name()
    region.category.index()

for level in scene.levels:
    level.id

sim.close()
```

After landing #6, I will make another PR to add a this as a test

## Types of changes

### DIFF
--- a/src/esp/scene/Mp3dSemanticScene.cpp
+++ b/src/esp/scene/Mp3dSemanticScene.cpp
@@ -2,6 +2,7 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+#include "esp/scene/Mp3dSemanticScene.h"
 #include "SemanticScene.h"
 
 #include <algorithm>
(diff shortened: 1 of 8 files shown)


src/esp/scene/Mp3dSemanticScene.cpp
```diff
@@ -2,6 +2,7 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+#include "esp/scene/Mp3dSemanticScene.h"
```
msavva: Rest of the code adopts the style of including the sibling header without the relative path (i.e. only `Mp3dSemanticScene.h` in this case)

erikwijmans: Fixing and landing

PR #27, opened 2019-04-19
Speed up compilation using ALL THE THREADS

### LINKED ISSUE #24: Slow setup.py install
Can we switch https://github.com/facebookresearch/habitat-sim/blob/master/setup.py#L86 to `-j ` instead of `-j 4` This will use all available cpus regardless of the system. While it might be "a burden" while setting up, do we expect people to be multitasking while installing habitat-sim or do we want it to install as fast as possible? I would argue the former.

danielgordon10:
## Motivation and Context

Closes #24

## Types of changes

## Checklist


erikwijmans:
It now uses lots of threads, merging :)

PR #29, opened 2019-04-22
Generic instance mesh

erikwijmans:
## Motivation and Context

Adds support for different variants of instance labeled meshes.  Supports slight difference in ways to store things like the object ID and index buffer and support both triangle and quad meshes.

Switches the instance mesh rendering to indexing a texture as a 2D array to lookup the instance ID for each primitive instead of using a 4th element per vertex as the latter requires vertex doubling.

## How Has This Been Tested

The semantic sensor unit test passes :)

Also tested with an instance labelled quad-mesh.

Output from example.py:

test sem 00000

## Types of changes


src/esp/assets/FRLInstanceMeshData.h
```diff
@@ -23,20 +20,17 @@ namespace assets {
...
-    Magnum::GL::Buffer ibo;
-  };
-
-  InstanceMeshData() : BaseMesh(SupportedMeshType::INSTANCE_MESH){};
-  virtual ~InstanceMeshData(){};
+  FRLInstanceMeshData()
+      : GenericInstanceMeshData(SupportedMeshType::INSTANCE_MESH){};
+  virtual ~FRLInstanceMeshData(){};
 
   bool from_ply(const std::string& ply_file);
```
jia-git: Can we just rename from_ply to loadPLY, and make it override base class? I think from_ply is only called from couple places.

erikwijmans: Done

src/esp/assets/GenericInstanceMeshData.h
```diff
@@ -0,0 +1,70 @@
...
+#include <Magnum/GL/Mesh.h>
+
+#include <Magnum/GL/Texture.h>
+#include "BaseMesh.h"
+#include "esp/core/esp.h"
+
+namespace esp {
+namespace assets {
+
+Magnum::GL::Texture2D createInstanceTexture(float* data, const int texSize);
```
msavva: I would add a brief comment here as the purpose of this function is not immediately obvious

erikwijmans: Copied the comment on the implementation to here

PR #38, opened 2019-05-09
Setup py improvements

mosra:
## Motivation and Context

While #24 said that

> do we expect people to be multitasking while installing habitat-sim [...]?

I'm actually one of those crazy multitaskers :) 100% CPU wouldn't be a problem, but when running on all cores, *something* in the build demands 8 GB of RAM which I usually don't have, so my system comes to a grinding halt. (I still want to investigate what's responsible and if it's possible to tell `ninja` to not overschedule such jobs.)

While the default is still *all the cores*, it now allows to override the setting using the setuptools' builtin `--parallel` option. Together with redirecting it to use a pre-existing CMake build (which was set up manually for #35) dir my command-line looks like this:

```sh
python setup.py build_ext --build-temp=build --parallel 3
```

Apart from this, because I use some system deps, I'm not interested in `setup.py` implicitly pulling submodules that I am not interested in, so it runs `git submodule update` only the very first time.

## How Has This Been Tested

I'm now able to *both* have 163 tabs open in my browser *and* build Habitat :)

## Types of changes

- [x] Docs change / refactoring / ~~dependency upgrade~~ buildsystem update

## Checklist

- [x] My change requires a change to the documentation. -- maybe? please tell me where

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,7 +92,10 @@ class CMakeBuild(build_ext):
     def build_extension(self, ext):
         extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
 
-        if in_git():
+        # Init & update all submodules if not already (the user might be pinned
+        # on some particular commit or have working tree changes, don't destroy
+        # those)
+        if in_git() and not os.path.exists(".git/modules"):
             subprocess.check_call(
                 ["git", "submodule", "update", "--init", "--recursive"]
             )
--- a/setup.py
+++ b/setup.py
@@ -8,12 +8,14 @@
 Adapted from: http://www.benjack.io/2017/06/12/python-cpp-tests.html
 """
 
+import argparse
 import builtins
 import glob
 import json
 import os
 import os.path as osp
 import re
+import shlex
 import subprocess
 import sys
 from distutils.version import StrictVersion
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -92,7 +92,10 @@ def run(self):
     def build_extension(self, ext):
         extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
 
-        if in_git():
+        # Init & update all submodules if not already (the user might be pinned
+        # on some particular commit or have working tree changes, don't destroy
+        # those)
+        if in_git() and not os.path.exists(".git/modules"):
```
erikwijmans: I like the idea of this -- I have resorted to some pretty hacky hacks to get git to not update submodules (adding a random space or newline to a file and then not committing it being my favorite).

However, I also like that currently when people pull a new build, they don't have to worry about updating submodules.  Would adding a `--no-update-modules` flag be sufficient (I am about to submit a PR adding a `--cmake-args` flag, so I can add it to that)?

CC: @msavva

mosra: I see your point. Yes that would work too, I would just need to remember to always add that flag (which is okay) :)

A more general question from a `setup.py` noob -- do setuptools have something similar to cmake's cache? Like, you enable / set up some flags and properties and they get remembered for the next invocation; so you don't have to specify the whole command-line but can do just `./setup.py` and it will pick up the rest from *somewhere*, which is (like cmake's build dir) not commited to the git repository but private to your working copy? Or am I asking for the impossible?

erikwijmans: Yeah... unfortunately setup.py doesn't do that.  We could roll our own as pickling out `args` and reloading if it exists would work.....

msavva: I also like the idea.  I agree with @erikwijmans 's suggestion to make this into a `--no-update-modules` flag. It's good to accommodate for the expectation that everything should work through a naive pull-and-build.  I'm not aware of a way to make `setup.py` cache flags etc. :(

erikwijmans: Made a PR into this PR with a go at caching (it contains more than just that tho).

mosra:
Since #40 depends on this (and reverts a part), how should we proceed to merge? Should I remove the `.git/submodules` part from this one and let it be squashed into #40?

erikwijmans:
Merged #40 into this.  This should be good to go too!

Issue #36, opened 2019-05-07
Passing commands to pip install


danielgordon10:
It looks like the current `setup.py` script does not support the `--install-option` arg.
https://stackoverflow.com/questions/18725137/how-to-obtain-arguments-passed-to-setup-py-from-pip-with-install-option

mosra:
Nice, just today I was wondering if there's a canonical way to pass things through to the underlying cmake invocation. This could be it.

erikwijmans:
I've looked as this before.  Unfortunately install options don't work for both `install` and `build_ext`.   Which would limit us to just using `install` and `develop` for development workflow, but neither are particularly nice when you have both python and c++ code. The pattern I see from other repos that have both python and c++ code is to parse the args yourself in `setup.py` and then use environment variables to interact with pip.  We can add a `--cmake-args` command line arg and a `CMAKE_ARGS` environment variable

mosra:
I ended up configuring the cmake build manually and then redirecting `setup.py` to use the existing directory instead of creating its own, for example:

```sh
python setup.py install build_ext --build-temp=build
```

For my needs this has all the flexibility I want, however might not be convenient enough for others.

mathfac:
Should be solved.

PR #40, opened 2019-05-09
Setup.py -- Argprase, Cmake Args, and caching

erikwijmans:
## Motivation and Context

See #38 and #36 for more context

Rewrite the arg parsing in setup.py to use `argparse` as this allows for easier extension in the future and gives a nice `--help` message!

Adds a command line flag for passing arguments to cmake.  Note that due to `argparse`'s parsing, cmake args must be assed as `--cmake-args="...."`.  Doing `--cmake-args "...."` (no =) won't work :(
This can be passed to pip via the `CMAKE_ARGS` environment variable.

Adds caching of arguments passed to `setup.py` via a simple `json` file written into the build_temp dir.  Arguments can be cached by adding `--cache-args` to the invocation of `setup.py`, any current argument cache is ignored.  I have blacked listed `force_cmake` and `cache_args` from the cache.

@mosra I pointed this PR at your PR as we are both playing with things in setup.py

## How Has This Been Tested

Running around and building things.

## Types of changes


setup.py
```diff
@@ -122,32 +174,31 @@ def build_extension(self, ext):
...
-                ["cmake", ext.sourcedir] + cmake_args, cwd=self.build_temp, env=env
+                shlex.split("cmake -H{} -B{}".format(ext.sourcedir, self.build_temp))
+                + cmake_args,
+                env=env,
             )
 
-        subprocess.check_call(
-            ["cmake", "--build", "."] + build_args, cwd=self.build_temp
-        )
+        subprocess.check_call(shlex.split("cmake --build {}".format(self.build_temp)))
```
mosra: This is missing the `build_args`, so `--config` or `-jN` doesn't get set correctly. This worked for me:

```suggestion
        subprocess.check_call(shlex.split("cmake --build {}".format(self.build_temp)) + build_args)
```

erikwijmans: Whoops!

erikwijmans:
I’ll think about if there’s an easy way to cache the build_ext options. 

One thing you can do is add those to the setup.cfg as build_ext will read args from there also.

erikwijmans:
Give that a try.  I also made `--build-temp=build` the default as that is nicer for helping people debug things.

PR #46, opened 2019-05-20
setup.py: Can't do inplace=1 because that breaks the install

erikwijmans:
## Motivation and Context

setup.py makes no sense and the install breaks if you set inplace to be the default for build_ext!

## How Has This Been Tested

In a fresh conda env with `python setup.py install`

## Types of changes


erikwijmans:
Yeah, just tagging you for visibility :)

PR #53, opened 2019-05-28
[WIP] Test emscripten build for WebGL deployment

msavva:
## Motivation and Context

Preliminary testing of habitat-sim compiling through emscripten for WebGL deployment.  Currently, just a "pile of hacks" to get viewer utility binary working only with pre-loaded data package containing test scenes.

Before landing will need to:
- [ ] Figure out proper handling of dynamic asset loading (example dynamic data package loading [here](https://github.com/kripken/BananaBread/blob/master/cube2/js/game-setup.js))
- [ ] Remove hack ignoring `texSize` setting (see [here](https://github.com/facebookresearch/habitat-sim/blob/decouple-util-deps/src/esp/gfx/GenericShader.cpp#L166-L169))
- [x] Also compile with headless contexts so that things other than viewer utility work
- [ ] Lots of other stuff ...

## How Has This Been Tested

1. Download and install emscripten SDK from https://emscripten.org/
1. Make sure to source emscripten env variables before building (i.e. `source ~/code/emsdk/emsdk_env.sh` or similar)
1. Run `./build_js.sh`
1. Follow instructions at end of build to launch server and open viewer in browser tab

## Types of changes

## Checklist


src/esp/gfx/GenericShader.cpp
```diff
@@ -122,12 +122,14 @@ enum { TextureLayer = 0 };
...
-  Magnum::GL::Shader vert{Magnum::GL::Version::GL410,
-                          Magnum::GL::Shader::Type::Vertex};
-  Magnum::GL::Shader frag{Magnum::GL::Version::GL410,
-                          Magnum::GL::Shader::Type::Fragment};
+  // MAGNUM_ASSERT_GL_VERSION_SUPPORTED(Magnum::GL::Version::GL410);
+#ifdef BUILD_WEBGL
+  Magnum::GL::Version glVersion = Magnum::GL::Version::GLES300;
+#else
+  Magnum::GL::Version glVersion = Magnum::GL::Version::GL410;
+#endif
```
mosra: There's the same handy `CORRADE_TARGET_EMSCRIPTEN` macro that you can use for preprocessor branching -- no need to propagate `BUILD_WEBGL` from CMake either :) In this case, however, I think `MAGNUM_TARGET_WEBGL` would make more sense. [Docs](https://doc.magnum.graphics/magnum/Magnum_8h.html#define-members). These macros get defined whenever you include any Corrade or Magnum header.

src/esp/gfx/GenericShader.cpp
```diff
@@ -160,8 +162,11 @@ GenericShader& GenericShader::bindTexture(Magnum::GL::Texture2D& texture) {
 
   texture.bind(TextureLayer);
 
+// TODO this is a hack and terrible! Properly set texSize for WebGL builds
```
mosra: I guess having `bindTexture(Texture2D& texture, int size)` would be a solution, right? Unfortunately until ES 3.1 there's no way to query texture image sizes.

src/utils/datatool/CMakeLists.txt
```diff
@@ -1,9 +1,22 @@
...
+if(USE_SYSTEM_ASSIMP)
+  find_package(ASSIMP REQUIRED)
+else()
+  set(ASSIMP_BUILD_ASSIMP_TOOLS OFF CACHE BOOL "ASSIMP_BUILD_ASSIMP_TOOLS" FORCE)
+  set(ASSIMP_BUILD_TESTS OFF CACHE BOOL "ASSIMP_BUILD_TESTS" FORCE)
+  set(BUILD_SHARED_LIBS OFF CACHE BOOL "ASSIMP_BUILD_TESTS" FORCE)
+  add_subdirectory("${DEPS_DIR}/assimp" "${CMAKE_CURRENT_BINARY_DIR}/assimp")
+  include_directories(SYSTEM "${DEPS_DIR}/assimp/include")
+  # message(STATUS "Found ASSIMP in ${ASSIMP_INCLUDE_DIR}")
+endif()
```
mosra: Why not moving just the `find_package()` here and keeping the rest inside `dependencies.cmake`? That should work even when Assimp is not present.

Also, I'm not sure about  `"${CMAKE_CURRENT_BINARY_DIR}/assimp"`, that feels out of place :)

src/esp/gfx/CMakeLists.txt
```diff
@@ -1,14 +1,30 @@
...
+  GenericDrawable.cpp
+  GenericShader.cpp
+  RenderCamera.cpp
+  Renderer.cpp
+)
+
+# If not WebGL build, also include offscreen render-reliant code
+# TODO: this should be built as well (with emscripten windowless contexts)
+if(NOT BUILD_WEBGL)
+  list(APPEND gfx_SOURCES Simulator.cpp WindowlessContext.cpp)
```
mosra: The Magnum builtin `WindowlessEglApplication` works as-is on Emscripten, but I'm not sure about your modified version -- the extension APIs most definitely aren't defined there. I can take a look at this once I get to integrating your WindowlessContext modifications back into Magnum.

build_js.sh
```diff
@@ -0,0 +1,31 @@
...
+cmake ../src \
+    -DBUILD_WEBGL=ON \
+    -DBUILD_GUI_VIEWERS=ON \
+    -DBUILD_PYTHON_BINDINGS=OFF \
+    -DBUILD_DATATOOL=OFF \
+    -DBUILD_PTEX_SUPPORT=OFF \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_PREFIX_PATH="$EMSCRIPTEN" \
+    -DTARGET_GLES2=OFF \
+    -DTARGET_WEBGL=ON \
```
mosra: No such flag -- Magnum detects Emscripten based on the toolchain file and then enables WebGL implicitly.

build_js.sh
```diff
@@ -0,0 +1,31 @@
...
+    -DCMAKE_TOOLCHAIN_FILE="../src/deps/corrade/toolchains/generic/Emscripten-wasm.cmake" \
+    -DCMAKE_INSTALL_PREFIX="." \
+    -DCMAKE_CXX_FLAGS="-s FORCE_FILESYSTEM=1 -s TOTAL_MEMORY=1677721600 --preload-file $DATA_DIR/scene_datasets/habitat-test-scenes@/" \
```
mosra: I think [`-s ALLOW_MEMORY_GROWTH=1`](https://emscripten.org/docs/optimizing/Optimizing-Code.html#memory-growth) would be a better option here, it has no perf penalty on Wasm.

Just FYI -- unfortunately as far as I checked the `--preload-file` bundles the files as base64 inside `*.js` instead of (the obvious) binary inside `*.wasm`, which makes the download sizes *extra bad* :(

src/utils/viewer/viewer.html
```diff
@@ -0,0 +1,29 @@
...
+<head>
+  <meta charset="UTF-8" />
+  <title>Habitat WebGL Viewer</title>
+  <link rel="stylesheet" href="WebApplication.css" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+</head>
+
+<body>
+  <h1>Habitat WebGL Viewer</h1>
+  <div id="container">
```
mosra: Adding `class="fullsize"` will make it [occupy the whole browser window](https://doc.magnum.graphics/magnum/platforms-html5.html#platforms-html5-layout).

build_js.sh
msavva: Good to know that `ALLOW_MEMORY_GROWTH=1` has no perf penalty. Packaging the data with `--preload-file` is a temporary kludge that I used for testing. We can't rely on packaging all of our assets in this way.  A filesystem abstraction transparently supporting web assets is the proper solution to this 😅.

mathfac:
If this PR will be rebased on latest master we will get proper CI testing.

build_js.sh
```diff
@@ -0,0 +1,30 @@
...
+mkdir -p build_js
+cd build_js
+
+cmake ../src \
+    -DBUILD_GUI_VIEWERS=ON \
+    -DBUILD_PYTHON_BINDINGS=OFF \
+    -DBUILD_DATATOOL=OFF \
+    -DBUILD_PTEX_SUPPORT=OFF \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_PREFIX_PATH="$EMSCRIPTEN" \
```
abhiskk: The `$EMSCRIPTEN` variable should be explicitly set to the prefix where emscripten was built otherwise it picks up the wrong compiler.

msavva: Thanks @abhiskk for testing out this PR and support in my absence.  Regarding, the `$EMSCRIPTEN` variable: it is defined by sourcing the installed emscripten distribution's `emsdk_env.sh` file.  so it is in general a good idea to leave that not explicitly set in the script here.  Perhaps it would be good to include in the script a `source /path/to/emsdk_env.sh` statement though and ask people to modify as necessary.

mosra:
Since #133 got merged, this can be closed, right?

msavva:
> Since #133 got merged, this can be closed, right? Or is here something left that didn't get into #133 (JS bindings code)?

Yes, it should be getting closed fairly soon. There are indeed a few JS bindings here that will make their way in as a separate PR.

msavva:
Functionality in this PR has been cleanly merged in separate PRs (mostly #133 and #166).  Closing this PR.

PR #56, opened 2019-05-28
Example of adding additional actions

erikwijmans:
## Motivation and Context

We need more examples!

This example shows how introduce new control functions, register them, and then use them as actions for the agent!

## How Has This Been Tested

Ran the script and states look correct.  Would dumping images be useful?


examples/new_actions.py
```diff
@@ -0,0 +1,199 @@
...
+the following signature
+
+def my_new_control(scene_node: habitat_sim.SceneNode, actuation_spec: habitat_sim.ActuationSpec):
+    pass
+
+The scene_node is what the control function manipulates (or controls) and the actuation_spec
+contains any parameters needed by that control function.
+See habitat_sim/agent/default_controls.py for example control functions
+
+Control functions are registered use the habitat_sim.agent.controls.register_move_fn function.
```
msavva: typo `use` -> `using`

erikwijmans:
Landing.

Issue #54, opened 2019-05-28
Importing a mesh into Habitat, what 3D formats are supported?


medhagitnod:
Is it possible to import or create a 3D scene in Habitat from an .obj and .mtl file of the space so that the mesh can be explored using Habitat-sim agents?

msavva:
You can use tools such as [assimp](https://github.com/assimp/assimp) to package your obj+mtl format mesh to glTF v2 binary format which can be used with Habitat.  A rough outline of steps for a given scene mesh `<scene>.obj`:
1. `brew install assimp` on MacOS, or `apt-get install assimp` etc. in Linux
1. `assimp export <scene>.obj <scene>.glb -embtex`
1. Compile Habitat and use data tool to precompute navmesh for scene: `build/utils/datatool/datatool create_navmesh <scene>.glb <scene>.navmesh`
1. Run `examples/example.py --scene <scene>.glb` or `build/viewer <scene>.glb` (NOTE: this assumes the `<scene>.navmesh` file is in the same directory as `<scene>.glb`

You can also check that the glb mesh is packaged correctly by loading it into a viewer such as https://gltf-viewer.donmccurdy.com .  Hope the above information is helpful.  Feel free to follow up here if you encounter any trouble, or just let us know if all works as expected.

medhagitnod:
When I attempt to load the glb mesh into the viewer linked, I receive the following message: THREE.GLTFLoader: Legacy binary file detected. Use LegacyGLTFLoader instead. 
Will this affect my ability to load the mesh into Habitat? If so, what can I do to ensure the glb is packaged correctly?

msavva:
The output might have been exported as a glTF v1 format as opposed to v2 format.  You can force v2 by calling `assimp export <scene>.obj <scene>.glb -fglb2 -embtex` instead.

edbeeching:
I have had similar issue with conversion with the assimp package. I found the [obj2gltf](https://github.com/AnalyticalGraphicsInc/obj2gltf) package to work out of the box. 

usage: obj2gltf -i house.obj -o house.glb

mosra:
[...]

PR #57, opened 2019-05-29
Use AnySceneImporter for loading arbitrary formats

mosra:
## Motivation and Context

As mentioned in #54, current code had hardcoded only support for glTF files and in order to import other formats, users were forced to convert them to glTF. That's making users' life unnecessarily hard -- we're using Assimp anyway, so why not use it for general data as well. This PR enables effortless import of "any" formats using [AnySceneImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1AnySceneImporter.html) -- FBX, Blend, ..., basically anything it can recognize (and load a plugin for).

On the buildsystem side it's just re-enabling AnySceneImporter and AssimpImporter that were commented out right now, plus making the Assimp finding & linking a bit more robust.

Some formats (such as OBJ or PLY) don't provide scene hierarchy, in that case the code takes a wild guess and tries to import the first found mesh with the first found material.

## How Has This Been Tested

Built and tested with both system and embedded Assimp.

Loading PLY, OBJ and various other formats in the viewer utility, explicitly verified also with files that have no meshes and/or no materials, however I'm a bit unsure about the bounds checks -- let me know if these are okay.

## Types of changes

## Checklist


src/esp/assets/ResourceManager.cpp
```diff
@@ -475,8 +484,20 @@ bool ResourceManager::createScene(Importer& importer,
     for (auto objectID : sceneData->children3D()) {
       createObject(importer, info, metaData, sceneNode, drawables, objectID);
     }
+  } else if (importer.mesh3DCount() && meshes_[metaData.meshIndex.first]) {
+    // no default scene --- standalone OBJ/PLY files, for example
+
+    // create scene parent node with transformation aligning to global frame
+    auto& sceneNode = parent.createChild();
+    const quatf transform =
+        quatf::FromTwoVectors(info.frame.front(), geo::ESP_FRONT);
```
msavva: This should probably be `quatf::FromTwoVectors(info.frame.up(), geo::ESP_UP);` instead. We explicitly define the up vector for most of our assets, but leave front to the default in many cases.

mosra: This is verbatim copied from the branch above (see line 480), should I change both?

mosra:
> in the changes I've been making in #53 I've actually separated out assimp as a dependency as it was only used for building the datatool binary

@msavva actually, checking the SceneLoader.cpp, I propose rewriting the code to use AnySceneImporter as well instead of Assimp directly -- it only does manual mesh data import + fixing up the gravity orientation, right? Then we're depending on Assimp only as an (optional) plugin.

> OTOH, maybe we ought to try out building assimp on emscripten anyways. :smile:

I'm not saying it's impossible, but once it's just a plugin it's easy to replace it with dedicated importers for just the formats we want to support on the web (for example PLY, OBJ and glTF), saving *several megabytes* (the native binary is 8.5 MB on my machine, for reference). Assimp has a lot of functionality we don't need at all, it's not just import but also export and several complex tools for mesh processing.

msavva:
> > in the changes I've been making in #53 I've actually separated out assimp as a dependency as it was only used for building the datatool binary
> 
> @msavva actually, checking the SceneLoader.cpp, I propose rewriting the code to use AnySceneImporter as well instead of Assimp directly -- it only does manual mesh data import + fixing up the gravity orientation, right? Then we're depending on Assimp only as an (optional) plugin.
> 
> > OTOH, maybe we ought to try out building assimp on emscripten anyways. 😄
> 
> I'm not saying it's impossible, but once it's just a plugin it's easy to replace it with dedicated importers for just the formats we want to support on the web (for example PLY, OBJ and glTF), saving _several megabytes_ (the native binary is 8.5 MB on my machine, for reference). Assimp has a lot of functionality we don't need at all, it's not just import but also export and several complex tools for mesh processing.

Sounds good on both counts!

src/esp/assets/ResourceManager.cpp
msavva: Ah! Let's leave the current behavior in place for both branches but add a "TODO ought to align with ESP_UP" above them.

mosra: #61 by @erikwijmans fixes that, so I propose merging that one first -- and then I update the above to be the same as done in #61.

mathfac:
If this PR will be rebased on the latest master we will get proper CI testing.

mosra:
Rebased on master, adapted to the change from #61.

msavva:
@mosra now that this PR is rebased and passed through CI, it should be ready to land on master and take advantage of the additional format support. 🙂

Please squash-and-merge into the master branch.

PR #61, opened 2019-05-30
Fix coordinate transforms

erikwijmans:
## Motivation and Context

#57 raised that the transform on the meshes is weird!  This is actually a bug that, due to serendipity, actually works.

This PR fixes this by first fixing `CoordinateFrame::rotationWorldToFrame`, using that to implement a `CoordinateFrame::rotationFrameToWorld`, then using that to rotate the mesh to ESP coordinates.

CC: @mosra (no need for review, just tagging for visibility)

## How Has This Been Tested

Via pytest (rendering tests would complain) and via viewer

## Types of changes


mosra:
OK, so once this is in I'll do the corresponding change in #57.

src/esp/geo/CoordinateFrame.h
```diff
@@ -40,6 +40,10 @@ class CoordinateFrame {
   //! to coordinates in this CoordinateFrame
   quatf rotationWorldToFrame() const;
 
+  //! Returns quaternion representing rotation taking world coordinates
+  //! to coordinates in this CoordinateFrame
```
msavva: Perhaps clearer to say `rotation taking direction vectors in world coordinates to direction vectors in this CoordinateFrame` (since applying just this quat to world coordinates without accounting for translations doesn't account for the full `world coordinates to CoordinateFrame coordinates` transformation)

src/esp/assets/Asset.cpp
```diff
@@ -25,6 +25,8 @@ AssetInfo AssetInfo::fromPath(const std::string& path) {
   } else if (endsWith(path, ".glb")) {
     // assumes MP3D glb with gravity = -Z
     info.type = AssetType::MP3D_MESH;
+    // Create a coordinate from for the mesh by rotating the default ESP
```
msavva: typo `from for -> for`

erikwijmans:
Issues addressed

msavva:
> Issues addressed

Looks great -- let's land it!

PR #64, opened 2019-05-31
Moving to Magnum Python bindings

mosra:
## Motivation and Context

Moves away from AttachedObject / SceneNode APIs to Magnum's own [python bindings](https://doc.magnum.graphics/python/). Opening a draft PR so we can discuss the API changes as they gradually appear. **Early feedback welcome.**

Things to do:

- [x] Make this work with system-wide installed `magnum-bindings`
  - [x] Make this work with `magnum-bindings` as a subproject (how to install these / merge the two `setup.py`?)
- [x] Replace Eigen-based wrappers for transformation APIs in AttachedObject and SceneNode with APIs directly from `Magnum::SceneGraph`
  - [x] Adapt the Python code to the above
  - [x] Publish minor updates to `magnum` and `magnum-bindings` needed to achieve the above
- [x] Cionvert AttachedObject to be a SceneGraph feature
  - [x] Adapt the Python code to the above
  - [x] ~~Rename to Feature or something like that?~~
  - [x] Get completely rid of it? ~~It defines a type, can't~~
- [x] Agree on API consistency (where to draw the line between using Magnum math and numpy/Eigen) -- will be done gradually over next PRs to keep this change from growing beyond reasonable bounds
- [x] Agree on `using namespace Magnum` / `from magnum import *` vs `namespace Mn = Magnum` / `import magnum as mn` -- will be done gradually over next PRs to keep this change from growing beyond reasonable bounds

## How Has This Been Tested

Note: in the current state, it won't compile on your side, since there is no Homebrew package for magnum-bindings and the in-project integration is not done yet.

- [x] CI succeeds for habitat-sim
- [ ] CI fails for habitat-api


src/esp/sensor/Sensor.cpp
```diff
@@ -47,12 +49,12 @@ void Sensor::setTransformationFromSpec() {
...
 
-  resetTransformation();
+  node().resetTransformation();
 
-  translate(spec_->position);
-  rotateX(spec_->orientation[0]);
-  rotateY(spec_->orientation[1]);
-  rotateZ(spec_->orientation[2]);
+  node().translate(Magnum::Vector3(spec_->position));
+  node().rotateX(Magnum::Rad(spec_->orientation[0]));
```
erikwijmans: I believe these should be in degrees

mosra: I actually had to look twice to be sure -- the (now deleted) `AttachedObject::rotateX()` had a parameter `angleInRad` and there was another `rotateXInDegree()` (which is not used here).

erikwijmans:
Just wanted to say that 
image 
is a great thing to see :)

mathfac:
If this PR will be rebased on latest master we will get proper CI testing.

mosra:
Almost 1000 lines deleted now -- the `SceneNode` and `AttachedObject` are now just tiny objects -- one managing the node ID and the other managing the attached object type. All tests are passing for me except one which fails for me on master as well:

```
E       AssertionError: Incorrect semantic_sensor output
E       assert 13581879105.047745 < (0.015 * 62022.41997052356)
```

Feedback on the overall direction appreciated, I still need to rebase on master to fix conflicts with the recently merged Greedy Follower (and of course bundle the bindings).

erikwijmans:
Interesting that it fails on master. Do you have the buggy version on MP3D (we fixed that in early April and it required you to re-download the dataset)?

habitat_sim/simulator.py
```diff
@@ -73,18 +75,14 @@ def reconfigure(self, config: Configuration):
...
             return
 
-        for agent in self.agents:
-            agent.detach()
-
         # NB: Configure backend last as this gives more time for python's GC
         # to delete any previous instances of the simulator
-        self._config_agents(config)
+        # TODO: can't do the above, sorry -- the Agent constructor needs access
+        # to self._sim.get_active_scene_graph()
```
msavva: Doesn't seem to be a problem to me.  In any case, this is just eagerness on our end to make it easier for python GC, and there is no correctness or stability concern AFAIK, right @erikwijmans ?  So we can just remove the note above.

habitat_sim/agent/agent.py
```diff
@@ -99,12 +100,18 @@ class Agent(object):
         graph in python is dangerous due to differences in c++ and python memory management
     """
 
+    # TODO: is there a way to convince attrs to construct `body` from a
+    # `SceneNode` argument in the constructor? This works around that by saving
+    # scene node into an attribute and then passing that to AttachedObject
+    # constructor in a post-init. UGLY.
```
erikwijmans: I am not aware of one, but you can add `init=False` to the `@attr.s(...)` (https://www.attrs.org/en/stable/api.html#attr.s) and then implement `__init__` method yourself!

habitat_sim/simulator.py
erikwijmans: There is a stability concern.  I have seen crashes due to trying to having two OpenGL contexts in the same process when rapidly creating and delete instances of the simulator.  Python's GC doesn't get around to cleaning up the old instance before we try to create a new instance.

habitat_sim/agent/agent.py
mosra: I tried a few attempts at providing my own `__init__`, but failed miserably -- though I only tried to add `init=False` to the `body` `attrib`, not to the others. I never used attrs before, but if I understand correctly, when I have my own `__init__`, I need to manually init all the attributes, right?

habitat_sim/simulator.py
mosra: Do we *need* to recreate the GL context, tho? I think it would make sense to keep it around all the time and recreate only the scene and GL objects.

mosra:
@msavva re `SceneNode` / `AttachedObject` -- I spent quite some time thinking about this, but couldn't come up with a nice solution. As I see it based on the code:

- the `SceneNode` contains an ID of *the node*, which is then used by the shader to provide an object ID.  The ID AFAICT maps 1:1 to object ID from importer. Because not every `SceneNode` has a corresponding `AttachedObject`, I realized I couldn't move this property to `AttachedObject` and get rid of `SceneNode`.
- the `AttachedObject`, which is now a feature attached to some object, not an object itself, contains a 
"feature type". Again because not every `SceneNode` has a corresponding `AttachedObject`, I couldn't move this property to `SceneNode` and get rid of `AttachedObject`.

My question is -- what all nodes need to have the ID and for what? If it's mainly used to annotate drawn objects in the shader, then the ID could be moved to the drawable implementation and then we don't need `SceneNode` anymore. Does that make sense?

Also, I didn't do any renaming of `SceneNode` / `AttachedObject` yet (to keep the diff small) -- now if we get rid of `SceneNode`, what do you think could be the new name for `AttachedObject`? Just `Feature`? Something more descriptive?

habitat_sim/agent/agent.py
erikwijmans: There are two places you can provide `init=False`.  The first is on the attributes themselves,

```python
@attr.s(auto_attribs=True)
class MyClass:
    my_attrib: int = attr.ib(init=False, default=5)


print(MyClass())
# MyClass(my_attrib=5)

MyClass(6)
# Traceback (most recent call last):
#   File "tmp.py", line 11, in <module>
#     MyClass(6)
# TypeError: __init__() takes 1 positional argument but 2 were given
```
In this case, attrs will still write `__init__` and initialize `my_attrib`, but `my_attrib` won't be an argument in `__init__`

The second is on the decorator
```python
@attr.s(init=False, auto_attribs=True)
class MyClass:
    my_attrib: int

    def __init__(self, my_attrib_value_getter):
        self.my_attrib = my_attrib_value_getter()


print(MyClass(lambda: 12))
# MyClass(my_attrib=12)
```
In this case, `attrs` will not write an `__init__` method and so you can do whatever you want it :).  It also won't initialize or create any attributes (even if you provide them a default value, i.e. if I didn't set `self.my_attrib` above, you'd get an `AttributeError: 'MyClass' object has no attribute 'my_attrib'` if you tried to access it), but you'll still get all the other niceties of attrs like print methods!

habitat_sim/simulator.py
erikwijmans: Yeah, it is currently a super-hacky way to control GPU memory usage.  Whenever we swap a scene, we just kill the entire simulator backend as that dump all assets off the GPU :)

erikwijmans: Could we use a singleton for the OpenGL context?  That way the context will survive through destruction of the simulator?

mosra: Singleton as in making the `WindowlessContext` independent from the `Simulator` (instead being protected there) and assigning it to a global `gl_context = WindowlessContext()`, for example? Yes that would work.

> it is currently a super-hacky way to control GPU memory usage

If you delete the whole `Simulator` instance every time and keep the GL context alive, it will cascade-delete all GL objects anyway -- and that's a perfectly valid way to clean up. Not "super hacky" at all :)

Is this a blocking issue for this PR (will it cause crashes in regular usage)? I have [platform.{glx,egl}.WindowlessApplication](https://doc.magnum.graphics/python/magnum.platform.egl.WindowlessApplication.html) exposed in Python, but currently without the additions you have in WindowlessContext -- would need to integrate those first.

erikwijmans: If we can't rapidly delete and recreate instances of the simulator, this will causes crashes in regular uses, habitat-api kills and recreates the simulator whenever the scene swaps, this is the hacky part :).

We could do the singleton in C++ via something like `WindowlessContext::Instance()` (would have to check to see how that works with threads and multi-processing).

There are also other ways to handle this, like forcing python GC to run, or having an explicit "close" method on the c++ side instead of letting the destructor handle it (as then we won't be reliant on the whims of python GC).

Let's definitely keep pushing this along and if it does cause crashing issues, we will figure it out!

mosra: Having the GL context "singleton" / global on the Python side is not an option (`gl_context = WindowlessContext()` as I mentioned above)? That feels the most clean to me.

erikwijmans: Yeah, a python Singleton would work also 🙂

habitat_sim/agent/agent.py
mosra: Thank you, works great!

mosra:
Ok, so I'm *almost* there:

- :white_check_mark: rebased on current `master` and adapted to changes related to Greedy Follower
- :white_check_mark: the `magnum-bindings` repo can now build a single `*.so` for the corrade/magnum bindings if the libs are built statically
- :white_check_mark: it's integrated as a submodule and built together with the other things
- :white_check_mark: it's integrated into `setup.py`, although in a bit interesting way
- :white_check_mark: the CI can build and install it, and the `sim tests` pass
- :red_circle: but the `api tests` abort, so I need to look at that -- is there any way I could get more info from the CI run than just "aborted"? I suspect it's due to the GL context issues we discussed above :sweat_smile: 
- :red_circle: also the above unticked boxes and discussion about `AttachedObject` / `SceneNode` are not resolved yet

habitat_sim/agent/agent.py
```diff
@@ -99,12 +100,22 @@ class Agent(object):
...
-    def __attrs_post_init__(self):
+    agent_config: AgentConfiguration
+    sensors: SensorSuite
+    controls: ObjectControls
+    body: hsim.AttachedObject
+
+    def __init__(
+        self,
+        scene_node: hsim.SceneNode,
+        agent_config=AgentConfiguration(),
```
erikwijmans: Fix `agent_config=AgentConfiguration()` anti-pattern

mosra:
Latest update:

- the `AttachedObject` is now fully gone, type is moved to `SceneNode` as we agreed on (1.1k lines removed now, yay!)
- all features now have a `node()` / `node` accessor returning `SceneNode`
- no magnum types exposed to habitat-api

however -- every time I seem to run into another problem that explodes into a billion additional tasks and it's kinda never ending :sweat_smile: :
 
- I'm not rebased on master yet, because the linters on the CI have a different idea about how python imports should be sorted compared to the precommit hooks on my machine (will need to sort that out)
- on master, the CI is failing with assertion inside setuptools (see https://circleci.com/gh/facebookresearch/habitat-sim/255) -- apparently it doesn't like how I'm recursively calling setup.py from magnum-bindings (why wasn't it failing before? suggestions to do this better?)
- I temporarily switched to `pytest -s` in `.circleci.yml` to investigate why it aborts and it doesn't abort anymore but rather fails at random places due to missing dependencies or *somehow* giving errors in code that's no longer there (see https://circleci.com/gh/facebookresearch/habitat-sim/263 and the error in `_strafe_impl`) ... is it all caused by me switching from `python setup.py test` to `pytest -s`?

For the remaining things we agreed on personally (like rotation representation, moving the rest of the internals to use magnum types etc.) I think it's better to do these in a completely separate PR, as this is already quite big :)

Any help/feedback appreciated.

erikwijmans:
It seems like calling setup.py in setup.py isn't gonna work.  I am not surprised (setup.py is very old and very clunky, but nothing has come about to dethrone it so....)

Ideas:
1. Can we have magnum-python build standalone?  i.e. `pip install src/deps/magnum-bindings/src/python`?
2. We decide for now, no `pip install .` for habitat_sim and just have people do `build.sh` or `python setup.py install` followed by `pip install build/deps/magnum-bindings/src/python`

I'm not in-love with either of these, but eventually we need to start distributing habitat_sim and magnum via `conda` so maybe this will be a reason to start doing that.

erikwijmans:
For `isort` complaining, make sure you are on the latest version of isort, that all the dependencies are installed and `/path/to/habitat_sim` is in your python path.  If you keep having trouble, I am happy to format and push.

erikwijmans:
The test that is failing is on the habitat-api side: https://github.com/facebookresearch/habitat-api/blob/master/examples/new_actions.py#L35

So that is expected to fail for now

erikwijmans:
The tests are now failing as the double context thing seems to have been taken care of, i.e. 
```
import habitat_sim

agent_cfg = habitat_sim.AgentConfiguration()
sim_cfg = habitat_sim.SimulatorConfiguration()
sim_cfg.scene.id = "./data/scene_datasets/habitat-test-scenes/van-gogh-room.glb"
cfg = habitat_sim.Configuration(sim_cfg, [agent_cfg])

for _ in range(100):
    sim = habitat_sim.Simulator(cfg)
    sim.close()
```
works

erikwijmans:
For building, there seems to be an option 3:
Add a `pip install /path/to/build/deps/magnum-bindings/src/python` to the end of setup.py

mosra:
> The tests are now failing as the double context thing seems to have been taken care of

I did numerous changes, but this is *not* one of them :) OTOH, if still needed, a recent commit in Magnum could make such problems a bit easier to fix: https://github.com/mosra/magnum/commit/11a94c16a39c6bf63408171e032c454d5e945db8

> For `isort` complaining

Thanks, `habitat_sim` was *not* in PYTHONPATH and that was making most of the differences locally. I saw your https://github.com/facebookresearch/habitat-sim/commit/b44d27a1d28a1b569629a3e42fa978571a92453c, I did this change by modifying `pyproject.toml` instead, is that okay?

```toml
known_third_party = ["corrade", "magnum", "habitat_sim"]
```

> The test that is failing is on the habitat-api side: https://github.com/facebookresearch/habitat-api/blob/master/examples/new_actions.py#L35

ahhh dang :sweat_smile: the *exact same file* with the exact same `_strafe_impl()` is in the sim repo and there I updated it, so I was *extremely* confused why it is showing the old code again 

> Add a `pip install /path/to/build/deps/magnum-bindings/src/python` to the end of setup.py

Will do that, then -- thanks :+1: Is there any *disadvantage* compared to doing `python /path/to/.../setup.py install`?

erikwijmans:
> Will do that, then -- thanks 👍 Is there any disadvantage compared to doing python /path/to/.../setup.py install?

I don't think so.  I think using `pip install` is better in the case as pip will then track which packages are installed properly (`pip install` uses `setup.py install` internally).

erikwijmans:
> Added `known_third_party = ["corrade", "magnum"]`

Yep! makes sense

erikwijmans:
> I did numerous changes, but this is not one of them :) OTOH, if still needed, a recent commit in Magnum could make such problems a bit easier to fix: mosra/magnum@11a94c1

Fixing the double-context thing properly instead of just hoping python GC wins the race and deletes the old one before we make a new one.  Looking at that commit, seems like it would just be one or two additional lines as we'd need access to the old context?

mosra:
> seems like it would just be one or two additional lines as we'd need access to the old context?

Depends on what all needs to get destroyed -- if it's just the Context instance, `makeCurrent(nullptr)` would allow you to create a new one, while Context's destructor doesn't require itself to be active during destruction, so deleting the old one can then happen at any time.

However if there are random textures, meshes and buffers still present while the old one needs to be destroyed and destruction of those isn't in any way "ordered before" destruction of the old one (for example being members of the same class that owns the Context instance), then you still need the old one to be active while its GL objects are deleted (as the destructors are doing various state tracking cleanup).

erikwijmans:
> However if there are random textures, meshes and buffers still present while the old one needs to be destroyed

Ah, yes, there will be meshes and textures that need to be destroyed and are part of the old context :(

mosra:
OK, then let's go with the global WindowlessContext singleton instead as discussed above? https://github.com/facebookresearch/habitat-sim/pull/64#discussion_r292365015

I would say exposing a (completely opaque) WindowlessContext to Python and requiring users to instantiate it before anything else could do the trick.

habitat_sim/simulator.py
msavva: Agreed.  Allowing a python singleton for the context is a good abstraction imo.  It would be good to make its explicit construction from the python side optional though so that we don't introduce additional boilerplate that is necessary to create a simulator instance.

msavva:
> OK, then let's go with the global WindowlessContext singleton instead as discussed above? [#64 (comment)](https://github.com/facebookresearch/habitat-sim/pull/64#discussion_r292365015)
> 
> I would say exposing a (completely opaque) WindowlessContext to Python and requiring users to instantiate it before anything else could do the trick.

Agree -- one comment though: ideally we can do this while also allowing for a functional simulator backend to be instantiated purely on the C++ side (i.e. without needing to wrap with a Python driver that does the context instantiation). [This comment is motivated by the desire to avoid future "intialization logic" duplication for using hsim without Python (e.g. #53)].

mosra:
As far as I can see, the way @erikwijmans did this in https://github.com/facebookresearch/habitat-sim/commit/3b58fddcc2f77eae5a5436f1fb302cc5c2995cb4 there's no reason why the same couldn't "just work" also on C++ side.

erikwijmans:
> check if magnum is installed when pip install is used and ask use build.sh instead.

People can still do `python setup.py install`.  It is just `pip` that is being a problem child.

PR #67, opened 2019-06-05
Add linting to CI

erikwijmans:
## Motivation and Context

Add linting to CI

## How Has This Been Tested

With the CI!


erikwijmans:
There is one caveat with this:  `clang-format-8` (which is what is on MacOS) gives different results on `clang-format-7` (which is the most recent version available on common ubuntu version), so it is kinda annoying to get formatting correct on linux.

erikwijmans:
Although `clang-format-8` is available on [linuxbrew](https://docs.brew.sh/Homebrew-on-Linux) and via `conda install clangdev -c conda-forge` so simple enough.

erikwijmans:
@mathfac @msavva  does this make sense to add?

PR #70, opened 2019-06-07
Use controls to report if a collision occurred

erikwijmans:
## Motivation and Context

Our current collision logic is a little buggy :( It also requires knowing if an action can results in a collision (i.e. move_forward can result in a collision but turn_left cannot); this problem will only get worse know that we allow people to define their own actions.

This PR seeks to resolve this by having the controls check if a collision occurred by checking to see if the filtering step caused us to move less than expected.

@msavva, thoughts in the API?

## How Has This Been Tested

Visually/with the collision test in the api side.

## Types of changes


habitat_sim/simulator.py
```diff
@@ -139,9 +139,12 @@ def last_state(self):
 
     def step(self, action):
         self._num_total_frames += 1
-        self._default_agent.act(action)
+        collided = self._default_agent.act(action)
         self._last_state = self._default_agent.get_state()
+
         observations = self.get_sensor_observations()
+        observations["collided"] = collided
```
msavva: It would be good to document this added field somewhere

PR #71, opened 2019-06-09
Better docs

erikwijmans:
## Motivation and Context

@mosra took over this from @erikwijmans :scream:! What is done:

- Just the Python part right now, C++ part will follow in a separate PR.
- Integrated everything what was originally in this PR and reformatted for m.css
- Added a config file for m.css python docs generator + intersphinx inventories for crosslinking to dependencies
- Added a submodule and a helper script for generating the docs

What is explicitly *not* done:

- Docs CI setup, as I anticipate *huge* issues with latex there (can someone do this instead? can we build the docs locally for now?)
- A (light) style that fits the rest of the Habitat website. Based on our previous conversation, @msavva promised to take a look :)
- This is a braindead conversion of existing content and the docs are *extremely* outdated in various places. Someone needs to go over everything and fix stuff.

A great way to give feedback would be if you could make it working locally. Links to help get you acquainted with the tool:

- https://mcss.mosra.cz/documentation/python/
- https://mcss.mosra.cz/plugins/sphinx/

## How Has This Been Tested

Sample is at https://tmp.mosra.cz/habitat-sim/ .


msavva:
Vehemently agree @erikwijmans! 👍 Thanks for kicking this off.

I was actually just now investigating a small fix to get the doc build setup working on sphinx 2.1.0 (we used to build on 1.7.6) -- will push it shortly.

mosra:
More docs is always better! :+1: 

@msavva Are we still planning to eventually generate both c++ and python docs using [m.css](https://mcss.mosra.cz/documentation/) (like magnum has)? In that case I could jump on this right after I'm done with #64 -- the ultimate goal is full compatibility with Sphinx so ideally everything that's written now could be used as-is, but realistically we *might* need to adapt a few things here and there. So I think the sooner I get on this, the less we would need to rewrite/adapt.

mathfac:
@msavva, @erikwijmans lets refresh this PR and merge it. As was discussed @msavva can add a high level description of Habitat Sim design and components. Maybe, we can re-use `A. Habitat Platform Details` chapter from [the technical report](https://arxiv.org/pdf/1904.01201.pdf). Then we can divide work into smaller chunks.

msavva:
> @msavva, @erikwijmans lets refresh this PR and merge it. As was discussed @msavva can add a high level description of Habitat Sim design and components. Maybe, we can re-use `A. Habitat Platform Details` chapter from [the technical report](https://arxiv.org/pdf/1904.01201.pdf). Then we can divide work into smaller chunks.

Thanks for pinging this PR alive @mathfac 🙂Yes, let's get this into a mergeable state soon.  I will likely have some cycles later this week to sketch out barebones docs for the main elements on the hsim side. @mosra -- what's a good reference point for the raw rst doc style? Or should I just mirror usage in Magnum? 🙂

mosra:
I'd say follow the Sphinx docs for now (here's [function docs](http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists)) and for now focus mainly on the high-level prose content instead of documenting particular properties or methods. From the other side I'll be gradually adding Sphinx-compatible features and we'll meet somewhere in the middle. 

I was actually able to build *some* auto-generated docs just through module inspection, so I'll continue iterating on that. Looking at the documentation you have here, it's actually quite close to what the parser would expect. One thing it will *definitely* do for you is the type annotation -- there's no need to do that manually in the docs.

If you want a reference from Magnum, here's the python rst-formatted docs: https://github.com/mosra/magnum-bindings/tree/master/doc/python ... and for the C++ part, look at any header in Magnum.

mosra:
@erikwijmans do you have any idea why python lint is failing here while in #105 (which this is based off of) it's completely happy? I didn't change any CI config, the only *vaguely* relevant change is in `.editorconfig` but I doubt it would be looking there :)

erikwijmans:
Looks like `isort` is is not parsing the `[*.{py,rst}]` and is doing indent of 2

mosra:
Blargh, okay, I'll format the config differently. Didn't expect it would look there as well, isn't it doing a bit too much?

mosra:
Pushed a new version with fixed conflicts against current master. There's a new commit showcasing a bit more advanced connection between code and docs for [d3_40_colors_hex](https://tmp.mosra.cz/habitat-sim/habitat_sim.utils.html#d3_40_colors_hex) -- primarily to show possibilities, I'm not saying it *should* be in the docs like this :wink: 

```rst

.. py:data:: habitat_sim.utils.d3_40_colors_hex
    :summary: Color map for semantic ID rendering.

    Same as `d3_40_colors_rgb`, but in a hexadecimal representation.

    .. include:: ../habitat_sim/utils.py
        :code: py
        :start-after: # [d3_40_colors_hex]
        :end-before: # [/d3_40_colors_hex]
        :filters: string_hex_colors

```

image

Besides that:

- the submodule is now in `docs/m.css` to not make it seem like it's needed for using the project itself
- it should no longer require Pelican to run (@erikwijmans can you confirm?)
- the output is placed into `build/docs/habitat-sim` instead of `build/doc` so I can put the C++ docs next to it in the immediately following PR

PR #73, opened 2019-06-11
Tests: fix a "variable referenced before assignment" error

mosra:
## Motivation and Context

I Ctrl-C'd while running the `test_greedy_follower` test and instead of `KeyboardInterrupt` it ended up with a "variable referenced before assignment" error on the line with `raise e`. This fixes it simply by copying it from the similar code below.

## How Has This Been Tested

Everything compiles and runs, tests pass and do the expected thing during Ctrl-C.

## Types of changes


erikwijmans:
Thank you!

PR #85, opened 2019-06-24
Support no sensors

erikwijmans:
## Motivation and Context

Sometimes we don't need sensors (generating datasets for instance), so it is nice to not need to load-up the mesh and get a rendering context.  This PR allows the simulator to be configured without any sensors.

## How Has This Been Tested

I adapted the test for the semantic annotations to use its own instance of the simulator loaded up with no sensors -- this ensures that the semantic scene graph is also loaded.  This test would crash due to having two OpenGL contexts if the no-sensor thing didn't work :)

### DIFF
--- a/tests/test_sensors.py
+++ b/tests/test_sensors.py
@@ -98,6 +98,9 @@ def test_sensors(scene, has_sem, sensor_type, sim, make_cfg_settings):
     ) < 1.5e-2 * np.linalg.norm(gt.astype(np.float)), f"Incorrect {sensor_type} output"
 
 
+# Tests to make sure that no sensors is supported and doesn't crash
+# Also tests to make sure we can have multiple instances
+# of the simulator with no sensors
 def test_smoke_no_sensors(make_cfg_settings):
     sims = []
     for scene in _test_scenes:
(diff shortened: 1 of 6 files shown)


erikwijmans:
I put the smoke-test for the no-sensor thing there as there is no other non-rendering test that touches the cpp code not related to rendering.

erikwijmans:
Added a simple smoke test for no sensors.  I've also been using this branch to train locally, so I am confident it is stable :)

tests/test_sensors.py
```diff
@@ -96,3 +96,17 @@ def test_sensors(scene, has_sem, sensor_type, sim, make_cfg_settings):
     assert np.linalg.norm(
         obs[sensor_type].astype(np.float) - gt.astype(np.float)
     ) < 1.5e-2 * np.linalg.norm(gt.astype(np.float)), f"Incorrect {sensor_type} output"
+
+
+def test_smoke_no_sensors(make_cfg_settings):
```
msavva: Probably good to add a brief comment here to explain the interpretation of `test_smoke_no_sensors` (i.e. that the "no sensor" configuration should be a valid, non-crashing configuration), otherwise the test function naming is a little bit confusing.

PR #91, opened 2019-06-28
Bullet Physics Integration #1

hzyjerry:
## Motivation and Context

- First PR towards physical interactions with Habitat 😊
- **Early feedbacks are severely welcomed** 🎉
- Laying the foundation for upcoming development (articulated agents & objects, robotics, etc) requires moderate refactoring of existing habitat code and data organizations.

## Types of changes

- Introduced `esp/assets/PhysicsManager` (similar to `esp/assets/ResourceManager` in style) to track & manage states of physics simulation, such as bulletWorld, FPS, timestep, objects etc
- Introduced `esp/physics/BulletObject` as an extended SceneNode. BulletObject interacts with PhysicsManager to create collision mesh, set object properties, etc.
- [x] Supporting APIs: generic object state update, instancing
    - [x] `loadObject(xyz, rotation)`
    - [x] `setObject(ID, xyz, rotation)`
    - [x] `removeObject(ID)`
    - [x] Move functions from `viewer.cpp` to habitat core
    - [x] Support object instancing by ID
- [x] Testing

## Challenges 💥
- File formats: getting GLB/obj/ply objects to all work correctly (E.g. Bullet mesh uses `PHY_INTEGER`, incompatible with `Magnum::UnsignedInt`)
- Object: requires preprocessing before we can use them
- Speed: physics simulation slows down significantly when there are >20 objects
- Physical metadata: how to store physical properties (friction, restitution, mass)

## Checklist

- [x] Magnum mesh -> Bullet mesh object Interface
    - [x] Integrate Physics Loader based on Magnum [Issue](https://github.com/mosra/magnum-integration/issues/20)
    - [x] Load GLB object/scene composite mesh
    - [x] Load FRL object/scene composite mesh (**bullet can handle <= 1.5M triangles**)
    - [x] Load PLY object/scene composite mesh
- [x] Update habitat object loader to be more physics-ready
    - [x] Currently GLB/FRL/PLY loaders work differently. A uniform `getMeshData()` function can better interface with PhysicsManager to create collision mesh
- [x] Use smart pointers
- [x] Simple object Interactions
    - [x] Apply force by object id
    - [x] Apply impulse by object id (optional)
- [x] Prepare demo script
    - [x]  `View.cpp` with `--enbable-physics` option
    - [x]  `View.cpp` backward compatible
- [x] Physical Object management in `PhysicsManager` and `ResourceManager`
    - [x] Refactoring: separate `loadObject` and `addObject`
    - [x] ~Support cloning object by cloning sceneNode~ (decided to push to next PR)
    - [x] Test and support instancing
- [x] JSON Config files
    - [x] Object config for: friction, mass, restitution, center of mass
    - [x] Scene config for: list of objects, scene name, scene layout
- [x] Prepare object meshes
    - [x] Pre-processing: we require all objects to be convex-decomposed. Provide preprocessing script for end users.
    - [x] Provide starter object mesh files

## PR Milestones
[July 2 GLB Scene](https://www.dropbox.com/s/ndqneibzhs2w087/Habitat-Object-PR91%2301.mov)
[July 4 Surreal Scene](https://www.dropbox.com/s/4lba2jwi0mtn69x/Habitat-Object-PR91%2303%2025%25.mov)
[July 5 Replica Scene](https://www.dropbox.com/s/bcc7yi4tfg55ngq/Habitat-Object-PR91%2304%2050%25.mov)
Tested on Gibson & MP3D
Screen Shot 2019-08-02 at 10 39 16 AM


src/esp/assets/FRLInstanceMeshData.h
```diff
@@ -43,10 +50,19 @@ class FRLInstanceMeshData : public GenericInstanceMeshData {
...
   virtual Magnum::GL::Mesh* getMagnumGLMesh() override;
 
+  const std::vector<vec3f> get_vbo();
+  const std::vector<int> get_ibo();
+
  protected:
   std::vector<vec4f> cpu_vbo;
   std::vector<vec3uc> cpu_cbo;
 
+  std::vector<vec3f>* cpu_vbo_3   = nullptr;
```
erikwijmans: We name class members with and `_`, i.e. `cpu_vbo_3_` as it makes the implementation code more readable (I see that that wasn't followed for the existing implementation here, oops).

src/esp/assets/FRLInstanceMeshData.h
```diff
@@ -43,10 +50,19 @@ class FRLInstanceMeshData : public GenericInstanceMeshData {
...
+  std::vector<uint32_t>* tri_ibo  = nullptr;
+  //std::vector<int>* tri_ibo_i     = nullptr;
+  std::vector<float>* cbo_float   = nullptr;
```
erikwijmans: All these should be unique pointers or shared pointers (depending on if they are needed by other classes or not).
That way you don't have to delete them yourself.

src/esp/assets/PhysicsManager.h
```diff
@@ -0,0 +1,118 @@
...
+  //! instances have to remove themselves from it on destruction
+  Magnum::BulletIntegration::DebugDraw      _debugDraw{Magnum::NoCreate};
+  btDbvtBroadphase*                         _bBroadphase;
+  btDefaultCollisionConfiguration*          _bCollisionConfig;
+  btCollisionDispatcher*                    _bDispatcher;
+  //! btCollisionDispatcher*                    _bDispatcher;
+  btSequentialImpulseConstraintSolver*      _bSolver;
+  //! btDiscreteDynamicsWorld                   
+  //      _bWorld{&_bDispatcher, &_bBroadphase, &_bSolver, &_bCollisionConfig};
+  btDiscreteDynamicsWorld*                  _bWorld;
```
erikwijmans: Same comment about unique/shared pointers and naming scheme for all the things here.

src/esp/assets/ResourceManager.cpp
```diff
@@ -58,6 +63,231 @@ bool ResourceManager::loadScene(const AssetInfo& info,
...
+          continue;
+        }
+        meshGroup.push_back(&meshData);
+      } 
+
+      // PLY Instance mesh
+      else if (info.type == AssetType::INSTANCE_MESH) {
+        GenericInstanceMeshData* insMeshData = 
+            dynamic_cast<GenericInstanceMeshData*>(meshes_[mesh_i].get());
+        quatf quatf = quatf::FromTwoVectors(info.frame.front(), geo::ESP_FRONT);
```
erikwijmans:
```suggestion
        quatf R_world_frame = info.frame.rotationFrameToWorld();
```

Naming a quatf `quatf` is very scary :)

src/esp/assets/ResourceManager.cpp
```diff
@@ -58,6 +63,231 @@ bool ResourceManager::loadScene(const AssetInfo& info,
...
+
+  return meshSuccess;
+}
+
+bool ResourceManager::loadObject(const AssetInfo& info,
+                                 PhysicsManager& _physicsManager,
+                                 scene::SceneNode* parent,
+                                 bool attach_physics, /* = false */
+                                 DrawableGroup* drawables /* = nullptr */,
+                                 physics::BulletRigidObject** physNode) {
```
erikwijmans: This hack that we did to get it working is scary and shouldn't get to live :).  Returning an Optional<scene::SceneNode*> is the way to do this correctly.  `attach_physics` will get to decided if the returned node is a pointer to a `physics::BulletRigidObject` that was down-casted or a pointer to a SceneNode.

src/esp/assets/ResourceManager.cpp
```diff
@@ -58,6 +63,231 @@ bool ResourceManager::loadScene(const AssetInfo& info,
...
+}
+
+void ResourceManager::shiftMeshDataToOrigin(GltfMeshData* meshDataGL) {
+  Magnum::Trade::MeshData3D& meshData = *(meshDataGL->getMeshData());
+  float minX = 999999.9f;
+  float maxX = -999999.9f;
+  float minY = 999999.9f;
+  float maxY = -999999.9f;
+  float minZ = 999999.9f;
+  float maxZ = -999999.9f;
```
erikwijmans: This can be done using Magnum's eigen intergation with the following pattern:
```
Eigen::Array3f min = <whatever magnum's method is>(meshData.positions(0)[0]).array()
...
for ()
    min = min.cwiseMin(<whatever magnum's method is>(meshData.positions(0)[vi]).array())
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -227,15 +461,27 @@ bool ResourceManager::loadGeneralMeshData(const AssetInfo& info,
     MeshMetaData metaData;
     loadTextures(*importer, &metaData);
     loadMaterials(*importer, &metaData);
-    loadMeshes(*importer, &metaData);
+    loadMeshes(*importer, &metaData, shiftOrigin);
```
erikwijmans: Is the origin shift needed?  That will probably break all of our current episode datasets.  We will also have to make sure we propagate the new origin to the annotations for MP3D.

src/esp/assets/ResourceManager.cpp
mosra: Or just use Magnum and do it in an oneliner :P 

```cpp
std::pair<Vector3, Vector3> bounds = Math::minmax<Vector3>(meshData.positions(0));
```

Note you may need to include `<Magnum/Math/FunctionsBatch.h>` and `<Corrade/Containers/ArrayViewStl.h>` for this to work.

erikwijmans: Fancy :)

src/esp/physics/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+)
+
+
+## Enable physics profiling
+#add_compile_definitions(BT_ENABLE_PROFILE=0)
+#add_definitions(-DBT_ENABLE_PROFILE)
+
+set_directory_properties(PROPERTIES CORRADE_USE_PEDANTIC_FLAGS ON)
+
+if(OpenMP_CXX_FOUND)
```
bigbike: Are you using OpenMP?

src/esp/assets/ResourceManager.h
```diff
@@ -74,7 +97,7 @@ class ResourceManager {
   bool createScene(Importer& importer,
                    const AssetInfo& info,
                    const MeshMetaData& metaData,
-                   scene::SceneNode& parent,
+                   scene::SceneNode& sceneNode,
```
bigbike: Why do you need to rename it?

src/esp/assets/FRLInstanceMeshData.h
```diff
@@ -55,6 +69,8 @@ class FRLInstanceMeshData : public GenericInstanceMeshData {
 
   // Gravity direction of the mesh, this is a STATIC
   vec3f orig_gravity_dir;
+
+  void compute_buffer();
```
bigbike: computeBuffer();

Also, would you please leave a comment on what you are doing in this function.

src/esp/assets/PhysicsManager.cpp
```diff
@@ -0,0 +1,266 @@
...
+  physObject->syncPose();
+  return objectSuccess;
+}
+
+
+bool PhysicsManager::isMeshPrimitiveValid(CollisionMeshData& meshData) {
+  if (meshData.primitive != Magnum::MeshPrimitive::Triangles) {
+    if (meshData.primitive == Magnum::MeshPrimitive::Lines) {
+      LOG(INFO) << "Primitive Lines";
+    }
```
bigbike: please use "else"
Or consider using "switch".

src/esp/assets/PhysicsManager.h
bigbike: I agree with Erik.
I saw you were managing the memory by your own code.
If it is not time critical (meaning the overhead of using smart pointers can be ignored), use the smart pointer (here, it should be unique pointers) to keep the system maintainable.

src/esp/assets/PhysicsManager.h
```diff
@@ -0,0 +1,97 @@
...
+
+  void getPhysicsEngine();
+
+  //! Check if mesh primitive type is valid for bullet physics engine
+  bool isMeshPrimitiveValid(CollisionMeshData& meshData);
+
+  //! ==== physics engines ====
+  //! The world has to live longer than the scene because RigidBody
+  //! instances have to remove themselves from it on destruction
+  Magnum::BulletIntegration::DebugDraw      _debugDraw{Magnum::NoCreate};
```
bigbike: See the coding convention.
make the underscore at the end.
debugDraw_
The same logic applies to any following member variables.

src/esp/assets/MeshData.h
```diff
@@ -7,23 +7,70 @@
...
   //! Vertex colors
-  std::vector<vec3f> cbo;
+  std::vector<vec3f>    cbo;
   //! Index buffer
   std::vector<uint32_t> ibo;
 };
 
+//! Reference to vertices and
+//! Usage: (1) for creating collision mesh in Bullet
+struct CollisionMeshData {
```
bigbike: If it is a different kind of mesh data, can you please create a pair of new files (.h/.cpp)?

CollisionMeshData.h / .cpp

src/esp/assets/PhysicsManager.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+namespace esp {
+namespace assets {
+
+bool PhysicsManager::initPhysics(
+      scene::SceneNode*     node,
+      bool                  do_profile) {
+  LOG(INFO) << "Initializing Physics Engine...";
+
+  _bCollisionConfig = new btDefaultCollisionConfiguration();
```
bigbike: Consider using unique pointers...

src/esp/assets/PhysicsManager.cpp
```diff
@@ -0,0 +1,266 @@
...
+    const MeshMetaData& metaData,
+    std::vector<CollisionMeshData> meshGroup,
+    physics::BulletRigidObject* physObject) {
+
+  // Test Mesh primitive is valid
+  for (CollisionMeshData& meshData: meshGroup) {
+    if (!isMeshPrimitiveValid(meshData)) {return false;}
+  }
+
+  bool objectSuccess;
```
bigbike: Always initialize it.
objectSuccess = false; // or true, depending on your case

src/esp/assets/PhysicsManager.cpp
```diff
@@ -0,0 +1,266 @@
...
+  float mass;
```
bigbike: mass = 0.0f.

Are you sure bullet is using float instead of double?
In physically based simulation, it is super rare to use single precision floating point to do the computation. It causes significant numerical instability sometimes.

bigbike:
By the way, I did not even find function calls, such as addRigidBody(...) which is necessary to register the rigid objects in the bullet physics world.

src/esp/assets/PhysicsManager.cpp
mosra: `btScalar` is usually a float, yes (but can be configured to be a double instead (which is very rare OTOH)). This is for the interfaces, not sure what they're using internally for calculations.

src/esp/gfx/Viewer.h
```diff
@@ -37,17 +39,42 @@ class Viewer : public Magnum::Platform::Application {
...
   scene::ObjectControls controls_;
   Magnum::Vector3 previousPosition_;
 
   bool computeActionPath_;
+  bool enablePhysics_;
+  bool surreal_mesh = false;
+  bool replica_mesh = false;
+  bool vangoth_mesh = true;
+  bool castle_mesh = false;
+
```
hzyjerry: These are very hacky way to specify initial camera location, based on which mesh we're using.

src/esp/assets/FRLInstanceMeshData.h
hzyjerry: Good catch! The function will be removed.

src/esp/assets/FRLInstanceMeshData.cpp
```diff
@@ -129,6 +129,33 @@ bool FRLInstanceMeshData::loadPLY(const std::string& ply_file) {
     xyzid.head<3>() = xyz_esp;
   }
 
+  // Store vertex buffer without instance id
+  cpu_vbo_3 = new std::vector<vec3f>(cpu_vbo.size());
+  for (int i = 0; i < cpu_vbo.size(); ++i) {
+    (*cpu_vbo_3)[i] = cpu_vbo[i].head<3>();
+  }
+
+  // Compute tri index buffer
```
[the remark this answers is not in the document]
hzyjerry: Updated. Thanks

src/esp/assets/MeshData.h
hzyjerry: Good point.

src/esp/assets/PhysicsManager.cpp
hzyjerry: Updated

src/esp/assets/PhysicsManager.h
hzyjerry: Updated

src/esp/assets/ResourceManager.h
hzyjerry: Original `parent` is much better. Renamed back.

src/esp/gfx/CMakeLists.txt
```diff
@@ -20,6 +20,8 @@ find_package(Magnum
     MeshTools
     SceneGraph
     Shaders
+    # Sdl2Application   # Tried to make physics example work
+    TextureTools
```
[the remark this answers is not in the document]
hzyjerry: Nope, removed. Thanks!

src/cmake/FindMagnumIntegration.cmake
```diff
@@ -70,6 +70,9 @@
 #   DEALINGS IN THE SOFTWARE.
 #
 
+## TODO(JH) here forcing local with bullet magnum-integration
+set(WITH_BULLETINTEGRATION ON CACHE BOOL "WITH_BULLETINTEGRATION" FORCE)
```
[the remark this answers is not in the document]
hzyjerry: Updated to `set(WITH_BULLET ON CACHE BOOL "WITH_BULLET" FORCE)` in `dependencies.cmake`

src/CMakeLists.txt
```diff
@@ -80,4 +84,4 @@ if (BUILD_TEST)
 endif()
 
 # pybind bindings
-add_subdirectory(esp/bindings)
+#add_subdirectory(esp/bindings)
```
hzyjerry: I had some conflict with py bindings before when developing. Need to enable this back

setup.py
```diff
@@ -72,7 +72,7 @@ def build_parser():
         help="""Caches the arguements sent to setup.py
         and reloads them on the next invocation.  This argument is not cached""",
     )
-
+    # --cmake-args="-DWITH_BULLET=ON"
```
hzyjerry: This is needed only if users want to build with bullet. Not sure if we should always set it to `ON` though

src/esp/assets/FRLInstanceMeshData.cpp
```diff
@@ -129,6 +129,33 @@ bool FRLInstanceMeshData::loadPLY(const std::string& ply_file) {
...
+    (*tri_ibo)[triIdx + 4] = quadIdx + 2;
+    (*tri_ibo)[triIdx + 5] = quadIdx + 3;
+  }
+
+  // Store indices, facd_ids in Magnum MeshData3D format such that
+  // later they can be accessed.
+  // Note that normal and texture data are not stored
+  meshData_ = Magnum::Trade::MeshData3D(Magnum::GL::MeshPrimitive::Triangles,
+      reinterpret_cast<std::vector<Magnum::UnsignedInt>>(tri_ibo->data()),
+      reinterpret_cast<std::vector<std::vector<Magnum::Vector3>>>(cpu_vbo_3->data()));
```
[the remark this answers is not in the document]
hzyjerry: Updated in the latest version :)

hzyjerry:
> By the way, I did not even find function calls, such as addRigidBody(...) which is necessary to register the rigid objects in the bullet physics world.

Yup. If you're looking for the lowest level function, it's inside `esp/physics/BulletObject.cpp` where `bWorld_->addCollisionObject(bCollisionBody_);`

However this function is intentionally hidden from the user. What happens is 

1. `ResourceManager.cpp` internally calls `physicsManager.initObject()` or `physicsManager.initScene()`
2. `PhysicsManager.cpp` internally creates `BulletRigidObject` and calls `object.initializeObject()` and `object.initializeScene()`. This both initializes the object (sets shape, weight, etc) and adds it into `PhysicsManager -> bWorld_`

src/esp/assets/PhysicsManager.cpp
```diff
@@ -0,0 +1,246 @@
...
+    const MeshMetaData& metaData,
+    std::vector<CollisionMeshData> meshGroup,
+    physics::BulletRigidObject* physObject) {
+
+  // Test Mesh primitive is valid
+  for (CollisionMeshData& meshData: meshGroup) {
+    if (!isMeshPrimitiveValid(meshData)) {return false;}
+  }
+
+  float mass = 0.0f;    // TODO (JH) weight is currently hardcoded
```
bigbike: Saw the mass = 0.0f. Is it because the "scene" is static object in the btWorld?

src/esp/assets/PhysicsManager.cpp
```diff
@@ -0,0 +1,246 @@
...
+  }
+}
+
+void PhysicsManager::getPhysicsEngine() {}
+
+// Bullet Mesh conversion adapted from:
+// https://github.com/mosra/magnum-integration/issues/20
+bool PhysicsManager::initScene(
+    const AssetInfo& info,
+    const MeshMetaData& metaData,
```
bigbike: Did you need to use the metaData in this function?

src/esp/assets/PhysicsManager.cpp
```diff
@@ -0,0 +1,246 @@
...
+      case Magnum::MeshPrimitive::LineLoop:
+        LOG(ERROR) << "Invalid primitive Line loop"; break;
+      case Magnum::MeshPrimitive::LineStrip:
+        LOG(ERROR) << "Invalid primitive Line Strip"; break;
+      case Magnum::MeshPrimitive::TriangleStrip:
+        LOG(ERROR) << "Invalid primitive Triangle Strip"; break;
+      case Magnum::MeshPrimitive::TriangleFan:
+        LOG(ERROR) << "Invalid primitive Triangle Fan"; break;
+      default:
+        LOG(ERROR) << "Invalid primitive " << int(meshData.primitive);
```
bigbike: better to have "break" for default as well.

src/esp/assets/PhysicsManager.h
```diff
@@ -0,0 +1,96 @@
...
+  void getPhysicsEngine();
+
+  //! Check if mesh primitive type is valid for bullet physics engine
+  bool isMeshPrimitiveValid(CollisionMeshData& meshData);
+
+  //! ==== physics engines ====
+  //! The world has to live longer than the scene because RigidBody
+  //! instances have to remove themselves from it on destruction
+  Magnum::BulletIntegration::DebugDraw      debugDraw_{Magnum::NoCreate};
+  btDbvtBroadphase*                         bBroadphase_;
```
bigbike: These should be the unique pointers

src/esp/physics/BulletObject.h
```diff
@@ -0,0 +1,59 @@
...
+#include "esp/assets/BaseMesh.h"
+#include "esp/assets/MeshData.h"
+#include "esp/assets/GenericInstanceMeshData.h"
+#include "esp/assets/FRLInstanceMeshData.h"
+#include "esp/core/esp.h"
+
+namespace esp {
+namespace physics {
+
+class BulletRigidObject : public scene::SceneNode {
```
bigbike: We never ever inherit any class from the sceneNode to avoid "double deallocation", and to avoid ownership problems in pyBindings (if this class needs to be exposed in py which is very likely). This is because the SG memory management is automatically handled by magnum lib. Not so sure if the above theory still holds. If it is, then this is a problem here.

A "camera, agent, sensor, or light etc." is treated as an attached object to the scene node.
I believe the "rigid object" should follow the same logic.

More details (Pay attention to the figures in the PR description):
https://github.com/fairinternal/habitat-sim/pull/40

mosra: This will get fully resolved with #64.

src/esp/assets/PhysicsManager.cpp
```diff
@@ -109,15 +104,16 @@ bool PhysicsManager::initScene(
   } else {                                                  // GLB mesh data
     LOG(INFO) << "Initialize GLB scene";
   }
-  sceneSuccess = physObject->initializeScene(info, mass, meshGroup, *bWorld_);
+  sceneSuccess = physObject->initializeScene(info, mass, meshGroup, *bWorld_.get());
```
erikwijmans: `.get()` is redundant.  `*bWorld_` will dereference the underlying pointer.

hzyjerry: Thanks

src/esp/assets/PhysicsManager.cpp
erwincoumansepic: Bullet author here: I would highly recommend using double precision, that's our default in the context of robotics and reinforcement learning. Instead of implementing your own C++ and Python binding, you may also learn a bit more about [PyBullet](https://docs.google.com/document/d/10sXEhzFRSnvFcl3XxNGhnD4N2SedqwdAvK3dsihxVUA/edit#heading=h.2ye70wns7io3). PyBullet also shows the usage of articulated bodies (btMultiBody) instead of just btRigidBody, and Inverse Kinematics, Inverse Dynamics etc. Happy to discuss more, just catch me on twitter or email.

aclegg3: Hey Erwin, thanks for the comment. We are absolutely looking into PyBullet and would love to chat sometime.

erwincoumans: In PyBullet, it is easy to create more complex robots in btMultiBody (generalized coordinates) or optionally as btRigidBody+btTypeConstraints. Enabling btMultiBody is not easy to get right. Since you already have Python, it may be an easy integration. Happy to chat.

erwincoumans: You can see some of the features of PyBullet in its quickstart guide. We use it a lot for reinforcement learning and robotics. For example, there is a deep_mimic version, re-implementing Jason Peng's motion imitation, and various quadruped robots and arm grasping setups. See https://docs.google.com/document/d/10sXEhzFRSnvFcl3XxNGhnD4N2SedqwdAvK3dsihxVUA/edit#heading=h.2ye70wns7io3

src/esp/assets/ResourceManager.cpp
```diff
@@ -333,11 +333,63 @@ int ResourceManager::loadObject(const std::string objPhysConfigFilename)
   //2. construct a physicsObjectMetaData
   PhysicsObjectMetaData physMetaData;
 
-  //3. load/check_for render and collision mesh metadata
-  //    ALEX NOTE: these paths should be relative to the properties file
+  //ALEX NOTE: these paths should be relative to the properties file
   std::string propertiesFileDirectory = objPhysConfigFilename.substr(0,
       objPhysConfigFilename.find_last_of("/"));
```
mosra: [Utility::Directory::path()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1Directory.html#a222baead7128ec0cadc433e3a8c0059f), there's also a lot more utilities for path handling

src/esp/assets/ResourceManager.cpp
```diff
@@ -568,38 +602,30 @@ std::string ResourceManager::getObjectConfig(int objectID) {
...
+
+    minCorner[0] = fmin(minCorner.x(), pos.x());
+    minCorner[1] = fmin(minCorner.y(), pos.y());
+    minCorner[2] = fmin(minCorner.z(), pos.z());
   }
-  LOG(INFO) << "Shifting data origin";
-  Magnum::Matrix4 transform = Magnum::Matrix4::translation(Magnum::Vector3(
-      -(maxX + minX) / 2, -(maxY + minY) / 2, -(maxZ + minZ) / 2));
+
+  return (maxCorner+minCorner)/2.0;
```
mosra: I feel like I already said this in another review of this exact code but can't find it right now -- this whole loop could be simplified to roughly the following, making use of [Math::minmax()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Math.html#ac0524e916ab96970c6b00e99ecc6e912):

```cpp
#include <Corrade/Containers/ArrayViewStl.h>
#include <Magnum/Math/FunctionsBatch.h>

...

return Magnum::Range3D{Magnum::Math::minmax<Vector3>(meshData.positions)}.center();
```

In general, if you are ever looping over numbers to get their bounds or doing repetitive calculations over vector components like above, there's a high chance Magnum already has a well-tested function for that. Just search around, it'll save you *massive* amounts of time ;)

build.sh
```diff
@@ -27,6 +27,7 @@ done
 my_hostname=`hostname`
 if [[ $my_hostname =~ "fair" ]]; then
   module purge
+  module load anaconda3/5.0.1
```
erikwijmans: This will end-up breaking the build on the fair cluster if you are using your own conda install.

src/esp/assets/ResourceManager.cpp
```diff
@@ -349,9 +349,17 @@ int ResourceManager::loadObject(const std::string objPhysConfigFilename,
     MeshMetaData meshMetaData = resourceDict_[filename];
     scene::SceneNode& newNode = parent->createChild();
     AssetInfo renderMeshinfo = AssetInfo::fromPath(filename);
-
+    Magnum::PluginManager::Manager<Importer> manager{
+#ifdef MAGNUM_BUILD_STATIC
+        "./"
+#endif
```
mosra: FYI, this `#ifdef` is not needed since https://github.com/mosra/corrade/commit/1329656464ed4afbeaa1ff96ad0f0f5ec3149a0f, simply instantiating the manager is enough. Same below :)

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -0,0 +1,435 @@
...
+  // Current not supported
+  /*if (isScene_)
+    return;
+  else
+    bObjectRigidBody_->setCenterOfMassTransform(
+        btTransform(Magnum::Matrix4<float>::translation(COM)));*/
+}
+
+void BulletRigidObject::setInertia(const Magnum::Vector3& inertia) {
+  if (isScene_)
```
bigbike: braces.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -0,0 +1,223 @@
...
+      ->setRestitutionCoefficient(restitutionCoefficient);
+}
+
+const double BulletPhysicsManager::getMargin(const int physObjectID) {
+  if (existingObjects_.count(physObjectID) > 0) {
+    return std::dynamic_pointer_cast<BulletRigidObject>(
+               existingObjects_.at(physObjectID))
+        ->getMargin();
+  } else {
+    return -1.0;
```
bigbike: magic number.

src/esp/physics/PhysicsManager.h
```diff
@@ -0,0 +1,210 @@
...
+  //! ==== physics engines ====
+  enum PhysicsSimulationLibrary { NONE, BULLET };
+  PhysicsSimulationLibrary activePhysSimLib_ = NONE;  // default
+
+  //! Used to keep track of all sceneNodes that have physical properties
+  scene::SceneNode* physicsNode_ = nullptr;
+  std::shared_ptr<physics::RigidObject> sceneNode_ = nullptr;
+
+  //! ==== dynamic object resources ===
+  std::map<int, std::shared_ptr<physics::RigidObject>> existingObjects_;
```
erikwijmans: `RigidObject` current inherits from a `Magnum::SceneGraph::Object` and thus can't be shared nor can you own that memory, so anywhere with `std::shared_ptr<RigidObject>` will cause a double free.  @mosra I believe making `RigidObject` inherit from `Magnum::SceneGraph::AbstractFeature3D` is the correct way to allow for a persistent reference to a `Magnum::SceneGraph::Object`?

erikwijmans: This will end up being fine as long as the parent object doesn't get deleted before any of the objects in the list.  A map to `unique_ptr` _might_ end up being okay, it isn't really a unique ptr, but it might be okay.

erikwijmans: Idea:  We store the `physics::RigidObject` just as raw pointers.  Then we have a physics manager per scene and the resource manager is then responsible for tying the lifetime of the physics manager and the scene graph it is managing objects within.

mosra: > I believe making `RigidObject` inherit from `Magnum::SceneGraph::AbstractFeature3D` is the correct way to allow for a persistent reference to a `Magnum::SceneGraph::Object`?

Yes, that's the designed way to do that. I would then rename it to not have `Object` in its name, tho, to avoid confusing inheritance assumptions. (`RigidBody` for example?)

> fine as long as the parent object doesn't get deleted before any of the objects in the list

While working on the Python bindings I made it possible to override this implicit memory management behavior if you implement your own `doErase()` on either the object or the feature ([docs](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1LinkedListItem.html#acb954b145234f6ff5ae43f9f2a951b0e), hopefully understandable enough). In this case, if it gets implemented on the feature, the feature stays alive even after its owner object gets destroyed and then the proposed `unique_ptr` would take care of its deletion when appropriate.

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -0,0 +1,435 @@
...
+  //! Sync render pose with physics
+  bWorld_ = bWorld;
+  initialized_ = true;
+  syncPose();
+  return true;
+}  // end BulletRigidObject::initializeObject
+
+bool BulletRigidObject::removeObject() {
+  bWorld_->removeRigidBody(bObjectRigidBody_.get());
+  bWorld_ = nullptr;
```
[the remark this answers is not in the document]
mosra: Curious -- why? :) Both operations are equivalent and I like to use the `= nullptr` (or even `= {}`) because (for me at least) it doesn't make it any less clear.

src/esp/physics/RigidObject.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+bool RigidObject::isActive() {
+  // Alex NOTE: no active objects without a physics engine... (kinematics don't
+  // count)
+  return false;
+}
+
+RigidObject::~RigidObject() {
+  if (initialized_) {
+    LOG(INFO) << "Deleting object ";
```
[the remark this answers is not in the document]
mosra: Why do we need the destruction logs, anyway? It just spams the output for me.

src/esp/physics/RigidObject.h
```diff
@@ -0,0 +1,121 @@
...
+  // ==== Transformations ===
+  //! Need to overwrite a bunch of functions to update physical states
+  virtual SceneNode& setTransformation(
+      const Magnum::Math::Matrix4<float>& transformation);
+  virtual SceneNode& setTranslation(const Magnum::Math::Vector3<float>& vector);
+  virtual SceneNode& setRotation(
+      const Magnum::Math::Quaternion<float>& quaternion);
+
+  virtual SceneNode& resetTransformation();
+  virtual SceneNode& translate(const Magnum::Math::Vector3<float>& vector);
```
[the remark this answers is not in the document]
mosra: Why? The `SceneNode` is full of these, the magnum math types are fully exposed to python now.

src/esp/physics/RigidObject.h
```diff
@@ -0,0 +1,121 @@
...
+  virtual SceneNode& rotateLocal(
+      const Magnum::Math::Rad<float> angleInRad,
+      const Magnum::Math::Vector3<float>& normalizedAxis);
+
+  virtual SceneNode& rotateX(const Magnum::Math::Rad<float> angleInRad);
+  virtual SceneNode& rotateY(const Magnum::Math::Rad<float> angleInRad);
+  virtual SceneNode& rotateZ(const Magnum::Math::Rad<float> angleInRad);
+  virtual SceneNode& rotateXLocal(const Magnum::Math::Rad<float> angleInRad);
+  virtual SceneNode& rotateYLocal(const Magnum::Math::Rad<float> angleInRad);
+  virtual SceneNode& rotateZLocal(const Magnum::Math::Rad<float> angleInRad);
```
mosra: For all `Magnum::Math::Type<float>` you can say just `Magnum::Type` (and for `double` it would be `Magnum::Typed`).

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -0,0 +1,435 @@
...
+    //! Add to compound shape stucture
+    bObjectShape_->addChildShape(t, bObjectConvexShapes_.back().get());
+  }
+  //! Set properties
+  bObjectShape_->setMargin(margin);
+
+  btVector3 bInertia =
+      btVector3(physicsObjectAttributes.getMagnumVec3("inertia"));
+
+  if (bInertia[0] == 0. && bInertia[1] == 0. && bInertia[2] == 0.) {
```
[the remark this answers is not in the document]
mosra: The proper implementation that gives correct results also for very small and very large numbers is much more complicated than that, tho. There's [Math::TypeTraits::equals()](https://doc.magnum.graphics/magnum/structMagnum_1_1Math_1_1TypeTraits.html#a8e7ba417cfae3d00b717a17afed16152) that you can use for this.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -0,0 +1,481 @@
...
+  existingObjects_.erase(physObjectID);
+  deallocateObjectID(physObjectID);
+  return physObjectID;
+}
+
+int PhysicsManager::allocateObjectID() {
+  if (!recycledObjectIDs_.empty()) {
+    int recycledID = recycledObjectIDs_.back();
+    recycledObjectIDs_.pop_back();
+    return recycledID;
```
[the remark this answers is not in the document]
mosra: Just a note -- on the GPU side of things, object ID is currently just 32 bit, in case of per-primitive object IDs just 22 bit (due to them being in a float texture), and I was thinking about reducing it to just 16 to improve performance. So maybe we should add some assertions to the shaders to avoid th IDs getting truncated.

src/esp/assets/Attributes.cpp
```diff
@@ -0,0 +1,259 @@
...
+  doubleMap_ = std::map<std::string, double>();
+  intMap_ = std::map<std::string, int>();
+  stringMap_ = std::map<std::string, std::string>();
+  magnumVec3Map_ = std::map<std::string, Magnum::Vector3>();
+  vecStringsMap_ = std::map<std::string, std::vector<std::string> >();
+}
+
+// return true if any container has the key
+bool Attributes::exists(std::string key) {
+  if (doubleMap_.count(key) > 0)
```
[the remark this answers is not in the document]
msavva: 👍 The `README.md` has publicly-visible code style pointers: https://github.com/facebookresearch/habitat-sim#code-style

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -373,6 +373,20 @@ const Magnum::Vector3 BulletRigidObject::getInertia() {
...
+const Magnum::Matrix3 BulletRigidObject::getInertiaMatrix() {
+  if (rigidObjectType_ == SCENE) {
+    const Magnum::Matrix3 inertia = Magnum::Matrix3();
+    return inertia;
+  } else {
+    const Magnum::Vector3 vecInertia = getInertiaVector();
+    const Magnum::Matrix3 inertia =
+        Magnum::Matrix3(Magnum::Vector3(vecInertia.x(), 0, 0),
+                        Magnum::Vector3(0, vecInertia.y(), 0),
+                        Magnum::Vector3(0, 0, vecInertia.z()));
```
mosra: Not complaining, just a suggestion: there's [Matrix3::fromDiagonal()](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1RectangularMatrix.html#a2b5423dcaab476595937428092187123) so you don't need to type it all out ;)

src/esp/assets/Attributes.cpp
```diff
@@ -0,0 +1,259 @@
...
+    stringVec.erase(position);
+}
+
+// return a formated string exposing the current contents of the attributes maps
+std::string Attributes::listAttributes() {
+  std::string attributes =
+      "List of attributes: \n----------------------------------------\n";
+
+  attributes += "\nDoubles: \n";
+  for (auto it = doubleMap_.begin(); it != doubleMap_.end(); ++it) {
```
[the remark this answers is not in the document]
erikwijmans: Yay for auto and range-based for loops tho :)

src/esp/assets/ResourceManager.h
```diff
@@ -33,67 +45,113 @@ class Drawable;
...
+  //! Load Object data and store internally
+  //! Does not instantiate (physics & drawable)
+  //! Return index in physicsObjectList_
+  int loadObject(const std::string objPhysConfigFilename,
+                 scene::SceneNode* parent,
+                 DrawableGroup* drawables);
+
+  // load an object into the physicsObjectLibrary_ from a physics properties
+  // filename
+  int loadObject(const std::string objPhysConfigFilename);
```
erikwijmans:
```suggestion
  int loadObject(const std::string& objPhysConfigFilename);
```

src/esp/physics/PhysicsManager.h
```diff
@@ -0,0 +1,211 @@
...
+  };
+
+  virtual ~PhysicsManager();
+
+  //============ Initialization =============
+  // load physical properties and setup the world
+  // do_profile indicates timing for FPS
+  virtual bool initPhysics(
+      scene::SceneNode* node,
+      assets::PhysicsManagerAttributes physicsManagerAttributes);
```
erikwijmans:
```suggestion
      const assets::PhysicsManagerAttributes& physicsManagerAttributes);
```
I believe this can be a const reference.

src/esp/physics/PhysicsManager.h
```diff
@@ -0,0 +1,211 @@
...
+  // Stores references to a set of drawable elements
+  using DrawableGroup = Magnum::SceneGraph::DrawableGroup3D;
+
+  //============ Object/Scene Instantiation =============
+  //! Initialize scene given mesh data
+  //! Only one scene per simulation
+  //! The scene could contain several components
+  virtual bool addScene(
+      const assets::AssetInfo& info,
+      assets::PhysicsSceneAttributes& physicsSceneAttributes,
```
erikwijmans:
```suggestion
      const assets::PhysicsSceneAttributes& physicsSceneAttributes,
```

Same here?

src/esp/physics/PhysicsManager.h
```diff
@@ -0,0 +1,211 @@
...
+  virtual const double getMargin(const int physObjectID) { return 0.0; };
+  virtual void setMargin(const int physObjectID, const double margin){};
+
+  // =========== Debug functions ===========
+  int checkActiveObjects();
+
+  //============ Interact with objects =============
+  // NOTE: engine specifics handled by objects themselves...
+  void applyForce(const int physObjectID,
+                  Magnum::Vector3& force,
```
erikwijmans:
```suggestion
                  const Magnum::Vector3& force,
```

I believe this is an input only argument? i.e. `applyForce` does not change the value of `force`?

src/esp/physics/RigidObject.cpp
```diff
@@ -0,0 +1,176 @@
...
+  return true;
+}
+
+bool RigidObject::isActive() {
+  // NOTE: no active objects without a physics engine... (kinematics don't
+  // count)
+  return false;
+}
+
+void RigidObject::applyForce(Magnum::Vector3& force, Magnum::Vector3& relPos) {
```
erikwijmans: All of these functions that don't use their arguments can be defined as `(const Magnum::Vector3&, const Magnum::Vector3&)` to deal with the compiler warning and to signify that the methods don't use that argument.

src/esp/gfx/Viewer.h
```diff
@@ -11,11 +11,15 @@
...
 #include <Magnum/SceneGraph/Camera.h>
+#include <Magnum/Timeline.h>
 
 #include "esp/agent/Agent.h"
 #include "esp/assets/ResourceManager.h"
 #include "esp/gfx/RenderCamera.h"
 #include "esp/nav/PathFinder.h"
+#include "esp/physics/PhysicsManager.h"
+#include "esp/physics/RigidObject.h"
+//#include "esp/physics/bullet/BulletPhysicsManager.h"
```
erikwijmans: Should this be deleted or does it need an `#ifdef <thing>` guard?

src/CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ include(cmake/dependencies.cmake)
 # include source dirs
 include_directories(${PROJECT_SOURCE_DIR})
 
+# Physics
+add_definitions(-DBT_ENABLE_PROFILE)
```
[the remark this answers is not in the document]
mosra: Doesn't this make everything *really* slow? (That's what profiling usually does, in my experience.) Shouldn't that be configurable via an `option()` (and the recently-added `configure_file()`) instead?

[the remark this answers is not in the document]
erwincoumans: It shouldn't make things very slow. In addition, you can call an API that makes it almost for free. See Bullet/src/LinearMath/btQuickprof.h

src/esp/assets/ResourceManager.cpp
aclegg3: Default shift is false. If shifted, the transform is now stored for query. This is only expected to be used for object meshes, not static scenes.

src/esp/gfx/Viewer.h
aclegg3: Only the Bullet specific classes need ifdef right now. We didn't need that include here, so removed.

src/esp/gfx/Simulator.cpp
```diff
@@ -202,5 +203,111 @@ bool operator!=(const SimulatorConfiguration& a,
...
+}
+
+// apply forces and torques to objects
+void Simulator::applyTorque(const Magnum::Vector3& tau,
+                            const int objectID,
+                            const int sceneID) {
+  if (physicsManager_ != nullptr && sceneID >= 0 && sceneID < sceneID_.size()) {
+    physicsManager_->removeObject(objectID);
+  }
+}
```
mosra: :thinking:  This isn't using `tau` in any way. Bad copypaste from above? :)

aclegg3: Yes, thanks for catching this. 😂

examples/example.py
```diff
@@ -48,24 +49,46 @@ def make_settings():
...
+        % (settings["width"], settings["height"], perf["total_time"]),
+        "frame time %0.3f ms (%0.1f FPS)" % (perf["frame_time"] * 1000.0, perf["fps"]),
+    )
+    print(" ============================================================== ")
 
-assert perf["fps"] > args.test_fps_regression, (
-    "FPS is below regression threshold: %0.1f < %0.1f"
-    % (perf["fps"], args.test_fps_regression)
-)
+    # assert perf["fps"] > args.test_fps_regression, (
```
mathfac: @aclegg3, maybe I missed the comment. Can you remind me why we commented out the speed regression check??

aclegg3: Should be there when not running with physics. May have been commented out in the rush to verify physics pybind, sorry for that. We will need a separate threshold for the new physics benchmarks, but we can add this back when enable_physics is false.

aclegg3: Should be there when not running with physics. May have been commented out in the rush to verify physics pybind, sorry for that. We will need a separate threshold for the new physics benchmarks, but we can add this back when enable_physics is false.

mathfac: @aclegg3, no worries, just wanted to check there is no serious reason to disable it.

PR #103, opened 2019-07-15
Fix bindings cyclic type references and type annotations

mosra:
## Motivation and Context

- The binding code was referencing types that weren't known yet to pybind (either described later or not at all), causing the exported functions to have C++ names in their signatures, making them unusable from Python.
- Various Python APIs had wrong type annotations. In particular, the `np.array()` is a function, the correct type is `np.ndarray`. Unfortunately that's all we can annotate for now; type, dimensionality and size needs to wait until https://github.com/numpy/numpy/issues/7370.

Discovered while running [m.css](https://mcss.mosra.cz) on the project to generate docs for the Python side.

## How Has This Been Tested

It builds? :sunglasses: 

This brings no functional change, so should be easy to review and merge.

## Types of changes

### DIFF
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -50,18 +50,18 @@ class SixDOFPose(object):
         rotation (np.quaternion): unit quaternion rotation
     """
 
-    position: np.array = np.zeros(3)
+    position: np.ndarray = np.zeros(3)
     rotation: Union[np.quaternion, List] = np.quaternion(1, 0, 0, 0)
 
 
 @attr.s(auto_attribs=True, slots=True)
 class AgentState(object):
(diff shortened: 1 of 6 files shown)


erikwijmans:
If it build, it ships! Thank you for fixing these.  I will attempt to remember that `np.array` is not a type (I always forget that....)

mosra:
I really hope proper type annotations for numpy types are going to land sometime soon so we can use `np.ndarray[float, [3]]` and such. The current way doesn't really tell much about the expected type, being only slightly better than doing `def foo() -> object` :sweat_smile:

mosra:
Ready to merge? Or should I wait for one review? :)

erikwijmans:
Should be good to land!

PR #105, opened 2019-07-15
Rework the __all__ property of Python modules, phase out the bindings module

mosra:
## Motivation and Context

In preparation for reworked docs (#71), reorganizing the `__all__` property of all modules to make more sense and be inspectable by doc generation tools. In particular:

- Avoiding extra deep hierarchies like `habitat_sim.agent.controls` but also not putting everything directly into `habitat_sim`.
- Taking stuff out of the `bindings` module and reorganizing it to other modules. The `bindings` module only grouped things coming from C++ until now and that semantically didn't have any value.
- Adding random things that were missing from `__all__`, being invisible to doc generators, IDE autocompletion etc.

What is *not* done:

- Getting rid of the old `bindings` module. It's kept there for backwards compatibility with existing code.
- Removing the `from foo import *` in the root `__init__.py` for the same reason.
- I don't see why the native module should be deep down in `_ext.bindings.habitat_sim_bindings`, I think having it as `_habitat_sim` *next* to the pure Python `habitat_sim` would be perfectly okay.
- The general structure and naming could be improved to be more clear, but I don't think that's something I should be doing -- you have a much better idea about the desired overall project structure than me.

## How Has This Been Tested

Docs generated with [m.css](https://mcss.mosra.cz) now reference all classes as expected, has a flattened class list and doesn't include the `bindings` module anymore. Sample is at https://tmp.mosra.cz/habitat-sim/

### DIFF
--- a/habitat_sim/__init__.py
+++ b/habitat_sim/__init__.py
@@ -9,7 +9,45 @@ import builtins
 __version__ = "0.1.2"
 
 if not getattr(builtins, "__HSIM_SETUP__", False):
+    # TODO: all those import * should get removed, kept only for compatibility
+    #   with existing code
     from .nav import *
     from .agent import *
     from .simulator import *
-    from .bindings import *
+
+    # Bindings has a geo module, but we have our own. So take everything except
+    # that.
+    from .bindings import (
+        SceneNodeType,
+        GreedyFollowerCodes,
+        GreedyGeodesicFollowerImpl,
+        MultiGoalShortestPath,
+        PathFinder,
+        PinholeCamera,
+        SceneGraph,
+        SceneNode,
+        Sensor,
+        SensorSpec,
+        SensorType,
+        ShortestPath,
+        SimulatorConfiguration,
+    )
+
+    from . import agent, geo, gfx, logging, nav, scene, sensor, simulator, utils
+    from ._ext.habitat_sim_bindings import MapStringString
+
+    __all__ = [
+        "agent",
+        "nav",
+        "sensors",
+        "errors",
+        "geo",
+        "gfx",
+        "logging",
+        "nav",
+        "scene",
+        "sensor",
+        "simulator",
+        "utils",
+        "MapStringString",
+    ]
--- a/habitat_sim/agent/controls/__init__.py
+++ b/habitat_sim/agent/controls/__init__.py
@@ -8,6 +8,7 @@ from habitat_sim.agent.controls.controls import (
     ActuationSpec,
     ObjectControls,
     SceneNodeControl,
+    move_func_map,
     register_move_fn,
 )
 from habitat_sim.agent.controls.default_controls import *
@@ -16,6 +17,7 @@ from habitat_sim.agent.controls.pyrobot_noisy_controls import PyRobotNoisyActuat
 __all__ = [
     "ActuationSpec",
     "ObjectControls",
+    "move_func_map",
     "SceneNodeControl",
     "register_move_fn",
     "PyRobotNoisyActuationSpec",
--- a/habitat_sim/agent/__init__.py
+++ b/habitat_sim/agent/__init__.py
@@ -6,3 +6,5 @@
 
 from .agent import *
 from .controls import *
+
+__all__ = agent.__all__ + controls.__all__
--- a/habitat_sim/bindings/__init__.py
+++ b/habitat_sim/bindings/__init__.py
@@ -4,6 +4,9 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+# TODO: this whole thing needs to get removed, kept just for compatibility
+#   with existing code
+
 modules = [
     "SceneNodeType",
     "GreedyFollowerCodes",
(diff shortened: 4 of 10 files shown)


erikwijmans:
Yeah, there are lots of `*.py` files to keep them small and reasonably self-contained.  The reason to not to `_agent.py` is that you need to do absolute imports in other `*.py` files that are part of the same folder to avoid circular imports, i.e. `default_controls.py` needs to do `from habitat_sim.agent.agent import AgentState`

erikwijmans:
> Is it important to have a distinction between C++ and pure Python classes by having a module named habitat_sim.bindings? What about putting them instead into modules named habitat_sim.scene, habitat_sim.gfx, habitat_sim.geo to mirror the C++ side?

Breaking up our python bindings into multiple modules would definitely be good and would help reduce the absurd about of time to link `habitat_sim_bindings.so`

mosra:
I actually meant creating submodules inside the single `*.so`, using `m.def_submodule()`. Since it's usually just one or two classes in each submodule, I don't think creating one `*.so` per submodule makes sense.

According to my measurements, the most absurd link time is now for `_magnum.so`, due to all the math types exposed. And after I moved most stuff out of the habitat bindings into `magnum.scenegraph`, link time of `habitat_sim_bindings.so` is no longer an issue I think :)

erikwijmans:
Ah, that makes sense also :)

mosra:
@msavva @erikwijmans I'd like to have this resolved and merged with a clear decision how to go forward with module organization in the project, because all documentation efforts kinda depend on having this cleared up.

Docs generated using this PR are at https://tmp.mosra.cz/tmp/habitat-docs/classes.html, please have a look and either confirm that the hierarchy makes sense or suggest improvements. Thank you.

mosra:
Okay, so I'll try pulling them out of the `bindings` and into modules of the same names as the underlying C++ code is structured.

> Is that something which would be easy to do?

The user-facing part yes, just populating `__all__` at correct places. For now (in the scope of this PR) I'll try to preserve backwards compatibility, keeping the `bindings` module as-is.

> Other than that, the generated docs look awesome to me!

There's less than 2% of what *should* be there :)

erikwijmans:
> The user-facing part yes

Just user-facing is fine by me!

mathfac:
@mosra thank you for sorting things out with user facing interfaces. I support @erikwijmans and @msavva regarding merging C++ classes into Python namespaces. That make sense also for future approach, when we can move some Python functionality to C++ for performance improvements and user won't even notice that.
Side node: 
`class ObjectControls and class SceneNodeControl` look more related to general scene manipulation than agent. Another challenge is naming of C++ and Python Sensor classes.

mosra:
@erikwijmans @msavva this is now ready for review from my side. Next I'm updating #71 and basing it off this commit.

mosra:
Pushed an updated version where existing code is actually able to run and pass tests. But the root `__init__.py` is a mess, the codebase will really need a cleanup pass once this gets merged ;)

habitat_sim/agent/controls/__init__.py
```diff
@@ -8,6 +8,7 @@
     ActuationSpec,
     ObjectControls,
     SceneNodeControl,
+    move_func_map,
```
erikwijmans: This map is really more of an internal thing.  You shouldn't interact with it directly IMO.  You can if you need to, but I don't think it should be officially in the public API.

mosra: It's referenced from here, which triggered me to expose it: https://tmp.mosra.cz/habitat-sim/habitat_sim.agent.ObjectControls.html#action

erikwijmans: Ah.  Makes sense then.  For the doc for https://tmp.mosra.cz/habitat-sim/habitat_sim.agent.html#move_func_map, I am not 100% sure how that is generated, but it looks like it is just a print.  You can try https://docs.python.org/3/library/pprint.html for a much nicer looking print :)

mosra: Right now it's just what `str(thing)` returns, yes, as that worked well for basic constants and enums (see e.g. https://doc.magnum.graphics/python/magnum.math.html#data). This is on my TODO list, I'm thinking about completely omitting the value if it's too long -- for long dicts and lists (also https://tmp.mosra.cz/habitat-sim/habitat_sim.utils.html#data) the listing doesn't add much value anyway. If the contents are really important I think there should be a detailed doc block for it with a human-readable list instead. (Or the relevant snippet included as a highlighted code block, at least -- easy with the `.. include::` directive.)

Btw., this is exactly the feedback I'm looking for, so thanks a lot! :+1:

habitat_sim/__init__.py
```diff
@@ -9,7 +9,45 @@
...
+        SceneGraph,
+        SceneNode,
+        Sensor,
+        SensorSpec,
+        SensorType,
+        ShortestPath,
+        SimulatorConfiguration,
+    )
+
+    from . import agent, geo, gfx, logging, nav, scene, sensor, simulator, utils
```
erikwijmans:
```suggestion
    import agent, geo, gfx, logging, nav, scene, sensor, simulator, utils
```

mosra: As I said on Slack, this didn't really work :)

PR #106, opened 2019-07-15
Switch to a multiplatform mmap implementation.

mosra:
## Motivation and Context

Should help with #84 -- we don't need to reinvent things that are already implemented and tested :) Also made the code a bit more `const`-correct and simplified some pointer chasings.

## How Has This Been Tested

It builds for me on Linux.

~~:warning: Even though the change is pretty straightforward, there's not much to screw up and the underlying API has been extensively tested in other contexts, unfortunately I'm afraid I don't have any PTex mesh data to give this a *real* test, sorry.~~ Successfully tested with Replica.

## Types of changes

### DIFF
--- a/src/esp/gfx/WindowlessContext.cpp
+++ b/src/esp/gfx/WindowlessContext.cpp
@@ -4,9 +4,11 @@
 
 #include "WindowlessContext.h"
 
-#ifdef __APPLE__
+#include <Corrade/configure.h>
+
+#if defined(CORRADE_TARGET_APPLE)
 #include <Magnum/Platform/WindowlessCglApplication.h>
-#elif __linux__
+#elif defined(CORRADE_TARGET_UNIX)
 #include <Magnum/Platform/GLContext.h>
 
 #ifdef __ESP_USE_EGL__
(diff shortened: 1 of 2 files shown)


dhruvbatra:
> unfortunately I'm afraid I don't have any PTex mesh data to give this a real test, sorry.

I believe you can find them here: 
https://github.com/facebookresearch/Replica-Dataset

mosra:
Ah! :) Will check those out tomorrow, then.

src/esp/gfx/WindowlessContext.cpp
```diff
@@ -4,9 +4,11 @@
 
 #include "WindowlessContext.h"
 
-#ifdef __APPLE__
+#include <Corrade/configure.h>
+
+#ifdef CORRADE_TARGET_APPLE
```
erikwijmans:
```suggestion
#if defined(CORRADE_TARGET_APPLE)
```
?

mosra: Sure, if you prefer that :) In my case I always pick the shortest variant.

Updated (and rebased on master as well).

mosra:
Ok, it took ages to download, but everything seems to work the same way as before! :)

dhruvbatra:
Great, thanks!

PR #107, opened 2019-07-15
Performance enhancements

erikwijmans:
## Motivation and Context

Slightly faster implementations of things since a lot of the internal logic now uses magnum types instead of numpy types.

## How Has This Been Tested

Via a profiling script using python's `cProfile` package.

## Types of changes

Performance enhancement

### DIFF
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -169,16 +169,17 @@ class Agent(object):
     def get_state(self) -> AgentState:
         habitat_sim.errors.assert_obj_valid(self.body)
         state = AgentState(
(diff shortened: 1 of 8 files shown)


mosra:
You merged so fast I didn't have time to review :sweat_smile: 

One minor API usability thing -- I'm not sure if the new property wouldn't better be named `absolute_translation` instead of `translation`, all other absolute properties are prefixed with `absolute_` as well.

Besides that, does this improve performance *beyond* what was there before #64 got merged? Or do we still have some catching-up to do?

erikwijmans:
Performance seems to be about where it was before #64, if not slightly better, with #104 + this PR

PR #108, opened 2019-07-16
Make installing magnum always the default

erikwijmans:
## Motivation and Context

The behavior of not installing magnum if you already have it is nice for development and incremental builds, but not nice for user.  This PR makes reinstalling magnum the default and adds a `--skip-reinstall-magnum` flag to the build to skip the reinstall.

## How Has This Been Tested

`build.sh` and `setup.py install` now both always install magnum by default.

## Types of changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -312,20 +312,6 @@ class CMakeBuild(build_ext):
 
 
 if __name__ == "__main__":
-    try:
-        import magnum
-
-        has_magnum = True
-    except ImportError:
-        has_magnum = False
-
-    if not has_magnum and is_pip():
-        raise RuntimeError(
-            "Must have magnum already installed when installing habitat_sim"
-            "via pip due to limitations of setup.py\n"
-            "You can install magnum via 'python setup.py install'"
-        )
-
     assert StrictVersion(
         "{}.{}".format(sys.version_info[0], sys.version_info[1])
     ) >= StrictVersion("3.6"), "Must use python3.6 or newer"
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -352,7 +361,7 @@ def load(filename):
         _cmake_build_dir, "deps", "magnum-bindings", "src", "python"
     )
 
-    if not has_magnum:
+    if not args.skip_reinstall_magnum or not has_magnum:
```
mosra:
```suggestion
    if not args.skip_install_magnum:
```

Maybe `--skip-install-magnum` instead and make it skip *always*, even if magnum doesn't seem to be installed? I would consider that a "pro" option where you want it to reliably skip installing magnum no matter what was the reason for `ImportError` being raised.

erikwijmans: Makes sense and simplifies logic :)

erikwijmans: I will also remove the yelling about pip.  The fact that things work with pip is kinda shadow supported and the yelling thing actually doesn't work as pip does some weird things with the environment so `has_magnum` is always false when installing via pip anyways.

setup.py
```diff
@@ -352,7 +347,7 @@ def load(filename):
         _cmake_build_dir, "deps", "magnum-bindings", "src", "python"
     )
 
-    if not has_magnum:
+    if not args.skip_reinstall_magnum and not is_pip():
         subprocess.check_call(shlex.split(f"pip install {pymagnum_build_dir}"))
     else:
         print("Assuming magnum bindings are already installed")
```
mosra:
```suggestion
         print("Assuming magnum bindings are already installed (or we're inside pip and ¯\\_(ツ)_/¯)")
```

I would still yell *a bit*, tho -- better to have the message overly verbose than hiding something important. Feel free to change it to something reasonable, of course :)

mosra: Dang *why* it made that extra indentation there :see_no_evil: GitHub, you're buggy.

erikwijmans: Seems like the unicode character isn't being parsed correctly :(

mosra: oh well, but that's a normal katakana character? is this python 2 or...? :D

erikwijmans: It’s 3.6 so who knows

PR #114, opened 2019-07-19
Sensors of different resolutions and GPU-GPU transfer

### LINKED ISSUE #45: Support two RGB sensors with different resolution
## Steps to reproduce

  1. Enable 2 RGB sensors with same resolution and different location.
  2. Capture observations from both sensors.

## Observed Results

  * Generated frames are broken.

cc @bhavanajain for more details

erikwijmans:
## Motivation and Context

Currently, our renderer owns a global frame buffer which all sensors share for rendering.  This causes problems if you have sensors of different resolutions.  This PR introduces a `gfx::RenderTarget` which contains the various rendering buffers and framebuffer.  Each sensor has its *own*  `gfx::RenderTarget`, alleviating this issue.

`gfx::RenderTarget` was also made with GPU-GPU transfer via @jia-git's GLTensor in mind, so that is included too :)  At 256x256, I see a 1k FPS (from 3k to 4k) speed-up using GPU-GPU instead of GPU-CPU-GPU and a 3k FPS (from 700 to 3.7K) speed-up at 1024x1024.

~Unfortunately, it seems we finally need to make the OpenGL context a singleton as destructing with CUDA-GL interop takes a while and we get double context errors.~

To recap:
* Sensors now own their frame buffer via a `RenderingTarget`
* CUDA-GL Interop is here!
~* The GL Context is a singleton :(~

**EDIT:** Got rid of the singleton (a context singleton is actually an awful idea and doesn't work).  I was just careful and explicit with the ordering of calls to `del` to get stuff to delete.

Note:  This PR is _not_ 2.6k lines.  About 1.8k of those lines are two helper cuda header files :)

Resolves #45 

## How Has This Been Tested

Via the rendering unit test!

And visually :)

## Types of changes


src/esp/bindings/CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ target_link_libraries(habitat_sim_bindings
     sensor
 )
 
+if(WITH_GL_TENSOR)
+  target_compile_definitions(habitat_sim_bindings PRIVATE __ESP_WITH_GL_TENSOR__)
```
mosra: I'd stronly advise against using doubly-underscored identifiers. `__Uppercase` as well as `_Uppercase` is reserved by the standard, `_lowercase` is safe.

Could it be just `ESP_WITH_GL_TENSOR`? Or if ESP isn't unique enough then `HABITAT_ESP_WITH_GL_TENSOR`? That would make it easier to grep for as well I think.

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -0,0 +1,190 @@
...
+  void renderEnter() {
+    framebuffer_.bind();
+    framebuffer_.clear(GL::FramebufferClear::Color |
+                       GL::FramebufferClear::Depth);
+    framebuffer_.clearColor(1, Vector4{});
+    framebuffer_.clearColor(2, Vector4ui{});
+    framebuffer_.bind();
+  }
+
+  void renderExit() { glBindFramebuffer(GL_FRAMEBUFFER, 0); }
```
mosra: `GL::defaultFramebuffer.bind()` instead. Calling `glBind*()` directly messes up with Magnum's state tracker and you'd need to call `GL::Context::resetState()` to avoid unpredictable behavior.

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+#include "RenderingTarget.h"
+#include "magnum.h"
+
+using namespace Magnum;
+
+namespace esp {
+namespace gfx {
+struct RenderingTarget::Impl {
+  Impl(WindowlessContext::ptr context, int height, int width)
```
mosra: I'm sure there's some convention I'm not aware about but ... why is it height first? :) What about using a `Vector2i` for that instead to make the argument order clear? (Instantiating the class would be then just a matter of wrapping the args in `{width, height}`, or, in case of python, `[width, height]`, nothing verbose).

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+  void readFrameRgba(uint8_t* ptr) {
+    framebuffer_.mapForRead(GL::Framebuffer::ColorAttachment{0});
+    Image2D rgbaImage =
+        framebuffer_.read(Range2Di::fromSize({0, 0}, framebufferSize_),
+                          {PixelFormat::RGBA8Unorm});
+    uint8_t* src_ptr = rgbaImage.data<uint8_t>();
+    std::memcpy(
+        ptr, src_ptr,
+        framebufferSize_[0] * framebufferSize_[1] * 4 * sizeof(uint8_t));
```
mosra: Um, I must have missed this. Gimme a while, I'll give you an API that allows you to read directly into `ptr` without doing an extra allocation and copy (and without needing to go through Eigen on the bindings side). 

(It could be done with `Image2D` aready with a bunch of hacks, assumptions, wishes, and custom deleters, but it's far from pretty.)

src/esp/gfx/RenderingTarget.h
```diff
@@ -0,0 +1,40 @@
...
+
+  gltensor::GLTensorParam::ptr glTensorParamRgba() const;
+  gltensor::GLTensorParam::ptr glTensorParamDepth() const;
+  gltensor::GLTensorParam::ptr glTensorParamId() const;
+
+  void readFrameRgba(uint8_t* ptr);
+  void readFrameDepth(float* ptr);
+  void readFrameObjectId(uint32_t* ptr);
+
+  const Magnum::Vector2i framebufferSize() const;
```
mosra: The first `const` is unneeded, I think :)

src/deps/gl_tensor/README.md
```diff
@@ -0,0 +1,74 @@
...
+    cd test
+    python setup.py build
+    python setup.py install
+
+Library paths to OpenGL and EGL may need to be added to `LD_LIBRARY_PATH` env variable, on devfair machines, it looks something like  
+    
+    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib/x86_64-linux-gnu/nvidia-opengl:/usr/lib/x86_64-linux-gnu
+
+## Limitations
+1. The plugin depends on Pytorch 0.4.1. Pytorch 1.0 has different C++ API for creating tensors, so the plugin needs to be updated.
```
jia-git: It seems to work with Pytorch 1.0, so this line can be removed.

src/esp/gfx/RenderingTarget.cpp
erikwijmans: Whoops!  Looks like I forgot to put a fixme to find the correct Magnum API :)

src/esp/gfx/RenderingTarget.cpp
erikwijmans: height, width is common when treating images an array (which is super common for computer vision people) -- I originally had it as a vector, but kept in passing `[height, width]` and breaking things....

I can change to width, height in the constructor, but I will keep them separate so I don't have to debug that again :)

src/deps/gl_tensor/README.md
erikwijmans: Yep, I updated it to work with PyTorch 1.0+

src/esp/gfx/RenderingTarget.cpp
erikwijmans: Maybe just Image2D -> python buffer?  It'll need to get into a numpy array at some point :)

src/esp/gfx/RenderingTarget.cpp
mosra: OK, was not aware of that. Then keep it in whatever form is the least surprising for the target audience :)

src/esp/gfx/RenderingTarget.cpp
mosra: What I have currently in-progress is reading directly into a preallocated numpy array (so fully zero-copy). `Image2D -> buffer` would still be at least one copy.

I'll submit this in a separate PR once I'm done.

src/esp/bindings/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ target_link_libraries(habitat_sim_bindings
 )
 
 if(WITH_GL_TENSOR)
-  target_compile_definitions(habitat_sim_bindings PRIVATE __ESP_WITH_GL_TENSOR__)
+  target_compile_definitions(habitat_sim_bindings PRIVATE __esp_with_gl_tensor__)
```
mosra: Eh, sorry, should have mentioned that double underscores *anywhere* are reserved too :sweat_smile: [Specs.](https://en.cppreference.com/w/cpp/language/identifiers)

src/esp/gfx/Renderer.cpp
```diff
@@ -23,109 +23,29 @@ namespace esp {
...
     camera.draw(drawables);
-    renderExit();
   }
 
   void draw(sensor::Sensor& visualSensor, scene::SceneGraph& sceneGraph) {
     ASSERT(visualSensor.isVisualSensor());
 
     // set the modelview matrix, projection matrix of the render camera;
     sceneGraph.setDefaultRenderCamera(visualSensor);
+    sceneGraph.getDefaultRenderCamera().getMagnumCamera().setViewport(
```
bigbike: Fancy method chaining.

Maybe we can upgrade the "setDefaultRenderCamera" in the SceneGraph.h to
void setDefaultRenderCamera(sensor::Sensor& sensor, Magnum::Vector2i* framebufferSizePtr = nullptr);
In this way, you can hide the getMagnumCamera ()... inside this function with:

if (nullptr != framebufferSizePtr) {
// set the frame buffer size to the default camera;
}

It would be more convenient for the other developers.

src/esp/gfx/RenderingTarget.cpp
bigbike: I agree with @mosra . W x H is the convention for computer graphics (e.g., 1920 x 1080). Also please note that we have "width" ahead of "height" almost everywhere in the C++ backend. I am not against using H x W, but please be careful when you work on the code bridging the python and C++.

src/esp/gfx/Renderer.cpp
bigbike: Another question:
The default camera is a state machine I believe. Once we set its resolution, do we need to restore the default value of it?

src/esp/gfx/Renderer.cpp
```diff
@@ -23,109 +23,29 @@ namespace esp {
...
+        visualSensor.renderingTarget()->framebufferSize());
 
     draw(sceneGraph.getDefaultRenderCamera(), sceneGraph.getDrawables());
```
bigbike: Maybe a bit out of topic, a couple questions to @mosra related to "instancing" :
Can a drawable be shared by a couple of scene nodes? (Background, e.g., there are 20 identical chairs in the scene, and I would like to load the mesh just once, create just 1 drawable, and make it shared among 20 scene nodes.)
Is there a way to quickly clone a sub-SceneGraph (along with all the attached drawables)? (e.g., a sub-SceneGraph representing a humanoid robot)

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "magnum.h"
+
+using namespace Magnum;
+
+namespace esp {
+namespace gfx {
+struct RenderingTarget::Impl {
+  Impl(WindowlessContext::ptr context, int height, int width)
+      : context_{context},
+        framebufferSize_(width, height),
```
bigbike: Be careful, it is w x h here.

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -0,0 +1,190 @@
...
+                     {2, GL::Framebuffer::ColorAttachment{2}}});
+    CORRADE_INTERNAL_ASSERT(
+        framebuffer_.checkStatus(GL::FramebufferTarget::Draw) ==
+        GL::Framebuffer::Status::Complete);
+  }
+
+  void renderEnter() {
+    framebuffer_.bind();
+    framebuffer_.clear(GL::FramebufferClear::Color |
+                       GL::FramebufferClear::Depth);
```
bigbike: Any other buffer need to be cleaned as well?

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -0,0 +1,190 @@
...
+    framebuffer_.clearColor(2, Vector4ui{});
+    framebuffer_.bind();
+  }
+
+  void renderExit() { GL::defaultFramebuffer.bind(); }
+
+  gltensor::GLTensorParam::ptr glTensorParam() const {
+    auto param = std::make_shared<gltensor::GLTensorParam>();
+
+    param->height_ = framebufferSize_[1];
```
bigbike: You may consider:
define the following two in the core/esp.h and use them everywhere in the code.
const int WIDTH = 0;
const int HEIGHT = 1;

src/esp/gfx/Simulator.h
```diff
@@ -61,8 +61,10 @@ class Simulator {
 
   void saveFrame(const std::string& filename);
 
+  RenderingTarget::ptr createRenderingTarget(int height, int width);
```
bigbike: I am not a broken radio, but I still would like to suggest using the convention W x H so that it is uniform everywhere in the C++ backend (including Magnum), which is less error-prone in future development.

src/esp/sensor/Sensor.cpp
```diff
@@ -53,12 +56,33 @@ void Sensor::setTransformationFromSpec() {
...
 }
 
+gltensor::GLTensorParam::ptr Sensor::glTensorParam() const {
+  if (tgt_ == nullptr)
+    throw std::runtime_error("Sensor has no rendering target");
+
+  switch (spec_->sensorType) {
+    case SensorType::COLOR:
+      return tgt_->glTensorParamRgba();
+
```
bigbike: break; // even it is not necessary, it is good to have it to make the code less error-prone in the future.

src/esp/sensor/Sensor.cpp
```diff
@@ -53,12 +56,33 @@ void Sensor::setTransformationFromSpec() {
...
+    case SensorType::DEPTH:
+      return tgt_->glTensorParamDepth();
+
```
bigbike: break;

src/esp/sensor/Sensor.cpp
```diff
@@ -53,12 +56,33 @@ void Sensor::setTransformationFromSpec() {
...
+    case SensorType::SEMANTIC:
+      return tgt_->glTensorParamId();
+
```
bigbike: break;

src/esp/sensor/Sensor.cpp
```diff
@@ -53,12 +56,33 @@ void Sensor::setTransformationFromSpec() {
...
+    default:
+      throw std::runtime_error(
+          "GLTensor only suppots color, depth, and semantic sensors");
```
bigbike: break;

src/esp/gfx/RenderingTarget.cpp
erikwijmans: Yes, it's w x h here as that is what magnum assumes.

src/esp/gfx/RenderingTarget.cpp
erikwijmans: I used H x W here as we pass the sensor resolutions as H x W https://github.com/facebookresearch/habitat-sim/blob/master/examples/settings.py#L43

src/esp/gfx/Renderer.cpp
erikwijmans: Made `setDefaultRenderCamera` check to see if the sensor has a rendering target and if it does, set the framebuffer size on the camera

src/esp/gfx/RenderingTarget.cpp
bigbike: > I used H x W here as we pass the sensor resolutions as H x W https://github.com/facebookresearch/habitat-sim/blob/master/examples/settings.py#L43

I will advise in pybindings which are exposed to the external users (main audience perhaps are from CV community), you use H X W, while in the C++ backend infra (the simulator), you use W X H. 

Does it make sense?

src/esp/gfx/RenderingTarget.cpp
erikwijmans: It differs if we are using graphics or computer vision conventions.  Since we interact with both all the time, it's gonna differ all the time.  i.e the numpy arrays and GL Tensor (which is targeting CV people) have things in h x w format.

src/esp/gfx/RenderingTarget.cpp
erikwijmans: I believe that it is  (this `renderEnter()` function has been here the whole time, just use to be on the renderer).

erikwijmans:
Swapped things to W x H format (I said H x W in the comment as that is the convention that is ingrained into my head....)

src/esp/gfx/Renderer.cpp
bigbike: > Made `setDefaultRenderCamera` check to see if the sensor has a rendering target and if it does, set the framebuffer size on the camera

I have another question above that you may miss.
Suppose the resolution was set by Sensor A previously. Now Sensor B is going to render something, which does not have a buffer target and would like to render it in the default buffer, would the resolution need to be reset in that case?

erikwijmans:
gpu2gpu_vs_cpu
@dhruvbatra

src/esp/gfx/RenderingTarget.cpp
erikwijmans: @mosra -- Is there any reason why I should set the frame-buffer back to default?  It actually has a non-trivial impact on performance to do so for some reason.

src/esp/gfx/Renderer.cpp
erikwijmans: I guess maybe querying the sensor's spec to figure out the resolution makes more sense?  Also, AFAIK, there currently isn't a default framebuffer.

src/esp/gfx/Renderer.cpp
mosra: > Can a drawable be shared by a couple of scene nodes?

No. But a couple of drawables can reference the same mesh/shader/texture multiple times, avoiding the need to load them more than once.

> Is there a way to quickly clone a sub-SceneGraph (along with all the attached drawables)? 

We discussed this with @hzyjerry a couple days ago on Slack, see there for details. Current plan is that he'll make it as part of #91, I think?

src/esp/gfx/RenderingTarget.cpp
mosra: Why not just `framebufferSize.y()`? No constants needed.

src/esp/sensor/Sensor.cpp
mosra: :thinking: I fear that may cause some compilers to complain about unreachable code, don't remember which tho.

src/esp/gfx/Simulator.h
```diff
@@ -62,6 +62,10 @@ class Simulator {
   void saveFrame(const std::string& filename);
 
   RenderingTarget::ptr createRenderingTarget(int width, int height);
+  inline RenderingTarget::ptr createRenderingTarget(
+      const Magnum::Vector2i& size) {
+    return createRenderingTarget(size[0], size[1]);
+  }
```
mosra: Why not have just the `Vector2i` overload? On the call site it's just about doing `create_rendering_target([width, height])` if you have those separate (or `{}` in case of C++). Magnum does the same, following Python's philosophy of "provide only one way to do things" there's never separate vector/scalar overloads.

Then I would change `createRenderingTarget()` and `RenderingTarget` constructor to also take `Vector2i` instead of two scalars, so it's consistent and one doesn't need to pack/unpack the components every time.

Btw, `inline` inside class definition is redundant, that's implicit ([spec](https://en.cppreference.com/w/cpp/language/inline), see the second paragraph).

src/esp/sensor/Sensor.h
```diff
@@ -78,6 +78,10 @@ class Sensor : public Magnum::SceneGraph::AbstractFeature3D {
 
   SensorSpec::ptr specification() const { return spec_; }
 
+  inline Magnum::Vector2i framebufferSize() const {
+    return {spec_->resolution[1], spec_->resolution[0]};
```
mosra: I would put a **big** comment here pointing out the order. Because it took me an awful lot of time to realize that :)

src/esp/gfx/RenderingTarget.cpp
mosra: I don't know, you tell me, it was your code at first :) If I see correctly, the previous implementation didn't have this?

Switching framebuffers has always a big perf impact, yes, so ideally you want to have as little of those as possible. Unless something else later in the code *expects* that the default framebuffer is bound, I would say switching back is not needed. Or there could be an explicit binding of the default framebuffer later on the python side.

Thinking further -- because the framebuffers differ *only* in their size, why not have just one global framebuffer (like you had previously), that's as large as the largest sensor and then render / read just a part of it? All Magnum APIs are able to work with framebuffer subranges, but not sure if GLTensor can.

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -0,0 +1,193 @@
...
+        GL::Framebuffer::Status::Complete);
+  }
+
+  void renderEnter() {
+    framebuffer_.bind();
+    framebuffer_.clear(GL::FramebufferClear::Color |
+                       GL::FramebufferClear::Depth);
+    framebuffer_.clearColor(1, Vector4{});
+    framebuffer_.clearColor(2, Vector4ui{});
+    framebuffer_.bind();
```
mosra: There's a lot of (usually quite slow) extra clearing done here. The `clear()` clears all attachments that match the flags, so the following calls are not needed, alternatively you could also `clearDepth()` and remove the `clear()`.

The double `bind()` is not needed either, once is enough.

src/esp/gfx/RenderingTarget.cpp
erikwijmans: Yep, doesn't seem like there is a need to bind the default framebuffer :)

RE 1 big framebuffer:  This is something that I had considered, but the ownership problem for CUDA-GL interop becomes difficult to manage (i.e. who register a given renderbuffer with CUDA-GL).  CUDA-GL also doesn't directly support reading only part of a framebuffer so we'd have to do the slicing ourselves.

habitat_sim/simulator.py
```diff
@@ -247,23 +273,32 @@ def get_observation(self):
...
-                self._buffer.reshape(
-                    (
-                        self._spec.resolution[0],
-                        self._spec.resolution[1],
-                        self._spec.channels,
-                    )
-                ),
-                axis=0,
-            ).copy()
+        with self._sensor_object.rendering_target as tgt:
```
erikwijmans: What are people thoughts about getting rid of the clones/copies here?  They are kinda slow (especially the one on the GPU tensor) and for a lot of use-cases, they aren't needed (for RL training, we just immediately copy the image into a big tensor of images).

src/esp/gfx/RenderingTarget.cpp
```diff
@@ -108,29 +119,53 @@ struct RenderingTarget::Impl {
 
   void readFrameRgba(uint8_t* ptr) {
     framebuffer_.mapForRead(GL::Framebuffer::ColorAttachment{0});
+
+#if defined(CORRADE_TARGET_UNIX) && !defined(CORRADE_TARGET_APPLE)
+    framebuffer_.read(Range2Di::fromSize({0, 0}, framebufferSize_),
+                      colorPixelBuffer_, GL::BufferUsage::StreamRead);
+    auto src_ptr = colorPixelBuffer_.buffer().data();
```
mosra: The `.data()` does a copy into the CPU memory (and there's another copy inside the GPU memory before that from the FB to the PBO), so I'm not sure if this improves anything :) You might want to `map()` the buffer instead (don't forget to `unmap()` after), but since it's GL the driver might be doing a CPU memory copy underneath anyway.

PBOs should work on Mac without issues as well (or at least magnum's tests claim so), so no need for the `#ifdef` I think.

erikwijmans: Yeah, I've seen FBO->PBO->CPU be faster than FBO->CPU previously, so I thought I'd see if that was true here also.  Doesn't seem to matter.

erikwijmans:
gpu2gpu_vs_cpu
@dhruvbatra

dhruvbatra:
Thanks for the plot. A couple of request:
-- Can we get the x-axis going out to 1024. 
-- Can we run it 5 times and plot means+shaded error bars? I suspect that will remove the presumably spurious *increase* in speed from 128 to 256 for GPU->GPU.

erikwijmans:
gpu2gpu_vs_cpu
Yep, does seem like that jump was spurious

erikwijmans:
gpu2gpu_vs_cpu
With steps per second instead of frames per second

erikwijmans:
@mosra @jia-git @msavva @bigbike -- I took a stab at integrating gl-tensor directly into habitat-sim, that way there isn't a need to compile a second package.  I did this by making the rendering target also take a GPU device pointer and read directly into that.  PyTorch does export the device pointer of tensors to python, so this is a feasible solution.  It also more directly allows people to use other GPU tensor libraries if they'd like.  Thoughts?

msavva:
> @mosra @jia-git @msavva @bigbike -- I took a stab at integrating gl-tensor directly into habitat-sim, that way there isn't a need to compile a second package. I did this by making the rendering target also take a GPU device pointer and read directly into that. PyTorch does export the device pointer of tensors to python, so this is a feasible solution. It also more directly allows people to use other GPU tensor libraries if they'd like. Thoughts?

This is great imo -- it is far preferable to remove the overhead of dealing with a separate package, and the added generality for alternative interop implementations is an added bonus.

src/esp/bindings/bindings.cpp
```diff
@@ -382,6 +348,73 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
...
+          [](RenderingTarget& self,
+             Eigen::Ref<Eigen::Matrix<uint32_t, Eigen::Dynamic, Eigen::Dynamic,
+                                      Eigen::RowMajor>>& img) {
+            self.readFrameObjectId(img.data());
+          },
+          py::arg("img").noconvert(), R"()")
+#ifdef ESP_WITH_GPU_GPU
+      .def("read_frame_rgba_gpu",
+           [](RenderingTarget& self, size_t devPtr) {
+             self.readFrameRgbaGPU(reinterpret_cast<uint8_t*>(devPtr));
```
erikwijmans: Just wanted to draw attention to this wonkiness -- This is the pattern needed for passing pointers from python to c++.

erikwijmans:
Merged in #120 

I also have decided to get rid of the extract copy calls as they aren't needed for the RL workload and the extra copy on the pytorch tensor is really quite costly.

mosra:
I'm really interested in the updated plot now :)

erikwijmans:
Never-mind.  We already have logic in habitat-api that fails if we don't do the copy :(

erikwijmans:
We should look into trying to fix that, but for now, copies it is.

mosra:
@erikwijmans how long will this need until it's ready to be merged into master? I'm asking because it contains a bunch of changes I could use on master as well (dependency updates etc.). Or should I cherry-pick those and submit separately?

erikwijmans:
It’s good to go on my end.

erikwijmans:
> Thinking about it a bit more, for the RenderTarget readout functions, the ideal would have been to re-use the existing functions and parameterize them on GPU-to-GPU readout mode vs not (and potentially other modes in the future) instead of increasing the API surface area

PyBind does support methods over-loaded by type, but I do like the explicit naming.  The GPU->GPU transfer functions just take a pointer, and it's unclear that that point _needs_ to be a CUDA device pointer otherwise.

mosra:
> It’s good to go on my end.

@erikwijmans In #140 everything is slightly newer, so if you're okay with that, I'd merge that first :)

> do you also have any opinions on the second point, from your prior experience?

@msavva you mean changing it from a compile-time error (if someone uses a GPU/GPU function while it's not compiled) to a runtime error, everything looking seemingly good until you call the function, and then throwing/asserting/aborting? Not sure if that's a good idea. In Magnum I'm always preferring compile-time errors, the most prominent case is when you port from desktop to WebGL -- as soon as your code starts to compile, it will most probably also work correctly. With those being runtime erros instead, porting would become a nightmare.

msavva:
> @msavva you mean changing it from a compile-time error (if someone uses a GPU/GPU function while it's not compiled) to a runtime error, everything looking seemingly good until you call the function, and then throwing/asserting/aborting? Not sure if that's a good idea. In Magnum I'm always preferring compile-time errors, the most prominent case is when you port from desktop to WebGL -- as soon as your code starts to compile, it will most probably also work correctly. With those being runtime erros instead, porting would become a nightmare.

Good point about run-time vs compile-time errors.  Agree that compile-time is far preferable to run-time.  The distinction I was trying to make was more about conditional variation in the user-facing API exposed through the header files (i.e. what functions are declared vs not).  I think it makes sense to minimize conditional variation in the header files (while still having compile-time errors as appropriate in the actual implementation.

mosra:
> The distinction I was trying to make was more about conditional variation in the user-facing API exposed through the header files (i.e. what functions are declared vs not).

This can be done quite cleanly with a generated `configure.h` file (I was suggesting this for the recently merged WebGL stuff as well) -- even if we try really hard, at some point it will inevitably happen that we need to have something conditionally enabled in the header. Almost every project I can think of has such a file and it's not a hurdle to users at all (compared to forcing them to explicitly pass `-DSOME_THING_ENABLED` to the compiler, for example).

> I think it makes sense to minimize conditional variation in the header files (while still having compile-time errors as appropriate in the actual implementation.

Assuming the declaration will be there but not the implementation, when users try to call such functions, they will only get a linker error (and open an issue here, thinking the project is broken). This should be a compile error instead, with the declaration (or anything related to it) not being present either.

msavva:
> This can be done quite cleanly with a generated `configure.h` file (I was suggesting this for the recently merged WebGL stuff as well) -- even if we try really hard, at some point it will inevitably happen that we need to have something conditionally enabled in the header. Almost every project I can think of has such a file and it's not a hurdle to users at all (compared to forcing them to explicitly pass `-DSOME_THING_ENABLED` to the compiler, for example).

Fully support that -- @erikwijmans are you game to try out using the `configure.h` approach in this PR? 🙂 

> Assuming the declaration will be there but not the implementation, when users try to call such functions, they will only get a linker error (and open an issue here, thinking the project is broken). This should be a compile error instead, with the declaration (or anything related to it) not being present either.

Good point again -- I was somehow assuming that both declaration and implementation would exist, but that the implementation would internally throw a compile-time error...

erikwijmans:
@mosra I am not entirely certain what a `configure.h` file allows us to do here. What would go in the `configure.h`?  Would I generate the header for `RenderTarget.h` conditionally?

erikwijmans:
> Fully support that -- @erikwijmans are you game to try out using the configure.h approach in this PR? 🙂

I will open a second PR for doing that

erikwijmans:
> RenderTarget is a much more widely used form for the introduced RenderingTarget class (enough that there's a wikipedia entry 🙂). I'd change the name to the former, and as an added bonus we need four fewer characters.

@msavva  done.  I will do the `configure.h` thing in a separate PR.

src/esp/core/Buffer.h
```diff
@@ -33,14 +34,14 @@ class Buffer {
...
-  virtual ~Buffer() { dealloc(); }
+  virtual ~Buffer() { dealloc(); };
 
  protected:
   void alloc();
   void dealloc();
 
  public:
-  void* data = nullptr;
+  Corrade::Containers::Array<char> data;
```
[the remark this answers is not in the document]
msbaines: SGTM. RAII is definitely preferred to raw pointers.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    checkCudaErrors(cudaGraphicsMapResources(1, &colorBufferCugl_, 0));
+
+    cudaArray* array = nullptr;
+    checkCudaErrors(
+        cudaGraphicsSubResourceGetMappedArray(&array, colorBufferCugl_, 0, 0));
+    const int widthInBytes = framebufferSize().x() * 4 * sizeof(uint8_t);
+    checkCudaErrors(cudaMemcpy2DFromArray(devPtr, widthInBytes, array, 0, 0,
+                                          widthInBytes, framebufferSize().y(),
+                                          cudaMemcpyDeviceToDevice));
```
msbaines: Have you considered using CUDA's EGL interoperability? Using EGLImage or EGLStream, would avoid even the GPU-GPU copy. EGLStream would even potentially save on some context switching overhead. It would also be a more general solution allowing you to plug into more than just CUDA.

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EGL.html

src/esp/gfx/Renderer.cpp
```diff
@@ -86,39 +41,9 @@ struct Renderer::Impl {
...
   }
-
-  void readFrameRgba(uint8_t* ptr) {
-    framebuffer_.mapForRead(GL::Framebuffer::ColorAttachment{0});
-    Image2D rgbaImage =
```
msbaines: It would be cool if Image2D, or something like it, could wrap an EGLImage. You could then keep the pixels in GPU memory and only read into CPU memory if the user asked for a CPU pointer. Similar to CVPixelBuffer's CVPixelBufferLockBaseAddress.

https://developer.apple.com/documentation/corevideo/cvpixelbuffer
https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglCreateImage.xhtml

src/esp/gfx/RenderTarget.cpp
erikwijmans: I haven't looked into this before, but I believe as long as we are using OpenGL, we absolutely need a GPU->GPU transfer as OpenGL makes no guarantees that it won't move the memory of frame buffers when you aren't actively reading (and you can't render when you are doing so).  The CUDA docs also seem to support this with the `Note that mapping of EGL resources is performed with the graphics API agnostic, resource mapping interface described in Graphics Interopability.` statement.

src/esp/gfx/Renderer.cpp
mosra: There's [GL::BufferImage](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1BufferImage.html), wrapping a PBO, but that's probably not very interoperable with EGL either?

I have to admit I don't have much experience with EGLImages, need to update my knowledge first :)

msbaines: > There's [GL::BufferImage](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1BufferImage.html), wrapping a PBO, but that's probably not very interoperable with EGL either?
> 

Cool. I've never actually used PBOs before but it looks like they allow you to copy GPU memory into the CPU address space in an efficient way.

EGLImage lets you create a reference (not a copy) to an existing GL object (Texture, RenderBuffer) that you can share with another EGL client. You can also share an EGLImage between processes using a platform-specific EGL extension or API (e.g. https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_image_pixmap.txt)

To actually access the EGLImage in CPU memory, you'd need to do a copy.  You'd need to import the EGLImage as a RenderBuffer or Texture and then glReadPixels() or maybe use a PBO to access the data. EGLImage is useful if you want to pass the data by reference to another EGL client (CUDA, another EGL rendering context, etc).

https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_image_base.txt
https://www.khronos.org/registry/OpenGL/extensions/OES/OES_EGL_image.txt
https://docs.nvidia.com/cuda/cuda-for-tegra-appnote/index.html

erikwijmans:
I can move the commit to deal with all the warnings introduced in #91 to a different PR if needed, but I needed to do that here as my linter was freaking out, making it hard to debug.

erikwijmans:
I have changed the way the the CUDAGL-Interop gets built to rely solely on CUDA (I am using APIs that are supported by, at least, CUDA>=8.0 so that seems fine) -- there will also be more CUDA things going forward, so soon this won't be the only thing that uses CUDA.  One question:  Should building with CUDA be the default if the system has it, or should you have to specify `--with-cuda` on the build?  I can see arguments both ways, currently it'll build with CUDA if you have it and you use `--no-cuda` to turn CUDA off even if you have it.

src/esp/sensor/Sensor.h
```diff
@@ -95,6 +97,16 @@ class Sensor : public Magnum::SceneGraph::AbstractFeature3D {
 
   SensorSpec::ptr specification() const { return spec_; }
 
+  //! Returns the size of the framebuffer corresponding to the sensor's
+  //! resolution
+  Magnum::Vector2i framebufferSize() const {
+    // NB: The sensors resolution is in H x W format as that more cleanly
```
msavva: Typos:
`sensors` -> `sensor's`
`pratice` -> `practice`
`and DL` -> `and DL.`
`expected` -> `expect`

src/esp/sim/SimulatorWithAgents.cpp
```diff
@@ -87,6 +87,12 @@ agent::Agent::ptr SimulatorWithAgents::addAgent(
 
   auto& agentNode = agentParentNode.createChild();
   agent::Agent::ptr ag = agent::Agent::create(agentNode, agentConfig);
+
+  // Add render targets to the agents sensors
```
msavva: `agents` -> `agent's`

habitat_sim/simulator.py
```diff
@@ -147,9 +156,13 @@ def semantic_scene(self):
         return self._sim.semantic_scene
 
     def get_sensor_observations(self):
+        for _, sensor in self._sensors.items():
```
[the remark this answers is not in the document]
erikwijmans: Since this PR uses as many framebuffers as there are sensors, the draw calls for each sensor are now independent.  Breaking the the draw and read into two separate blocks allows OpenGL to do a better job scheduling them.

erikwijmans:
Addressed some of @msavva's comments.  I am happy to move the RenderTarget logic into the renderer if others also agree.  I don't have a real preference either way.

src/esp/core/Buffer.h
```diff
@@ -26,12 +27,12 @@ enum class DataType {
 
 class Buffer {
  public:
-  explicit Buffer(){};
+  explicit Buffer() {}
   explicit Buffer(const std::vector<size_t> shape, const DataType dataType) {
```
mosra: Oh, the `Buffer` is still here? Somehow I was under the impression that it got removed/replaced already :)

src/esp/core/Buffer.h
```diff
@@ -40,7 +41,7 @@ class Buffer {
   void dealloc();
 
  public:
-  void* data = nullptr;
+  Corrade::Containers::Array<uint8_t> data;
   size_t totalBytes = 0;
```
mosra: The `totalBytes` is redundant now, `data.size()` contains the same information.

src/esp/core/CMakeLists.txt
```diff
@@ -38,4 +45,4 @@ target_link_libraries(core
 
 target_include_directories(core
   PUBLIC
-    ${PROJECT_BINARY_DIR})
+    "${PROJECT_BINARY_DIR}")
```
mosra: Minor thing, but CMake doesn't behave like Bash in this case and correctly preserves spaces inside variables. So the quotes are unnecessary, the same as they would be unnecessary in the above `configure_file()`.

src/esp/gfx/DepthUnprojection.h
```diff
@@ -65,8 +73,13 @@ class DepthShader : public Magnum::GL::AbstractShaderProgram {
    */
   DepthShader& bindDepthTexture(Magnum::GL::Texture2D& texture);
 
+  /**
+   * @brief The flags passed to the Constructor
+   */
+  const Flags& flags() const { return flags_; }
```
mosra: Could be just `Flags` returned. It's just a fancy integer, after all.

src/esp/gfx/DepthUnprojection.h
```diff
@@ -2,6 +2,8 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+#pragma once
+
```
mosra: Oh no, I forgot a header guard? :scream:

src/esp/gfx/Simulator.h
```diff
@@ -111,27 +123,27 @@ class Simulator {
...
+  double getWorldTime();
 
  protected:
-  Simulator() {}
-  std::unique_ptr<WindowlessContext> context_ = nullptr;
+  Simulator(){};
+
+  WindowlessContext::uptr context_ = nullptr;
   std::shared_ptr<Renderer> renderer_ = nullptr;
+  std::unique_ptr<DepthShader> depthShader_ = nullptr;
```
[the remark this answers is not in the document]
mosra: :+1: to what @msavva says -- Simulator shouldn't need to care about specifics of rendering, Renderer feels like a much better place.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -0,0 +1,296 @@
...
+  void renderEnter() {
+#ifdef CORRADE_TARGET_EMSCRIPTEN
+    framebuffer_.clearDepth(1.0);
+    framebuffer_.clearColor(0, Magnum::Color4{});
+    framebuffer_.clearColor(1, Magnum::Vector4ui{});
+#else
+    framebuffer_.clear(GL::FramebufferClear::Color |
+                       GL::FramebufferClear::Depth);
+    framebuffer_.clearColor(1, Magnum::Vector4ui{});
+#endif
```
mosra: Forgive me if I forgot already, but what was the reason for doing this? Is the second variant faster on desktop?

src/esp/gfx/CMakeLists.txt
```diff
@@ -76,6 +78,19 @@ target_include_directories(gfx
     "${DEPS_DIR}/glad/include"
 )
 
+if(BUILD_WITH_CUDA)
+  find_package(CUDA 8.0)
+  if(CUDA_FOUND)
+    target_include_directories(gfx PUBLIC
+      "${CUDA_INCLUDE_DIRS}"
+      ${CMAKE_CURRENT_LIST_DIR}/cuda_helpers)
```
mosra: If I see correctly, the cuda headers are needed just in cpp files, so this could be `PRIVATE` instead :)

src/esp/gfx/CMakeLists.txt
```diff
@@ -76,6 +78,19 @@ target_include_directories(gfx
     "${DEPS_DIR}/glad/include"
 )
 
+if(BUILD_WITH_CUDA)
+  find_package(CUDA 8.0)
+  if(CUDA_FOUND)
```
mosra: I'm a fan of things either reliably working or reliably not working, so instead of silently building without CUDA when it's not found, I would

```cmake
if(NOT CUDA_FOUND)
    message(FATAL_ERROR "CUDA was not found. Either install it or disable BUILD_WITH_CUDA.")
endif()
```

... because otherwise the message about CUDA not being found gets lost somewhere wayy up in the log output and users wouldn't know that until they try using those APIs.

Same goes for the `ESP_BUILD_WITH_CUDA` done in `core/CMakeLists.txt` -- actually, I'd move this whole thing one level up so the finding and checking is done just on a single place.

src/esp/core/Buffer.h
mosra: IIRC on Slack we discussed using the [StridedArrayView](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1StridedArrayView.html) instead, since it's a superset of the features provided by this one, it's fully tested, has a conversion to any numpy array and is generally better integrated to Magnum. But I don't have the full context either, maybe @angelxuanchang and @msbaines could comment here? :)

In case it's not clear -- I'm not saying this PR should be taking any action on this, just got surprised it wasn't handled already elsewhere.

mosra: > StridedArrayView requires the shape of the data to be known at compile time

Not sure if we are talking about the same thing -- it only needs to know the dimensionality, *not* sizes. (There's StaticArrayView, which has compile-time sizes, but that's a different thing.) Dimensionality is, in this particular case, 2. When potentially used for audio (as @angelxuanchang mentioned on Slack) it would be 1, for volumetric data it would be 3, but that's it. Based on my experience, having dimensionality templated makes C++ code much easier to reason about, with no need for allocting an unbounded `std::vector` holding the size, with no need to check for its size at runtime, similarly as having a distinct type for two- and three-component vectors is beneficial. And I don't think this makes bindings harder -- for things with different dimension count you'd need a completely different set of APIs anyway, so why not a type as well.

mosra: I see, that's a valid concern. :+1: 

A different idea, then -- what about always using the max dimension count, so 3D? Then there's no templating not inheritance hierarchy needed; if the observation is one-dimensional, the second and third dimension is always one; if the observation is two-dimensional, the third dimension is always one. Furthermore, on the C++ side, converting between different dimensions is just a matter of calling [slice&lt;N&gt;()](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1StridedArrayView.html#a66c3b7ae7c67cad7896b68202ebff081), so if some code would really expect just one/two dimensions, the conversion is an one-liner (of course with a check that nothing got lost when dropping the extra dimensions):

```cpp
Containers::StridedArrayView3D<float> observations;
Containers::StridedArrayView1D<float> samples = observations.slice<1>();
```

"Inflating" to more dimensions works similarly, setting size of the newly added dimensions to `1`.

mosra: I *knew* you'd point that out :stuck_out_tongue_winking_eye:  

The container can be "type-erased" (`char`, e.g.) and then cast to a concrete type at the point of use. That's what [Image::pixels()](https://doc.magnum.graphics/magnum/classMagnum_1_1Image.html#a2fa5831183fcd118cb3e4040a258e895) does already, for example.

PR #117, opened 2019-07-24
Sample from truncated normal

erikwijmans:
## Motivation and Context

After some more chats with people, we've decided two things make more sense:

* Sample the noise from a truncated normal such that the noise never exceeds 3 sigma
* Sample the noise such that robot always moves at least a little bit in its intended direction. i.e. doesn't turn left when it mean to turn right.  I've set this little bit to mean at least 5% of its intended amount as this allows all actuation amounts to still be well defined.  For actuations without too much noise, large actuation amounts will pretty quick get into the 3 sigma regime, so noise will go back to being independent of actuation amount

This involves adding a dependency on scipy as I need to be able to sample from a truncated normal.  I can make this a soft dependency if this feels like to big of a dependency.

## How Has This Been Tested

Via tests.  I increased the actuation amount in the unit test as the assumption that the noise is approximately gaussian is now only true for reasonably large actuation amounts.

The rollouts image looks very similar to as before:
noisy_paths

## Types of changes

### DIFF
--- a/habitat_sim/agent/controls/pyrobot_noisy_controls.py
+++ b/habitat_sim/agent/controls/pyrobot_noisy_controls.py
@@ -8,6 +8,7 @@
 import attr
 import magnum as mn
 import numpy as np
+import scipy.stats
 
 import habitat_sim.bindings as hsim
 from habitat_sim import utils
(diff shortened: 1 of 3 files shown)


danielgordon10:
What are the red lines outside the boundaries?

erikwijmans:
When the agent slides along a wall, it can slide around a corner.

PR #119, opened 2019-07-25
remove notebook

abhiskk:
## Motivation and Context

Remove jupyter notebook as it has been moved under documentation (refer [here](https://github.com/facebookresearch/habitat-api/pull/155))


setup.cfg
```diff
@@ -6,6 +6,6 @@ build_temp=build
 
 [tool:pytest]
 addopts = --verbose -rsxX -q --nbval-lax
```
erikwijmans: Also need to remove --nbval-lax here

.circleci/config.yml
```diff
@@ -189,7 +189,7 @@ jobs:
               pip install nbval
```
erikwijmans: We can also remove the `pip install nbval` here

msavva:
@abhiskk this PR looks good to land as soon as the unused `nbval` dependency is removed.

PR #120, opened 2019-07-25
Zero-copy transfer of image data

mosra:
## Motivation and Context

Following [the discussion on #114](https://github.com/facebookresearch/habitat-sim/pull/114#discussion_r305228969) this updates the code to avoid an unnecessary memcpy of each frame. I set the `gpu-gpu` branch as base so you see just my changes here, not all of #114 as well.

I was not sure what should be exposed to Python and what not, exposing the whole framebuffer in order to make it possible for the user to call [gl.Framebuffer.read()](https://doc.magnum.graphics/python/magnum.gl.Framebuffer.html#read-2c68b) seemed *a bit* much, OTOH exposing just a [MutableArrayView](https://doc.magnum.graphics/python/corrade.containers.MutableArrayView.html) felt too typeless, so I'm exposing [MutableImageView](https://doc.magnum.graphics/python/magnum.MutableImageView2D.html), which contains size and format metadata as well. I'm of course open for suggestions.

Besides that, what you get as a side effect of updating the submodules is:

- up to 2x faster conversion between magnum math types (and array views) and numpy thanks to buffer protocol being now implemented in pure C Python APIs instead of the allocation-happy pybind11 wrapper, together with a lot of related tech debt cleaned up 
- all image view-related APIs exposed to Python (except general image import/export, that'll come in the next batch)
- all framebuffer-related APIs exposed to Python (except textures, that'll come in the next batch)

I'm still not sure about the additonal `.copy()` done on the Python side, however even with that I'm interested in how it benchmarks against the previous code.

## How Has This Been Tested

Pytest is happy on my machine :heavy_check_mark:  ... however test data download on the CI is failing with 403?

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -8,6 +8,7 @@ import os.path as osp
 from typing import Dict, List, Optional
 
 import attr
+import magnum as mn
 import numpy as np
 
 import habitat_sim.bindings as hsim
@@ -297,14 +298,23 @@ class Sensor:
             if self._spec.gpu2gpu_transfer:
                 return self._gl_tensor.tensor().flip(0).squeeze(-1).clone()
             else:
+                size = self._sensor_object.framebuffer_size
                 if self._spec.sensor_type == hsim.SensorType.SEMANTIC:
-                    tgt.read_frame_object_id(self._buffer)
+                    tgt.read_frame_object_id(
+                        mn.MutableImageView2D(mn.PixelFormat.R32UI, size, self._buffer)
+                    )
                     return np.flip(self._buffer, axis=0).copy()
                 elif self._spec.sensor_type == hsim.SensorType.DEPTH:
-                    tgt.read_frame_depth(self._buffer)
+                    tgt.read_frame_depth(
+                        mn.MutableImageView2D(mn.PixelFormat.R32F, size, self._buffer)
+                    )
                     return np.flip(self._buffer, axis=0).copy()
                 else:
-                    tgt.read_frame_rgba(self._buffer)
+                    tgt.read_frame_rgba(
+                        mn.MutableImageView2D(
+                            mn.PixelFormat.RGBA8UNORM, size, self._buffer
+                        )
+                    )
                     return np.flip(
                         self._buffer.reshape(
                             (
--- a/src/esp/bindings/bindings.cpp
+++ b/src/esp/bindings/bindings.cpp
@@ -22,6 +22,7 @@ using namespace py::literals;
 #include "esp/sensor/PinholeCamera.h"
 #include "esp/sensor/Sensor.h"
 
+#include <Magnum/Python.h>
 #include <Magnum/SceneGraph/Python.h>
 
 using namespace esp;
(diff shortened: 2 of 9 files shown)


habitat_sim/simulator.py
```diff
@@ -297,14 +298,23 @@ def get_observation(self):
             if self._spec.gpu2gpu_transfer:
                 return self._gl_tensor.tensor().flip(0).squeeze(-1).clone()
             else:
+                size = mn.Vector2i(self._spec.resolution[1], self._spec.resolution[0])
```
erikwijmans:
```suggestion
                size = self.__sensor_object.framebuffer_size
```

mosra: Ah! I *knew* it was there somewhere :)

erikwijmans: Yeah, it should probably be on the rendering target also :)

mosra: Ultimately the rendering target could expose the whole Framebuffer class, I think. Or (crazy thinking) be implemented completely in Python.

src/esp/bindings/bindings.cpp
```diff
@@ -359,38 +360,18 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
...
-            self.readFrameObjectId(img.data());
-          },
-          py::arg("img").noconvert(), R"()")
+          "Reads RGBA frame into passed img in uint8 byte format.")
+      .def("read_frame_depth",
+           [](RenderingTarget& self, const Magnum::PyImageView<2, char>& view) {
+             self.readFrameDepth(view);
+           })
+      .def("read_frame_object_id",
+           [](RenderingTarget& self, const Magnum::PyImageView<2, char>& view) {
```
erikwijmans: Why is this a `char` here?

habitat_sim/simulator.py
```diff
@@ -297,14 +298,23 @@ def get_observation(self):
             if self._spec.gpu2gpu_transfer:
                 return self._gl_tensor.tensor().flip(0).squeeze(-1).clone()
             else:
+                size = self.__sensor_object.framebuffer_size
```
erikwijmans:
```suggestion
                size = self._sensor_object.framebuffer_size
```

Looks like an additional underscore got added here (my bad!)

src/esp/bindings/bindings.cpp
mosra: The basic `ImageView<dimensions, T>` class uses `char`/`const char` to distinguish between mutable and immutable version. The C++ API has nice typedefs for all (`ImageView2D`, `MutableImageView2D` and so on) that but I didn't bother doing that for this python wrapper class because it's used only very rarely. (The `PyImageView` is needed in order to ensure the memory owner is correctly refcounted, btw.)

habitat_sim/simulator.py
mosra: Uh oh. Was tired and didn't spot the error either. Should have tested before pushing.

src/esp/bindings/bindings.cpp
erikwijmans: Nope, it doesn't seem like that's needed.  Just wanted to make sure `char` is correct given that the underlying datatype is `uint32_t` :)

mosra: The image itself is typeless (would be a template hell otherwise). For typed access there's the `pixels()` function ([C++ API](https://doc.magnum.graphics/magnum/classMagnum_1_1ImageView.html#a13bedd03f65287197a040b32ac51ac7f), [Python API](https://doc.magnum.graphics/python/magnum.MutableImageView2D.html#pixels)), returning a n-dimensional array view that you can then cast to a concrete type based on the actual pixel format. Also check the [related blog post](https://blog.magnum.graphics/backstage/multidimensional-strided-array-views/) :)

PR #124, opened 2019-07-30
Further improvements to python-side framebuffer reading

mosra:
## Motivation and Context

Another addition to #114, with `gpu-gpu` being the base branch. Again deleting more code than it adds, I'm slowly removing all code here, sorry :laughing: 

~~**Note:** the last commit in this PR is a bit controversional, so you might want to look at the [diff without this change](https://github.com/facebookresearch/habitat-sim/compare/gpu-gpu...gpu-gpu-additions^) first ;)~~ last commit removed, it unconvered a serious problem I will need to look at.

What this does:

- the Magnum Python bindings now have a better way to expose classes that need to refcount their dependencies (such as `ArrayView`, `ImageView` or `GL::Framebuffer`), meaning there are no `PyImageView` types and such and it's possible to just directly expose the classes to Python without any hassle.
- I went around all Clang warnings for the bindings project and fixed/suppressed them (including those annoying ones related to `py::self -= py::self`) -- you should get a cleaner build output now
- a bunch of other random fixes / updates (see commit messages for details)
- ~~In the last commit I wanted see how it would look when the framebuffer would be exposed all the way to python and all reading was done in `simulator.py`. Not sure if this is what you want or not, so feedback welcome! :)~~ will submit that separately

## How Has This Been Tested

Pytest is happy on my machine. ~~CI fails due to `GL::Context: no current context`? :scream:~~

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -209,7 +209,7 @@ class Sensor:
         self._sensor_object = self._agent.sensors.get(sensor_id)
         self._spec = self._sensor_object.specification()
 
-        # NB: Graphics APIs tend to use H x W, so use that here
+        # NB: Graphics APIs tend to use W x H, so use that here
         self._sensor_object.bind_rendering_target(
             self._sim.create_rendering_target(self._sensor_object.framebuffer_size)
         )
--- a/src/esp/bindings/bindings.cpp
+++ b/src/esp/bindings/bindings.cpp
@@ -22,6 +22,7 @@ using namespace py::literals;
 #include "esp/sensor/PinholeCamera.h"
 #include "esp/sensor/Sensor.h"
 
+#include <Magnum/ImageView.h>
 #include <Magnum/Python.h>
 #include <Magnum/SceneGraph/Python.h>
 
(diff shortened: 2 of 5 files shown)


erikwijmans:
Seems like `tgt.framebuffer.map_for_read` has an issue with the EGL context

erikwijmans:
Looks like calling anything method on `tgt.framebuffer` throws an error with the EGL context.

mosra:
OK, back to drawing board then (tested with GLX here and that worked) :sweat_smile: 

If you take everything except the last commit, does the rest look (and work) okay?

erikwijmans:
Huh.  It also crashes on MacOS for me :/

erikwijmans:
For the other commits, all looks good but some `framebufferSize_` needs to be replaced with `framebufferSize()` in the GPU_GPU transfer part.

mosra:
> some `framebufferSize_` needs to be replaced with `framebufferSize()`

Sorry about that, pushed a fixed version (again except the last commit, need to look into it).

mosra:
@erikwijmans I removed the last commit, will submit it again once I figure out a proper way to fix it. The rest should be pretty safe to merge :)

PR #132, opened 2019-08-06
Support Rendering Replica models in the simulator (under CONSTRUCTION)

bigbike:
## Motivation and Context
Initialize the PR to get early feedbacks.

Purpose: 
This PR is to upgrade the simulator so that it can render the Replica model released by FRL (https://github.com/facebookresearch/Replica-Dataset).

The master version can render the semantic scene of the replica model correctly. However, it cannot render the replica model (with the old ptex textures (.rgb format), or the new ones (.hdr format)) on either MacOS or Linux.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist


src/esp/assets/PTexMeshData.cpp
```diff
@@ -602,35 +616,56 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+
     std::cout << "\rLoading atlas " << iMesh + 1 << "/"
-              << renderingBuffers_.size() << "... ";
+              << renderingBuffers_.size() << " from " << rgbFile << "... ";
     std::cout.flush();
 
-    Cr::Containers::Array<const char, Cr::Utility::Directory::MapDeleter> data =
-        Cr::Utility::Directory::mapRead(rgbFile);
-    const int dim = static_cast<int>(std::sqrt(data.size() / 3));  // square
+    const size_t numBytes = io::fileSize(rgbFile);
```
bigbike: I would like to load the textures from the file. Such textures are not in common picture format (e.g., png, jpg, tga), but some customized format in binary.
In ReplicaSDK, they mapped a texture (rgbFile) to the memory, and uploaded it to a texture (atlas). The code is as follows.
     const size_t numBytes = std::experimental::filesystem::file_size(rgbFile);
     // Open file
     int fd = open(std::string(rgbFile).c_str(), O_RDONLY, 0);
     void* mmappedData = mmap(NULL, numBytes, PROT_READ, MAP_PRIVATE |
MAP_POPULATE, fd, 0);
     meshes[i]->atlas.Upload(mmappedData, GL_RGB, GL_UNSIGNED_BYTE);
     munmap(mmappedData, numBytes);
     close(fd);

@mosra : Do you think the current implementation is proper?
Do you have better ideas in terms of leveraging Magnum to do the same work?

src/esp/assets/PTexMeshData.cpp
```diff
@@ -602,35 +616,56 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+
+    // atlas
+    // the size of each image is dim x dim x 3 (RGB), which equals to numBytes
     Magnum::ImageView2D image(Magnum::PixelFormat::RGB8UI, {dim, dim}, data);
     renderingBuffers_[iMesh]
         ->tex.setWrapping(Magnum::GL::SamplerWrapping::ClampToEdge)
         .setMagnificationFilter(Magnum::GL::SamplerFilter::Linear)
         .setMinificationFilter(Magnum::GL::SamplerFilter::Linear)
-        // .setStorage(1, GL::TextureFormat::RGB8UI, image.size())
+        // .setStorage(1, Magnum::GL::TextureFormat::RGB8UI, image.size())
```
bigbike: @mosra : 
The code in the master version commented it out.
Do I need to set the storage here?

bigbike:
Another question:
 I quote your comment in the previous PR: "Looking at the sources, the only thing that requires GLSL 4.30 is the SSBO in atlas.glsl (which could be replaced with GL::BufferTexture, as is already done here?). So I'd say choosing GL 4.1/4.2 on both would make the code simpler?"
So my understandings are,

-  On macOS, we use 4.1 while on linux, we use 4.2;
-  If we use GL::bufferTexture to handle the SSBO in atlas.glsl, which requires 4.2, that implies rendering Replica models cannot be supported on macOS;

Do I understand it correctly?
Thank you very much!

dhruvbatra:
CC: @jstraub

src/esp/assets/PTexMeshData.cpp
```diff
@@ -602,35 +616,56 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
   for (size_t iMesh = 0; iMesh < renderingBuffers_.size(); ++iMesh) {
-    const std::string rgbFile =
-        atlasFolder_ + "/" + std::to_string(iMesh) + "-color-ptex.rgb";
-    if (!io::exists(rgbFile)) {
-      ASSERT(false, "Can't find " + rgbFile);
-    }
+    const std::string rgbFile = Corrade::Utility::Directory::join(
+        atlasFolder_, std::to_string(iMesh) + "-color-ptex.rgb");
+
+    ASSERT(io::exists(rgbFile), Error : Cannot find the rgb file);
```
mosra: There's [CORRADE_ASSERT()](https://doc.magnum.graphics/corrade/Assert_8h.html#a83f7361970111951c88f1564a4f148e8) btw, which could make `esp/logging.h` obsolete -- and it supports printing everything that [Utility::Debug](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html) can :) 

```suggestion
    CORRADE_ASSERT(io::exists(rgbFile), "Error : Cannot find the rgb file");
```

Also, for `io::exists` there's [Directory::exists()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1Directory.html#a0299a9e5b18b4187615c2fd8e256d586) ... and and and

... 🤔 should I open a PR moving everything from `esp/logging.h` and `io/io.h` to builtin Corrade functionality? :smile: (cc: @msavva @erikwijmans)

src/esp/assets/PTexMeshData.cpp
mosra: It was commited this way by @msavva in October 2018, I don't know what was the original reason :) Generally, calling `setSubImage()` without `setStorage()` or `setImage()` being called first will result in a GL error and nothing being done, so I guess that's what is happening here? (Unless the texture storage is being set up somewhere else, but I don't see any such code anywhere else.)

If you are on linux you can run the viewer with `--magnum-gpu-validation on` and it'll tell you in the console output about any GL errors. (Mac doesn't implement GL debug output, so there you'd need to test for `GL::Renderer::error()` manually by hand.)

mosra:
> If we use GL::bufferTexture to handle the SSBO in atlas.glsl, which requires 4.2, that implies rendering Replica models cannot be supported on macOS

No. SSBOs are since GL 4.3 and those aren't supported on macOS, as it's stuck on 4.1 + some extensions on top. `GL::BufferTexture` requires just GL 3.1, and it definitely *is* supported there. So if we go with 4.1 and `GL::BufferTexture` we're all good on Mac.

src/esp/assets/PTexMeshData.cpp
bigbike: Allow me to reply this one first before I read the other comments.
I found an issue in the previous code that the "dim" was wrong, which resulted from "numBytes". 

After reading your comment, assuming the mapRead() did exactly the same, then it means getting the "numBytes" using "data.size()" is incorrect.

What is the correct way to get this number?

I would change it back.

(I am not aware of the #106 or #84 since I was on leave for a couple of months. Good to know.)

Thank you very much for the comment.

src/esp/assets/PTexMeshData.cpp
bigbike: I am developing everything on the Mac. I guess I will have to require a linux box. :)

So here, if I understand you correctly, I should actively call setStorage(). Correct?

src/esp/assets/PTexMeshData.cpp
```diff
@@ -34,7 +38,10 @@ void PTexMeshData::load(const std::string& meshFile,
   ASSERT(io::exists(atlasFolder));
 
   // Parse parameters
-  const auto& paramsFile = atlasFolder + "/parameters.json";
+  // careful: when using "join", no leading forward slash in the file name.
+  // otherwise the corrade would think it is absolute path
```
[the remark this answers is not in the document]
bigbike: I left the comment here, because I removed the forward slash from the original code.

bigbike:
@jstraub:
I found "calculateAdjacency()" was time-consuming. I believe that is why the openmp is employed.
Can we pre-compute it (or compute it once on the 1st run) and store the adjacency to the disk?

src/esp/assets/PTexMeshData.cpp
mosra: Yes, or change the below to `setImage()` ... but `setStorage()` is the preferred thing to do.

src/esp/assets/ResourceManager.cpp
```diff
@@ -105,14 +105,21 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
...
+    const auto atlasDir = [=]()->std::string{
+      // backwards compatibility
+      if (Corrade::Utility::String::endsWith(filename, "ptex_quad_mesh.ply")) {
+        return (Corrade::Utility::String::stripSuffix(filename, "ptex_quad_mesh.ply") +
+        "ptex_textures");
+      }
+        // officially released Replica dataset
+        return (Corrade::Utility::String::stripSuffix(filename, "mesh.ply") +
+        "textures");
+    };
```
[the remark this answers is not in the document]
mosra: Oh well, sorry, missed that part, thought it's just cutting off a different filename :see_no_evil:

src/esp/assets/PTexMeshData.cpp
mosra: Huh. The array returned from `mapRead()` should have its size equal to number of bytes in that file (and my tests so far were confirming that). The rest of the code looks exactly the same, so that would mean `mapRead()` returns a wrong size?

What does `numBytes` give and what `data.size()`?

bigbike:
> > If we use GL::bufferTexture to handle the SSBO in atlas.glsl, which requires 4.2, that implies rendering Replica models cannot be supported on macOS
> 
> No. SSBOs are since GL 4.3 and those aren't supported on macOS, as it's stuck on 4.1 + some extensions on top. `GL::BufferTexture` requires just GL 3.1, and it definitely _is_ supported there. So if we go with 4.1 and `GL::BufferTexture` we're all good on Mac.

Good to know. This is great news. 
Before, I thought GL::BufferTexture would require 4.2.

mosra:
> I found "calculateAdjacency()" was time-consuming.

What does the function do, exactly? To me it looks like for every face it finds out IDs of adjacent faces (plus some orientation encoding after)? If that's so, that could be done without the hashmap, without nested vectors, in `O(n)` time and just one or two allocations instead of `~3n`. I did this very recently for [MeshTools::generateSmoothNormals()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a1055154900bcbc7d87b72d863f14ce54) and have it still fesh in my head, so I could extract the code out to a reusable utility that we could use here. (The code in question, for reference, is [here](https://github.com/mosra/magnum/blob/e2610bff18fec836ed2e93942a0b312244e413bf/src/Magnum/MeshTools/GenerateNormals.cpp#L97-L141).)

src/esp/assets/PTexMeshData.cpp
bigbike: Good to know. It is also possible my testing code has some bugs.
Allow me to take another look.

bigbike: Just tested it again. The numbers are identical. There must be something wrong in my previous debugging code. My bad. I sincerely apologize for the confusion. Sorry, Vladimir.

mosra: No problem ;) In any case, I'm not claiming my implementation is bug-free either, so if you see something strange with it, let me know.

bigbike:
> Have you seen [my comment](https://github.com/facebookresearch/habitat-sim/pull/132#issuecomment-519177744) ? :) I think the algorithm could be rewritten in a way that doesn't make it take ages to run (and thus no need to save/restore it), but I need a clarification that my assumption is correct.

Yes, I saw it, which is very cool if we can optimize their algorithm to O(N). But such code is not immediately available yet, and I have to debug and run the code hundreds of times a day. To ease my pain, such save/load functions are rather helpful at the moment.

We can treat them as temporary code here, and remove them later if they are indeed not necessary.
My opinion is that I prefer to save/load any data that can be pre-computed or pre-processed regardless the computation complexity.

src/esp/gfx/PTexMeshDrawable.cpp
```diff
@@ -18,16 +18,19 @@ PTexMeshDrawable::PTexMeshDrawable(
...
       tileSize_(ptexMeshData.tileSize()),
-      exposure_(ptexMeshData.exposure()) {}
+      exposure_(ptexMeshData.exposure()),
+      gamma_(ptexMeshData.gamma()),
+      saturation_(ptexMeshData.saturation()) {}
 
 void PTexMeshDrawable::draw(const Magnum::Matrix4& transformationMatrix,
                             Magnum::SceneGraph::Camera3D& camera) {
-  adjTex_.bind(1);
+  adjFaces_.bind(1);
```
bigbike: @mosra : adjFaces is the buffer texture, and it is bound to 1. Not sure if 1 is correct binding point.

src/shaders/ptex-default-gl410.frag
```diff
@@ -29,7 +29,7 @@ const uint FACE_MASK = 0x3FFFFFFF;
 
 int GetAdjFace(int face, int edge, out int rot) {
   // uint data = meshAdjFaces[face * 4 + edge];
-  uint data = uint(texelFetch(meshAdjFaces, face * 4 + edge));
+  uint data = texelFetch(meshAdjFaces, face * 4 + edge).r;
```
bigbike: @mosra : Is it the correct way to fetch the unsigned int in the buffer texture? 
BTW, please note: I changed the meshAdjFaces from sampleBuffer to usamplerBuffer.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -582,54 +660,75 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
 
-    currentMesh->adjTex.setBuffer(Magnum::GL::BufferTextureFormat::R32UI,
-                                  currentMesh->abo);
+    currentMesh->adjFaces.setBuffer(Magnum::GL::BufferTextureFormat::R32UI,
+                                    currentMesh->abo);
     currentMesh->abo.setData(adjFaces[iMesh],
                              Magnum::GL::BufferUsage::StaticDraw);
+
+    // experiment code (may not work):
+    // using GL_LINES_ADJACENCY here to send quads to geometry shader
```
bigbike: @mosra: I believe it is the most suspicious part.
The purpose is to send quads to geometry shader using GL_LINES_ADJACENCY.
The corresponding code in RelicaSDK is:
`
glDrawElements(GL_LINES_ADJACENCY, mesh.ibo.num_elements, mesh.ibo.datatype, 0);
`
Did I set it correctly using magnum?

mosra: Yes, I think so (does it work?).

src/shaders/ptex-default-gl410.frag
mosra: If the shader compiler doesn't complain, yes :)

src/esp/gfx/PTexMeshDrawable.cpp
mosra: I don't know -- I don't see any `layout(binding = N)`  on the shader side, or anything related in the constructor, for either this texture or the other, which leads me to think that this is the root of the problem. To be clear -- Magnum doesn't have any implicit binding points or anything, so you can't expect the shader to pick (samplers, uniforms, attributes) up unless you explicitly define the location / binding points for these as well. 

*(This comment is getting a bit long, so bear with me please.)*

What's the designated way of doing this in Magnum (and what you are doing here differently) is, step-by-step ([docs for reference](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1AbstractShaderProgram.html#GL-AbstractShaderProgram-subclassing) -- look there for more details about each point):

- For each uniform, defining `layout(location = N)` in the shader code, and then using the same `N` on C++ side. (Extensions needed for that are available on all systems we care about, so no need to save results of `uniformLocation()` on the C++ side.) Especially *not* using `setUniform(uniformLocation("some string"), value)` each time in every frame, all those string operations are extremely wasteful.
- For each input attribute, `typedef`ing it in the header with location corresponding to the `layout()` qualifier in the shader code *and* type matching the type you are most likely to have the vertex data in. GL pads the remaining components with `(0, 0, 0, 1)` so for example, if you pass a `Vector2` and have `vec4` on the shader side, you'll get `(x, y, 0, 1)` there -- no need to supply the `0` and `1` explicitly. Here you're using `Vector4` and I don't see a reason to supply the fourth component, so it could be `Vector3` I think, saving 25% of memory. Of course the `PTexMeshData` needs to be updated as well.
- For each sampler, either doing `layout(binding = N)` in the shader code or calling `setUniform(uniformLocation("sampler name"), N)` in the constructor. Here you need the latter, as macOS doesn't have `GL_ARB_shading_language_420pack`. Then, for each sampler providing `bindSomething()` APIs on the C++ side, using the same `N` as a parameter to `texture.bind()`. Here you have `bindTexture()` just for one of the samplers and it has the binding point as a parameter, which doesn't make sense, and in the other case you're calling `bind()` manually, with the `1` having no relation to the shader at all. What I'm usually doing is defining an enum with the binding point IDs and then using those enums for both the `bind()` and `setUniform()` calls -- so for example:

  ```cpp
  enum: UnsignedInt {
    AtlasTextureUnit = 0,
    AdjacencyTextureUnit = 1
  };

  ...

  PTexMeshShader::PTexMeshShader() {
    setUniform(uniformLocation("atlasTex"), AtlasTextureUnit);
    setUniform(uniformLocation("meshAdjFaces"), AdjacencyTextureUnit);
  }

  PTexMeshDhader& PTexMeshShader::bindAtlasTexture(GL::Texture2D& texture) {
    texture.bind(AtlasTextureUnit);
    return *this;
  }

  PTexMeshDhader& PTexMeshShader::bindAdjacencyTexture(GL::BufferTexture& texture) {
    texture.bind(AdjacencyTextureUnit);
    return *this;
  }
  ```

  This way the shader API on the C++ side is complete and you can be sure that you are always binding the right type of a texture to the right binding point. It's generally advised to choose mutually exclusive texture units for each shader (unless it's expected to use the same textures in each), as you save on texture rebinding when switching between different shaders. So e.g. if the generic shader makes use of units 0-3, the PTex shader would take units 4 and 5.

bigbike: Thank you so much for the informative and invaluable comment! It saves me from a lot of guesswork on Magnum usage.

src/shaders/ptex-default-gl410.frag
bigbike: The previous code will not be complained by the compiler either. But I doubt if that is correct. So I changed it.

bigbike:
Updated a version based on your comment.

The code is suffering segfault now. I am debugging it.

src/esp/gfx/PTexMeshDrawable.cpp
mosra: > But "Explicit uniform location" is not supported until 4.3.

Right, sorry. For some reason I thought it got in much earlier.

> It makes things complicated. But do you think it is necessary or it can reduce the overhead?

Uh oh, i meant something else -- the `uniformLocation()`, due to all its string operations, is the problem, not `setUniform()`. So call it once in the constructor, save the ID and then do just `setUniform(exposureUniform_, value)`. Check the performance plot in #151 -- just by going from `GenericShader` (which calls `uniformLocation()` every time) to magnum's `Shaders::Flat` (which caches the locations), I was able to shave some additional milliseconds.

Long time ago I tried to cache the uniform values as well, but it didn't prove any measurable performance improvement. And the maintenance overhead was absolutely not worth it. What could make sense is going with UBOs, but those are known to be even slower than classic `setUniform()` calls on Mac and some Intel GPUs, so ... ¯\\\_(ツ)_/¯

src/esp/assets/PTexMeshData.cpp
```diff
@@ -582,55 +660,77 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
     currentMesh->ibo.setData(submeshes_[iMesh].ibo,
```
bigbike: @mosra:
I am debugging the segfault.
After this "setData", do you think currentMesh->ibo.size() would equal to submeshes_[iMesh].ibo.size()?
Same for the vbo.

bigbike: I am asking this question is because I observed the size of currentMesh->ibo.size() was 4x of the submeshes_[iMesh].ibo.size(). 
I suspect currentMesh->ibo.size() returns the bytes of the entire array.
If this is the case, later "setCount(currentMesh->ibo.size())" would be wrong.

mosra: > After this "setData", do you think currentMesh->ibo.size() would equal to submeshes_[iMesh].ibo.size()?

The naming is making this harder than it should be, first I assumed the two `ibo`s are both `std::vector`s but they're actually not :) Calling [GL::Buffer::size()](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Buffer.html#a5e9a38c6e0d1e3e818d8648c0ac141b7) returns a number of bytes, since the buffer is a typeless piece of memory. And it's a query into GL, which is best avoided (moreover this function is not available at all in OpenGL ES or WebGL, but I'm not sure if that matters here) -- if you have the same information available CPU-side, it's recommended to prefer using that -- the `size()`, `data()`, `imageSize()`, ... functions on GL objects are meant mainly for debugging purposes.

Few lines below in the diff I see this -- could it be related?

```diff
-        .setCount(currentMesh->ibo.size() / 2)
+        .setCount(currentMesh->ibo.size())
```

bigbike: Yes, that is the problem. We actually need to pass the "number of primitives" to the setCount(), not the "number of bytes". That is why no matter if we divide it by 2, it will segfault anyway.

bigbike: Do you have any other example in your docs that shows how to correctly use ibo, vbo?
The "textured triangle" demo is a bit too simple in this case, in my opinion.

mosra: The Primitives example expands on that, and the GL::Mesh documentation is very extensive.

src/esp/gfx/Viewer.cpp
```diff
@@ -87,8 +90,17 @@ Viewer::Viewer(const Arguments& arguments)
   agentBodyNode_ = &rootNode_->createChild();
   cameraNode_ = &agentBodyNode_->createChild();
 
+  /*
   cameraNode_->translate({0.0f, cameraHeight, 0.0f});
   agentBodyNode_->translate({0.0f, 0.0f, 5.0f});
+  */
+
+  vec3f eye(0.0f, 0.0f, 1.0f);
```
bigbike: This is the debug code, and it is temporary. Will definitely remove them after it is finished.

src/esp/gfx/Viewer.cpp
```diff
@@ -49,6 +49,9 @@ Viewer::Viewer(const Arguments& arguments)
       .addBooleanOption("enable-physics")
       .addOption("physicsConfig", "./data/default.phys_scene_config.json")
       .setHelp("physicsConfig", "physics scene config file")
+//      .addOption("lookAt", "0.0f,1.5f,0.0f,1.5f,1.0f,0.0f,1.0f,0.0f")
```
bigbike: @mosra : I am planning to offer an option to let the user provide the the projection parameters of the camera (eye, target, up). 
Question: if the input is a string as it shows here (0.0f,1.5f,0.0f,1.5f,1.0f,0.0f,1.0f,0.0f), do we have convenient functions in Corrade to parse it?

bigbike:
Screenshot 2019-08-23 17 12 56




We are almost there.
The only problem I have is the artifact on the edges of the quads.

src/esp/gfx/Viewer.cpp
mosra: Yes -- parse it as a Matrix4, tho it has to be delimited by spaces and without the `f` -- so 

```cpp
.addOption("lookAt", "0.0 1.5 0.0 1.5 1.0 0.0 1.0 0.0")
```

then you can get it conveniently as. Note the storage is column-major so depending on what you expect the users to do you might want to `transpose()` it after reading.

```cpp
Matrix4 lookAt = args.value<Matrix4>("lookAt");
```

By the way, some options are with `-` (`"enable-physics"`), some are camelCase -- can we stick to one style for good UX? So `"look-at"` in this case.

src/esp/gfx/PTexMeshShader.cpp
```diff
@@ -67,13 +67,14 @@ PTexMeshShader::PTexMeshShader() {
...
-  tileSizeUniform_ = uniformLocation("tileSize");
-  widthInTilesUniform_ = uniformLocation("widthInTiles");
+  // std::map<std::string, int> uniformLocations;
+  uniformLocations_["MVP"] = uniformLocation("MVP");
+  uniformLocations_["exposure"] = uniformLocation("exposure");
+  uniformLocations_["gamma"] = uniformLocation("gamma");
+  uniformLocations_["saturation"] = uniformLocation("saturation");
+  uniformLocations_["clipPlane"] = uniformLocation("clipPlane");
+  uniformLocations_["tileSize"] = uniformLocation("tileSize");
+  uniformLocations_["widthInTiles"] = uniformLocation("widthInTiles");
```
mosra: Wait... really? The main point of caching uniform locations was to *reduce* the number of allocations and string operations. This is even worse than calling `uniformLocation()` in every setter. What was wrong with the previous approach?

mosra: No. With the integer, the sequence of operation is:

1. fetch the ID from memory
2. call `glUniform()` with that ID

With strings, however:

1. call `std::strlen()` on the C char array to get its length
2. allocate memory for a temporary `std::string` of given length
3. copy thta C char array to that allocated memory
4. do a lookup in `std::map` / `std::unordered_map` using the temporary `std::string`
    - either *O(log n)* `std::strcmp()` operations (and lots of memory jumps)
    - or a hash calculation and then at least one `std::strcmp()` operation (and lots of memory jumps)
5. fetch the ID from memory
6. call `glUniform()` with that ID
7. deallocate the temporary string used for key lookup

src/esp/assets/PTexMeshData.cpp
```diff
@@ -572,67 +760,120 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+    // equals to numBytes
+    Magnum::ImageView2D image(Magnum::PixelFormat::RGB16F, {dim, dim}, data);
+    const int mipLevelCount = 1;
     renderingBuffers_[iMesh]
-        ->tex.setWrapping(Magnum::GL::SamplerWrapping::ClampToEdge)
-        .setMagnificationFilter(Magnum::GL::SamplerFilter::Linear)
-        .setMinificationFilter(Magnum::GL::SamplerFilter::Linear)
-        // .setStorage(1, GL::TextureFormat::RGB8UI, image.size())
+        ->tex
+        //.setWrapping(Magnum::GL::SamplerWrapping::ClampToEdge)
```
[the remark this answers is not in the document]
mosra: All of them? :) The upper bound of memory usage of mip levels is 4/3 of the original texture size, so leaving some of the lower levels doesn't really save much. So, [as the docs day](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Texture.html#GL-Texture-usage), `Math::log2(image.size().min()) + 1`.

mosra:
> The biggest issue that blocks this PR is the "fish net", the artifacts on the edges (see the previous pictures).

If this is just on macOS, then I can only guess ... unless you'd buy me a Mac to test on :) I don't know the PTex algorithm by heart, so here's just a mixed bag of ideas:

- Where do the artifacts happen? Primitive edges? On places where the texture discontinuities are?
- Is it static or changing from frame to frame?
- Where does the white color come from? Is it the clear color shining through? Does changing the clear color affect it? Could it be due to NaNs happening in the shader? Is there any bounds test in the shader that could cause this?
- What are the texture filtering parameters? Wrapping? Can it happen that you get a static color when you sample outside of the `[0 ; 1]` range?
- Is blending enabled? Do the artifacts change if you enable/disable it?
- Do the artifacts change when you enable/disable multisampling?

src/esp/assets/PTexMeshData.cpp
```diff
@@ -572,67 +760,120 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+    saveAdjacency(adjFaceFilename, adjFaces);
+    LOG(INFO) << "Done: it is computed and saved to: " << adjFaceFilename;
+  } else {
+    LOG(INFO) << "Done: Loaded it from " << adjFaceFilename;
   }
 
+  // exit(-1);
+
+  int maxBufferTextureTexels = 0;
+  glGetIntegerv(GL_MAX_TEXTURE_BUFFER_SIZE, &maxBufferTextureTexels);
```
mosra: There's `GL::BufferTexture::maxSize()` btw. ;) Also if you search for `GL_MAX_TEXTURE_BUFFER_SIZE` in magnum docs, it'll lead you to that function.

mosra:
> I suspect the buffer textures somehow on mac do not work.

Interesting -- any chance you could build Magnum with tests enabled (`WITH_TESTS` and `WITH_GL_TESTS` CMake options) and then run `src/Magnum/GL/Test/BufferTextureGLTest` on Mac? If I remember correctly, there weren't any issues with those last time I checked, but I could misremember. Unfortunately Mac doesn't have a debug output, so checking for GL errors can't be done without printf-debugging-style checking for `GL::Renderer::error()` every few lines :(

Sometimes Mac doesn't like particular texture formats, can you try changing from `R32UI` to e.g. `RGBA32UI` or `RGBA16UI`?

mosra:
Yes, I know, you mentioned that above -- it means the buffer texture doesn't work at all there for some reason. Got any success running the tests or checking for GL errors?

src/esp/gfx/PTexMeshDrawable.cpp
```diff
@@ -18,18 +20,45 @@ PTexMeshDrawable::PTexMeshDrawable(
...
+      .setSaturation(saturation_)
+      .setAtlasTextureSize(tex_, tileSize_)
+      .bindAtlasTexture(tex_)
+      .bindAdjFacesBufferTexture(adjFaces_)
       .setMVPMatrix(camera.projectionMatrix() * transformationMatrix);
   mesh_.draw(ptexMeshShader);
+
+  //XXX
+  // ASSERT(Magnum::GL::Renderer::Error::NoError == Magnum::GL::Renderer::error());
+  switch (Magnum::GL::Renderer::error()) {
```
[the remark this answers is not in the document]
mosra: Does the InvalidOperation happen on Linux too? 

- If not, then that is a potential source of the problem :) 
- If yes, can you run on Linux with `--magnum-gpu-validation on` passed to the application? (Or setting the `MAGNUM_GPU_VALIDATION` environment variable to `ON`, in case you can't pass arguments.) That should tell you the origin of the warning in the console output. Ideally code that works shouldn't be emitting any errors.

mosra:
Ah, right, of course -- thanks a lot for the investigation, this made it clear -- because you're adding a freshly created buffer to the texture first and only then filling it with data, GL complains that the buffer hasn't been created yet (`glGenBuffers()` doesn't create it). The `GL::BufferTexture::setBuffer()` had an assumption that only already created buffers are passed to it. This didn't manifest on Linux because all other platforms support the `ARB_direct_state_access` extension that has `glCreateBuffers()` which doesn't suffer from this issue.

In mosra/magnum@7b43ab5fc24c345170b78a264931ecca612ebe8e I updated `setBuffer()` to account for this case -- and the test is also expanded. Could you pull latest `master` and run the `GLBufferTextureGLTest` again? I'm interested in the output now. Using latest `master` in Habitat should also make the buffer texture work now, without any additional patching.

What I'm still not sure about is the segfault you were getting -- in the test I'm calling `setData()` before `setBuffer()` and not getting any crash. So that has to be unrelated. Any chance you could get me a backtrace? Would be very interested in where it blows up.

Thanks again! :+1:

mosra:
> Can you make it fast-forward to 7b43ab5 as you mentioned?

But I need *you* to test it first :)

```sh
cd src/deps/magnum
git checkout master
```

mosra:
Oh, that's great it's reproducible with the test alone. Can you get a backtrace of that, please? If that's a macOS driver bug, I could try to work around it on the engine side. 

For the submodules, you can pass `--no-update-submodules` to `setup.py` (got added in https://github.com/facebookresearch/habitat-sim/pull/38) :)

bigbike:
The simulator was suffering the segfault as well with the latest magnum master.
It was also from the "setData". I tried to disable it, and simply used:
```
    glBindBuffer(GL_TEXTURE_BUFFER, currentMesh->adjFacesBuffer.id());
    glBufferData(GL_TEXTURE_BUFFER, sizeof(uint32_t) * adjFaces[iMesh].size(), (void*)adjFaces[iMesh].data(), GL_STATIC_DRAW);
```
It did not work either. 

It has nothing to do with the order sequence now. What was the change in the setBuffer? It may cause the failure here.

mosra:
Sorry for the late replies, having too much on my plate :/

> Something is wrong with the mosra/magnum@7b43ab5

I'm pretty confident it's a macOS driver bug, *not* a problem with magnum. Nevertheless, the engine *has to* deal with the problem and make it non-crashy, somehow. The `createIfNotAlready()` is simply calling `glBindBuffer()` in case the buffer isn't created yet. So equivalent to the crashy pure-GL code [you tried above](https://github.com/facebookresearch/habitat-sim/pull/132#issuecomment-525577400) -- again I don't think this is due to anything conflicting with magnum. Proof by statistics -- as far as I have checked, this order of operations isn't causing any crashes or GL errors anywhere else, only on macOS.

So we're looking at a driver bug. At this point I'm not sure about anything anymore, to be frank, and I guess I really need some access to a Mac to debug this. 

A totally random idea -- can you apply the following patch to the test and run it again? It's silly (and not required by the spec at all), but maybe that is the missing bit that causes the driver to blow up.

```patch
diff --git a/src/Magnum/GL/Test/BufferTextureGLTest.cpp b/src/Magnum/GL/Test/BufferTextureGLTest.cpp
index 29d95d90d..104953f4f 100644
--- a/src/Magnum/GL/Test/BufferTextureGLTest.cpp
+++ b/src/Magnum/GL/Test/BufferTextureGLTest.cpp
@@ -207,7 +207,7 @@ void BufferTextureGLTest::setBufferEmptyFirst() {
     #endif
 
     BufferTexture texture;
-    Buffer buffer;
+    Buffer buffer{Buffer::TargetHint::Texture};
     texture.setBuffer(BufferTextureFormat::RGBA8UI, buffer);
 
     MAGNUM_VERIFY_NO_GL_ERROR();

```

bigbike:
@mosra :

Tried. Segfault.
Do not worry. I will help you to find this bug and have a workaround.

If you are a contingent worker at FB, can you ask the HR or related staff to get a macbook pro? See if they can ship one to your country. I do not think it is a problem!

Sorry, I later have 1 hour meeting, will be back in the afternoon. Feel free to go to sleep:)

mosra:
Uh. 

Plan Z, then: can you try with plain textures, then? Similarly to what I did with the Primitive ID texture in the Primitive ID shader. I'm pretty sure this will fail on larger models and be slower than BufferTexture, but could be a workable short-term solution to get this running on Mac:

- `TextureFormat::R32UI`
- `texelFetch()` in the shader to avoid texture filtering and all that unnecessary stuff
- `Texture1D` instead of `Texture2D` might get you bigger max size limits

bigbike:
> Uh.
> 
> Plan Z, then: can you try with plain textures? Similarly to what I did with the Primitive ID texture in the Primitive ID shader. I'm pretty sure this will fail on larger models and be slower than BufferTexture, but could be a workable short-term solution to get this running on Mac:
> 
> * `TextureFormat::R32UI`
> * `texelFetch()` in the shader to avoid texture filtering and all that unnecessary stuff
> * `Texture1D` instead of `Texture2D` might get you bigger max size limits

On Mac, I disabled the buffer texture in the shader. And it gave fairly good results:



test rgba 00000

**The current status:**
It works 100% on Linux. It gives pretty good result on Mac with the workaround.

**On Mac:**
I believe somehow the buffer texture works (comparing it did not work at all), but I cannot confirm the value (adjFaces within) is correct on GPU when the shader uses it. (by the way, Do you have any ideas how to verify it?). Once I can verify the buffer texture values are correct, I can go further to debug the "fish net" artifacts. 

Current workaround: Disable the buffer texture. Just not use the adjFaces.

**Thinking**
adjFaces are ONLY used in texture filtering. The effects brought by it can only be visible when you zoom in quite a bit on the texture. I understand such seamless filtering is rather critical for industry level, film production (e.g., to avoid lighting artifacts when using displacement maps). But considering our project and application, I think the image quality is excellent. So at this moment, I think the practical solution is to fully enable all the functionalities on Linux, while on Mac we disable the buffer texture. (We will try to completely fix it on Mac, once we resolve the segfault in Magnum).  
Keep this in our mind that we only use Mac for development visualization the scene. We never train any models using the MacBook. We do the training in the linux boxes. So as long as the linux version is OK, we can go ahead like this.

About the segfault on Mac, and the magnum solution:
As a computer graphics guy for many years, I feel the pain from you. I would like to thoroughly understand what happened, and if there is a workaround. So I will keep debugging it with you. 

For the Texture1D idea, I think it is a good idea to try. I would like to see at least some types of "buffer" work, and can pass the adjFaces to the shader correctly on Mac. But like you said, it would fail when the number of texels was big, plus it was slower. So I may not adopt it as our solution here. I will still use the texture buffer in this case. How do you think?

mosra:
> Do you have any ideas how to verify it?

Magnum has [a framework for shader tests](https://doc.magnum.graphics/magnum/classMagnum_1_1DebugTools_1_1CompareImage.html) and it recently got improved even further for easier iteration. The most used builtin shaders are now all tested with it (see https://github.com/mosra/magnum/tree/master/src/Magnum/Shaders/Test). So if it would be feasible to craft small enough data (and expected output) to test on, then I think this would be a way to go. With the test scaffolding ready, the adjacency could be then (for example) verified by a specially crafted texture where the adjacent faces have a completely different color (and the test checks that the gradient in between is monotonic and doesn't include colors from wrong faces).

Yes, it's some nontrivial work upfront, but it would be useful far beyond this bug I think -- later refactorings will be much easier to do, not to mention opening possibilities for benchmark-guided optimizations and saving us a lot of time when we get to making things work on Vulkan.

> Simulator segfaults with the magnum master

This is unfortunate, but I still believe Magnum does the right thing (and so I don't think the change should be reverted) -- the test that triggers the crash on Mac caused GL_INVALID_OPERATION on all drivers before, now it works everywhere except Mac.

> On Mac, I disabled the buffer texture in the shader.

So it's basically not used at all now, right? Which means if you'd `#ifdef` out the crashy part on Mac, the rendering wouldn't change and you could keep using current Magnum master, right?

> I would like to thoroughly understand what happened, and if there is a workaround.

Unless I get access to a Mac, I'm afraid I can't debug this any further. I have some ideas, but with this way of "remote debugging" it would cost both you and me an extreme amount of time and I think that time could be spent more wisely elsewhere.

bigbike:
@mosra:
Good to see your suggestions. The debugTools seems very interesting. I am going to learn more from your online docs.
As suggested by the other team members in the engineering meeting, I am going to split this PR into a couple of small ones to make the code review easier for the others.

bigbike:
We divided it into a couple of PRs, which were committed in.
The system now can support PTex Mesh rendering.
Many thanks are given to all the reviewers. especially @mosra !
Close this PR.

PR #133, opened 2019-08-06
Webgl viewer

msbaines:
## Motivation and Context

WebGL build of habitat-sim. Currently only the viewer is supported.

## How Has This Been Tested

1. Download and install emscripten SDK from https://emscripten.org/
1. Make sure to source emscripten env variables before building (i.e. `source ~/code/emsdk/emsdk_env.sh` or similar)
1. Run `./build_js.sh`
1. Follow instructions at end of build to launch server and open viewer in browser tab

## Types of changes

## Checklist


src/CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ cmake_minimum_required(VERSION 3.10)
 
 project(esp)
 
+if(CMAKE_SYSTEM_NAME MATCHES "Emscripten")
+  set(EMSCRIPTEN 1)
+endif()
```
mosra: This isn't needed anymore, was fixed with https://github.com/mosra/toolchains/commit/e33a22f2400b29b8d9492f8686fc00ed2c309a1a and that's pulled in with latest Corrade I think.

OTOH I would strongly advise to use `CORRADE_TARGET_EMSCRIPTEN` or `MAGNUM_TARGET_WEBGL` instead of `if(EMSCRIPTEN)` in the CMakeLists, because those have a clearly defined semantics and are consistent with other platform defines.

build_js.sh
```diff
@@ -0,0 +1,29 @@
...
+cd build_js
+
+cmake ../src \
+    -DBUILD_GUI_VIEWERS=ON \
+    -DBUILD_PYTHON_BINDINGS=OFF \
+    -DBUILD_DATATOOL=OFF \
+    -DBUILD_PTEX_SUPPORT=OFF \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_PREFIX_PATH="$EMSCRIPTEN" \
+    -DTARGET_GLES2=OFF \
```
mosra: I think `TARGET_GLES2=OFF` would be better set inside the `if(NOT USE_SYSTEM_MAGNUM)` of `dependencies.cmake`.

build_js.sh
```diff
@@ -0,0 +1,29 @@
...
+    -DCMAKE_TOOLCHAIN_FILE="../src/deps/corrade/toolchains/generic/Emscripten-wasm.cmake" \
+    -DCMAKE_INSTALL_PREFIX="." \
+    -DCMAKE_CXX_FLAGS="-s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1 --preload-file $DATA_DIR/scene_datasets/habitat-test-scenes@/" \
+
+cmake --build . -- -j 4
+cmake --build . --target install -- -j 4
+
+echo "Done building."
+echo "Run:"
+echo "python -m http.server"
```
mosra: `python3 -m http.server`? I got a lot of complaints about this from people who still suffer from Python 2 being the default on their system :)

src/esp/gfx/Viewer.cpp
```diff
@@ -34,8 +34,8 @@ Viewer::Viewer(const Arguments& arguments)
       controls_(),
       previousPosition_() {
   Utility::Arguments args;
-  args.addArgument("file")
-      .setHelp("file", "file to load")
+  args.addNamedArgument("scene")
+      .setHelp("scene", "scene file to load")
```
mosra: I would instead do something like

```cpp
#ifndef CORRADE_TARGET_EMSCRIPTEN
args.addArgument("scene");
#else
args.addNamedArgument("scene");
#endif
```

so people on desktop don't have to say `./viewer --scene bla.gltf` but just `./viewer bla.gltf`, like it was until now :)

src/cmake/dependencies.cmake
```diff
@@ -112,23 +117,28 @@ if(NOT USE_SYSTEM_MAGNUM)
   set(WITH_STBIMAGEIMPORTER ON CACHE BOOL "WITH_STBIMAGEIMPORTER" FORCE)
   set(WITH_STBIMAGECONVERTER ON CACHE BOOL "WITH_STBIMAGECONVERTER" FORCE)
   set(WITH_SDL2APPLICATION OFF CACHE BOOL "WITH_SDL2APPLICATION" FORCE)
+  set(WITH_GLFWAPPLICATION OFF CACHE BOOL "WITH_GLFWAPPLICATION" FORCE)
   set(WITH_EIGEN ON CACHE BOOL "WITH_EIGEN" FORCE) # Eigen integration
-  set(WITH_PYTHON ON CACHE BOOL "" FORCE) # Python bindings
+  if(BUILD_PYTHON_BINDINGS)
+    set(WITH_PYTHON ON CACHE BOOL "" FORCE) # Python bindings
```
mosra: Hmm maybe also doing 

```cmake
if(BUILD_PYTHON_BINDINGS)
  add_subdirectory("${DEPS_DIR}/magnum-bindings")
endif()
```

down below, instead of adding it unconditionally?

src/esp/gfx/Viewer.cpp
msavva: Agree, that's a good suggestion -- thanks @mosra

mosra:
One thing I forgot -- this happened recently: https://blog.magnum.graphics/announcements/new-emscripten-application-implementation/

I'm not saying this PR should switch to it ("don't fix what's not broken") but I'd suggest keeping this possibility in mind and switching to it later, as the new implementation is superior to `Sdl2Application` in every way.

src/CMakeLists.txt
msbaines: The one issue I ran into using CORRADE_TARGET_EMSCRIPTEN is that it is not available until the add_subdirectory of corrade which does not occur until near the end of src/cmake/dependencies.cmake.

The habitat-sim build seems to be primarily using UNIX, APPLE, and WIN32 so I was also trying to be consistent with that.

I will move sourcing of corrade cmake to earlier in the build. Maybe into the top-level CMakeLists.txt like in magnum CMakeLists.txt does:

# Find Corrade first so we can check on the target
find_package(Corrade REQUIRED Utility)

That way we can follow a style similar to magnum and use CORRADE_TARGET_* consistently and in place of UNIX, APPLE, etc.

mosra: Right, sorry, forgot about this "chicken and egg" problem :) I'm not sure if moving `add_subdirectory(corrade)` earlier is feasible, it also needs to keep working with `USE_SYSTEM_MAGNUM`.

The only place where `CORRADE_TARGET_EMSCRIPTEN` is not defined yet is until `add_subdirectory(corrade)` is called in `dependencies.cmake`, right? Then I think using `EMSCRIPTEN` in that one file until the `add_subdirectory()` could be okay. But feel free to do whatever seems clearer to you.

In any case, the above piece of code is not needed anymore as it's set by the toolchain now.

hzyjerry:
It seems that CI complains that tinyobj is not cleanly removed and building with `setup.py` is failing. I was able to build after running `git rm --cached src/deps/tinyobjloader/`

msbaines:
> It seems that CI complains that tinyobj is not cleanly removed and building with `setup.py` is failing (strangely: https://circleci.com/gh/facebookresearch/habitat-sim/988?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).
> 
> I was able to build after running `git rm --cached src/deps/tinyobjloader/`

Your change seems to be adding the tinyobjloader submodule back:

```
diff --git a/src/deps/tinyobjloader b/src/deps/tinyobjloader
new file mode 160000
index 0000000..fd06fa4
--- /dev/null
+++ b/src/deps/tinyobjloader
@@ -0,0 +1 @@
+Subproject commit fd06fa49e4e3f4811df4ee566a0ea41832eb0cca
```

Probably do some git-submodule weirdness.

You should be able to fix by removing the submodule like you did (via git-rm) and then committing that change.

PR #141, opened 2019-08-08
Add SimulatorWithAgents

msbaines:
## Motivation and Context

This change creates a new Simulator subclass, SimulatorWithAgents, which functions similar to the Simulator python class. We plan to use this class for the js port initially and then migrate python over to it eventually.

## How Has This Been Tested

Build.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/Buffer.h
+++ b/src/esp/core/Buffer.h
@@ -1,0 +1,53 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include "esp/core/esp.h"
+
+namespace esp {
+namespace core {
+
+// Enumeration of data
+enum class DataType {
+  DT_NONE = 0,
+  DT_INT8 = 1,
+  DT_UINT8 = 2,
+  DT_INT16 = 3,
+  DT_UINT16 = 4,
+  DT_INT32 = 5,
+  DT_UINT32 = 6,
+  DT_INT64 = 7,
+  DT_UINT64 = 8,
+  DT_FLOAT = 9,
+  DT_DOUBLE = 10,
+};
+
+class Buffer {
+ public:
+  explicit Buffer(){};
+  explicit Buffer(const std::vector<size_t> shape, const DataType dataType) {
+    this->shape = shape;
+    this->dataType = dataType;
+    alloc();
+  };
+  void clear();
+  virtual ~Buffer() { dealloc(); }
+
+ protected:
+  void alloc();
+  void dealloc();
+
+ public:
+  void* data = nullptr;
+  size_t totalBytes = 0;
+  size_t totalSize = 0;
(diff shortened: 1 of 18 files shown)


src/esp/sensor/Sensor.cpp
```diff
@@ -20,10 +20,14 @@ Sensor::Sensor(scene::SceneNode& node, SensorSpec::ptr spec)
...
-Observation Sensor::getObservation() {
+bool Sensor::getObservation(gfx::Simulator& sim, Observation& obs) {
   // TODO fill out observation
-  Observation obs{};
-  return obs;
+  return false;
+}
+
+bool Sensor::getObservationSpace(ObservationSpace& space) {
+  // TODO fill out observation spacek
```
msavva: typo: `spacek` -> `space`

erikwijmans:
Maybe it makes sense to take the current thing we call the simulator and rename it `HabitatSimBackend`?  Then we can have this the `c++` front-end and the existing thing be the `python` front end?

src/esp/core/Buffer.h
```diff
@@ -0,0 +1,53 @@
...
+  DT_UINT16 = 4,
+  DT_INT32 = 5,
+  DT_UINT32 = 6,
+  DT_INT64 = 7,
+  DT_UINT64 = 8,
+  DT_FLOAT = 9,
+  DT_DOUBLE = 10,
+};
+
+class Buffer {
```
erikwijmans: Any reason why this takes a datatype enum instead of being templated  i.e if you want a double buffer, just get a `Buffer<double>`?

erikwijmans: Never mind, this makes sense with the way it's being used.

msbaines: > Maybe it makes sense to take the current thing we call the simulator and rename it `HabitatSimBackend`? Then we can have this the `c++` front-end and the existing thing be the `python` front end?

Since the js work is new and exploratory, the approach we are taking with the js port  is to extend the c++ port first and then rationalize the python port once we have a good understand of the architecture. That way we avoid churn and destabilization of the python port during the exploratory phase. The downside is that you don't get the benefits of an integrated code base (testing, etc). Does that sound reasonable?

msbaines: > Any reason why this takes a datatype enum instead of being templated i.e if you want a double buffer, just get a `Buffer<double>`?

We don't know the type at compile time.

erikwijmans:
> Since the js work is new and exploratory

Does it make sense to have a `js` branch that the exploration is done within then?  There can still be PRs and such, but just into that branch until things settle down.

msavva:
> > Since the js work is new and exploratory
> 
> Does it make sense to have a `js` branch that the exploration is done within then? There can still be PRs and such, but just into that branch until things settle down.

I don't think that's a good idea, right now at least.  We are making some fundamental changes to the C++ api which if we bifurcate on right now will lead to massive amounts of pain when trying to rebase and bring things back into alignment down the line.

PR #142, opened 2019-08-09
Use EmscriptenApplication instead of SDL2Application for WebGL.

msbaines:
## Motivation and Context

Emscripten is preferred over SDL2.

## How Has This Been Tested

Verified the WebGL Viewer still works.

## Types of changes

## Checklist


mosra:
Yay, you were fast :)

I'm not able to try this out right now (in the middle of a shader refactor), but there's one important difference when porting from Sdl2Application -- framebuffer size is no longer the same as window size on HiDPI screens (macbook's internal display, for example), which means the mouse events here will behave differently ("slower"): https://github.com/facebookresearch/habitat-sim/blob/8ddc21f3bcb5a89343595d94c58e721bf3f72b1e/src/esp/gfx/Viewer.cpp#L140-L185

In particular:

- `event.position()` is in window coordinates, but the `positionOnSphere()` operates relative to framebuffer size, so to get framebuffer-relative coordinates you need to do `Vector2{event.position()*framebufferSize()}/Vector2{windowSize()}` first
- (not directly related to HiDPI but) the scroll event reports a floating-point offset instead of 0 / -1 / 1, so incorporating it into the equation directly (instead of the `?:`) would make it behave more natural when the mouse or touchpad has a wheel acceleration (most of them have nowadays)

Of course try it out to check if the new behavior feels consistent with the old one.

If you run the app with `?magnum-log=verbose` appended to the URL, it'll print DPI scaling to browser's console. On the Mac you should see a value of 2 I think, indicating a HiDPI screen. [Obligatory docs here](https://doc.magnum.graphics/magnum/classMagnum_1_1Platform_1_1EmscriptenApplication.html#Platform-EmscriptenApplication-dpi).

I'm looking at Magnum's Viewer example (from which this viewer was adapted) and there it's unfortunately not adapted yet so you can't look there for an inspiration -- on my TODO list, sorry.

src/esp/gfx/Viewer.h
```diff
@@ -6,7 +6,9 @@
 
 #include <Magnum/configure.h>
 #ifdef MAGNUM_TARGET_WEBGL
-#include <Magnum/Platform/Sdl2Application.h>
+// Undefine when using SDL2 Emscripten.
+#define EMSCRIPTEN_INPUT
```
mosra: This is not used anywhere -- probably a leftover from a debugging session? :)

PR #143, opened 2019-08-09
Use configure.h

erikwijmans:
## Motivation and Context

We are starting to have a number of `#ifdef <blah>` for conditional builds.  @mosra brought up that having a `configure.h` file which is conditionally generated by cmake makes this much cleaner.  This PR swaps our `target_compile_definitions` to this.  I also renamed things to `ESP_WITH_*` so that we have a consistent way of specifying if we are building things or not.

## How Has This Been Tested

It builds!


src/esp/core/CMakeLists.txt
```diff
@@ -1,5 +1,21 @@
 file(GLOB_RECURSE core_SOURCES "*.cpp")
 
+if(NOT BUILD_GUI_VIEWERS AND UNIX AND NOT APPLE)
```
mosra:
```suggestion
if(NOT BUILD_GUI_VIEWERS AND CORRADE_TARGET_UNIX AND NOT CORRADE_TARGET_APPLE)
```

... because the builtin `UNIX` is unfortunately defined for Emscripten as well, which is probably not what you want. The `CORRADE_TARGET_UNIX` explicitly isn't.

(Sorry for pushing for these all the time :sweat_smile:)

src/esp/assets/ResourceManager.cpp
```diff
@@ -149,8 +149,8 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
 
   return true;
 #else
-  LOG(ERROR)
-      << "PTex support not enabled. Define BUILD_PTEX_SUPPORT when building.";
+  LOG(ERROR) << "PTex support not enabled. Define ESP_WITH_PTEX_SUPPORT when "
```
mosra: This naming inconsistency between `BUILD_*` and `WITH_*` might be a problem for users -- could the CMake `option()` be renamed to `WITH_*` as well? Then the error message would be clearer as

```suggestion
  LOG(ERROR) << "PTex support not enabled. Enable the WITH_PTEX_SUPPORT CMake option when "
```

erikwijmans: I can change things to `BUILD_*` to be consistent with the current cmake options

mosra: Or that, I don't have a preference. In Magnum the difference between `WITH` and `BUILD` is kinda blurry also, so ¯\\\_(ツ)_/¯

src/esp/core/configure.h.cmake
```diff
@@ -0,0 +1,11 @@
...
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#cmakedefine BUILD_EGL_SUPPORT
+
+#cmakedefine BUILD_PTEX_SUPPORT
+
+#cmakedefine BUILD_GLOG_SHIM
```
mosra: Wait but now it all lost the `ESP_` prefix, especially the `BUILD_EGL_SUPPORT` is quite likely to conflict with something else ... and there's `ESP_USE_GLOG_SHIM` but this is `BUILD_GLOG_SHIM`.

Can it be `BUILD_*` on CMake side and `ESP_BUILD_*` on C++ side? :)

erikwijmans: Yes :)

erikwijmans:
Yeah, I need to do a better job reviewing my commits before pushing them as I currently just use the PR to do that as I do like git's interface to do that :)

PR #147, opened 2019-08-13
Fix WebGL glClear issue.

msbaines:
Fixes the following GL_ERROR on WebGL:

glClear: can't be called on integer buffers

The problem is that glClear will clear all glDrawBuffers.

## Motivation and Context

Before this change, the main color_attachement was not clearing.

## How Has This Been Tested

Tested with work-in-progress embind patch.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/Renderer.cpp
+++ b/src/esp/gfx/Renderer.cpp
@@ -62,7 +62,7 @@ struct Renderer::Impl {
   }
 
   inline void renderEnter() {
-    framebuffer_.clear(GL::FramebufferClear::Depth);
+    framebuffer_.clearDepth(1.0);
     framebuffer_.clearColor(0, Color4{});
     framebuffer_.clearColor(1, Color4{});
     framebuffer_.clearColor(2, Vector4ui{});
--- a/src/esp/gfx/Renderer.cpp
+++ b/src/esp/gfx/Renderer.cpp
@@ -62,9 +62,9 @@ struct Renderer::Impl {
   }
 
   inline void renderEnter() {
-    framebuffer_.clear(GL::FramebufferClear::Color |
-                       GL::FramebufferClear::Depth);
-    framebuffer_.clearColor(1, Vector4{});
+    framebuffer_.clear(GL::FramebufferClear::Depth);
+    framebuffer_.clearColor(0, Color4{});
+    framebuffer_.clearColor(1, Color4{});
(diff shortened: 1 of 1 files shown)


src/esp/gfx/Renderer.cpp
```diff
@@ -62,9 +62,9 @@ struct Renderer::Impl {
   }
 
   inline void renderEnter() {
-    framebuffer_.clear(GL::FramebufferClear::Color |
-                       GL::FramebufferClear::Depth);
-    framebuffer_.clearColor(1, Vector4{});
+    framebuffer_.clear(GL::FramebufferClear::Depth);
+    framebuffer_.clearColor(0, Color4{});
+    framebuffer_.clearColor(1, Color4{});
     framebuffer_.clearColor(2, Vector4ui{});
```
mosra: To use just a single API it could be `clearDepth()` as well, but I don't have a preference.

erikwijmans:
In #114 I’m switching to just a clear call as multiple clear calls has a noticeable impact on performance. So I guess we’ll need to ifdef that method?

erikwijmans:
In #114 I’m switching to just a clear call as multiple clear calls has a noticeable impact on performance. So I guess we’ll need to ifdef that method?

msbaines:
> In #114 I’m switching to just a clear call as multiple clear calls has a noticeable impact on performance. So I guess we’ll need to ifdef that method?

Using glClear to clear an integer buffer may be undefined behavior.

The WebGL implementation in the browser will police input for undefined behavior before calling into the driver. I suspect that is what is causing this error to only happen on WebGL and not when running natively.

The spec doesn't seem to say anything about integer buffers and glClear but it does say the following about glClearBuffer* and integer buffers:

> For all forms of ClearBuffer* and ClearNamedFramebuffer*, the result of
> these commands is undefined if no conversion between the type of the specified
> value and the type of the buffer being cleared is defined (for example, if ClearBufferiv is called for a fixed- or floating-point buffer, or if ClearBufferfv is called
> for a signed or unsigned integer buffer). This is not an error

OpenGL 4.6 (Core Profile) - 17.4.3.1 Clearing Individual Buffers https://www.khronos.org/registry/OpenGL/specs/gl/glspec46.core.pdf

So its not an error but it is undefined. If the performance benefit of a single clear is compelling you could create a driver/hardware whitelist to enable the functionality on driver/hardware combinations that have been verified.

PR #152, opened 2019-08-14
Use Phong shader for rendering scenes

mosra:
## Motivation and Context

before:

chair3-before

after:

chair3-after

This PR depends on #151 (which depends on #150), and only changes use of Flat to Phong, together with adding a fine-tuned light setup. What you *don't* see here is a completely fresh set of tests for all shader combinations in Magnum itself, which makes adding new features (such as vertex color, object ID output, normal mapping, alpha masking, ...) much easier and less error prone. This will also enables many new features in the near future, such as instanced rendering for physics objects.

## How Has This Been Tested

Viewer looks prettier than before, but I think some ground truth data need to be updated for this to pass tests. @erikwijmans could you take care of that? :pray: 

Perf measurements -- there's *expected* slowdown compared to #150 and #151, as the shader is getting much more data and doing much more work now. However it's not *too* bad I would say -- it will get much worse with PBR: 

image

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -14,7 +14,7 @@
 #include <Magnum/Math/Range.h>
 #include <Magnum/Math/Tags.h>
 #include <Magnum/PixelFormat.h>
-#include <Magnum/Shaders/Flat.h>
+#include <Magnum/Shaders/Phong.h>
 #include <Magnum/Trade/AbstractImporter.h>
 #include <Magnum/Trade/ImageData.h>
 #include <Magnum/Trade/MeshObjectData3D.h>
@@ -657,27 +657,44 @@ Magnum::GL::AbstractShaderProgram* ResourceManager::getShaderProgram(
 
       case COLORED_SHADER: {
         shaderPrograms_[COLORED_SHADER] =
-            std::make_shared<Magnum::Shaders::Flat3D>(
-                Magnum::Shaders::Flat3D::Flag::ObjectId);
+            std::make_shared<Magnum::Shaders::Phong>(
+                Magnum::Shaders::Phong::Flag::ObjectId, 3 /*lights*/);
       } break;
 
       case VERTEX_COLORED_SHADER: {
         shaderPrograms_[VERTEX_COLORED_SHADER] =
-            std::make_shared<Magnum::Shaders::Flat3D>(
-                Magnum::Shaders::Flat3D::Flag::ObjectId |
-                Magnum::Shaders::Flat3D::Flag::VertexColor);
+            std::make_shared<Magnum::Shaders::Phong>(
+                Magnum::Shaders::Phong::Flag::ObjectId |
+                    Magnum::Shaders::Phong::Flag::VertexColor,
+                3 /*lights*/);
       } break;
 
       case TEXTURED_SHADER: {
         shaderPrograms_[TEXTURED_SHADER] =
-            std::make_shared<Magnum::Shaders::Flat3D>(
-                Magnum::Shaders::Flat3D::Flag::ObjectId |
-                Magnum::Shaders::Flat3D::Flag::Textured);
+            std::make_shared<Magnum::Shaders::Phong>(
+                Magnum::Shaders::Phong::Flag::ObjectId |
+                    Magnum::Shaders::Phong::Flag::DiffuseTexture,
+                3 /*lights*/);
       } break;
 
       default:
         return nullptr;
         break;
+    }
+
+    /* Default setup for Phong, shared by all models */
+    if (type == COLORED_SHADER || type == VERTEX_COLORED_SHADER ||
+        type == TEXTURED_SHADER) {
+      using namespace Magnum::Math::Literals;
+
+      static_cast<Magnum::Shaders::Phong&>(*shaderPrograms_[TEXTURED_SHADER])
+          .setLightPositions({Magnum::Vector3{10.0f, 10.0f, 10.0f} * 100.0f,
+                              Magnum::Vector3{-5.0f, -5.0f, 10.0f} * 100.0f,
+                              Magnum::Vector3{0.0f, 10.0f, -10.0f} * 100.0f})
+          .setLightColors({0xffffff_rgbf * 0.8f, 0xffcccc_rgbf * 0.8f,
+                           0xccccff_rgbf * 0.8f})
+          .setSpecularColor(0x11111100_rgbaf)
+          .setShininess(80.0f);
     }
   }
   return shaderPrograms_[type].get();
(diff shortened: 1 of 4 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -83,28 +83,45 @@ Magnum::GL::AbstractShaderProgram* ResourceManager::getShaderProgram(
...
         break;
     }
+
+    /* Default setup for Phong, shared by all models */
+    if (type == COLORED_SHADER || type == VERTEX_COLORED_SHADER ||
+        type == TEXTURED_SHADER) {
+      using namespace Magnum::Math::Literals;
+
+      static_cast<Magnum::Shaders::Phong&>(*shaderPrograms_[TEXTURED_SHADER])
+          .setLightPositions({Magnum::Vector3{10.0f, 10.0f, 10.0f} * 100.0f,
```
bigbike: Are these light positions randomly picked?
My idea is to put the lights on the corners of the bounding box of the scene, which can be adaptive.

mosra: Yes, those are just a verbatim copy from [magnum-player](https://magnum.graphics/showcase/player). There it was hand-tuned to give a nicely balanced lighting (and I think it's not bad here either), but something else could fit here even better.

The lights are currently relative to the camera so you always get the scene uniformly lit -- not sure if making them fully static would achieve the same, maybe balance them out? Definitely might need some fine-tuning :)

bigbike: > The lights are currently relative to the camera so you always get the scene uniformly lit -- not sure if making them fully static would achieve the same, maybe balance them out? Definitely might need some fine-tuning :)

I may misunderstand you:
When the camera moves, the lights follow?

I think it might be more realistic if the lights are fixed (though the scene looks ugly from some angles), especially if we would like to add shadows to the objects in the scene (we talked about it recently).

mosra: > When the camera moves, the lights follow?

Yes, because that looked like the best tradeoff for me -- as there are no shadows, having a room affected by a light coming through walls from some other room will not look realistic *at all*. 

IIRC back in April we talked about some models having designated light positions that could be used for giving the lights static positions? That could make sense to do together with shadows.

This PR is quite experimental, adding a completely new feature where it's unclear how it should be used. So please clone it, play with it, change light positions / colors / ... to see what fits best. I won't like to merge it without having a consensus on how the visuals should look like :)

msbaines:
I was looking at Phong.frag and it seem like its using the Phong shading model instead of the half-angle Blinn-Phong shading model. Blinn-Phong is more efficient and what is used by threejs and what was was specified in the original OpenGL fixed function pipeline. I think its also more physically accurate but I have no idea if that's true. shininess values need to be scaled though to get the same look. You may get a performance increase with Blinn-Phong.

```
        /* Add specular color, if needed */
        if(intensity > 0.001) {
            highp vec3 reflection = reflect(-normalizedLightDirection, normalize\
dTransformedNormal);
            mediump float specularity = clamp(pow(max(0.0, dot(normalize(cameraD\
irection), reflection)), shininess), 0.0, 1.0);
            fragmentColor += vec4(finalSpecularColor.rgb*specularity, finalSpecu\
larColor.a);
        }

```
https://threejs.org/docs/#api/en/materials/MeshPhongMaterial
https://www.khronos.org/registry/OpenGL/specs/gl/glspec21.pdf

mosra:
PR description updated with something actually presentable.

One question remains for this to be actually usable -- how to distinguish what is a mesh that already has shading baked in and what is not ... in the case above I'm branching on whether the mesh has normals (because the room doesn't have but the chairs do), but that of course doesn't work in general (PLYs are usually normal-less, some scans can have normals as well...). Other ideas?

cc: @erikwijmans @msavva

erikwijmans:
We have in the past talked about adding a `habitat_config.json` to the root directory for various datasets to specify things like the gravity direction.  We could revive that idea to specify whether or not to use phong?

mosra:
Just realized -- for glTF that info could be already encoded in the file, since there are extensions like [KHR_materials_unlit](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_unlit/README.md), but that's just glTF. (And then other stuff like various alternative PBR material specifications etc.) Scans loaded from OBJ / PLY can't really have that kind of info, tho.

aclegg3:
Could we include this functionality as optional and choose a default for now? Maybe we add an option to the SceneConfiguration and then work on another PR with logic to intelligently set that option for our datasets? We could set default to flat shading without breaking any existing expectations and then set phong in places where we have some way to confirm it.

mosra:
SceneConfiguration ... would mean editing a JSON file? Not a fan of config files :)

The main logic that seems to branch over various dataset types is in [esp/assets/Asset.cpp](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/assets/Asset.cpp), and the `AssetType` enum in particular, right? So there it could be extended to detect whether it needs to render flat or shaded, and pick a shader type based on that.

Looking there, it seems to pick *any* GLB file as a MP3D dataset (i.e., unshaded), so that check would need to be extended to verify the GLB file is really from MP3D and not somethine

aclegg3:
> SceneConfiguration ... would mean editing a JSON file? Not a fan of config files :)

Actually I was suggesting adding an optional flag or list of flags to the [SceneConfiguration](https://github.com/facebookresearch/habitat-sim/blob/781910bd06f9ce9621fedbab2ae7c46cd02df595/src/esp/scene/SceneConfiguration.h#L14) struct with the default being FLAT. We could pass that through at scene loading time giving the option to setup a phong shaded scene programmatically. This struct is also in the python bindings and [used by Simulator class to determine how to load the scene](https://github.com/facebookresearch/habitat-sim/blob/781910bd06f9ce9621fedbab2ae7c46cd02df595/src/esp/gfx/Simulator.cpp#L55). Shader could be passed to loadScene at that point. Giving the user a way to manually override our automated shader type might be a nice feature anyway for odd use cases we don't test.

> The main logic that seems to branch over various dataset types is in [esp/assets/Asset.cpp](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/assets/Asset.cpp), and the `AssetType` enum in particular, right? So there it could be extended to detect whether it needs to render flat or shaded, and pick a shader type based on that.
> 
> Looking there, it seems to pick _any_ GLB file as a MP3D dataset (i.e., unshaded), so that check would need to be extended to verify the GLB file is really from MP3D and not something else.

Agreed, but this will likely be an ongoing process as we dig into what differentiates various datasets and could be another PR. Similarly, the AssetInfo is loaded in Simulator and if no override flag is present could be used to fork on shader type as we build more aware default behavior.

dhuber23b:
Is there any update on this PR?  It has been a couple of months since any comments have been made.  If it is not ready, I will look into baking in textures on my models, since they look flat when inserted into the scene.

dhuber23b:
Here's an example.  Notice that the chair has no shading.

mosra:
Last we discussed, @aclegg3 promised to take over. The remaining blocker is deciding when to use flat shading (for photogrammetry scenes with lighting baked in) and when phong (artificial objects, House3D etc.). I don't have enough knowledge about the various datasets Habitat is using to implement this detection & decision making.

aclegg3:
Closing this based on other recent progress. (#500, #489)

PR #153, opened 2019-08-15
Initial javascript bindings.

msbaines:
## Motivation and Context

This change implements the initial javascript bindings.

There is also an html application that is similar to viewer except we are driving a real habitat simulator instance and presented the actual frames seen from the rgba_camera sensor.

## How Has This Been Tested

Tested binding.html in chrome browser.

## Types of changes

## Checklist


src/esp/bindings_js/bindings.html
```diff
@@ -0,0 +1,73 @@
...
+</head>
+
+<body>
+  <h1>Habitat</h1>
+  <div id="container">
+    <div id="sizer">
+      <div id="expander">
+        <div id="listener">
+          <canvas class="hidden" id="canvas"></canvas>
+          <canvas id="canvas2d" width="640" height="480" style="transform: scaleY(-1)"></canvas>
```
mosra: I was not involved in anything related to this at all so I have no idea about the decisions made ... but why does this need two canvases?

msbaines: The first canvas is for WindowlessEmscriptenApplication.js. It requires a canvas but I'm not sure how it is used. The second canvas is for presenting frames from the rgba_sensor.

mosra: AFAIK a WebGL context can't be created without a canvas, so this is a hidden canvas just for that. There's a pretty new experimental API for "headless contexts" but last time I checked the browser support was poor.

If the second canvas is used for presenting frames, why not use a non-windowless application here instead and present those through WebGL, like the viewer does?

msbaines: That's a good Idea. I will try to do that in a follow-up change. Currently, I have to color-space convert from linearSRGB to SRGB (i.e. apply gamma of 2.2) because canvas assumes SRGB for ImageData but the sensor data is linear. I wouldn't need to color convert if I used WebGL since texture are linear unless you specify an srgb format.

mosra: Oh, interesting. I have sRGB conversion among my TODOs for WebGL, because unlike all other GL flavors it doesn't provide any way to enable sRGB conversion on the default framebuffer. So I would need to do that manually in a shader.

PR #154, opened 2019-08-15
Height and width were swapped in PinholeCamera.

msbaines:
## Motivation and Context

This was causing frames from rgba_sensor to be stretched.

The bug was recently introduced by me in PR#141. Since know one yet uses this API, no one was affected by this bug.

Before:

before

After:

after

## How Has This Been Tested

With this fix, frames pulled from the rgba_sensor are no longer stretched when specifying a width of 640 and height of 480. 

## Types of changes

## Checklist

### DIFF
--- a/src/esp/sensor/PinholeCamera.cpp
+++ b/src/esp/sensor/PinholeCamera.cpp
@@ -58,9 +58,8 @@ bool PinholeCamera::getObservation(gfx::Simulator& sim, Observation& obs) {
   // TODO: Get appropriate render with correct resolution
   std::shared_ptr<gfx::Renderer> renderer = sim.getRenderer();
   vec3i resolution = renderer->getSize();
-  if (resolution[0] != spec_->resolution[0] ||
-      resolution[1] != spec_->resolution[1]) {
-    renderer->setSize(spec_->resolution[0], spec_->resolution[1]);
+  if (resolution[0] != width_ || resolution[1] != height_) {
+    renderer->setSize(width_, height_);
   }
   if (spec_->sensorType == SensorType::SEMANTIC) {
     // TODO: check sim has semantic scene graph


msbaines:
The test scenes in the repository look OK. This will require more investigation.

mosra:
I was running into this pretty often as well. Would it help if `spec_->resolution` was a struct with *named* `width` and `height` attributes instead of a two-component array of values without any defined order? Not sure how that would fit into the Python side, tho.

msavva:
> I was running into this pretty often as well. Would it help if `spec_->resolution` was a struct with _named_ `width` and `height` attributes instead of a two-component array of values without any defined order? Not sure how that would fit into the Python side, tho.

That is a good idea, and would help us prevent future confusion.  Having a simple struct and exposing that through pybind would likely help us better handle the change of convention between the gfx APIs and the tensor dimensions on the python side.

msbaines:
Agree. We need to do something about using vector for dimensions where height and width are different order. I will work on a proposal for that. I don't want to fix that in this change because I think it will require some discussion. The order used in resolution is also part of the public python API so cannot be changed so easily:

src/esp/bindings/bindings.cpp:      .def_readwrite("resolution", &SensorSpec::resolution)
examples/stereo_agent.py:    left_rgb_sensor.resolution = [512, 512]

So cool if I just land this fix for now and in parallel working on a long term solution.

msavva:
> Agree. We need to do something about using vector for dimensions where height and width are different order. I will work on a proposal for that. I don't want to fix that in this change because I think it will require some discussion. The order used in resolution is also part of the public python API so cannot be changed so easily:
> 
> src/esp/bindings/bindings.cpp: .def_readwrite("resolution", &SensorSpec::resolution)
> examples/stereo_agent.py: left_rgb_sensor.resolution = [512, 512]
> 
> So cool if I just land this fix for now and in parallel working on a long term solution?

Sounds good to me.  I believe this bug fix shouldn't break any python-side logic but I'm not 100% certain.

msbaines:
> Sounds good to me. I believe this bug fix shouldn't break any python-side logic but I'm not 100% certain.

This bug fix only affects a function that was added by me in SimulatorWithAgents PR #141. So currently the only users of this function are the js bindings code.

PR #155, opened 2019-08-16
CI: adapt to be consistent with habitat-api

mosra:
## Motivation and Context

Doing the same as here: https://github.com/facebookresearch/habitat-api/commit/dd75f18fb81e00810126a46c23c812e75230e86f#diff-1d37e48f9ceff6d8030570cd36286a61

## How Has This Been Tested

I expect the CI to not be red anymore.


mathfac:
@mosra thanks for fixing this!

PR #161, opened 2019-08-19
Fix a bunch of silly crashes.

### LINKED ISSUE #157: Segfault in Viewer Utility
## Steps to reproduce

  1. Compile the code; 
  2. run viewer with a scene without providing the navmesh;
  3.

## Observed Results

   segfault.

## Expected Results

  Viewer should performs as a virtual touring system for the scene.

## Relevant Code

The problem comes from the following line in viewer.cpp
```
const vec3f position = pathfinder_->getRandomNavigablePoint();

```
It segfaults if the pathfinder is not initialized before this line. 

@aclegg3

mosra:
## Motivation and Context

First commit fixes #157, accessing pathfinder only if it's loaded. Reproducible by doing the following:

```sh
cd path/to/mp3d/82sE5b5pLXE
path/to/viewer 82sE5b5pLXE_semantic.ply    # boom, there's no 82sE5b5pLXE_semantic.navmesh
```

Second commit fixes the following:

```sh
cd path/to/any/house
path/to/viewer house.json         # boom
path/to/viewer $(pwd)/house.json  # doesn't boom
```

## How Has This Been Tested

Doesn't crash as much as before -- however the pathfinder should be more robust I think, accessing it while it's not loaded should not segfault.

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -6,6 +6,7 @@
 
 #include <Corrade/Containers/ArrayViewStl.h>
 #include <Corrade/PluginManager/Manager.h>
+#include <Corrade/Utility/Directory.h>
 #include <Corrade/Utility/String.h>
 #include <Magnum/EigenIntegration/GeometryIntegration.h>
 #include <Magnum/EigenIntegration/Integration.h>
@@ -49,6 +50,8 @@
 #include "esp/gfx/PTexMeshDrawable.h"
 #include "esp/gfx/PTexMeshShader.h"
 #endif
+
+namespace Cr = Corrade;
 
 namespace esp {
 namespace assets {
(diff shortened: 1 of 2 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -1099,10 +1102,12 @@ bool ResourceManager::loadSUNCGHouseFile(const AssetInfo& houseInfo,
                                          scene::SceneNode* parent,
                                          DrawableGroup* drawables) {
   ASSERT(parent != nullptr);
-  const std::string& houseFile = houseInfo.filepath;
+  std::string houseFile = Cr::Utility::Directory::join(
```
erikwijmans: Maybe only do this if the path doesn't exist?  Otherwise people always have to specify a relative path.

mosra: It always exists ;) ... only it's relative in some cases (relative to CWD) and absolute in others -- in both cases `exists()` returns `true`.

And, (I think I mentioned this a few times already) `join()` works the same as Python's `os.path.join()` -- so if you `join(whatever, absolutePath)`, you *always* get `absolutePath` unchanged.

erikwijmans: Ah!  I didn't know python's `join` did that :)

mosra: In my opinion that's *the* main feature of it that makes `join()` so immensely useful and failproof :)

PR #162, opened 2019-08-19
Clean up the primitive ID textured shader, resolve the WebGL TODO

mosra:
## Motivation and Context

Follow-up to #151, doing a cleanup pass over the `PrimitiveIDTexturedShader`. There was a few outdated assumptions about how GPUs work, unnecessarily large used data types and inconsistent attribute definitions. In short:

- Using a rectangular 16-bit texture instead of a 32-bit power-of-two texture, saving more than half of the memory. The textures can be neither power-of-two nor square, those restrictions are useful only in order to save memory when having mipmapping (which is not the case here). Aassuming the object IDs never go over 65k, we can also save a half of the memory by using a smaller type.
- Using `texelFetch()` instead of `texture()` -- that way we sidestep all coordinate rounding, filtering etc., making the texel access faster, *and* we don't need to pass texture size to the shader. That nicely resolves the WebGL TODO as well.
- Using `usampler2D` instead of `sampler2D`, avoiding more rounding
- Defining vertex attributes consistently with Magnum's builtin shaders (so the mesh can be evenutally used with those as well) -- the original definition was matching neither the shader code nor the mesh setup, which made it useless
- Using packed 24bit data types for colors instead of expanding them to three-component floats

## How Has This Been Tested

Tests pass locally. Unfortunately the PLY files can't be tested on the CI, so the green tick doesn't mean it works everywhere -- if you can, please try it out locally with MP3D data as well. The `FRLInstanceMeshData` update is done with best intentions in mind but as far as I understood it's a dead code.

All in all, this update could have a significant effect on rendering performance of PLY files -- didn't measure, tho.

### DIFF
--- a/src/esp/assets/GenericInstanceMeshData.cpp
+++ b/src/esp/assets/GenericInstanceMeshData.cpp
@@ -249,9 +249,9 @@ void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
   renderingBuffer_->mesh.setPrimitive(Magnum::GL::MeshPrimitive::Triangles)
       .setCount(cpu_ibo_.size() * 3)
       .addVertexBuffer(renderingBuffer_->vbo, 0,
-                       Magnum::GL::Attribute<0, Magnum::Vector3>{})
+                       gfx::PrimitiveIDTexturedShader::Position{})
       .addVertexBuffer(renderingBuffer_->cbo, 0,
-                       Magnum::GL::Attribute<1, Magnum::Color3>{})
+                       gfx::PrimitiveIDTexturedShader::Color3{})
       .setIndexBuffer(renderingBuffer_->ibo, 0,
                       Magnum::GL::MeshIndexType::UnsignedInt);
 
--- a/src/esp/assets/GenericInstanceMeshData.cpp
+++ b/src/esp/assets/GenericInstanceMeshData.cpp
@@ -27,8 +27,12 @@
 
 #include "esp/core/esp.h"
 #include "esp/geo/geo.h"
+#include "esp/gfx/PrimitiveIDTexturedShader.h"
 #include "esp/io/io.h"
 #include "esp/io/json.h"
+
+namespace Cr = Corrade;
+namespace Mn = Magnum;
 
 namespace esp {
 namespace assets {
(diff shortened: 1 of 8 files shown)


src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -42,33 +46,30 @@ void copyTo(std::shared_ptr<tinyply::PlyData> data, std::vector<T>& dst) {
...
-                    Magnum::SamplerWrapping::ClampToEdge})
-      .setStorage(1, Magnum::GL::TextureFormat::R32F, image.size())
-      .setSubImage(0, {}, image);
-
-  return tex;
+Mn::GL::Texture2D createInstanceTexture(
+    Cr::Containers::ArrayView<std::uint32_t> objectIds) {
+  const Mn::Vector2i textureSize{
+      gfx::PrimitiveIDTexturedShader::PrimitiveIDTextureWidth,
+      int(objectIds.size() +
```
erikwijmans:
```suggestion
      int(std::ceil(float(objectIds.size()) / gfx::PrimitiveIDTexturedShader::PrimitiveIDTextureWidth))
```

mosra: Well. I'm usually going over code and replacing `ceil(float())` with the pure-integer solution, not the other way around :laughing:

Are you sure this would work correctly for values that have more than ~23 bits of precision? One PLY file has 9M of primitives, which is 24 bits. I'd trust the integer solution there instead ;)

erikwijmans: I pretty much always prefer `ceil(float())` for readability :) -- The precision definitely isn't an issue here.  The precision just needs to be good enough to separate things by `gfx::PrimitiveIDTexturedShader::PrimitiveIDTextureWidth` (which, since it's a power of 2, can be perfectly represented a float).

erikwijmans: But this is entirely a personal preference, I am fine with it either way :)

mosra:
Pushed the typo fix, since it's a non-trivial change I'd like to have at least one thorough review.

And since this fixes the WebGL TODO, any chance somebody could test that? :)

mosra:
Oh, this is not merged yet? Can somebody give me one more approval, please?

PR #164, opened 2019-08-20
GPU-based depth unprojection

mosra:
## Motivation and Context

Added a test and a benchmark for the existing code, applied an even better optimization by @msbaines from #150 and then added a GPU-based implementation -- both a depth-only shader and a depth unprojection shader, which is meant to be used in GPU-to-GPU cases to avoid a CPU roundtrip.

While benchmarking I realized there's a huge perf penalty coming from the branch that converts depth values at the far plane to `0.0f`. Run the benchmark for yourself, results for posterity here:

image

In general, the CPU code could be sped up about five times by removing the branching -- ~~according to @erikwijmans this is done because users expect depth values at far plane to be zero (while for me, coming from GPU background, values at far plane should have the depth value of far plane). I'm leaving the decision on you, but as far as I'm concerned, if this proves to be a perf bottleneck, I'd vote for removing the branching.~~ :heavy_check_mark: Mostly mitigated by doing the branching in a separate loop, see [my update below](https://github.com/facebookresearch/habitat-sim/pull/164#issuecomment-523413599).

OTOH, the GPU code doesn't seem to be affected by the branch -- both variants run roughly the same.

## How Has This Been Tested

While the shaders themselves aren't used anywhere yet (assuming those will get integrated into #114 by @erikwijmans), there's a full-blown test and a benchmark for the whole thing. This is the first test that uses [Corrade::TestSuite](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html) -- I'm happy to answer questions about its usage here.

Additionally, the unprojection algorithm docs are now directly in the header, written using *real* math -- I'm aiming to submit a PR with [m.css](https://mcss.mosra.cz)-based doc setup later this week (unless something with a higher priority comes up again): 

image

### DIFF
--- a/src/esp/gfx/DepthUnprojection.cpp
+++ b/src/esp/gfx/DepthUnprojection.cpp
@@ -99,15 +99,20 @@ Mn::Vector2 calculateDepthUnprojection(const Mn::Matrix4& projectionMatrix) {
 
 void unprojectDepth(const Mn::Vector2& unprojection,
                     Cr::Containers::ArrayView<Mn::Float> depth) {
-  for (std::size_t i = 0; i != depth.size(); ++i) {
-    /* We can afford using == for comparison as 1.0f has an exact
-       representation and the depth is cleared to exactly this value. */
-    if (depth[i] == 1.0f) {
-      depth[i] = 0.0f;
-      continue;
-    }
+  for (Mn::Float& d : depth) {
+    d = unprojection[1] / (d + unprojection[0]);
+  }
 
-    depth[i] = unprojection[1] / (depth[i] + unprojection[0]);
+  /* Change pixels on the far plane to be 0. Done in a separate loop to allow
+     the optimizer to vectorize the above better.  */
+  const Mn::Float farDepth = unprojection[1] / (1.0f + unprojection[0]);
+  for (Mn::Float& d : depth) {
+    /* We can afford using == for comparison as 1.0f has an exact
+       representation, the depth was cleared to exactly this value and the
+       calculation is done exactly the same way in both cases -- thus the
+       result should be bit-exact. */
+    if (d == farDepth)
+      d = 0.0f;
   }
 }
 
--- a/src/esp/gfx/DepthUnprojection.cpp
+++ b/src/esp/gfx/DepthUnprojection.cpp
@@ -1,0 +1,87 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "DepthUnprojection.h"
+
+#include <Corrade/Containers/ArrayView.h>
+#include <Magnum/Math/Functions.h>
+#include <Magnum/Math/Matrix4.h>
+
+namespace Cr = Corrade;
+namespace Mn = Magnum;
+
+namespace esp {
+namespace gfx {
+
+Mn::Matrix2x2 calculateDepthUnprojection(const Mn::Matrix4& projectionMatrix) {
+  /* Inverted projection matrix to unproject the depth value and chop the
+     near plane off. We don't care about X/Y there and the corresponding
+     parts of the matrix are zero as well so take just the lower-right part
+     of it (denoted a, b, c, d).
+
+      x 0 0 0
+      0 y 0 0
+      0 0 a b
+      0 0 c d
+
+     Doing an inverse of just the bottom right block is enough as well -- see
+     https://en.wikipedia.org/wiki/Block_matrix#Block_diagonal_matrices for
+     a proof.
+
+     Taking a 2-component vector with the first component being Z and second
+     1, the final calculation of unprojected Z is then
+
+      | a b |   | z |   | az + b |
+      | c d | * | 1 | = | cz + d |
+
+  */
+  auto unprojection =
+      Mn::Matrix2x2{Mn::Math::swizzle<'z', 'w'>(projectionMatrix[2]),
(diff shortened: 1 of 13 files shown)


src/esp/gfx/Renderer.cpp
```diff
@@ -144,29 +103,9 @@ struct Renderer::Impl {
...
-      /* The following is
-
-          (az + b) / (cz + d)
-
-         See the comment in draw() above for details. */
-      ptr[i] =
-          Math::fma(depthUnprojection_[0][0], z, depthUnprojection_[1][0]) /
-          Math::fma(depthUnprojection_[0][1], z, depthUnprojection_[1][1]);
-    }
+    unprojectDepth(depthUnprojection_,
```
msbaines: Why not unproject into ptr to avoid the memcpy?

src/esp/gfx/DepthUnprojection.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+void unprojectDepth(const Mn::Vector2& unprojection,
+                    Cr::Containers::ArrayView<Mn::Float> depth) {
+  for (std::size_t i = 0; i != depth.size(); ++i) {
+    /* We can afford using == for comparison as 1.0f has an exact
+       representation and the depth is cleared to exactly this value. */
+    if (depth[i] == 1.0f) {
+      depth[i] = 0.0f;
+      continue;
```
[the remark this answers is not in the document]
mosra: Discussed that with @erikwijmans yesteday: https://github.com/facebookresearch/habitat-sim/pull/150#discussion_r315403984 -- not even `?:` made any difference (and I would assume the optimizer would figure out the `continue` as well).

Tried builtin_expect as well. No change either -- the branch predictor cache works well even without, I assume.

src/esp/gfx/Renderer.cpp
mosra: Originally it was like this but then I put the memcpy back when extracting this code to a function. Ultimately the operation *should* be done in-place, with no memcpy, so I future-proofed the function to operate in-place as well. The zero-copy update is waiting in #114.

mosra:
Further thoughts on the branching impact while I'm fighting EGL:

- Not doing an in-place operation would allow me to first calculate everything and then do a second loop, zeroing all outputs where `input == 1.0f`. According to my tests from yesterday this made the operation run only ~1 µs instead of ~2.5 µs. But that goes against the zero-copy perf improvements scheduled in #114.
- I could extract the far plane from the projection matrix and then do a fuzzy test against that on the output values instead of testing for `1.0f` on the input values. But since that's a fuzzy comparison, it may throw away more pixels than strictly necessary. Also might be more costly than a simple `==`.

Opinions? :)

erikwijmans:
The point of zero-copy was for performance, so if a copy improves performance, do a copy!

bigbike:
@mosra :

There is a typo in the last line of the sceenshot of your docs:

d + a', not d - a'

src/esp/gfx/DepthUnprojection.h
```diff
@@ -0,0 +1,141 @@
...
+  \end{array}
+@f]
+
+Finally, to reduce the amount of operations in @ref unprojectDepth(), we
+integrate the constants into @f$ a @f$ and @f$ b @f$, returning @f$ a' @f$ and
+@f$ b' @f$: @f[
+  \begin{array}{rcl}
+    -z & = & \cfrac{b}{2d + (a - 1)} \\[1.25em]
+    -z & = & \cfrac{\frac{1}{2}b}{d + \frac{1}{2}(a - 1)} \\[1.25em]
+    -z & = & \cfrac{b'}{d - a'} ; ~~~~ a' = \frac{1}{2}(a - 1),
```
bigbike: d + a'

mosra: Oh well. Thanks a lot. Making typos in important documentation and then preserving them *forever* in a screenshot is my speciality :see_no_evil:

mosra:
After successfully taming EGL yesterday (@erikwijmans, thanks for the debugging help) the CI now runs green and shows this output for the test:

image

In addition, I modified the CPU implementation to do the branching in a separate loop while keeping the calculation in-place, which resulted in a nice speedup compared to before:

image

And finally I played with [GCC's function multi-versioning](https://lwn.net/Articles/691932/), which should make the binary opt-in for SSE4.2 / AVX2 support in case your CPU & system has the support -- so *theoretically* there should be no need for `-march=native`. Unfortunately the CI has only GCC 5.4 (FMV is fully implemented only since GCC 6), so there it didn't result in any difference. Additionally it seems that AVX512 is not supported by those runners, so I wasn't able to discover the blistering speed of server-grade hardware either. If you are on Clang, FMV is in the works [but not there yet](https://reviews.llvm.org/D51650).

@erikwijmans @msbaines @bigbike @msavva Can you have a look again at the updated code? Ideally if you could pull this branch and run the benchmark locally, with either:

- `-march=native` on Clang, the result ratio should be pretty similar to what I had above,
- or GCC 6+, with FMV even without `-march=native` the speed should be roughly the same (here it was only about ~5% worse compared to native build). And if you can access an AVX512 CPU (it should report it in `/proc/cpuinfo`), let me know -- I'd like to add a FMV annotation for that one too :)

I'm not yet totally sure whether the FMV annotations are a good idea or not if one can use `-march=native`, but I'd vote for merging them in -- we can always remove or define them out if they start to be problematic.

erikwijmans:
We have avx512 on Georgia Tech's cluster!

PR #166, opened 2019-08-20
Add hsim js bindings build to CircleCI.

msbaines:
## Motivation and Context

- Add javascript binding build to CircleCI in order to prevent breakages of emscripten build.
- To enable this, I needed to add logic in build_js.sh to build corrade-rc.

## How Has This Been Tested

CircleCI

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,19 @@ jobs:
               sudo apt-get --yes --force-yes install cuda
               touch ./cuda_installed
               nvidia-smi
+      - run:
+          name: Install emscripten
+          no_output_timeout: 20m
+          background: true
+          command: |
+              if [ ! -f ~/emscripten_installed ]
+              then
+                git clone -q https://github.com/emscripten-core/emsdk.git ~/emsdk
+                cd ~/emsdk
+                ./emsdk install 1.38.38
+                ./emsdk activate 1.38.38
+                touch ~/emscripten_installed
+              fi
       - restore_cache:
           keys:
             - conda-{{ checksum "habitat-sim/.circleci/config.yml" }}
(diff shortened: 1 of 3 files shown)


build_js.sh
```diff
@@ -7,6 +7,21 @@ git submodule update --init --recursive
...
+cmake ../src \
+    -DBUILD_GUI_VIEWERS=OFF \
+    -DBUILD_PYTHON_BINDINGS=OFF \
+    -DBUILD_ASSIMP_SUPPORT=OFF \
+    -DBUILD_DATATOOL=OFF \
+    -DBUILD_PTEX_SUPPORT=OFF
+cmake --build . --target corrade-rc
+popd
+
+export PATH=${CORRADE_RC_PATH}:${PATH}
```
mosra: Why not passing `-DCORRADE_RC_EXECUTABLE="$(pwd)"/build_corrade-rc/deps/corrade/src/Corrade/Utility/corrade-rc` directly to CMake instead? Modifying `PATH` might do unwanted things, especially when Emscripten is involved.

msbaines:
It is now passing and all feedback should be addressed.

.circleci/config.yml
```diff
@@ -192,14 +203,22 @@ jobs:
               export PYTHONPATH=$PYTHONPATH:$(pwd)
               GTEST_COLOR=yes ./build.sh --headless --run-tests
               pytest
+      - run:
```
mathfac: This step increases whole testing for 4 min. Unfortunately, when `background: true` option is used step can't fail the testing and is consider successful. Let's keep it as is right now and parallelize later when we have more understanding what other tests we will run for JS build, physics, C++.

PR #168, opened 2019-08-21
WebGL: Add support for setting initial position and location.

msbaines:
## Motivation and Context

This is required for reproducibility.

## How Has This Been Tested

Verified that agent is now always in the same location at start.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/bindings.html
+++ b/src/esp/bindings_js/bindings.html
@@ -50,13 +50,18 @@
         sensorSpecifications: sensorConfigs
       };
 
+      const agentState = {
+        position: [6.35314, 0.0099186, 19.0241],
+        rotation: [0, 0.946352, 0, 0.323138]
+      };
+
       let sceneConfig = new Module.SceneConfiguration();
       sceneConfig.id = "skokloster-castle.glb";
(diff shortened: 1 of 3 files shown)


angelxuanchang:
This is useful functionality and the changes looks pretty good.  My one suggestion would be to rename the this.state in simenv_embind.js to this.initialAgentState (to make it clear that is it agent state and that is used in reset).  Other comment is that we need to start to have more documentation - in this case, specify what happens with unspecified parameters in the agent state (such as velocity, force, etc - I believe they are all set to 0).

msbaines:
> This is useful functionality and the changes looks pretty good. My one suggestion would be to rename the this.state in simenv_embind.js to this.initialAgentState (to make it clear that is it agent state and that is used in reset). Other comment is that we need to start to have more documentation - in this case, specify what happens with unspecified parameters in the agent state (such as velocity, force, etc - I believe they are all set to 0).

Yes, initialAgentState is much clear. Will add documentation in a separate change. I am thinking of using JSDoc annotations and [documentationjs](https://documentation.js.org/) for the documentation generation.

Issue #42, opened 2019-05-10
Update to use modern CMake practices


mosra:
I'll tackle this once I'm done with current priority tasks. Things to do:

- [ ] replace `include_directories()` with `target_include_directories()`, preferring `PRIVATE` where possible; if a dependency provides a CMake target use that instead
- [ ] remove use of `file(GLOB)`, list sources explicitly (makes the build more predictable without things getting broken when there are untracked files in the working copy)
- [ ] convert `if(${FOO})` to just `if(FOO)`
- [ ] look into TODOs with [CMAKE_CXX_VISIBILITY_PRESET](https://github.com/facebookresearch/habitat-sim/blob/e7d115f034020ae7b91b2d1566ec787fe188580f/src/CMakeLists.txt#L37)

Skylion007:
@mosra Can this be closed?

mosra:
Why? There's still a few good-to-have items from the list that are not done yet. Feel free to tackle those, if you have time.

aclegg3:
@mosra any more updates here?

mosra:
Hmm. Do we have any other place to list such low-priority tech debt TODOs? In my case (with Magnum repos) I have a bunch of such long-lived "aggregate issues" full of checkboxes that act more as a reminder of what needs to be *eventually* solved, and I opened this one with a similar intent. 

But if the aim here is to only have an issue open if it's likely to be resolved soon, feel free to close.

aclegg3:
I'll leave this open for now until we establish such a list and then we can migrate the remainder there.

PR #174, opened 2019-08-27
Lots of warnings and all warnings as errors for esp code

msbaines:
## Motivation and Context

Catch more bugs at compile-time.

## How Has This Been Tested

Build.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/CMakeLists.txt
+++ b/src/esp/CMakeLists.txt
@@ -1,0 +1,30 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# https://embeddedartistry.com/blog/2017/3/7/clang-weverything
+# TODO(msb) enable CORRADE_USE_PEDANTIC
+# set_directory_properties(PROPERTIES CORRADE_USE_PEDANTIC_FLAGS ON)
+
+option(BUILD_WARNINGS_AS_ERRORS "Build with warnings as errors" OFF)
+if (BUILD_WARNINGS_AS_ERRORS)
+  if (MSVC)
+    add_compile_options(/WX)
+  else()
+    # TODO(msb) remove -Wall once we enable CORRADE_USE_PEDANTIC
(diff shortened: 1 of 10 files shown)


src/esp/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# https://embeddedartistry.com/blog/2017/3/7/clang-weverything
+# lots of warnings and all warnings as errors
+# TODO(msb) turn on -Wextra
+add_compile_options(-Wall -Werror)
```
mosra: Please don't :) `-Werror` breaks as soon as a new version of a compiler decides to throw warnings for something that didn't warn before. Or older compilers tend to warn on something that is treated as a standard practice in later versions and is just too annoying in the long run. We should still strive for a zero-warning build, but without suffering too much :)

This is also Clang- ang GCC-specific and since we'll have a MSVC build soon (#84 and mosra/magnum-bindings#2), let's not make the porting harder for no reason ;) What I suggested elsewhere and have in TODOs in #42 is `CORRADE_USE_PEDANTIC_FLAGS` -- it contains a broad set of useful warnings, works on MSVC as well (and also disables ones that are just noise). It can be used like this:

```cmake
set_directory_properties(PROPERTIES CORRADE_USE_PEDANTIC_FLAGS ON)
```

Oh -- this got already used on a few places since #91, so if you can, search for `CORRADE_USE_PEDANTIC_FLAGS`
and remove the other occurences, as they would be redundant now :) Also it enables `-Wextra`, so it might bring you a bit more work.

mosra: Sure, just saw the commit -- why not go all the way and enable `CORRADE_USE_PEDANTIC_FLAGS`, then? Did it result it many more complaints? :)

mosra: OK, makes sense :+1: I was under a false assumption that it got in a pretty good shape already :)

PR #186, opened 2019-09-03
Fix the imports!

erikwijmans:
## Motivation and Context

I didn't rebase #114 properly before landing.  Let's fix that!

## How Has This Been Tested

Tests


mosra:
The cleanup is happening much sooner than I expected, ha! :+1: 

If the python docs from #71 still generate correctly without any new warnings, then I'm happy too :)

erikwijmans:
> If the python docs from #71 still generate correctly without any new warnings, then I'm happy too :)

Works!

PR #188, opened 2019-09-04
Cleanup physics things

erikwijmans:
## Motivation and Context

- Use configure.h introduced in #143.
- Fix a segfault I found while testing the bullet build.
- Add a `--with-bullet` alias to match the other `--with-*`.

## How Has This Been Tested

Locally

## Types of changes

### DIFF
--- a/src/esp/gfx/Viewer.h
+++ b/src/esp/gfx/Viewer.h
@@ -61,8 +61,11 @@ class Viewer : public Magnum::Platform::Application {
                                    const Magnum::Vector2i& position);
 
   assets::ResourceManager resourceManager_;
-  std::shared_ptr<physics::PhysicsManager> physicsManager_;
+  // SceneManager must be before physicsManager_ as the physicsManager_
+  // assumes that it "owns" things owned by the scene manager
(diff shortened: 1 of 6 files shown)


src/esp/gfx/Viewer.h
```diff
@@ -61,8 +61,11 @@ class Viewer : public Magnum::Platform::Application {
                                    const Magnum::Vector2i& position);
 
   assets::ResourceManager resourceManager_;
-  std::shared_ptr<physics::PhysicsManager> physicsManager_;
+  // SceneManager must be before physicsManager_ as the physicsManager_
+  // assumes that it "owns" things owned by the scene manager
```
mathfac: Sounds like double ownership. Did you mean expect instances already initiated in `SceneManager`?

erikwijmans: No.  It's kinda double ownership.  The physics manager's logic is simpler if it assumes it owns the magnum scene nodes it manipulates.  This is a fine thing to assume as long as the scene graph those nodes are in has a longer lifetime than the physics manager.

PR #189, opened 2019-09-04
Redwood noisy depth

erikwijmans:
## Motivation and Context

Adds the noise model from here: http://redwood-data.org/indoor/dataset.html

There are 2 implementations:
* A numba based python implementation that runs at ~400 FPS at 256x256
* A CUDA implementation that runs at ~5k FPS at 256x256 or ~200k FPS at 256x256 when using the GPU->GPU transfer from #114 !  I am gonna call ~200k FPS as "fast enough".

@msavva @bigbike @mosra What are your thoughts on how I am currently providing arguments to the sensor noise models?  I just added a `py::dict` to the python binding of the sensor spec so that people can pass arbitrary arguments in it.

TODO

- [x] Doc strings
- [x] License headers
- [x] Add citations
- [x] Figure out how to get the args for constructing noise models -- @mosra's suggestion to have a python only property seems to be a nice middle ground between PySensorSpec and just doing `py::dynamic_attr()`
- [x] Switch to a proper registry pattern in a different PR and use that here -- #204 
- [x] Remove the `tmp.py` file.  That has just been useful for testing :)

## How Has This Been Tested

Via the tests and visually!

redwood_depth

depth

## Types of changes


dhruvbatra:
> A CUDA implementation that runs at ~5k FPS at 256x256 or ~200k FPS at 256x256 when using the GPU->GPU transfer from #114 !

200k?!! Damn. 

Are those two images w/ and w/o noise? 

Request -- is it easy to show images with varying levels of noise? (Or is that not feasible for this noise type?)

erikwijmans:
> Are those two images w/ and w/o noise?

Yes they are :)

> Request -- is it easy to show images with varying levels of noise? (Or is that not feasible for this noise type?)

We could make that doable with this noise model.  There are two different Gaussian that are sampled from and I could make the variance on those changeable.

erikwijmans:
redwood_depth

redwood_depth
redwood_depth

redwood_depth


I add a simple multiplier parameter which I just multiply both the standard deviations by.  Above is 2x, 3x, 4x, 5x.  Seems to fairly smoothly make things worse.

src/esp/bindings/bindings.cpp
```diff
@@ -46,6 +50,13 @@ SceneNode* nodeGetter(T& self) {
     throw py::value_error{"feature not valid"};
   return &self.node();
 };
+
+struct PySensorSpec : SensorSpec {
+  py::dict noise_model_kwargs;
```
mosra: Since this field is never used from C++, what about adding this attribute from Python side only? Pybind-exposed types don't have `__dict__` by default (just `__slots__`), defining `SensorSpec` with [py::dynamic_attributes()](https://pybind11.readthedocs.io/en/stable/classes.html#dynamic-attributes) would make that work. Then, in Python, you could

```py
sensor.noise_model_kwargs = {
    "gpu_device_id": self._sim.gpu_device
}

...

if hasattr(sensor, 'noise_model_kwargs'):
    ...
```

Alternatively, to avoid the `PySensorSpec`, using `def_property()` to define the `noise_model_kwargs` as a property that gets pulled out with `py::getattr()` if it's there, and returning an empty `py::dict` if not.

mosra: Hm, and why not just this, once `SensorSpect` has the `__dict__`:

```py
sensor.gpu_device_id = self._sim.gpu_device

...
```

I think these could be added directly as instance attributes without losing any clarity.

erikwijmans: The `gpu_device_id` was really only meant as a argument to the noise model.  I will rewrite to clarify that.

erikwijmans: > Pybind-exposed types don't have __dict__ by default (just __slots__), defining SensorSpec with py::dynamic_attributes() would make that work. 

I had originally used this, but that felt wrong as its highly undiscoverable and seems harder to document properly.

erikwijmans: > Alternatively, to avoid the PySensorSpec, using def_property() to define the noise_model_kwargs as a property that gets pulled out with py::getattr() if it's there, and returning an empty py::dict if not.

I will give this a try

erikwijmans:
This is ready for review.

habitat_sim/sensors/noise_models/sensor_noise_model.py
```diff
@@ -0,0 +1,48 @@
...
+    r"""Base class for all sensor noise models
+
+    :param gpu_device_id: The ID of the CUDA device to use (only applicable to
+    noise models that proivde a CUDA implementation and, generally, if habitat-sim was build with ``--with-cuda``)
+    """
+    gpu_device_id: Optional[int] = None
+
+    @staticmethod
+    @abc.abstractmethod
+    def is_valid_sensor_type(sensor_type: SensorType):
```
mathfac:
```suggestion
    def is_valid_sensor_type(sensor_type: SensorType) -> bool:
```

habitat_sim/sensors/noise_models/sensor_noise_model.py
```diff
@@ -0,0 +1,48 @@
...
+        r"""Used to determine whether or not the noise model
+        is applicable to the sensor type
+
+        :return: True if this noise model can be applied to this sensor input type
+        """
+        pass
+
+    @abc.abstractmethod
+    def apply(self, sensor_observation):
```
mathfac:
```suggestion
    def apply(self, sensor_observation) -> Observations:
```

habitat_sim/simulator.py
```diff
@@ -391,7 +398,7 @@ def get_observation(self):
                 else:
                     tgt.read_frame_rgba_gpu(self._buffer.data_ptr())
 
-                return self._buffer.flip(0).clone()
```
mathfac: Got it, you moved copying logic to `NoNoise`. Still will copying/cloning be valid for all types of sensors?

setup.py
```diff
@@ -377,6 +377,7 @@ def load(filename):
         # add custom build_ext command
         cmdclass=dict(build_ext=CMakeBuild),
         zip_safe=False,
+        include_package_data=True,
```
mathfac: Are we adding some assets?

src/esp/sensor/RedwoodNoiseModel.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "RedwoodNoiseModel.h"
+
+namespace esp {
+namespace sensor {
+
+namespace {
+
+struct CudaDeviceContext {
+  CudaDeviceContext(int deviceId) {
```
mathfac:
```suggestion
  CudaDeviceContext(const int deviceId) {
```

src/esp/sensor/RedwoodNoiseModel.cpp
```diff
@@ -0,0 +1,78 @@
...
+ private:
+  bool setDevice = false;
+  int currentDevice = -1;
+};
+
+}  // namespace
+
+RedwoodNoiseModelGPUImpl::RedwoodNoiseModelGPUImpl(
+    const Eigen::Ref<const Eigen::RowMatrixXf> model,
+    int gpuDeviceId,
```
mathfac: Maybe `const`s.

habitat_sim/sensors/noise_models/no_noise_model.py
```diff
@@ -0,0 +1,23 @@
...
+from habitat_sim.sensors.noise_models.sensor_noise_model import SensorNoiseModel
+
+
+@registry.register_noise_model(name="None")
+class NoSensorNoiseModel(SensorNoiseModel):
+    @staticmethod
+    def is_valid_sensor_type(sensor_type):
+        return True
+
+    def apply(self, x):
```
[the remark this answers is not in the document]
erikwijmans: We had a copy before-hand, so I just rolled the copy into this as it felt cleaner to always apply a noise model and have a no-noise model which just returns a copy.

habitat_sim/sensors/noise_models/sensor_noise_model.py
erikwijmans: I don't believe we have a generic `Observations` type currently defined.

setup.py
erikwijmans: Yes, the distortion model.

habitat_sim/sensors/noise_models/redwood_depth_noise_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+            # Downsample
+            d = gt_depth[y - y % 2, x - x % 2]
+            # If the depth is greater than 10, the sensor will just return 0
+            if d >= 10.0:
+                noisy_depth[j, i] = 0.0
+            else:
+                # Distort
+                undistorted_d = _undistort(
+                    int(x / xmax * 639.0 + 0.5), int(y / ymax * 479.0 + 0.5), d, model
```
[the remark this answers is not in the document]
erikwijmans: 480x640 is the resolution of the sensor used to create the distortion model originally.  This logic just rescales any given observation to 480x640 so that it can query the distortion model.

habitat_sim/simulator.py
erikwijmans: For all visual sensors, yes. This current sensor class is really for just visual sensors.  When we introduce other sensors, I would imagine those will have their own separate python wrapper.

src/esp/sensor/RedwoodNoiseModel.cpp
mosra: And I'd say `explicit` as well :)

docs/noise_models.rst
```diff
@@ -0,0 +1,53 @@
...
+.. py:class:: habitat_sim.sensors.noise_models.RedwoodDepthNoiseModel
+    :summary: Redwood Noise Model for PrimSense depth cameras
+
+    Accessible from the registry under the name ``"RedwoodDepthNoiseModel"``
+
+    Implements the noise model provided by http://redwood-data.org/indoor/dataset.html
+
+    If you use this noise model, please cite::
+
+        @inproceedings{choi2015robust,
```
mosra: Putting this into `.. code:: bibtex` will make it highlighted the same way as is on https://aihabitat.org/#citing-habitat

erikwijmans: Ooooo, fancy.

src/esp/sensor/RedwoodNoiseModel.h
```diff
@@ -0,0 +1,85 @@
...
+   *                          can be used to increase or decrease the noise
+   *                          level
+   */
+  RedwoodNoiseModelGPUImpl(const Eigen::Ref<const Eigen::RowMatrixXf> model,
+                           const int gpuDeviceId,
+                           const float noiseMultiplier);
+
+  /**
+   * @brief Simulates noisy depth from clean depth.  The input is assumed to be
+   * on the CPU and the output will be on the CPU
```
msavva: Maybe add: what happens if assumption does not hold?

src/esp/sensor/RedwoodNoiseModel.h
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @param[in] depth  Clean depth, i.e. depth from habitat's depth shader
+   * @return Simulated noisy depth
+   */
+  Eigen::RowMatrixXf simulateFromCPU(
+      const Eigen::Ref<const Eigen::RowMatrixXf> depth);
+
+  /**
+   * @brief Similar to @ref simulateFromCPU() but the input and output are
+   * assumed to be on the GPU
```
msavva: Maybe add: what happens if assumption does not hold?

src/esp/sensor/RedwoodNoiseModel.h
erikwijmans: Bad things happen.  Segfaults happen :)

src/esp/sensor/RedwoodNoiseModel.h
erikwijmans: Bad things happen. Segfaults happen :)

src/esp/sensor/RedwoodNoiseModel.cu
```diff
@@ -0,0 +1,181 @@
...
+#include <algorithm>
+
+#include <cuda_runtime.h>
+#include <curand_kernel.h>
+
+namespace {
+const int MODEL_N_DIMS = 5;
+const int MODEL_N_COLS = 80;
+
+__device__ float undistort(const int _x,
```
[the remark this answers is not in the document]
erikwijmans: I added the link, http://www.alexteichman.com/octo/clams/ but that doesn't really help the implementation read-ability.  I just took the code from here: http://redwood-data.org/indoor/data/simdepth.py and made it not slow.

habitat_sim/sensors/noise_models/no_noise_model.py
```diff
@@ -0,0 +1,24 @@
...
+@registry.register_noise_model(name="None")
+class NoSensorNoiseModel(SensorNoiseModel):
+    @staticmethod
+    def is_valid_sensor_type(sensor_type: SensorType) -> bool:
+        return True
+
+    def apply(self, x):
+        if isinstance(x, np.ndarray):
+            return x.copy()
+        else:
```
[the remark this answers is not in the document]
erikwijmans: Added a check to see if `x` is a torch tensor.

PR #190, opened 2019-09-04
Support ptex mesh - step 1: fix texture binding points, number of primitives.

bigbike:
We would like to support ptex mesh rendering in our simulator.

Status before step 1:
Code was implemented, no compile error; 
Cannot render the ptex mesh in the utility viewer or in headless mode;
The viewer can be run but returns a black screen;

In this PR:
-) it fixes the texture binding points for adjFaces (buffer texture) and atlas textures;
-) it sets correct number of primitives when setup the magnum;
-) it updates the Asset.cpp to recognize the ptex mesh in publicly released Replica data set;

## Motivation and Context

1st step to support ptex mesh rendering

## How Has This Been Tested
Local. No compile error.
If would eventually work (tested in #132) once all the mandatory fix are done.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/Asset.cpp
+++ b/src/esp/assets/Asset.cpp
@@ -14,12 +14,11 @@ using namespace Corrade::Utility::String;
 AssetInfo AssetInfo::fromPath(const std::string& path) {
   AssetInfo info{AssetType::UNKNOWN, path};
 
-  if (endsWith(path, "semantic_quad_mesh.ply")) {
-    info.type = AssetType::FRL_INSTANCE_MESH;
-  } else if (endsWith(path, "_semantic.ply")) {
+  if (endsWith(path, "_semantic.ply")) {
     info.type = AssetType::INSTANCE_MESH;
-  } else if (endsWith(path, "ptex_quad_mesh.ply")) {
+  } else if (endsWith(path, "mesh.ply")) {
     info.type = AssetType::FRL_PTEX_MESH;
+    info.frame = {quatf::FromTwoVectors(geo::ESP_GRAVITY, -vec3f::UnitZ())};
   } else if (endsWith(path, "house.json")) {
     info.type = AssetType::SUNCG_SCENE;
   } else if (endsWith(path, ".glb")) {
--- a/src/esp/assets/Asset.cpp
+++ b/src/esp/assets/Asset.cpp
@@ -14,16 +14,10 @@ using namespace Corrade::Utility::String;
 AssetInfo AssetInfo::fromPath(const std::string& path) {
   AssetInfo info{AssetType::UNKNOWN, path};
 
-  if (endsWith(path, "mesh_semantic.ply")) {
-    info.type = AssetType::FRL_INSTANCE_MESH;
-  } else if (endsWith(path, "_semantic.ply")) {
-    ASSERT(!endsWith(path, "mesh_semantic.ply"));
+  if (endsWith(path, "_semantic.ply")) {
     info.type = AssetType::INSTANCE_MESH;
   } else if (endsWith(path, "mesh.ply")) {
     info.type = AssetType::FRL_PTEX_MESH;
-    // the gravity in replica dataset is -Z;
-    // create a coordinate for the mesh by rotating the default ESP
-    // coordinate frame to -Z gravity
     info.frame = {quatf::FromTwoVectors(geo::ESP_GRAVITY, -vec3f::UnitZ())};
   } else if (endsWith(path, "house.json")) {
     info.type = AssetType::SUNCG_SCENE;
--- a/src/esp/assets/Asset.cpp
+++ b/src/esp/assets/Asset.cpp
@@ -14,10 +14,16 @@ using namespace Corrade::Utility::String;
 AssetInfo AssetInfo::fromPath(const std::string& path) {
   AssetInfo info{AssetType::UNKNOWN, path};
 
-  if (endsWith(path, "_semantic.ply")) {
+  if (endsWith(path, "mesh_semantic.ply")) {
+    info.type = AssetType::FRL_INSTANCE_MESH;
+  } else if (endsWith(path, "_semantic.ply")) {
+    ASSERT(!endsWith(path, "mesh_semantic.ply"));
     info.type = AssetType::INSTANCE_MESH;
   } else if (endsWith(path, "mesh.ply")) {
     info.type = AssetType::FRL_PTEX_MESH;
+    // the gravity in replica dataset is -Z;
+    // create a coordinate for the mesh by rotating the default ESP
+    // coordinate frame to -Z gravity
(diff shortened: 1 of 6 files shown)


bigbike:
@all:

It is ready for review. Thanks.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -602,8 +602,13 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
                              Magnum::GL::BufferUsage::StaticDraw);
-    currentMesh->mesh.setPrimitive(Magnum::GL::MeshPrimitive::LinesAdjacency)
-        .setCount(currentMesh->ibo.size() / 2)
+    currentMesh->mesh
+        .setPrimitive(Magnum::GL::MeshPrimitive::LinesAdjacency)
+        // Warning:
+        // CANNOT use currentMesh.ibo.size() when calling
+        // setCount because that returns the number of bytes of the buffer, NOT
+        // the index counts
+        .setCount(submeshes_[iMesh].ibo.size())
```
erikwijmans: Maybe `currentMesh->ibo.size() / sizeof(currentMesh->ibo[0])` then?

erikwijmans:
Yeah, I do think we should be shooting for adding at least minimal documentation for new API.  More detailed docs for more complicated pieces.

src/esp/assets/Asset.cpp
```diff
@@ -14,12 +14,11 @@ using namespace Corrade::Utility::String;
 AssetInfo AssetInfo::fromPath(const std::string& path) {
   AssetInfo info{AssetType::UNKNOWN, path};
 
-  if (endsWith(path, "semantic_quad_mesh.ply")) {
-    info.type = AssetType::FRL_INSTANCE_MESH;
```
erikwijmans: Should we also remove `FRLInstanceMeshData.*`?  This was the only place that code could have possibly be run IIRC.

mosra: Good point, I think so too.

bigbike: image

The latest "FRL_INSTANCE_MESH", which has the name "mesh_semantic.ply" and can be actually handled (and it is handled) by the AssetType::INSTANCE_MESH (see the picture above).

Is there any different (in terms of rendering, model training etc.) when we use FRL_INSTANCE_MESH and INSTANCE_MESH? Any exclusive functionalities designed for each type? If there is no difference, I think we can deprecate one (e.g., FRL_INSTANCE_MESH).

Thanks for rising up this issue. Yes, even if we decide to deprecate one, we can ship it in some other PR.

src/esp/assets/PTexMeshData.cpp
bigbike: @erikwijmans : The naming is an issue. See the discussions about this piece of code if you are interested in #132.
A following PR will be made very soon to rename them.

bigbike:
> Yeah, I do think we should be shooting for adding at least minimal documentation for new API. More detailed docs for more complicated pieces.

@erikwijmans @mosra : Would any of you give my a sample .cpp or .h, which contains such formatted docs, so that I can follow?

mosra:
https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/DepthUnprojection.h is pretty complete I would say. Or look into any Magnum header for examples.

I still need to add a guide for setting up all dependencies and generating docs, working on that. Ping me on Slack if you get into a trouble :)

src/esp/assets/Asset.cpp
```diff
@@ -14,12 +14,17 @@ using namespace Corrade::Utility::String;
 AssetInfo AssetInfo::fromPath(const std::string& path) {
   AssetInfo info{AssetType::UNKNOWN, path};
 
-  if (endsWith(path, "semantic_quad_mesh.ply")) {
+  if (endsWith(path, "mesh_semantic.ply")) {
```
bigbike: I just restored the FRL_INSTANCE_MESH.
WARNING: It will segfault now.
The old code somehow cannot handle the new replica instance model.
This put me in a dilemma: 
a) fix the segfault;
b) retire the FRL_INSTANCE_MESH directly in the future;

src/esp/assets/Asset.cpp
erikwijmans: The `FRL_INSTANCE_MESH` was for that initial scan of the apartment.  For replica, I rewrote the instance mesh exporter to be similar to the MP3D format so that we could load it via generic instance mesh.  The replica meshes do not get loaded by `FRL_INSTANCE_MESH`/ can't get loaded as that type.

src/esp/assets/Asset.cpp
erikwijmans: No no no.  `semantic_quad_mesh.ply` was the correct ends with.  That mesh format is not for the replica meshes. See my other comment.

bigbike: @erikwijmans : Then this is the old format with the SurrealApt, and I do not think we need to support it any more.
Shall we retire this type?

src/esp/assets/Asset.cpp
bigbike: @erikwijmans : Good to know. Then we never ever use the "initial scan" in the released replica dataset. Correct?
Shall we just retire it? (I should revert my changes again.) 
I remembered you suggested we did not have to support the old replica format any more, and we agreed in the group conversation a couple of weeks ago.

src/esp/assets/Asset.cpp
erikwijmans: Yes

src/esp/assets/Asset.cpp
erikwijmans: Yeah, that scan was never publicly release and I don't think anyone needs to load it for any reason.

src/esp/gfx/PTexMeshShader.h
```diff
@@ -27,11 +27,12 @@ class PTexMeshShader : public Magnum::GL::AbstractShaderProgram {
 
   explicit PTexMeshShader();
 
-  PTexMeshShader& bindTexture(Magnum::GL::Texture2D& texture,
-                              uint32_t textureUnit = 0) {
-    texture.bind(textureUnit);
-    return *this;
-  }
+  // ======== texture binding ========
+  // Note: the texture binding points are explicitly specified in .cpp
```
msavva: This looks like a comment that should go in the implementation file. Someone looking at the header to understand the API wouldn't a priori have any expectations of explicit binding point specification or care about GL version constraints, and this would be confusing text, in my opinion

src/esp/assets/Asset.cpp
msavva: Somehow, I was under the impression we had already deprecated it a while back... In any case, agree, there should be a separate PR to deprecate it fully and avoid any future confusion like the above.

bigbike:
Made changes to address reviewers' concern.
Commit it in to unblock the future PR.

PR #191, opened 2019-09-05
Make semantic mesh shader work with WebGL2

msbaines:
The semantic mesh shader relies on gl_PrimitiveID which is not
available in GLES3.0 and hence WebGL2. Instead, create a combined
set of indices.

Tested with 17DRP5sb8fy_semantic.ply:

vertices before: 1522546
vertices after:  1572308

Since the increase is only 3%, the benefits of having a separate
gl_PrimitiveID shader are not large enough to warrant the complexity
of supporting two shaders.

I've only tested one mesh so the overhead could be larger on other
meshes.

## Motivation and Context

The semantic mesh shader relies on gl_PrimitiveID which is not
available in GLES3.0 and hence WebGL2. Instead, create a combined
set of indices.

## How Has This Been Tested

Tested with 17DRP5sb8fy_semantic.ply:

vertices before: 1522546
vertices after:  1572308

## Types of changes

## Checklist


src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -229,17 +255,24 @@ void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
...
+      Magnum::GL::Buffer::TargetHint::ElementArray);
+  renderingBuffer_->ibo.setData(indices, Magnum::GL::BufferUsage::StaticDraw);
   renderingBuffer_->mesh.setPrimitive(Magnum::GL::MeshPrimitive::Triangles)
-      .setCount(cpu_ibo_.size() * 3)
+      .setCount(cpu_ibo_.size())
+      .addVertexBuffer(renderingBuffer_->obo, 0,
+                       gfx::PrimitiveIDTexturedShader::ObjectId{
+                           gfx::PrimitiveIDTexturedShader::ObjectId::DataType::
+                               UnsignedShort})
       .addVertexBuffer(renderingBuffer_->vbo, 0,
```
mosra: Now that we have three buffers, I think the `obo`, `vbo`, `cbo` could be interleaved into a single buffer (you could use [MeshTools::interleave()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a006bda8a1901669404adcc50b4b68735)) -- it'll give the GPU a better memory locality in the vertex shader and the setup would be just a single `addVertexBuffer()` call:

```cpp
// order of arguments to interleave() and to addVertexBuffer() has to match ;)
renderingBuffer_->vbo.setData(MeshTools::interleave(cpu_vbo_, cpu_cbo_, cpu_obo_); 
renderingBuffer_->mesh.addVertexBuffer(renderingBuffer_->vbo, 0,
    gfx::PrimitiveIDTexturedShader::Position{},
    gfx::PrimitiveIDTexturedShader::Color3{
              gfx::PrimitiveIDTexturedShader::Color3::DataType::UnsignedByte,
              gfx::PrimitiveIDTexturedShader::Color3::DataOption::Normalized}),
    gfx::PrimitiveIDTexturedShader::ObjectId{
               gfx::PrimitiveIDTexturedShader::ObjectId::DataType::UnsignedShort});
```

For an even more minimalistic setup, you could `std::move(vbo)` into the `addVertexBuffer()` call (and the same for `setIndexBuffer()`), then you need to store just the mesh and all buffers are owned by it ([docs on this](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Mesh.html#GL-Mesh-buffer-ownership)).

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -40,10 +42,32 @@ namespace assets {
 namespace {
 template <typename T>
 void copyTo(std::shared_ptr<tinyply::PlyData> data, std::vector<T>& dst) {
-  dst.resize(data->count);
+  dst.resize(data->buffer.size_bytes() / sizeof(T));
```
erikwijmans: This was done like that for a reason -- it makes sure that you didn't try to read a `uint32_t` field as `uint16_t`, which would be bad.

mosra: `ASSERT(data->count*sizeof(T) == data->buffer.size_bytes());` then?

erikwijmans: Yeah, that'll do it too :)

erikwijmans:
I ran over all of MP3D.  Top I saw was a 6% increase in the number of vertices.  It'll probably be less in Replica as the annotations are cleaner.  Will look at that later.

msbaines:
Just realized that I need to fix FRLInstanceMeshData also.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -40,10 +42,32 @@ namespace assets {
...
+    auto result = table.emplace(val, cursor);
+    if (result.second) {
+      data[cursor++] = val;
+    }
+    resultIndices.push_back(result.first->second);
+  }
+  data.resize(cursor);
+
+  return resultIndices;
+}
```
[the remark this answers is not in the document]
msbaines: Sounds good.

src/esp/gfx/PrimitiveIDTexturedShader.h
```diff
@@ -30,6 +30,8 @@ class PrimitiveIDTexturedShader : public Magnum::GL::AbstractShaderProgram {
   typedef Magnum::Shaders::Generic3D::Position Position;
   //! @brief vertex colors
   typedef Magnum::Shaders::Generic3D::Color3 Color3;
+  //! @brief object ids
+  typedef Magnum::GL::Attribute<5, Magnum::UnsignedInt> ObjectId;
```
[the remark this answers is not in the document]
msbaines: Will do. I also need to update the FRL code paths.

src/esp/assets/GenericInstanceMeshData.cpp
msbaines: I will try these. I may punt some of this to a separate change if it gets too involved.

erikwijmans:
Seems to be a little higher in Replica (highest is about 8%), but still fine.

mosra:
As I said above, even the 8% increase should be less used memory *in total*, no? And memory access is the main thing that matters for performance in this case.

src/esp/assets/GenericInstanceMeshData.cpp
msbaines: Yeah, this was not ideal but was required since cpu_ibo_ is now a std::vector<uint32_t> instead of a std::vector<vec3ui>. This was needed in order to use cpu_ibo_ with MeshTools::combineIndexedArrays.

Will use the new ASSERT suggested by mosra.

src/esp/assets/GenericInstanceMeshData.cpp
msbaines: I tried interleave but it didn't work right away. Will debug it later and submit as a follow-up patch.

```
diff --git a/src/esp/assets/GenericInstanceMeshData.cpp b/src/esp/assets/GenericInstanceMeshData.cpp
index 931f218..58329b5 100644
--- a/src/esp/assets/GenericInstanceMeshData.cpp
+++ b/src/esp/assets/GenericInstanceMeshData.cpp
@@ -11,6 +11,7 @@
 #include <Magnum/ImageView.h>
 #include <Magnum/Math/Functions.h>
 #include <Magnum/MeshTools/CombineIndexedArrays.h>
+#include <Magnum/MeshTools/Interleave.h>
 #include <Magnum/MeshTools/RemoveDuplicates.h>
 #include <Magnum/PixelFormat.h>
 #include <Magnum/Trade/Trade.h>
@@ -241,26 +242,21 @@ void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
       std::make_pair(std::cref(cpu_ibo_), std::ref(cpu_vbo_)),
       std::make_pair(std::cref(cpu_ibo_), std::ref(cpu_cbo_)));
   LOG(INFO) << "Mandeep: after: " << cpu_vbo_.size();
-  renderingBuffer_->obo.setData(objectIds_,
-                                Magnum::GL::BufferUsage::StaticDraw);
-  renderingBuffer_->vbo.setData(cpu_vbo_, Magnum::GL::BufferUsage::StaticDraw);
-  renderingBuffer_->cbo.setData(cpu_cbo_, Magnum::GL::BufferUsage::StaticDraw);
+  renderingBuffer_->vbo.setData(
+      Mn::MeshTools::interleave(cpu_vbo_, cpu_cbo_, objectIds_),
+      Magnum::GL::BufferUsage::StaticDraw);
   renderingBuffer_->ibo.setTargetHint(
       Magnum::GL::Buffer::TargetHint::ElementArray);
   renderingBuffer_->ibo.setData(indices, Magnum::GL::BufferUsage::StaticDraw);
   renderingBuffer_->mesh.setPrimitive(Magnum::GL::MeshPrimitive::Triangles)
       .setCount(cpu_ibo_.size())
-      .addVertexBuffer(renderingBuffer_->obo, 0,
-                       gfx::PrimitiveIDTexturedShader::ObjectId{
-                           gfx::PrimitiveIDTexturedShader::ObjectId::DataType::
-                               UnsignedShort})
       .addVertexBuffer(renderingBuffer_->vbo, 0,
-                       gfx::PrimitiveIDTexturedShader::Position{})
-      .addVertexBuffer(
-          renderingBuffer_->cbo, 0,
-          gfx::PrimitiveIDTexturedShader::Color3{
-              gfx::PrimitiveIDTexturedShader::Color3::DataType::UnsignedByte,
-              gfx::PrimitiveIDTexturedShader::Color3::DataOption::Normalized})
+                       gfx::PrimitiveIDTexturedShader::Position{},
+                       gfx::PrimitiveIDTexturedShader::Color3{
+                           gfx::PrimitiveIDTexturedShader::Color3::DataType::UnsignedByte,
+                           gfx::PrimitiveIDTexturedShader::Color3::DataOption::Normalized},
+                       gfx::PrimitiveIDTexturedShader::ObjectId{
+                            gfx::PrimitiveIDTexturedShader::ObjectId::DataType::UnsignedShort})
       .setIndexBuffer(renderingBuffer_->ibo, 0,
                       Magnum::GL::MeshIndexType::UnsignedInt);

```

msbaines:
> As I said above, even the 8% increase should be less used memory _in total_, no? And memory access is the main thing that matters for performance in this case.

Performance should be better since the shader is simpler and doesn't require a texel fetch. Once we get interleave working it should be better still.

src/esp/assets/GenericInstanceMeshData.cpp
mosra: Hmm. Don't see anything wrong with the diff. How exactly it didn't work? Black screen of death?

Maybe the driver is unhappy about data alignment there? If I'm counting correctly, each vertex is 17 bytes (12 bytes for position, 3 for color, 2 for object ID). Can you try adding 3 byte padding after the color to make it 20 bytes (and so the floats are always at 4-byte-aligned positions and object ID at 2-byte-aligned)? It's done by specifying a number in both `interleave()` and `addVertexBuffer()`:

```cpp
renderingBuffer_->vbo.setData(MeshTools::interleave(cpu_vbo_, cpu_cbo_, 3, objectIds_)); 
renderingBuffer_->mesh.addVertexBuffer(renderingBuffer_->vbo, 0,
    gfx::PrimitiveIDTexturedShader::Position{},
    gfx::PrimitiveIDTexturedShader::Color3{
          gfx::PrimitiveIDTexturedShader::Color3::DataType::UnsignedByte,
          gfx::PrimitiveIDTexturedShader::Color3::DataOption::Normalized}),
    3,
    gfx::PrimitiveIDTexturedShader::ObjectId{
           gfx::PrimitiveIDTexturedShader::ObjectId::DataType::UnsignedShort});
```

Of course now, in case this makes it work, this begs a question if this interleaving really helps perf (the memory locality is better, but at the same time, each vertex loads 3 wasted bytes). I guess we won't know until we benchmark :)

msbaines: Yep. That was it. I forgot about alignment. Thanks!

src/esp/assets/GenericInstanceMeshData.h
```diff
@@ -18,25 +17,12 @@
...
-`(gl_PrimitiveID % width, gl_PrimitiveID / width)`.
-*/
-Magnum::GL::Texture2D createInstanceTexture(
-    Corrade::Containers::ArrayView<uint32_t> objectIds);
-
 class GenericInstanceMeshData : public BaseMesh {
  public:
   struct RenderingBuffer {
     Magnum::GL::Mesh mesh;
     Magnum::GL::Buffer vbo;
```
erikwijmans: Maybe rename this to be something like `interleavedVertColorIdBuffer`.  I was confused when I saw `Magnum::GL::Buffer cbo;` deleted in the diff.

src/esp/gfx/PrimitiveIDTexturedDrawable.cpp
```diff
@@ -13,19 +13,16 @@ PrimitiveIDTexturedDrawable::PrimitiveIDTexturedDrawable(
     scene::SceneNode& node,
     PrimitiveIDTexturedShader& shader,
     Magnum::GL::Mesh& mesh,
-    Magnum::SceneGraph::DrawableGroup3D* group /* = nullptr */,
-    Magnum::GL::Texture2D* texture /* = nullptr */)
-    : Drawable{node, shader, mesh, group}, texture_(texture) {}
+    Magnum::SceneGraph::DrawableGroup3D* group /* = nullptr */)
+    : Drawable{node, shader, mesh, group} {}
 
 void PrimitiveIDTexturedDrawable::draw(
```
erikwijmans: `PrimitiveIDTexturedDrawable` -> `PrimitiveIDDrawable`

src/esp/assets/GenericInstanceMeshData.h
mosra: With my suggestion to [std::move() the buffers into the mesh](https://github.com/facebookresearch/habitat-sim/pull/191#discussion_r321323044), this would become just a `GL::Mesh` in the end.

PR #195, opened 2019-09-06
Physics namespace docs and resulting code cleanup.

aclegg3:
## Motivation and Context

Document Habitat-sim physics functionality introduced in PR #91 .

Some code cleanup exposed as the result of documentation. (yay!)

## How Has This Been Tested

Local doc building.

## Types of changes

## Checklist


src/esp/physics/PhysicsManager.h
```diff
@@ -227,9 +230,10 @@ class PhysicsManager {
 
   /** @brief Set the friction coefficient of the scene collision geometry. See
    * @ref sceneNode_.
-   * @param The scalar friction coefficient of the scene geometry.
+   * @param frictionCoefficient The scalar friction coefficient of the scene
+   * geometry.
    */
-  virtual void setSceneFrictionCoefficient(const double){};
+  virtual void setSceneFrictionCoefficient(const double frictionCoefficient){};
```
erikwijmans: This will end up bringing an unused parameter warning back.  On way to deal wit this is to not have the definition inline here.  Just do `virtual void setSceneFrictionCoefficient(const double frictionCoefficient);` then have `void PhysicsManager::setSceneFrictionCoefficient(const double){};` in the `.cpp`

aclegg3: Thanks, Erik. This works well.

mosra: Or put the [CORRADE_UNUSED](https://doc.magnum.graphics/corrade/Macros_8h.html#aaf99b187cce2ff1eadddf1507ace2d69) on it :)

erikwijmans: Oooooo.  That seems even better :)

src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -13,88 +17,381 @@
...
-  //! All components are wrapped into one rigidBody_
   std::vector<std::unique_ptr<btConvexHullShape>> bObjectConvexShapes_;
+
+  //! Object data: All components of the collision shape
   std::unique_ptr<btCompoundShape> bObjectShape_;
+
+  /** @brief Object data: All components of a @ref RigidObjectType::OBJECT are
+   * wrapped into one <A
+   * HREF="https://pybullet.org/Bullet/BulletFull/classbtRigidBody.html">
+   * btRigidBody</A>
```
mosra: ~~You can use Markdown links here, these look a bit less ugly -- `[btRigidBody](https://pybullet.org/Bullet/BulletFull/classbtRigidBody.html)`~~

Actually, I think we could hook up a Doxygen tagfile for all these, similarly how it's done for STL and Magnum type crosslinking. I'll take a look at that. (Sorry that you had to manually add all these.)

aclegg3: Copy paste does get old, haha. If you have an optimization I'm happy to apply it!

mosra: I looked around and Bullet doesn't seem to provide any tagfile on its own, so I would need to generate one myself from the sources. Gonna do that tomorrow, it's getting late here :)

mosra: Done: #222. Have fun :)

aclegg3: Thanks again, mosra! I'll port these as soon as #222 is live.

aclegg3:
Any more comments before we merge this in as a first draft?

PR #196, opened 2019-09-06
Support ptex mesh - step 3: better naming

bigbike:
This PR is to rename a number of buffers, textures to avoid ambiguity and increase the readability of the code. 

-) rename the buffers on GPU side in PTexMeshData;
-) rename variables in PTexMeshDrawable;

## Motivation and Context

It caused a lot of confusion in the past, making it difficult to read, maintain and debug.

## Types of changes

## Checklist


bigbike:

bigbike:
By the way, please take a look at the CI failure. It says `ModuleNotFoundError: No module named 'Cython'`, which I do not think caused by this PR.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -25,8 +25,6 @@
 static constexpr int ROTATION_SHIFT = 30;
 static constexpr int FACE_MASK = 0x3FFFFFFF;
 
-namespace Cr = Corrade;
```
msavva: What is the rationale for getting rid of the shorter namespace alias? We are using this pervasively in other parts of the code (e.g., `ResourceManager`) and I don't think readability is impacted.

bigbike: I understand it may save you a couple of characters when coding. But I actually think it does decrease the readability. When I started this project and read this file, I actually searched what the "Cr" was.

Yes, it is used in a couple of files. But I see more places are using Corrade:: directly. And I prefer this way since it is straightforward.

msavva: `grep -rl 'Cr = Corrade' src/esp | wc` shows 7 cpp files for me.  In my opinion, namespace aliasing in cpp files is quite reasonable, anyone interested in looking at implementation detail will quickly see the alias at the top of the file and understand.  @mosra @erikwijmans @msbaines do you have opinions about this?

msavva: Also, I would say this change (removing a namespace alias) is orthogonal to the "member name improvements" in this PR.  If we make a decision one way or the other, I would still advocate namespace alias changes be done in a separate PR regardless.

erikwijmans: No preference from me.  `Cor<tab>` vs `Cr<tab>` are about the same length.  I can both see an argument for increasing readability by using `Corrade` and increasing readability by reducing the length of names (`Cr::Utility::Directory::MapDeleter` seems better than `Corrade::Utility::Directory::MapDeleter`, for instance).

bigbike: > anyone interested in looking at implementation detail will quickly see the alias at the top of the file and understand. 

Then the one needs to search it to find his answer, like me. If you put Corrade directly in place, one understands what it is, and where it is from immediately. This is based on what actual happens to me.

At the very beginning, I did not connect "Cr" to "Corrade" in my mind. I thought it was something other than Magnum or Corrade. So I believed it decreased the readability. Even someone can guess it is "Corrade", I think she still needs to search it to confirm it. May I understand why you insist it is a must-have here? Again, what only I can see is that it saves you a couple of characters when coding.

Creating a PR involves a lot of overheads. I do not think such issue is big enough to start a completely new PR. Splitting #132 into different PRs (and a lot more are on the way) already slows me down a lot. Creating another one because of this is just too much.

This PR is called "Renaming". But the essence is to increase the readability and to avoid confusion. In this sense, I think putting this change here is quite reasonable, which will not introduce too much burden to the reviewers. Plus the aforementioned overheads of a new PR. I would resolve this issue here. 

@msavva : If you are extremely uncomfortable with this change, I can put it back. But I still stick to my opinion. I simply hope not to quibble about it, and it will not slow me down on improving the simulator. I already start feeling bad to spend so much time writing comments on such a trivial matter.

mosra: My original suggestion couple of months back was to do

```cpp
namespace esp {
    using namespace Magnum;
    using namespace Corrade;
}
```

or, on python side

```py
from corrade import *
from magnum import *
```

but that's a highly opinionated thing to do -- some projects I know of were perfectly happy to go that route, while for some it was a *heresy*. What I suggested next (and what *some* other projects are doing as well), was to use

```cpp
namespace Cr = Corrade;
namespace Mn = Magnum;
```

and on python side

```py
import corrade as cr
import magnum as mn
```

and continue that way, as it's a good middle ground between having lines short enough and not lumping everything into a single namespace. For example, in case of numpy, `import numpy as np` is *so widely understood* that you don't even pause when encountering `np.array`.  So if you need to pause when seeing `Cr`, my humble opinion is that it wasn't used often enough to become ubiquitous similarly as `np` is :)

I'm just a library developer and so I'm not in a position where I should force you to one particular style, the above were just my recommendations -- and we agreed on the `Cr` / `Mn` style back when we were dicussing this, IIRC. A part of the current confusion problem is that those recommendations came quite late in the process, which means the code is using partially one style, partially another. While I'm trying to stick to a single style for new code, existing code is still waiting to be updated to match that.

mosra: While designing the two libs, I was thinking about this a lot -- should I prefix all types and functions (like Qt does, for example) and have them in the root namespace instead? That would be a clutter you can't get rid of. Should I put everything into an easy-to-type namespace like `M::`? Not sure either. I ended up with the current state, giving the user a freedom to do whatever fits.

msavva: Agree with @mosra 's point about readability through widespread use -- moreover the current pattern is nicely consistent between python and C++ (another good rationale for the alias).

As @mosra points out most of the confusion is because we are currently not entirely consistent in other parts of the code.  We did decide in a #habitat-dev discussion ~May to go with the namespace alias above, but now is a good point in time to re-affirm that decision -- or if there are good rationales against, to reverse it. 😅

@msbaines @aclegg3 @angelxuanchang do you have an opinion one way or the other?

I hate to be a stickler, but regardless, I would advocate against making namespace alias changes as part of this PR. The PR summary and changes bullet point list at the top makes no mention of this consideration, and this is certainly a decision with broader implications.

angelxuanchang: I think @mosra's suggestion is very reasonable to me.

What is most important in cases like this is to have:
1. Consistent and agreed upon conventions of referring to modules.  This should be documented somewhere so the convention is clear and that the issue is not revisited.
2. No arbitrary changing of convention without discussion so that code can remain consistent
3. Small to the point PRs that are as close to "one change" as possible so that PRs do not feel like an overhead.

mosra: > This should be documented somewhere

:+1: Someone should be taking notes on what to populate a Coding Style document with :)

> I would advocate against making namespace alias changes as part of this PR.

Definitely. Even if you (or anybody else as a PR author) don't like the conventions used, changing them only adds extra lines to the diff, extra overhead for each reviewer.

If done separately in a dedicated PR instead, this change could be simply a global replace of `Magnum::` with `Mn::`, `magnum.` with `mn.` and then adding `namespace Mn = Magnum` or `import magnum as mn` (and removing `using namespace Magnum;` if it's still somewhere) until everything compiles again and tests pass. Then, if it's really just that, with no other changes requiring special attention, such a PR would be extremely easy to review.

angelxuanchang: Do we have an existing coding style document?  I can help track of what we need to put into the coding style document.

msavva: Indeed we do, will share with all in the dev slack channel -- thanks @angelxuanchang !

msbaines: I like the namespace alias. Currently, it's a bit confusing in that it is not followed everywhere. Once we resolve that, I think it will become second nature.

bigbike:
Let me know if you have more comments about this PR before I merge it. Thanks.

bigbike:
Sincerely thank all the reviwers' work. Merge it.

PR #198, opened 2019-09-07
Support ptex mesh - step 4: load and set atlas textures correctly

bigbike:
## Motivation and Context
This PR is to load and set the atlas textures (.hdr files) in correct format using magnum;

Must-have for rendering pets meshes

## How Has This Been Tested
local (see #132)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/PTexMeshData.cpp
+++ b/src/esp/assets/PTexMeshData.cpp
@@ -12,8 +12,10 @@
 #include <Corrade/Containers/Array.h>
 #include <Corrade/Containers/ArrayView.h>
 #include <Corrade/Containers/ArrayViewStl.h>
+#include <Corrade/Utility/Assert.h>
 #include <Corrade/Utility/Directory.h>
 #include <Magnum/GL/BufferTextureFormat.h>
+#include <Magnum/GL/TextureFormat.h>
 #include <Magnum/ImageView.h>
 #include <Magnum/PixelFormat.h>
 
@@ -614,28 +616,40 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
                         Magnum::GL::MeshIndexType::UnsignedInt);
   }
 
+  // load atlas data and upload them to GPU
+  LOG(INFO) << "loading atlas textures: ";
   for (size_t iMesh = 0; iMesh < renderingBuffers_.size(); ++iMesh) {
-    const std::string rgbFile =
-        atlasFolder_ + "/" + std::to_string(iMesh) + "-color-ptex.rgb";
-    if (!io::exists(rgbFile)) {
-      ASSERT(false, "Can't find " + rgbFile);
-    }
-    LOG(INFO) << "\rLoading atlas " << iMesh + 1 << "/"
-              << renderingBuffers_.size() << "... ";
-    LOG(INFO).flush();
+    const std::string hdrFile = Corrade::Utility::Directory::join(
+        atlasFolder_, std::to_string(iMesh) + "-color-ptex.hdr");
 
-    Cr::Containers::Array<const char, Cr::Utility::Directory::MapDeleter> data =
-        Cr::Utility::Directory::mapRead(rgbFile);
-    const int dim = static_cast<int>(std::sqrt(data.size() / 3));  // square
-    Magnum::ImageView2D image(Magnum::PixelFormat::RGB8UI, {dim, dim}, data);
+    CORRADE_ASSERT(io::exists(hdrFile), "Error : Cannot find the .hdr file ", );
+
+    LOG(INFO) << "Loading atlas " << iMesh + 1 << "/"
+              << renderingBuffers_.size() << " from " << hdrFile << ". ";
+
+    Corrade::Containers::Array<const char,
+                               Corrade::Utility::Directory::MapDeleter>
+        data = Corrade::Utility::Directory::mapRead(hdrFile);
+    // divided by 6, since there are 3 channels, R, G, B, each of which takes 1
+    // half_float (2 bytes)
+    const int dim = static_cast<int>(std::sqrt(data.size() / 6));  // square
+
+    // atlas
+    // the size of each image is dim x dim x 3 (RGB) x 2 (half_float), which
+    // equals to numBytes
+    Magnum::ImageView2D image(Magnum::PixelFormat::RGB16F, {dim, dim}, data);
+    const int mipLevelCount = Magnum::Math::log2(image.size().min()) + 1;
+    const int mipLevel = 0;
+
     renderingBuffers_[iMesh]
         ->tex.setWrapping(Magnum::GL::SamplerWrapping::ClampToEdge)
         .setMagnificationFilter(Magnum::GL::SamplerFilter::Linear)
         .setMinificationFilter(Magnum::GL::SamplerFilter::Linear)
-        // .setStorage(1, GL::TextureFormat::RGB8UI, image.size())
-        .setSubImage(0, {}, image);
+        .setStorage(mipLevelCount, Magnum::GL::TextureFormat::RGB16F,
+                    image.size())
+        .setSubImage(mipLevel, {}, image)
+        .generateMipmap();
   }
-  LOG(INFO) << "... done" << std::endl;
 
   buffersOnGPU_ = true;
 }


bigbike:

Thanks.

bigbike:
> LGTM! 👍 Should be good to land after the Corrade namespace alias is reintroduced.

Thanks for reviewing it. 
I have different opinions about the abbreviation.
See my comments in other PR.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -614,28 +616,40 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+    // divided by 6, since there are 3 channels, R, G, B, each of which takes 1
+    // half_float (2 bytes)
+    const int dim = static_cast<int>(std::sqrt(data.size() / 6));  // square
+
+    // atlas
+    // the size of each image is dim x dim x 3 (RGB) x 2 (half_float), which
+    // equals to numBytes
+    Magnum::ImageView2D image(Magnum::PixelFormat::RGB16F, {dim, dim}, data);
+    const int mipLevelCount = Magnum::Math::log2(image.size().min()) + 1;
+    const int mipLevel = 0;
```
mosra: Do we really need to have the mip level in a dedicated variable? :) That's what you usually do if you don't have pre-calculated mips, upload just the first one and let `generateMipmap()` populate the others. Extending this style further, it would need a `const Vector2i offset = {}` as well. I don't think either is necessary.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -614,28 +616,40 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+        atlasFolder_, std::to_string(iMesh) + "-color-ptex.hdr");
+
+    CORRADE_ASSERT(io::exists(hdrFile), "Error : Cannot find the .hdr file ", );
+
+    LOG(INFO) << "Loading atlas " << iMesh + 1 << "/"
+              << renderingBuffers_.size() << " from " << hdrFile << ". ";
+
+    Corrade::Containers::Array<const char,
+                               Corrade::Utility::Directory::MapDeleter>
+        data = Corrade::Utility::Directory::mapRead(hdrFile);
```
mosra: I wonder, does Replica plan to supply the textures in some real file format (with size metadata, compression and such)? With half-floats it could easily be an EXR. Storing (and loading) the data uncompressed is quite wasteful IMHO and this guesswork with calculating image size is error-prone.

bigbike: I see. To answer this question, we will have to ask our colleagues at FRL who wrote the replicaSDK. 
This code is simply following that.

src/esp/assets/PTexMeshData.cpp
bigbike: Yes, the reason is to improve the code readability, and also make the program less error-prone. I learned it these years when I developed the infra at FB product team.

Ideally code should be self-explanatory. '0' and `Magnum::Math::log2(image.size().min()) + 1` are somehow magic number and expression here without knowing the actual meaning unless reading the header files (We are fortunate enough as we can just read your docs instead of going deeply into Magnum to find the .h file). You are the author of the Magnum, so perhaps you do not see this as a problem. Even worse, a moment later, you may forget, and you will have to check it again.

When I start the project, the application simply gave me a "black screen" without rendering anything. So every function was suspicious at that moment. I had to check each parameter passed to the function, see if we pass the correct one, and pass it properly  (that was how I found the issue about the number we passed to the setCount() was wrong, which caused the segfault. Remember?) This process is quite painful. Using the "redundant" variables here, it eases the pain a bit since I only check it once, and will not forget it when I look it again later. 

Also, it prevents errors sometimes. Considering a function, 'Foo(int a, int b)', the compiler will not rise any complain if anyone accidentally passes "b" as "a" and "a" as "b" when calling it. It is not necessary the case here, but the logic is the same.

And yes, ideally it would need a `const Vector2i offset = {}`; I admit I made a tradeoff here. I can add it if we agree on it.

src/esp/assets/PTexMeshData.cpp
mosra: Sure -- I'm not saying this code is to blame ;) It just felt strange to me, so I figured you would know the reason why they did it like this.

src/esp/assets/PTexMeshData.cpp
mosra: I'm trying hard to avoid confusing APIs like `setFoo(0, {}, -3, true, false, 74)` and I think the `setSubImage()` is still acceptable :) Too bad C++ doesn't have a possibility to use named parameters like Python, then it could be crystal clear:

```cpp
.setSubImage(mipLevel=0, offset={}, image)
```

As a middle ground, if you feel this is really a problem, what about this instead? Similarly to what you did in #200 yesterday:

```cpp
.setSubImage(
  0,  // mipLevel
  {}, // offset
  image)
```

bigbike:
Let me know if you have more comments about this PR before I merge it. Thanks.

src/esp/assets/PTexMeshData.cpp
msavva: Agreed with @mosra 's suggestion above, it will make parameter interpretation clear while not selectively introducing additional local variables.

bigbike: By simply comparing to the "named parameters" in python, what I implemented here is closer than what you suggested.

"Self-explanatory" means code is well understood even without the comments. This is exactly what I did here, and in my humble opinion, it is better than the middle ground solution. 

I do not think it make the code unnecessarily complicated. In fact, to keep it simple,  I compromised as I did not replace the `{}` by another variable, `offset`.

Performance lost? Apparently not. Even if it has, it is totally negligible.

I really do not see any good reason to give it up. But if it make you more comfortable, I am actually going to change it. Not a big deal to me, though.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -616,29 +618,41 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
-    if (!io::exists(rgbFile)) {
-      ASSERT(false, "Can't find " + rgbFile);
-    }
-    LOG(INFO) << "\rLoading atlas " << iMesh + 1 << "/"
-              << renderingBuffers_.size() << "... ";
-    LOG(INFO).flush();
+    const std::string hdrFile = Cr::Utility::Directory::join(
+        atlasFolder_, std::to_string(iMesh) + "-color-ptex.hdr");
+
+    CORRADE_ASSERT(io::exists(hdrFile), "Error : Cannot find the .hdr file ", );
```
bigbike: @mosra :
For the 2nd parameter, I cannot do something like, "Error : Cannot find the .hdr file " << hdrFile as the compiler returns large number of errors.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -616,29 +618,41 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+              << renderingBuffers_.size() << " from " << hdrFile << ". ";
 
     Cr::Containers::Array<const char, Cr::Utility::Directory::MapDeleter> data =
-        Cr::Utility::Directory::mapRead(rgbFile);
-    const int dim = static_cast<int>(std::sqrt(data.size() / 3));  // square
+        Cr::Utility::Directory::mapRead(hdrFile);
+    // divided by 6, since there are 3 channels, R, G, B, each of which takes 1
+    // half_float (2 bytes)
+    const int dim = static_cast<int>(std::sqrt(data.size() / 6));  // square
+
```
msbaines: Maybe add a debug assertion that:

dim*dim*6 = data.size()

bigbike: Done.

bigbike:
Let me know if you have further suggestions. Thanks a lot.

src/esp/assets/PTexMeshData.cpp
mosra: Does it complain also if you `#include <Corrade/Utility/DebugStl.h>`?

bigbike: I did include a header file, but maybe not this one. Let me try. Thanks.

Issue #121, opened 2019-07-27
Proposal: replace glog with Corrade::Utility::Debug


mosra:
I'm seeing a lot of code like

```cpp
LOG(INFO) << "User provided object inertia " 
          << bInertia.x() << " " << bInertia.y() << " " << bInertia.z();
```

or a `switch` that looks like 

```cpp
      case Magnum::MeshPrimitive::Lines:
        LOG(ERROR) << "Invalid primitive: Lines";
        break;
      case Magnum::MeshPrimitive::Points:
        LOG(ERROR) << "Invalid primitive: Points";
        break;
      case Magnum::MeshPrimitive::LineLoop:
        LOG(ERROR) << "Invalid primitive Line loop";
        break;
      ...
```

It's making me sad, because it wastes programmer time -- logging / printf debugging should not involve any mental overhead of "how the heck do I print this variable". The [Utility::Debug](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html) class in Magnum can do the same as just:

```cpp
Debug{} << "User provided object inertia" << bInertia;
```

and it can also print *any* container, almost all Magnum enum values (as names, not numbers, of course), it has output redirection (and muting), debug/warning/error levels, colored output if you want to get fancy and it can [print images](https://blog.magnum.graphics/backstage/multidimensional-strided-array-views/#enter-strided-array-views) as well.

I never used glog, but looking at the project page there doesn't even seem to be any real docs for it, apart from a [raw HTML file](https://github.com/google/glog/blob/master/doc/glog.html). How well maintained that project is? For the WebGL build in #53 we'd need to use something else anyway, so I would propose switching away from it completely.

My question is -- is there any glog feature that `Utility::Debug` doesn't have and you need it? One thing I'm aware of is that it doesn't prefix the output with file/line by default, but there's a (not yet merged) proposal that adds it as an opt-in feature on supported compilers: https://github.com/mosra/corrade/pull/55

msavva:
Thanks @mosra for making this proposal!  It will certainly make things easier for us down the line if we can leverage logging from Magnum (and get free WebGL build support).  Overall, it does seem that switching from `glog` to `Utility::Debug` will bring benefits without giving up too much.

[...]

PR #200, opened 2019-09-09
Add an eigen format to control the way vectors are printed.

bigbike:
## Motivation and Context
As titled.
This PR improves the format of the output vectors, and makes it a bit easier to read.

Before this change:
image

After this change:
image

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/Viewer.cpp
+++ b/src/esp/gfx/Viewer.cpp
@@ -347,28 +347,38 @@ void Viewer::keyPressEvent(KeyEvent& event) {
     case KeyEvent::Key::A:
       controls_(*agentBodyNode_, "moveLeft", moveSensitivity);
       LOG(INFO) << "Agent position "
-                << Eigen::Map<vec3f>(agentBodyNode_->translation().data());
+                << Eigen::Map<vec3f>(agentBodyNode_->translation().data())
+                       .format(eigenFlatFormat_);
       break;
     case KeyEvent::Key::D:
       controls_(*agentBodyNode_, "moveRight", moveSensitivity);
       LOG(INFO) << "Agent position "
-                << Eigen::Map<vec3f>(agentBodyNode_->translation().data());
+                << Eigen::Map<vec3f>(agentBodyNode_->translation().data())
+                       .format(eigenFlatFormat_);
       break;
     case KeyEvent::Key::S:
       controls_(*agentBodyNode_, "moveBackward", moveSensitivity);
       LOG(INFO) << "Agent position "
-                << Eigen::Map<vec3f>(agentBodyNode_->translation().data());
+                << Eigen::Map<vec3f>(agentBodyNode_->translation().data())
+                       .format(eigenFlatFormat_);
       break;
     case KeyEvent::Key::W:
       controls_(*agentBodyNode_, "moveForward", moveSensitivity);
       LOG(INFO) << "Agent position "
-                << Eigen::Map<vec3f>(agentBodyNode_->translation().data());
+                << Eigen::Map<vec3f>(agentBodyNode_->translation().data())
+                       .format(eigenFlatFormat_);
       break;
     case KeyEvent::Key::X:
       controls_(*agentBodyNode_, "moveDown", moveSensitivity, false);
+      LOG(INFO) << "Agent position "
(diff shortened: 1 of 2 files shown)


bigbike:
Thanks.

src/esp/gfx/Viewer.h
```diff
@@ -84,6 +84,15 @@ class Viewer : public Magnum::Platform::Application {
...
   Magnum::Timeline timeline_;
+
+  const Eigen::IOFormat eigenFlatFormat_{Eigen::StreamPrecision,
+                                         Eigen::DontAlignCols,
+                                         ", ",  // coef separator
+                                         ", ",  // row separator
+                                         "",    // row prefix
+                                         "",    // col prefix
+                                         "[",   // mat prefix
+                                         "]"};  // mat suffix
```
mosra: I encountered [this in core/esp.h](https://github.com/facebookresearch/habitat-sim/blob/1d4bd7bc8c73f6f4674e6e2c553930eff1a0d606/src/esp/core/esp.h#L32-L41) when going over the code few months ago and assumed all code was using it already:

```cpp
//! Eigen JSON string format specification
static const IOFormat
    kJsonFormat(StreamPrecision, DontAlignCols, ",", ",", "", "", "[", "]");

//! Write Eigen matrix types into ostream in JSON string format
template <typename T, int numRows, int numCols>
std::ostream& operator<<(std::ostream& os,
                         const Matrix<T, numRows, numCols>& matrix) {
  return os << matrix.format(kJsonFormat);
}
```

For the things printed in viewer, are those not `Matrix` instances and thus it doesn't work? Are those `Eigen::Array`s? Or some `Eigen::Map<Something, Something>`? What about adding another `operator<<` overload there instead so you don't need to specify `.format()` every time?

src/esp/gfx/Viewer.cpp
```diff
@@ -347,28 +347,38 @@ void Viewer::keyPressEvent(KeyEvent& event) {
...
       controls_(*agentBodyNode_, "moveDown", moveSensitivity, false);
+      LOG(INFO) << "Agent position "
+                << Eigen::Map<vec3f>(agentBodyNode_->translation().data())
+                       .format(eigenFlatFormat_);
       break;
     case KeyEvent::Key::Z:
       controls_(*agentBodyNode_, "moveUp", moveSensitivity, false);
+      LOG(INFO) << "Agent position "
+                << Eigen::Map<vec3f>(agentBodyNode_->translation().data())
+                       .format(eigenFlatFormat_);
```
mosra: The `translation()` is a `Magnum::Vector3`, right? I think this could be another TODO for #121 :)

src/esp/gfx/Viewer.h
msavva: Thanks for catching this @mosra -- indeed, an `operator<<` overload in core/esp.h ought to be able to handle formatting without the need to explicitly call `format()` every time.

bigbike: Good to know the similar format was actually defined in the esp.h.
Will take a look, and see if it can be used. Thanks.

src/esp/core/esp.h
```diff
@@ -30,8 +30,14 @@ typedef Matrix<uint32_t, 4, 1> Vector4ui;
 typedef Matrix<uint64_t, 4, 1> Vector4ul;
 
 //! Eigen JSON string format specification
-static const IOFormat
-    kJsonFormat(StreamPrecision, DontAlignCols, ",", ",", "", "", "[", "]");
+static const IOFormat kJsonFormat(StreamPrecision,
+                                  DontAlignCols,
+                                  ",",   // coef separator
```
bigbike:
```suggestion
                                  ", ",   // coef separator
```
Can I add a "space" behind the comma for the coef, and row separator? It looks better in my opinion.

The following is the one WITHOUT the space:
image

And the one with the space:
image

bigbike: Please let me know your thoughts. Thanks.

msavva: Looks good to me.

bigbike:
Can I add a "space" behind the comma for the coef, and row separator?

PR #202, opened 2019-09-10
Speedup incremental build of WebGL

msbaines:
Currently an incremental build requires an install which requires
all targets be built. By avoid an install, I can avoid building
the viewer.js target, cutting my incremental build time in half.

## Motivation and Context

Currently an incremental build requires an install which requires
all targets be built. By avoid an install, I can avoid building
the viewer.js target, cutting my incremental build time in half.

## How Has This Been Tested

Build.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/CMakeLists.txt
+++ b/src/esp/bindings_js/CMakeLists.txt
@@ -14,12 +14,21 @@ target_link_libraries(hsim_bindings
 
 set_target_properties(hsim_bindings PROPERTIES LINK_FLAGS "--bind")
 
-# copy JS/HTML/CSS scaffolding for WebGL build
-install(FILES
-  bindings.html
-  bindings.css
-  navigate.js
-  simenv_embind.js
-  ${MAGNUM_WINDOWLESSEMSCRIPTENAPPLICATION_JS}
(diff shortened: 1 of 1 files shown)


src/esp/bindings_js/CMakeLists.txt
```diff
@@ -14,12 +14,21 @@ target_link_libraries(hsim_bindings
...
-  simenv_embind.js
-  ${MAGNUM_WINDOWLESSEMSCRIPTENAPPLICATION_JS}
-  ${MAGNUM_WEBAPPLICATION_CSS}
-  DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+# copy JS/HTML/CSS resources for WebGL build
+set(resources
+    bindings.html
+    bindings.css
+    navigate.js
+    simenv_embind.js
```
apsdehal: Can we use a glob pattern here, otherwise this will need to be updated every time we add a new js or html file?

erikwijmans: We recently removed using file globs: https://github.com/facebookresearch/habitat-sim/pull/158

PR #203, opened 2019-09-11
Support ptex mesh - step 5: improve the shader and rendering code

bigbike:
## Motivation and Context
As titled.
-) added illumination parameters, gamma, saturation etc.;
-) refactored the PTexMeshShader a bit, moving function implementations to .cpp file;
-) renamed a couple of functions in PTexMeshShader;
-) cached the uniform positions;

## How Has This Been Tested
local built

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/PTexMeshData.h
+++ b/src/esp/assets/PTexMeshData.h
@@ -63,13 +63,13 @@ class PTexMeshData : public BaseMesh {
   virtual Magnum::GL::Mesh* getMagnumGLMesh(int submeshID) override;
 
   float exposure() const;
-  void setExposure(const float& val);
+  void setExposure(float val);
 
   float gamma() const;
-  void setGamma(const float& val);
+  void setGamma(float val);
 
   float saturation() const;
-  void setSaturation(const float& val);
+  void setSaturation(float val);
 
  protected:
   void loadMeshData(const std::string& meshFile);
@@ -77,9 +77,14 @@ class PTexMeshData : public BaseMesh {
   float splitSize_ = 0.0f;
   uint32_t tileSize_ = 0;
 
-  // based on ReplicaSDK
+  // initial values are based on ReplicaSDK
+  //! @brief expsure, the amount of light per unit area reaching the image
   float exposure_ = 0.025f;
-  float gamma_ = 1.6969f;
+
+  //! @brief gamma, the exponent applied in the gamma correction
+  float gamma_ = 1.0f / 1.6969f;
+
+  //! @brief saturation, the intensity of a color
   float saturation_ = 1.5f;
 
   std::string atlasFolder_;
--- a/src/esp/assets/PTexMeshData.h
+++ b/src/esp/assets/PTexMeshData.h
@@ -78,7 +78,7 @@ class PTexMeshData : public BaseMesh {
   uint32_t tileSize_ = 0;
 
   // initial values are based on ReplicaSDK
-  //! @brief expsure, the amount of light per unit area reaching the image
+  //! @brief exposure, the amount of light per unit area reaching the image
   float exposure_ = 0.025f;
 
   //! @brief gamma, the exponent applied in the gamma correction
--- a/src/esp/assets/PTexMeshData.h
+++ b/src/esp/assets/PTexMeshData.h
@@ -45,8 +45,6 @@ class PTexMeshData : public BaseMesh {
 
   // ==== geometry ====
   void load(const std::string& meshFile, const std::string& atlasFolder);
-  float exposure() const;
-  void setExposure(const float& val);
   uint32_t tileSize() const { return tileSize_; }
 
   const std::vector<MeshData>& meshes() const;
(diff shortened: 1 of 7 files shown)


bigbike:
This diff is ready to be reviewed. Thanks.

src/esp/assets/PTexMeshData.h
```diff
@@ -64,12 +62,26 @@ class PTexMeshData : public BaseMesh {
...
   virtual Magnum::GL::Mesh* getMagnumGLMesh(int submeshID) override;
 
+  float exposure() const;
+  void setExposure(const float& val);
+
+  float gamma() const;
+  void setGamma(const float& val);
+
+  float saturation() const;
+  void setSaturation(const float& val);
```
mosra: It would be great to add docstrings here, if nothing else at least a note about the inversion of gamma that's done inside.

Or to the shader. Or both.

src/esp/gfx/PTexMeshShader.cpp
```diff
@@ -87,5 +90,38 @@ PTexMeshShader& PTexMeshShader::bindAdjFacesBufferTexture(
...
+
+PTexMeshShader& PTexMeshShader::setAtlasTextureSize(
+    Magnum::GL::Texture2D& texture,
+    uint32_t tileSize) {
+  setUniform(tileSizeUniform_, (int)tileSize);
+
+  // Image size in given mip level 0
+  int mipLevel = 0;
+  int widthEntry = 0;
+  const auto width = texture.imageSize(mipLevel)[widthEntry];
```
mosra: I get that this code was already there and you just moved it, but what about `texture.imageSize(mipLevel).x()`?

src/esp/assets/PTexMeshData.h
```diff
@@ -64,12 +62,26 @@ class PTexMeshData : public BaseMesh {
   virtual void uploadBuffersToGPU(bool forceReload = false) override;
   virtual Magnum::GL::Mesh* getMagnumGLMesh(int submeshID) override;
 
+  float exposure() const;
+  void setExposure(const float& val);
```
mosra: Just `float val`, maybe? The reference doesn't save anything (is actually slower than passing by-value in this case).

Same in the two cases below.

src/esp/assets/PTexMeshData.h
bigbike: OK. I am working on it.

src/esp/gfx/PTexMeshShader.cpp
bigbike: Good suggestion.

src/esp/assets/PTexMeshData.h
bigbike: Definitely you are right. Good catch!

src/esp/gfx/PTexMeshShader.h
```diff
@@ -23,19 +23,51 @@ namespace gfx {
...
+   *! @brief Bind the buffer texture containing the adjacent faces
+   *! @return Reference to self (for method chaining)
+   */
   PTexMeshShader& bindAdjFacesBufferTexture(Magnum::GL::BufferTexture& texture);
 
   // ======== set uniforms ===========
+  /**
+   *! @brief Set modelview and projection matrix to the uniform on GPU
+   *! @return Reference to self (for method chaining)
+   */
```
mosra: Can we maybe decide on a consistent docstring style? :) It's `//!` on some places, #195 uses `/**` and here it's `*!`. (Actually, does Doxygen even understand these without putting exclamation marks to the output? I know only of `/*!`, but not `*!`).

Cc: @aclegg3 @msavva @erikwijmans

bigbike: @mosra :
Are you commenting on the latest version? I did not see exclamation marks on my side (no new version was pushed after your comments).
The docstrings in this diff was following the format in the PrimitiveIDShader.h

bigbike: image

mosra: Uh oh, sorry, GitHub is acting up and showing it like this in the conversation here

image

but when I open the file diff, those are not there anymore. Not sure what's up. Apologies for the noise.

bigbike: @mosra : NO worries! Sometimes GitHub behaves a bit abnormal. I totally understand since I vaguely remembered I had similar experience before.

bigbike:
Please let me know your concerns on this PR. Thanks!

bigbike:
@mosra : do you have more comments on this PR?

src/esp/gfx/PTexMeshShader.cpp
```diff
@@ -87,5 +90,36 @@ PTexMeshShader& PTexMeshShader::bindAdjFacesBufferTexture(
...
+}
+
+PTexMeshShader& PTexMeshShader::setAtlasTextureSize(
+    Magnum::GL::Texture2D& texture,
+    uint32_t tileSize) {
+  setUniform(tileSizeUniform_, (int)tileSize);
+
+  // get image width in given mip level 0
+  int mipLevel = 0;
+  const auto width = texture.imageSize(mipLevel).x();
```
mosra: Since the rendering uses geometry shaders, I assume it's not planned to be running on WebGL, right? (If it would be, the `imageSize()` would be problematic, but otherwise not at all.)

bigbike: @msbaines: May I ask you what our latest plan is on rendering the ptex mesh on WebGL?

bigbike: @mosra : Yeah, the geometry shader is a problem as it is not available in WebGL.

bigbike:
I will merge it to master to unblock myself this afternoon.

src/esp/gfx/PTexMeshShader.cpp
msbaines: I am currently planning on using the vertex color information from the ply mesh but at some point would like to migrate to using the ptex textures. Not sure when. It will depend on whether the vertex color-only is OK. The other issue is that the textures are 100s of MB each so may not be feasible for WebGL.

mosra: Texture size could be fixed by using Basis texture compression (see https://github.com/mosra/magnum-plugins/pull/62, I'm in process of integrating that), ideally the compressed textures being upstreamed back to Replica.

For vertex colors, when I checked how a PLY file looks compared to the textured GLB, it was quite underwhelming :) For human eyes and a 4K screen at least, when rendering tiny images the difference would probably be minimal.

PR #204, opened 2019-09-11
Registry

erikwijmans:
## Motivation and Context

With PR #189 adding another thing that makes sense to also use a registry for, it is time to swap to a proper registry pattern!

This registry is slightly different than the one in habitat-api as it's just a class instead of a global singleton instance. The global singleton instance was hard to generate proper docs for while generating docs for a class was super simple.  I don't believe there is any functional difference between the two.

## How Has This Been Tested

Via the existing tests.  

Note that the CI is expected to fail for this PR as habitat-api has an example which uses habitat-sim's registry that will need to be updated.

## Types of changes

- Breaking change (fix or feature that would cause existing functionality to change)


docs/pages/new-actions.rst
```diff
@@ -20,8 +20,8 @@ actuation_spec contains any parameters needed by that control function.
 See ``habitat_sim/agent/controls/default_controls.py`` for more example
 controls.
 
-Controls are registered using the `habitat_sim.agent.register_move_fn()`
-function. This function takes the functor to register and, optionally, the name
+Controls are registered using `habitat_sim.registry.register_move_fn`.
```
mosra: Oh, I was always appending the `()` to function references (to distinguish them from variables/properties/modules)... and I did that for the soon-to-be-PR'd habitat-api docs rework as well. Please let me know if I should stop doing that :)

habitat_sim/registry.py
```diff
@@ -0,0 +1,92 @@
...
+def _camel_to_snake(name):
+    s1 = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", name)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", s1).lower()
+
+
+class registry:
+    r"""registry is a central source of truth in Habitat-Sim
+
+    Taken from Pythia, it is inspired from Redux's
+    concept of global store. registry maintains mappings of various information
```
mosra: > This registry is slightly different than the one in habitat-api

AH, that's why this docstring felt like a deja vu to me :)

erikwijmans: Yep :)

docs/pages/new-actions.rst
erikwijmans: I like that!  I didn't realize that's why you had done that so it didn't register that I should keep it.

PR #211, opened 2019-09-13
Physics tests: kinematics

aclegg3:
## Motivation and Context
New circleCI tests for physics functions introduced in PR #91  

CI download test object dataset

Test initialization of default PhysicsManager and related code path for "enable_physics" flag.

Test object loading, instancing, removal, kinematics functions, etc...

## How Has This Been Tested

pytest circleCI

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Testing

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #213, opened 2019-09-16
Physics tests: Install and run Habitat-sim with Bullet physics

aclegg3:
## Motivation and Context

Test installation of Bullet physics and habitat-sim with Bullet physics in circleCI.
Test python exposed dynamics with various physics implementations (only Bullet currently).

Leaving TODOs for later PR additions. We have no caching of dynamic state currently so I'm not querying or asserting anything about that yet. This PR is primarily to test bullet install, build with bullet, and run the simulator with dynamics.

## How Has This Been Tested

Local (MacOS) and devfair (Ubuntu) pytest.

## Types of changes

- [x] Testing
## Checklist


src/esp/gfx/Simulator.h
```diff
@@ -88,7 +88,7 @@ class Simulator {
   int getPhysicsObjectLibrarySize();
 
   // remove object objectID instance in sceneID
-  void removeObject(const int objectID, const int sceneID = 0);
+  int removeObject(const int objectID, const int sceneID = 0);
```
mosra: Maybe document what the return value is? It wasn't immediately clear to me :)

aclegg3: Good point, I'll do a documentation pass over some of these Simulator functions once the first physics doc PR is merged and link to relevant functions.

aclegg3: Addressed with #211 as you commented there :)

aclegg3:
I'd like to get this test (especially the build part) into master before the release. I'll write a separate PR to add assertions and additional tests once we have force/torque caching and more interesting time-stepping schemes.

.circleci/config.yml
```diff
@@ -237,6 +238,11 @@ jobs:
 
               while [ ! -f ~/miniconda/pytorch_installed ]; do sleep 2; done # wait for Pytorch
               pytest
+
+              #re-install with bullet and run physics tests again
+              rm -r build
```
erikwijmans: I don't believe this is needed

mosra: Yeah -- if you don't `rm` it, the dependencies wouldn't need to be rebuilt :)

tests/test_physics.py
```diff
@@ -73,3 +73,52 @@ def test_kinematics(sim):
...
+        # TODO: assertion?
+
+        # TODO: impulses
+
+        # test getting observation
+        obs = sim.step(random.choice(list(hab_cfg.agents[0].action_space.keys())))
+
+        # check that time is increasing in the world
+        assert sim.get_world_time() > prev_time
+        prev_time = sim.get_world_time()
```
mosra: Random idea on how to test this -- each rendered frame should be quite different (objects moving around). Is it easy to verify?

.circleci/config.yml
erikwijmans: Fortunately, ccache to the rescue, but still unneeded.

tests/test_physics.py
aclegg3: Yeah, that sounds like one way to test this. Remove the set_translation, make the scene deterministic, spawn the objects within camera view, and then check that they are moving. 

We could also check the state of the objects for change, or try to setup a physical scenario such as an anti-gravity force or something like this.

aclegg3: >...try to setup a physical scenario such as an anti-gravity force...
 
I implemented this option with assertions and a deterministic scene.

.circleci/config.yml
```diff
@@ -254,6 +255,10 @@ jobs:
 
               while [ ! -f ~/miniconda/pytorch_installed ]; do sleep 2; done # wait for Pytorch
               pytest
+
+              #re-install with bullet and run physics tests again
+              python setup.py install --headless --bullet
```
erikwijmans: This should actually be `./build.sh` as the tests will only pickup in-tree bindings and don't work with the setup.py install method (grumble grumble grumble)

aclegg3: Odd that the tests were passing after the rebuild, then. Do you think they were picking up the previous build?

.circleci/config.yml
```diff
@@ -254,6 +255,10 @@ jobs:
 
               while [ ! -f ~/miniconda/pytorch_installed ]; do sleep 2; done # wait for Pytorch
               pytest
+
+              #re-install with bullet and run physics tests again
+              ./build.sh --headless --bullet
```
mathfac: Maybe we can do `./build.sh --headless --bullet` by default as it will do more test coverage. And do testing without bullet as secondary testing for compatibility.

.circleci/config.yml
erikwijmans: Yeah, they were picking up things from the previous invocation of `build.sh`

aclegg3: :expressionless:

PR #214, opened 2019-09-16
Blit to defaultFrameBuffer to directly render in EmScripten Build

apsdehal:
## Motivation and Context

In our Web demo, we are currently taking observation from Simulator and then pasting it on 2d canvas which is very slow. In this change, we directly blit the offscreen render target's RGBA buffer to default frame buffer which will render in the browser as canvas. This also removes custom rendering code on frontend side and relies totally on simulator now to provide that information.

## How Has This Been Tested

./build_js.sh and then manually testing in browser

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/bindings.html
+++ b/src/esp/bindings_js/bindings.html
@@ -15,8 +15,7 @@
     <div id="sizer">
       <div id="expander">
         <div id="listener">
-          <canvas class="hidden" id="canvas"></canvas>
-          <canvas id="canvas2d" width="640" height="480" style="transform: scaleY(-1)"></canvas>
+          <canvas id="canvas" width="640" height="480"></canvas>
           <canvas id="semantic" width="640" height="480" style="transform: scaleY(-1)"></canvas>
           <canvas id="radar"></canvas>
           <pre id="log"></pre>
@@ -86,7 +85,7 @@
       const simenv = new SimEnv(config, episode, 0);
       const agent = simenv.addAgent(agentConfig);
       const task = new NavigateTask(simenv, {
-        canvas: document.getElementById('canvas2d'),
+        canvas: document.getElementById('canvas'),
         semantic: document.getElementById('semantic'),
         radar: document.getElementById('radar'),
         status: document.getElementById('status')
--- a/src/esp/bindings_js/navigate.js
+++ b/src/esp/bindings_js/navigate.js
@@ -56,9 +56,7 @@ class NavigateTask {
   }
 
   renderImage() {
-    // Not using this as of now, but useful for later tracking purposes
-    // Also, required to render everything on sim side
-    this.lastObs = this.sim.getObservation("rgb", null);
+    this.sim.displayObservation("rgb");
     this.renderRadar();
   }
 
--- a/src/esp/bindings_js/navigate.js
+++ b/src/esp/bindings_js/navigate.js
@@ -55,12 +55,6 @@ class NavigateTask {
     this.components.status.innerHTML = text;
   }
 
-  applyGamma(data, gamma) {
-    for (let i = 0; i < data.length; i++) {
-      data[i] = Math.pow(data[i]/255.0, gamma) * 255;
-    }
-  }
-
   renderImage() {
     // Not using this as of now, but useful for later tracking purposes
     // Also, required to render everything on sim side
@@ -69,6 +63,10 @@ class NavigateTask {
   }
 
   renderSemanticImage() {
+    if (this.semanticObjects.size() == 0) {
+      return;
+    }
+
     const obs = this.sim.getObservation("semantic", null);
     this.semantic_data = obs.getData();
     let data = this.semantic_data;
@@ -76,19 +74,19 @@ class NavigateTask {
     // TOOD(msb) implement a better colorization scheme
     for (let i = 0; i < 640*480; i++) {
       if (data[i*4] & 1) {
-	this.semanticImageData.data[i*4] = 255;
+        this.semanticImageData.data[i*4] = 255;
       } else {
-	this.semanticImageData.data[i*4] = 0;
+        this.semanticImageData.data[i*4] = 0;
       }
       if (data[i*4] & 2) {
-	this.semanticImageData.data[i*4+1] = 255;
+        this.semanticImageData.data[i*4+1] = 255;
       } else {
-	this.semanticImageData.data[i*4+1] = 0;
+        this.semanticImageData.data[i*4+1] = 0;
       }
       if (data[i*4] & 4) {
-	this.semanticImageData.data[i*4+2] = 255;
+        this.semanticImageData.data[i*4+2] = 255;
       } else {
-	this.semanticImageData.data[i*4+2] = 0;
+        this.semanticImageData.data[i*4+2] = 0;
       }
       this.semanticImageData.data[i*4 + 3] = 255;
     }
@@ -131,9 +129,7 @@ class NavigateTask {
 
   render() {
     this.renderImage();
-    if (this.semanticObjects.size() > 0) {
-      this.renderSemanticImage();
-    }
+    this.renderSemanticImage();
   }
 
   handleAction(action) {
--- a/src/esp/bindings_js/navigate.js
+++ b/src/esp/bindings_js/navigate.js
@@ -12,9 +12,8 @@ class NavigateTask {
   constructor(sim, components) {
     this.sim = sim;
     this.components = components;
-    this.imageCtx = components.canvas.getContext("2d");
+    this.lastObs = null;
     let shape = this.sim.getObservationSpace("rgb").shape;
-    this.imageData = this.imageCtx.createImageData(shape.get(1), shape.get(0));
     this.semanticCtx = components.semantic.getContext("2d");
     shape = this.sim.getObservationSpace("semantic").shape;
     this.semanticImageData = this.semanticCtx.createImageData(shape.get(1), shape.get(0));
@@ -63,11 +62,9 @@ class NavigateTask {
   }
 
   renderImage() {
-    const obs = this.sim.getObservation("rgb", null);
-    this.imageData.data.set(obs.getData());
-    // convert from linear to sRGB gamma
-    this.applyGamma(this.imageData.data, 2.2);
-    this.imageCtx.putImageData(this.imageData, 0, 0);
+    // Not using this as of now, but useful for later tracking purposes
+    // Also, required to render everything on sim side
+    this.lastObs = this.sim.getObservation("rgb", null);
     this.renderRadar();
   }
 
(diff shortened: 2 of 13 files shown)


src/esp/gfx/RenderTarget.cpp
```diff
@@ -114,6 +114,17 @@ struct RenderTarget::Impl {
 
   void renderExit() {}
 
+  void blitRgbaToDefault() {
+    GL::defaultFramebuffer.clearColor(Color4{1, 0, 0, 1});
```
msbaines: Not needed.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -114,6 +114,17 @@ struct RenderTarget::Impl {
...
+    GL::defaultFramebuffer
+        .clear(GL::FramebufferClear::Color | GL::FramebufferClear::Depth)
+        .bind();
```
msbaines: I don't think the bind is needed. blit will bind.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -114,6 +114,17 @@ struct RenderTarget::Impl {
...
+    framebuffer_.mapForRead(RgbaBuffer);
```
msbaines: I don't think mapForRead is needed either.

src/esp/bindings_js/navigate.js
```diff
@@ -63,11 +62,9 @@ class NavigateTask {
   }
 
   renderImage() {
-    const obs = this.sim.getObservation("rgb", null);
-    this.imageData.data.set(obs.getData());
-    // convert from linear to sRGB gamma
-    this.applyGamma(this.imageData.data, 2.2);
```
msbaines: You can delete applyGamma.

src/esp/bindings_js/bindings.html
```diff
@@ -86,7 +85,7 @@ <h1>Habitat</h1>
       const simenv = new SimEnv(config, episode, 0);
       const agent = simenv.addAgent(agentConfig);
       const task = new NavigateTask(simenv, {
-        canvas: document.getElementById('canvas2d'),
+        canvas: document.getElementById('canvas'),
```
msbaines: I think we can delete canvas since its no longer needed.

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -91,6 +96,9 @@ bool PinholeCamera::getObservation(gfx::Simulator& sim, Observation& obs) {
 
   renderTarget().renderExit();
 
+  if (is_webgl_build_) {
```
erikwijmans: You can use `#ifdef CORRADE_TARGET_EMSCRIPTEN` here instead.

src/esp/gfx/RenderTarget.cpp
erikwijmans: Yeah, the call to `GL::AbstractFramebuffer::blit` should be the only thing you need to do here.

src/esp/gfx/RenderTarget.cpp
erikwijmans: Actually, this may need to stay as otherwise blit won't know which attachment to blit from.

msbaines: You're right. The mapForRead is necessary.

src/esp/sensor/PinholeCamera.cpp
apsdehal: IMO, this would be better if we have to use this check at multiple places. Also, if we find a better way to do this then final change will need to happen only at one place.

src/esp/bindings_js/bindings.html
apsdehal: The canvas element is still needed for "mousedown" event.

src/esp/sensor/PinholeCamera.cpp
erikwijmans: In general, you should use `#ifdef` where you can as that signifies that whether or not this code runs is determined at compile time, which is both nice for your reader and for the compiler (as it can trivially remove that code).  If we were requiring c++17, we could use `if constexpr`, so the syntax to do `#ifdef` will eventually get cleaner :)

If you feel very strongly about `is_webgl_build_`, we should at least do

```
const bool is_webgl_build_ = 
#ifdef CORRADE_TARGET_EMSCRIPTEN
    true;
#else
    false;
#endif
```

So that the compiler can, hopefully, optimize out the if statements (we should avoid adding unnecessary jumps to something so core as getting observations).

With all this said, it does seem weird that the sensor is responsible for calling `renderTarget().blitRgbaToDefault()`. Maybe add either an argument to `SimulatorWithAgents::getAgentObservation` to do the blit or add a `SimulatorWithAgents::blitRgbaToDefault(int agentId, const std::string& sensorId)` method and call that from JS?

src/esp/bindings_js/navigate.js
```diff
@@ -56,42 +55,38 @@ class NavigateTask {
...
+
     const obs = this.sim.getObservation("semantic", null);
     this.semantic_data = obs.getData();
     let data = this.semantic_data;
 
     // TOOD(msb) implement a better colorization scheme
     for (let i = 0; i < 640*480; i++) {
       if (data[i*4] & 1) {
-	this.semanticImageData.data[i*4] = 255;
+        this.semanticImageData.data[i*4] = 255;
```
erikwijmans: If you guys want, `clang-format` will format JS code!  Just update the regex here: https://github.com/facebookresearch/habitat-sim/blob/master/.circleci/config.yml#L48 and here: https://github.com/facebookresearch/habitat-sim/blob/master/.pre-commit-config.yaml#L30

src/esp/gfx/RenderTarget.cpp
erikwijmans: Looks like you'll also need to remove https://github.com/facebookresearch/habitat-sim/blob/f8c5a1b55f90a9f9706abd104412e0a9e96d3faf/.clang-format#L2 as otherwise it'll only try to format c++

src/esp/sensor/PinholeCamera.cpp
msbaines: SimulatorWithAgents::blitRgbaToDefault sounds reasonable. Maybe a different name though like displayObservation?. It would also be nice to avoid the readFrameRgba in the blit path since its expensive and the data isn't used.

src/esp/gfx/RenderTarget.cpp
apsdehal: I will open up a separate PR for formatting and linting JS.

src/esp/sensor/PinholeCamera.cpp
apsdehal: Thanks both. The overall motivation of Erik's original comment is much more clear now. I will soon push a fix.

src/esp/gfx/RenderTarget.cpp
erikwijmans: No idea why this comment showed up in this thread :)

apsdehal:
I have updated the code and made changes according to the comments (while taking some decisions on my own as well). Happy to hear more thoughts.

src/esp/sensor/PinholeCamera.h
```diff
@@ -53,6 +58,7 @@ class PinholeCamera : public Sensor {
   float near_ = 0.001f;  // near clipping plane
   float far_ = 1000.0f;  // far clipping plane
   float hfov_ = 35.0f;   // field of vision (in degrees)
+  bool is_webgl_build_ = false;
```
mosra: I don't see this used anywhere anymore ... a leftover? :)

erikwijmans: Same comment :)

apsdehal: Thanks. Fixed it.

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -55,15 +55,14 @@ bool PinholeCamera::getObservation(gfx::Simulator& sim, Observation& obs) {
...
 
-  // Make sure we have memory
-  if (buffer_ == nullptr) {
-    // TODO: check if our sensor was resized and resize our buffer if needed
-    ObservationSpace space;
-    getObservationSpace(space);
-    buffer_ = core::Buffer::create(space.shape, space.dataType);
-  }
-  obs.buffer = buffer_;
+  drawObservation(sim);
```
erikwijmans: The `renderTarget().renderEnter();` and `renderTarget().renderExit();` calls can be put into `drawObservation` as they are only needed for drawing.

src/esp/sensor/PinholeCamera.h
```diff
@@ -36,8 +36,25 @@ class PinholeCamera : public Sensor {
...
   virtual bool getObservationSpace(ObservationSpace& space) override;
 
+  virtual bool displayObservation(gfx::Simulator& sim) override;
+
+  /**
+   * @brief Draw an observation using simulator's renderer
+   * @param[in] sim Instance of Simulator class for which the observation needs
+   *                to be drawn
+   */
+  void drawObservation(gfx::Simulator& sim);
```
msbaines: This can be private or protected.

src/esp/sensor/PinholeCamera.h
```diff
@@ -36,8 +36,25 @@ class PinholeCamera : public Sensor {
...
+
+  /**
+   * @brief Read the observation that was rendered by the simulator
+   * @param[in,out] obs Instance of Observation class in which the observation
+   *                    will be stored
+   */
+  void readObservation(Observation& obs);
```
msbaines: This can be private or protected.

apsdehal:
Addressed the latest comments. Merging after tests pass. :) Thanks everyone for comments.

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -88,6 +99,17 @@ bool PinholeCamera::getObservation(gfx::Simulator& sim, Observation& obs) {
...
         obs.buffer->data});
   }
+}
+
+bool PinholeCamera::displayObservation(gfx::Simulator& sim) {
+  if (!hasRenderTarget()) {
+    return false;
+  }
+
+  renderTarget().renderEnter();
```
erikwijmans: This should have been deleted

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -88,6 +99,17 @@ bool PinholeCamera::getObservation(gfx::Simulator& sim, Observation& obs) {
...
+
+  drawObservation(sim);
+  renderTarget().blitRgbaToDefault();
 
   renderTarget().renderExit();
```
erikwijmans: Same with this here

PR #217, opened 2019-09-17
Use clang-format for js code.

msbaines:
## Motivation and Context

We would like to use clang-format for js.

## How Has This Been Tested

Ran clang-format on js code and verified that running clang-format on C++ code caused no diff.

## Types of changes

## Checklist


.pre-commit-config.yaml
```diff
@@ -27,5 +27,5 @@ repos:
       name: Run clang-format
       entry: clang-format -i -style=file
       types: [text]
-      files: '.*\.cpp|.*\.h'
+      files: '.*\.cpp|.*\.h|.*\.js'
```
erikwijmans:
```suggestion
      files: '.*\.cpp|.*\.h|.*\.js$'
```

src/esp/bindings_js/simenv_embind.js
```diff
@@ -11,7 +11,8 @@ class SimEnv {
    * @param {number} agentId - default agent id
    */
   constructor(config, episode, agentId) {
-    this.sim = new Module.Simulator(config);;
+    this.sim = new Module.Simulator(config);
+    ;
```
erikwijmans:
```suggestion
```

src/esp/bindings_js/navigate.js
```diff
@@ -12,26 +12,30 @@ class NavigateTask {
...
-    this.radarCtx = components.radar.getContext("2d");
+    components.canvas.onmousedown = e => {
+      this.handleMouseDown(e);
+    } this.radarCtx = components.radar.getContext('2d');
     this.actions = [
-      { name: 'moveForward', key: 'w' },
-      { name: 'lookLeft', key: 'a', },
-      { name: 'lookRight', key: 'd' },
-      { name: 'done', key: ' ' }
+      {name: 'moveForward', key: 'w'}, {
```
apsdehal: This formatting doesn't look right. Maybe worth to take a look if there is a flag that needs to be enabled.

src/esp/bindings_js/navigate.js
```diff
@@ -12,26 +12,30 @@ class NavigateTask {
...
-    let shape = this.sim.getObservationSpace("rgb").shape;
+    this.imageCtx = components.canvas.getContext('2d');
+    let shape = this.sim.getObservationSpace('rgb').shape;
     this.imageData = this.imageCtx.createImageData(shape.get(1), shape.get(0));
-    this.semanticCtx = components.semantic.getContext("2d");
-    shape = this.sim.getObservationSpace("semantic").shape;
-    this.semanticImageData = this.semanticCtx.createImageData(shape.get(1), shape.get(0));
+    this.semanticCtx = components.semantic.getContext('2d');
+    shape = this.sim.getObservationSpace('semantic').shape;
+    this.semanticImageData =
```
apsdehal: Same here. Maybe we should move to black. Worth checking if there are options there.
Black would make this
```js
this.semanticImageData = this.semanticCtx.createImageData(
    shape.get(1), shape.get(0)
)
```
which looks much nicer.

erikwijmans: Does black work on javascript?  I thought it was python only

apsdehal: Oh yeah. Sorry, working too much with python. There is a way to make it work with JavaScript but it is not worth it. In my mind, I was actually talking about eslint.

apsdehal: ESLint is very nice in doing all this and very customizable.

msbaines: Does black support javascript?

I wasn't able to find any option to fix this.

src/esp/bindings_js/navigate.js
msbaines: I was able to fix this a little bit.

msbaines:
Abandoning this change. Will investigate ESLint instead since it is more widely used and clang-format has issues: weird formatting in cases and one instance where it broke code (closure).

PR #218, opened 2019-09-17
Update Magnum for the latest goodies

mosra:
## Motivation and Context

With this you get:

- Magnum Python Bindings working on Windows (helping @cegbertOculus and #84)
- Corrade and Magnum fixed to have global data are shared across `*.so` modules, which unblocks further usage of Magnum Python bindings in the code -- feel free to *finally* start touching all GL APIs from Python!
  - however not for Windows yet, there the current workaround is to build everything as shared and not static, fix is scheduled to ship in the 2019.09 release
- Python bindings extended with:
  - SDL/GLFW application improvements (key/mouse bindings, ...), which means the GUI Python things could use those instead of `cv2`
  - Texturing
  - Full texture-enabled Flat and Phong shaders as well as all interfaces for compiling your own shaders (see the Python Textured Triangle example, if you know how to do it in C++, Python is very similar)
  - Plugin management and asset import (currently just images and meshes, scene hierarchy and materials coming later)
  - Vector swizzles
  - Lots of other polish and loose-end-tying
- Updated Assimp importer, with a fix for ambient textures that I got reported privately on Slack
- Related updates in Magnum / Corrade, needed to make the above possible, one rather minor but notable thing is addition of
[Math::gather()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Math.html#aeda0faa04b927f5c7aa2bbc3a8794d7d) and [Math::scatter()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Math.html#a04d57d5b0e229035590046fed3496502)

:arrow_right: A big overview of what all went into Magnum Python Bindings and how it works under the hood is in this article, published earlier today: https://blog.magnum.graphics/announcements/introducing-python-bindings/ :sparkles: 

## How Has This Been Tested

I'm running against Magnum `master` every time I compile Habitat anyway (`USE_SYSTEM_MAGNUM` enabled), so I don't expect any new issues popping up. And the CI seems to agree with me :)

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -391,7 +391,7 @@ class Sensor:
             else:
                 tgt.read_frame_rgba(
                     mn.MutableImageView2D(
-                        mn.PixelFormat.RGBA8UNORM,
+                        mn.PixelFormat.RGBA8_UNORM,
                         size,
                         self._buffer.reshape(self._spec.resolution[0], -1),
                     )


erikwijmans:
> SDL/GLFW application improvements (key/mouse bindings, ...), which means the GUI Python things could use those instead of cv2

cc: @angelxuanchang

dhruvbatra:
> > SDL/GLFW application improvements (key/mouse bindings, ...), which means the GUI Python things could use those instead of cv2
> 
> cc: @angelxuanchang

Also, cc: @msbaines and @apsdehal

PR #223, opened 2019-09-18
[esp/bindings_js] add support for top-down map

msbaines:
## Motivation and Context

Add support for top-down map.

## How Has This Been Tested

Screen Shot 2019-09-18 at 12 15 07 PM

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/topdown.js
+++ b/src/esp/bindings_js/topdown.js
@@ -1,0 +1,92 @@
+/**
+ * TopDownMap class
+ */
+// eslint-disable-next-line no-unused-vars
+class TopDownMap {
+  // PUBLIC methods.
+
+  /**
+   * Create top-down map.
+   * @param {PathFinder} pathFinder - pathFinder used to create map
+   * @param {HTMLCanvasElement} canvas - canvas element for drawing map
+   */
+  constructor(pathFinder, canvas) {
+    this.pathFinder = pathFinder;
+    this.canvas = canvas;
+    this.ctx = canvas.getContext("2d");
+    this.imageData = this.createMap();
+  }
+
+  /**
+   * Draw the topdown map to its canvas.
+   */
+  draw() {
+    this.ctx.putImageData(this.imageData, 0, 0);
+  }
+
+  // PRIVATE methods.
+
+  /*
+   * Produces map using a regular grid sampling of navigability.
+   */
+  createMap() {
+    const canvas = this.canvas;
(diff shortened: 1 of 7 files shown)


src/esp/bindings_js/topdown.js
```diff
@@ -0,0 +1,90 @@
...
+   * Draw the topdown map to its canvas.
+   */
+  draw() {
+    this.ctx.putImageData(this.imageData, 0, 0);
+  }
+
+  // PRIVATE methods.
+
+  createMap() {
+    let ctx = this.ctx;
```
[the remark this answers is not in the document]
apsdehal: Let me actually look into ESLint rule for this. It will autofix those.

src/esp/bindings_js/topdown.js
```diff
@@ -0,0 +1,90 @@
...
+    let y = bounds.min[heightComponent] + increment / 2;
+    for (i = 0; i < heightInPixels; i++) {
+      let x = bounds.min[widthComponent] + increment / 2;
+      for (j = 0; j < widthInPixels; j++) {
+        let point = [0, 0, 0];
+        point[widthComponent] = x;
+        point[heightComponent] = y;
+        let isNavigable = pathFinder.isNavigable(point, 0.5);
+        if (isNavigable) {
+          imageData.data[i * stride + j * 4] = 0xff;
```
apsdehal: Maybe make the color constant for easy change?

src/esp/bindings_js/topdown.js
```diff
@@ -0,0 +1,90 @@
...
+          imageData.data[i * stride + j * 4 + 2] = 0xff;
+          imageData.data[i * stride + j * 4 + 3] = 0xff;
+        }
+        x += increment;
+      }
+      y += increment;
+    }
+    return imageData;
+  }
+}
```
mosra: Totally random comment: wouldn't this exact functionality be useful also on the C++/Python side (or alternatively, isn't there one already)? In that case maybe have it implemented in C++ and push only the final binary image to JS?

msbaines: There is a python implementation but not a C++ implementation. I wanted to write a C++ implementation but couldn't come up with an interface I was happy with. I would need to pass in the sampling distance, orientation, and color (or return a 2D vector of booleans). I would still need to js or python specific code for doing the final render. However, a C++ implementation would be way more efficient. I will revisit implementing in C++ in a future change once I can figure out an API I'm happy with.

PR #225, opened 2019-09-19
Replace ASSERT with CORRADE_ASSERT in PTexMeshData.cpp

bigbike:
If you look into the implementation of ASSERT, you will find that it actually cannot output any error message when the condition is failed.

Apply the CORRADE_ASSERT in the Corrade library instead.

In this PR:
-) replaced all the ASSERT with CORRADE_ASSERT;
-) added "Error: " to some of the error messages;

## Types of changes

## Checklist


bigbike:

src/esp/assets/PTexMeshData.cpp
```diff
@@ -512,7 +538,7 @@ void PTexMeshData::parsePLY(const std::string& filename,
       colorOffsetBytes = offsetSoFarBytes;
       offsetSoFarBytes += colorBytes;
     } else {
-      ASSERT(false);
+      CORRADE_ASSERT(false, "Error: unknown.", );
```
mosra: `CORRADE_ASSERT_UNREACHABLE()` would be more descriptive in this case I think.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -469,13 +492,16 @@ void PTexMeshData::parsePLY(const std::string& filename,
...
     }
 
     // Check things make sense.
-    ASSERT(numVertices > 0);
-    ASSERT(positionDimensions > 0);
+    CORRADE_ASSERT(numVertices > 0,
+                   "Error: number of vertices is not greater than 0", );
+    CORRADE_ASSERT(
+        positionDimensions > 0,
+        "Error: the dimensions of the position is not greater than 0", );
```
mosra: Is the `Error: ` prefix really needed? :) In every case it's the last thing you see before it aborts.

In Magnum itself I'm instead prefixing every assert with class/function name to make it easier to figure out where it blew up. In this case it would be e.g. `PTexMeshData::parsePly(): the dimensions...`.

(Also, I'm planning to add configurable message prefixes / coloring in the near future.)

src/esp/assets/PTexMeshData.cpp
bigbike: Good to know. will change.

src/esp/assets/PTexMeshData.cpp
bigbike: Good suggestion. I used to do what you recommended as well, and then I stopped to do so since the ASSERT, which used LOG, would output the filename and line number.

@mosra: Another question to you about the CORRADE_ASSERT, it seems it would generate a run-time error instead of exiting the program (like e.g., exit(-1) does)?

mosra: A bit more background on this: there's `CORRADE_INTERNAL_ASSERT()` which doesn't have a message but OTOH prints file/line info. That one is designed for internal assertions which shouldn't ever fire as a result of bad API usage by the library user -- and since the internal error messages would be in most cases just `"internal state that's too complex to explain in a single sentence is totally messed up, sorry for your bad experience"` I decided to not have them there.

OTOH I thought showing file/line info in `CORRADE_ASSERT()` would be mostly useless for the user as they rather need to know what function *name* caused everything to blow up (as opposed to knowing it's line 2765 of AbstractTexture.cpp, for example). So for those there's just a human-readable formatted message.

Basically when deciding which of them to use, I think whether the assertion can happen rather from bad usage on the API user side (in which case it's more important to show *why* is it wrong), or it's just an internal sanity check that shouldn't hit when just passing bad parameters to a function (in which case file/line info is better). ... Hope those design decisions still make sense now :)

> it seems it would generate a run-time error instead of exiting the program (like e.g., exit(-1) does)?

Yes, it calls `std::abort()` instead of `std::exit()` -- reason is that the former generates a proper backtrace (or enters a debugger on some systems), which is what you usually want when an assert hits. The `exit()` would just exit without giving you any chance to know from where it was fired.

bigbike: OK, I see your point on the design decisions. I totally understand you. 

To decide which one to use, we need to think about the purpose of the error message (to expedite debugging or FYI only), and who is the audience (the developer or the user of the system).

Though it is pretty vague in some cases, it is not hard to determine.

I will go over all the places I changed, and update accordingly.

bigbike: Oh, you used `std::abort()`. That is why.

Yes, the returned backtrace is rather useful sometimes for the developer. 

Again, if the audience is a user of our simulator, in my humble opinion, it is more comfortable for her to see the program is terminated normally (exit()) instead of abnormally (abort()).

erikwijmans: But it didn't terminate normally -- It terminated due to a failed assertion!

bigbike: > But it didn't terminate normally -- It terminated due to a failed assertion!

Yes, indeed, that is actually what happened. I agree, but that is just from the developer's perspective.

But from a user's view, maybe it is more comfortable if the system can "pretend" to quit "normally",meaning it exits without a run-time error (otherwise, there would be an annoying pop-up dialog-box on mac e.g.,).

bigbike:
Code is updated. Let me know if you have more concerns. Thanks.

src/esp/assets/PTexMeshData.cpp
erikwijmans: I definitely don't think people want things to crash with no indication that anything went wrong.  It seems very for them to figure out that something is broken if we don't inform them in a fairly loud way.

mosra: If you want to `exit()` with a message, use `Fatal{} << blah;`. It exits the "normal" way, it's what `Utility::Arguments` do when you pass a wrong thing on the command line. Asserts are loud by definition and yes, are meant for developers. If a non-developer hits an assert, it's the app developers fault this wasn't checked and handled gracefully.

erikwijmans: Hmmm.  Maybe something like `FATAL_IF(<blah>, <msg>)` would be useful?  Some of the stuff here is checking if a file exists, so a user could in theory hit that.  The assert style synatx is nice than `if (<blah>) Fatal{} << <msg>;` IMO

mosra: As I understand it, asserts are macros mainly so there's a possibility to compile them out in optimized builds. Apart from that I'm trying to avoid macros as much as possible, so the non-macroy `if()` is no problem for me.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -35,12 +36,22 @@ namespace assets {
...
 void PTexMeshData::load(const std::string& meshFile,
                         const std::string& atlasFolder) {
-  ASSERT(io::exists(meshFile));
-  ASSERT(io::exists(atlasFolder));
+  if (!io::exists(meshFile)) {
+    Cr::Utility::Fatal{-1} << "PTexMeshData::load: Mesh file " << meshFile
+                           << " does not exist.";
+  }
+  if (!io::exists(atlasFolder)) {
+    Cr::Utility::Fatal{-1} << "PTexMeshData::load: The atlasFolder "
```
bigbike: @mosra : I used -1 as the exit code. Hope it is fine.

bigbike:
Let me know if you have more concerns before I merge it in.

src/esp/assets/PTexMeshData.cpp
mosra: Default if you don't specify anything is `1`. Anything except zero is fine I guess, and since we don't document return codes anywhere, the value probably doesn't matter at all.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -678,12 +725,18 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
...
+                     << submeshID << " is out of range.",
+                 nullptr);
   return renderingBuffers_[submeshID].get();
 }
 
 Magnum::GL::Mesh* PTexMeshData::getMagnumGLMesh(int submeshID) {
-  ASSERT(submeshID >= 0 && submeshID < renderingBuffers_.size());
+  CORRADE_ASSERT(submeshID >= 0 && submeshID < renderingBuffers_.size(),
+                 "PTexMeshData::uploadBuffersToGPU: the submesh ID "
+                     << submeshID << " is out of range.",
```
mosra:
```suggestion
                     << submeshID << "is out of range.",
```

Neither with `Debug` / `Warning` / `Error` / `Fatal` nor with the assert macros you need to wrap values with spaces, that's done automatically to save you extra typing. Putting the space there would make it doubly-spaced. [Docs.](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html)

Same in all other cases ;)

bigbike:
Thanks all reviewers.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -35,12 +36,22 @@ namespace assets {
...
+  }
+  if (!io::exists(atlasFolder)) {
+    Cr::Utility::Fatal{-1} << "PTexMeshData::load: The atlasFolder"
+                           << atlasFolder << "does not exist.";
+  }
 
   // Parse parameters
   const auto& paramsFile = atlasFolder + "/parameters.json";
-  ASSERT(io::exists(paramsFile));
+  if (io::exists(paramsFile)) {
```
bigbike: This is a bug.

PR #226, opened 2019-09-19
[build_js.sh] add support for Ninja

msbaines:
## Motivation and Context

Ninja is much faster for incremental builds.

## How Has This Been Tested

Build.

## Types of changes

## Checklist

### DIFF
--- a/build_js.sh
+++ b/build_js.sh
@@ -7,9 +7,14 @@ git submodule update --init --recursive
 
 DATA_DIR="$(pwd)/data/"
 
+GENERATOR="Unix Makefiles"
+if which -s ninja; then
+  GENERATOR="Ninja"
+fi
+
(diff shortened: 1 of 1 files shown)


msbaines:
Abandoning. Discussed with Amanpreet. This can be done much simpler using a wrapper script or exported bash function.

msbaines:
> By the way, did you know there's a new [CMAKE_GENERATOR](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR.html) env var that you can use to force `-G Ninja` being done by default? It's new since CMake 3.15.

Cool. Thanks much for pointing this out! I had written a bash wrapper function to set -G Ninja but this is much simpler. I think Amanpreet had also mentioned it to me but it slipped my mind at that the time.

PR #227, opened 2019-09-19
Physics benchmarking

aclegg3:
## Motivation and Context

We would like to analyze the impact of non-static object manipulation through the new physics integration introduced by PR #91 .

This PR adds benchmarking for use of kinematic(default) and dynamic(Bullet physics) objects in habitat sim to benchmark.py with rgb and rgbd sensor modes.

Added new benchmark for time inside step function.

Added benchmark script parameters for number and type of object to simulate during benchmark.

To accommodate testing, MotionType has been added to python bindings in a physics module.

Bug fix: timestep was not making from config file to PhysicsManager initialization.

## How Has This Been Tested

Local and devfair benchmarking:

### Setup:
- 10 cheezit boxes are loaded into a grid above the skokloster castle scene center table.
- Simplified bounding box collision meshes are used for dynamics.
- Each call to simulator.step() requests 1/60 sec simulation duration.
- Fixed timestep of 0.004166... is used to step the simulation forward. Requested simulation duration is truncated to increments of the fixed timestep (4 updates per sim step)
- Each step, all active objects are moved (kinematic) or activated and shoved (dynamic).
- FPS to realtime factor can be computed as: FPS\*4\*0.004
- run python examples/benchmark.py --enable_physics

### Results of kinematics benchmark on devfair:
image

### Results of dynamics benchmark on devfair:

With 10 cheezit objects (simplified):
image

With 10 chefcan objects (complex meshes):
image

### Benchmark notes:
- Includes rendering of 10 additional objects when phys is enabled.
- Cheezit box is likely the simplest collision mesh we will find
- Kinematics (with additional object rendering) results in 6-9% slowdown.
- Dynamics with rgb seems similarly costly to adding depth sensor for 10 simple (cheezit) objects. However, this cost compounds when using rgbd with physics. 10 more complex objects (chefcan) results in greatly reduced performance, as expected.
- Realtime factor for this setup (with rgbd) is ~2.5x on 1 process and ~8x on 5 processes.

## Types of changes

## Checklist


dhruvbatra:
Quite useful! Thank you!

> Kinematics (with additional object rendering) results in 6-9% slowdown.

Very cool! 

> Dynamics with rgb seems similarly costly to adding depth sensor for 10 simple objects. However, this cost compounds when using rgbd with physics.

Still, physics+RGBD at ~500 FPS is a great starting point!

erikwijmans:
Definitely seems like your devfair's load is hurting things.  I can ~1550 RGB FPS at 128x128 on mine.

erikwijmans:
I am getting closer to ~300 FPS for bullet mode for phys_rgb at 128x128 (and really all the resolutions) :)

examples/benchmark.py
```diff
@@ -102,8 +127,28 @@
...
+            row += "\t%-8.1f" % value.get("fps")
         print(row)
     print(
         " =============================================================================="
     )
+
+    # also print the average time per simulation step (including object perturbations)
+    if args.enable_physics:
+        print(
+            " ================ Performance (step time: seconds) NPROC={} ===================================".format(
```
mosra: Thanks for showing also step time and not just FPS :+1: A minor thing: could the frame time be in milliseconds instead of seconds? There's a lot of leading zeros, which makes a difference between 0.00012 and 0.0012 look less significant than it would with 0.12 and 1.2 :)

PR #231, opened 2019-09-20
Physics usability improvements

mosra:
## Motivation and Context

Sweeps away a bunch of roadblocks I ran into when crafting screenshots for #152.

- Naming the Bullet-enabling CMake var consistently, turning it into a CMake option so it's discoverable even for people that don't use `setup.py` at all (i.e., me). Not a user-facing change as the public `setup.py` variable is still named the same.
- Using proper directory joining for physics object filenames. This allows absolute filenames to be used inside the JSON configs (i.e., `/home/mosra/Data/fair/1199943_collision.glb`) without those being turned into `./data//home/mosra/Data/fair/1199943_collision.glb`.
- Printing the physics simulation stats every frame actually slows everything down quite a lot (depending on how slow is your terminal), so not doing that. After that change I got back to 60 FPS in the viewer.
- Improving the physics config file passing.
     * Use an absolute default path so it doesn't depend on current working directory.
     * Instead of dying on an assert inside rapidjson, check file existence and fail gracefully with a clear message.
     * Use consistent naming scheme for the option.
- Exiting after an error shouldn't use a success return code.
- `--enable-physics` w/o Bullet enabled is most probably an error, so warn on that. People *might* still want to do this, so not making this a hard error. (Alternatively, `--enable-physics` could be compiled out when Bullet is not enabled, if that makes more sense for you.)
- Improving usability of the banana throwing. When all files fail to load, it would crash because of a modulo with zero. And when physics is not enabled, the O key would silently do nothing, which isn't great either.
- Enable 4x MSAA for the viewer so the imported models are less aliased. Performed well even on my Intel 630 on a 4K screen, and I expect even less perf issues for you all with NVidias :)

## How Has This Been Tested

Creating those screenshots was quite a soak test for all the viewer functionality, and it survived.


src/esp/gfx/Viewer.cpp
```diff
@@ -126,6 +136,13 @@ Viewer::Viewer(const Arguments& arguments)
   renderCamera_->node().setTransformation(
       cameraNode_->absoluteTransformation());
 
+#ifndef ESP_BUILD_WITH_BULLET
+  if (enablePhysics_)
+    LOG(WARNING) << "Physics was enabled but the project is built without "
+                    "Bullet support, objects added with the O key won't "
+                    "interact with the environment in any way.";
+#endif
```
aclegg3: Note that one use case is to define "physics simulator": "none" in default.phys_scene_config.json to enable kinematic objects only. This is also the default behavior when "bullet" is set in config and Habitat-sim was not installed with Bullet. In this case [ResourceManager line 136](https://github.com/facebookresearch/habitat-sim/blob/be178edd9a5e71fb5964631748597c6ac6668b2f/src/esp/assets/ResourceManager.cpp#L136) prints an error). Maybe we should modify that text to be more precise instead of placing another warning here to avoid warning when the user intends for kinematics?

mosra: You know better than me what's the proper place where to put this and what should it say, so feel free to push commits on top of this -- I promise to not force-push :)

In my case I didn't get *any* warning when I simply used the default-provided files with default settings, so that's why I put it here.

src/esp/assets/ResourceManager.cpp
```diff
@@ -133,7 +133,11 @@ bool ResourceManager::loadScene(
     _physicsManager.reset(new physics::BulletPhysicsManager(this));
     defaultToNoneSimulator = false;
 #else
-    LOG(ERROR) << "trying to use BULLET engine, but not installed";
+    LOG(WARNING)
+        << ":\n---\nPhysics was enabled and Bullet physics engine was "
+           "specified, but the project is built without Bullet support. "
+           "Objects added to the scene will be restricted to kinematic updates "
+           "only. Reinstall with --bullet to enable Bullet dynamics.\n---";
```
mosra: `with --with-bullet`? :)

Looking again, this error was actually there, but so buried among the others that I didn't see it. Now it is nicely loud and clear.

PR #237, opened 2019-09-23
Store and draw axis-aligned bounding boxes for loaded object meshes

aclegg3:
## Motivation and Context

Bounding boxes are useful for many reasons. We want to compute them automatically for each mesh loaded into the resource manager so we can query them later. We were already using them to compute center of mass (COM) transformations to enable physics. 

This update stores the computed bounding boxes for each mesh.

RigidObjects now own a bounding box union of their component meshes. They also have access to a pointer to a SceneNode to draw their bounding box. A toggle function for drawing has been added to the PhysicsManager to manage these bounding box nodes.

This also paves the way for adding arbitrary primitives to the scene (for debugging/visualization) and computing collision meshes from bounding objects. The first step is a structure in ResourceManager to hold references to primitive meshes and a method to add them to the SceneGraph at some node.

## How Has This Been Tested

Local testing with viewer.cpp.

Toggle bounding box draw for a random object with 'B' key.

Results:
toggle_BB

## Types of changes

## Checklist


dhruvbatra:
Looks great!

> Rendering bounding boxes is off be default and currently requires re-compile to turn on.

Can we change that to not require re-compile? I can imagine people using this for visualization purposes.

mosra:
@aclegg3 just FYI: there's a [BulletIntegration::DebugDraw](https://doc.magnum.graphics/magnum/classMagnum_1_1BulletIntegration_1_1DebugDraw.html) meant to be used for visualization purposes. The advantage of it is that it's all drawn in a single call and Bullet can provide more info like object interactions, labels, constraints etc. (Originally there were debug visualizers in the same style as you did now, but in the end I removed that functionality in favor of Bullet's own DebugDraw.

It's used in the Magnum Bullet example, if you want to see it in action.

aclegg3:
Thanks @mosra, I know the DebugDraw code is there, but I'd like to have the bounding boxes and rendering of primitives native to Habitat-sim for a few reasons. The big ones are to leverage the speed of kinematics (no simulator) while not losing functionality and to later allow different simulators without a change to the API for debugging. A bonus is to enable adding primitives as simulation objects and for visual debugging as a user feature.

aclegg3:
> Looks great!
> 
> > Rendering bounding boxes is off be default and currently requires re-compile to turn on.
> 
> Can we change that to not require re-compile? I can imagine people using this for visualization purposes.

Absolutely! I'm working on a slightly less clunky design.

mosra:
> but I'd like to have the bounding boxes and rendering of primitives native to Habitat-sim 

Okay, then I'll prioritize instancing support in the builtin shaders etc. to have this running at a reasonable speed at least :)

src/esp/scene/SceneNode.cpp
```diff
@@ -25,5 +25,48 @@ SceneNode& SceneNode::createChild() {
...
+const Magnum::Range3D& SceneNode::computeCumulativeBB() {
+  // first copy from your precomputed mesh bb
+  cumulativeBB_ = Magnum::Range3D(meshBB_);
+  auto* child = children().first();
+
+  while (child != nullptr) {
+    // try catch before cast incase a non scene node child is attached here.
+    // Better way?
+    try {
+      SceneNode* child_node = static_cast<SceneNode*>(child);
```
erikwijmans: You can use `dynamic_cast` here https://en.cppreference.com/w/cpp/language/dynamic_cast -- that'll return a nullptr if the cast is invalid and you can check for that.

aclegg3:
> Do the bounding boxes get calculated for scene meshes? If so, what is the load-time impact?

Every mesh which passes through `ResourceManager::loadMeshes` gets a local bounding box.
This does not appear to have a noticeable effect on load-time (tested with skokloster and replica apartment 0).

However, only objects added through the `PhysicsManager::addObject` will have a cumulativeBoundingBox computed.

src/esp/assets/ResourceManager.cpp
```diff
@@ -90,6 +92,10 @@ bool ResourceManager::loadScene(const AssetInfo& info,
     // to the void
   }
 
+  // once a scene is loaded, we should have a GL::Context so load the primitives
+  Magnum::Trade::MeshData3D cube = Magnum::Primitives::cubeWireframe();
```
msbaines: Add a:

`namespace Mn = Magnum;`

to the top of the file and use Mn:: instead.

src/esp/assets/ResourceManager.cpp
```diff
@@ -830,6 +837,7 @@ bool ResourceManager::loadGeneralMeshData(
     loadTextures(*importer, &metaData);
     loadMaterials(*importer, &metaData);
     loadMeshes(*importer, &metaData, shiftOrigin, translation);
+
```
msbaines: Maybe remove this white-space change to avoid git history churn.

src/esp/scene/SceneNode.cpp
```diff
@@ -25,5 +25,44 @@ SceneNode& SceneNode::createChild() {
...
+const Magnum::Range3D& SceneNode::computeCumulativeBB() {
+  // first copy from your precomputed mesh bb
+  cumulativeBB_ = Magnum::Range3D(meshBB_);
+  auto* child = children().first();
+
+  while (child != nullptr) {
+    SceneNode* child_node = dynamic_cast<SceneNode*>(child);
+    if (child_node != nullptr) {
+      child_node->computeCumulativeBB();
+      std::vector<Magnum::Vector3> corners;
```
msbaines: For readability, could this be extracted into a getTransformedBB function?

src/esp/scene/SceneNode.h
```diff
@@ -53,6 +54,12 @@ class SceneNode : public MagnumObject {
     return this->absoluteTransformation().translation();
   }
 
+  Magnum::Range3D meshBB_;
```
msbaines: Why are these public?

src/esp/assets/ResourceManager.cpp
```diff
@@ -945,14 +953,20 @@ void ResourceManager::loadMeshes(Importer& importer,
...
     // see if the mesh needs to be shifted
     if (shiftOrigin) {
       // compute BB center if necessary ([0,0,0])
-      if (offset[0] == 0 && offset[1] == 0 && offset[2] == 0)
-        offset = -computeMeshBBCenter(gltfMeshData);
-      // translate the mesh if necessary
-      if (!(offset[0] == 0 && offset[1] == 0 && offset[2] == 0))
+      if (offset[0] == 0 && offset[1] == 0 && offset[2] == 0) {
+        offset = -gltfMeshData->BB.center();
```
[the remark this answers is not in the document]
aclegg3: You are correct, all meshes get the same translation currently. If we are transforming mesh hierarchies, we want all of them transformed together so that later their transforms will relate correctly. Really this shifting was intended to align single object mesh origins with their centers of mass so we could directly consume those meshes to create collision objects. This will not work properly for mesh hierarchies and we need a larger PR to address this down the line. I'll create an issue for this for now.

src/esp/scene/SceneNode.h
aclegg3: To avoid the bloat of public getter/setter functions.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -496,5 +496,28 @@ double PhysicsManager::getAngularDamping(const int physObjectID) {
   }
 }
 
+void PhysicsManager::setObjectBBDraw(int physObjectID,
+                                     DrawableGroup* drawables,
+                                     bool drawBB) {
+  if (existingObjects_.count(physObjectID) > 0) {
```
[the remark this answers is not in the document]
aclegg3: Sure, it should probably crash to request an invalid primitive instead of failing silently.

src/esp/assets/ResourceManager.cpp
aclegg3: Sure, but for consistency we should then update all uses in the file. I'll create a separate PR for this also.

src/esp/scene/SceneNode.h
msbaines: Yeah, it does feel like it adds bloat but for consistency with the rest of the code and the [coding style](https://google.github.io/styleguide/cppguide.html#Access_Control) these should be private. Public members break encapsulation by exposing implementation details in the interface making it harder to change code and ensure interface safety. They prevent you from making accessors read-only or write-only. They also allow clients to store a pointer to such a member, making code harder to reason about.

The one case where data members should be public is for a [pure data structure](https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes). In that case, the trailing under-score is omitted.

I'm hoping to add [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) support soon in order to have the presubmit check for the coding style.

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
aclegg3: Opened an issue for this: #301

src/esp/physics/PhysicsManager.cpp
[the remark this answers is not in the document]
aclegg3: PR for this: #305

PR #239, opened 2019-09-23
[esp/bindings_js] load scene files bases on URL parameters

msbaines:
Instead of using a large .data file, we preload the scene files
into the virtual filesystem based on URL scene parameter. This is
a workaround until Corrade PR#39 is ready.

This should speed up startup since we now only load the files we
need.

## Motivation and Context

Speedup startup by only loading files we need for the current scene.

## How Has This Been Tested

Verified that the new URL parameter scene works.

## Types of changes

## Checklist

- [ X I have completed my CLA (see **CONTRIBUTING**)

### DIFF
--- a/src/esp/bindings_js/index.js
+++ b/src/esp/bindings_js/index.js
@@ -1,12 +1,25 @@
-/*global Module */
+/* global FS, Module */
 
 import WebDemo from "./modules/web_demo";
 import VRDemo from "./modules/vr_demo";
 import "./bindings.css";
 
-Module["onRuntimeInitialized"] = function() {
-  console.log("hsim_bindings initialized");
+Module.preRun.push(() => {
+  let scene = "skokloster-castle.glb";
+  for (let arg of window.location.search.substr(1).split("&")) {
+    let [key, value] = arg.split("=");
+    if (key === "scene" && value) {
+      scene = value;
+    }
+  }
+  FS.createPreloadedFile("/", scene, scene, true, false);
+  Module.scene = scene;
+  const navmesh = scene.substr(0, scene.lastIndexOf(".")) + ".navmesh";
+  FS.createPreloadedFile("/", navmesh, navmesh, true, false);
+});
 
+Module.onRuntimeInitialized = () => {
+  console.log("hsim_bindings initialized");
   let demo;
   if (window.vrEnabled) {
     if (navigator && navigator.getVRDisplays) {
--- a/build_js.sh
+++ b/build_js.sh
@@ -5,8 +5,6 @@ set -e
 
 git submodule update --init --recursive
 
-DATA_DIR="$(pwd)/data/"
-
 mkdir -p build_corrade-rc
 pushd build_corrade-rc
 cmake ../src \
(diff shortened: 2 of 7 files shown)


src/esp/bindings_js/index.js
```diff
@@ -1,12 +1,26 @@
...
 import VRDemo from "./modules/vr_demo";
 import "./bindings.css";
 
-Module["onRuntimeInitialized"] = function() {
-  console.log("hsim_bindings initialized");
+Module.preRun.push(() => {
+  let arg;
+  let scene;
+  for (arg of window.location.search.substr(1).split("&")) {
+    let [key, value] = arg.split("=");
```
apsdehal: I would add a default value here and make sure that args has length 2 before assigning it to `key, value`.

src/esp/bindings_js/index.js
```diff
@@ -1,12 +1,26 @@
...
+    if (key == "scene") {
```
apsdehal:
```suggestion
    if (key === "scene") {
```

src/esp/bindings_js/index.js
```diff
@@ -1,12 +1,26 @@
...
+      scene = value;
+    }
+  }
+  FS.createPreloadedFile("/", scene, scene, true, false);
+  Module.scene = scene;
+  let navmesh = scene.replace(/\..*/, ".navmesh");
```
apsdehal: To cover all kind of cases
```suggestion
  const navmesh = scene.substr(0, scene.lastIndexOf(".")) + ".navmesh";
```

src/esp/bindings_js/index.js
```diff
@@ -1,12 +1,26 @@
...
 
 import WebDemo from "./modules/web_demo";
 import VRDemo from "./modules/vr_demo";
 import "./bindings.css";
 
-Module["onRuntimeInitialized"] = function() {
-  console.log("hsim_bindings initialized");
+Module.preRun.push(() => {
+  let arg;
+  let scene;
```
apsdehal: NIT: To make it clear that this is url, maybe name it sceneUrl?

src/esp/bindings_js/index.js
```diff
@@ -1,12 +1,26 @@
...
+  for (arg of window.location.search.substr(1).split("&")) {
+    let [key, value] = arg.split("=");
+    if (key == "scene") {
+      scene = value;
+    }
+  }
+  FS.createPreloadedFile("/", scene, scene, true, false);
```
apsdehal: Does this cache it once it is loaded?

msbaines: In chrome inspector, I see that it is served from the disk cache.

src/esp/bindings_js/index.js
msbaines: Its also the filename we are using in the virtual filesystem.

src/esp/bindings_js/index.js
msbaines: Good idea! That also preserves the prior behavior.

PR #242, opened 2019-09-24
[esp/bindinds_js] topdown: don't assume floor is at y=0

msbaines:
Replica models have a floor that is non-zero. We were assuming
a floor of zero. Also, made the top-down map generation more
dynamic in order to support going up and down floors.

## Motivation and Context

Top-down map was not rendering correctly for replica scenes.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/modules/topdown.js
+++ b/src/esp/bindings_js/modules/topdown.js
@@ -17,22 +17,20 @@ class TopDownMap {
     this.widthComponent = 0;
     this.heightComponent = 2;
     this.scale = 1.0;
-    this.imageData = this.createMap();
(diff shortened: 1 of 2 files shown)


src/esp/bindings_js/modules/topdown.js
```diff
@@ -49,9 +51,16 @@ class TopDownMap {
    * @param {vec3f} position - new position
    */
   moveTo(position) {
-    let [x, y] = this.convertPosition(position);
-    this.ctx.lineTo(x, y);
-    this.ctx.stroke();
+    if (
```
apsdehal: A comment would be helpful here.

PR #245, opened 2019-09-25
[esp/bindings_js] add back support for semantic information

msbaines:
My earlier change where I allowed specifying the scene broke
semantic mesh support. Adding back semantic mesh support as
a URL parameter.

## Motivation and Context

Get semantic mesh working again with WebGL port.

## How Has This Been Tested

Tested via http://0.0.0.0:8000/esp/bindings_js/bindings.html?scene=17DRP5sb8fy.glb&semantic=mp3d

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/index.js
+++ b/src/esp/bindings_js/index.js
@@ -2,20 +2,31 @@
 
 import WebDemo from "./modules/web_demo";
 import VRDemo from "./modules/vr_demo";
+import { defaultScene } from "./modules/defaults";
 import "./bindings.css";
 
+function preload(file) {
+  FS.createPreloadedFile("/", file, file, true, false);
+}
+
 Module.preRun.push(() => {
-  let scene = "skokloster-castle.glb";
+  let config = {};
+  config.scene = defaultScene;
   for (let arg of window.location.search.substr(1).split("&")) {
     let [key, value] = arg.split("=");
-    if (key === "scene" && value) {
-      scene = value;
+    if (key && value) {
+      config[key] = value;
     }
   }
-  FS.createPreloadedFile("/", scene, scene, true, false);
+  const scene = config.scene;
+  preload(scene);
   Module.scene = scene;
-  const navmesh = scene.substr(0, scene.lastIndexOf(".")) + ".navmesh";
-  FS.createPreloadedFile("/", navmesh, navmesh, true, false);
+  const fileNoExtension = scene.substr(0, scene.lastIndexOf("."));
+  preload(fileNoExtension + ".navmesh");
+  if (config.semantic === "mp3d") {
+    preload(fileNoExtension + ".house");
+    preload(fileNoExtension + "_semantic.ply");
+  }
 });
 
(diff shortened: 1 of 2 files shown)


src/esp/bindings_js/index.js
```diff
@@ -5,17 +5,25 @@ import VRDemo from "./modules/vr_demo";
...
+      config[key] = value;
     }
   }
+  const scene = config.scene;
   FS.createPreloadedFile("/", scene, scene, true, false);
   Module.scene = scene;
   const navmesh = scene.substr(0, scene.lastIndexOf(".")) + ".navmesh";
   FS.createPreloadedFile("/", navmesh, navmesh, true, false);
+  if (config.semantic === "mp3d") {
+    const house = scene.substr(0, scene.lastIndexOf(".")) + ".house";
```
apsdehal: Convert extension replacement code and preload into a function now?

src/esp/bindings_js/index.js
```diff
@@ -5,17 +5,25 @@ import VRDemo from "./modules/vr_demo";
 import "./bindings.css";
 
 Module.preRun.push(() => {
```
apsdehal: At this point, it will be better to move this to its own class which has functions like preloadMp3d etc.

src/esp/bindings_js/index.js
```diff
@@ -5,17 +5,25 @@ import VRDemo from "./modules/vr_demo";
...
-  let scene = "skokloster-castle.glb";
+  let config = {};
+  config.scene = "skokloster-castle.glb";
```
apsdehal: Also, use defaults.js file here to export and load defaults (forgot in last PR).

src/esp/bindings_js/index.js
```diff
@@ -5,17 +5,25 @@ import VRDemo from "./modules/vr_demo";
...
   FS.createPreloadedFile("/", scene, scene, true, false);
   Module.scene = scene;
   const navmesh = scene.substr(0, scene.lastIndexOf(".")) + ".navmesh";
   FS.createPreloadedFile("/", navmesh, navmesh, true, false);
+  if (config.semantic === "mp3d") {
+    const house = scene.substr(0, scene.lastIndexOf(".")) + ".house";
+    FS.createPreloadedFile("/", house, house, true, false);
+    const semantic = scene.substr(0, scene.lastIndexOf(".")) + "_semantic.ply";
+    FS.createPreloadedFile("/", semantic, semantic, true, false);
+  }
```
mosra: Just checking we're all still on the same page -- once I'm done with the filesystem abstraction APIs (mosra/corrade#39), everything that's added here and in #239 could be moved back to C++, right?

msbaines: I think we are on the same page. I am looking forward to deleting this code once ([mosra/corrade#39](https://github.com/mosra/corrade/pull/39)) lands. I'd much rather just use the C++ filesystem API then have to preload files into the virtual filesystem. This is definitely a temporary hack.

PR #246, opened 2019-09-25
cmake: prefer the python3 executable over python

mosra:
## Motivation and Context

In a response to https://github.com/mosra/magnum-bindings/issues/4. On macOS (without any special setup at least), `python` defaults to python 2, and this override here is causing the version 2 to be picked up first.

@erikwijmans actually, I'm not sure about the whole check at all -- why was `find_package(PythonInterp 3.6 REQUIRED)` not enough? This override after is not doing any version checks and if version 2 gets picked up, it just happily continues without any indication of things being wrong. What would break if this whole part was not there?

## How Has This Been Tested

`/usr/bin/python` shouldn't be picked up by accident anymore.

### DIFF
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -88,16 +88,20 @@ target_compile_definitions(Detour
   DT_VIRTUAL_QUERYFILTER)
 
 if(BUILD_PYTHON_BINDINGS)
-  # python interpreter
+  # Before calling find_package(PythonInterp) search for python executable not
+  # in the default paths to pick up activated virtualenv/conda python
+  find_program(PYTHON_EXECUTABLE
+    # macOS still defaults to `python` being Python 2, so look for `python3`
+    # first
+    NAMES python3 python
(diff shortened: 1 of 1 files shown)


erikwijmans:
This may have been fixed with cmake at some point, but last I knew, `find_package(PythonInterp 3.6 REQUIRED)` sometimes wouldn't respect conda/virtual environments.  (IIRC, it's because cmake has some default search paths that it always searches first).

mosra:
Looking into https://github.com/Kitware/CMake/blob/master/Modules/FindPythonInterp.cmake, it's just a fancy version-checking wrapper over `find_executable(PYTHON_EXECUTABLE python)`, and as far as I can see, there is no difference in default search paths between what the Find module does and what the Habitat code has.

mosra:
OK, so all

msbaines:
Happy to see this fixed. I ran into this when I first tried to get habitat-sim working on Mac (without conda). Will let Erik approve.

PR #235, opened 2019-09-21
Added issues template that will make easy to categorize issues

mathfac:
## Motivation and Context
- Added issues template that will make easy to categorize issues
Screenshot 2019-09-20 19 24 47

- Improved `isort` and `black` compatibility: added `toml/pyproject` support to isort precommit and added [an option recommended for compatibility](https://github.com/psf/black#how-black-wraps-lines). Added [seed-isort-config](https://github.com/asottile/seed-isort-config) pre-commit hook to manage third parties list.

- Added `collect_env.py` script to capture information about system environment for debug purposes
**Script result for Linux installation:**
```
ENVIRONMENT INFO:
Platform: Linux-4.15.0-29-generic-x86_64-with-debian-buster-sid
Machine: x86_64
Processor: x86_64
Libc version: glibc 2.9
Mac version:
Python version: 3.7.0
Architecture: 64bit
Win version:
System OS: Linux
Release: 4.15.0-29-generic
Version: #31-Ubuntu SMP Tue Jul 17 15:39:52 UTC 2018
Operational System: linux
GCC version: gcc (GCC) 7.1.0
CMAKE version: cmake version 3.10.1
NVIDIA-SMI: Wed Sep 25 15:34:09 2019
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 410.79       Driver Version: 410.79       CUDA Version: 10.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Quadro GP100        Off  | 00000000:81:00.0 Off |                    0 |
| 26%   40C    P0    33W / 235W |      0MiB / 16278MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Quadro GP100        Off  | 00000000:82:00.0 Off |                    0 |
| 26%   42C    P0    34W / 235W |      0MiB / 16278MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
Pip packages versions:
numba==0.45.1
numpy==1.17.2
numpy-quaternion==2019.7.23.15.26.49
tqdm==4.19.7
Conda packages versions:
attrs                     19.1.0                    <pip>  19.1.0
numba                     0.36.2          np114py36hc6662d5_0
numba                     0.45.1                    <pip>
[...]

PR #252, opened 2019-09-27
docs: sync doc theme with website updates, random fixes.

mosra:
## Motivation and Context

To be consistent with https://aihabitat.org main site theme. This affects habitat-api as well, as that repo looks here for the CSS.

While generating a fresh batch of docs, I discovered a crashy corner-case with the `typing` module, so this pulls in a slightly fresher m.css as well. And in #235 the `habitat_sim.utils.common` submodule got lost, so adding it back (and then reverting the temporary hack for docs so the quat-related functions aren't on two places).

## How Has This Been Tested

Things look consistent again, doc generation is neither crashing nor complaining too much.

### DIFF
--- a/habitat_sim/utils/__init__.py
+++ b/habitat_sim/utils/__init__.py
@@ -4,9 +4,13 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-# Added for backward compatibility with Habitat-API
+# quat_from_angle_axis and quat_rotate_vector imports
+# added for backward compatibility with Habitat-API
 # TODO @maksymets: remove after habitat-api/examples/new_actions.py will be
 # fixed
+
+
+from habitat_sim.utils import common
 from habitat_sim.utils.common import quat_from_angle_axis, quat_rotate_vector
 
-__all__ = ["quat_from_angle_axis", "quat_rotate_vector"]
+__all__ = ["quat_from_angle_axis", "quat_rotate_vector", "common"]
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -651,19 +651,19 @@ STRICT_PROTO_MATCHING  = NO
 # list. This list is created by putting \todo commands in the documentation.
 # The default value is: YES.
 
-GENERATE_TODOLIST      = YES
+GENERATE_TODOLIST      = NO
 
 # The GENERATE_TESTLIST tag can be used to enable (YES) or disable (NO) the test
 # list. This list is created by putting \test commands in the documentation.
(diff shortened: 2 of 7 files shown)


mosra:
Updated based on a request by @abhiskk, similar to https://github.com/facebookresearch/habitat-api/pull/219.

habitat_sim/utils/__init__.py
```diff
@@ -9,4 +9,4 @@
 # fixed
 from habitat_sim.utils.common import quat_from_angle_axis, quat_rotate_vector
```
erikwijmans: Probably needs to have a `from habitat_sim.utils import common` added?

mosra: I admit that, due to the work I did on all the python inspection, I have a deeper-than-necessary knowledge of the module internals and I'm quite sure it works this way too, since `common` get pulled in as a side effect of `from habitat_sim.utils.common import thing`.

Added that explicitly for clarity, thanks.

mosra:
Ugh, I have to admit that commiting *anything* with those commit hooks and CI checks is a total productivity killer. What would take fifteen seconds to do is a 20-minute war against `isort` and `black`, trying various `SKIP=thing git commit` until it comes through (and then being yelled at by the CI again because *somehow* my local config is different from it).

This time `isort` and `black` have a dispute over one newline -- `black` [wants it in](https://circleci.com/gh/facebookresearch/habitat-sim/2878?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) but `isort` [doesn't](https://circleci.com/gh/facebookresearch/habitat-sim/2871?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). I'm not used to tools bullying me like that and I don't like that at all. Any ideas?

erikwijmans:
Yeah.... it is very annoying when isort and black disagree on something.  I have no idea how to get them to agree on this one.

I just changed the file to
```python
#!/usr/bin/env python3

# Copyright (c) Facebook, Inc. and its affiliates.
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.

# quat_from_angle_axis and quat_rotate_vector imports
# added for backward compatibility with Habitat-API
# TODO @maksymets: remove after habitat-api/examples/new_actions.py will be
# fixed


from . import common
from .common import quat_from_angle_axis, quat_rotate_vector

__all__ = ["quat_from_angle_axis", "quat_rotate_vector", "common"]

```
and all seems to be good.

mosra:
Yay. Thanks for restoring my sanity :)

Issue #233, opened 2019-09-20
Agent CI test failing at random


aclegg3:
## Steps to reproduce

  1. Run pytest tests/test_agent.py

## Observed Results

  Sometimes the test will fail on np.allclose() with small error.

![image](https://user-images.githubusercontent.com/1445143/65342739-3273dd00-db88-11e9-9c3d-a4f5a5fe5ede.png)

## Expected Results

The test should pass. It will pass most of the time. Re-running the same failed CI test will often succeed.

## Relevant Code
test_agent.py:76
Maybe the `atol=1e-5` is too tight a bound here?

```
assert np.allclose(s1.postion, s2.position, atol=1e-5)
```

msbaines:
I'm starting to hit this pretty regularly also but I don't think I hit this at all until maybe 1 week ago.

PR #258, opened 2019-09-30
[tests] test_agent.py: make test_change_state more deterministic

msbaines:
This test is randomly failing (#233). Try to make it deterministic
so that it always fails or always passes.

## Types of changes

## Checklist

### DIFF
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -50,7 +50,8 @@ def test_set_state():
 
 
 def test_change_state():
-    np.random.seed()
+    random_state = np.random.get_state()
+    np.random.seed(233)
     scene_graph = habitat_sim.SceneGraph()
     agent = habitat_sim.Agent(scene_graph.get_root_node().create_child())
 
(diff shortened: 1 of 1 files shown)


tests/test_agent.py
```diff
@@ -50,6 +50,7 @@ def test_set_state():
 
 
 def test_change_state():
+    np.random.seed()
```
apsdehal: It is generally a good idea to set back to random number generator for cleanup at the end of your test if you do this.

For getting the state of random generator you can use `state = np.random.get_state()` and then at the end of the test you can set it back using `np.random.set_state(state)`.

erikwijmans: Doesn't `np.random.seed()` just read a "random" number from `/dev/urandom`?  I think you need to provide a seed, like `np.random.seed(0)`.

apsdehal: Yes, I agree. It needs to have some number to be consistent. By default, it just takes current timestamp as the seed.

msbaines: @erikwijmans is right:

https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.random.RandomState.html

> Random seed used to initialize the pseudo-random number generator. Can be any integer between 0 and 2**32 - 1 inclusive, an array (or other sequence) of such integers, or None (the default). If seed is None, then RandomState will try to read data from /dev/urandom (or the Windows analogue) if available or seed from the clock otherwise.

PR #259, opened 2019-10-01
[utils/viewer] move viewer code out of esp/gfx library

msbaines:
## Motivation and Context

Viewer code does not really belong in the esp/gfx library.

## How Has This Been Tested

Ran viewer and took a screenshot.

## Types of changes

## Checklist


bigbike:
@msbaines : The filename is supposed to start with a capital letter. So it should be "Viewer.cpp", not "viewer.cpp".

msbaines:
> @msbaines : The filename is supposed to start with a capital letter. So it should be "Viewer.cpp", not "viewer.cpp".

Yeah, I wasn't sure what the convention was. Whether to name it Viewer.cpp or keep it as viewer.cpp. I kept it as viewer.cpp since that was the current name of this file. It is also consistent with utils/datatool/datatool.cpp. viewer and datatool are the name of the binaries so maybe this convention is OK. I don't have strong opinions on this as long as we are consistent. I am happy to rename viewer.cpp to Viewer.cpp but then we should also change datatool.cpp to Datatool.cpp.

bigbike:
Oh, I see.
When we started this project last year, I remembered our decision was to follow the FB C++ naming conventions found internally. You can search CppStyle internally. I think we should follow it.

bigbike:
Yes, we may need to change the datatool.cpp as well.

PR #262, opened 2019-10-01
[esp/bindings_js] fix Observation leak

msbaines:
We were leaking observations. Fix by allocating one observation and
repeatedly passing it to getObservation. Alternatively, we could
explicitly call delete() on the observation when we are done with it.

## Motivation and Context

Fix leak.

## How Has This Been Tested

Verified that we no longer leak observations.

## Types of changes

## Checklist


src/esp/bindings_js/modules/simenv_embind.js
```diff
@@ -66,9 +66,8 @@ class SimEnv {
    * @param {number} sensorId - id of sensor
```
apsdehal: Update params too, making sure observation is required.

msbaines:
> So, passing null leaks off the observation inside the simulator? Maybe I am missing something as I didn't understand why is that the case. Maybe explain in the comment to the commit?

I updated the commit comment as follows:

Alternatively, we could explicitly call delete() on the observation when we are done with it.

PR #264, opened 2019-10-02
[esp/bindings_js] simplify event handling logic by using event.key

msbaines:
This will also make it easier to handle arrow keys. I also modified the even to keydown because it is more reliable and reduces latency.

## Motivation and Context

First step to getting arrow keys to work.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/modules/navigate.js
+++ b/src/esp/bindings_js/modules/navigate.js
@@ -167,10 +167,9 @@ class NavigateTask {
   }
 
   bindKeys() {
-    document.addEventListener("keyup", event => {
-      const key = String.fromCharCode(event.which).toLowerCase();
+    document.addEventListener("keydown", event => {
       for (let a of this.actions) {
-        if (key === a.key) {
+        if (event.key === a.key) {
           this.handleAction(a.name);
           break;
         }


src/esp/bindings_js/modules/navigate.js
```diff
@@ -167,10 +167,9 @@ class NavigateTask {
   }
 
   bindKeys() {
-    document.addEventListener("keyup", event => {
-      const key = String.fromCharCode(event.which).toLowerCase();
+    document.addEventListener("keydown", event => {
       for (let a of this.actions) {
-        if (key === a.key) {
+        if (event.key === a.key) {
```
apsdehal: Some browsers still don't support event.key. Depends on what browsers we are targeting.

Ideally it should be

```
if (e.key) {
     const key = e.key;
} else {
    const code = e.which || e.keyCode;
    const key  = String.fromCharCode(code);
}
```

msbaines: It looks like it is supported by all the major browsers. We are limited to WebGL2 and WebAssembly so require a modern version.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Browser_compatibility

apsdehal: Ah, I forgot about that. Go ahead with the change then.

PR #269, opened 2019-10-03
[esp/agent] rename lookLeft/Right to turnLeft/Right

msbaines:
lookLeft is not correct. We are actually turning the agent body.
Since only SimulatorWithAgents is using this it should only
affect WebGL. Fixed a bunch of places where Agent.h was included
but not used in order to confirm that only WebGL was affected.

## Motivation and Context

I want to clean up and rationalize actions in Agent.h before adding lookUp and lookDown.

## How Has This Been Tested

Verified that turning still works in WebGL demo.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/agent/Agent.cpp
+++ b/src/esp/agent/Agent.cpp
@@ -16,12 +16,9 @@ using Magnum::EigenIntegration::cast;
 namespace esp {
 namespace agent {
 
-const std::set<std::string> Agent::BodyActions = {
-    "moveRight", "moveLeft", "moveForward", "moveBackward", "turnLeft",
-    "turnRight",
-    // TODO lookLeft and lookRight should not be body actions
-    // turnLeft and turnRight will take their place
-    "lookLeft", "lookRight"};
+const std::set<std::string> Agent::BodyActions = {"moveRight",   "moveLeft",
+                                                  "moveForward", "moveBackward",
+                                                  "turnLeft",    "turnRight"};
 
 Agent::Agent(scene::SceneNode& agentNode, const AgentConfiguration& cfg)
     : Magnum::SceneGraph::AbstractFeature3D(agentNode),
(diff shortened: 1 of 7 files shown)


mathfac:
Looks like more things need to be changed: https://github.com/facebookresearch/habitat-sim/search?q=lookLeft&unscoped_q=lookLeft.

erikwijmans:
> Looks like more things need to be changed: https://github.com/facebookresearch/habitat-sim/search?q=lookLeft&unscoped_q=lookLeft.

Those are fine.  The difference between a lookLeft and turnLeft isn't the implementation, its whether or not it operates on the agent's body or on the sensors.

msbaines:
> Looks like more things need to be changed: https://github.com/facebookresearch/habitat-sim/search?q=lookLeft&unscoped_q=lookLeft.

All the ones that are relevant should now be addressed.

msbaines:
> Those are fine. The difference between a lookLeft and turnLeft isn't the implementation, its whether or not it operates on the agent's body or on the sensors.

I agree. That's why I'm making this change. In the cases I've addressed with this change, we were moving the agents body and calling it lookLeft.

msbaines:
> > Those are fine. The difference between a lookLeft and turnLeft isn't the implementation, its whether or not it operates on the agent's body or on the sensors.
> 
> I agree. That's why I'm making this change. In the cases I've addressed with this change, we were moving the agents body and calling it lookLeft.

To handle look, we should make the sensor a separate node (that is a child of the agent node) in the scene graph if that is not the case already.

erikwijmans:
> To handle look, we should make the sensor a separate node (that is a child of the agent node) in the scene graph if that is not the case already.

Yep, that is exactly how it is done :)

erikwijmans:
> I agree. That's why I'm making this change. In the cases I've addressed with this change, we were moving the agents body and calling it lookLeft.

Yeah, we made this change on the python side, but never put it into cpp

PR #270, opened 2019-10-04
Expose snap point

erikwijmans:
## Motivation and Context

Expose projectToPoly as `snap_point`.  Sometimes it is useful to be able to snap a point to the nearest navigable point, people had been using `pathfinder.try_step(pt, pt)` to do that, but this has some weird edge cases.

## How Has This Been Tested

Via the test!  If you swap `snap_point(pt)` for `try_step(pt, pt)`, it finds a failure very fast.

## Types of changes

### DIFF
--- a/tests/test_snap_point.py
+++ b/tests/test_snap_point.py
@@ -18,13 +18,14 @@ def test_data():
     return pf, pf.get_random_navigable_point()
 
 
-@hypothesis.given(nx=st.floats(-20, 20), ny=st.floats(-5, 5), nz=st.floats(-20, 20))
+@hypothesis.given(
+    nudge=st.tuples(st.floats(-20, 20), st.floats(-5, 5), st.floats(-20, 20))
+)
 @hypothesis.settings(max_examples=int(1e3))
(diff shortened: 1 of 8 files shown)


tests/test_snap_point.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.fixture(scope="function")
+def test_data():
+    pf = habitat_sim.PathFinder()
+    pf.load_nav_mesh(
+        "data/scene_datasets/habitat-test-scenes/skokloster-castle.navmesh"
+    )
+
+    return pf, pf.get_random_navigable_point()
```
msbaines: Will this be deterministic?

erikwijmans: It's not (seeding the pathfinder is pretty much impossible due to recast using just raw `rand()` and `srand` only kinda working).

I have tried a number of invocations and let hypothesis run out to 100k examples and it couldn't find anything.

msbaines: Maybe add a small comment saying that this isn't deterministic.

erikwijmans: Done

Issue #284, opened 2019-10-08
How to create your own scene?


dhuber23b:
I need to create a blank scene for use in Habitat.  If that is not possible, I would like to create a scene with only a plane.

What are the requirements for this?  I tried replicating the information found in the replica header, but I get an error:
F1008 14:08:38.440702 53716416 Simulator.cpp:189] Check failed: activeSemanticSceneID_ >= 0 (-1 vs. 0)

Here is the ply file I am using:
ply
format ascii 1.0
comment Created by Blender 2.80 (sub 75) - www.blender.org, source file: 'cone.blend'
element vertex 3
property float x
property float y
property float z
property float nx
property float ny
property float nz
element face 1
property list uchar uint vertex_indices
property uint16 object_id
end_header
0.000000 0.031314 -1.000000 0.087754 0.445488 -0.890977 
0.000000 2.031314 -0.000000 0.087754 0.445488 -0.890977 
0.195090 0.031314 -0.980785 0.087754 0.445488 -0.890977 
3 0 1 2 0

aclegg3:
Hey Daniel, you can create an empty scene in habitat by setting the scene filepath (e.g. `sim_cfg.scene.id`) to `"NONE"`. This will not create a semantic scene in current master branch. You can see the branch in PR #285 for a fix and example of the results.

aclegg3:
> ... I get an error:
> F1008 14:08:38.440702 53716416 Simulator.cpp:189] Check failed: activeSemanticSceneID_ >= 0 (-1 vs. 0)

This should be fixed by the patch in PR #285.

dhuber23b:
Thank you , Alex.  I will give it a try.

PR #285, opened 2019-10-09
Semantic scene ID assigned by default

aclegg3:
## Motivation and Context

Related to issue #284. A semantic scene ID should be assigned by default so that semantic rendering can be done regardless of the provided annotations. For example on the empty "NONE" scene.

In the future we would like to provide a method for setting those semantic annotations for physics objects programmatically.

Comments welcome.

## How Has This Been Tested

Local testing.

Empty scene with physics object rendered via rgb and semantic sensors:
image
(Image courtesy of a script by Daniel Huber.)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/Simulator.cpp
+++ b/src/esp/gfx/Simulator.cpp
@@ -121,16 +121,17 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
                                    &semanticDrawables);
       }
       LOG(INFO) << "Loaded.";
-    }else{
+    } else {
       activeSemanticSceneID_ = activeSceneID_;
       // instance meshes and suncg houses contain their semantic annotations
       // empty scene has none to worry about
-      if ( !(sceneInfo.type == assets::AssetType::SUNCG_SCENE ||
-        sceneInfo.type == assets::AssetType::INSTANCE_MESH ||
-        sceneFilename.compare(assets::EMPTY_SCENE) == 0)) {
-          //TODO: programmatic generation of semantic meshes when no annotiations are provided.
-          LOG(WARNING)
-            << ":\n---\n The active scene does not contain semantic annotations. \n---";
+      if (!(sceneInfo.type == assets::AssetType::SUNCG_SCENE ||
+            sceneInfo.type == assets::AssetType::INSTANCE_MESH ||
+            sceneFilename.compare(assets::EMPTY_SCENE) == 0)) {
+        // TODO: programmatic generation of semantic meshes when no annotiations
+        // are provided.
+        LOG(WARNING) << ":\n---\n The active scene does not contain semantic "
+                        "annotations. \n---";
       }
     }
   }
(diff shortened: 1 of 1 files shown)


src/esp/gfx/Simulator.cpp
```diff
@@ -101,6 +101,12 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
       throw std::invalid_argument("Cannot load: " + sceneFilename);
     }
 
+    activeSemanticSceneID_ = activeSceneID_;
```
erikwijmans: This is the only line that is needed.  And 132 to 136 can go.

aclegg3:
> This maybe somewhat confusing for people as it'll seem like instance renderings for datasets without them (i.e. Gibson) exist.

The primary use case of this is programmatic assignment of semantics to physics objects and loading those objects into arbitrary or empty scenes. Some use cases may not care about structural semantics and only those for imported objects. 

Maybe we should for now just add the empty scene as a criterion for doing this along with:
```
if (sceneInfo.type == assets::AssetType::SUNCG_SCENE ||
        sceneInfo.type == assets::AssetType::INSTANCE_MESH) 
{ 
activeSemanticSceneID_ = activeSceneID_;
}
```

aclegg3:
> This maybe somewhat confusing for people as it'll seem like instance renderings for datasets without them (i.e. Gibson) exist.

Ok, now this only affects the EMPTY_SCENE.

erikwijmans:
Making it just for the empty scene seems like a band-aid that we will just have to revisit in a couple months when someone wants a scene in the background but wants segmentation of added objects.  Probably enable it for all scenes but add a comment about this and add a logging statement when a scene is loaded without an instance-segmentation mesh?

src/esp/gfx/Simulator.cpp
```diff
@@ -125,7 +125,8 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
 
     // instance meshes and suncg houses contain their semantic annotations
     if (sceneInfo.type == assets::AssetType::SUNCG_SCENE ||
-        sceneInfo.type == assets::AssetType::INSTANCE_MESH) {
+        sceneInfo.type == assets::AssetType::INSTANCE_MESH ||
+        sceneFilename.compare(assets::EMPTY_SCENE) == 0) {
```
mathfac: Let's add TODO to create mesh programmatically in future.

src/esp/gfx/Simulator.cpp
```diff
@@ -121,13 +121,17 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
...
-        sceneInfo.type == assets::AssetType::INSTANCE_MESH ||
-        sceneFilename.compare(assets::EMPTY_SCENE) == 0) {
+    }else{
       activeSemanticSceneID_ = activeSceneID_;
+      // instance meshes and suncg houses contain their semantic annotations
+      // empty scene has none to worry about
+      if ( !(sceneInfo.type == assets::AssetType::SUNCG_SCENE ||
+        sceneInfo.type == assets::AssetType::INSTANCE_MESH ||
+        sceneFilename.compare(assets::EMPTY_SCENE) == 0)) {
+          //TODO: programmatic generation of semantic meshes when no annotiations are provided.
```
erikwijmans: > //TODO: programmatic generation of semantic meshes when no annotiations are provided.

That sounds like a massive research project, not a TODO

PR #286, opened 2019-10-09
Set modelview, projection matrices, viewport respectively from sensor to render camera

bigbike:
## Motivation and Context

Set the camera matrices (model view, proj), viewport **uniformly** from sensor to the render camera;
When people create their own visual sensor (other than pinhole camera), they can inherit from the "Sensor" class, and customize and implement their own version of aforementioned functions;

## How Has This Been Tested

Test it with the example model (the castle):
Screen Shot 2019-10-08 at 4 48 56 PM

## Types of changes

## Checklist

### DIFF
--- a/src/esp/sensor/PinholeCamera.cpp
+++ b/src/esp/sensor/PinholeCamera.cpp
@@ -37,6 +37,10 @@ PinholeCamera& PinholeCamera::setProjectionMatrix(
 
 PinholeCamera& PinholeCamera::setTransformationMatrix(
     gfx::RenderCamera& targetCamera) {
+  CORRADE_ASSERT(!scene::SceneGraph::isRootNode(targetCamera.node()),
+                 "PinholeCamera::setTransformationMatrix: target camera cannot "
+                 "be on the root node of the scene graph",
+                 *this);
   Magnum::Matrix4 absTransform = this->node().absoluteTransformation();
   Magnum::Matrix3 rotation = absTransform.rotationScaling();
   Magnum::Math::Algorithms::gramSchmidtOrthonormalizeInPlace(rotation);
@@ -55,11 +59,12 @@ PinholeCamera& PinholeCamera::setTransformationMatrix(
   // obtain the *absolute* transformation from the sensor node,
   // apply it as the *relative* transformation between the camera and
   // its parent
-  auto parent = targetCamera.node().parent();
-  // if parent is the root node, then skip it!
-  if (parent != nullptr) {
+  auto camParent = targetCamera.node().parent();
+  // if camera's parent is the root node, skip it!
+  if (!scene::SceneGraph::isRootNode(
+          *static_cast<scene::SceneNode*>(camParent))) {
     relativeTransform =
-        parent->absoluteTransformation().inverted() * relativeTransform;
+        camParent->absoluteTransformation().inverted() * relativeTransform;
   }
   targetCamera.node().setTransformation(relativeTransform);
   return *this;
--- a/src/esp/sensor/PinholeCamera.cpp
+++ b/src/esp/sensor/PinholeCamera.cpp
@@ -59,7 +59,7 @@ PinholeCamera& PinholeCamera::setTransformationMatrix(
   // if parent is the root node, then skip it!
   if (parent != nullptr) {
     relativeTransform =
-        parent->absoluteTransformation().inverted() * absTransform;
+        parent->absoluteTransformation().inverted() * relativeTransform;
   }
   targetCamera.node().setTransformation(relativeTransform);
   return *this;
--- a/src/esp/sensor/PinholeCamera.cpp
+++ b/src/esp/sensor/PinholeCamera.cpp
@@ -35,25 +35,33 @@ PinholeCamera& PinholeCamera::setProjectionMatrix(
   return *this;
 }
 
-PinholeCamera& PinholeCamera::setModelViewMatrix(
+PinholeCamera& PinholeCamera::setTransformationMatrix(
     gfx::RenderCamera& targetCamera) {
-  Magnum::Matrix4 T = this->node().absoluteTransformation();
-  Magnum::Matrix3 R = T.rotationScaling();
-  Magnum::Math::Algorithms::gramSchmidtOrthonormalizeInPlace(R);
+  Magnum::Matrix4 absTransform = this->node().absoluteTransformation();
+  Magnum::Matrix3 rotation = absTransform.rotationScaling();
+  Magnum::Math::Algorithms::gramSchmidtOrthonormalizeInPlace(rotation);
 
   VLOG(1) << "||R - GS(R)|| = "
-          << Eigen::Map<mat3f>((R - T.rotationShear()).data()).norm();
+          << Eigen::Map<mat3f>((rotation - absTransform.rotationShear()).data())
+                 .norm();
 
-  T = Magnum::Matrix4::from(R, T.translation()) *
-      Magnum::Matrix4::scaling(T.scaling());
+  auto relativeTransform =
+      Magnum::Matrix4::from(rotation, absTransform.translation()) *
+      Magnum::Matrix4::scaling(absTransform.scaling());
 
   // set the transformation to the camera
   // so that the camera has the correct modelview matrix for rendering;
   // to do it,
   // obtain the *absolute* transformation from the sensor node,
-  // apply it as the *relative* transformation between the default camera and
-  // its parent, which is rootNode_.
-  targetCamera.node().setTransformation(T);
+  // apply it as the *relative* transformation between the camera and
+  // its parent
+  auto parent = targetCamera.node().parent();
+  // if parent is the root node, then skip it!
+  if (parent != nullptr) {
+    relativeTransform =
+        parent->absoluteTransformation().inverted() * absTransform;
+  }
+  targetCamera.node().setTransformation(relativeTransform);
   return *this;
 }
 
--- a/src/esp/sensor/PinholeCamera.cpp
+++ b/src/esp/sensor/PinholeCamera.cpp
@@ -3,8 +3,8 @@
 // LICENSE file in the root directory of this source tree.
 
 #include <Magnum/ImageView.h>
-#include <Magnum/PixelFormat.h>
 #include <Magnum/Math/Algorithms/GramSchmidt.h>
+#include <Magnum/PixelFormat.h>
 
 #include "PinholeCamera.h"
 #include "esp/gfx/DepthUnprojection.h"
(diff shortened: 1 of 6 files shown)


bigbike:
Amanpreet @apsdehal : I added 3 virtual APIs to the base class. Thus you can develop your sub-class and implement these functions to satisfy your needs.

bigbike:
This PR is ready to be reviewed. Thanks!

bigbike:
@msavva : 
Your comments make sense. 
This PR is to come up with some simple solution to quickly unblock Aman. After all, we do not have the hierarchy (e.g., sensor --> visual sensor --> pinhole camera) yet. So I have no choice but to put these functions in the top level (Sensor). It is temporary, and will be changed when we upgrade the Sensor module (We just started a project addressing this issue after the meeting yesterday).

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -28,8 +29,37 @@ void PinholeCamera::setProjectionParameters(SensorSpec::ptr spec) {
...
+PinholeCamera& PinholeCamera::setModelViewMatrix(
+    gfx::RenderCamera& targetCamera) {
+  Magnum::Matrix4 T = this->node().absoluteTransformation();
+  Magnum::Matrix3 R = T.rotationScaling();
+  Magnum::Math::Algorithms::gramSchmidtOrthonormalizeInPlace(R);
+
+  VLOG(1) << "||R - GS(R)|| = "
+          << Eigen::Map<mat3f>((R - T.rotationShear()).data()).norm();
+
+  T = Magnum::Matrix4::from(R, T.translation()) *
```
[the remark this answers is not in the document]
bigbike: I just copied and pasted whatever we had to the new functions.
Good suggestion. I will change them.

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -28,8 +29,37 @@ void PinholeCamera::setProjectionParameters(SensorSpec::ptr spec) {
...
 }
 
-void PinholeCamera::setProjectionMatrix(gfx::RenderCamera& targetCamera) {
+PinholeCamera& PinholeCamera::setProjectionMatrix(
+    gfx::RenderCamera& targetCamera) {
   targetCamera.setProjectionMatrix(width_, height_, near_, far_, hfov_);
+  return *this;
+}
+
+PinholeCamera& PinholeCamera::setModelViewMatrix(
```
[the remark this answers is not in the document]
bigbike: It makes sense. Will change the function names to make them consistent.

Thanks!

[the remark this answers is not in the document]
erikwijmans: Agreed on the naming.  I also think this isn't the modelview matrix?   i.e. the modelview matrix transformations points from model coordinates to camera coordinates, but I believe the matrix here transforms from camera coordinates to model coordiantes.

src/esp/sensor/PinholeCamera.cpp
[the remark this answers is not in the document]
erikwijmans: Yeah... I wrote this code and I go back and forth on uppercase variables names for matrices.  Uppercase is pretty much the universal way to specify a matrix in math/linear algebra, so that is why I do it.

src/esp/sensor/Sensor.h
```diff
@@ -114,8 +114,14 @@ class Sensor : public Magnum::SceneGraph::AbstractFeature3D {
 
   virtual bool isVisualSensor() { return false; }
 
-  // visual sensor should implement and override this function
-  virtual void setProjectionMatrix(gfx::RenderCamera& targetCamera){};
+  // visual sensor should implement and override the following functions
+  virtual Sensor& setProjectionMatrix(gfx::RenderCamera& targetCamera) {
```
erikwijmans: I think it would be better to have the sensor provide all these things via getters rather than giving them the camera and having them set it.  That way people don't have to be concerned with the specifics of how to set the projection matrix and just need to provide it.

bigbike:
> I'm not sure we should support setting view matrix in the generic camera API. By setting the view matrix, you are overriding the scene graph.

The name is causing the confusion. We are not setting the view matrix here. We are setting the model matrix (the absolute transformation matrix) from sensor to the render camera.

> Adding APIs for setting camera matrices seems like a leaky abstraction. It feels like view matrix should really be a derived read-only property if available at all. When I think of modeling a camera, the properties that come to mind are resolution, focal length, field of view, position, and orientation. We should be able to derive the appropriate matrices from those properties. Ideally, the API should be about modeling a real-world camera.

Yes, as a user, she only needs to set a "camera" (actually a sensor) using the most intuitive parameters (e.g., as you mentioned,  resolution, focal length, field of view, position, orientation).
The functions provided in this PR is for the developers, like Aman.

> I think VRCamera is really a special case. I don't think we need to allow setting view matrix for regular camera. I took a quick look at Unity's camera API (https://docs.unity3d.com/ScriptReference/Camera.html) and they only allow setting a custom view matrix for stereo camera. They do not allow setting the view matrix for the regular camera and only allow setting the projection matrix in order to support special cases like water.

Yes, to achieve the same transformation effect, we can convert any change to the view matrix to an operation on the model matrix. (That means, e.g., to take a photo of an object in a particular angle, you either move your camera or your object). 
Here we only change the model matrix. I will update the function name.


> ThreeJS is similar (https://threejs.org/docs/#api/en/cameras/Camera). It only allows setting the projection matrix. Setting the projection matrix is not recommended though. You are expected to use ProjectionCamera or OrthographicCamera which set the matrix for you.

Projection matrix needs to be updated in certain particular cases, like in Aman's project.
We need to provide some functions to satisfy the developer's needs. For example, I may also create an overload function, which has a Matrix4 (the proj matrix itself) as an input parameter.

> For regular (non-VR) camera, setting the projection matrix could make sense (as an advanced feature) in certain cases. But I can't think of any use case for setting the view matrix (unity does not allow it).

Yes, as an advanced feature.
No, we do not set the view matrix, we set the transformation matrix (model matrix) instead.

src/esp/sensor/Sensor.h
bigbike: Such a getter function will return a matrix (e.g., a 4x4 matrix), which is an intermediate value passing to the "render camera". Certainly we can have such getter functions. And I can see such functions can be very useful in some cases. 
But if it is better, I doubt.
I prefer the way used here:
```sensor.setProjectMatrixTo(camera);```
A user treats the function as a "black box" and she does not need to be concerned with the implementation details underlying.

File: src/esp/sensor/PinholeCamera.cpp [hunk 4]
```diff
@@ -28,8 +29,37 @@ void PinholeCamera::setProjectionParameters(SensorSpec::ptr spec) {
...
+  T = Magnum::Matrix4::from(R, T.translation()) *
+      Magnum::Matrix4::scaling(T.scaling());
+
+  // set the transformation to the camera
+  // so that the camera has the correct modelview matrix for rendering;
+  // to do it,
+  // obtain the *absolute* transformation from the sensor node,
+  // apply it as the *relative* transformation between the default camera and
+  // its parent, which is rootNode_.
+  targetCamera.node().setTransformation(T);
```
bigbike: This function has a BUG after refactoring.
The purpose is to set a "render camera" (input parameter) so that it the same "absolute" transformation as the sensor.
But the ```setTransformation``` here will only set the "relative" transformation w.r.t. the parent node. What if the "render camera" is deep in the kinematic chain?

Before, it is not a bug because it was setting the "default render camera" which attaches to the root node of the SG.

Now this new function makes it more generic by passing a regular "render camera", which causes the problem.

src/esp/sensor/Sensor.h
erikwijmans: It depends on what you think the user wants to do.  Does the user want to be able to set the set the projection matrix on a new camera, or does the user want to be able to add a new sensor?  This API makes sense for the former, doing this with a getter makes sense for the latter.

bigbike:
Ran and test it with example.py. It looks normal.

test rgba 00019

src/esp/sensor/Sensor.h
bigbike: @erikwijmans : Yes, certainly we can add new getter functions as you suggested.

src/esp/sensor/PinholeCamera.cpp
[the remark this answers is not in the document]
bigbike: > modelview = view x model;
> Here, the view is an Identity (because the "camera" (sensor) orientation and position is already considered in the model matrix).
> Here, the transformation matrix (model matrix) is to transfer any vertices from the model local frame to the world frame.

@erikwijmans: Actually what I said in the previous message was wrong. I apologize.
Here, we set the transformation matrix (denoted as T) from sensor to the camera. So T is the transformation matrix for the "render camera" now. T will transform from camera frame to the world frame ( you are right at this point). 
The actual view matrix = inverse(T), which will be computed and applied by the magnum library.

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -28,8 +29,50 @@ void PinholeCamera::setProjectionParameters(SensorSpec::ptr spec) {
...
+  // to do it,
+  // obtain the *absolute* transformation from the sensor node,
+  // apply it as the *relative* transformation between the camera and
+  // its parent
+  auto camParent = targetCamera.node().parent();
+  // if camera's parent is the root node, skip it!
+  if (!scene::SceneGraph::isRootNode(
+          *static_cast<scene::SceneNode*>(camParent))) {
+    relativeTransform =
+        camParent->absoluteTransformation().inverted() * relativeTransform;
```
msbaines: LGTM.

src/esp/sensor/PinholeCamera.cpp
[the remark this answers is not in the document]
erikwijmans: Yeah, this is what I was originally thinking and wonder if Aman had run into this, but apparently webVR calls the modelview matrix the matrix that transforms from camera coordinates into the world frame, so who knows if the convention I am use to is actually universal or not.

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -28,8 +29,50 @@ void PinholeCamera::setProjectionParameters(SensorSpec::ptr spec) {
...
+  return *this;
+}
+
+PinholeCamera& PinholeCamera::setTransformationMatrix(
+    gfx::RenderCamera& targetCamera) {
+  CORRADE_ASSERT(!scene::SceneGraph::isRootNode(targetCamera.node()),
+                 "PinholeCamera::setTransformationMatrix: target camera cannot "
+                 "be on the root node of the scene graph",
+                 *this);
+  Magnum::Matrix4 absTransform = this->node().absoluteTransformation();
```
msbaines: @bigbike explained this logic to me offline but it wasn't initially obvious to me what was going on here. It all just seem like a no-op. But @bigbike explained that we are correcting for the fact that numerical errors in the rotation basis don't apply evenly so we need to correct for this by regenerating the basis.

Would be nice to have a comment to this affect with perhaps a reference citation. Its always nice when fancy math comes with a citation.

Issue #283, opened 2019-10-08
Reconfigure does not work with objects if image types changes


dhuber23b:
## 🐛 Bug

You can add objects to the scene using the physics library capability in Habitat.  If you initially set the image types to none and configure the simulator and then later add new image types when you call reconfigure, the object is not added (or at least it is not displayed).

## To Reproduce

Attaching a zip file with code that reproduces the problem.  It uses the replica data, which is too large to attach, but the problem should be reproducible with any simple scene.

## System Info
ENVIRONMENT INFO:
Platform: Darwin-18.7.0-x86_64-i386-64bit
Machine: x86_64
Processor: i386
Libc version:
Mac version: 10.14.6
Python version: 3.6.2+
Architecture: 64bit
Win version:
System OS: Darwin
Release: 18.7.0
Version: Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64
Operational System: darwin
GCC version: Apple clang version 11.0.0 (clang-1100.0.33.5)
CMAKE version: cmake version 3.15.4
NVIDIA-SMI:
Pip packages versions:

Conda packages versions:

[reconfigure_bug.zip](https://github.com/facebookresearch/habitat-sim/files/3704067/reconfigure_bug.zip)

aclegg3:
Hey Daniel, can you tell me what error output you get? It may be an issue with semantics for objects added through the physicsManager. If I initialize the simulator with no image types `{}` then ask for `{"color":True, "depth":True}` and reconfigure, it works fine.

dhuber23b:
Alex, 

There is no error message.  It just does not show the object.  Maybe I am not setting it up correctly.  You can see the steps I use in the sample code.  Below is the text of the main script that reproduces it:

from collections import namedtuple
import habitat_sim
from habitat_sim.utils import common
from habitat_sim.utils.common import d3_40_colors_rgb
import matplotlib.pyplot as plt
import numpy as np
import os
import random
import shutil
from PIL import Image
from typing import List

Pose = namedtuple('Pose', ['translation', 'quaternion'])

def make_habitat_cfg(settings, scene_file, image_types):
    sim_cfg = habitat_sim.SimulatorConfiguration()
    sim_cfg.gpu_device_id = 0
    sim_cfg.scene.id = scene_file
    sim_cfg.enable_physics = True
[...]

PR #287, opened 2019-10-09
Simulator::addObject should add to the active scene graph

aclegg3:
## Motivation and Context

In response to issue #283.

`Simulator::addObject` takes a `sceneID` parameter for multi-scene use case. Since we don't support multi-scene physics presently, this should always add to the active scene graph instead.

This will need to be changed back in the future as we support multi-scene simulation.

## How Has This Been Tested

Local testing with multiple `sim.reconfigure()` to create several sceneIDs in the Simulator.

## Types of changes

## Checklist


erikwijmans:
IMO, it would make more sense to just set the default value of `sceneID` to `activeSceneID_`.

aclegg3:
> IMO, it would make more sense to just set the default value of `sceneID` to `activeSceneID_`.

Good thought, but there are multiple levels of default value = 0 as the function is passed through python bindings and into the simulator.py.

PR #293, opened 2019-10-11
Default to random start position and allow URI decoding

apsdehal:
## Motivation and Context

Our default start position has been hardcoded for castle scene. We need to rely back on SimulatorWithAgents to set a default position for us. You can reset back to default episode by adding `&useDefaultEpisode=1` to the back of your url.

This allows van-gogh and replica apartment to be rendered inside web browser.

Other than that some of the link sharing websites URI encode the URLs so they should be decoded back on our end.

## How Has This Been Tested

Tests have been added for the changes that were made in this PR. And as mentioned in chat on the URLs shared there.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/index.js
+++ b/src/esp/bindings_js/index.js
@@ -12,7 +12,8 @@ import "./bindings.css";
 import {
   checkWebAssemblySupport,
   checkWebgl2Support,
-  getInfoSemanticUrl
+  getInfoSemanticUrl,
+  buildConfigFromURLParameters
 } from "./modules/utils";
 
 function preload(url) {
@@ -28,12 +29,7 @@ function preload(url) {
 Module.preRun.push(() => {
   let config = {};
   config.scene = defaultScene;
-  for (let arg of window.location.search.substr(1).split("&")) {
-    let [key, value] = arg.split("=");
-    if (key && value) {
-      config[key] = value;
-    }
-  }
+  buildConfigFromURLParameters(config);
   const scene = config.scene;
   Module.scene = preload(scene);
   const fileNoExtension = scene.substr(0, scene.lastIndexOf("."));
@@ -62,6 +58,7 @@ Module.onRuntimeInitialized = () => {
   if (!demo) {
     demo = new WebDemo();
   }
+
   demo.display();
 };
 
--- a/src/esp/bindings_js/modules/simenv_embind.js
+++ b/src/esp/bindings_js/modules/simenv_embind.js
@@ -19,10 +19,14 @@ class SimEnv {
    * @param {Object} episode - episode to run
    * @param {number} agentId - default agent id
    */
-  constructor(config, episode, agentId) {
+  constructor(config, episode = {}, agentId = 0) {
     this.sim = new Module.Simulator(config);
     this.episode = episode;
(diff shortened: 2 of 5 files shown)


src/esp/bindings_js/index.js
```diff
@@ -28,12 +29,7 @@ function preload(url) {
 Module.preRun.push(() => {
   let config = {};
   config.scene = defaultScene;
-  for (let arg of window.location.search.substr(1).split("&")) {
-    let [key, value] = arg.split("=");
-    if (key && value) {
-      config[key] = value;
-    }
-  }
+  buildConfigFromURLParameters(config);
```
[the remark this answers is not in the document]
apsdehal: It doesn't propagate everything passed in URL parameters yet. I agree with you that it is an important feature. Let me add that change in this PR as well.

src/esp/bindings_js/modules/simenv_embind.js
```diff
@@ -19,10 +19,14 @@ class SimEnv {
    * @param {Object} episode - episode to run
    * @param {number} agentId - default agent id
    */
-  constructor(config, episode, agentId) {
+  constructor(config, episode = {}, agentId = 0) {
```
msbaines: Why not just use null as the default value? That way you would only have to check for falsiness.

src/esp/bindings_js/modules/simenv_embind.js
```diff
@@ -31,8 +35,10 @@ class SimEnv {
    */
   reset() {
     this.sim.reset();
-    const agent = this.sim.getAgent(this.selectedAgentId);
-    agent.setState(this.initialAgentState, true);
+    if (this.initialAgentState !== null) {
```
msbaines: Why not:

`if (this.initialAgentState) {
`

Seems like it would be more robust and simpler.

apsdehal: I am generally wary of comparing by using truthy value of the object itself. It can mean a lot of things in JavaScript. That's why I like to be specific to what I am compare to.

src/esp/bindings_js/modules/simenv_embind.js
apsdehal: I was thinking of future changes. Episode also contains goal, so it might not be sufficient to set it to null yet. So, setting it to object allows us to assume there will always be an object and then we can do object level compares.

apsdehal: For e.g. I won't be able to do `episode.goal` without check if episode is null. (That would be one extra check if default value is not `{}`)

src/esp/bindings_js/index.js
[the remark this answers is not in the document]
apsdehal: I talked with Mandeep and since we are using WindowlessEmscriptenApplication.js, logs are going into pre element with id log. It should be the right time to go ahead and move to EmscriptenApplication.js. I will make that change in separate PR. (If you want to debug, logs should be there in the pre element which is hidden).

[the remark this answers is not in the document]
mosra: Ah, that's why :) Thank you.

PR #304, opened 2019-10-16
Blit from frameBuffer to defaultFrameBuffer viewport

apsdehal:
## Motivation and Context
- Before, we used to blit from same source rectangle to same destination
rectangle
- Instead, we should blit from source framebuffer viewport to default
frame buffer's viewport
- Also, instead of hardcoded source rectangle, we will take the source rectangle from `frameBuffer_.viewport`

## How Has This Been Tested

Tested on viewer.html and webvr.html

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -117,8 +117,10 @@ struct RenderTarget::Impl {
   void blitRgbaToDefault() {
     framebuffer_.mapForRead(RgbaBuffer);
     GL::AbstractFramebuffer::blit(framebuffer_, GL::defaultFramebuffer,
-                                  {{}, framebufferSize()},
-                                  GL::FramebufferBlit::Color);
+                                  framebuffer_.viewport(),
+                                  GL::defaultFramebuffer.viewport(),
+                                  GL::FramebufferBlit::Color,
+                                  GL::FramebufferBlitFilter::Nearest);
   }
 
   void readFrameRgba(const MutableImageView2D& view) {
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -116,6 +116,8 @@ struct RenderTarget::Impl {
 
   void blitRgbaToDefault() {
     framebuffer_.mapForRead(RgbaBuffer);
+    ASSERT(framebuffer_.viewport() == GL::defaultFramebuffer.viewport());
+
     GL::AbstractFramebuffer::blit(
         framebuffer_, GL::defaultFramebuffer, framebuffer_.viewport(),
         GL::defaultFramebuffer.viewport(), GL::FramebufferBlit::Color,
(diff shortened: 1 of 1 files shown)


src/esp/gfx/RenderTarget.cpp
```diff
@@ -116,9 +116,10 @@ struct RenderTarget::Impl {
 
   void blitRgbaToDefault() {
     framebuffer_.mapForRead(RgbaBuffer);
-    GL::AbstractFramebuffer::blit(framebuffer_, GL::defaultFramebuffer,
-                                  {{}, framebufferSize()},
-                                  GL::FramebufferBlit::Color);
+    GL::AbstractFramebuffer::blit(
+        framebuffer_, GL::defaultFramebuffer, framebuffer_.viewport(),
+        GL::defaultFramebuffer.viewport(), GL::FramebufferBlit::Color,
+        GL::FramebufferBlitFilter::Nearest);
```
mosra: `Linear`? Or it looks *too* bad? :) Btw., how can it happen that the framebuffers have a different size? I'd assume the `framebuffer_` gets created from default FB's size and then resized as appropriate on webgl canvas / window resize (application's `viewportEvent()`).

apsdehal: I can check Linear. Nearest was the default. Thanks.

About different size of frame buffers: They still have same size but the viewport can have different start and end points, so I am just taking that in consideration. For WebVR, first paint has defaultFrameBuffer's viewport as `{{0, 0}, {halfWidth, height}}`, and the second paint will have it as `{{halfWidth, 0}, {width, height}}`. So, source rectangle stays same as whatever was the viewport of the `framebuffer_` and destination rectangle would be the current viewport of the defaultFrameBuffer.

mosra: Ah, I see :) If the size is the same, then `Nearest` makes sense, yup ... but otherwise it might look quite bad, depending on the ratio. So maybe instead add an "explanatory assert" that the sizes are equal?

msbaines: Maybe an assertion to confirm that source and destination are the same size to verify that no interpolation is happening. The interpolation method only matters if the sizes are different.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -116,6 +116,8 @@ struct RenderTarget::Impl {
 
   void blitRgbaToDefault() {
     framebuffer_.mapForRead(RgbaBuffer);
+    ASSERT(framebuffer_.viewport() == GL::defaultFramebuffer.viewport());
```
mosra: :thinking: I think you wanted to do `framebuffer_.viewport().size() == GL::defaultFramebuffer.viewport().size()`, right? Because this will blow up if the offsets are different (which I think they are, in case of stereo rendering).

apsdehal: That's right. Let me fix it in a separate PR.

PR #305, opened 2019-10-16
PhysicsManager CHECK cleanup

aclegg3:
## Motivation and Context

This PR replaces if statements in PhysicsManager and BulletPhysicsManager based on the validity of object IDs with CHECK().

As @msbaines mentioned in a review for PR #237, PhysicsManager functions taking an object ID as input should not fail silently when an invalid object ID is provided. This may allow hidden bugs to crop up. For example: setting a force or torque on an object which does not exist and then receiving no indication that the set failed resulting in extended debugging. Instead, such an invalid set should be fatal and produce a clear error message. 

## How Has This Been Tested

Local testing.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/PhysicsManager.cpp
+++ b/src/esp/physics/PhysicsManager.cpp
@@ -84,7 +84,7 @@ int PhysicsManager::addObject(const std::string& configFile,
 }
 
 int PhysicsManager::removeObject(const int physObjectID) {
-  CHECK(existingObjects_.count(physObjectID) > 0);
+  assertIDValidity(physObjectID);
   existingObjects_.at(physObjectID)->removeObject();
   delete existingObjects_.at(physObjectID);
   existingObjects_.erase(physObjectID);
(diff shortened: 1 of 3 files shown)


src/esp/physics/PhysicsManager.cpp
```diff
@@ -84,10 +84,7 @@ int PhysicsManager::addObject(const std::string& configFile,
 }
 
 int PhysicsManager::removeObject(const int physObjectID) {
-  if (existingObjects_.count(physObjectID) == 0) {
-    LOG(ERROR) << "Failed to remove object: no object with ID " << physObjectID;
-    return ID_UNDEFINED;
-  }
+  assertIDValidity(physObjectID);
```
msbaines: Should the object destructor call removeObject()? That way you never forget to call it.
delete should be avoid and RTTI used instead. Could we use unique_ptr instead here? The last three lines would then boil down to:

existingObjects_.erase(physObjectID);

aclegg3: This comes back to the SceneNode problem. SceneNode objects are owned completely by the SceneGraph. I have not gotten around to converting RigidObject from a SceneNode to an AbstractFeature3D as suggested in reviews for PR #91 . We use raw pointers to denote that PhysicsManager does not own the RigidObject. 

That said, the removal process could probably be improved. I'll lump that in with the AbstractFeature3D re-design.

msbaines: If we don't own the object, why are we calling delete on it below on line 89?

aclegg3: We have to remove the SceneNode from the SceneGraph when the object no longer exists. I agree the current situation is not entirely self-consistent. The simulator was not initially designed for this sort of dynamic restructuring.

PR #307, opened 2019-10-16
[esp/bindings_js] add initial support for object nav

msbaines:
Object and best-view IoU is specified by a URL parameter. If the
user presses the spacebar and IoU of object exceeds 50% of best-view
the task is successfully completed.

## Motivation and Context

Required for object nav.

## How Has This Been Tested

Verified that object nav is working.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings_js/modules/object_sensor.js
+++ b/src/esp/bindings_js/modules/object_sensor.js
@@ -33,19 +33,11 @@ class ObjectSensor {
       }
     }
     for (let i = 0; i < objects.size(); i++) {
-      const object = objects.get(i);
-      if (
-        object &&
-        object.category &&
-        object.category.getIndex("") === categoryIndex
-      ) {
+      const object = objects.get(i) || {};
+      if (object.category && object.category.getIndex("") === categoryIndex) {
         this.objectIds.push(i);
       }
     }
-
-    console.log(this.objectIds);
-    console.log(semanticShape);
-    console.log("looking for " + objectClass);
   }
 
   /**
--- a/src/esp/bindings_js/modules/object_sensor.js
+++ b/src/esp/bindings_js/modules/object_sensor.js
@@ -1,0 +1,87 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+/**
+ * ObjectSensor class
+ */
+class ObjectSensor {
+  // PUBLIC methods.
+
+  /**
+   * Create a scope for object navigation task.
+   * @param {Object} searchRect - rect for Intersection-over-Union
+   * @param {Object} semanticShape - shape of semantic observation
+   * @param {SemanticScene} semanticScene - semantic information for scene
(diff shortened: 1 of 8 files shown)


src/esp/bindings_js/modules/object_sensor.js
```diff
@@ -0,0 +1,87 @@
...
+      if (
+        object &&
+        object.category &&
+        object.category.getIndex("") === categoryIndex
+      ) {
+        this.objectIds.push(i);
+      }
+    }
+
+    console.log(this.objectIds);
```
apsdehal: Remove?

src/esp/bindings_js/modules/object_sensor.js
```diff
@@ -0,0 +1,87 @@
...
+    let categoryIndex = -1;
+    for (let i = 0; i < categories.size(); i++) {
+      const category = categories.get(i);
+      if (category && category.getName("") === objectClass) {
+        categoryIndex = category.getIndex("");
+        break;
+      }
+    }
+    for (let i = 0; i < objects.size(); i++) {
+      const object = objects.get(i);
```
apsdehal:
```suggestion
      const object = objects.get(i) || {};
```

src/esp/bindings_js/modules/object_sensor.js
```diff
@@ -0,0 +1,87 @@
...
+      if (
+        object &&
```
apsdehal:
```suggestion
```

src/esp/bindings_js/modules/object_sensor.js
```diff
@@ -0,0 +1,87 @@
...
+    console.log("looking for " + objectClass);
+  }
+
+  /**
+   * Compute IOU for for a given class.
+   * @param {Observation} semanticObservation - observation to search
+   * @param {String} className - object class to search for
+   */
+  computeIOU(observation) {
+    const width = this.semanticShape.get(1);
```
apsdehal: Move to utils.

apsdehal: Might not make sense to move to utils with objectIds check below.

src/esp/bindings_js/modules/object_sensor.js
```diff
@@ -0,0 +1,87 @@
...
+        if (this.objectIds.includes(observation[j * width + i])) {
+          if (inSearchArea && i >= left && i < right) {
+            maskPixelsIn++;
+          } else {
+            maskPixelsOut++;
+          }
+        }
+      }
+    }
+    if (maskPixelsIn === 0) {
```
apsdehal: Won't this will be automatically captured in statement below. For speed purposes, checking this in the start would be better without even iterating.

msbaines: This value is computed in the loop.

apsdehal: I meant this check is not needed as without this check it will still return zero.

I thought you were saving on rectArea calculation. If that was the case, this can be precomputed by checking top, left, bottom, right of both triangles beforehand.

Issue #301, opened 2019-10-15
Bullet collision objects for rigid bodies with mesh hierarchies are incorrect.


aclegg3:
## 🐛 Bug

The current implementation creates rigid objects from loaded meshes by directly indexing into asset memory. Meshes are shifted to align object center of mass with origin (as required by Bullet physics) and then consumed directly to create collision objects. This works for single mesh objects (such as our test dataset). However, if the loaded object has a hierarchy of meshes including transforms (such as the Wayfair dataset), we should build a compound collision object encapsulating this hierarchy. 

Note: rendering and kinematics work, this bug only concerns the underlying physical collision mesh used for simulation.

## Command

`./build/viewer --enable-physics --debug-bullet -- "NONE" `
## To Reproduce

1. <del> Pull the `bullet-debug-draw` branch (to get Bullet debug visualization) #298 </del> 
 Edit: on `master` now.
2. Add objects (such as wayfair objects) with a mesh transformation hierarchy to the `default.phys_scene_config.json` file.
2. Modify viewer.cpp to set objects to KINEMATIC upon creation:
```
void Viewer::addObject(std::string configFile) {
...
    physicsManager_->setObjectMotionType(physObjectID, physics::MotionType::KINEMATIC);
...
}
```
3. `./build/viewer --enable-physics --debug-bullet -- "NONE"`
4. Add objects with 'o' and remove with 'u'

Results show bullet collision meshes (white wireframes) do not match rendered objects:
![bullet_hierarchal_mesh_bug](https://user-images.githubusercontent.com/1445143/66873659-d8cfba00-ef5d-11e9-8123-042abb6e3ea0.gif)

## Expected behavior

Collision object meshes should align with rendered object meshes.

## Additional context

Relevant redesign and solution is in progress.

## System Info

Please copy and paste the output from the
[environment collection script](https://github.com/facebookresearch/habitat-sim/tree/master/habitat_sim/utils/collect_env.py)
(or fill out the checklist below manually).

You can run the script with:
```
# For security purposes, please check the contents of collect_env.py before running it.
python habitat_sim/utils/collect_env.py
```

 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - GPU models and configuration:
 - CUDA version:
 - PIP/conda dependencies packages versions
 - Any other relevant information:

mosra:
[...]

PR #309, opened 2019-10-16
Store mesh transform hierarchy (Issue #301 fix part 1/2)

aclegg3:
## Motivation and Context

This is the first step toward a solution for Issue #301 .

In order to build collision objects from mesh hierarchies, we first need to store them. 

Since the hierarchies are metadata, this PR adds a `MeshTransformNode` data structure to the `MeshMetaData` structure. When a mesh is loaded into assets the hierarchy is computed and stored.

Now that the hierarchy of transforms is available, we no longer need Importer to reload the source files in order to add objects to the scene graph with `addComponent`. This has been refactored.

No previously exposed functionality should change as a result of this refactor.

## How Has This Been Tested

Local testing with physics.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/MeshMetaData.h
+++ b/src/esp/assets/MeshMetaData.h
@@ -5,9 +5,45 @@
 #pragma once
 
 #include "esp/core/esp.h"
+#include "esp/gfx/magnum.h"
 
 namespace esp {
 namespace assets {
+
+//! Some mesh files include a transformation hierarchy. A @ref MeshTransformNode
+//! stores this hierarchy and indices for the meshes at each level such that it
+//! can be reused to instances meshes later.
+struct MeshTransformNode {
+  //! mesh ID within @ref MeshMetaData::meshIndex
+  int meshIDLocal;
+
+  //! material ID within @ref MeshMetaData::materialIndex
+  int materialIDLocal;
+
+  //! Object ID in the original file
+  int componentID;
+
+  std::vector<MeshTransformNode> children;
+
+  //! Node local transform
+  Magnum::Matrix4 T;
+
+  MeshTransformNode() {
+    meshIDLocal = ID_UNDEFINED;
+    materialIDLocal = ID_UNDEFINED;
+    componentID = ID_UNDEFINED;
+  };
+
+  MeshTransformNode(const MeshTransformNode& val) {
+    componentID = val.componentID;
+    meshIDLocal = val.meshIDLocal;
+    materialIDLocal = val.materialIDLocal;
+    T = Magnum::Matrix4(val.T);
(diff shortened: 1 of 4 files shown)


src/esp/assets/MeshMetaData.h
```diff
@@ -5,10 +5,46 @@
...
+  //! material ID within @ref MeshMetaData::materialIndex
+  int materialIDLocal;
+
+  //! Object ID in the original file
+  int componentID;
+
+  std::vector<MeshTransformNode> children;
+
+  //! Node local transform
+  Magnum::Matrix4 T;
```
erikwijmans: Let's name this more explicitly.  I like the notation of `T_<to>_<from>`, so if this goes matrix that transforms things world to local coordinates, it would be `T_local_world`

I like this notation because you can write `pt_local = T_local_world * pt_world` and it's clear what happens because the `world`'s "cancel" out and you left with just `local`.

This matrix is likely `T_world_local` in this notation.

aclegg3: Sure, this sounds reasonable. In this notation `T_parent_local` may be more accurate. 👍

src/esp/assets/MeshMetaData.h
```diff
@@ -5,10 +5,48 @@
...
+    componentID = ID_UNDEFINED;
+  };
+
+  //! copy constructor which duplicates the @ref MeshTransformNode tree of which
+  //! val is the root.
+  MeshTransformNode(const MeshTransformNode& val) {
+    componentID = val.componentID;
+    meshIDLocal = val.meshIDLocal;
+    materialIDLocal = val.materialIDLocal;
+    T_parent_local = Magnum::Matrix4(val.T_parent_local);
```
mosra: What's the reason to use `snake_case` with first letter upper-case? To me that's quite confusing when reading the code.

erikwijmans: It's not really `snake_case`, it is to denote subscripting like in latex.   I proposed this naming scheme for transforms as I really like it :) It is `T_<to>_<from>` as you can then write something like `pt_<to> = T_<to>_<from> pt_<from>` and you can think of it as the `<from>`s "canceling" out.  Same thing with `T_<a>_<c> = T_<a>_<b> * T_<b>_<c>`.  But it does break normal code style.

erikwijmans: Yep, `transformFromLocalToParent` is also fine by me :) -- I use `T_parent_local` and have grown to really like it (although I do admit remembering the notation took a couple tries), but it is merely a suggestion and is localized enough that we can change it!

PR #310, opened 2019-10-16
Bullet collision compound from hierarchy of meshes (Issue #301 fix part 2/2)

aclegg3:
## Motivation and Context

Now that the mesh hierarchy is stored (PR #309) , this PR exposes necessary assets and enables `BulletRigidObject` to construct a [btCompoundShape ](https://pybullet.org/Bullet/BulletFull/classbtCompoundShape.html) from a number of [btConvexHullShapes](https://pybullet.org/Bullet/BulletFull/classbtConvexHullShape.html) (for `RigidObjectType::OBJECT`) or [btBvhTriangleMeshShape](https://pybullet.org/Bullet/BulletFull/classbtBvhTriangleMeshShape.html) (for `RigidObjectType::SCENE`) with transforms accumulated from the stored hierarchy.

Mesh shifting, the previous method of aligning single mesh object origins with center of mass (COM) has been removed and replaced with a `SceneNode` shifting approach which generalizes to mesh hierarchies.

Note that these changes do break some previous assumptions and an update to the provided test object dataset `http://dl.fbaipublicfiles.com/habitat/objects_v0.1.zip` will be required upon merge. Specifically, `.glb` objects are now loaded with their global frame transforms properly (this was a hidden bug before), so collision shapes exported before as '.obj' files for `cheezit.glb` and `chefcan.glb` match the erroneously un-rotated meshes. Newly exported '.glb' versions with matching local_to_world rotation frames must be uploaded to remedy this.

## How Has This Been Tested

Local testing with Bullet debugDraw:

### Before (Issue #301 ):
bullet_hierarchal_mesh_bug

### After:
fixed_bullet_heirarchies

### Before refactoring mesh shift to play nice with hierarchal meshes we saw strange physics behavior for these objects:
heirarchal_COM_bug

### After:
chair_drop_bugfix

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -450,6 +450,9 @@ int ResourceManager::loadObject(const std::string& objPhysConfigFilename) {
         }
       }
       physicsObjectAttributes.setMagnumVec3("COM", COM);
+      // set a flag which we can find later so we don't override the desired COM
+      // with BB center.
+      physicsObjectAttributes.setBool("COM_provided", true);
     }
   }
 
@@ -524,26 +527,10 @@ int ResourceManager::loadObject(const std::string& objPhysConfigFilename) {
   AssetInfo renderMeshinfo;
   AssetInfo collisionMeshinfo;
 
-  Magnum::Vector3 COM = physicsObjectAttributes.getMagnumVec3("COM");
-  bool shiftMeshOrigin = !(COM[0] == 0 && COM[1] == 0 && COM[2] == 0);
-
   //! Load rendering mesh
-  // TODO: unify the COM move: don't want
-  // simplified meshes to result in different rendering and collision
-  // COMs/origins
   if (!renderMeshFilename.empty()) {
     renderMeshinfo = assets::AssetInfo::fromPath(renderMeshFilename);
-
-    if (shouldComputeMeshBBCenter) {  // compute the COM from BB center
-      renderMeshSuccess =
-          loadGeneralMeshData(renderMeshinfo, nullptr, nullptr, true);
-    } else if (shiftMeshOrigin) {  // use the provided COM
-      renderMeshSuccess =
-          loadGeneralMeshData(renderMeshinfo, nullptr, nullptr, true,
-                              -physicsObjectAttributes.getMagnumVec3("COM"));
-    } else {  // mesh origin already at COM
-      renderMeshSuccess = loadGeneralMeshData(renderMeshinfo);
-    }
+    renderMeshSuccess = loadGeneralMeshData(renderMeshinfo);
     if (!renderMeshSuccess) {
       LOG(ERROR) << "Failed to load a physical object's render mesh: "
                  << objPhysConfigFilename << ", " << renderMeshFilename;
@@ -552,26 +539,12 @@ int ResourceManager::loadObject(const std::string& objPhysConfigFilename) {
   //! Load collision mesh
   if (!collisionMeshFilename.empty()) {
     collisionMeshinfo = assets::AssetInfo::fromPath(collisionMeshFilename);
-    if (shouldComputeMeshBBCenter) {  // compute the COM from BB center
-      collisionMeshSuccess =
-          loadGeneralMeshData(collisionMeshinfo, nullptr, nullptr, true);
-    } else if (shiftMeshOrigin) {  // use the provided COM
-      collisionMeshSuccess =
-          loadGeneralMeshData(collisionMeshinfo, nullptr, nullptr, true,
-                              -physicsObjectAttributes.getMagnumVec3("COM"));
-    } else {  // mesh origin already at COM
-      collisionMeshSuccess = loadGeneralMeshData(collisionMeshinfo);
-    }
+    collisionMeshSuccess = loadGeneralMeshData(collisionMeshinfo);
     if (!collisionMeshSuccess) {
       LOG(ERROR) << "Failed to load a physical object's collision mesh: "
                  << objPhysConfigFilename << ", " << collisionMeshFilename;
     }
   }
-
-  // NOTE: if we want to save these after edit we need to save the moved
-  // mesh or save the original COM as a member of RigidBody...
-  physicsObjectAttributes.setMagnumVec3("COM", Magnum::Vector3(0));
-  // once we move the meshes, the COM is aligned with the origin...
 
   if (!renderMeshSuccess && !collisionMeshSuccess) {
     // we only allow objects with SOME mesh file. Failing
@@ -608,8 +581,6 @@ int ResourceManager::loadObject(const std::string& objPhysConfigFilename) {
     CollisionMeshData& meshData = gltfMeshData->getCollisionMeshData();
     meshGroup.push_back(meshData);
   }
-  //! Properly align axis direction
-  // NOTE: this breaks the collision properties of some files
   collisionMeshGroups_.emplace(objPhysConfigFilename, meshGroup);
   physicsObjectConfigList_.push_back(objPhysConfigFilename);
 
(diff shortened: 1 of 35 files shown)


src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -62,15 +64,34 @@ class BulletRigidObject : public RigidObject {
...
+   * in a flat manner for efficiency.
+   * @param bCompound The @ref btCompoundShape being constructed.
+   * @param T The cumulative local-to-world transformation matrix constructed by
+   * composition down the @ref MeshTransformNode tree to the current node.
+   * @param meshGroup Access structure for collision mesh data.
+   * @param node The current @ref MeshTransformNode in the recursion.
+   */
+  void constructBulletConvexCompoundFromMeshes(
+      std::unique_ptr<btCompoundShape>& bCompound,
+      Magnum::Matrix4& T,
```
erikwijmans: Same request about matrix naming :)

erikwijmans: Also, should this be a const reference?  If not, this should be a return value.

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -103,11 +103,41 @@ bool BulletRigidObject::initializeScene(
...
 
+// recursively create the convex mesh shapes and add them to the compound in a
+// flat manner by accumulating transformations down the tree
+void BulletRigidObject::constructBulletConvexCompoundFromMeshes(
+    std::unique_ptr<btCompoundShape>& bCompound,
+    Magnum::Matrix4& T,
+    const std::vector<assets::CollisionMeshData>& meshGroup,
+    const assets::MeshTransformNode& node) {
+  Magnum::Matrix4 cT = T * node.T_parent_local;
+  btTransform t(cT);
```
erikwijmans:
```suggestion
```

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -103,11 +103,41 @@ bool BulletRigidObject::initializeScene(
...
 }  // end BulletRigidObject::initializeScene
 
+// recursively create the convex mesh shapes and add them to the compound in a
+// flat manner by accumulating transformations down the tree
+void BulletRigidObject::constructBulletConvexCompoundFromMeshes(
+    std::unique_ptr<btCompoundShape>& bCompound,
+    Magnum::Matrix4& T,
+    const std::vector<assets::CollisionMeshData>& meshGroup,
+    const assets::MeshTransformNode& node) {
+  Magnum::Matrix4 cT = T * node.T_parent_local;
```
erikwijmans: Assuming that `T` is `T_world_parent` (seems like it is from the doc string), then `cT` is `T_world_local`?

src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -62,15 +64,34 @@ class BulletRigidObject : public RigidObject {
...
+   * @param meshGroup The collision mesh data for the object.
    * @return true if initialized successfully, false otherwise.
    */
   bool initializeObject(
       const assets::PhysicsObjectAttributes& physicsObjectAttributes,
+      std::shared_ptr<btDiscreteDynamicsWorld> bWorld,
+      const assets::MeshMetaData& metaData,
+      const std::vector<assets::CollisionMeshData>& meshGroup);
+
+  /**
```
erikwijmans: Yay!  Thank you for the doc string.

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -126,36 +156,10 @@ bool BulletRigidObject::initializeObject(
   //! Iterate through all mesh components for one object
   //! The components are combined into a convex compound shape
   bObjectShape_ = std::make_unique<btCompoundShape>();
-  for (const assets::CollisionMeshData& meshData : meshGroup) {
-    Corrade::Containers::ArrayView<Magnum::Vector3> v_data = meshData.positions;
-    Corrade::Containers::ArrayView<Magnum::UnsignedInt> ui_data =
-        meshData.indices;
+  Magnum::Matrix4 T;  // TODO: this should translate to the COM
```
erikwijmans: Doesn't look like this is used elsewhere, so deleted and do
```
constructBulletConvexCompoundFromMeshes(bObjectShape_, Magnum::Matrix4{Magnum::Math::IdentityInit}, meshGroup,
metaData.root);
```

Looks like Magnum does the right thing, (thank you @mosra) and `Magnum::Matrix4{}` would suffice :)

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -103,11 +103,41 @@ bool BulletRigidObject::initializeScene(
...
+
+    bObjectConvexShapes_.emplace_back(std::make_unique<btConvexHullShape>(
+        static_cast<const btScalar*>(mesh.positions.data()->data()),
+        mesh.positions.size(), sizeof(Magnum::Vector3)));
+
+    // TODO: need this for child shapes?
+    // bObjectConvexShapes_.back()->setMargin(margin);
+
+    //! Add to compound shape stucture
+    bObjectShape_->addChildShape(t, bObjectConvexShapes_.back().get());
```
erikwijmans:
```suggestion
    bObjectShape_->addChildShape(btTransform{<whatever the name is>}, bObjectConvexShapes_.back().get());
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -745,6 +747,10 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
         scene::SceneNode& node = parent->createChild();
         const quatf transform = info.frame.rotationFrameToWorld();
         node.setRotation(Magnum::Quaternion(transform));
+        Magnum::Matrix4 R = Magnum::Matrix4::from(
+            Magnum::Quaternion(transform).toMatrix(), Magnum::Vector3());
+        resourceDict_[filename].root.T_parent_local =
+            R * resourceDict_[filename].root.T_parent_local;
```
aclegg3: Is there is nicer way to construct this Transform from an Eigen `quatf`?

aclegg3:
CI tests are failing on frame equality assertions for test_sensors. Now that I am storing the transformation for each component of the loaded file, I'm rolling the `info.frame.rotationFrameToWorld()` transform into the root of the hierarchy rather than its parent. This likely breaks the expected `render_state` used for comparison in CI since the transform will have moved down the tree. Open to thoughts here. I could save that transformation elsewhere like `MeshMetaData::T_world_local` and convert things to expect this variable. Thoughts?

src/esp/assets/ResourceManager.cpp
```diff
@@ -886,7 +894,9 @@ bool ResourceManager::loadGeneralMeshData(
     }  // forceReload
 
     const quatf transform = info.frame.rotationFrameToWorld();
-    newNode.setRotation(Magnum::Quaternion(transform));
```
erikwijmans: Why was this deleted.

src/esp/assets/ResourceManager.cpp
```diff
@@ -745,6 +747,10 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
         scene::SceneNode& node = parent->createChild();
         const quatf transform = info.frame.rotationFrameToWorld();
         node.setRotation(Magnum::Quaternion(transform));
```
erikwijmans: But this was kept?

erikwijmans:
@aclegg3 -- I think I see the bug.  The bug has to due to loading a scene, swapping scenes, and then reloading that scene.  When you reload a scene, all the assets get cached and the `R` will end up getting applied multiple times in `metaData.root.T_parent_local = R * metaData.root.T_parent_local;` 

If you move the `metaData.root.T_parent_local = R * metaData.root.T_parent_local;` step to only be done when the file isn't loaded, everything passes.

src/esp/assets/ResourceManager.cpp
aclegg3: These functions work slightly differently. `loadGeneralMeshData` makes use of `addComponent` to handle mesh hierarchies. `loadPTexMeshData` assumes all meshes are already transformed by all but the global rotation transform (and as written assumes only 1 mesh, `... MeshMetaData(index, index) ...` ). I've added the global transform to the solitary `MeshTransformNode` purely for consistency here, it is likely not to be sued until collision mesh construction.

aclegg3:
I noticed issues with loading arbitrary .glb files as scenes (like the wayfair objects themselves). The fix fits nicely into the re-design of this PR using MeshTransformNode to build the collision objects. Once this was done, mesh shifting no longer applied correctly to objects, so rather than push unstable functionality for the sake of small PRs, I rolled the refactor for bounding box shift in also. I'll re-frame the PR description to reflect this.

PR #313, opened 2019-10-18
Allow agent reference to be kept

erikwijmans:
## Motivation and Context

Habitat-sim currently crashes if you do

```
agent = sim.get_agent(0)
sim.close()
```

as the sensors have OpenGL related things, but the simulator owns the OpenGL context.

Also makes `sensors` on the agent private, so we don't have to go down this rabbit hole further.

## How Has This Been Tested

With the test!

## Types of changes

### DIFF
--- a/tests/test_simulator.py
+++ b/tests/test_simulator.py
@@ -1,3 +1,4 @@
+import multiprocessing
 import random
 
 import examples.settings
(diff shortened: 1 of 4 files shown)


tests/test_simulator.py
```diff
@@ -40,3 +41,31 @@ def test_empty_scene(sim):
...
+
+    sim_cfg.scene.id = "data/scene_datasets/habitat-test-scenes/van-gogh-room.glb"
+
+    sim = habitat_sim.Simulator(habitat_sim.Configuration(sim_cfg, [agent_config]))
+
+    agent = sim.get_agent(0)
+
+    sim.close()
+    sim = None
+
```
mathfac: Can we add 2 sequential `sim` reinitialization.

erikwijmans: How about this.

PR #322, opened 2019-10-29
Bug fix: add "reshape" functionality to utility viewer

bigbike:
In the viewport event, we only reset the viewport but do not specify what to do when window size is changed.
This PR is to update the projection matrix to handle it.

## Motivation and Context
As titled.
The OLD version looks like this:
Initial screen:
image

After window size was changed:
image

## How Has This Been Tested

Local test:
Initial screen:
image

After window size was changed:
image

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -364,7 +364,7 @@ void Viewer::viewportEvent(ViewportEvent& event) {
   int width = event.windowSize()[0];
   int height = event.windowSize()[1];
   renderCamera_->setProjectionMatrix(width, height, znear_, zfar_, hfov_);
-} 
+}
 
 void Viewer::mousePressEvent(MouseEvent& event) {
   if (event.button() == MouseEvent::Button::Left)
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -364,7 +364,7 @@ void Viewer::viewportEvent(ViewportEvent& event) {
   int width = event.windowSize()[0];
   int height = event.windowSize()[1];
   renderCamera_->setProjectionMatrix(width, height, znear_, zfar_, hfov_);
-}  // namespace
+} 
 
 void Viewer::mousePressEvent(MouseEvent& event) {
   if (event.button() == MouseEvent::Button::Left)
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -104,6 +104,10 @@ class Viewer : public Magnum::Platform::Application {
 
   bool drawObjectBBs = false;
 
+  const float hfov_ = 90.0f;
+  const float znear_ = 0.01f;
+  const float zfar_ = 1000.0f;
+
   Magnum::Timeline timeline_;
 };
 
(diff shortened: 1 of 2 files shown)


bigbike:

src/utils/viewer/viewer.cpp
```diff
@@ -359,7 +360,11 @@ void Viewer::drawEvent() {
 void Viewer::viewportEvent(ViewportEvent& event) {
   GL::defaultFramebuffer.setViewport({{}, framebufferSize()});
   renderCamera_->getMagnumCamera().setViewport(event.windowSize());
-}
+
+  int width = event.windowSize()[0];
+  int height = event.windowSize()[1];
+  renderCamera_->setProjectionMatrix(width, height, znear_, zfar_, hfov_);
+}  // namespace
```
mosra: Why the `// namespace`?

src/utils/viewer/viewer.cpp
```diff
@@ -359,7 +360,11 @@ void Viewer::drawEvent() {
 void Viewer::viewportEvent(ViewportEvent& event) {
   GL::defaultFramebuffer.setViewport({{}, framebufferSize()});
   renderCamera_->getMagnumCamera().setViewport(event.windowSize());
```
mosra: The `SceneGraph::Camera` class can handle aspect ratio preservation in the `setViewport()` call automatically -- you only need to call [Camera::setAspectRatioPolicy()](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1Camera.html#a98f200ca05c1e0a7fa06845b3510f8f2) with a desired value.

Looking at the code, [here](https://github.com/facebookresearch/habitat-sim/blob/eb191d19165a32101a3745bde03518ddfe86155e/src/esp/gfx/RenderCamera.cpp#L36) it's set to `NotPreserved` for some reason, which effectively disables such handling.

bigbike: I see. Good to know. I do not know the purpose either. Will mark it, and solve it in recent "sensor/camera redesign" project (See slack for more details. Hope you can participate in this project. :-) )

mosra: Well, I commented here because doing it via `setAspectRatioPolicy()` would result in *less code* overall. Merging this as-is just piled up more unnecessary tech debt in my opinion.

`NotPreserved` is the default, and it probably makes sense for rendering headless -- there you have no arbitrary framebuffer resizes anyway. But having that there doesn't prevent the *viewer* from redefining it to handle this differently when having a GUI.

msavva: Completely agree with @mosra .  There is no reason to keep the `NotPreserved` here and add extra code to implement behavior we could achieve by just changing the flag.  Let's have a followup PR to remove the extra code + extra comments, and just switch the flag instead.

bigbike: I was thinking when @msavva wrote it, he must have some reasons, so I was trying to respect that by not touching it, and leaving it to the future discussion.
But since he also leans towards changing it now, I am happy to write a new PR to roll back this one, and fix the issue in the other way.

src/utils/viewer/viewer.cpp
bigbike: Oh, this is because the "vim" sometimes would automatically add the brackets for the user (totally unnecessarily), and if the user did not pay attention to it, and applied the clang-format, such "wrong" comments would be auto-added.

Issue #323, opened 2019-10-30
Enable clang-tidy


msbaines:
## 🚀 Feature

Automate coding style checking.

## Motivation

Developers will be more productively navigating the code and making changes if a consistent
style is followed.

## Pitch

clang-tidy also support static analysis.

mathfac:
The task looks finished.

msbaines:
clang-tidy is now enabled but I am still enabling some checks. I can close this and open separate tasks for the currently disabled checks.

PR #324, opened 2019-10-30
[clang-tidy] initial checkin

msbaines:
Just checking one directory for now. Will send follow-up PRs to
fix other directories.

Issue: #323

## Motivation and Context

Automatically enforce coding style.

## Types of changes

## Checklist

- [  ] I have updated the documentation accordingly.


mosra:
Forgive me if the answer is "yes of course" but since I never used Clang Tidy, I have to ask -- does this check function/variable/... naming as well? Such as the case here: https://github.com/facebookresearch/habitat-sim/pull/309/files#r340844325

Thanks! :)

msbaines:
> Forgive me if the answer is "yes of course" but since I never used Clang Tidy, I have to ask -- does this check function/variable/... naming as well? Such as the case here: https://github.com/facebookresearch/habitat-sim/pull/309/files#r340844325
> 
> Thanks! :)

Not yet but should be possible to add. Looks like google coding style is still WiP upstream. The list of available checks so far:

https://clang.llvm.org/extra/clang-tidy/checks/list.html

run-clang-tidy.py
```diff
@@ -0,0 +1,388 @@
+#!/usr/bin/env python
```
erikwijmans: Can we move this to a `tools` or something folder?

msbaines:
All feedback addressed. Ready for review.

tools/run-clang-tidy.py
```diff
@@ -0,0 +1,388 @@
+#!/usr/bin/env python
```
mathfac: @msbaines is this custom script for Habitat or it can be `pip` installed?

msbaines: The script comes from here:

https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py

I don't believe its available via pip install.

On Mac, you will also need to:

brew install llvm

PR #325, opened 2019-10-30
[clang-tidy] fix various issues reported by clang-tidy

msbaines:
For details on any specific check, please see:

https://releases.llvm.org/9.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/list.html

Fixed the following issues:

habitat-sim/src/esp/assets/GltfMeshData.cpp:11:20: error: default arguments on virtual or override methods are prohibited [google-default-arguments,-warnings-as-errors]
void GltfMeshData::uploadBuffersToGPU(bool forceReload) {
                   ^
habitat-sim/src/esp/assets/Asset.cpp:12:1: error: do not use namespace using-directives; use using-declarations instead [google-build-using-namespace,-warnings-as-errors]
using namespace Corrade::Utility::String;
^
habitat-sim/src/esp/assets/GenericInstanceMeshData.cpp:226:31: error: default arguments on virtual or override methods are prohibited [google-default-arguments,-warnings-as-errors]
void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
                              ^
habitat-sim/src/esp/nav/PathFinder.cpp:135:39: error: non-const reference parameter 'islandVerts', make it const or use a pointer [google-runtime-references,-warnings-as-errors]
                  std::vector<vec3f>& islandVerts) {
                                      ^
habitat-sim/src/esp/physics/PhysicsManager.cpp:170:22: error: default arguments on virtual or override methods are prohibited [google-default-arguments,-warnings-as-errors]
void PhysicsManager::stepPhysics(double dt) {
                     ^

## Motivation and Context

Fix clang-tidy issues so that I can turn clang-tidy on for more directories.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -81,7 +81,7 @@ class IslandSystem {
         if (navMesh->isValidPolyRef(startRef) &&
             (polyToIsland_.find(startRef) == polyToIsland_.end())) {
           uint32_t newIslandId = islandRadius_.size();
-          expandFrom(navMesh, filter, newIslandId, startRef, islandVerts);
+          expandFrom(navMesh, filter, newIslandId, startRef, &islandVerts);
 
           // The radius is calculated as the max deviation from the mean for all
           // points in the island
@@ -132,9 +132,9 @@ class IslandSystem {
                   const dtQueryFilter* filter,
                   const uint32_t newIslandId,
                   const dtPolyRef& startRef,
-                  std::vector<vec3f>& islandVerts) {
+                  std::vector<vec3f>* islandVerts) {
     polyToIsland_.emplace(startRef, newIslandId);
-    islandVerts.clear();
+    islandVerts->clear();
 
     // Force std::stack to be implemented via an std::vector as linked
     // lists are gross
@@ -151,7 +151,7 @@ class IslandSystem {
       navMesh->getTileAndPolyByRefUnsafe(ref, &tile, &poly);
 
       for (int iVert = 0; iVert < poly->vertCount; ++iVert) {
-        islandVerts.emplace_back(
+        islandVerts->emplace_back(
             Eigen::Map<vec3f>(&tile->verts[poly->verts[iVert] * 3]));
       }
 
--- a/src/esp/physics/PhysicsManager.h
+++ b/src/esp/physics/PhysicsManager.h
@@ -182,7 +182,7 @@ class PhysicsManager {
    * increments of @ref fixedTimeStep_.
    * @param dt The desired amount of time to advance the physical world.
    */
-  virtual void stepPhysics(double dt = 0.0);
+  virtual void stepPhysics(double dt);
 
   // =========== Global Setter functions ===========
 
--- a/src/esp/assets/Asset.cpp
+++ b/src/esp/assets/Asset.cpp
@@ -9,7 +9,7 @@
 namespace esp {
 namespace assets {
 
-using namespace Corrade::Utility::String;
+using Corrade::Utility::String::endsWith;
 
 AssetInfo AssetInfo::fromPath(const std::string& path) {
   AssetInfo info{AssetType::UNKNOWN, path};
--- a/src/esp/assets/GenericInstanceMeshData.h
+++ b/src/esp/assets/GenericInstanceMeshData.h
@@ -32,7 +32,7 @@ class GenericInstanceMeshData : public BaseMesh {
   virtual bool loadPLY(const std::string& plyFile);
 
   // ==== rendering ====
-  virtual void uploadBuffersToGPU(bool forceReload = false) override;
+  virtual void uploadBuffersToGPU(bool forceReload) override;
   RenderingBuffer* getRenderingBuffer() { return renderingBuffer_.get(); }
 
   virtual Magnum::GL::Mesh* getMagnumGLMesh() override;
--- a/src/esp/assets/GltfMeshData.h
+++ b/src/esp/assets/GltfMeshData.h
@@ -22,7 +22,7 @@ class GltfMeshData : public BaseMesh {
 
   virtual ~GltfMeshData(){};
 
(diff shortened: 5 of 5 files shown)


erikwijmans:
Maybe just me, but I disagree with
> habitat-sim/src/esp/nav/PathFinder.cpp:135:39: error: non-const reference parameter 'islandVerts', make it const or use a pointer [google-runtime-references,-warnings-as-errors]
std::vector& islandVerts)

References nicely deal with ownership, which, IMO, is the more confusing thing than whether or not an argument is mutable.

src/esp/nav/PathFinder.cpp
```diff
@@ -132,9 +132,9 @@ class IslandSystem {
                   const dtQueryFilter* filter,
                   const uint32_t newIslandId,
                   const dtPolyRef& startRef,
-                  std::vector<vec3f>& islandVerts) {
+                  std::vector<vec3f>* islandVerts) {
```
mosra: This is in direct contrast with what Magnum's coding guidelines have (and I'm well aware of this Google coding style rule and have a different opinion):

- In Magnum, a pointer indicates a "nullable" parameter (whereas here passing a `nullptr` would crash), while a reference indicates a parameter that can't be null. This is in order to move the null-checking responsibility to the caller of the function, not the function itself (where, in order to be safe, it would need a lot of `assert(islandVerts)` noise everywhere). A consistency in this particular rule is crucial, otherwise you can't trust the API to accept `nullptr` in pointer parameters.
- If a function has "out" parameters, it's explicitly said so in the documentation as `@param[out]` (for example, [MeshTools::flipNormals()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#af4417b043b3de224d43edf1748f880fc))

Can this particular rule be disabled? Because it'll inevitably lead to Magnum-inspired APIs being *almost* the same, except for `&` parameters being converted to `*`.

src/esp/physics/PhysicsManager.h
```diff
@@ -182,7 +182,7 @@ class PhysicsManager {
    * increments of @ref fixedTimeStep_.
    * @param dt The desired amount of time to advance the physical world.
    */
-  virtual void stepPhysics(double dt = 0.0);
+  virtual void stepPhysics(double dt);
```
mosra: I understand why it complains (because an override could be defining a different default), but nevertheless I think defaults in overrideable APIs are useful. What I'm doing in Magnum is, roughly the following, based on the ages-old [Virtuality article by Herb Sutter](http://www.gotw.ca/publications/mill18.htm):

```cpp
// non-virtual, public
void stepPhysics(double dt = 0.0) {
    // do various error checking here, so the (re)implementations don't need to duplicate that
    doStepPhysics(dt);
}

// virtual with no default parameters, ideally protected or private
virtual void doStepPhysics(double dt);
```

src/esp/nav/PathFinder.cpp
msbaines: > This is in direct contrast with what Magnum's coding guidelines have (and I'm well aware of this Google coding style rule and have a different opinion):
> 
> * In Magnum, a pointer indicates a "nullable" parameter (whereas here passing a `nullptr` would crash), while a reference indicates a parameter that can't be null. This is in order to move the null-checking responsibility to the caller of the function, not the function itself (where, in order to be safe, it would need a lot of `assert(islandVerts)` noise everywhere). A consistency in this particular rule is crucial, otherwise you can't trust the API to accept `nullptr` in pointer parameters.
> * If a function has "out" parameters, it's explicitly said so in the documentation as `@param[out]` (for example, [MeshTools::flipNormals()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#af4417b043b3de224d43edf1748f880fc))
> 
> Can this particular rule be disabled? Because it'll inevitably lead to Magnum-inspired APIs being _almost_ the same, except for `&` parameters being converted to `*`.
> 
> EDIT: just saw the comment from @erikwijmans -- I guess we agree on this one :)

I think there are pros and cons to both approaches. Avoiding a null-check is  nice. Being able to easily reason about mutability in the context of the call is also nice. I slightly prefer the former but not strongly. However, I think consistency with Magnum is an even more important concern.

I will remove this check. I will try to add a check for the Magnum convention in the future.

src/esp/physics/PhysicsManager.h
msbaines: Thanks much for the pointer to this excellent article. Great read!

If we were to follow the approach outlined in the article, then defaults in overrideable APIs would not be needed since the public interface is not overrideable.

Since this is contentious, I will remove it. My immediate goal is to enable coding style verification and get a baseline setup.

msbaines: You could use an overload instead of an override in order to get default behavior which is defined in exactly one place.

src/esp/nav/PathFinder.cpp
mosra: > References nicely deal with ownership

> Raw pointers should not be used to transfer ownership. To transfer ownership, a smart pointer should be used.

I think you're both saying the same thing, just differently :) If a function takes a reference, it's clear it *does not* take ownership (nobody calling `delete` on it and such). On the other hand, some 3rd APIs might be designed pre-C++11 (so no smart pointers) and C APIs have just pointers, which means if there's a pointer, it's less clear if there's some ownership transfer happening or not.

src/esp/physics/PhysicsManager.h
mosra: True, good idea -- just so we're on the same page, you mean a non-virtual overload with one parameter less that calls the virtual function with the default value, right? Like

```cpp
void stepPhysics() {
    stepPhysics(0.0);
}
```

However I have a feeling some static analyzers complain when you have overloaded virtuals.

msbaines:
Change of plans.

I'm going to change my original clang-tidy to disable any check that fails. I will then send CLs one at a time to turn on tests. That will allow for discussion on a test by test basis.

PR #331, opened 2019-10-31
Revert #322 and add "reshape" functionality to utility viewer

bigbike:
-) revert the extra code and comments in #322;
-) move setAspectRatioPolicy out of the setProjectionMatrix;
-) set the correct policy in the viewer;

## Motivation and Context

See #322 

## How Has This Been Tested
Local test:

Original window:
image

After window size was changed:
image

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -104,10 +104,6 @@ class Viewer : public Magnum::Platform::Application {
 
   bool drawObjectBBs = false;
 
-  const float hfov_ = 90.0f;
-  const float znear_ = 0.01f;
-  const float zfar_ = 1000.0f;
-
   Magnum::Timeline timeline_;
 };
 
@@ -182,7 +178,10 @@ Viewer::Viewer(const Arguments& arguments)
 
   int width = viewportSize[0];
   int height = viewportSize[1];
-  renderCamera_->setProjectionMatrix(width, height, znear_, zfar_, hfov_);
+  float znear = 0.01f;
+  float zfar = 1000.0f;
+  float hfov = 90.0f;
+  renderCamera_->setProjectionMatrix(width, height, znear, zfar, hfov);
   renderCamera_->getMagnumCamera().setAspectRatioPolicy(
       Magnum::SceneGraph::AspectRatioPolicy::Extend);
 
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -176,11 +176,11 @@ Viewer::Viewer(const Arguments& arguments)
   rgbSensorNode_->translate({0.0f, rgbSensorHeight, 0.0f});
   agentBodyNode_->translate({0.0f, 0.0f, 5.0f});
 
-  renderCamera_->setProjectionMatrix(viewportSize[0],  // width
-                                     viewportSize[1],  // height
-                                     0.01f,            // znear
-                                     1000.0f,          // zfar
-                                     90.0f);           // hfov
+  renderCamera_->setProjectionMatrix(viewportSize.x(),  // width
+                                     viewportSize.y(),  // height
+                                     0.01f,             // znear
+                                     1000.0f,           // zfar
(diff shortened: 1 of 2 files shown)


bigbike:

bigbike:
> But this does not revert any of the new code added in 322... :/
> 
> Unless I'm missing something, `renderCamera_->getMagnumCamera().setViewport(event.windowSize());` will do the job, and you only need to `setAspectRatioPolicy()` in the constructor, as you do here. None of the `hfov_` / ... saving or `renderCamera_->setProjectionMatrix()` in the viewport is needed.

I sincerely apologize for my careless mistake. (This is because I git checked out this viewer.cpp file multiple times, and forgot to remove it at the last checkout.)

Updated.

mosra:
... and what about [this part of the viewer.cpp](https://github.com/facebookresearch/habitat-sim/pull/322/files#diff-12a33243a58a88f83a63a13dfe799584R107-R185)? Seems unnecessary to save those if they're used just in one place.

```diff
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -104,6 +104,10 @@ class Viewer : public Magnum::Platform::Application {
 
   bool drawObjectBBs = false;
 
+  const float hfov_ = 90.0f;
+  const float znear_ = 0.01f;
+  const float zfar_ = 1000.0f;
+
   Magnum::Timeline timeline_;
 };
 
@@ -176,12 +180,9 @@ Viewer::Viewer(const Arguments& arguments)
   rgbSensorNode_->translate({0.0f, rgbSensorHeight, 0.0f});
   agentBodyNode_->translate({0.0f, 0.0f, 5.0f});
 
-  float hfov = 90.0f;
   int width = viewportSize[0];
   int height = viewportSize[1];
-  float znear = 0.01f;
-  float zfar = 1000.0f;
-  renderCamera_->setProjectionMatrix(width, height, znear, zfar, hfov);
+  renderCamera_->setProjectionMatrix(width, height, znear_, zfar_, hfov_);
 
   // Load navmesh if available
   const std::string navmeshFilename = io::changeExtension(file, ".navmesh");
```

bigbike:
Yes, in this case we do not have to save them.
Updated. Take a look.

src/utils/viewer/viewer.cpp
```diff
@@ -180,9 +176,13 @@ Viewer::Viewer(const Arguments& arguments)
   rgbSensorNode_->translate({0.0f, rgbSensorHeight, 0.0f});
   agentBodyNode_->translate({0.0f, 0.0f, 5.0f});
 
-  int width = viewportSize[0];
-  int height = viewportSize[1];
-  renderCamera_->setProjectionMatrix(width, height, znear_, zfar_, hfov_);
+  renderCamera_->setProjectionMatrix(viewportSize[0],  // width
+                                     viewportSize[1],  // height
```
mosra: `.x()` and `.y()` instead of `[0]` and `[1]`? :) then you won't need to comment `// width` `// height`

src/utils/viewer/viewer.cpp
```diff
@@ -180,9 +176,13 @@ Viewer::Viewer(const Arguments& arguments)
   rgbSensorNode_->translate({0.0f, rgbSensorHeight, 0.0f});
   agentBodyNode_->translate({0.0f, 0.0f, 5.0f});
 
-  int width = viewportSize[0];
-  int height = viewportSize[1];
-  renderCamera_->setProjectionMatrix(width, height, znear_, zfar_, hfov_);
+  renderCamera_->setProjectionMatrix(viewportSize.x(),  // width
+                                     viewportSize.y(),  // height
```
bigbike: Can I keep the comment here so that we understand the order of input parameters in the function? That means we know here is w x h, not h x w.

mosra: Sure, it was just an idea how to make it "self-documenting".

bigbike:
Cool. Thanks a lot for the valuable suggestions.
Now, I think it is good to merge it.

PR #333, opened 2019-11-01
Recompute navmesh programmatically

aclegg3:
## Motivation and Context

Our current workflow is to pre-compute a navmesh fore each scene using the `utils/datatool` and save this navmesh to file. When the simulator is initialized and loads a scene, we look for a specified navmesh to load with the scene. Each navmesh is generated with a set of assumptions about the agent and scene including navmesh cell density and size, agent radius, height, max pathable slope, etc...

As we begin to generate scenes with multiple agents, load/configure agents, and dynamically modify the static environment, pre-computation of the navmesh may quickly become limiting and programmatic generation a more attractive option. 

This PR adds configurable programmatic computation of the active scene navmesh through the `Simulator` class in python and `Viewer` demo in C++.

Future TODO: Future PRs will include the option to recompute the navmesh including `STATIC` type `RigidObject`s.

## How Has This Been Tested

Local with scenes:
- `NONE` (empty scene, nothing to do)
- `skokloster-castle.glb`
  -  default settings recomputed navmesh includes roof, is there a way to cull this?
- `van-gogh-room.glb`
- Replica: `apartment_0/habitat/mesh_semantic.ply`
- gibson scenes

Added a python CI test for this feature: `tests/recompute_navmesh.py`
 
## Types of changes

## Checklist


src/esp/assets/ResourceManager.cpp
```diff
@@ -1207,5 +1207,49 @@ bool ResourceManager::loadSUNCGHouseFile(const AssetInfo& houseInfo,
   return true;
 }
 
+vec3f toEig(Magnum::Vector3 v) {
```
erikwijmans: This can be done via `Mn::EigenIntegration::cast<vec3f>`

aclegg3: Thanks, hoping someone would have a shortcut for this. :+1:

habitat_sim/simulator.py
```diff
@@ -284,6 +284,9 @@ def apply_torque(self, torque, object_id, scene_id=0):
     def get_world_time(self, scene_id=0):
         return self._sim.get_world_time()
 
+    def recompute_navmesh(self, navmesh_settings):
+        self.pathfinder = self._sim.recompute_navmesh(navmesh_settings)
```
erikwijmans: IMO, it makes more sense to have the pathfinder be recomputed in place. i.e. `self._sim.recompute_navmesh` should take a reference to the current pathfinder and return nothing.  This'll keep weird bugs from happening where the new pathfinder doesn't get picked up by something else that has a reference to the old one.

aclegg3: `self.pathfinder.load_nav_mesh` loads the navmesh from a file which was generated with some fixed settings, so new settings cannot be used.

Original intent was to allow the navmesh to be stored for the case that you want multi-agent support or some other use of multiple navmeshes with different settings.

erikwijmans: Due to nice division of responsibilities, something like `self.pathfinder.recompute_navmesh(navmesh_settings)` isn't possible without giving the pathfinder way more information than it needs access to do its job 99% of the time.

src/esp/assets/ResourceManager.h
```diff
@@ -134,11 +134,25 @@ class ResourceManager {
     return meshes_[meshIndex]->meshTransform_;
   };
 
-  const MeshMetaData& getMeshMetaData(std::string filename) {
+  /**
+   * @brief Public accessor for @ref MeshMetaData. See @ref resourceDict_.
+   * @param filename The identifying string key for the asset.
+   * @return The @ref MeshMetaData object for the asset.
+   */
+  const MeshMetaData& getMeshMetaData(const std::string filename) {
```
mosra:
```suggestion
  const MeshMetaData& getMeshMetaData(const std::string& filename) {
```

src/esp/gfx/Simulator.h
```diff
@@ -263,6 +260,16 @@ class Simulator {
...
 
+  /**
+   * @brief Compute the navmesh for the simulator's current active scene.
+   * @param navMeshSettings The @ref nav::NavMeshSettings instance to
+   * parameterize the navmesh construction.
+   * @return The resulting completed @ref Pathfinder object or nullptr if
+   * failed.
+   */
+  std::shared_ptr<nav::PathFinder> recomputeNavMesh(
+      nav::NavMeshSettings& navMeshSettings);
```
mosra: This should be either `const&` or explicitly documented that it's a `@param[in,out]` and *why* a mutable access is needed.

examples/demo_runner.py
```diff
@@ -314,6 +314,11 @@ def init_common(self):
 
         self._sim = habitat_sim.Simulator(self._cfg)
 
+        # test recompute navmesh
+        navmesh_settings = habitat_sim.NavMeshSettings()
```
mosra:
```suggestion
        navmesh_settings = habitat_sim.nav.NavMeshSettings()
```

to follow that it's exposed through the `nav` package?

src/esp/bindings/ShortestPathBindings.cpp
```diff
@@ -43,6 +43,29 @@ void initShortestPathBindings(py::module& m) {
       .def_readwrite("geodesic_distance",
                      &MultiGoalShortestPath::geodesicDistance);
 
+  py::class_<NavMeshSettings, NavMeshSettings::ptr>(m, "NavMeshSettings")
+      .def(py::init(&NavMeshSettings::create<>))
```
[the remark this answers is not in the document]
aclegg3: These are bindings of C++ functions, so the doc strings would be on the C++ side. `PathFinder.h` is not documented yet.

tests/test_greedy_follower.py
```diff
@@ -46,7 +47,8 @@ def pbar():
 
 
 @pytest.mark.parametrize("test_navmesh", test_navmeshes)
-def test_greedy_follower(test_navmesh, scene_graph, pbar):
+@pytest.mark.parametrize("recompute_navmesh", [True, False])
```
erikwijmans: Two requests:  
1) Can you lave this test unchanged? I have been using it if I need to make changes to the greedy follower to test it on all ~600 navmeshes, so this test needs to be really fast.

2) For the navmesh test, I think a possible good test would be as follows
-- Compute the (geodesic) shortest path between some pairs of points
-- Recompute the navmesh with the base setting, recompute paths, they should match
-- Recompute navmesh on very different settings (probably half the agent radius?), some, if not all, paths should be different.

aclegg3: Thanks for the suggestion. 👍 Done.

aclegg3:
Ready for review.

habitat_sim/simulator.py
erikwijmans: I still think `self._sim.recompute_navmesh(self.pathfinder, navmesh_settings)` is cleaner and less likely to run into weird bugs where the changes to the pathfinder don't get propagated (i.e. something still has a reference to the old pathfinder).

src/esp/gfx/Simulator.cpp
```diff
@@ -350,5 +350,21 @@ double Simulator::getWorldTime() {
   return NO_TIME;
 }
 
+std::shared_ptr<nav::PathFinder> Simulator::recomputeNavMesh(
```
msbaines: Use a unique_ptr here. The caller can convert to shared_ptr if they want to .

https://herbsutter.com/2013/05/29/gotw-89-solution-smart-pointers/

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -262,6 +265,15 @@ void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
       .setIndexBuffer(std::move(indices), 0,
                       Mn::GL::MeshIndexType::UnsignedInt);
 
+  cpu_ibo_ = new_indices;
```
msbaines: cpu_ibo_ = std::move(new_indices);

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -262,6 +265,15 @@ void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
...
+
+  collisionMeshData_.positions =
```
msbaines: We are setting collisionMeshData_ already in loadPly and now here also. I think it may be possible to only set collisionMeshData_ in one place by moving some of the logic in this function to loadPly. It also feels like this a bug fix that should really be its own change.

src/esp/assets/ResourceManager.cpp
```diff
@@ -944,8 +944,8 @@ bool ResourceManager::loadGeneralMeshData(
     const quatf transform = info.frame.rotationFrameToWorld();
     Magnum::Matrix4 R = Magnum::Matrix4::from(
         Magnum::Quaternion(transform).toMatrix(), Magnum::Vector3());
-    resourceDict_[filename].root.T_parent_local =
-        R * resourceDict_[filename].root.T_parent_local;
+    resourceDict_[filename].root.transformFromLocalToParent =
```
msbaines: This name change should probably be a separate change.

habitat_sim/simulator.py
aclegg3: Sure, so we alter the function to take an existing pathfinder and reseat it, rather than returning a new one. Would still suite my purpose of allowing multiple pathfinders to exist. I'll make this change. 👍

src/esp/gfx/Simulator.h
```diff
@@ -6,6 +6,7 @@
 
 #include "esp/core/esp.h"
 #include "esp/core/random.h"
+#include "esp/nav/PathFinder.h"
```
msbaines: I don't think this is necessary. There is already a forward declaration below.

aclegg3: For `NavMeshSettings`

src/esp/assets/ResourceManager.cpp
aclegg3: Done.

src/esp/gfx/Simulator.h
msbaines: Ah. To be consistent, you may want to consider forward declaring NavMeshSettings.

habitat_sim/simulator.py
erikwijmans: Sounds good!  You can of course make an overloaded version for all these things that creates a new pathfinder, but since this method is modifying `self.pathfinder` it seems cleaner to do it in place :)

src/tests/ResourceManagerTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+      Magnum::Vector3{1, 1, 1},   Magnum::Vector3{1, -1, -1},
+      Magnum::Vector3{1, 1, -1},  Magnum::Vector3{1, -1, 1},
+      Magnum::Vector3{1, 1, -1},  Magnum::Vector3{-1, -1, -1},
+      Magnum::Vector3{-1, 1, -1}, Magnum::Vector3{1, -1, -1},
+      Magnum::Vector3{1, 1, 1},   Magnum::Vector3{-1, 1, -1},
+      Magnum::Vector3{-1, 1, 1},  Magnum::Vector3{1, 1, -1},
+      Magnum::Vector3{1, -1, 1},  Magnum::Vector3{-1, -1, -1},
+      Magnum::Vector3{-1, -1, 1}, Magnum::Vector3{1, -1, -1},
+      Magnum::Vector3{1, 1, 1},   Magnum::Vector3{-1, -1, 1},
+      Magnum::Vector3{-1, 1, 1},  Magnum::Vector3{1, -1, 1}};
```
mosra: FYI -- the constructors are not explicit, so you can do just `{-1, 1, 1}` for example ;)

src/tests/ResourceManagerTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+  std::vector<uint32_t> indexGroundTruth{
+      0,  1,  2,  0,  3,  1,  4,  5,  6,  4,  7,  5,  8,  9,  10, 8,  11, 9,
+      12, 13, 14, 12, 15, 13, 16, 17, 18, 16, 19, 17, 20, 21, 22, 20, 23, 21};
+
+  for (size_t vix = 0; vix < joinedBox->vbo.size(); vix++) {
+    // Cr::Utility::Debug() << joinedBox->vbo[vix] << " vs " <<
+    // vertGroundTruth[vix];
+    ASSERT_EQ(vertGroundTruth[vix],
+              Magnum::Vector3(joinedBox->vbo[vix][0], joinedBox->vbo[vix][1],
+                              joinedBox->vbo[vix][2]));
```
mosra: Could be just `Magnum::Vector3(joinedBox->vbo[vix])` if you `#include <Magnum/EigenIntegration/Integration.h>` :)

src/esp/gfx/Simulator.cpp
aclegg3: With re-design this is no longer an issue. :)

src/esp/gfx/Simulator.h
aclegg3: Done.

src/esp/assets/GenericInstanceMeshData.cpp
aclegg3: Good point. I've create a new PR #389 for this including your previous suggestion above. 👍

aclegg3:
This PR is now down to navmesh computation code only and ready for review. 😄

src/utils/viewer/viewer.cpp
```diff
@@ -201,17 +207,26 @@ Viewer::Viewer(const Arguments& arguments)
...
+      recomputeNavMesh(file, navMeshSettings);
+    }
   }
 
   // connect controls to navmesh if loaded
   if (pathfinder_->isLoaded()) {
+    // TODO: some scenes could have pathable roof polygons. We are not filtering
+    // those starting points here.
+    vec3f position = pathfinder_->getRandomNavigablePoint();
+    agentBodyNode_->setTranslation(Vector3(position));
```
msbaines: How are lines 227 and 228 related to this change?

src/utils/viewer/viewer.cpp
```diff
@@ -297,6 +312,22 @@ void Viewer::pushLastObject() {
   physicsManager_->applyForce(objectIDs_.back(), force, rel_pos);
 }
 
+void Viewer::recomputeNavMesh(const std::string& sceneFilename,
+                              nav::NavMeshSettings& navMeshSettings) {
+  std::shared_ptr<nav::PathFinder> pf = std::make_shared<nav::PathFinder>();
```
msbaines: nav::PathFinder::ptr pf = nav::PathFinder::create();

src/utils/viewer/viewer.cpp
```diff
@@ -297,6 +312,22 @@ void Viewer::pushLastObject() {
...
+
+  std::unique_ptr<assets::MeshData> joinedMesh =
```
msbaines: assets::MeshData::uptr joinedMesh =

src/utils/viewer/viewer.cpp
```diff
@@ -201,17 +207,26 @@ Viewer::Viewer(const Arguments& arguments)
...
-
-    const vec3f position = pathfinder_->getRandomNavigablePoint();
-    agentBodyNode_->setTranslation(Vector3(position));
+  if (file.compare(esp::assets::EMPTY_SCENE) != 0) {
+    const std::string navmeshFilename = io::changeExtension(file, ".navmesh");
+    if (io::exists(navmeshFilename) && !args.isSet("recompute-navmesh")) {
+      LOG(INFO) << "Loading navmesh from " << navmeshFilename;
+      pathfinder_->loadNavMesh(navmeshFilename);
+    } else {
+      // TODO: agent radius is not configured here
```
msbaines: TODO(alexanderwclegg)

I haven't turned it on yet but there is a clang-tidy warning that verifies that TODOs have a name attached.

src/esp/gfx/Simulator.cpp
```diff
@@ -350,5 +350,19 @@ double Simulator::getWorldTime() {
   return NO_TIME;
 }
 
+bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
+                                 const nav::NavMeshSettings& navMeshSettings) {
+  std::unique_ptr<assets::MeshData> joinedMesh =
```
msbaines: assets::MeshData::uptr joinedMesh =

src/utils/viewer/viewer.cpp
aclegg3: Without these, the agent's position is hardcoded in line 199 resulting in cases such as: 
image

I guess this could be its own PR 😆

src/utils/viewer/viewer.cpp
aclegg3: Noted. I'll assume then that we intend to use TODO for concrete goals only. 🤔  I'll refrain from using TODO to make notes as in this case. 😉

src/utils/viewer/viewer.cpp
msbaines: You could make it a separate PR (generally preferred) or it should be mentioned explicitly in the commit message. Reviewing the history of the code (to locate a regression, etc) becomes difficult when the commit message and code do not agree.

msbaines:
https://google.github.io/styleguide/cppguide.html#TODO_Comments

src/esp/gfx/Simulator.cpp
```diff
@@ -350,5 +350,19 @@ double Simulator::getWorldTime() {
   return NO_TIME;
 }
 
+bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
+                                 const nav::NavMeshSettings& navMeshSettings) {
+  assets::MeshData::uptr joinedMesh =
+      resourceManager_.createJoinedCollisionMesh(config_.scene.id);
+
+  if (!pathfinder.build(navMeshSettings, *joinedMesh)) {
```
erikwijmans: This call is going to leak memory here: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/nav/PathFinder.cpp#L523 and here: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/nav/PathFinder.cpp#L549

Issue #320, opened 2019-10-26
How to set object category?


dhuber23b:
When I manually add a physics object to the scene from the library, it has a semantic category of -1.  Is there a way I can set this to something of my choosing?

aclegg3:
Hey Daniel, this feature has been discussed and is WIP. I'll take a look at the engineering effort and see what we can do to prioritize this.

aclegg3:
Update on this issue: #341 is a prototype which allows you to manually configure the semantic category. Improvements to come.

aclegg3:
Update: configurable semantic id is now enabled by #668.

Issue #336, opened 2019-11-02
Same mesh for semantics and color (MP3D)


dhruvbatra:
## 🚀 Feature
Use the same mesh for semantics as we do for color/texture (for MP3D dataset). 

More generally, develop a standard protocol/format for information (color, semantics, instance annotation, material properties, etc) on meshes. 

## Motivation

As stated in our README, the semantic sensor in MP3D houses currently requires the use of additional 3D meshes with orders of magnitude more geometric complexity leading to reduced performance. This leads to a significant reduction in rendering speed when using the semantics sensor. 

My understanding (from talking to @msavva) is that this is because semantic annotation was done on more detailed meshes, which made the lives of the annotators easy. 

Now, we should transfer this information to the color texture meshes. 

This will increase FPS and reduce memory footprint (we won't have to load 2 sets of meshes for the same env), meaning that more environments will be loadable on the same GPU.  

## Details/Subtleties

Semantics information will need to be copied over as additional texture to the color texture meshes. @msavva has tried direct copy (and has some code available), but it won't solve the problem because sometimes there aren't enough faces/meshes on an object/surface to just copy over the semantic information. So we'll need to create semantic textures. 

This will get complicated with Basis because we don’t want to compress semantics colors.

aclegg3:
We also want to enable programmatic setting of semantic IDs for objects added dynamically to the scene. Unifying the assets this way would make this much more straightforward.

erikwijmans:
For added objects, this is actually reasonably straightforward as the entire drawable has the same object ID.  This is what the objectId parameter in the GenericDrawable is for: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/GenericDrawable.h#L24

mosra:
> This will get complicated with Basis because we don’t want to compress semantics colors.

We don't need to use Basis for everything -- here a `R8Uint` format of a much smaller resolution than the color texture would suffice I think. For integer data Basis is bad choice anyway because all formats they support assume a normalized `[0; 1]` range.

mathfac:
Clarifying question: @msavva how do you find correspondence between semantic and original meshes for MP3D?

msavva:
Correspondence is mostly reasonable.  There are a few issues due to differences in geometry, but I believe that this projection will actually mostly help to improve the situation (by getting rid of floating geometry artifacts in the semantic mesh).

PR #341, opened 2019-11-05
Configurable physics object semantics

aclegg3:
## Motivation and Context
When adding arbitrary objects to the scene, users may want to configure the semantic ID as in issue #320.

With the current implementation we enable integer setting of the semanticId in the configuration file. This requires the user to know the mapping between semantic label and integer id. This should be improved in the future to allow specifying new or existing object classes instead of integer ids only.

This also changes `SceneNode::id_` assignment in `addComponent` to occur for all `SceneNode`'s associated with an object (even those not containing meshes). 

## How Has This Been Tested

Added a C++ test.

Local testing in Replica apartment_0 with `<object>.phys_properties.json` updated to specify:
- cheezit
  - `"semantic id": 10`
- banana
  - `"semantic id": 5`
- chefcan
  - `"semantic id": 2`

rgb_scene:
test rgba 00000

before object semantics:
test sem 00000

after object semantics:
test sem 00000

## Types of changes

## Checklist


aclegg3:
https://github.com/facebookresearch/habitat-sim/pull/413#discussion_r362987043

@msavva, @dhruvbatra, @mosra 

This discussion seems very relevant to this PR. 😄 

🤔 My understanding was that the `SceneNode::id_` is the semantic type identifier of the object. Therefore when a mesh is rendered with no vertex level semantic identifier, it is used to flat shade the mesh stored at the node. 

I arrived at this conclusion from this issue discussion: https://github.com/facebookresearch/habitat-sim/issues/336#issuecomment-549017586.

This PR is about exposing that value for modification and configuration. 

Big distinction here is that the `PhysicsManager::existingObjects_` is a map of individual object instances to integer ids (e.g., could be 10 of the same asset, but each would have a unique id here. When referring to these locally in the code, I usually call them `objectId`. However, all of these objects may have the same `SceneNode::id_`. If we want scene nodes to somehow refer back to their id in `PhysicsManager::existingObjects_`, then we should create a new id for this purpose and rename the existing `SceneNode::id_` used for semantic rendering to something more obvious unless there is another purpose for this value in the current implementation.

Open to thoughts! 😀

bigbike:
continue the discussion (#413) here:
>I don't know that it changes the guaranteed relationship between a rigid object and a SceneNode or the need to id them.
The 1:1 correspondence between a rigid object and a SceneNode is guaranteed. And each rigid object certainly needs an id.
My opinion is that you can put this 'id' in the `RigidObject`, not in the 'SceneNode'. physics-specific properties should not be added to the base class "SceneNode'.

bigbike:
Clarify: my reply is based on my current understanding:
- Only the physics-enabled objects have object ids. The other objects (e.g., the scene, the cameras) do not have ids.
- We give ids to objects, not scene graph nodes. So each scene graph node does not have an id. (the existing id_ in SceneNode is used to store semantic info.)

kylevedder:
Has there been any additional work on this PR or others to get object semantics for physics objects merged? This is a blocking issue for our work so we're willing to devote engineering resources to getting this feature added.

aclegg3:
> Has there been any additional work on this PR or others to get object semantics for physics objects merged? 

Hey @kylevedder, thanks for checking in on this. 👍 
We have recently picked back up on this feature as other users are also blocked here. We plan to complete this in the next week or two. Stay tuned.

aclegg3:
Purpose of this PR filled with #668. Closing.

PR #342, opened 2019-11-06
[clang-tidy] enable clang-analyzer-cplusplus.NewDeleteLeaks

msbaines:
* added SceneNodee:addFeature to avoid naked new

## Motivation and Context

Enable static analysis to check for new/delete leaks.

## How Has This Been Tested

Verified replica_v1 apartement_0 still loads with viewer.

## Types of changes

## Checklist


src/esp/scene/SceneNode.h
```diff
@@ -39,6 +39,13 @@ class SceneNode : public MagnumObject {
   SceneNodeType getType() { return type_; }
   void setType(SceneNodeType type) { type_ = type; }
 
+  // Add a feature. Used to avoid naked new ond make intent clearer.
```
mathfac:
```suggestion
  // Add a feature. Used to avoid naked `new` and makes intent clearer.
```

msbaines:
> Looks good. Maybe, worth to add more directories to clang-tidy scope first and then enable rules. Otherwise it will be painful to add new directories with all rules enabled. What do you think?

We currently run clang-tidy on all of src/esp. Was only planning on using it for our code and not deps. The only remaining directory to add it to is utils which is just two source files.

bigbike:
I really wish I could see this diff much earlier.
I need the reference to the newly created "features" (here, the drawables) for development. And such needs are not uncommon.

PR #343, opened 2019-11-07
Add FPS counter in utility viewer

bigbike:
As titled.
This PR is to display the runtime FPS on the screen, which can be rather useful when developing rendering related project (such as rendering acceleration project).

## Types of changes

## Checklist


bigbike:
@mosra : Currently, the program cannot be compiled and linked correctly. 

The error message:
```
CMake Error at /usr/local/Cellar/cmake/3.15.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Magnum (missing: TextureTools Text)
```

What is missing here?

Also, I would sincerely appreciate if you can share your advices on the code. (It is still under construction though, e.g., I need to compute the FPS to display).

mosra:
These libs are [disabled here](https://github.com/facebookresearch/habitat-sim/blob/38dadb63708c38dfae91015601f879f937c9f098/src/cmake/dependencies.cmake#L128-L131) because nothing needed them until now. Additionally, you'll need to

- enable `WITH_STBTRUETYPEFONT` and link to `MagnumPlugins::StbTrueTypeFont`
- bundle some TTF font and load it (be sure to check its license)
- call `updateText()` often enough (but not too often, so for example every 15th frame) to have the text updated, `timeline_.lastFrameTime()` should give you last frame time, so average that over the 15 frames for example

src/utils/viewer/viewer.cpp
```diff
@@ -105,6 +115,11 @@ class Viewer : public Magnum::Platform::Application {
   bool drawObjectBBs = false;
 
   Magnum::Timeline timeline_;
+
+  Cr::Containers::Pointer<Mn::Text::Renderer2D> performanceText_;
```
mathfac: Maybe worth to factor out Text drawing in separate tiny class that will have `init`, `draw`, `update` to keep viewer clean and make code re-usable for time when we replace `viewer` with simulator usage.

bigbike:
@mosra : Would you take a look when you have time on Mon.?
Let me know if I missed something here. Thanks a lot.

src/cmake/dependencies.cmake
```diff
@@ -127,8 +127,9 @@ if(NOT USE_SYSTEM_MAGNUM)
 
   # These are enabled by default but we don't need them right now -- disabling
   # for slightly faster builds. If you need any of these, simply delete a line.
-  set(WITH_TEXT OFF CACHE BOOL "" FORCE)
-  set(WITH_TEXTURETOOLS OFF CACHE BOOL "" FORCE)
+  # XXX
+  set(WITH_TEXT ON CACHE BOOL "WITH_TEXT" FORCE)
```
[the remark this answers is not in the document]
mosra: delete, those are enabled by default

src/cmake/dependencies.cmake
```diff
@@ -160,6 +161,7 @@ if(NOT USE_SYSTEM_MAGNUM)
   endif()
 
   if(BUILD_GUI_VIEWERS)
+    set(WITH_STBTRUETYPEFONT ON CACHE BOOL "WITH_STBTRUETYPEFONT" FORCE)
```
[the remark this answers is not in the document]
mosra: The docstring is only ever used when you're in CMake GUI or `ccmake` and hover the particular option. In Magnum itself these options [are documented to say which exact library they enable](https://github.com/mosra/magnum/blob/c4fe09f5dff544ecf9961fd760932f19d4c9fdb4/CMakeLists.txt#L79-L101) but here I don't see a reason to bother (and moreover, just copying the variable name to the docstring achieves nothing).

So when I started doing modifications to this file, I began using just `""` as it's shorter. But didn't want to add unrelated changes, so the original code stays as it was before.

src/utils/viewer/viewer.cpp
```diff
@@ -515,9 +574,19 @@ void Viewer::keyPressEvent(KeyEvent& event) {
   }
   renderCamera_->node().setTransformation(
       rgbSensorNode_->absoluteTransformation());
+
+  // XXX This will cause segfault
+  updateText();
```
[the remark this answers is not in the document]
mosra: backtrace?

src/utils/viewer/resources.conf
```diff
@@ -0,0 +1,4 @@
+group=fonts
+
+[file]
+filename=DejaVuSans.ttf
```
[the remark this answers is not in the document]
mosra: This one is pretty big, Source Sans Pro [on Google Fonts](https://fonts.google.com/specimen/Source+Sans+Pro) is only ~110 kB if you pick just basic Latin.

src/utils/viewer/viewer.cpp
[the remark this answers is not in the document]
bigbike: @mosra : The bt is as follows:

I noticed that it says "TextureTools::atlas(): requested atlas size Vector(512, 512) is too small to fit 70 Vector(90, 108) textures. Generated atlas will be empty.". Would the segfault be caused by this?

```
WARNING: Logging before InitGoogleLogging() is written to STDERR
I1111 12:19:44.038406 16127424 viewer.cpp:210] Loading navmesh from /Users/yili/models/data/scene_datasets/habitat-test-scenes/skokloster-castle.navmesh
TextureTools::atlas(): requested atlas size Vector(512, 512) is too small to fit 70 Vector(90, 108) textures. Generated atlas will be empty.
I1111 12:19:47.641803 16127424 viewer.cpp:221] position=0-5.14033 0.0099186 0024.1693 rotation=0000000000 -0.0980171 0000000000 000.995185
I1111 12:19:47.641839 16127424 viewer.cpp:223] Distance to closest obstacle: 0.19586
viewer was compiled with optimization - stepping may behave oddly; variables may not be available.
Process 29100 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x30)
    frame #0: 0x00000001000557af viewer`Magnum::Text::AbstractFont::lineHeight() const [inlined] Magnum::Text::AbstractFont::isOpened(this=0x0000000101033e40) const at AbstractFont.h:281:40 [opt]
   278          #endif
   279
   280          /** @brief Whether any file is opened */
-> 281          bool isOpened() const { return doIsOpened(); }
   282
   283          /**
   284           * @brief Open font from raw data
Target 0: (viewer) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x30)
  * frame #0: 0x00000001000557af viewer`Magnum::Text::AbstractFont::lineHeight() const [inlined] Magnum::Text::AbstractFont::isOpened(this=0x0000000101033e40) const at AbstractFont.h:281:40 [opt]
    frame #1: 0x00000001000557ac viewer`Magnum::Text::AbstractFont::lineHeight(this=0x0000000101033e40) const at AbstractFont.cpp:229 [opt]
    frame #2: 0x0000000100053b42 viewer`Magnum::Text::(anonymous namespace)::renderVerticesInternal(font=0x0000000101033e40, cache=0x00007ffeefbff438, size=0, text="FPS: 6e+01", alignment=TopRight) at Renderer.cpp:79:53 [opt]
    frame #3: 0x0000000100054963 viewer`Magnum::Text::AbstractRenderer::render(this=0x0000000101032d10, text=<unavailable>) at Renderer.cpp:364:40 [opt]
    frame #4: 0x000000010000cf63 viewer`(anonymous namespace)::Viewer::updateText(this=<unavailable>) at viewer.cpp:587:21 [opt]
    frame #5: 0x0000000100008759 viewer`(anonymous namespace)::Viewer::keyPressEvent(this=0x00007ffeefbff170, event=<unavailable>) at viewer.cpp:579:3 [opt]
    frame #6: 0x00000001004a4482 viewer`Magnum::Platform::GlfwApplication::setupCallbacks()::$_4::__invoke(GLFWwindow*, int, int, int, int) [inlined] Magnum::Platform::GlfwApplication::setupCallbacks(window=<unavailable>, key=262, (null)=0, action=1, mods=0)::$_4::operator()(GLFWwindow*, int, int, int, int) const at GlfwApplication.cpp:507:17 [opt]
    frame #7: 0x00000001004a4455 viewer`Magnum::Platform::GlfwApplication::setupCallbacks(window=<unavailable>, key=262, (null)=<unavailable>, action=1, mods=0)::$_4::__invoke(GLFWwindow*, int, int, int, int) at GlfwApplication.cpp:501 [opt]
    frame #8: 0x00000001004af037 viewer`-[GLFWContentView keyDown:](self=0x000000010162ae70, _cmd=<unavailable>, event=0x000000010143c7f0) at cocoa_window.m:663:5 [opt]
    frame #9: 0x00007fff36839367 AppKit`-[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:] + 6840
    frame #10: 0x00007fff36837667 AppKit`-[NSWindow(NSEventRouting) sendEvent:] + 478
    frame #11: 0x00007fff366d7889 AppKit`-[NSApplication(NSEvent) sendEvent:] + 2953
    frame #12: 0x00000001004afa62 viewer`-[GLFWApplication sendEvent:](self=0x0000000101119c80, _cmd=<unavailable>, event=0x000000010143c7f0) at cocoa_window.m:904:9 [opt]
    frame #13: 0x00000001004b18e7 viewer`_glfwPlatformPollEvents at cocoa_window.m:1590:9 [opt]
    frame #14: 0x00000001004a3ca5 viewer`Magnum::Platform::GlfwApplication::exec(this=0x00007ffeefbff170) at GlfwApplication.cpp:601:9 [opt]
    frame #15: 0x0000000100005a36 viewer`main(argc=2, argv=<unavailable>) at viewer.cpp:592:1 [opt]
    frame #16: 0x00007fff650383d5 libdyld.dylib`start + 1
```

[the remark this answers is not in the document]
mosra: Looks to be because the `font` passed to the `Text::Renderer2D` constructor gets deleted at the end of Viewer constructor. You need to save it as a class member, together with the `fontManager`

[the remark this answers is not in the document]
mosra: And re the warning message -- try to have the font size only e.g. 16, not 110. 110 was needed for the text example because there it's *huge*, here you don't need that much. Play with the value until it looks sharp on the screen -- best results are when the font size is exactly 2x larger than what appears on the screen (so, 2x supersampled).

bigbike:
It should work now. This diff is ready to be reviewed.

image

src/utils/viewer/viewer.cpp
```diff
@@ -352,14 +406,32 @@ void Viewer::drawEvent() {
...
+      .setColor(Color3{1.0f})
+      .bindVectorTexture(cache_.texture());
+  performanceText_->mesh().draw(textShader_);
+
   swapBuffers();
-  timeline_.nextFrame();
   redraw();
+
+  totalFrames_++;
+  if ((totalFrames_ & fpsUpdateFrequencyMask) == 0) {
```
erikwijmans: Why not 
```suggestion
  if ((totalFrames_ % fpsUpdateFrequencyMask) == 0) {
```

src/utils/viewer/viewer.cpp
erikwijmans: I second this.  Will make it much easier in the long run.

bigbike: This is the 1st time to introduce it to the system, and it is non-trivial (as you can see it in this PR.)
I will have to make it work first.
The further improvement will be in the future.

src/utils/viewer/viewer.cpp
bigbike: Because it is a bit faster.

src/utils/viewer/viewer.cpp
erikwijmans: I get doing things like this to make it work, but I don't agree with not fixing it up once it's working.  The "and it is non-trivial (as you can see it in this PR.)" further reinforces that it should be compartmentalized so when someone decided to refactor the viewer (and it's definitely overdue for one), they don't have to go through what you went through to get it working again.

bigbike: @erikwijmans :
I appreciate your comments, which makes a lot of sense. Please do not get me wrong. I certainly pay attention to the quality of the code.

- Code to display text on GUI is even more non-trivial if written with raw OpenGL from scratch. Magnum library already encapsulates and hides a large amount of tedious details, and provided high quality APIs to be used here. So the code (implemented in this PR) may  not be that ugly to start with, first of all.

- "not fixing it up once it's working." I never say or imply that I would not "fix" it. Actually "improve" would be a more proper word here, since there is nothing wrong with the current implementation. I said "further improvement is in the future". It may just not be in this PR. 

- Why not make the improvement in this PR? I need to clarify that I am quite open to this suggestion. But it will slow me down. This is the 1st PR in the "rendering acceleration" project, and it is the simplest one. My plan was to make a quick PR (to be committed 1-2 days) and put my focus on the following challenging work, which needs much more time. I have a lot of experience on this topic and I know which part is the most difficult one. So I have to prioritize the work, and invest my time wisely given the limited time budget (to hit the Q4 goal).

- I cared about the quality of the code. Actually because of this reason, I was thinking to postpone the improvement because it needs a careful design, and it takes time. I indeed thought about it last week. The core idea of the improvement is to create a higher level, helper class to further encapsulate and drive the fontManager, font, cache, Renderer2D etc. Two ways to do it. One is to do it in the Magnum library, and the other is to do it in our libraries. (I was planning to hear the opinions from Vladimir first). Either way relates to a lot of detailed design decisions, user APIs, location of the fonts etc.

Again, I am willing to do the improvement here, if eventually we agree it must be done in this PR. The above bulletin points are simply to let you understand a bit more about my situation and thoughts.

Thanks.

src/utils/viewer/viewer.cpp
mosra: I don't think that matters in this context. In a tight loop over thousands of elements probably yes, but here any additional latency from the modulo op is negligible compared to function call overhead, VSync waiting and such. I'd vote for having the modulo and not requiring this to be `2^n - 1`, as that's simpler.

src/utils/viewer/viewer.cpp
mosra: Eh. What about this, then -- I'll briefly pause my work on texture compression / asset import and create some `DebugTools::FpsCounter` thing that wraps this up into an "oneliner" with a similar convenience/flexibility tradeoffs like [DebugTools::screenshot()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1DebugTools.html#a66cd81ac668af7f8dffcd27ee47b2019) we're already using in the viewer.

That means all work put into this so far will be thrown away, but if that saves you from spending even more time on it, why not. And I was planning to implement such a thing anyway, only you needed it sooner than I could have done it.

src/utils/viewer/viewer.cpp
bigbike: I learned it a couple of years ago when implementing the lock-less ring buffer. It is a bit faster and will benefit e.g., a server program which needs to run for long time (no need to be in a tight loop). It comes natural to me, so I did not think too much about it when I wrote the code.

If you guys do not like it, I will change it using modulo. This program is not that time-critical after all. I agree.

src/utils/viewer/viewer.cpp
bigbike: @mosra : I appreciate the help you offered. Thank you. 
Actually I knew you are busy with the texture staff, so I started it by myself.

What I really need is a generic "text display" class, much more than a "FpsCounter". Besides FPS, it will also help me to display e.g., how many drawables or triangles are culled later. I think such an advanced class (a wrapper) will be built on top of the fontManager, font, cache, projMtx etc (as I mentioned in previous comments) to simplify the operations here. Like I say one proper way is to do it within magnum, and the other way is to do it in our lib. Does it make sense to you? 

Let me know if you are going to do it or let me do it. I am happy to.

bigbike: @mosra: I am going to code it up. You can focus on your current work. :-)

src/utils/viewer/viewer.cpp
```diff
@@ -105,6 +120,16 @@ class Viewer : public Magnum::Platform::Application {
   bool drawObjectBBs = false;
 
   Magnum::Timeline timeline_;
+
+  // performance counter
+  uint64_t totalFrames_ = 0;
+  float timeDuration_ = 0.0f;
+  Cr::Containers::Pointer<Mn::Text::Renderer2D> performanceText_;
+  Mn::PluginManager::Manager<Mn::Text::AbstractFont> fontManager_;
```
bigbike: Bug: this is Cr::PluginManager.
I am wondering why the compiler never complain it.

src/utils/viewer/viewer.cpp
```diff
@@ -515,9 +587,15 @@ void Viewer::keyPressEvent(KeyEvent& event) {
   }
   renderCamera_->node().setTransformation(
       rgbSensorNode_->absoluteTransformation());
+
   redraw();
 }
 
+void Viewer::updateText(float fps) {
+  // update the text occasionally
+  performanceText_->render(Cr::Utility::formatString("FPS: {:.1f}", fps));
```
bigbike: @mosra:
I do not see the description of this function in your doc, btw.
https://doc.magnum.graphics/magnum/classMagnum_1_1Text_1_1Renderer.html#pub-static-methods

bigbike: Oh, never mind. It is here:
https://doc.magnum.graphics/magnum/classMagnum_1_1Text_1_1AbstractRenderer.html#a1730154b16f3884ef10b29b7582a759b

bigbike:
- built a advanced class on top of fontManager, font, cache, renderer2D, projectMtx;
- fonts were placed in a folder "fonts" at the same level as "shaders";
- applied the library in the viewer;
- tested it with different character colors;
image

bigbike:
@mosra :
- About alignment: I found only "TopRight" worked here. I tried "TopLeft" and hoped the text was shown on top left of the window but it gave me nothing instead. "CenterLeft" did not work as well. The text was shifted a bit to the right, as I could only see "FPS", and the numbers were out of the screen. So certainly it did not behave as what I expected.

I read your doc about the alignment. For example:
```TopLeft: Text start and top is at origin```
What does it mean? origin of what coordinate system?

**Comment by bigbike** [comment]
Also I noticed something is rather confusing in the CI test. The error message reads:
```
-- Could NOT find Magnum (missing: GlfwApplication Sdl2Application WindowlessGlxApplication WindowlessWglApplication) 
Building with ptex support
-- Found Corrade: /home/circleci/project/habitat-sim/src/deps/corrade/src  found components:  Containers rc Utility 
-- Found Corrade: /home/circleci/project/habitat-sim/src/deps/corrade/src  found components:  Containers rc Utility PluginManager 
-- Found Magnum: /home/circleci/project/habitat-sim/src/deps/magnum/src  found components:  Trade GL TextureTools AnyImageImporter AnySceneImporter MeshTools SceneGraph Shaders AnyImageConverter Text 
-- Found Magnum: /home/circleci/project/habitat-sim/src/deps/magnum/src  found components:  Trade TextureTools GL AnyImageImporter Text 
CMake Error at cmake/FindMagnumPlugins.cmake:401 (file):
  file failed to open for reading (No such file or directory):

    /home/circleci/project/habitat-sim/src/deps/magnum-plugins/src/MagnumPlugins/StbTrueTypeFont/configure.h
Call Stack (most recent call first):
  esp/gfx/CMakeLists.txt:54 (find_package)


CMake Error at /opt/cmake/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find MagnumPlugins (missing: StbTrueTypeFont)
Call Stack (most recent call first):
  /opt/cmake/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindMagnumPlugins.cmake:435 (find_package_handle_standard_args)
  esp/gfx/CMakeLists.txt:54 (find_package)
```

**Comment by bigbike** [comment]
Tried to compile it on the linux devserver. Same error.

File: src/utils/viewer/viewer.cpp (hunk 1 above)
Comment (mosra) [reply src/utils/viewer/viewer.cpp]: It would be first just for FPS, later extended with info coming from [ARB_pipeline_statistics_query](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_pipeline_statistics_query.txt) etc., possibly plotting a graph of the values over time. Once I'm done with the asset pipeline rework, I'll need it "for marketing purposes" to monitor resource use before/after :)

File: src/utils/viewer/viewer.cpp (hunk 7 above)
Comment (mosra) [reply src/utils/viewer/viewer.cpp]: Because of https://github.com/mosra/magnum/blob/98232f383adc1ca1fdce877bc426951c23a89521/src/Magnum/Magnum.h#L40

File: src/utils/viewer/viewer.cpp [hunk 9]
```diff
@@ -352,14 +373,27 @@ void Viewer::drawEvent() {
     physicsManager_->debugDraw(projM * camM);
   }
 
+  if (displayPerformanceInfo) {
+    Mn::Color3 textColor{1.0f};
+    textRenderer_->draw(textColor);
```
mosra: `textRenderer_->draw(0xffffff_rgbf)`, since we have the hexadecimal / CSS color literals?

src/utils/viewer/viewer.cpp
```diff
@@ -210,17 +224,21 @@ Viewer::Viewer(const Arguments& arguments)
   renderCamera_->node().setTransformation(
       rgbSensorNode_->absoluteTransformation());
 
-  timeline_.start();
+  // fonts
+  textRenderer_.reset(new gfx::TextRenderer(
+      Mn::Vector2(Mn::GL::defaultFramebuffer.viewport().size()).aspectRatio()));
```
mosra:
```suggestion
      Mn::Vector2(framebufferSize()).aspectRatio()));
```

(I still need to update all docs and example code to use this shorter variant, sorry.)

src/utils/viewer/viewer.cpp
```diff
@@ -105,6 +113,12 @@ class Viewer : public Magnum::Platform::Application {
   bool drawObjectBBs = false;
 
   Magnum::Timeline timeline_;
+
+  // performance counter
+  bool displayPerformanceInfo = true;
+  uint64_t totalFrames_ = 0;
+  float timeDuration_ = 0.0f;
```
mosra: What is this for? Never used, as far as I could see.

src/utils/viewer/viewer.cpp
```diff
@@ -510,14 +544,23 @@ void Viewer::keyPressEvent(KeyEvent& event) {
                                          drawObjectBBs);
       }
     } break;
+    case KeyEvent::Key::One: {
+      displayPerformanceInfo = !displayPerformanceInfo;
```
mosra: Bit trick, if you want: 

```suggestion
      displayPerformanceInfo ^= true;
```

src/utils/viewer/viewer.cpp
```diff
@@ -352,14 +373,27 @@ void Viewer::drawEvent() {
...
+    textRenderer_->draw(textColor);
+  }
+
   swapBuffers();
-  timeline_.nextFrame();
   redraw();
+
+  totalFrames_++;
+  if (displayPerformanceInfo && (totalFrames_ % fpsUpdateFrequency) == 0) {
+    updateText(totalFrames_ / timeline_.previousFrameTime());
```
mosra: :thinking: so ... let's say you have 1500 `totalFrames_` and previous frame time was 0.016 seconds. That'll give you a FPS of 93750. Or am I missing something?

Also, what about some moving average? That's what I originally meant when I said "update every 15 frames".

src/esp/gfx/TextRenderer.h
```diff
@@ -0,0 +1,86 @@
...
+                        float fontSize = 20.0f,
+                        int cacheSize = 512,
+                        const std::string& characters = defaultCharacters);
+
+  /**
+   * @brief Create a renderer to render text on screen
+   * @return renderer ID
+   */
+  size_t createRenderer(
+      float onScreenCharacterSize = 0.08f,
```
mosra: I would make the characters a bit smaller -- `0.04f`? (but keeping the glyph cache texture size a bit higher, as supersampling will make the text look crisper).

src/esp/gfx/TextRenderer.cpp
```diff
@@ -0,0 +1,112 @@
...
+  CORRADE_ASSERT(glyphCount > 0,
+                 "TextRenderer::createRenderer: the copacity" << glyphCount
+                                                              << "is illegal.",
+                 0);
+
+  text_.emplace_back(
+      std::move(Cr::Containers::Pointer<Mn::Text::Renderer2D>()));
+  size_t rendererId = text_.size() - 1;
+  text_[rendererId].reset(new Mn::Text::Renderer2D(
+      *font_, cache_, onScreenCharacterSize, alignment));
```
mosra: Why not just

```cpp
text_.emplace_back(Cr::Containers::pointer<Mn::Text::Renderer2D>(*font_, cache_, onScreenCharacterSize, alignment));
```

No need to `std::move()` or `reset()` anything.

src/esp/gfx/TextRenderer.h
```diff
@@ -0,0 +1,86 @@
...
+   * information.
+   */
+  void updateText(const std::string& text, size_t rendererId = 0);
+
+ protected:
+  Cr::PluginManager::Manager<Mn::Text::AbstractFont> fontManager_;
+  // TODO: we can build an array to support multiple fonts
+  Cr::Containers::Pointer<Mn::Text::AbstractFont> font_;
+  Mn::Text::GlyphCache cache_;
+  Mn::Shaders::Vector2D textShader_;
```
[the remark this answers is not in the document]
bigbike: The TextRenderer class in current shape is quite self-contained and thus independent.
Your comments got me thinking if I should put the entire class to the resource manager. Would the fonts be the assets?
The shaders in the resource manager were used to render the assets -- meshes. So if we consider the fonts to be an asset, then yes, we need to put the shader (even the entire class) in the resource manager. But I think that makes things complicated.

I put them here simply because after the fonts are loaded, we immediately build the bitmap font for the selected characters, and we implement rendering functions for them, all of which are related to the "gfx".

src/esp/gfx/TextRenderer.cpp
bigbike: Oh, thanks. I tried to run fast, so you will see a lot of place in this code, where I just followed the example program in the doc and without changing too much.

src/esp/gfx/TextRenderer.cpp
```diff
@@ -0,0 +1,112 @@
...
+TextRenderer& TextRenderer::draw(const Mn::Color3& color, size_t rendererId) {
+  CORRADE_ASSERT(
+      rendererId < text_.size(),
+      "TextRenderer::draw: the rendererId" << rendererId << "is out of range.",
+      *this);
+  textShader_
+      .setTransformationProjectionMatrix(
+          textProjection_ *
+          Mn::Matrix3::translation(
+              1.0f / textProjection_.rotationScaling().diagonal()))
```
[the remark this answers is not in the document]
bigbike: Again, this is copy+paste from your example program.

src/esp/gfx/TextRenderer.h
```diff
@@ -0,0 +1,86 @@
...
+class TextRenderer {
+ public:
+  /**
+   * @brief Constructor
+   */
+  // aspectRatio = width / height
+  explicit TextRenderer(float viewportAspectRatio = 1.33333f,
+                        float fontSize = 20.0f,
+                        int cacheSize = 512,
+                        const std::string& characters = defaultCharacters);
```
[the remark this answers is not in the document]
bigbike: Sorry, my bad. Yes, it is amateur. It will become a disaster when the header is included by multiple .cpp.
Good catch. I wish I can came up with the same idea as yours to prevent such silly mistake.

src/esp/gfx/TextRenderer.h
```diff
@@ -0,0 +1,86 @@
...
+      float onScreenCharacterSize = 0.08f,
+      size_t glyphCount = 60,
+      Mn::Text::Alignment alignment = Mn::Text::Alignment::TopRight);
+
+  /**
+   *  @brief Draw the text by the specified renderer (default 0) with specified
+   * color
+   *  @return Reference to self (for method chaining)
+   */
+  TextRenderer& draw(const Mn::Color3& color, size_t rendererId = 0);
```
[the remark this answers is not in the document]
bigbike: OK, will do.

src/esp/gfx/TextRenderer.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   *  @brief Update the aspect ratio of the viewport
+   *  @return Reference to self (for method chaining)
+   *  @notification DO NOT forget to call it in the viewportEvent()
```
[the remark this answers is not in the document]
bigbike: OK, will change.

src/utils/viewer/viewer.cpp
bigbike: Oh, will remove. 
This is because I read your previous message, and tried to introduce a variable to sum up the time elapsed since start() was called. But it is not needed.

src/utils/viewer/viewer.cpp
bigbike: Somehow I found it from your docs.
Will change.

src/utils/viewer/viewer.cpp
bigbike: OK, will change.

src/utils/viewer/viewer.cpp
bigbike: You mixed ```previousFrameTime()``` and ```previousFrameDuration()```. 
```
previousFrameTime 
Time at previous frame (in seconds)
Returns time elapsed since start() was called. If the timeline is stopped, the function returns 0.0f.
```
So here it gives you the average FPS ever since the simulator runs, which is not what I meant.
Good catch. I need to change it.

src/utils/viewer/viewer.cpp
bigbike: Hahaha. :)

src/esp/gfx/TextRenderer.h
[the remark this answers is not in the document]
bigbike: The TextRenderer class in current shape is quite self-contained and thus independent.
Your comments got me thinking if I should put the entire class to the resource manager. Would the fonts be the assets?
The shaders in the resource manager were used to render the assets -- meshes. So if we consider the fonts to be an asset, then yes, we need to put the shader (even the entire class) in the resource manager. But I think that makes things complicated.

I put them here simply because after the fonts are loaded, we immediately build the bitmap font for the selected characters, and we implement rendering functions for them, all of which are related to the "gfx".

I will keep this class here until we decide it really needs to be moved to the resource manager.

src/esp/gfx/TextRenderer.cpp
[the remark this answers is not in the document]
bigbike: > > Also I noticed something is rather confusing in the CI test.
> 
> It's because `WITH_STBTRUETYPEFONT` is enabled only if `BUILD_GUI_VIEWERS` is enabled ;)


I checked my code where I set it when BUILD_GUI_VIEWERS is enabled.
```
if(BUILD_GUI_VIEWERS)
    set(WITH_STBTRUETYPEFONT ON CACHE BOOL "WITH_STBTRUETYPEFONT" FORCE)
```
When I tried to compile it on linux, I used ```build.sh --headless```.  Wouldn't it set the BUILD_GUI_VIEWERS to false?

src/utils/viewer/viewer.cpp
```diff
@@ -352,14 +373,30 @@ void Viewer::drawEvent() {
...
+  if (displayPerformanceInfo) {
+    textRenderer_->draw(0xffffff_rgbf);
+  }
+
   swapBuffers();
-  timeline_.nextFrame();
   redraw();
+
+  totalFrames_++;
+  timeDuration_ += timeline_.previousFrameDuration();
```
bigbike: Note: now I used the `previousFrameDuration` not the `previousFrameTime`.
I simply compute the average FPS for every `fpsUpdateFrequency`. 

You would like to have a moving window (window size = fpsUpdateFrequency)? Then I have to implement a small ring buffer (buffer size = fpsUpdateFrequency), and more importantly, that means I need to update the text every frame (otherwise, if I update it in every `fpsUpdateFrequency`, this is what I am doing here.)

src/esp/gfx/TextRenderer.cpp
```diff
@@ -0,0 +1,141 @@
...
+TextRenderer& TextRenderer::draw(const Mn::Color3& color, size_t rendererId) {
+  CORRADE_ASSERT(
+      rendererId < text_.size(),
+      "TextRenderer::draw: the rendererId" << rendererId << "is out of range.",
+      *this);
+  auto entry = locations_[rendererId];
+  textShader_
+      .setTransformationProjectionMatrix(
+          textProjection_ *
+          Mn::Matrix3::translation(
```
bigbike: > Why so complicated? Why not, for example, have the text projection equal to window size and the text size in pixels? That'll make sure the text size stays the same when you resize / maximize the window.

Sorry, I cannot understand the sentence: `have the text projection equal to window size and the text size in pixels?`. Based on what you said later, I assume you would like to display the text in fixed size (no matter what the window size is, the text size is fixed.) Can you give me the expression of the projection matrix? (this is the most straightforward way.)

If I understand you correctly, then what if the window size becomes tiny even smaller than the text size?

src/esp/gfx/TextRenderer.h
```diff
@@ -0,0 +1,101 @@
...
+
+class TextRenderer {
+ public:
+  /**
+   * @brief Constructor
+   */
+  // aspectRatio = width / height
+  explicit TextRenderer(float viewportAspectRatio = 1.33333f,
+                        float fontSize = 20.0f,
+                        int cacheSize = 1024);
```
bigbike: Is this the cache texture size you mentioned in your previous comments? 
Or the fontSize above it?

src/esp/gfx/TextRenderer.cpp
```diff
@@ -0,0 +1,141 @@
...
+      alignment = Mn::Text::Alignment::TopRight;
+      break;
+    default:
+      Cr::Utility::Fatal{-1} << "TextRenderer::createRenderer: The location"
+                             << "is not supported.";
+      break;
+  }
+  text_.emplace_back(
+      Cr::Containers::Pointer<Mn::Text::Renderer2D>(new Mn::Text::Renderer2D(
+          *font_, cache_, onScreenCharacterSize, alignment)));
```
bigbike: Thank you for your explanation in the previous comments.
Now I understand e.g., the TopLeft means the topleft corner of the text is located at the origin. And more importantly, the origin (0, 0) is originally at the center of the window. So the TopLeft is (-1, 1), TopRight is (1, 1), BottomLeft is (-1, -1) and BottomRight is (1, -1).

bigbike:
@msbaines : The iamGUI looks very pretty. We are using glfw application framework here. Changing it to iamGUI is another topic other than this PR.

This PR is to leverage the existing text libraries in magnum to display some real-time information on the screen.

bigbike:
> Removing my objection. Talked to Yili offline. I didn't realize the urgency of this work.
> 
> Long term, I think we should introduce something like imgui to enable visualizations and the ability to reconfigure at run-time.

Thank you for the understanding. 

Actually, what you proposed here makes a lot of sense. I like the demos from the iamGUI, which is pretty. We need to consider using it in near future.


(I should correct myself: if we employ iamgui, it does not replace the glfw. Instead it works with it.)

src/utils/viewer/viewer.cpp
mosra: Yeah, that's what I'm saying :) most of the occurences where this is used was written *before* `framebufferSize()` was a thing, so I need to update all those docs.

src/esp/gfx/TextRenderer.h
mosra: Can't tell what are the right values without playing with them myself :) You have two degrees of freedom here:

- size of the pre-rendered glyphs (and cache size is just large enough to contain them all, no point in having it too large ... I guess maybe even 256x256 could be enough)
- size of the text on-screen

These are orthogonal (size of one doesn't affect size of the other), for given size of the text on screen you basically need to pick large enough `fontSize` to have it rendered not too blurry but also not too aliased.

src/utils/viewer/viewer.cpp
mosra: Yeah, that's why I suggested to have all this implemented on the side of Magnum -- because it's not feasible to have ring buffer, Kahan Sum, moving average and all that implemented & tested here just for the needs of a FPS counter of a viewer utility.

Let's keep it as you have it now, we can always revisit later.

src/esp/gfx/TextRenderer.cpp
mosra: I cleaned up and updated the text example to show this better -- please see https://github.com/mosra/magnum-examples/blob/91faa149da5def876558ea93d1ac47d3df0f37f3/src/text/TextExample.cpp#L105-L113

src/cmake/dependencies.cmake
```diff
@@ -160,6 +155,7 @@ if(NOT USE_SYSTEM_MAGNUM)
   endif()
 
   if(BUILD_GUI_VIEWERS)
+    set(WITH_STBTRUETYPEFONT ON CACHE BOOL "" FORCE)
```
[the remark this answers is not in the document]
mosra: Well but that's what I'm saying. With `./build.sh --headless`, `BUILD_GUI_VIEWERS` is `OFF`, which means `WITH_STBTRUETYPEFONT` is not `ON`, which means the `gfx` library depends on `StbTrueTypeFont` that's not there.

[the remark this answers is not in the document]
bigbike: Ah, I see. Thanks.
That means I need to move `WITH_STBTRUETYPEFONT` out of the if clause.

src/esp/gfx/TextRenderer.cpp
bigbike: I see. Thanks a lot.
The example code is much clearer than before.
When you say "mesh origin", I understand the "mesh" is for the 2D text, not the model in the scene. (I know it sounds silly, but for readers who have no ideas how the text are displayed using modern OpenGL, it can be confusing. They do not know there is a "mesh" for a character.)

src/esp/gfx/TextRenderer.h
bigbike: I see. so no need to set the cacheSize so big, and whether the "text" look sharp or not depends on the `fontSize` here and the text size on screen (set later).

src/utils/viewer/viewer.cpp
bigbike: Make sense.
By the way, in my previous projects, we simply use double to increase the precision, and never think Kahan summation is needed. Maybe it is because my old projects were computation heavy and thus the FPS was too low. :-) Have you experienced any case that it is a must-have? Just curious. I hope to understand and learn.

src/utils/viewer/viewer.cpp
bigbike: This should be double, by the way.

src/esp/gfx/TextRenderer.cpp
mosra: By *mesh* I mean the `GL::Mesh` that's populated by the `render()` function and later drawn, yes.

src/utils/viewer/viewer.cpp
mosra: Probably not so important in this case where we have a dozen values to average each time, but would definitely make a difference when calculating an average FPS over the whole application run time -- with floats, it'd start losing precision in the microsecond range already after 15 minutes. With other measured quantities such as sample fill rate or primitive emission it could be even sooner. To give you a perspective, [in this test](https://github.com/mosra/magnum/blob/debddf658571cd2a653c6ffe82bc39f245a51ba7/src/Magnum/Math/Algorithms/Test/KahanSumTest.cpp#L87-L100) I sum `1.0f` 100 million times and while Kahan Sum gives me 1.0e8, the naive approach *stops* at 1.66667e7 and doesn't go higher because the newly added values get rounded off during addition.

Doubles would be sufficient for precise calculations when the app is running days, but not much more I think.

wojga:
n00b question: is there really no off-the-shelf text renderer that we could use to overlay text as part of some 2D post-processing pipeline?
Does @msbaines suggestion to use imgui make sense?

mosra:
@wojga ImGui is the most effortless way to achieve this, and besides text it allows you to have similarly effortless toggles, buttons etc. It's not the prettiest, but it's the most versatile.

Magnum's text rendering was a bit neglected over the past years, so the APIs are not as simple to use as they could be -- especially regarding sane defaults or having some "good enough" builtin font. But I'd say it's definitely easier to use the `Text` library than dragging in a random third party lib for text rendering. Let me know if I should prioritize coding up some easy-to-integrate way to display FPS and other things (like I mentioned above) to minimize the amount of code needed here -- until then I'm keeping my focus on the primary tasks :)

bigbike:
@wojga : Take a look at #355. We are not against using iamgui. Actually we will close this PR when #355 is committed in.

When I started this PR, I tried not to introduce another 3rd party library to the system, just improving the text rendering using the existing Magnum modules. The reason was the same as @mosra described.

bigbike:
Close it since we have #355.

PR #346, opened 2019-11-11
Fix action space path

### LINKED ISSUE #111: --compute_action_shortest_path says method not found
Hello, I'm trying to use example.py to make the agent save png image while traversing the shortest path to the goal position. When I try to use the option --compute_shortest_path, it says in the demo_runner.py file (line 260), the simulator object has no attribute named "make_action_pathfinder(agent_id)". And i indeed cannot find the method. I'm wondering is this a bug to be fixed? or is there any other way to accomplish what I want to do with the example.py? 

Thank you!

erikwijmans:
## Motivation and Context

Actually hook-up the greedy follower to the action space path part of the example (this never got done for some reason).

Fixes #111 

## How Has This Been Tested

Via `python examples/example.py --compute_action_shortest_path`

## Types of changes


erikwijmans:
Added a test

mathfac:
Surprised that PR decreases coverage, [the follower looks pretty uncovered in master](https://codecov.io/gh/facebookresearch/habitat-sim/src/master/habitat_sim/nav/greedy_geodesic_follower.py).

mosra:
Looks like it counts `import` lines into coverage, and one of these got removed, which makes it a net negative :)

erikwijmans:
The follower should be really well covered in master: https://github.com/facebookresearch/habitat-sim/blob/master/tests/test_greedy_follower.py

This PR is just hooking it up correctly in the example script

PR #348, opened 2019-11-12
Save 25% of memory usage on vertex position buffer (PTex Mesh)

bigbike:
## Motivation and Context
As titled.

Using Vector3 instead of Vector4 for a vertex position in PTex mesh

Reduce memory footprint 

## How Has This Been Tested
local test with utility viewer.

image

## Types of changes

## Checklist


bigbike:

erikwijmans:
Will the shader need to change here: https://github.com/facebookresearch/habitat-sim/blob/master/src/shaders/ptex-default-gl410.vert#L1 ?

bigbike:
> Will the shader need to change here: https://github.com/facebookresearch/habitat-sim/blob/master/src/shaders/ptex-default-gl410.vert#L1 ?

Oh, this is because OpenGL will fill the 4th component, w, with 1.0 automatically for the user if only a 3-dimensional vertex position is provided. I did not realize it would cause confusion here.

If you really feel bad about it, maybe I can change it to vec3. Then the following code will be modified as:
```
gl_Position = MVP * vec4(position, 1.0);
```
@mosra: Do you think it is necessary?

mosra:
No, keep the `position` as `vec4`. That's what I'm doing [for magnum's builtin shaders](https://github.com/mosra/magnum/blob/98232f383adc1ca1fdce877bc426951c23a89521/src/Magnum/Shaders/Flat3D.vert#L64) as well.

src/esp/assets/PTexMeshData.cpp
```diff
@@ -665,7 +665,7 @@ void PTexMeshData::parsePLY(const std::string& filename,
                    "not greater than 0", );
   }
 
-  meshData.vbo.resize(numVertices, vec4f(0, 0, 0, 1));
+  meshData.vbo.resize(numVertices, vec3f(0, 0, 0));
```
msbaines: You should assert that positionDimensions < 4.

src/esp/assets/PTexMeshData.h
```diff
@@ -22,7 +22,7 @@ namespace assets {
 class PTexMeshData : public BaseMesh {
  public:
   struct MeshData {
-    std::vector<vec4f> vbo;
+    std::vector<vec3f> vbo;
     std::vector<vec4f> nbo;
```
msbaines: Couldn't nbo also be a vec3?

erikwijmans:
> Oh, this is because OpenGL will fill the 4th component, w, with 1.0 automatically for the user if only a 3-dimensional vertex position is provided. I did not realize it would cause confusion here.

Ah, I didn't realize OpenGL would do that.  How nice of it :)

No need to change it as people will rarely actually look at the shaders themselves, I just wanted to check for my own knowledge.

src/esp/assets/PTexMeshData.h
mosra: As far as I can see, neither `nbo` nor `cbo` is used right now (only populated). Is that correct?

bigbike: > As far as I can see, neither `nbo` nor `cbo` is used right now (only populated). Is that correct?

Yes, you are right. I checked them as well before this PR.

msbaines:
You may be able to get a really nice speedup if you used the mmapped buffer directly instead of copying vertex attributes a vertex at a time.

mosra:
@msbaines That's precisely what I'm working on to enable with the asset import rework (happening in https://github.com/mosra/magnum/pull/371 -- description might be a bit outdated tho). Together with that we could get rid of all the copies of data stored CPU-side.

erikwijmans:
> Together with that we could get rid of all the copies of data stored CPU-side.

For vertices at least, we do need CPU-side copies for #333

msbaines:
> > Together with that we could get rid of all the copies of data stored CPU-side.
> 
> For vertices at least, we do need CPU-side copies for #333

Ah. Very good point. I already broke this once optimizing mesh loading:) You could still keep a CPU-side copy around with mmap. You'd just need to manage the file handle and mapping.

We really need to write tests for this behavior to avoid breaking it.

PR #349, opened 2019-11-12
[clang-tidy] enable undefined-bool-conversion check

msbaines:
## Motivation and Context

This check was initially disabled because it was failing.

## How Has This Been Tested

Build

## Types of changes

## Checklist


mosra:
Argh sorry about this one. This (the error the CI produces) at https://github.com/facebookresearch/habitat-sim/blob/9af1f62bd8fe9b41149c08e9a22f29decb94863f/src/esp/bindings/bindings.cpp#L46-L53 is a dirty leftover from the time when I ported to Magnum's own Python bindings. When Python's GC comes into play, the invariants in C++ code (where a feature is always attached to a scenegraph object) no longer hold and this was a check for that. I'll see if that can be exposed / checked for in a saner way.

msbaines:
> Argh sorry about this one. This (the error the CI produces) at
> 
> https://github.com/facebookresearch/habitat-sim/blob/9af1f62bd8fe9b41149c08e9a22f29decb94863f/src/esp/bindings/bindings.cpp#L46-L53
> 
> is a dirty leftover from the time when I ported to Magnum's own Python bindings. When Python's GC comes into play, the invariants in C++ code (where a feature is always attached to a scenegraph object) no longer hold and this was a check for that. I'll see if that can be exposed / checked for in a saner way.

Should I just remove this check. It may get optimized out. If it does not get removed, I don't think it will work. What would cause the reference to be NULL?

msbaines:
> > What would cause the reference to be NULL?
> 
> Like I'm saying above -- when C++ deletes an object that still has some features referenced from python, `node()` of those features becomes null. So removing the check is not a solution.
> 
> Is this critical (do I need to jump on this _immediately_)? Can we live with this thing disabled for a while?

No. Not critical. I can ignore the warning for this particular code if you think that is best.

I agree that removing the check is not a solution. However, I'm not sure leaving the check in is a solution either. It gives a false sense of security. clang will optimize out the check for optimization level O1 and above:

https://godbolt.org/z/JTv_Nc

This is because the check is relying on undefined behavior (null dereference).

Spent some time looking through the logic and now understand where the nullptr is coming from:

1) When the LinkedList is destroyed it calls clear()
2) clear() calls erase() on each item
3) erase() calls doErase() which calls cut()
4) cut() sets _list to nullptr.

mosra:
> if you think that is best

I'm not saying it's best, not at all. Problem is that i need to expose this check *somehow* through the scenegraph APIs in order to make it possible to check for this condition without relying on UB, and I don't know how to do that yet without adding an API that says 

> "look, `thing.object()` will *always* return a valid reference so you can totally trust it ... except for this one little case where Python's refcounting gets involved and in that case you might want to call `isThisFeatureReallyStillAttachedToAnObjectLikeItIsSupposedToBe()` in order to avoid dereferencing a null pointer"

To save time for both you and me, can we put this PR aside for a while (and thus keep the bool conversion check disabled globally) until I figure a non-confusing solution? :) Then I'll push to this branch and fix it properly. Thanks!

msbaines:
> > if you think that is best
> 
> I'm not saying it's best, not at all. Problem is that i need to expose this check _somehow_ through the scenegraph APIs in order to make it possible to check for this condition without relying on UB, and I don't know how to do that yet without adding an API that says
> 
> > "look, `thing.object()` will _always_ return a valid reference so you can totally trust it ... except for this one little case where Python's refcounting gets involved and in that case you might want to call `isThisFeatureReallyStillAttachedToAnObjectLikeItIsSupposedToBe()` in order to avoid dereferencing a null pointer"
> 
> To save time for both you and me, can we put this PR aside for a while (and thus keep the bool conversion check disabled globally) until I figure a non-confusing solution? :) Then I'll push to this branch and fix it properly. Thanks!

Why not just disable it on a line-by-line basis like I'm doing here? What is the concern? I think this is a good check. Its catching something legitimate. I'm disabling it for this case but would like to prevent new occurrences.

msbaines:
> Okay then :) Will follow up with #353.

Thanks much!

PR #350, opened 2019-11-13
Fix tryStep edge cases

erikwijmans:
## Motivation and Context

Currently, tryStep returns {NAN, NAN, NAN} if you give it a non-navigable, just having it return the start position seems cleaner as returning in all NANs results in a crash if you try to render from that location (which makes sense, but just not moving seems cleaner/less confusing).

Also, I found out that `moveAlongSurface` isn't guaranteed to return a point on the surface of the navmesh, only within a polygon by [x, z].  I added an explicit call to get the height of the polygon.  It seems like almost all the time, the height isn't different, but sometimes it is a bit different.

I needed to make the comparison fuzzier for the rendering test to accommodate the newly added larger MP3D mesh (we had be running into bugs in the resource manager that were only apparent on larger meshes, so I added this one). Note, this was also failing on master for me and all other tests passed with the stricter version on this PR.

## How Has This Been Tested

I had been getting NaNs very very very infrequently when running large scale RL in replica, and with these changes, that has gone away.  Existing tests also still pass.

## Types of changes

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -879,14 +879,9 @@ T PathFinder::tryStep(const T& start, const T& end) {
   // navmesh, it seems to be in 99.9% of cases for us, but there are some
   // extreme edge cases where it won't be, so explicitly get the height of the
   // surface at the endPoint and set its height to that.
-  {
-    float endPointHeight;
-    // Note, this will never fail as endPoint is always within in the poly
-    // polys[numPolys - 1]
-    navQuery_->getPolyHeight(polys[numPolys - 1], endPoint.data(),
-                             &endPointHeight);
-    endPoint[1] = endPointHeight;
-  }
+  // Note, this will never fail as endPoint is always within in the poly
+  // polys[numPolys - 1]
+  navQuery_->getPolyHeight(polys[numPolys - 1], endPoint.data(), &endPoint[1]);
 
   // Hack to deal with infinitely thin walls in recast allowing you to
   // transition between two different connected components
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -873,10 +873,12 @@ T PathFinder::tryStep(const T& start, const T& end) {
     return start;
   }
 
-  // According to recast's docs, the endPoint is not guaranteed to be actually
-  // on the surface of the navmesh, it seems to be in 99.9% of cases for us, but
-  // there are some extreme edge cases where it won't be, so explicitly get the
-  // height of the surface at the endPoint and set its height to that.
+  // According to recast's code
+  // (https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMeshQuery.cpp#L2006-L2007),
+  // the endPoint is not guaranteed to be actually on the surface of the
+  // navmesh, it seems to be in 99.9% of cases for us, but there are some
+  // extreme edge cases where it won't be, so explicitly get the height of the
+  // surface at the endPoint and set its height to that.
   {
     float endPointHeight;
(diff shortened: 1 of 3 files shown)


src/esp/nav/PathFinder.cpp
```diff
@@ -850,20 +850,43 @@ T PathFinder::tryStep(const T& start, const T& end) {
...
+  // (https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMeshQuery.cpp#L2006-L2007),
+  // the endPoint is not guaranteed to be actually on the surface of the
+  // navmesh, it seems to be in 99.9% of cases for us, but there are some
+  // extreme edge cases where it won't be, so explicitly get the height of the
+  // surface at the endPoint and set its height to that.
+  {
+    float endPointHeight;
+    // Note, this will never fail as endPoint is always within in the poly
+    // polys[numPolys - 1]
+    navQuery_->getPolyHeight(polys[numPolys - 1], endPoint.data(),
```
msbaines: Why not just pass &endPoint[1] here? You would avoid the temporary and the scope.

tests/test_sensors.py
```diff
@@ -123,7 +123,7 @@ def test_sensors(scene, has_sem, sensor_type, gpu2gpu, sim, make_cfg_settings):
     # Different GPUs and different driver version will produce slightly different images
     assert np.linalg.norm(
         obs[sensor_type].astype(np.float) - gt.astype(np.float)
-    ) < 1.5e-2 * np.linalg.norm(gt.astype(np.float)), f"Incorrect {sensor_type} output"
+    ) < 2e-2 * np.linalg.norm(gt.astype(np.float)), f"Incorrect {sensor_type} output"
```
msbaines: This is unrelated and should be probably be a separate change.

src/esp/nav/PathFinder.cpp
erikwijmans: I didn't want to dig through the code to make sure that is safe (it seems kinda scary to me).

tests/test_sensors.py
erikwijmans: Makes sense.  Will make a separate PR.

src/esp/nav/PathFinder.cpp
msbaines: What is the concern exactly? I don't see the hazard.

However, on line 888, you are unconditionally updating endPoint[1]. If getPolyHeight were to somehow fail and not update endPointHeight, you would be copying uninitialized memory to endPoint[1]. Passing &endPoint[1], directly would avoid this issue.

tests/test_sensors.py
erikwijmans: https://github.com/facebookresearch/habitat-sim/pull/352

src/esp/nav/PathFinder.cpp
erikwijmans: Looks like it's fine.  I generally prefer to not alias pointers.  There are optimizations the compiler can make if you don't alias pointers (although you need to use `__restrict__` to specify that they aren't aliased).

msbaines: Interesting, I knew about strict-aliasing and type-punning (not the case here) which can cause undefined behavior but was not aware of issues of non type-punned aliasing. But yeah, looks like if you use __restrict__ you can avoid a copy in certain cases. I don't think this is one of those cases though. The main benefit seems to be when there is a loop iterating over an array of data.

https://software.intel.com/en-us/articles/pointer-aliasing-and-vectorization
https://devblogs.nvidia.com/cuda-pro-tip-optimize-pointer-aliasing/

erikwijmans:
> Could you add test-case to src/tests/NavTest.cpp which checks for the new behavior.

Added a new test for the case of a non-navigable point.

PR #355, opened 2019-11-17
[viewer] add FPS counter using imgui

msbaines:
## Motivation and Context

Would be cool to be able to use imgui for HeadsUpDisplay

Screen Shot 2019-11-16 at 9 35 21 PM

## How Has This Been Tested

Ran viewer and verified that FPS counter is displayed correctly.

## Types of changes

## Checklist


src/cmake/dependencies.cmake
```diff
@@ -31,6 +31,8 @@ else()
   set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${DEPS_DIR}/eigen-git-mirror/cmake")
 endif()
 
+set(IMGUI_DIR "${DEPS_DIR}/imgui")
```
mosra: can you wrap this in `if(NOT IMGUI_DIR)` so it works also with my system-wide magnum installation? :)

src/utils/viewer/viewer.cpp
```diff
@@ -352,6 +367,31 @@ void Viewer::drawEvent() {
     physicsManager_->debugDraw(projM * camM);
   }
 
+  _imgui.newFrame();
+
+  ImGui::Begin("main", NULL,
+               ImGuiWindowFlags_NoDecoration | ImGuiWindowFlags_NoBackground |
+                   ImGuiWindowFlags_AlwaysAutoResize);
+  ImGui::SetWindowFontScale(2.0);
+  ImGui::Text("%.1f FPS", Double(ImGui::GetIO().Framerate));
```
bigbike: What is the strategy to compute the fps in "iamgui"? Do they measure it in an average moving window (e.g., moving 30 frames), or simply return the overall average FPS, ever since the simulator starts?

bigbike:
Also, would you consider adding a boolean switch to turn on/off the info display?
Check the `displayPerformanceInfo` in #343.

src/utils/viewer/viewer.cpp
msbaines: Framerate is calculated over last 120 frames:

https://github.com/ocornut/imgui/blob/57dc34f4e850c29ba9be7886afc7d07486a7ef59/imgui.h#L1445
https://github.com/ocornut/imgui/blob/ca30ef4269150e71a950106c6cfa8f94898581da/imgui.cpp#L3629

PR #359, opened 2019-11-20
[clang-tidy] enable google-build-using-namespace

msbaines:
## Motivation and Context

Enable one more clang-tidy check.

## How Has This Been Tested

Build.

## Types of changes

## Checklist


src/esp/bindings/bindings.cpp
```diff
@@ -25,20 +26,47 @@ using namespace py::literals;
...
+using esp::scene::SuncgObjectCategory;
+using esp::scene::SuncgRegionCategory;
+using esp::scene::SuncgSemanticObject;
+using esp::scene::SuncgSemanticRegion;
+using esp::sensor::Observation;
+using esp::sensor::PinholeCamera;
+using esp::sensor::Sensor;
+using esp::sensor::SensorSpec;
+using esp::sensor::SensorSuite;
+using esp::sensor::SensorType;
```
mosra: Do we need this, considering we scatter the bindings across multiple modules again on the Python side? To make it easier on both sides, I would say best would be to have `gfx`, `physics`, `scene`, ... submodules generated in the bindings (without using `using`, so it's clear through which submodule should given API be exposed) and then making use of those on the `Python` side instead of have things manually cherry-picked via `__all__`.

Of course there's a question whether this should be done as a part of this PR or separately, but based on previous experience the tech debt TODOs tend to stay in the code forever and so I'd say it should be in this PR. After all, the purpose of this PR is to fix the root cause of the Clang Tidy warnings, not merely silencing them, right? :)

src/esp/gfx/test/DepthUnprojectionTest.cpp
```diff
@@ -43,6 +43,7 @@ struct DepthUnprojectionTest : Mn::GL::OpenGLTester {
   void benchmarkGpuUnprojectExisting();
 };
 
+// NOLINTNEXTLINE(google-build-using-namespace)
 using namespace Mn::Math::Literals;
```
mosra: Is there anything I can do on the engine side to not need to have this annotation in *every* source file that uses the literals? How does Clang Tidy behave with `using namespace std::chrono::literals`? Do you need to do the same or is there something that whitelists that namespace?

erikwijmans:
You likely need to add 
```
#include <pybind11/eigen.h>
#include <pybind11/stl.h>
#include <pybind11/functional.h>
```
everywhere

msbaines:
> You likely need to add
> 
> ```
> #include <pybind11/eigen.h>
> #include <pybind11/stl.h>
> #include <pybind11/functional.h>
> ```
> 
> everywhere

Gotcha. Thanks!

msbaines:
All feedback addressed and tests now passing.

src/esp/gfx/test/DepthUnprojectionTest.cpp
```diff
@@ -25,6 +25,9 @@
 namespace Cr = Corrade;
 namespace Mn = Magnum;
 
+using Mn::Math::Literals::operator""_degf;
+using Mn::Math::Literals::operator""_rgbf;
```
mosra: I still don't agree with Tidy about this one, but I guess we can't do any better. 

Found https://reviews.llvm.org/D33010 that makes the check not complain for standard literals, but such hardcoded unextendable whitelisting makes me kinda angry :/

erikwijmans:
I think LTO is the most time consuming part of the bindings.  Building the `.o` is peanuts compared to linking :)

msbaines:
ping

PR #365, opened 2019-11-26
--cov-append for second pytest

erikwijmans:
## Motivation and Context

We need to add `--cov-append` for the second invocation of pytest in habitat-sim

Also moves the `--cov-report=xml --cov=./` flags to just the CI as those aren't needed locally

## How Has This Been Tested

On the CI!


mathfac:
PR improves `src/esp/assets/ResourceManager.cpp` quite strange, as impact was expected for Python code only.

erikwijmans:
Yeah, I have no idea what that's about.  Our python code does touch a lot of cpp code tho

PR #371, opened 2019-12-04
Remove zero area polys

### LINKED ISSUE #370: projectToPoly method validation fails when Agent is moving forward diagonal to a wall.
## 🐛 Bug
The agent goes out of bounds and the episodes ends because of segmentation fault.

## Command
I ran the example below

## To Reproduce

1. Set the agent in a position close to a wall, for example:
**map**: data/scene_datasets/mp3d/2n8kARJN3HM/2n8kARJN3HM.navmesh
**start_pos**: [-0.20505089, -3.468762, 2.312384]
**start_rot**: quaternion(0.41104656457901, 0, 0.91161435842514,0)
2. Set the FORWARD_STEP_SIZE: 2.00
3. Trigger the "MOVE_FORWARD" action.

Those steps will make the agent fall out of bounds in the second dimension and not the first. So when the projection procedure checks for errors it cannot determine that there is a NAN in the position. Therefore the agent act method wrongly returns that there is not a collision after the action is performed. Which is in fact not true, there is a collision or the agent is "out of bounds". 

Code Sample:

` import habitat
import cv2
from habitat.sims.habitat_simulator.actions import HabitatSimActions

FORWARD_KEY="w"
LEFT_KEY="a"
RIGHT_KEY="d"
FINISH="f"

def transform_rgb_bgr(image):
    return image[:, :, [2, 1, 0]]

def example():
    print("Starting habitat Environment")
    env = habitat.Env(
        config=habitat.get_config("configs/tasks/pointnav.yaml")
    )
    print("Environment creation successful")
    observations = env.reset()
    cv2.imshow("RGB", transform_rgb_bgr(observations["rgb"]))

    print("Agent stepping around inside environment.")
   
   
    initial_state = env._sim.get_agent_state(0)
    init_translation = initial_state.position
    init_rotation = initial_state.rotation
       
    print(init_translation, init_rotation)
    print(observations["heading"])
   
    count_steps = 0
    while not env.episode_over:
        keystroke = cv2.waitKey(0)

        if keystroke == ord(FORWARD_KEY):
            action = "MOVE_FORWARD"
            print("action: FORWARD")
        elif keystroke == ord(LEFT_KEY):
            action = "TURN_LEFT"
            print("action: LEFT")
        elif keystroke == ord(RIGHT_KEY):
            action = "TURN_RIGHT"
            print("action: RIGHT")
        elif keystroke == ord(FINISH):
            action = "STOP"
            print("action: FINISH")
        else:
            print("INVALID KEY")
            continue

        observations = env.step(action)
        count_steps += 1
        print(observations["heading"])
        #print("Destination, distance: {:3f}, theta(radians): {:.2f}".format(
        #    observations["pointgoal"][0], observations["pointgoal"][1]))
        cv2.imshow("RGB", transform_rgb_bgr(observations["rgb"]))
       
        initial_state = env._sim.get_agent_state(0)
        init_translation = initial_state.position
        init_rotation = initial_state.rotation
       
        print(init_translation, init_rotation)

    print("Episode finished after {} steps.".format(count_steps))

if __name__ == "__main__":
    example()`

## Expected behavior

A collision occurs and the agent doesn't move to the void. 

## Additional context

The validation for projectToPoly should be in the three dimensions but currently is done in only one dimension. Which causes an error when walking near the walls diagonally since the resulting projection becomes ( number, NAN, number) which is not properly validated. The file in question is under **src/esp/nav/PathFinder.cpp**

`std::tuple<dtStatus, dtPolyRef, vec3f> projectToPoly(
    const T& pt,
    const dtNavMeshQuery* navQuery,
    const dtQueryFilter* filter) {
  // Defines size of the bounding box to search in for the nearest polygon.  If
  // there is no polygon inside the bounding box, the status is set to failure
  // and polyRef == 0
  constexpr float polyPickExt[3] = {2, 4, 2};  // [2 * dx, 2 * dy, 2 * dz]
  dtPolyRef polyRef;
  // Initialize with all NANs at dtStatusSucceed(status) == true does NOT mean
  // that it found a point to project to..........
  vec3f polyXYZ{NAN, NAN, NAN};
  dtStatus status = navQuery->findNearestPoly(pt.data(), polyPickExt, filter,
                                              &polyRef, polyXYZ.data());

  // So let's call it a failure if it didn't actually find a point....
  **if (std::isnan(polyXYZ[0]))
    status = DT_FAILURE;**

  return std::make_tuple(status, polyRef, polyXYZ);
}`

## System Info

image

erikwijmans:
## Motivation and Context

There are some zero-area polygons in the navmeshes for some reason.  Those cause issues (#370).  This PR marks the zero are polygons as disabled, effectively removing them from the navigation mesh.

I have also pulled in the latest changes from recast as those do contain stability enhancements for methods that we heavily rely on.

Fixes #370 

## How Has This Been Tested

Via the test point in #370 (I can add that point to an explicit test, but its an MP3D mesh and we don't have them on the CI).

## Types of changes

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -603,8 +603,8 @@ struct NavMeshTileHeader {
 float polyArea(const dtPoly* poly, const dtMeshTile* tile) {
   float area = 0;
   // Code to iterate over triangles from here:
-  // https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMesh.cpp#L684
-  const unsigned int ip = (unsigned int)(poly - tile->polys);
+  // https://github.com/recastnavigation/recastnavigation/blob/57610fa6ef31b39020231906f8c5d40eaa8294ae/Detour/Source/DetourNavMesh.cpp#L684
+  const std::ptrdiff_t ip = poly - tile->polys;
   const dtPolyDetail* pd = &tile->detailMeshes[ip];
   for (int j = 0; j < pd->triCount; ++j) {
     const unsigned char* t = &tile->detailTris[(pd->triBase + j) * 4];
@@ -643,9 +643,12 @@ void PathFinder::removeZeroAreaPolys() {
     for (int jPoly = 0; jPoly < tile->header->polyCount; ++jPoly) {
       // Get the polygon reference from the tile and polygon id
       dtPolyRef polyRef = navMesh_->encodePolyId(iTile, tile->salt, jPoly);
-      const dtPoly* poly = 0;
-      const dtMeshTile* tmp;
+      const dtPoly* poly = nullptr;
+      const dtMeshTile* tmp = nullptr;
       navMesh_->getTileAndPolyByRefUnsafe(polyRef, &tmp, &poly);
+
+      CORRADE_INTERNAL_ASSERT(poly != nullptr);
+      CORRADE_INTERNAL_ASSERT(tmp != nullptr);
 
       if (polyArea(poly, tile) < 1e-5) {
         navMesh_->setPolyFlags(polyRef, POLYFLAGS_DISABLED);
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -628,6 +628,32 @@ float polyArea(const dtPoly* poly, const dtMeshTile* tile) {
 }
 }  // namespace
 
+// Some polygons have zero area for some reason.  When we navigate into a zero
+// area polygon, things crash.  So we find all zero area polygons and mark
+// them as disabled/not navigable.
+void PathFinder::removeZeroAreaPolys() {
+  // Iterate over all tiles
+  for (int iTile = 0; iTile < navMesh_->getMaxTiles(); ++iTile) {
+    const dtMeshTile* tile =
+        reinterpret_cast<const dtNavMesh*>(navMesh_)->getTile(iTile);
+    if (!tile)
+      continue;
+
+    // Iterate over all polygons in a tile
+    for (int jPoly = 0; jPoly < tile->header->polyCount; ++jPoly) {
+      // Get the polygon reference from the tile and polygon id
+      dtPolyRef polyRef = navMesh_->encodePolyId(iTile, tile->salt, jPoly);
+      const dtPoly* poly = 0;
+      const dtMeshTile* tmp;
+      navMesh_->getTileAndPolyByRefUnsafe(polyRef, &tmp, &poly);
+
+      if (polyArea(poly, tile) < 1e-5) {
+        navMesh_->setPolyFlags(polyRef, POLYFLAGS_DISABLED);
+      }
+    }
+  }
+}
+
 bool PathFinder::loadNavMesh(const std::string& path) {
   FILE* fp = fopen(path.c_str(), "rb");
   if (!fp)
@@ -703,30 +729,7 @@ bool PathFinder::loadNavMesh(const std::string& path) {
   navMesh_ = mesh;
   bounds_ = std::make_pair(bmin, bmax);
 
-  // Some polygons have zero area for some reason.  When we navigate into a zero
-  // area polygon, things crash.  So we find all zero area polygons and mark
-  // them as disabled/not navigable.
-
-  // Iterate over all tiles
-  for (int iTile = 0; iTile < navMesh_->getMaxTiles(); ++iTile) {
-    const dtMeshTile* tile =
-        reinterpret_cast<const dtNavMesh*>(navMesh_)->getTile(iTile);
-    if (!tile)
-      continue;
-
-    // Iterate over all polygons in a tile
-    for (int jPoly = 0; jPoly < tile->header->polyCount; ++jPoly) {
-      // Get the polygon reference from the tile and polygon id
-      dtPolyRef polyRef = navMesh_->encodePolyId(iTile, tile->salt, jPoly);
-      const dtPoly* poly = 0;
-      const dtMeshTile* tmp;
-      navMesh_->getTileAndPolyByRefUnsafe(polyRef, &tmp, &poly);
-
-      if (polyArea(poly, tile) < 1e-5) {
-        navMesh_->setPolyFlags(polyRef, POLYFLAGS_DISABLED);
-      }
-    }
-  }
+  removeZeroAreaPolys();
 
   return initNavQuery();
 }
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -603,6 +603,7 @@ struct NavMeshTileHeader {
 float polyArea(const dtPoly* poly, const dtMeshTile* tile) {
   float area = 0;
   // Code to iterate over triangles from here:
+  // https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMesh.cpp#L684
   const unsigned int ip = (unsigned int)(poly - tile->polys);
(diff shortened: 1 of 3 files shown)


tests/test_snap_point.py
```diff
@@ -20,7 +20,7 @@ def test_data():
 
 
 @hypothesis.given(
-    nudge=st.tuples(st.floats(-20, 20), st.floats(-5, 5), st.floats(-20, 20))
+    nudge=st.tuples(st.floats(-10, 10), st.floats(-2.5, 2.5), st.floats(-10, 10))
```
erikwijmans: Hypothesis was complaining about the search rejecting too many points, so I made it slightly narrower (this is still much larger than the mesh so it'll search everywhere).

src/esp/nav/PathFinder.cpp
```diff
@@ -673,6 +702,32 @@ bool PathFinder::loadNavMesh(const std::string& path) {
...
+
+    // Iterate over all polygons in a tile
+    for (int jPoly = 0; jPoly < tile->header->polyCount; ++jPoly) {
+      // Get the polygon reference from the tile and polygon id
+      dtPolyRef polyRef = navMesh_->encodePolyId(iTile, tile->salt, jPoly);
+      const dtPoly* poly = 0;
+      const dtMeshTile* tmp;
+      navMesh_->getTileAndPolyByRefUnsafe(polyRef, &tmp, &poly);
+
+      if (polyArea(poly, tile) < 1e-5) {
```
aclegg3: Curious to know how many polygons have area below this threshold in our datasets and whether they sit on the edge of the mesh or not, but I think this looks fine.

src/esp/nav/PathFinder.cpp
```diff
@@ -673,6 +702,32 @@ bool PathFinder::loadNavMesh(const std::string& path) {
 
   navMesh_ = mesh;
   bounds_ = std::make_pair(bmin, bmax);
+
+  // Some polygons have zero area for some reason.  When we navigate into a zero
+  // area polygon, things crash.  So we find all zero area polygons and mark
+  // them as disabled/not navigable.
+
+  // Iterate over all tiles
```
aclegg3: Actually, can we break this out as a function so that I can also use it for recompute navmesh later?

src/esp/nav/PathFinder.cpp
erikwijmans: There were 3 for the one in #370

src/esp/nav/PathFinder.cpp
erikwijmans: Done

src/esp/nav/PathFinder.cpp
```diff
@@ -597,8 +597,63 @@ struct NavMeshTileHeader {
   dtTileRef tileRef;
   int dataSize;
 };
+
+// Calculate the area of a polygon by iterating over the triangles in the detail
+// mesh and computing their area
+float polyArea(const dtPoly* poly, const dtMeshTile* tile) {
+  float area = 0;
+  // Code to iterate over triangles from here:
+  // https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMesh.cpp#L684
```
msbaines: This reference is not stable. Use a specific commit instead of master.

src/esp/nav/PathFinder.cpp
```diff
@@ -597,8 +597,63 @@ struct NavMeshTileHeader {
...
+    const dtMeshTile* tile =
+        reinterpret_cast<const dtNavMesh*>(navMesh_)->getTile(iTile);
+    if (!tile)
+      continue;
+
+    // Iterate over all polygons in a tile
+    for (int jPoly = 0; jPoly < tile->header->polyCount; ++jPoly) {
+      // Get the polygon reference from the tile and polygon id
+      dtPolyRef polyRef = navMesh_->encodePolyId(iTile, tile->salt, jPoly);
+      const dtPoly* poly = 0;
```
msbaines: nullptr

src/esp/nav/PathFinder.cpp
```diff
@@ -597,8 +597,63 @@ struct NavMeshTileHeader {
...
+      const dtMeshTile* tmp;
```
msbaines: Initialize this also?

src/esp/nav/PathFinder.cpp
```diff
@@ -597,8 +597,63 @@ struct NavMeshTileHeader {
...
   int dataSize;
 };
+
+// Calculate the area of a polygon by iterating over the triangles in the detail
+// mesh and computing their area
+float polyArea(const dtPoly* poly, const dtMeshTile* tile) {
+  float area = 0;
+  // Code to iterate over triangles from here:
+  // https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMesh.cpp#L684
+  const unsigned int ip = (unsigned int)(poly - tile->polys);
```
msbaines: Use C++ casts. In this case, you could avoid the cast entirely by using ptrdiff_t..

src/esp/nav/PathFinder.cpp
```diff
@@ -597,8 +597,63 @@ struct NavMeshTileHeader {
...
 }  // namespace
 
+// Some polygons have zero area for some reason.  When we navigate into a zero
+// area polygon, things crash.  So we find all zero area polygons and mark
+// them as disabled/not navigable.
+void PathFinder::removeZeroAreaPolys() {
+  // Iterate over all tiles
+  for (int iTile = 0; iTile < navMesh_->getMaxTiles(); ++iTile) {
+    const dtMeshTile* tile =
+        reinterpret_cast<const dtNavMesh*>(navMesh_)->getTile(iTile);
```
msbaines: Why is this reinterpret_cast needed?

src/esp/nav/PathFinder.cpp
msbaines: nullptr

src/esp/nav/PathFinder.cpp
msbaines: initialize this also?

src/esp/nav/PathFinder.cpp
```diff
@@ -597,8 +597,63 @@ struct NavMeshTileHeader {
...
+      continue;
+
+    // Iterate over all polygons in a tile
+    for (int jPoly = 0; jPoly < tile->header->polyCount; ++jPoly) {
+      // Get the polygon reference from the tile and polygon id
+      dtPolyRef polyRef = navMesh_->encodePolyId(iTile, tile->salt, jPoly);
+      const dtPoly* poly = 0;
+      const dtMeshTile* tmp;
+      navMesh_->getTileAndPolyByRefUnsafe(polyRef, &tmp, &poly);
+
```
msbaines: Assert that poly and tmp are not nullptr?

src/esp/nav/PathFinder.cpp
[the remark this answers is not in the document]
erikwijmans: The non-const `getTile` method is private.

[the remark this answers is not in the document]
msbaines: Ah, so there is a private non-const method and a public const method.

Why not use const_cast instead? It's safer. reinterpret_cast is rarely ever needed and is the least type-safe cast.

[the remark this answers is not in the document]
erikwijmans: Yep, using `const_cast` makes more sense here

erikwijmans:
I don't have really any idea why these zero area polygons happen so I don't know how to create one.  I have started an issue with recast about this and zero area polygons are not suppose to happen, so it's likely a corner case caused by the very odd geometry of reconstruction errors.

PR #372, opened 2019-12-06
[clang-tidy] enable google-readability-casting

msbaines:
C++-style casting is always preferred to C-style casts.

## Motivation and Context

Enforce coding style.

## How Has This Been Tested

Build.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -253,15 +253,16 @@ bool PathFinder::build(const NavMeshSettings& bs,
   cfg.cs = bs.cellSize;
   cfg.ch = bs.cellHeight;
   cfg.walkableSlopeAngle = bs.agentMaxSlope;
-  cfg.walkableHeight = (int)ceilf(bs.agentHeight / cfg.ch);
-  cfg.walkableClimb = (int)floorf(bs.agentMaxClimb / cfg.ch);
-  cfg.walkableRadius = (int)ceilf(bs.agentRadius / cfg.cs);
-  cfg.maxEdgeLen = (int)(bs.edgeMaxLen / bs.cellSize);
+  cfg.walkableHeight = static_cast<int>(ceilf(bs.agentHeight / cfg.ch));
+  cfg.walkableClimb = static_cast<int>(floorf(bs.agentMaxClimb / cfg.ch));
+  cfg.walkableRadius = static_cast<int>(ceilf(bs.agentRadius / cfg.cs));
+  cfg.maxEdgeLen = static_cast<int>(bs.edgeMaxLen / bs.cellSize);
   cfg.maxSimplificationError = bs.edgeMaxError;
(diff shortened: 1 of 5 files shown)


src/esp/nav/PathFinder.cpp
```diff
@@ -253,15 +253,16 @@ bool PathFinder::build(const NavMeshSettings& bs,
...
   cfg.ch = bs.cellHeight;
   cfg.walkableSlopeAngle = bs.agentMaxSlope;
-  cfg.walkableHeight = (int)ceilf(bs.agentHeight / cfg.ch);
-  cfg.walkableClimb = (int)floorf(bs.agentMaxClimb / cfg.ch);
-  cfg.walkableRadius = (int)ceilf(bs.agentRadius / cfg.cs);
-  cfg.maxEdgeLen = (int)(bs.edgeMaxLen / bs.cellSize);
+  cfg.walkableHeight = static_cast<int>(ceilf(bs.agentHeight / cfg.ch));
+  cfg.walkableClimb = static_cast<int>(floorf(bs.agentMaxClimb / cfg.ch));
+  cfg.walkableRadius = static_cast<int>(ceilf(bs.agentRadius / cfg.cs));
+  cfg.maxEdgeLen = static_cast<int>(bs.edgeMaxLen / bs.cellSize);
```
mosra: Is "constructor cast" allowed by this rule? I.e., `int(ceilf(...))`. Because I'm using it extensively myself and magnum docs are full of it :)

msbaines: functional cast expressions are not reported by this rule but are equivalent to a c-style cast so less safe than a static_cast.

```
char *p;
long i = long(p); // will compile
long j = static_cast<long>(p); // will not compile
```

mosra: Yes, I'm aware of the safety concerns.

mosra:
Btw., this is also available in GCC/Clang as `-Wold-style-cast`. I think it would make sense to enable it globally -- the warnings would get reported "closer to the source" than with Clang Tidy.

msbaines:
> old-style-cast

Good idea. I tried turning that on but it seems like it is triggering on third_party headers while the clang-tidy check is not (they may check at different phases of compile). I will investigate turning this on in a later change.

I recently become aware of this warning and some others while investigating coding conventions:

https://wiki.qt.io/Coding_Conventions#Conventions_for_public_header_files

Does this PR look good to land in its current state?

msbaines:
ping

mosra:
> seems like it is triggering on third_party headers while the clang-tidy check is not (they may check at different phases of compile)

For GCC and Clang this can be fixed by doing `[target_]include_directories(SYSTEM <path>)` instead of `[target_]include_directories(<path>)`. I saw a bunch of these in the codebase. Ideal would be to use `target_link_libraries()` with imported targets instead of manually adding the include dirs.

PR #380, opened 2019-12-13
[DO NOT accept][Experiment only] Some quick, hacky tests on frustum culling

bigbike:
## Motivation and Context

This PR is NOT intended to be committed into the master. It is just to get some general understandings of the code, and collect early feedback. 

Created a PR simply because in this way, people can directly comment on the code.

## Types of changes

## Checklist


bigbike:
@mosra :
The following are the current results. Certainly there is something wrong. Some mesh components are incorrectly culled. (I assume the rangeFrustum() is bug-free at this point).

image

image

bigbike:
@mosra: the code is quite hacky and a bit naive. I just quickly put something together to try.
If some of the code (especially the code related to Magnum) can be written in a more efficient way, do let me know.

src/utils/viewer/viewer.cpp
```diff
@@ -359,7 +372,43 @@ void Viewer::drawEvent() {
...
+      // for (int iDrawable = 0; iDrawable < total; ++iDrawable)
+      scene::SceneNode& node =
+          dynamic_cast<scene::SceneNode&>(drawables[iDrawable].object());
+      auto& aabb = node.getMeshBB();
+      Magnum::Math::Vector4<float> bbMin{aabb.min(), 1.0};
+      Magnum::Math::Vector4<float> bbMax{aabb.max(), 1.0};
+      auto modelview = renderCamera_->getMagnumCamera().cameraMatrix() *
+                       node.absoluteTransformation();
+      if (Magnum::Math::Intersection::rangeFrustum(
+              Magnum::Math::Range3D<float>{(modelview * bbMin).xyz(),
```
bigbike: It transforms the AABB from model frame to the camera frame.
Then the AABB is no longer an AABB, it is an OBB.
So cannot use rangeFrustum to do the culling any more.

bigbike:
Bug has been fixed.
The result is as follows. Now the code can generate same images with and without frustum culling.

Culling result:
image

Ground truth:
image

src/utils/viewer/viewer.cpp
mosra: You'd need to transform the frustum instead to keep AABBs as AABBs -- and I see you did exactly that :)

bigbike:
@mosra :
Another issue is about the "drawables".
They are created at the loading stage and put into a big drawable group "drawables_".
When culling is enabled, the visible drawables are moved to a new group for rendering, and moved back afterwards.
Do you think it is reasonable or any other better idea to do it? Thanks.

src/utils/viewer/viewer.cpp
```diff
@@ -359,7 +368,46 @@ void Viewer::drawEvent() {
...
+      // for (int iDrawable = 0; iDrawable < total; ++iDrawable)
+      scene::SceneNode& node =
+          dynamic_cast<scene::SceneNode&>(drawables[iDrawable].object());
+      // mvp = proj * modelview
+      auto mvp = cam.projectionMatrix() * cam.cameraMatrix() *
+                 node.absoluteTransformation();
+      Magnum::Math::Frustum<float> frustum;
+      frustum = Magnum::Math::Frustum<float>::fromMatrix(mvp);
+
+      if (Magnum::Math::Intersection::rangeFrustum(node.getMeshBB(), frustum)) {
```
mosra: Since >99% of the scene is currently static, I think it would make sense to have the AABBs *absolute*, to avoid the need to do all the above matrix ops again per each node -- basically calculating the frustum just from the current camera position/orientation.

And for dynamic/physics objects, to avoid having to recalculate the AABBs after each physics tick (unless Bullet can give you absolute AABBs "for free"), what about having bounding spheres instead? Then you would only need absolute translation of each and could again use a single `frustum` for all nodes. It *may* result in slightly more stuff being rendered than needed, but I think that would be balanced by simpler handling here.

src/utils/viewer/viewer.cpp
```diff
@@ -359,7 +368,46 @@ void Viewer::drawEvent() {
...
+    Magnum::SceneGraph::DrawableGroup3D visibleDrawables;
+
+    for (int iDrawable = total - 1; iDrawable >= 0; iDrawable--) {
+      // This, of course, will cause segfault.
+      // for (int iDrawable = 0; iDrawable < total; ++iDrawable)
+      scene::SceneNode& node =
+          dynamic_cast<scene::SceneNode&>(drawables[iDrawable].object());
+      // mvp = proj * modelview
+      auto mvp = cam.projectionMatrix() * cam.cameraMatrix() *
+                 node.absoluteTransformation();
```
mosra: I think what would be more performant (and simpler in code) is operating on the result of [cam.drawableTransformations()](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1Camera.html#a9a17bb008136d9e03687f815a189cda2), filtering that, and feeding that to the other [cam.draw()](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1Camera.html#a09546e57e1fc178f76cb49adb4b5a37e) overload -- you wouldn't need to add/remove drawables from the list and you'd get the transformations calculated as well.

The way it is now, the loop is doing quite a lot of redundant calculations (especially in deep object hierarchies, as `absoluteTransformation()` always goes all the way up the hierarchy with no chance of reusing what was already calculated before) and then `draw()` does that all again to calculate matrices passed to shaders (again with no chance of reusing what you already had calculated here).

src/utils/viewer/viewer.cpp
```diff
@@ -359,7 +368,46 @@ void Viewer::drawEvent() {
   int DEFAULT_SCENE = 0;
   int sceneID = sceneID_[DEFAULT_SCENE];
   auto& sceneGraph = sceneManager_.getSceneGraph(sceneID);
-  renderCamera_->getMagnumCamera().draw(sceneGraph.getDrawables());
+
+  // XXX
+  auto& drawables = sceneGraph.getDrawables();
+  auto total = drawables.size();
+  size_t visibles = 0;
+  auto& cam = renderCamera_->getMagnumCamera();
```
mosra: Can we use `auto` less (contrary to what [almost always auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/) suggests), please? :) It's easy when you have an IDE that gives you type hints, but unnecessarily hard when you don't (like here in GitHub's code view). In this particular case I was pretty sure `drawables` was a `std::vector` until I saw the `add()` below.

src/utils/viewer/viewer.cpp
bigbike: I read your doc: (https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1Drawable.html#SceneGraph-Drawable-draw-order)
- I did not get it so far how can we "draw just a subset" by using drawableTransformations(), std::sort() and cam.draw() overload?

- I read the magnum source code of "drawableTransformations()". It will apply the cameraMatrix. Basically it computes cameraMatrix * modelAbsTransformMatrix, which transforms the model to the camera space. Then again, the AABB is an OBB now. How can we do the frustum culling at this point?

bigbike: I think the [doc](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1Camera.html#a9a17bb008136d9e03687f815a189cda2) is a bit vague on function: drawableTransformations(...)
The description simply says: "Drawable transformations". 
I think a better version would be "Drawable transformations in camera space" (so it implies the model transformation matrix has been left-multiplied by cameraMatrix.)

bigbike:
@mosra : I sincerely appreciate your comments which is very helpful, and inspirational.

- Using reference to a drawable is a good idea. It solves my biggest concern. It avoids the move or copy operations on drawables. 

- I cannot use the function, cam.drawableTransformations(...) as it will post-multiply the camera matrix. I need the culling happened in model space. Otherwise AABB becomes an OBB.

- I read the source code of drawableTransformations(...). I will have to write a very similar function that only computes the absolute transformations for all drawables "efficiently".

- Too bad I "cannot see" any speed up either due to the reason you mentioned. I do not have a linux box to run the viewer. The only way to verify if there is performance gain is to implement it in class Renderer() and run it on headless mode. That is my next step. This PR is just a toy example.

- Absolute AABB for each static mesh component: This is a very good idea and should be implemented in my next PR. Let us forget about the dynamic objects temporarily.

- Use sphere in culling test so that we can deal with static and dynamic object uniformly. This is another good idea we can consider. If I understand correctly, the dynamic object (physics-enabled object) is also using AABB now. When an object moves, the initial AABB (NOT the current AABB) becomes an OBB, then the code computes the current AABB, which is the AABB of the OBB.

- We can use absolute AABB for static objects, and sphere for dynamic object for example.

- Can we use auto less? Yes, we can and we should. I was lazy and please forgive me. I used to oppose using "auto" a lot.

mosra:
> Basically it computes cameraMatrix * modelAbsTransformMatrix, which transforms the model to the camera space. Then again, the AABB is an OBB now.

Huh, hmm... you got me thinking for a good part of the day, but then I realized it actually doesn't matter that it would create an OBB. It doesn't matter in the current case because you use the transformation to calculate `frustum`, and the local AABB stays axis-aligned. So this code:

```cpp
      auto mvp = cam.projectionMatrix() * cam.cameraMatrix() *
                 node.absoluteTransformation();
```

would become the following, and then you would continue with creating the `frustum` exactly like before:

```cpp
      auto mvp = cam.projectionMatrix() * drawableTransformations[i].second;
```

and then you'd do the culling and pass the filtered list to `draw()`.

> Absolute AABB for each static mesh component

It wouldn't matter either for the next PR where you implement absolute AABBs, because there you would not be using the calculated transformations in any way except for passing them to the `draw()` later. Consider this code (which I'm going to put straight into the docs to remove the confusion you pointed out):

```cpp
struct CullableDrawable3D: SceneGraph::Drawable3D {
    Range3D aabb; /* Relative to world origin */

    // ...
};

/* Camera frustum relative to world origin */
const Frustum frustum = Frustum::fromMatrix(camera.projectionMatrix()*
    cameraObject.absoluteTransformationMatrix());

/* Erase all items that don't pass the frustum check */
std::vector<std::pair<std::reference_wrapper<SceneGraph::Drawable3D>, Matrix4>>
    drawableTransformations = camera.drawableTransformations(drawableGroup);
drawableTransformations.erase(std::remove_if(
    drawableTransformations.begin(), drawableTransformations.end(),
    [&](const std::pair<std::reference_wrapper<SceneGraph::Drawable3D>, Matrix4>& a) {
        Range3D aabb = static_cast<CullableDrawable3D&>(a.first.get()).aabb;
        return !Math::Intersection::rangeFrustum(aabb, frustum);
    }),
    drawableTransformations.end());

/* Draw just the visible part */
camera.draw(drawableTransformations);
```

A variant of this snippet (the `std::remove_if()` part, especially) would work for the current case as well I think. I hope I didn't make some brainfart on the way that makes this horribly broken / incorrect :)

bigbike:
Hahaha, now I know why OBB is a problem in my opinion.

In the current to-be-rendered frame, cameraMatrix stays the same. That means projMatrix * cameraMatrix can be computed just once, and thus save the computation, comparing to  auto `mvp = cam.projectionMatrix() * drawableTransformations[i].second; `

But if I computed projMatrix * cameraMatrix first, it means framework is transformed to camera frame. If the object is also transformed to this frame, AABB becomes OBB.

mosra:
Not sure if I understand what you mean.

> create a very similar function to drawableTransformations(...) where it ONLY computes the absolute transformations for all drawables

That's what I was about to do initially, but then realized is actually not needed and discarded the change.

> Of course I need another similar draw(...) overload in camera class to handle the rendering.

You wouldn't need a special `draw()` in that case, all that's needed would be multiplying the transformation matrices with `cam.cameraMatrix()` and then calling the original `draw()`. But again, I don't see why that's needed.

In an attempt to clear this up, there are two cases, and in neither I see any issue:

1. With relative AABBs, for each object you transform the frustum to be object-local, and then you use `mvp = cam.projectionMatrix() * drawableTransformations[i].second;`. Yes, this one multiplication could be saved, but that would mean you'd not have correct data to pass to `draw()` later, since it *needs* projection and transformation separately.
2. With absolute AABBs, you would not use the calculated transformations in any way, only pass them later to `draw()`.

bigbike:
@mosra: I see. If I save the computation by multiplying projection * cameraMtx first, then I will have to pay back before calling the draw(...) because it needs cameraMtx * modelabsTransformMtx. So all the savings have gone.

I wish to have a draw(...) that takes MVP matrix as parameter input (that is why when I said I need another similar draw(...) overload in previous message), but it seems strange since draw(...) is a function of the camera, and the _projectionMatrix is an internal member, and thus should not be part of the parameter input.


Another topic:
If I found the search dialog box in magnum doc page has fixed length. If the returned item is long, I cannot see the entire function specification (see the following screenshot). Is it possible to make it flexible?

Screen Shot 2019-12-18 at 9 37 52 AM

mosra:
> I wish to have a draw(...) that takes MVP matrix as parameter input (that is why when I said I need another similar draw(...) overload in previous message)

That would be a problem, because non-trivial shaders such as Phong require projection and transformation/normal matrices separately in order to correctly calculate lighting. That's the only reason why in this case and in `Drawable::draw()` the two are passed separately.

> fixed length

It gets wider if you enlarge the window, though there is an upper limit. I tried to tune it so it's usable in most cases, but `draw()` is one of the crazy outliers that break every layout.

bigbike:
I suspect the following code as you wrote in the previous message may not work. But I believe it is easy to fix. 

In my opinion you need to change:
```
const Frustum frustum = Frustum::fromMatrix(camera.projectionMatrix()*
    cameraObject.absoluteTransformationMatrix());
```
To
```
const Frustum frustum = Frustum::fromMatrix(camera.projectionMatrix()*
    camera.cameraMatrix());
```
And it would work.

Why?
We have `X = PVM * X_0` where `X_0` is the vertex position in model frame. `P, V, M` are projection, camera, model absolute transform matrices.

When you specify the `CullableDrawable3D`, the `aabb` within is w.r.t. world frame. It implies it considers `M * X_0`. Then the frustum must be extracted from `PV`.

That means ` frustum = Frustum::fromMatrix(camera.projectionMatrix()*
    camera.cameraMatrix());`

How do you think?


```cpp
struct CullableDrawable3D: SceneGraph::Drawable3D {
    Range3D aabb; /* Relative to world origin */

    // ...
};

/* Camera frustum relative to world origin */
const Frustum frustum = Frustum::fromMatrix(camera.projectionMatrix()*
    cameraObject.absoluteTransformationMatrix());

/* Erase all items that don't pass the frustum check */
std::vector<std::pair<std::reference_wrapper<SceneGraph::Drawable3D>, Matrix4>>
    drawableTransformations = camera.drawableTransformations(drawableGroup);
drawableTransformations.erase(std::remove_if(
    drawableTransformations.begin(), drawableTransformations.end(),
    [&](const std::pair<std::reference_wrapper<SceneGraph::Drawable3D>, Matrix4>& a) {
        Range3D aabb = static_cast<CullableDrawable3D&>(a.first.get()).aabb;
        return !Math::Intersection::rangeFrustum(aabb, frustum);
    }),
    drawableTransformations.end());
```

mosra:
Actually I originally had `cameraMatrix()` but then realized it was a mistake and changed the code to use `absoluteTransformationMatrix()` instead. This is why I think it's correct:

- every `aabb` is already relative to world origin
- and to cull those, we need also transformation of the camera in the same coordinate space, thus relative to world origin as well -- so `cameraObject.absoluteTransformationMatrix`
- `camera.cameraMatrix` is transformation of the world origin relative to the camera, thus an inverse of the above, so `cameraObject.absoluteTransformationMatrix^-1`
- `camera.cameraMatrix*aabb` (pseudocode) would be transformation of every AABB to be relative to the camera, but that's not what we want, as it would become an OBB

IOW, in your reasoning above, you consider `X` to be relative to camera, with projection applied, and `V` being `cameraMatrix()`. But since that turns AABBs into OBBs, we can't work in that coordinate space, but rather with `M * X_0` and `PV^-1`, which preserves AABBs.

(I hope my reasoning is sane though ... I'm getting quite uncertain at this point :sweat_smile: )

bigbike:
@mosra:
Careful. Here VP = P * V.  P is "right multiplied", not "left multiplied" by the matrix V. 

VP is the combined matrix that takes e.g. a vertex from world-space to clip-space. So extracting the planes from VP instead of P will result in world-space clip planes.

mosra:
Err ... that didn't really help, but (thanks to @Squareys) I now understood that `Frustum::fromMatrix()` is "making" the plane equations from `(PV)^-1`, which is `V^-1 P`, which is `cameraObject.absoluteTransformationMatrix*projectionMatrix.inverted()`, which finally makes sense to me.

TL;DR: you're right, it needs to be `Frustum::fromMatrix(camera.projectionMatrix()*camera.cameraMatrix())`.

bigbike:
- `(PV)^{-1}` is not equal to `V^{-1}P` unless `P1 is an involutory matrix. 

-The frustum planes are extracted directly (in clip space) from P, not P inverse. The same in the world space with "PV". So I cannot understand where (PV)^{-1} is involved.

Anyway, glad we have the same results now.

bigbike:
I suddenly came up with a good idea to identify if there is any speedup -- I just need to test it on ptex mesh!
To do it, I upgraded the resource manager so that it would compute the local aabbs for each sub-mesh in the ptex.

Then I tested it with one ptex mesh, the result was very promising! 

For a random angle, where 6 drawables of 12 were culled, the performance boosted from **35.7FPS** to **54.8FPS** (**+53.5%**)! See the attached picture.

(And this is with the code that has not been optimized yet)

Old, NO culling:
image

New, with culling:
image

bigbike:
@erikwijmans : If you would like to understand why I computed "absolute AABB" in #400. You can quickly read the comments in this PR. Thanks.

bigbike:
New functionalities have been committed in #390 #400 #472 #473.
Closing this toy example. 
Thank you @mosra for the invaluable discussions and advises.

PR #382, opened 2019-12-16
Resource manager docs

aclegg3:
## Motivation and Context

Documentation push for the ResourceManager and related files/classes. Some related refactoring and cleanup included.

Feedback, proofing, etc... welcome. 👍 

## How Has This Been Tested

Local build.

## Types of changes

## Checklist


aclegg3:
CI fails including cmake generated header `esp/gfx/configure.h` for doc linking in `ResourceManager.h`.  Works on local build. Happy for suggestions if anyone knows a simple fix. Otherwise it's not that important and can be removed.

erikwijmans:
Looks like `esp/gfx/configure.h` didn't get added to the diff :)

aclegg3:
> Looks like `esp/gfx/configure.h` didn't get added to the diff :)

I think this is actually built from `esp/gfx/configure.h.cmake` into `build/esp/gfx/configure.h`. I don't have that file locally in the `src` directory either. However, my local build works and the CI build failed. 🤔

src/esp/assets/ResourceManager.h
```diff
@@ -25,6 +28,7 @@
 #include "GltfMeshData.h"
 #include "MeshData.h"
 #include "MeshMetaData.h"
+#include "esp/gfx/configure.h"
```
erikwijmans: Why was this configure.h added?  The assets lib likely isn't getting linked against the gfx lib, so the include path for this file may not be being added.

src/esp/assets/BaseMesh.h
```diff
@@ -3,6 +3,12 @@
 // LICENSE file in the root directory of this source tree.
 
 #pragma once
+
+/** @file
+ * @brief enum @ref esp::assets::SupportedMeshType, Class @ref
+ * esp::assets::BaseMesh
```
mosra: Just a heads-up -- I remember Doxygen having strange issues with `@brief` that's on multiple lines. Maybe that isn't the case anymore and it got fixed, but be sure to check for suspicious-looking output ;)

src/esp/assets/BaseMesh.h
```diff
@@ -19,49 +25,141 @@
 namespace esp {
 namespace assets {
 
+/**
+  @brief Enumeration of mesh types supported by the simulator. Each entry
+  suggests a derived class of @ref BaseMesh implementing the specific storage
+  and processing interface to accomodate differing asset formats. Identifies the
+  derived variant of this object.
```
mosra: I'm personally making the brief quite short, one sentence at most, putting the rest in a separate paragraph like so:

```cpp
/**
  @brief Enumeration of mesh types supported by the simulator. 

  Each entry suggests a derived class of @ref BaseMesh implementing the specific storage
  and processing interface to accomodate differing asset formats. Identifies the
  derived variant of this object.
*/
```

Reason is that the brief description is used in API listings and there you don't need the full thing.

Btw., I'm usually not indenting the main doc blocks in any way to have the full 80-char width available for text. (Not sure if clang format agrees with me on that, tho.)

src/esp/assets/BaseMesh.h
```diff
@@ -19,49 +25,141 @@
...
+
+  /**
+   * @brief Meshes loaded from gltf format (i.e. .glb file). Object is likely
+   * type @ref GltfMeshData.
+   */
   GLTF_MESH = 2,
+
+  /**
+   * @brief Number of enumerated supported types.
+   */
```
mosra: Unlike everything else, with m.css enum values don't have brief docs as these would add too much noise to API listings. That means those get ignored and your enum values won't have any visible documentation. The script prints a warning if it encounters such issue, but I guess it'd be drowned among all other warnings right now anyway :)

(How to fix -- simply remove the `@brief`, leaving just plain paragraphs.)

(I'll stop complaining about `@brief` now, sorry :D)

src/esp/assets/BaseMesh.h
```diff
@@ -19,49 +25,141 @@
...
+  /**
+   * @brief Set the @ref SupportedMeshType, @ref type_, of this object.
+   * @param type The desired @ref SupportedMeshType to set.
+   * @return Whether or not the set was successful.
+   */
   bool setMeshType(SupportedMeshType type);
+
+  /**
+   * @brief Retrieve the @ref SupportedMeshType, @ref type_, of this object.
+   * @return the @ref SupportedMeshType, @ref type_, of this object.
```
mosra: Not sure if you need to repeat the `@ref SupportedMeshType` in the doc block -- the type can be clicked in the function signature as well (unless it's undocumented, which isn't).

Same elsewhere.

src/esp/assets/CollisionMeshData.h
```diff
@@ -14,11 +18,15 @@
...
+/**
+ * @brief Provides references to geometry and topology for an individual
+ * component of an asset for use in generating collision shapes for simulation.
+ * Usage: (1) for creating collision mesh/convex in @ref
+ * physics::BulletPhysicsManager and @ref physics::BulletRigidObject
+ */
 struct CollisionMeshData {
-  //! Primitive type (has to be triangle for Bullet to work)
+  //! Primitive type (has to be triangle for Bullet to work). See @ref
+  //! BulletRigidObject::constructBulletCompoundFromMeshes.
```
mosra: (Apologies if I'm not up-to-date on coding conventions, but) why `//!` here and `/**` everywhere else?

erikwijmans:
This is why it's failing: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/CMakeLists.txt#L66

The configure file needs to be run always.

aclegg3:
> This is why it's failing: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/CMakeLists.txt#L66
> 
> The configure file needs to be run always.

That looks like the culprit, thanks for finding this. I'll update it. 👍

src/esp/assets/BaseMesh.h
aclegg3: Thanks, I'll look out for this. Seems ok in this case.

src/esp/assets/BaseMesh.h
aclegg3: Looks like they don't get briefs as you said, but at least these are converted automatically into paragraphs. 🤔 I guess it makes sense to remove brief for specific enumerators but leave it for the enum type. 👍 

image

src/esp/assets/CollisionMeshData.h
aclegg3: No good reason, should be consistent. 👍

src/esp/assets/BaseMesh.h
aclegg3: Great point, thanks! 😄

aclegg3:
Any other thoughts on this before merge?

PR #383, opened 2019-12-16
[tests] Enable Code Coverage for JavaScript tests

apsdehal:
## Motivation and Context

Enables code coverage for JS files

## How Has This Been Tested

`npm run test_with_coverage` 

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,12 +10,14 @@ gpu: &gpu
 jobs:
   python_lint:
     docker:
-      - image: circleci/python:3.7
+      - image: circleci/python:3.6
     steps:
       - checkout
       - run:
           name: setup
           command: |
+              sudo apt-get update || true
+              sudo apt-get install -y python3.6-dev
               sudo pip install -U black "isort[pyproject]" numpy pytest sphinx pillow tqdm hypothesis regex
               sudo pip install -r requirements.txt --progress-bar off
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
       - run:
           name: setup
           command: |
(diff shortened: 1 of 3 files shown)


apsdehal:
@msbaines @mathfac Python lint job is failing due to unrelated unknown issue.

.circleci/config.yml
```diff
@@ -278,30 +290,25 @@ jobs:
               #TODO: instead of reinstall, do this with configuration
               ./build.sh --headless
               pytest --cov-report=xml --cov=./ --cov-append tests/test_physics.py tests/test_sensors.py
+
+              # Generate JS CodeConv
+              npm run test_with_coverage
```
mathfac: Maybe, makes sense to have a separate step for JS testing?

.circleci/config.yml
```diff
@@ -254,6 +254,18 @@ jobs:
...
+          name: Build Javascript bindings
+          command: |
+              # wait for npm and emscripten install
+              while [[ ! -f ~/emscripten_installed && ! -f ~/npm_deps_installed ]]; do sleep 2; done
+              cd habitat-sim
+              . ~/.bashrc
+              export PATH=$HOME/miniconda/bin:$PATH
+              . activate habitat
+              export EMSCRIPTEN=$HOME/emsdk/fastcomp/emscripten
+              nvm use v11.9.0
```
mathfac: For future that can be good to have versions mentioned somewhere in dependency or requirements specification and re-use it here.

apsdehal: Thanks. I will add to documentation in separate PR later.

.circleci/config.yml
apsdehal: This is inside `run sim tests` so it should be good to keep all tests in same place as it runs CPP as well as Python tests? We do have a separate step also `js_lint_and_test` which tests without code coverage.

Issue #374, opened 2019-12-10
Add support for importing semantics from https://3dscenegraph.stanford.edu for use with gibson dataset


msbaines:
## 🚀 Feature

We want to be able to use the semantic dataset from https://3dscenegraph.stanford.edu/ with the scene dataset from http://gibsonenv.stanford.edu/

msbaines:
The 3dscenegraph semantic dataset is currently limited to gibson_tiny. However, semantics for gibson_medium are expected to be release soon.

msbaines:
The mesh used for semantics is different from the mesh used for Habitat. The coordinate system is also different but both meshes. The Y and Z axis are switched but the origin is the same. We should be able to generate a semantic ply mesh from the origin .obj mesh by transformed vertex coordinates.

msbaines:
The semantic data from 3DSceneGraph is available via an [npz](https://docs.scipy.org/doc/numpy/reference/generated/numpy.savez.html) file.

To access the data:

`data = np.load(npz_path, allow_pickle=True)['output'].item()`

This will return the following dictionary:

`dict_keys(['building', 'room', 'object', 'camera', 'panorama'])`

### building

```
>>> pprint.pprint(data['building'])
{'floor_area': 35.04333970662052,
 'function': 'residential',
 'gibson_split': 'tiny',
 'id': 2,
 'name': 'Allensville',
 'num_cameras': 26,
 'num_floors': 1,
 'num_objects': 33,
 'num_rooms': 11,
 'object_inst_segmentation': array([[0.],
       [0.],
       [0.],
       ...,
       [0.],
       [0.],
       [0.]]),
 'object_voxel_occupancy': array([[0.],
       [0.],
       [0.],
       ...,
       [0.],
       [0.],
       [0.]]),
 'reference_point': array([0., 0., 0.]),
 'room_inst_segmentation': array([[10.],
       [ 8.],
       [ 8.],
       ...,
       [11.],
       [11.],
       [11.]]),
 'room_voxel_occupancy': array([[0.],
       [0.],
       [0.],
       ...,
       [0.],
       [0.],
       [0.]]),
 'size': array([9.76260114, 8.85856447, 2.50682107]),
 'volume': 201.63530725199752,
 'voxel_centers': array([[-0.98990329, -0.97658977, -0.02014603],
       [-0.98990329, -0.97658977,  0.07985397],
       [-0.98990329, -0.97658977,  0.17985397],
       ...,
[...]

PR #385, opened 2019-12-17
[datatool] add support for Gibson semantic mesh generation

msbaines:
It is a two-step process to create a Gibson semantic mesh. First
you need to extract the object_id table from the .npz file. Then
you can create the semantic_mesh from the extracted ids file
and the .obj file the .npz is based on.

Addresses: Issue #374

## Motivation and Context

Support Gibson semantic mesh.

## How Has This Been Tested

Tested semantic mesh by loading scene and semantic_mesh in WebGL demo app.

## Types of changes

## Checklist


src/utils/datatool/datatool.cpp
```diff
@@ -35,6 +38,88 @@ int createNavMesh(const std::string& meshFile, const std::string& navmeshFile) {
...
+  f << "property ushort object_id" << std::endl;
+  f << "end_header" << std::endl;
+
+  // We need to swap Y and Z to match the .glb files
+  for (size_t i = 0; i < numVerts; i++) {
+    float xyz[3];
+    unsigned char gray[] = {0x80, 0x80, 0x80};
+    xyz[0] = attrib.vertices[i * 3];
+    xyz[1] = -attrib.vertices[i * 3 + 2];
+    xyz[2] = attrib.vertices[i * 3 + 1];
```
erikwijmans: I am not sure what rotation this is (if it needs to stay, use a named quaternion to do it instead of just hard-coded logic), but we rotate the semantic ply meshes to -Y gravity assuming -Z gravity on load: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/assets/GenericInstanceMeshData.cpp#L208

tools/npZ2ids.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+
+# npz2ids - tool for extracting object_ids from 3dscenegraph dataset
+#           (https://3dscenegraph.stanford.edu/)
+
+import sys
+
+import numpy as np
```
erikwijmans: Please use `argparse` to do the argument parsing https://docs.python.org/3/library/argparse.html

tools/npZ2ids.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def main():
+    npz_path = sys.argv[1]
+    ids_path = sys.argv[2]
+    data = np.load(npz_path, allow_pickle=True)["output"].item()
+    f = open(ids_path, "wb")
```
erikwijmans: Use 
```
with open(ids_path, "wb") as f:
   ...
```

Safer and you won't need `f.close()`!

.gitmodules
```diff
@@ -49,3 +49,6 @@
 [submodule "src/deps/imgui"]
 	path = src/deps/imgui
 	url = https://github.com/ocornut/imgui.git
+[submodule "src/deps/tinyobjloader"]
+	path = src/deps/tinyobjloader
+	url = https://github.com/syoyo/tinyobjloader.git
```
mosra: Just a question -- what tinyobjloader can do but Assimp's OBJ loader can't?

src/utils/datatool/datatool.cpp
msbaines: I've fixed as you suggested. There are a couple of cons to this approach though. It will be slower and you'll lose some precision due to the floating point math. On the other hand, its probably easier to understand.

.gitmodules
msbaines: Assimp seemed too high-level and more tuned to rendering. I just needed a simple parser to get access to the vertices and faces. I looked into using Assimp and it looked a lot more complex. I also need to preserve face order so that it matches with semantic data which assumes face order.

src/utils/datatool/datatool.cpp
```diff
@@ -35,6 +38,89 @@ int createNavMesh(const std::string& meshFile, const std::string& navmeshFile) {
...
+  f << "property ushort object_id" << std::endl;
+  f << "end_header" << std::endl;
+
+  // We need to rotate to match .glb where -Z is gravity
+  const auto transform =
+      esp::quatf::FromTwoVectors(esp::vec3f::UnitY(), esp::vec3f::UnitZ());
+  for (size_t i = 0; i < numVerts; i++) {
+    unsigned char gray[] = {0x80, 0x80, 0x80};
+    float* components = &attrib.vertices[i * 3];
+    esp::vec3f vertex(components[0], components[1], components[2]);
```
erikwijmans:
```suggestion
    Eigen::Map<const esp::vec3f> vertex{components};
```

erikwijmans: You can use a map here if you want, the name "vertex" bellow will have to change tho

msbaines:
All feedback addressed. Ready for another round of review.

msbaines:
ping

src/utils/datatool/datatool.cpp
```diff
@@ -35,6 +38,89 @@ int createNavMesh(const std::string& meshFile, const std::string& navmeshFile) {
...
+  file.seekg(0, std::ios::beg);
+
+  std::vector<unsigned short> objectId(size / sizeof(short));
+  file.read(reinterpret_cast<char*>(objectId.data()), size);
+  if (!file) {
+    LOG(ERROR) << "Failed to load " << idsFile;
+    return 2;
+  }
+
+  size_t numVerts = attrib.vertices.size() / 3;
```
mathfac: Should we assert `attrib.vertices.size() % 3 == 0`?

src/utils/datatool/datatool.cpp
```diff
@@ -35,6 +38,89 @@ int createNavMesh(const std::string& meshFile, const std::string& navmeshFile) {
...
+  f << "element face " << shapes[0].mesh.num_face_vertices.size() << std::endl;
+  f << "property list uchar int vertex_indices" << std::endl;
+  f << "property ushort object_id" << std::endl;
+  f << "end_header" << std::endl;
+
+  // We need to rotate to match .glb where -Z is gravity
+  const auto transform =
+      esp::quatf::FromTwoVectors(esp::vec3f::UnitY(), esp::vec3f::UnitZ());
+  for (size_t i = 0; i < numVerts; i++) {
+    unsigned char gray[] = {0x80, 0x80, 0x80};
```
mathfac: nit `const unsigned char gray[] = {0x80, 0x80, 0x80};`, can be namespace const.

PR #389, opened 2019-12-18
Fix CollisionMeshData bug in GenericInstanceMeshData after mesh data manipulation

aclegg3:
## Motivation and Context

`CollisionMeshData` is used to reference mesh geometry during construction of collision objects for physics simulation. Manipulation of geometry data in `GenericInstanceMeshData::uploadBuffersToGPU` caused this reference set in `GenericInstanceMeshData::loadPLY` to be incorrect. This PR aims to fix this bug.

Moved from #333 as suggested by @msbaines.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/GenericInstanceMeshData.cpp
+++ b/src/esp/assets/GenericInstanceMeshData.cpp
@@ -235,13 +235,16 @@ void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
   // simply mmap a file with the vertex buffer and index buffer.
   std::vector<Mn::UnsignedInt> objectIdIndices = removeDuplicates(objectIds_);
   Mn::GL::Buffer vertices, indices;
-  indices.setTargetHint(Mn::GL::Buffer::TargetHint::ElementArray);
-  indices.setData(
+
+  std::vector<Magnum::UnsignedInt> new_indices =
       Mn::MeshTools::combineIndexedArrays(
           std::make_pair(std::cref(objectIdIndices), std::ref(objectIds_)),
           std::make_pair(std::cref(cpu_ibo_), std::ref(cpu_vbo_)),
-          std::make_pair(std::cref(cpu_ibo_), std::ref(cpu_cbo_))),
-      Mn::GL::BufferUsage::StaticDraw);
+          std::make_pair(std::cref(cpu_ibo_), std::ref(cpu_cbo_)));
+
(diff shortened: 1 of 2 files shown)


src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -262,6 +265,15 @@ void GenericInstanceMeshData::uploadBuffersToGPU(bool forceReload) {
       .setIndexBuffer(std::move(indices), 0,
                       Mn::GL::MeshIndexType::UnsignedInt);
 
+  cpu_ibo_ = std::move(new_indices);
+
+  collisionMeshData_.positions =
```
msbaines: Since this is duplicated from line 215, maybe move this into a helper?

mosra: Could be also just this, I think:

```cpp
collisionMeshData_.positions = Cr::Containers::arrayCast<Mn::Vector3>(Cr::Containers::arrayView(cpu_vbo_));
collisionMeshData_.indices = Cr::Containers::arrayCast<Mn::UnsignedInt>(Cr::Containers::arrayView(cpu_ibo_));
```

aclegg3:
> You may also want to consider adding a small test that verifies that collisionMesh is correct after uploadBuffersToGpu.

I think the only assets we have of this type are Replica which are not yet testable due to asset size. Once the RecomputeNavmesh test is active and we address the size issue, we can add a Replica scene to that test which will catch scrambled meshes. 

If you have another idea for a quick test to add we can do so. 🤔

msbaines:
> > You may also want to consider adding a small test that verifies that collisionMesh is correct after uploadBuffersToGpu.
> 
> I think the only assets we have of this type are Replica which are not yet testable due to asset size. Once the RecomputeNavmesh test is active and we address the size issue, we can add a Replica scene to that test which will catch scrambled CollisionMeshData.
> 
> If you have another idea for a quick test to add we can do so. 🤔

You could create a simple ply mesh with a cube. The test could be:

    GenericInstanceMeshData instanceMeshData;
    instanceMeshData.loadPLY(filepath);
    CollisionMeshData &collisionMesh1 =  instanceMeshData.getCollisionMeshData();
   // some assertions on collisionMesh1
    instanceMeshData.uploadBuffersToGPU(false);
   CollisionMeshData &collisionMesh2 =  instanceMeshData.getCollisionMeshData();
   // some assertions on collisionMesh2

PR #395, opened 2019-12-19
[build] initial support for sanitizers

msbaines:
## Motivation and Context

Use sanitizers to discover memory and threading issues.

## How Has This Been Tested

Modified a test to have a memory issue  and then verified that it was correctly reported.

## Types of changes

## Checklist


src/cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,56 @@
...
+
+# LeakSanitizer
+set(CMAKE_C_FLAGS_LSAN
+    "-fsanitize=leak -fno-omit-frame-pointer -g -O1"
+    CACHE STRING "Flags used by the C compiler during LeakSanitizer builds."
+    FORCE)
+set(CMAKE_CXX_FLAGS_LSAN
+    "-fsanitize=leak -fno-omit-frame-pointer -g -O1"
+    CACHE STRING "Flags used by the C++ compiler during LeakSanitizer builds."
+    FORCE)
```
mosra: I think LSan is a subset of ASan, so having just ASan should be enough.

src/cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,56 @@
...
+
+# UndefinedBehaviour
+set(CMAKE_C_FLAGS_UBSAN
+    "-fsanitize=undefined"
+    CACHE STRING "Flags used by the C compiler during UndefinedBehaviourSanitizer builds."
+    FORCE)
+set(CMAKE_CXX_FLAGS_UBSAN
+    "-fsanitize=undefined"
+    CACHE STRING "Flags used by the C++ compiler during UndefinedBehaviourSanitizer builds."
+    FORCE)
```
mosra: FYI, I'm not warning-free on msan/ubsan with magnum (got some annoying false positives all over the place), so these may produce a lot of noise. I'm warning-free with ASan and TSan tho, except for leaks in 3rd party libs (Assimp, NVidia drivers, ...).

src/cmake/Sanitizers.cmake
msbaines: Leaky sanitizer by itself can be used with much lower overhead (no instrumentation). This can be handy sometimes:

https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#stand-alone-mode

src/cmake/Sanitizers.cmake
msbaines: Currently I am only focused on enabling ASan for build but it is handy to have the other available to debug an issue.

src/cmake/Sanitizers.cmake
mosra: Coo, thanks, TIL! 👍

src/esp/gfx/test/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
 find_package(Magnum REQUIRED Primitives)
 
+if(NOT CMAKE_BUILD_TYPE MATCHES "ASAN")
```
mosra: Better would be to add a suppression file that matches `libnvidia-eglcore.so` -- we still want to be informed about *our* leaks :) Here's how I do it in magnum: 

- https://github.com/mosra/magnum/blob/master/package/ci/leaksanitizer.conf 
- https://github.com/mosra/magnum/blob/e72a0459c825621beb0298fd76c12b1eeaea8301/package/ci/travis-desktop.sh#L55

msbaines: Cool. Thanks for the pointer. I will try this.

eundersander:
Closing this stale PR.

PR #398, opened 2019-12-20
Fix memory leak in Simulator::recomputeNavMesh

aclegg3:
## Motivation and Context

This patch aims to fix a memory leak in Simulator::recomputeNavMesh as pointed out here: https://github.com/facebookresearch/habitat-sim/pull/333#discussion_r360183036

## How Has This Been Tested

Open to thoughts on memory leak testing. 🤔 

## Types of changes

## Checklist


aclegg3:
@erikwijmans Think we may also have a problem [here](https://github.com/facebookresearch/habitat-sim/blob/47f02620a9f1d8c2bbde97e8ebe4176d26c47f5f/src/esp/nav/PathFinder.cpp#L517) with navData?

`if (!dtCreateNavMeshData(&params, &navData, &navDataSize)) {`

src/esp/nav/PathFinder.cpp
```diff
@@ -519,6 +519,10 @@ bool PathFinder::build(const NavMeshSettings& bs,
       return false;
     }
 
+    if (navMesh_) {
+      dtFree(navMesh_);
```
msbaines: I think this should be dtFreeNavMesh. Otherwise, the destructor won't get called.

src/esp/nav/PathFinder.cpp
```diff
@@ -546,6 +550,9 @@ bool PathFinder::build(const NavMeshSettings& bs,
 }
 
 bool PathFinder::initNavQuery() {
+  if (navQuery_) {
+    dtFree(navQuery_);
```
msbaines: dtFreeNavQuery

src/esp/nav/PathFinder.cpp
```diff
@@ -546,6 +550,9 @@ bool PathFinder::build(const NavMeshSettings& bs,
 }
 
 bool PathFinder::initNavQuery() {
+  if (navQuery_) {
```
mathfac: Looks like canonical way is based on `free` method is:
 ```if (navQuery_) {
    dtFreeNavMeshQuery(navQuery_);
    navQuery_ = 0;
  }```

src/esp/nav/PathFinder.cpp
```diff
@@ -519,6 +519,10 @@ bool PathFinder::build(const NavMeshSettings& bs,
       return false;
     }
 
+    if (navMesh_) {
```
mathfac: Looks like canonical way is based on free method is:
```
  if (navMesh_) {
    dtFreeNavMesh(navMesh_);
    navMesh_ = 0;
  }
```

src/esp/nav/PathFinder.cpp
mosra: What do you mean? Setting pointers to zero after a free() might hide accidental double-frees, which is not good. Accidental use-after-free is pretty reliably caught by ASan, but accidental double-frees not if you reset the pointer to 0.

From my experience on previous projects, when we agreed to remove a `safeDelete()` helper that was doing exactly this, we immediately discovered a lot of previously hidden bugs where destructors were getting called more than once and similar serious issues.

mosra:
> Might be simpler to create a wrapper class.

There's this also: https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1ScopeGuard.html :)

erikwijmans:
A `std::unique_ptr` with a custom deleter would probably be best: https://en.cppreference.com/w/cpp/memory/unique_ptr

src/esp/nav/PathFinder.cpp
msbaines: > Use RAII:
> 
> std::unique_ptr<dtNavMesh, decltype(&dtFreeNavMesh)> navMesh_;
> 
> navMesh_ = std::unique_ptr<dtNavMesh, decltype(&dtFreeNavMesh)>(dtAllocNavMesh(), &dtFreeNavMesh);
> 
> Might be simpler to create a wrapper class.

This could be simplified to:

// In some utility header

struct NavMeshDeleter {
    void operator()(dtNavMesh* mesh) const {
        dtFreeNavMesh(mesh);
    }
};
using NavMeshUniquePtr = std::unique_ptr<dtNavMesh, NavMeshDeleter>;

// Usage:

NavMeshUniquerPtr navMesh_;
navMesh_ = NavMeshUniquePtr(dtAllocNavMesh());

erikwijmans:
> @erikwijmans Think we may also have a problem [here](https://github.com/facebookresearch/habitat-sim/blob/47f02620a9f1d8c2bbde97e8ebe4176d26c47f5f/src/esp/nav/PathFinder.cpp#L517) with navData?
> 
> `if (!dtCreateNavMeshData(&params, &navData, &navDataSize)) {`

@aclegg3 No, we don't, the navMesh will take ownership of `navData` later on.

aclegg3:
Thanks for all the discussion, I've implemented @msbaines solution. Thoughts?

src/esp/nav/PathFinder.h
```diff
@@ -15,13 +15,27 @@ class dtNavMeshQuery;
...
+void dtFreeNavMeshQuery(dtNavMeshQuery* navmeshQuery);
+
 namespace esp {
 // forward declaration
 namespace assets {
 struct MeshData;
 }
 namespace nav {
 
+struct NavMeshDeleter {
```
erikwijmans: These can be put in the class definition or in an additional `impl` namespace as to not populate the `nav` namespace.

aclegg3:
> Thank you @aclegg3. Just one suggestion :)

No problem, thanks for catching this. 👀

PR #399, opened 2019-12-20
Bullet margin fix and Aabb query

aclegg3:
## Motivation and Context
Bullet collision shapes have a [default margin on 0.04](https://pybullet.org/Bullet/BulletFull/btCollisionMargin_8h.html#ac3542bba5bfc2fcc6bbb2c8c991b947c). 

We allow the user to assign a single collision margin value for each rigid body. We assign this to the root of the btCompoundShape object which holds all sub-components of the collision shape. We have not been explicitly overriding the default margins for these sub-objects. This caused a difference between joined and component hierarchy objects in the actual collision margins of the whole, as sub-component margins are accumulated additively (i.e. default margins (0.04) on sub-components would be added to the user defined margin).

This PR fixes said issue.

Sub-thesis: 
 - Fix `PhysicsTest.cpp` build without `--bullet` issue. (identified by @matthewjmay ). 
 - Added here to avoid git churn because this PR contains the first functionality which requires the linking and better fix on master would be to remove the broken include.

## How Has This Been Tested

This PR introduces a test to be sure that different collision shape modes (scene, joined, hierarchal) will result in the correct margin. To do so, we add functionality for querying collision Aabb from Bullet and check this against ground truth for a known test asset composed of multiple parts. This test will be extended to encompass additional collision modes such as bounding boxes when added.

## Types of changes

## Checklist


aclegg3:
Ready for review.

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -177,6 +177,8 @@ bool BulletRigidObject::initializeObject(
   if (joinCollisionMeshes) {
     btTransform t;
     t.setIdentity();
+    bObjectConvexShapes_.back()->setMargin(0.0);
```
msbaines: I haven't dug too deep in the physics code yet but why is this here and not in constructBulletCompoundFromMeshes

src/tests/PhysicsTest.cpp
```diff
@@ -131,3 +134,62 @@ TEST_F(PhysicsManagerTest, JoinCompound) {
...
+
+    std::pair<Magnum::Vector3, Magnum::Vector3> AabbScene =
+        bPhysManager->getSceneCollisionShapeAabb();
+
+    std::pair<Magnum::Vector3, Magnum::Vector3> AabbOb0 =
+        bPhysManager->getCollisionShapeAabb(objectId0);
+    std::pair<Magnum::Vector3, Magnum::Vector3> AabbOb1 =
+        bPhysManager->getCollisionShapeAabb(objectId1);
+
+    std::pair<Magnum::Vector3, Magnum::Vector3> objectGroundTruth =
```
msbaines: Is this test testing all the places where new code was added.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -195,6 +195,18 @@ double BulletPhysicsManager::getSceneRestitutionCoefficient() const {
       ->getRestitutionCoefficient();
 }
 
+const std::pair<Magnum::Vector3, Magnum::Vector3>
```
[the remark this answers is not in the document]
aclegg3: Could be. Also could be used to construct a `Magnum::Range3D`. Currently, this just matches the return type format of the Bullet query function. Since it is not exposed to the base `PhysicsManager` API, I left it that way.

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: `constructBulletCompoundFromMeshes` is a recursive function. 

With the `join` flag, as in this case, it is building up a single convex shape from all components. This will not be complete until after the recursion is done. 

Without the `join` flag, each level of recursion builds a new object, so it makes sense to set margin upon construction in that case.

src/tests/PhysicsTest.cpp
aclegg3: Yes, there are 3 cases currently:
1. Scenes (concave mesh collision objects)
2. Joined objects (a single convex)
3. Hierarchal objects (a compound of convex components)

src/esp/physics/bullet/BulletRigidObject.cpp
msbaines: The constructBulletCompoundFromMeshes code might be easier to read with two functions. A top-level function and a recursive helper. It feels like functionality that should be part of constructBulletCompoundFromMeshes is part of initializeObject.

src/esp/physics/bullet/BulletPhysicsManager.cpp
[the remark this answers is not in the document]
msbaines: Structs are sometimes preferred to pair when the type is used in a large scope (especially if it is used in a header file). pairs are preferred for smaller scopes (e.g. a single function) and can result in hard to read code when used in a large scope. It is somewhat analogous to the use of the variable name "i" in a for loop.

src/tests/PhysicsTest.cpp
```diff
@@ -131,3 +134,62 @@ TEST_F(PhysicsManagerTest, JoinCompound) {
...
+        std::pair<Magnum::Vector3, Magnum::Vector3>({-1.1, -1.1, -1.1},
+                                                    {1.1, 1.1, 1.1});
+    std::pair<Magnum::Vector3, Magnum::Vector3> sceneGroundTruth =
+        std::pair<Magnum::Vector3, Magnum::Vector3>({-1.0, -1.0, -1.0},
+                                                    {1.0, 1.0, 1.0});
+
+    // Cr::Utility::Debug() << "aabb scene: " << AabbScene;
+    // Cr::Utility::Debug() << "aabb ob0: " << AabbOb0;
+    // Cr::Utility::Debug() << "aabb ob1: " << AabbOb1;
```
msbaines: These should be uncommented or removed.

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: Fair point. 👍 
A refactor like this should probably be a separate PR after this bug fix is merged.

Issue #390, opened 2019-12-18
Compute absolute AABB for mesh components in a static scene.


bigbike:
## 🚀 Feature
The scene consists of many mesh components. Compute a AABB  for each component in the world coordinate frame.

Note this AABB is different from the AABB computed in the component's local coordinate frame.

Also, if you rotate the local AABB to the world coordinate frame, which becomes an OBB, then you build an AABB of this OBB. This is an "AABB approximation", NOT what we request here.

## Motivation
To accelerate frustum culling.

More details, see discussion: #380

Skylion007:
@bigbike Can we close this issue or add more detail?

bigbike:
@Skylion007 Yes, we can. It has been resolved.

PR #400, opened 2019-12-21
[culling] Compute absolute AABB (NOT the existing local AABB) for each mesh component of the static scene (ptex, mp3d ONLY)

bigbike:
## Motivation and Context
- Computes an AABB in world space for each mesh component of the static scene;
- Applies to ptex mesh and MP3D mesh;
- Will not compute absolute AABB for dynamic object, or instance scene;

- Each mesh component corresponds to a static drawable. Static drawables have fixed absolute transformations during the runtime. Repeat computation is not necessary;

- absolute AABB will be used in the frustum culling process;

- more details: #380

Existing issue: #390 

## How Has This Been Tested
local

## Types of changes

## Checklist


bigbike:
Code is ready to be reviewed.

Thanks.

bigbike:
@msb:
I already added: // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
But clang-tidy still complains the potential memory leaks.

src/esp/assets/ResourceManager.h
```diff
@@ -255,17 +255,52 @@ class ResourceManager {
...
+  //
+  // pair: <Drawable, meshID>
+  //
+  // -) non-ptex mesh:
+  // meshID is the global index into meshes_.
+  // meshes_[meshID] is the BaseMesh corresponding to the drawable;
+  //
+  // -) ptex mesh:
+  // meshID is the index of the submesh corresponding to the drawable;
+  std::vector<std::pair<std::reference_wrapper<esp::gfx::Drawable>, uint32_t>>
```
erikwijmans: This vector is concerning.  Are we sure that we won't try to access the drawable after it gets deleted?

src/esp/assets/ResourceManager.cpp
```diff
@@ -649,6 +684,113 @@ Magnum::Range3D ResourceManager::computeMeshBB(BaseMesh* meshDataGL) {
...
+  }  // iEntry
+}
+
+std::vector<Mn::Matrix4> ResourceManager::computeAbsoluteTransformations() {
+  // sanity check
+  if (staticDrawableInfo_.size() == 0) {
+    return std::vector<Mn::Matrix4>{};
+  }
+
+  // basic assumption is that all the drawables are in the same scene;
```
erikwijmans: I think this assumption is wrong.  The resource manager may have multiple meshes loaded.  i.e. like an MP3D mesh where you have both the color mesh and semantic mesh loaded.  That'll also be true for semantic gibson meshes.

bigbike: See my comments above in ResourceManager.cpp. The program will not collect drawables from the semantic mesh.

src/esp/assets/ResourceManager.h
bigbike: This is a valid concern. However, it is unlikely happen.
This variable so far is used as a "temporary" variable that is to collect drawables, and meshIDs at the loading stage.  After that, it will not be used any more.

It is needed in multiple functions, so that I put it here as a member variable to avoid passing it as a function parameter.

I think solution is easy:
What I can do to address your concern is to clear this vector after its job is completed in loading stage.

bigbike:
> It seems that this PR introduces a second way to compute bounding boxes for a mesh as there is already the functionality to compute bounding boxes for physics related things in the code base. @bigbike @aclegg3 can you guys work together to figure out a way to unify this?

These 2 AABBs are different.
One is in local model frame, and the other is in world coordinate frame.

src/esp/assets/ResourceManager.cpp
bigbike: Also see my description in the "Motivation and Context" section.

bigbike:
> These 2 AABBs are different.
> One is in local model frame, and the other is in world coordinate frame.

Note:
If you apply absolute transformation to the local AABB, it would become an OBB in world space. Then if you create a "new AABB" of this OBB, this "new AABB" is different from the AABB that is computed here. The AABB computed here usually is much tighter.

bigbike:
The correctness of the computed absolute AABBs was tested in the branch called "culling": 
https://github.com/facebookresearch/habitat-sim/compare/culling

- When frustum culling is enabled, it uses the absolute AABBs computed in this PR.

- The rendering is visually correct with/without frustum culling enabled. This is tested and confirmed with one MP3D model, and one PTex model respectively.

- It improves the FPS and can be identified when loading the PTex model. (Cannot tell the FPS increase when loading MP3D model due to the v-sync lock on mac.)

See the attached:
NO Culling:
image

With Culling:
image

erikwijmans:
> > These 2 AABBs are different.
> > One is in local model frame, and the other is in world coordinate frame.
> 
> Note:
> If you apply absolute transformation to the local AABB, it would become an OBB in world space. Then if you create a "new AABB" of this OBB, this "new AABB" is different from the AABB that is computed here. The AABB computed here usually is much tighter.

Agreed, we'd need to compute an OBB in the first place.  Then that can be either converted to an AABB or transformed and converted to an AABB.

src/esp/assets/ResourceManager.cpp
```diff
@@ -65,6 +69,13 @@ namespace assets {
 bool ResourceManager::loadScene(const AssetInfo& info,
                                 scene::SceneNode* parent, /* = nullptr */
                                 DrawableGroup* drawables /* = nullptr */) {
+  // we only compute absolute AABB for evey mesh component when loading ptex
+  // mesh, or general mesh (e.g., MP3D)
+  if (info.type == AssetType::FRL_PTEX_MESH ||
+      info.type == AssetType::MP3D_MESH) {
+    computeAbsoluteAABBs_ = true;
```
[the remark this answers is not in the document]
erikwijmans: Maybe add a `staticDrawableInfo_.clear();` at the top of this method?  It should be a no-op all the time, but makes it clear that `staticDrawableInfo_.clear();` is related to only the current scene we are loading.

src/esp/assets/ResourceManager.h
erikwijmans: Is the drawable being used in anyway besides getting access to the node?  If so, why not store a reference to the scene node?  The scene node and drawable should have very similar lifetimes so it should be equally safe.

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
bigbike: Yes, it is addressed. See below.

src/esp/assets/ResourceManager.cpp
```diff
@@ -101,6 +112,32 @@ bool ResourceManager::loadScene(const AssetInfo& info,
...
+      computePTexMeshAbsoluteAABBs(*(meshes_[metaData.meshIndex.first].get()));
+    } else if (info.type == AssetType::MP3D_MESH) {
+      computeGeneralMeshAbsoluteAABBs();
+    }
+  }
+
+  if (computeAbsoluteAABBs_) {
+    computeAbsoluteAABBs_ = false;
+    // this is to prevent it from being misused in the future
+    staticDrawableInfo_.clear();
```
bigbike: @erikwijmans : I addressed your concern here, not at the beginning of this function.
Every time, after the job is done, it will be cleared.

src/esp/assets/ResourceManager.h
bigbike: Yes, good suggestion. In terms of functionality, both methods are the same. But I read my code, and what you suggested can potentially lead to simpler code. So why not? Let me try to change it.

src/esp/assets/ResourceManager.cpp
erikwijmans: Yeah, I saw this.  I was thinking a clear at the top may also be useful for readability.  That class member is really just a local variable to this function.

bigbike:
 @mosra: I appreciate if you can take a look as well. Thanks.

src/esp/assets/ResourceManager.cpp
```diff
@@ -751,8 +893,14 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
         scene::SceneNode& node = parent->createChild();
         const quatf transform = info.frame.rotationFrameToWorld();
         node.setRotation(Magnum::Quaternion(transform));
-        new gfx::PTexMeshDrawable{node, *ptexShader, *pTexMeshData, jSubmesh,
-                                  drawables};
+
+        // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
+        gfx::PTexMeshDrawable* d = new gfx::PTexMeshDrawable{
```
erikwijmans: I believe this can just be `new gfx::PTexMeshDrawable{node, *ptexShader, *pTexMeshData, jSubmesh, drawables};` again.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1221,12 +1371,23 @@ void ResourceManager::createDrawable(
   } else if (shaderType == INSTANCE_MESH_SHADER) {
     auto* shader =
         static_cast<gfx::PrimitiveIDShader*>(getShaderProgram(shaderType));
-    node.addFeature<gfx::PrimitiveIDDrawable>(*shader, mesh, group);
+
+    // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
+    gfx::PrimitiveIDDrawable* d =
+        new gfx::PrimitiveIDDrawable{node, *shader, mesh, group};
```
erikwijmans: This change can be undone.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1221,12 +1371,23 @@ void ResourceManager::createDrawable(
...
+      staticDrawableInfo_.emplace_back(node, *meshID);
+    }
   } else {  // all other shaders use GenericShader
     auto* shader =
         static_cast<Magnum::Shaders::Flat3D*>(getShaderProgram(shaderType));
-    node.addFeature<gfx::GenericDrawable>(*shader, mesh, group, texture,
-                                          objectId, color);
+
+    // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
+    gfx::GenericDrawable* d = new gfx::GenericDrawable{
```
erikwijmans: Same here.

src/esp/assets/ResourceManager.cpp
bigbike: Yes, right. Good catch. It was needed before as I stored drawables in previous commits. (Now I store the scene node).

src/esp/assets/ResourceManager.cpp
bigbike: Yes, right.

src/esp/assets/ResourceManager.cpp
bigbike: > That class member is really just a local variable

Yes, correct. I made it a member variable to avoid long and tedious argument list for a couple of member functions.
But will it make you more comfortable (putting it in the function arguments)?

src/esp/assets/BaseMesh.h
```diff
@@ -40,6 +40,9 @@ class BaseMesh {
 
   virtual Magnum::GL::Mesh* getMagnumGLMesh() { return nullptr; }
   virtual Magnum::GL::Mesh* getMagnumGLMesh(int) { return nullptr; }
+  Corrade::Containers::Optional<Magnum::Trade::MeshData3D>& getMeshData() {
```
aclegg3: Do you need this? Looks like you just want the vbo which should be accurate in `collisionMeshData_`. 🤔  I thought the original purpose of `collisionMeshData_` was to avoid public reference access to mesh data.

src/esp/assets/ResourceManager.cpp
```diff
@@ -65,6 +69,13 @@ namespace assets {
 bool ResourceManager::loadScene(const AssetInfo& info,
                                 scene::SceneNode* parent, /* = nullptr */
                                 DrawableGroup* drawables /* = nullptr */) {
+  // we only compute absolute AABB for evey mesh component when loading ptex
```
aclegg3: typo: every

src/esp/assets/ResourceManager.h
```diff
@@ -255,17 +255,53 @@ class ResourceManager {
...
+
+  void computePTexMeshAbsoluteAABBs(BaseMesh& baseMesh);
+
+  /**
+   * @brief Compute the absolute AABBs for drawables in general mesh (e.g.,
+   * MP3D)world space
+   */
+  void computeGeneralMeshAbsoluteAABBs();
+
+  /*
```
aclegg3: `/**`

src/esp/assets/ResourceManager.h
```diff
@@ -255,17 +255,53 @@ class ResourceManager {
...
+  // -) non-ptex mesh:
+  // meshID is the global index into meshes_.
+  // meshes_[meshID] is the BaseMesh corresponding to the drawable;
+  //
+  // -) ptex mesh:
+  // meshID is the index of the submesh corresponding to the drawable;
+  std::vector<
+      std::pair<std::reference_wrapper<esp::scene::SceneNode>, uint32_t>>
+      staticDrawableInfo_;
+  bool computeAbsoluteAABBs_ = false;
```
aclegg3: Reformat the comments for docs? 😃

src/esp/assets/BaseMesh.h
bigbike: By just looking at the name, `collisionMeshData_`, I was thinking it would only be used in the physics module. 

Good to know I can use it to get the geometry data. (It was really a pain point when I wrote this PR.)

bigbike:
@aclegg3 :
>Would be curious to know the difference between the approximate and precise bounding boxes computed over the Gibson or Replica datasets.

Good question. To answer it, I need to do experiment to collect data and compare the culling rate.

To compute the absolute AABB instead of approximation, better accuracy is indeed a purpose. However, there is another even more important reason -- save computation.

Because, culling test can only be done when frustum and the bounding box are in the same space (e.g., world, camera, or model space). And to make the test efficient, the bounding box must be a AABB. 

If we use the existing "local" AABB in the scene node (developed by you), we will have to transform the frustum by a constant transformation matrix from clip space to model space (because the bounding box is a AABB only in the model space). Simply put, that means we need to compute the "mvp" for every object in every frame. But with the absolute AABB, we only need to compute "vp" once for all objects in every frame.

src/esp/assets/BaseMesh.h
bigbike: I read the code.
Actually I will finish computing the absolute AABB in the default loadScene(...). It happened even before the collision data are initialized.

> I thought the original purpose of collisionMeshData_ was to avoid public reference access to mesh data.
I have no idea about this purpose.
I guess what I can do is to make this getMeshData(...) protected.

bigbike: Checked the utility Viewer. The answer is no. We have cases that will directly call the "default" loadScene without computing the collision mesh data.

src/esp/assets/ResourceManager.cpp
bigbike: @erikwijmans : I tried to put this "local" variable into the argument list for a couple of related functions. It looked very ugly, and thus significantly decreased the code readability. I will have to give it up.

I will put "a clear at the top" to address your concern.

src/esp/assets/BaseMesh.h
aclegg3: It looks like `BaseMesh::collisionMeshData_` is not set at all for PTex scene loading (with physics or not). 

However, for other asset types such as `Mp3dInstanceMeshData`, the `collisionMeshData_` is set during the mesh load (e.g. in `Mp3dInstanceMeshData::loadMp3dPLY`) which happens in default loadScene. In this case, the optional `BaseMesh::meshData_` doesn't seem to ever be set in favor of alternative buffers (e.g. `GenericInstanceMeshData::cpu_vbo_`) which are then pointed to by the `collisionMeshData_` directly. 

Looks like a possible refactor to unify this. 🤔

bigbike: We will not do frustum culling on instance meshes.
So it is OK now that BaseMesh::meshData_ is not set for Mp3dInstanceMeshData .

Yes, we need to unify it.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1206,13 +1358,14 @@ void ResourceManager::addPrimitiveToDrawables(int primitiveID,
...
   createDrawable(ShaderType::COLORED_SHADER, primitive_meshes_[primitiveID],
-                 node, drawables);
+                 node, Corrade::Containers::NullOpt, drawables);
 }
 
 void ResourceManager::createDrawable(
     const ShaderType shaderType,
     Magnum::GL::Mesh& mesh,
     scene::SceneNode& node,
+    Corrade::Containers::Optional<uint32_t> meshID, /* = NullOpt */
```
msbaines: We seem to be over-loading createDrawable with a lot o responsiblities. Not sure if the code might be simpler if it the new functionality were added to addMeshToDrawables instead.

src/esp/assets/ResourceManager.h
```diff
@@ -487,13 +487,48 @@ class ResourceManager {
...
+   * we will compute the absolute AABB pair: <Drawable's scene node, meshID>
+   *
+   * -) non-ptex mesh:
+   * meshID is the global index into meshes_.
+   * meshes_[meshID] is the BaseMesh corresponding to the drawable;
+   *
+   * -) ptex mesh:
+   * meshID is the index of the submesh corresponding to the drawable;
+   */
+  std::vector<
```
msbaines: Maybe use a proper struct instead of a pair since this is used in such a large scope.

It also seems like this should really be a local variable instead of a member for this.

src/esp/scene/SceneNode.h
```diff
@@ -75,6 +76,9 @@ class SceneNode : public MagnumObject {
   //! set local bounding box for meshes stored at this node
   void setMeshBB(Magnum::Range3D meshBB) { meshBB_ = meshBB; };
 
+  //! set the global bounding box for mesh stored in this node
+  void setAbsoluteAABB(Magnum::Range3D aabb) { aabb_ = aabb; };
```
msbaines: setter but no getter? How is this used? 

Oh. Its protected. private with a getter is generally preferred. Raw ivar access is fragile.

bigbike:
Synced up with @msbaines offline. His comments on data and code structures, tests makes a lot of sense. I really appreciated it. Will change accordingly.

src/esp/scene/SceneNode.h
bigbike: Yes, I actually have the getter in the branch "culling", which would be the next PR after this one committed in.

It looks like this:

```CPP
  Corrade::Containers::Optional<Magnum::Range3D> getAbsoluteAABB() {
    return aabb_;
  }
```

I can move it to this PR.

bigbike:
Ready for another round review while I am writing the unit tests.

src/esp/assets/ResourceManager.cpp
```diff
@@ -649,6 +687,113 @@ Magnum::Range3D ResourceManager::computeMeshBB(BaseMesh* meshDataGL) {
   return Magnum::Range3D{
       Magnum::Math::minmax<Magnum::Vector3>(meshData.positions)};
 }
+void ResourceManager::computePTexMeshAbsoluteAABBs(BaseMesh& baseMesh) {
```
erikwijmans: IMO, it makes sense to `#ifdef` the function definition itself instead of the implementation.  If we aren't building with ptex, then we can assume this method should/will never be called, so there shouldn't be any code that requires this blank implementation right?

bigbike: Makes sense to me. Will do.

bigbike:
Dear reviewers,

Just added the test as requested.
Sincerely appreciate if you can take another look. 
Thank you.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1257,10 +1413,12 @@ void ResourceManager::createDrawable(
   } else if (shaderType == INSTANCE_MESH_SHADER) {
     auto* shader =
         static_cast<gfx::PrimitiveIDShader*>(getShaderProgram(shaderType));
+
```
msbaines: unrelated white-space change

src/esp/assets/ResourceManager.cpp
```diff
@@ -1257,10 +1413,12 @@ void ResourceManager::createDrawable(
...
     node.addFeature<gfx::PrimitiveIDDrawable>(*shader, mesh, group);
   } else {  // all other shaders use GenericShader
     auto* shader =
         static_cast<Magnum::Shaders::Flat3D*>(getShaderProgram(shaderType));
+
```
msbaines: unrelated white-space change

src/esp/assets/ResourceManager.cpp
```diff
@@ -683,6 +721,113 @@ Magnum::Range3D ResourceManager::computeMeshBB(BaseMesh* meshDataGL) {
...
+std::vector<Mn::Matrix4> ResourceManager::computeAbsoluteTransformations() {
+  // sanity check
+  if (staticDrawableInfo_.size() == 0) {
+    return std::vector<Mn::Matrix4>{};
+  }
+
+  // basic assumption is that all the drawables are in the same scene;
+  // so use the 1st element in the vector to obtain this scene
+  auto* scene = dynamic_cast<Mn::SceneGraph::Scene<
```
msbaines: dynamic_cast to a pointer will return nullptr on error instead of throwing an exception so you should check the pointer or use a dynamic_cast to reference.

src/esp/assets/ResourceManager.cpp
msbaines: Still #ifdef'ing the implementation and not the whole function.

bigbike: I actually cannot add `#ifdef` to the entire function definition since it would cause the CI test failure. When compiling js, it complained undefined symbols. 

If you look at the other functions, e.g., `loadPTexMeshData`, we did the same thing, namely, adding the conditionals on the body, not the entire definition. So the function is just empty.

I guess I will just leave it like this.

@msbaines  @erikwijmans

erikwijmans: I think you also need to `#ifdef` it in the header file.

PR #407, opened 2019-12-28
[esp/scene] Added support of object's centers and bounding boxes to Gibson semantic scenes

mathfac:
## Motivation and Context
To leverage [3dscenegraph](https://3dscenegraph.stanford.edu/) semantic annotation spatial information added support of object's bounding boxes to Gibson semantic scenes. Related to issue https://github.com/facebookresearch/habitat-sim/issues/374 and depends on PR https://github.com/facebookresearch/habitat-sim/pull/406.

## How Has This Been Tested
Added the integration test and run it locally with `*.scn` data:
```
Running main() from /private/home/maksymets/habitat-sim/src/deps/googletest/googletest/src/gtest_main.cc
[==========] Running 2 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 1 test from GibsonSceneTest
[ RUN      ] GibsonSceneTest.Basic
[       OK ] GibsonSceneTest.Basic (2 ms)
[----------] 1 test from GibsonSceneTest (2 ms total)

[----------] 1 test from GibsonSemanticSimTest
[ RUN      ] GibsonSemanticSimTest.Basic
[       OK ] GibsonSemanticSimTest.Basic (5451 ms)
[----------] 1 test from GibsonSemanticSimTest (5451 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 2 test cases ran. (5453 ms total)
[  PASSED  ] 2 tests.
```

## Types of changes


src/esp/scene/GibsonSemanticScene.cpp
```diff
@@ -19,6 +19,18 @@ namespace scene {
...
+vec3f jsonToVec3f(const esp::io::JsonGenericValue& jsonObject) {
+  vec3f vec;
+  size_t dim = 0;
+  ASSERT(jsonObject.GetArray().Size() == vec.size());
+  for (const auto& element : jsonObject.GetArray()) {
+    vec[dim++] = element.GetFloat();
+  }
+  return vec;
+}
+}  // namespace
```
aclegg3: Should this live in `esp/io/json.*` with similar functions such as `jsonToString` ?

src/esp/scene/GibsonSemanticScene.cpp
```diff
@@ -67,7 +79,17 @@ bool SemanticScene::loadGibsonHouse(
...
+
+    const auto& jsonCenter = jsonObject["location"];
+    if (!jsonCenter.IsNull()) {
+      vec3f center = jsonToVec3f(jsonCenter);
+      const auto& jsonSize = jsonObject["size"];
+      vec3f size = vec3f::Zero();
+      if (!jsonSize.IsNull()) {
+        size = jsonToVec3f(jsonSize);
+      }
+      object->obb_ = geo::OBB(center, size, quatf::Identity());
```
erikwijmans: The center and size will need to transformed by the `worldRotation` parameter and the `worldRotation` will need to be to `quatf::FromTwoVectors(-vec3f::UnitZ(), esp::geo::ESP_GRAVITY)`.

aclegg3: Should that be applied to the transform of the OBB instead of the center/size directly, then?

mathfac: @erikwijmans I was sure that's already done during converting `.npz` files [here](https://github.com/facebookresearch/habitat-sim/blob/998f3f7ebd91e7359ae7c585da3e507d1a509fa3/tools/npz2scn.py#L16). Let me know, if I am wrong.

erikwijmans: The comment says `# Transform coordinates by rotating Y-axis to Z-axis` and that seems to be what the code is doing also, so we do need to go Z to Y again. (I do agree with decision btw, it makes sense to save the meshes and annotations in the same coordinate frame, -Z gravity) CC @msbaines 

@aclegg3 The center of the OBB is assumed to already be in the correct coordinate frame and won't get applied to the center.

src/esp/scene/GibsonSemanticScene.cpp
```diff
@@ -67,7 +79,17 @@ bool SemanticScene::loadGibsonHouse(
       scene.categories_.push_back(category);
       object->category_ = std::move(category);
     }
-    // TODO(msb) add support for aabb
+
+    const auto& jsonCenter = jsonObject["location"];
+    if (!jsonCenter.IsNull()) {
```
msbaines: Log a warning once for this.

src/esp/gfx/Simulator.cpp
```diff
@@ -57,6 +57,10 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
     houseFilename = cfg.scene.filepaths.at("house");
   }
 
+  if (!io::exists(houseFilename)) {
```
msbaines: You can avoid this by adding a "house" entry to your configuration in  cfg.scene.filepaths (see line 56)

msbaines:
The .obj mesh used for main gibson (what 3Dscenegraph is built from) has a rotated coordinate system when compared to the gibson gltf meshes we are using for rgb in Habitat. My utility rotates the coordinate system to match habitat and outputs to .ply semantic mesh. I transformed the bounding box info from the 3Dscenegraph .npz dataset in the same way. My goal was to have a consistent coordinated system. I verified the semantic meshes using the mesh highlighting logic we have in WebGL but did not have time before holidays to verify the bounding boxes were correct.

We have coordinate flipping logic in Asset.cpp that may be relevant here:

```
AssetInfo AssetInfo::fromPath(const std::string& path) {
  using Corrade::Utility::String::endsWith;
  AssetInfo info{AssetType::UNKNOWN, path};

  if (endsWith(path, "_semantic.ply")) {
    info.type = AssetType::INSTANCE_MESH;
  } else if (endsWith(path, "mesh.ply")) {
    info.type = AssetType::FRL_PTEX_MESH;
    info.frame = {quatf::FromTwoVectors(geo::ESP_GRAVITY, -vec3f::UnitZ())};
  } else if (endsWith(path, "house.json")) {
    info.type = AssetType::SUNCG_SCENE;
  } else if (endsWith(path, ".glb")) {
    // assumes MP3D glb with gravity = -Z
    info.type = AssetType::MP3D_MESH;
    // Create a coordinate for the mesh by rotating the default ESP
    // coordinate frame to -Z gravity
    info.frame = {quatf::FromTwoVectors(geo::ESP_GRAVITY, -vec3f::UnitZ())};
  }

  return info;
}
```

I have not studied the above code so don't know exactly what it does.

src/esp/gfx/Simulator.cpp
mathfac: Thank you, missed that logic.

src/esp/scene/GibsonSemanticScene.cpp
```diff
@@ -19,6 +19,18 @@ namespace scene {
 
 constexpr int kMaxIds = 10000; /* We shouldn't every need more than this. */
 
+namespace {
+vec3f jsonToVec3f(const esp::io::JsonGenericValue& jsonObject) {
```
msbaines: Should this be removed?

src/esp/scene/GibsonSemanticScene.cpp
```diff
@@ -67,7 +79,24 @@ bool SemanticScene::loadGibsonHouse(
       scene.categories_.push_back(category);
       object->category_ = std::move(category);
     }
-    // TODO(msb) add support for aabb
+
+    const auto& jsonCenter = jsonObject["location"];
+    if (!jsonCenter.IsNull()) {
```
msbaines: Code might be simpler if you checked center AND size at the top.

src/esp/gfx/Simulator.cpp
```diff
@@ -158,7 +161,9 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
         if (endsWith(houseFilename, ".house")) {
           scene::SemanticScene::loadMp3dHouse(houseFilename, *semanticScene_);
         } else if (endsWith(houseFilename, ".scn")) {
-          scene::SemanticScene::loadGibsonHouse(houseFilename, *semanticScene_);
+          scene::SemanticScene::loadGibsonHouse(
+              houseFilename, *semanticScene_,
+              quatf::FromTwoVectors(geo::ESP_GRAVITY, -vec3f::UnitZ()));
```
msbaines: Add a comment about this.

src/esp/io/json.h
```diff
@@ -7,6 +7,7 @@
 #include <cstdint>
 #define RAPIDJSON_NO_INT64DEFINE
 #include <rapidjson/document.h>
+#include "esp/core/esp.h"
```
msbaines: Why is this include needed?

src/esp/gfx/Simulator.cpp
```diff
@@ -53,6 +53,9 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
   }
 
   std::string houseFilename = io::changeExtension(sceneFilename, ".house");
+  if (!io::exists(houseFilename)) {
```
msbaines: Is this still needed? You can avoid this by adding a "house" entry to your configuration in cfg.scene.filepaths

mathfac: @msbaines by default, we are not filling a "house" entry in cfg.scene.filepaths. To support same behavior for `".house" (line 55) and ".scn" that make sense to leave it. Otherwise, wider refactoring needed and there will be no that seamless experience loading MP3D with semantic scenes.

src/esp/scene/GibsonSemanticScene.cpp
mathfac: I want to create OBB only if at least center is available, while size can be optional.

src/esp/io/json.h
```diff
@@ -26,6 +27,9 @@ JsonDocument parseJsonString(const std::string& jsonString);
 //! Return string representation of given JsonDocument
 std::string jsonToString(const JsonDocument& d);
 
+//! Return Vec3f coordinates representation of given JsonObject of array type
+esp::vec3f jsonToVec3f(const JsonGenericValue& jsonArray);
```
msbaines: Looks like you need to include esp/core/esp.h for esp::vec3f

src/esp/gfx/Simulator.cpp
```diff
@@ -158,7 +161,10 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
         if (endsWith(houseFilename, ".house")) {
           scene::SemanticScene::loadMp3dHouse(houseFilename, *semanticScene_);
         } else if (endsWith(houseFilename, ".scn")) {
-          scene::SemanticScene::loadGibsonHouse(houseFilename, *semanticScene_);
+          scene::SemanticScene::loadGibsonHouse(
+              houseFilename, *semanticScene_,
+              // Default World rotation for Gibson semantic data
+              quatf::FromTwoVectors(geo::ESP_GRAVITY, -vec3f::UnitZ()));
```
msbaines: mp3d does this as a default argument in the header file. Not sure which is better but we should try to be consistent

mathfac: @msbaines, that's great suggestion, thank you!

PR #409, opened 2019-12-28
[WIP][redesign] Initial class design for sensor api

apsdehal:
## Motivation and Context

As per the discussion I had with Mandeep, we need to first built a public exposing Sensor API for ease of using the sensors. This is first WIP step towards that.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/sensor/Base.h
+++ b/src/esp/sensor/Base.h
@@ -4,26 +4,27 @@
 
 #pragma once
 
-#include "esp/scene/SceneNode.h"
 #include "esp/io/json.h"
+#include "esp/scene/SceneNode.h"
 
 #include <string>
 
 namespace esp {
 namespace sensor {
 class BaseSensor : public Magnum::SceneGraph::AbstractFeature3D {
-  public:
-    explicit BaseSensor(scene::SceneNode& node, const esp::io::JsonDocument spec);
-    std::string getType();
-    std::string getUUID();
-    std::vector<char> getObservation();
-    std::vector<int> getObservationShape();
-    std::string getObservationDataType();
+ public:
+  explicit BaseSensor(scene::SceneNode& node, const esp::io::JsonDocument spec);
+  std::string getType();
+  std::string getUUID();
+  std::vector<char> getObservation();
+  std::vector<int> getObservationShape();
+  std::string getObservationDataType();
 };
 
 class SensorFactory {
-  public:
-    static BaseSensor create(scene::SceneNode& node, const esp::io::JsonDocument spec);
-}
-} // namespace sensor
-} // namespace esp
+ public:
+  static BaseSensor create(scene::SceneNode& node,
+                           const esp::io::JsonDocument spec);
+};
+}  // namespace sensor
+}  // namespace esp
--- a/src/esp/sensor/Base.h
+++ b/src/esp/sensor/Base.h
@@ -1,0 +1,29 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include "esp/scene/SceneNode.h"
+#include "esp/io/json.h"
+
+#include <string>
+
+namespace esp {
+namespace sensor {
+class BaseSensor : public Magnum::SceneGraph::AbstractFeature3D {
+  public:
+    explicit BaseSensor(scene::SceneNode& node, const esp::io::JsonDocument spec);
+    std::string getType();
+    std::string getUUID();
+    std::vector<char> getObservation();
+    std::vector<int> getObservationShape();
+    std::string getObservationDataType();
+};
+
+class SensorFactory {
+  public:
+    static BaseSensor create(scene::SceneNode& node, const esp::io::JsonDocument spec);
+}
+} // namespace sensor
+} // namespace esp
--- a/src/esp/sensor/Base.h
+++ b/src/esp/sensor/Base.h
@@ -11,20 +11,96 @@
 
 namespace esp {
 namespace sensor {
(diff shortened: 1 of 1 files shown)


src/esp/sensor/Base.h
```diff
@@ -0,0 +1,30 @@
...
+  std::string getType();
+  std::string getUUID();
+  std::vector<char> getObservation();
+  std::vector<int> getObservationShape();
+  std::string getObservationDataType();
+};
+
+class SensorFactory {
+ public:
+  static BaseSensor create(scene::SceneNode& node,
```
erikwijmans: This will need to return a pointer or a reference for dynamic casting to work.

src/esp/sensor/Base.h
```diff
@@ -0,0 +1,30 @@
...
+#include "esp/io/json.h"
+#include "esp/scene/SceneNode.h"
+
+#include <string>
+
+namespace esp {
+namespace sensor {
+class BaseSensor : public Magnum::SceneGraph::AbstractFeature3D {
+ public:
+  explicit BaseSensor(scene::SceneNode& node, const esp::io::JsonDocument spec);
```
erikwijmans: I am not sure if a `esp::io::JsonDocument` will work long term.  I think having a BaseSensorSpec class and then sensors implement their own would be better.  Documenting the json document will be a nightmare and you don't get the inherent type safety/compile time errors when using a struct/class.

apsdehal: I second you on this. Since the PR is WIP, I was going to iteratively refine this. The current plan is that the SensorFactory takes in a JSONDocument creates a specific SensorSpec object based on the type of sensor that is getting created. The final Sensor classes will only interact with SensorSpec objects.

apsdehal: Thank you for all of the comments. :)

Skylion007:
One use case the sensor redesign does not address is that it's currently impossible to access the object categories within the SemanticSensor or related sensors itself. Since currently the renderer only draws object ids, this means one has to iterate through object ids to access them and impossible to do that within the getObservation method. This makes extending or creating new types of sensors difficult. (For instance, if I want a sensor that only returns the object class id instead of the object instance id or a sensor which swaps the object instance/object class ids).

Being able to access the object categories or other scene/sim properties from within sensor types that may need them in the future would be a useful feature to consider when considering the redesign. It may also help create sensors that are stateful for when they are needed in the future.

src/esp/sensor/Base.h
```diff
@@ -0,0 +1,30 @@
...
+  std::string getType();
+  std::string getUUID();
+  std::vector<char> getObservation();
```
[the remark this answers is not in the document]
msavva: Agree with @erikwijmans but I think this might even be better wrapped into an `Observation` type which we can then endow with additional information about the observation down the line.  There's a stub for this that hasn't been hooked up yet at https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/sensor/Sensor.h#L67-L71 .  Perhaps we should adjust that to contain `std::vector<char>` and use it here?

src/esp/sensor/Base.h
msbaines: @erikwijmans I think for python and javascript it will be easier for users to deal with a json sensor spec than having to allocate an object and deal with a class hierarchy.

src/esp/sensor/Base.h
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * @brief Assign current observation of the sensor to the
+   *        passed argument
+   * @param[in, out] obs Pointer to the variable to which current observation
+   *                     will be assigned
+   *
+   * NOTE: Override in child sensor class
+   */
+  virtual bool getObservation(void* obs);
```
erikwijmans: This should be marked a pure virtual if a default implementation won't be provided.

src/esp/sensor/Base.h
```diff
@@ -0,0 +1,113 @@
...
+   */
+  virtual const std::string& getType() const = 0;
+
+  /**
+   * @brief Universally Unique Identifier for the current instance of the
+   *        sensor. Appends a unique string to UID defined by the
+   *        sensor specification.
+   * @return string UUID for the current instance
+   */
+  const std::string& getUUID() const;
```
erikwijmans: Same here.

src/esp/sensor/Base.h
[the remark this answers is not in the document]
msbaines: Agree with @msavva. Returning an Observation class instance makes sense. One piece of data I think of that might be useful is timestamp. An Observation class would also be useful for handling GPU copy.

src/esp/sensor/Base.h
apsdehal: This will be implemented in the base sensor implementation.

src/esp/sensor/Base.h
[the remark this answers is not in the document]
apsdehal: Thanks. Let's target that in a separate incremental PR. I will add a TODO here.

src/esp/sensor/Base.h
```diff
@@ -0,0 +1,113 @@
...
+   * NOTE: Override in child class
+   */
+  virtual const std::string& getObservationDataType() const = 0;
+
+ protected:
+  ESP_SMART_POINTERS(BaseSensor);
+};
+
+// Factory for creating sensors
+class SensorFactory {
```
bigbike: One file for one class. It is easier for the developers to find the class by filenames.

bigbike:
@apsdehal : The current shape of the code looks great!
It is a high impact project. Hope we can continue it when you are available. How do you think?

Also add @mosra  into the reviewer list.

eundersander:
Closing this stale PR.

PR #413, opened 2020-01-03
Add getObjectSceneNode function to PhysicsManager

aclegg3:
##  Motivation and Context

It is often desirable to fetch the SceneNode associated with a physics object. However, we provide many PhysicsManager functions to avoid direct manipulation of SceneNode states which may invalidate simulation state. Here we add a const reference getter for a query only object SceneNode.

## How Has This Been Tested

Some testing lines in existing physics C++ test function. Will likely be moved to another test function once this functionality is needed (upcoming).

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/PhysicsManager.cpp
+++ b/src/esp/physics/PhysicsManager.cpp
@@ -448,5 +448,11 @@ void PhysicsManager::setObjectBBDraw(int physObjectID,
         0, *existingObjects_[physObjectID]->BBNode_, drawables);
   }
 }
+
+const scene::SceneNode& PhysicsManager::getObjectSceneNode(int physObjectID) {
+  assertIDValidity(physObjectID);
+  return *existingObjects_[physObjectID];
+}
+
 }  // namespace physics
 }  // namespace esp
--- a/src/tests/PhysicsTest.cpp
+++ b/src/tests/PhysicsTest.cpp
@@ -82,6 +82,11 @@ TEST(PhysicsTest, JoinCompound) {
       for (int o = 0; o < num_objects; o++) {
         int objectId = physicsManager_->addObject(objectFile, nullptr);
         objectIds.push_back(o);
+
+        const esp::scene::SceneNode& node =
+            physicsManager_->getObjectSceneNode(objectId);
+        node.getId();
+
         Magnum::Matrix4 R{
             Magnum::Matrix4::rotationX(Magnum::Math::Rad<float>(-1.56)) *
             Magnum::Matrix4::rotationY(Magnum::Math::Rad<float>(-0.25))};
--- a/src/tests/PhysicsTest.cpp
+++ b/src/tests/PhysicsTest.cpp
@@ -85,7 +85,6 @@ TEST(PhysicsTest, JoinCompound) {
 
         const esp::scene::SceneNode& node =
             physicsManager_->getObjectSceneNode(objectId);
-        node.getId();
 
         Magnum::Matrix4 R{
             Magnum::Matrix4::rotationX(Magnum::Math::Rad<float>(-1.56)) *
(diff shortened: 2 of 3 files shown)


src/tests/PhysicsTest.cpp
```diff
@@ -82,6 +82,11 @@ TEST(PhysicsTest, JoinCompound) {
       for (int o = 0; o < num_objects; o++) {
         int objectId = physicsManager_->addObject(objectFile, nullptr);
         objectIds.push_back(o);
+
+        const esp::scene::SceneNode& node =
+            physicsManager_->getObjectSceneNode(objectId);
+        node.getId();
```
mosra: Maybe check that `node.getId() == objectId`? Because this only papers over code coverage, but doesn't test anything.

aclegg3: In this case, the value returned by `node.getId()` has nothing to do with the `objectId` besides an unforteunate variable name collision. `node.getId()` is the semantic object class while 'objectId' is an index in the `PhysicsManager`. 

I could check the node transform against `physicsManager_->getTransform()` to be more precise, however.

msavva: @aclegg3 (a thought triggered by @mosra 's comment, not feedback for this PR): since all physics-enabled objects managed by the `PhysicsManager` are guaranteed to also be `SceneNode`s, would it make sense to have our addressing scheme be consistent between the two? (i.e. we could assign object ids to physics-enabled `SceneNode`s and use those object ids to keep track of them in `PhysicsManager`).  It seems to me that would be much cleaner, ensure a single object id "source of truth" (regardless of being physics-enabled or not), and prevent potential pitfalls and misunderstandings as above.

bigbike: Just read Manolis's comment, and understand his point. I simply have some concerns from other aspects (not related to this PR), however.

Currently, the `rigidObject` is inherited from scene node. My concern is that how far we should go in this direction.

As we discussed before, ideally `rigidObject` should be subclassed from the "abstractFeature", like the existing render camera, sensors, lights etc. The advantage is obvious, e.g., smart pointers can be used for `rigidObject` (we cannot now), `rigidObject` can be "created and forgot" (handled automatically by magnum later).

So not so sure if we need to give ids to SceneNode representing physics-enabled objects.

aclegg3: Thanks @bigbike. I agree this is still a TODO, but I don't know that it changes the guaranteed relationship between a rigid object and a SceneNode or the need to id them.

Note: moved this conversation to https://github.com/facebookresearch/habitat-sim/pull/341#issuecomment-570728695 😄

msavva: Was about to make exact same comment as @aclegg3 🙂In any case, let's continue the conversation in the new thread.

bigbike: See my comments in #341 .

src/esp/physics/PhysicsManager.cpp
```diff
@@ -448,5 +448,11 @@ void PhysicsManager::setObjectBBDraw(int physObjectID,
         0, *existingObjects_[physObjectID]->BBNode_, drawables);
   }
 }
+
+const scene::SceneNode& PhysicsManager::getObjectSceneNode(int physObjectID) {
+  assertIDValidity(physObjectID);
```
bigbike: I will recommend:
-) CORRADE_ASSERT instead of CHECK within assertIDValidity;
-) std::unordered_map instead of std::map, which is faster, for existingObjects_;

But certainly none of this needs to be changed in this PR. Just FYI.

aclegg3: Thanks for the suggestions. Should do a refactor pass over some of that in the near future.

Issue #391, opened 2019-12-18
Make RenderCamera inherited from MagnumCamera


bigbike:
## 🚀 Feature
For historical reasons, the RenderCamera contains a MagnumCamera as the member variable. This request is to refactor the code so that it inherits from the MagnumCamera. 

## Motivation
- It immediately eliminates the getMagnumCamera(...) in the interface API;
- It gains all the existing APIs in MagnumCamera;
- The development could be easier in Frustum Culling project. Culling can happen in the RenderCamera class;

PR #422, opened 2020-01-10
Refactor RenderCamera to inherit from MagnumCamera

matthewjmay:
## Motivation and Context

Addresses #391 
- Eliminates the getMagnumCamera(...) in the interface API;
- gains all the existing APIs in MagnumCamera;
- The development could be easier in Frustum Culling project. Culling can happen in the RenderCamera class;

## How Has This Been Tested

- Manual testing using viewer app
- Manual testing ensuring `MagnumCamera` api accessible from python bindings
- habitat-sim and habitat-api have been grepped for usages of `RenderCamera`
- passes all tests

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/GfxBindings.cpp
+++ b/src/esp/bindings/GfxBindings.cpp
@@ -36,7 +36,7 @@ namespace gfx {
 void initGfxBindings(py::module& m) {
   // ==== RenderCamera ====
   py::class_<RenderCamera, Magnum::SceneGraph::PyFeature<RenderCamera>,
-             Magnum::SceneGraph::AbstractFeature3D,
+             Magnum::SceneGraph::Camera3D,
              Magnum::SceneGraph::PyFeatureHolder<RenderCamera>>(
       m, "Camera",
       R"(RenderCamera: The object of this class is a camera attached
@@ -47,12 +47,6 @@ void initGfxBindings(py::module& m) {
         Set this `Camera`'s projection matrix.
       )",
            "width"_a, "height"_a, "znear"_a, "zfar"_a, "hfov"_a)
-      .def("getProjectionMatrix", &RenderCamera::getProjectionMatrix, R"(
-        Get this `Camera`'s projection matrix.
-      )")
-      .def("getCameraMatrix", &RenderCamera::getCameraMatrix, R"(
-        Get this `Camera`'s camera matrix.
-      )")
       .def_property_readonly("node", nodeGetter<RenderCamera>,
                              "Node this object is attached to")
       .def_property_readonly("object", nodeGetter<RenderCamera>,
--- a/src/esp/gfx/RenderCamera.cpp
+++ b/src/esp/gfx/RenderCamera.cpp
@@ -11,12 +11,9 @@ namespace Mn = Magnum;
 namespace esp {
 namespace gfx {
 
-RenderCamera::RenderCamera(scene::SceneNode& node)
-    : Mn::SceneGraph::AbstractFeature3D{node} {
+RenderCamera::RenderCamera(scene::SceneNode& node) : MagnumCamera{node} {
   node.setType(scene::SceneNodeType::CAMERA);
-  camera_ = new MagnumCamera(node);
-  camera_->setAspectRatioPolicy(
-      Mn::SceneGraph::AspectRatioPolicy::NotPreserved);
+  setAspectRatioPolicy(Mn::SceneGraph::AspectRatioPolicy::NotPreserved);
 }
 
 RenderCamera::RenderCamera(scene::SceneNode& node,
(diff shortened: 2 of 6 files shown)


src/esp/gfx/RenderCamera.h
```diff
@@ -29,12 +29,10 @@ class RenderCamera : public Magnum::SceneGraph::AbstractFeature3D {
...
-        Magnum::SceneGraph::AbstractFeature3D::object());
+    return static_cast<scene::SceneNode&>(MagnumCamera::object());
   }
   const scene::SceneNode& object() const {
-    return static_cast<const scene::SceneNode&>(
-        Magnum::SceneGraph::AbstractFeature3D::object());
+    return static_cast<const scene::SceneNode&>(MagnumCamera::object());
   }
 
   void setProjectionMatrix(int width,
```
matthewjmay: We may want to consider returning reference to self (`RenderCamera&`) in both `setProjectionMatrix` overloads to stay consistent with `MagnumCamera::setProjectionMatrix`. Let me know your thoughts.

bigbike: Yes, correct. Let us do that, which is for method chaining so that you can call like object.methodA().methodB().methodC();

src/esp/bindings/GfxBindings.cpp
```diff
@@ -36,22 +36,25 @@ namespace gfx {
...
+           R"(
+        Set this `Camera`'s projection matrix via view configuration.
       )",
            "width"_a, "height"_a, "znear"_a, "zfar"_a, "hfov"_a)
-      .def("getProjectionMatrix", &RenderCamera::getProjectionMatrix, R"(
-        Get this `Camera`'s projection matrix.
-      )")
-      .def("getCameraMatrix", &RenderCamera::getCameraMatrix, R"(
-        Get this `Camera`'s camera matrix.
+      .def("setProjectionMatrix",
```
bigbike: @mosra @erikwijmans :

Now the RenderCamera is inherited from MagnumCamera, which I believe magnum has the py bindings for this signature (setProjectionMatrix), which makes it unnecessary here.

Is that correct?

src/esp/gfx/RenderCamera.h
```diff
@@ -43,17 +41,10 @@ class RenderCamera : public Magnum::SceneGraph::AbstractFeature3D {
...
 
-  mat4f getProjectionMatrix();
-
-  mat4f getCameraMatrix();
-
-  MagnumCamera& getMagnumCamera();
-
-  void draw(MagnumDrawableGroup& drawables);
+  // expose overload for directly setting projection matrix
+  using MagnumCamera::setProjectionMatrix;
```
bigbike: I understand that "if a derived class redefines base class member method (e.g., setProjectionMatrix here) then all the base class methods with same name become hidden in derived class".

However, we can still call that method in base class using: `derivedObject.::BaseClassName.projectionMatrix(...)`
So I do not think we need it here.

src/esp/bindings/GfxBindings.cpp
erikwijmans: Looks like there is just a property for the projection matrix itself: https://doc.magnum.graphics/python/magnum/scenegraph/Camera3D/#properties

So the get is redundant, but the setters aren't.

bigbike: @erikwijmans : We have two "setProjectionMatrix" here. The 1st one (line 46) definitely needs to be kept. I am talking about the 2nd one, which is just the original "setProjectionMatrix" defined in the base class.

I am asking if we need to keep this binding. My opinion is no, we do not need it since it is already done in Magnum python bindings.

erikwijmans: Ah, didn't look closely at the second one, yes, that is not needed.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -35,27 +32,11 @@ void RenderCamera::setProjectionMatrix(int width,
...
-
-mat4f RenderCamera::getCameraMatrix() {
-  return Eigen::Map<mat4f>(camera_->cameraMatrix().data());
-}
-
-MagnumCamera& RenderCamera::getMagnumCamera() {
-  return *camera_;
-}
-
-void RenderCamera::draw(MagnumDrawableGroup& drawables) {
```
bigbike: Oh, I need this draw(...) function in derived class so that I can do culling within.

Can we keep it? If you keep it, add a comment, e.g.,
// TODO: visibility culling

```CPP
void RenderCamera::draw(MagnumDrawableGroup& drawables) {
    this->MagnumCamera::draw(drawables);
}
```

src/tests/SimTest.cpp
```diff
@@ -53,3 +79,52 @@ TEST(SimTest, Reset) {
...
+      {static_cast<size_t>(pinholeCameraSpec->resolution[0]),
+       static_cast<size_t>(pinholeCameraSpec->resolution[1]), 4}};
+
+  ASSERT_EQ(obsSpace.spaceType, ObservationSpaceType::TENSOR);
+  ASSERT_EQ(obsSpace.dataType, esp::core::DataType::DT_UINT8);
+  ASSERT_EQ(obsSpace.shape, expectedShape);
+  ASSERT_EQ(observation.buffer->shape, expectedShape);
+
+  // Compare one pixel (center of image) with previously rendered ground truth
+  // TODO: compare more than one pixel
```
mosra: FYI (I guess you know already), if you ever get to this TODO, Corrade's TestSuite allows you to do this easily: https://doc.magnum.graphics/magnum/classMagnum_1_1DebugTools_1_1CompareImage.html

It would mean you'd need to port this away from googletest, but looking at the other parts of this test, it seems it would benefit from that :)

src/tests/SimTest.cpp
```diff
@@ -24,6 +35,21 @@ const std::string skokloster =
...
+  return std::abs(rhs - lhs) <= std::abs(tolerance);
+}
+
+bool pixelEqualWithChannelTolerance(const Mn::Color4ub& lhs,
+                                    const Mn::Color4ub& rhs,
+                                    int tolerance) {
+  return equalWithTolerance(lhs.r(), rhs.r(), tolerance) &&
+         equalWithTolerance(lhs.g(), rhs.g(), tolerance) &&
+         equalWithTolerance(lhs.b(), rhs.b(), tolerance) &&
+         equalWithTolerance(lhs.a(), rhs.a(), tolerance);
```
mosra: As I mentioned in my other comment, this is another thing Corrade's TestSuite has builtin -- basically the following, plus you get a clear message if the comparison fails (as opposed to "false is not true"):

```cpp
#include <Corrade/TestSuite/Tester/Compare/Numeric.h>

CORRADE_COMPARE_WITH(lhs, rhs, TestSuite::Compare::around(tolerance));
```

Though this is also related to the image comparison, and there the fuzzy compare is handled in an even more convenient way :)

PR #423, opened 2020-01-10
Bullet collision from bounding box

aclegg3:
## Motivation and Context

Primitive box shapes are much more efficient and stable than meshes for collision. 

This PR enables the use of an object's accumulated mesh bounding box to serve as its collision shape. It also allows this option to be configured in an object's physics config file. This option will override other options such as choice of collision mesh and mesh joining.

TODO in future PR: 
 - Add an option to build a compound of component bounding boxes for approximate concavity.

## How Has This Been Tested
C++ CI test.

Local testing in viewer:
sphere_mesh_vs_bb

## Types of changes

## Checklist


src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -166,20 +166,41 @@ bool BulletRigidObject::initializeObject(
...
   //! The components are combined into a convex compound shape
   bObjectShape_ = std::make_unique<btCompoundShape>();
-  // TODO: this should translate to the COM
-  constructBulletCompoundFromMeshes(Magnum::Matrix4{}, meshGroup, metaData.root,
-                                    joinCollisionMeshes);
 
-  // add the final object after joining meshes
-  if (joinCollisionMeshes) {
+  if (collisionFromBB_) {
+    // set a dummy shape for now. We will use the scene node BB object to
```
msbaines: Why do we need to set a dummy object?

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -220,6 +243,30 @@ bool BulletRigidObject::initializeObject(
...
+    // allow bullet to compute the inertia tensor if we don't have one
+    bObjectShape_->calculateLocalInertia(getMass(),
+                                         bInertia);  // overrides bInertia
+
+    LOG(INFO) << "Automatic BB object inertia computed: " << bInertia.x() << " "
+              << bInertia.y() << " " << bInertia.z();
+
+    setInertiaVector(Magnum::Vector3(bInertia));
+  }
+}
```
msbaines: The logic above is now duplicated. Could it be put into its own function?

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -166,20 +166,41 @@ bool BulletRigidObject::initializeObject(
...
+      t.setIdentity();
+      bObjectConvexShapes_.back()->setMargin(0.0);
+      bObjectConvexShapes_.back()->recalcLocalAabb();
+      bObjectShape_->addChildShape(t, bObjectConvexShapes_.back().get());
+    }
+    bObjectShape_->setMargin(margin);
+    bObjectShape_->recalculateLocalAabb();
+    btVector3 localAabbMin, localAabbMax;
+    bObjectShape_->getAabb(btTransform::getIdentity(), localAabbMin,
+                           localAabbMax);
```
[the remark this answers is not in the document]
aclegg3: I think that was leftover from testing. Good catch. 👍

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: Old design leftover. Changed.

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: These two instances are different enough that this wouldn't give us much. Since one is before initialization of the bObjectRigidBody_ and the other is after, most of the get/set code would still need to stay for each.

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -220,6 +243,30 @@ bool BulletRigidObject::initializeObject(
   return true;
 }
 
+void BulletRigidObject::setCollisionFromBB() {
+  btVector3 dim(cumulativeBB_.size() / 2.0);
+  bObjectShape_->removeChildShape(bGenericShapes_.back().get());
+  bGenericShapes_.clear();
+  bGenericShapes_.emplace_back(std::make_unique<btBoxShape>(dim));
```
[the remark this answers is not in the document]
aclegg3: We need to store the collision shapes since Bullet does not replicate them internally, this is also consistent with how we handle other collision shapes.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -97,6 +97,23 @@ int BulletPhysicsManager::makeRigidObject(
...
+int BulletPhysicsManager::addObject(const int objectLibIndex,
+                                    DrawableGroup* drawables) {
+  // Do default load first (adds the SceneNode to the SceneGraph and computes
+  // the cumulativeBB_)
+  int objID = PhysicsManager::addObject(objectLibIndex, drawables);
+
+  // Then set the collision shape to the cumulativeBB_ if necessary
+  if (objID != ID_UNDEFINED) {
+    BulletRigidObject* bro =
+        static_cast<BulletRigidObject*>(existingObjects_.at(objID));
```
erikwijmans: Do we ever expect to have multiple types of rigid objects in `existingObjects_`?  (i.e. some objects are bullet objects, others are simple kinematic ones?).  If so, a dynamic cast would be safer here.

src/esp/physics/bullet/BulletRigidObject.cpp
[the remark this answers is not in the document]
erikwijmans: Why is `bGenericShapes_` a vector?  I can't seem to find anywhere else that adds something to it, i.e. `bGenericShapes_.size() == 1` is always true.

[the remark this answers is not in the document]
aclegg3: Next step is to enable component level bounding box hierarchies for approximate concavities. This will be used soon.

src/esp/physics/bullet/BulletPhysicsManager.cpp
aclegg3: Current design is to set the MotionType of the object to specify kinematic, dynamic or static. To enable collisions between all types, they must still be BulletRigidObject type, so I think this is fine. If the user has enabled the BulletPhysicsManager then all objects in this list are BulletRigidObject. Will likely setup a different PhysicsManager type for dealing with multi-bodies.

PR #424, opened 2020-01-10
Bug fix: transformPoint used on vectors in viewer

aclegg3:
## Motivation and Context

Viewer functions for applying linear force and impulse to objects transform agent forward direction into world push direction. This was improperly done with `transformPoint` instead of `transformVector`.

## How Has This Been Tested

Local testing.

## Types of changes

## Checklist


aclegg3:
> Can the local testing be automated? That would provide context for the change and prevent this functionality from regressing in the future.

This seems like more of an application side semantic error than anything else. I tested by pushing an object from several orientations and noted that it moved in the view direction. Do we have any testing of the viewer demo application? Not sure that we should prioritize this if the underlying infra is tested. We were considering a re-design of this demo anyway.

PR #425, opened 2020-01-10
Bullet discrete contact testing

aclegg3:
## Motivation and Context

Discrete contact testing is useful for a variety of tasks including agent/object placement and collision filter functions. This PR introduces the first step: binary contact testing for a specific object with the collision world. 

Includes Simulator.cpp wrapper and python bindings for simulator.py.

This infra can later be extended to include a configurable contact distance threshold and collection of contact points, normals, friction directions, contact distances, and colliding object pair identifiers.

## How Has This Been Tested

C++ CI test for object/object and object/scene contact results.

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -285,6 +285,9 @@ class Simulator:
     def apply_torque(self, torque, object_id, scene_id=0):
         self._sim.apply_torque(torque, object_id, scene_id)
 
+    def contact_test(self, object_id, scene_id=0):
+        return self._sim.contact_test(object_id, scene_id)
+
     def get_world_time(self, scene_id=0):
         return self._sim.get_world_time()
 
--- a/src/esp/gfx/Simulator.cpp
+++ b/src/esp/gfx/Simulator.cpp
@@ -346,6 +346,13 @@ Magnum::Quaternion Simulator::getRotation(const int objectID,
   return Magnum::Quaternion();
 }
 
+bool Simulator::contactTest(const int objectID, const int sceneID) {
+  if (physicsManager_ != nullptr && sceneID >= 0 && sceneID < sceneID_.size()) {
+    return physicsManager_->contactTest(objectID);
+  }
+  return false;
+}
+
 double Simulator::stepWorld(const double dt) {
   if (physicsManager_ != nullptr) {
     physicsManager_->stepPhysics(dt);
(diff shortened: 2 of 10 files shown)


src/esp/gfx/Simulator.cpp
```diff
@@ -346,6 +346,13 @@ Magnum::Quaternion Simulator::getRotation(const int objectID,
   return Magnum::Quaternion();
 }
 
+bool Simulator::contactTest(const int objectID, const int sceneID) {
```
msbaines: Should we expose the physics manager instead of forwarding? It would simplify Simulator and avoid having to add forwarding code every time a new API is added.

habitat_sim/simulator.py
```diff
@@ -285,6 +285,9 @@ def apply_force(self, force, relative_position, object_id, scene_id=0):
     def apply_torque(self, torque, object_id, scene_id=0):
         self._sim.apply_torque(torque, object_id, scene_id)
 
+    def contact_test(self, object_id, scene_id=0):
```
bigbike: It it is an internal test, I am wonderingwhy we need to expose it on the python side?

I took a look at the other bindings above it, e.g., apply_torque, get_rotation, add_object. They are reasonable to be exposed here. But this one is different. I would like to hear from you first.

aclegg3: Good question. This function name may be misleading. It is not a unit test, but a discrete collision check. I'm happy to rename it if that is a problem. 

There are actually a couple of applications for this functionality on the python side. 

Examples include:
 - 3D step filter function (check if agent collides with env after moving)
 - Scene generation and 3D Agent placement (rejection sampling for contact free placement)

src/esp/gfx/Simulator.cpp
aclegg3: We've talked about this in the past. At that time the motivation was:
1) make C++ the primary API (implement functions in Simulator class in C++)
2) make physics functions callable from sim.function() instead of needing sim.physicsManager.function() for ease of use.

msbaines: For 2. you could grab a physicsManager reference and just call physicsManager.function(). But sounds reasonable to have everything go through Simulator.

Non-blocking: it would be nice to not have all the forwarding logic in Simulator.cpp. Maybe via multiple inheritance or an include.

PR #428, opened 2020-01-14
Imported objects configurable scaling

aclegg3:
## Motivation and Context

This change allows non-uniform scaling of objects via configuration of an object's template in `ResourceManager::physicsObjectLibrary_`.

This will allow users to load a single object template and spawn multiple instances with varying scales into the scene.

Note: negative scales are allowed, tested, and stable. Visually, this can invert the mesh normals, allowing the camera to see through the surface of the object due to back-face culling. I didn't restrict this since it could be considered a feature (e.g. invert the scale to see inside a shape from the outside).

## How Has This Been Tested
New C++ CI test.
Local testing in viewer:
scaling_boxes_test

## Types of changes

## Checklist

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,9 @@
 {
-  "extends": ["eslint:recommended", "plugin:prettier/recommended"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:prettier/recommended",
+    "prettier/@typescript-eslint"
+  ],
   "plugins": [
     "html"
   ],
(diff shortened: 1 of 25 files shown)


dhruvbatra:
Will be useful for domain randomization. 

Also, it's raining cheeseboxes :-).

aclegg3:
> The physics config parsing code is getting large enough that it may be time to pull it out of ResourceManager.cpp. We may also want to consider writing a code generator since the code is all following the same template.

Great point, I was thinking the same thing while writing this code. 😄

Issue #300, opened 2019-10-15
Make sim.reset reset agent and simulation to after reconfigure state


mathfac:
## 🚀 Feature

Make sim.reset reset an agent position and Simulator to `after reconfigure state`.

## Motivation
Now [sim.reset()](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/Simulator.cpp#L165) does nothing, while user expect it to call reconfigure with original config.

Additionally, remove [the comment ](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/Simulator.cpp#L172) as sampling agent position isn't expected to happen in C++.

## Alternatives
Similar as it is [done](https://github.com/facebookresearch/habitat-api/blob/master/habitat/sims/habitat_simulator/habitat_simulator.py#L420) in `habitat-api` now we can manually set the original agent state and get observations:
```
        for agent_id, _ in enumerate(self.config.AGENTS):
            agent_cfg = self._get_agent_config(agent_id)
            if agent_cfg.IS_SET_START_STATE:
                self.set_agent_state(
                    agent_cfg.START_POSITION,
                    agent_cfg.START_ROTATION,
                    agent_id,
                )

        return self._sim.get_sensor_observations()


```

## Additional context
Issue was surfaced by @piotr-bojanowski.

erikwijmans:
The agent config doesn't contain any initial position information  for habitat-sim: https://aihabitat.org/docs/habitat-sim/habitat_sim.agent.AgentConfiguration.html  (which is why this isn't done).

mathfac:
That makes sense for C++ part of sim, but Python part of Habitat-SIM has empty reset function while it should call [reconfigure](https://github.com/facebookresearch/habitat-sim/blob/823111e431df78bdb8c1224816912eb5aa74a08a/habitat_sim/simulator.py#L88) with original config.

erikwijmans:
A `reconfigure` also doesn't set the agent back to the initial state as the configuration doesn't contain that information (and unless the config changed, a reconfigure is a no-op).  We currently don't have a notion of an initial agent state and the agent is simply given a random state upon creation.

mathfac:
@erikwijmans, looks like `reconfigure ` is setting agent position [here](https://github.com/facebookresearch/habitat-sim/blob/823111e431df78bdb8c1224816912eb5aa74a08a/habitat_sim/simulator.py#L167).

mathfac:
[...]

PR #429, opened 2020-01-14
made reset function in Simulator class reset agent state to initial s…

mpiseno:
…tate

## Motivation and Context

Made the reset function in the Simulator class of simulator.py reset the agent to its initial state. Issue from here: https://github.com/facebookresearch/habitat-sim/issues/300

## How Has This Been Tested

Tested this by instantiating the simulator class and stepping through a few actions before calling reset, then comparing the initial and final state to ensure they were the same.

## Types of changes

Created a field in the Agent class called initial_state that gets set when initialize_agent() is called in simulator.py. This makes sure that the initial state is set with a valid position in the scene (not inside a wall or some weird place like that). Then when reset is called, each agent's state is set to its initial state.

### DIFF
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -123,6 +123,7 @@ class Agent(object):
         self.body = mn.scenegraph.AbstractFeature3D(scene_node)
         scene_node.type = hsim.SceneNodeType.AGENT
         self.reconfigure(self.agent_config)
+        self.initial_state = None
 
     def reconfigure(
         self, agent_config: AgentConfiguration, reconfigure_sensors: bool = True
@@ -189,13 +190,15 @@ class Agent(object):
 
         return state
 
-    def set_state(self, state: AgentState, reset_sensors: bool = True):
+    def set_state(self, state: AgentState, reset_sensors: bool = True, is_initial: bool = False):
         r"""Sets the agents state
 
         :param state: The state to set the agent to
         :param reset_sensors: Whether or not to reset the sensors to their
             default intrinsic/extrinsic parameters before setting their
             extrinsic state
+        :param is_initial: Whether this state is the initial state of the
+            agent in the scene. Used for resetting the agent at a later time
         """
         habitat_sim.errors.assert_obj_valid(self.body)
 
@@ -225,6 +228,9 @@ class Agent(object):
                 )
             )
             s.node.rotation = quat_to_magnum(state.rotation.inverse() * v.rotation)
+
+        if is_initial:
+            self.initial_state = state
 
     @property
     def scene_node(self):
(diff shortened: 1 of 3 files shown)


habitat_sim/simulator.py
```diff
@@ -87,6 +87,16 @@ def seed(self, new_seed):
 
     def reset(self):
         self._sim.reset()
+        for agent_id in range(len(self.agents)):
+            agent = self.get_agent(agent_id)
```
msbaines: Maybe this logic could go into an agent reset function?

habitat_sim/agent/agent.py
```diff
@@ -226,6 +229,9 @@ def set_state(self, state: AgentState, reset_sensors: bool = True):
             )
             s.node.rotation = quat_to_magnum(state.rotation.inverse() * v.rotation)
 
+        if is_initial:
+            self.initial_state = state
```
msbaines: Maybe we should set initial_state from agent.reconfigure?

msbaines:
You need to run the pre-commit in order to fix the python_lint error:

https://pre-commit.com/

msavva:
I haven't reviewed the code in depth, but I have a high-level comment: changing the default behavior of the reset call might not be a good idea since a lot of user code may already have been built up around the assumption that reset samples a new random initialization position.  I would propose adding an optional parameter to the reset call that allows for reset to the initial state as a way to do this without breaking backwards compatibility.

habitat_sim/agent/agent.py
erikwijmans: This makes sense to me.  I like the separation of agent configuration from agent state (that keeps the agent configuration from being specific to an environment).

habitat_sim/simulator.py
```diff
@@ -87,6 +87,16 @@ def seed(self, new_seed):
 
     def reset(self):
         self._sim.reset()
+        for agent_id in range(len(self.agents)):
```
mathfac: As mentioned in the issue https://github.com/facebookresearch/habitat-sim/issues/300 initial position of agents are not part of sim configuration right now. Therefore, best we can do just init agents position as it's done with `reconfigure` method:
```
        for i in range(len(self.agents)):
            self.initialize_agent(i)
```

mpiseno:
> I haven't reviewed the code in depth, but I have a high-level comment: changing the default behavior of the reset call might not be a good idea since a lot of user code may already have been built up around the assumption that reset samples a new random initialization position. I would propose adding an optional parameter to the reset call that allows for reset to the initial state as a way to do this without breaking backwards compatibility.

This make sense, but from my understanding reset was just a no-op before (not a resampling)

mpiseno:
Closing this pull request because I created a new one for the same issue that does not come from my fork, while this one does. The new pr is more up to date: https://github.com/facebookresearch/habitat-sim/pull/435

Issue #430, opened 2020-01-14
Errors when building navmesh on the fly for replica models. 


bigbike:
## 🐛 Bug
Built the utility viewer. Tried to load it with replica models.

It segfaults.

```
I0114 14:54:13.283030 169991 PTexMeshData.cpp:837] loading atlas textures: 
I0114 14:54:13.283100 169991 PTexMeshData.cpp:846] Loading atlas 1/6 from /home/yili/models/replica/room_1/textures/0-color-ptex.hdr. 
I0114 14:54:13.487493 169991 PTexMeshData.cpp:846] Loading atlas 2/6 from /home/yili/models/replica/room_1/textures/1-color-ptex.hdr. 
I0114 14:54:13.635692 169991 PTexMeshData.cpp:846] Loading atlas 3/6 from /home/yili/models/replica/room_1/textures/2-color-ptex.hdr. 
I0114 14:54:13.754685 169991 PTexMeshData.cpp:846] Loading atlas 4/6 from /home/yili/models/replica/room_1/textures/3-color-ptex.hdr. 
I0114 14:54:13.947890 169991 PTexMeshData.cpp:846] Loading atlas 5/6 from /home/yili/models/replica/room_1/textures/4-color-ptex.hdr. 
I0114 14:54:13.977864 169991 PTexMeshData.cpp:846] Loading atlas 6/6 from /home/yili/models/replica/room_1/textures/5-color-ptex.hdr. 
I0114 14:54:14.013339 169991 PathFinder.cpp:268] Building navmesh with -2147483648x-2147483648 cells
./run: line 3: 169991 Segmentation fault      (core dumped) __GL_SYNC_TO_VBLANK=0 ./build/utils/viewer/viewer /home/{user}/models/replica/room_1/mesh.ply
```

## Command
`# command that you run`

## To Reproduce

1.
2.
3.

## System Info

Please copy and paste the output from the
[environment collection script](https://github.com/facebookresearch/habitat-sim/tree/master/habitat_sim/utils/collect_env.py)
(or fill out the checklist below manually).

You can run the script with:
```
# For security purposes, please check the contents of collect_env.py before running it.
python habitat_sim/utils/collect_env.py
```

 - OS (e.g., Linux): Fedora 
 - How you installed PyTorch (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - GPU models and configuration:
 - CUDA version:
 - PIP/conda dependencies packages versions
[...]

PR #434, opened 2020-01-17
[bug fix][issue 430] load the navmesh for replica models

bigbike:
It will immediately fix the segfault when loading the replica models. #430 

For the long term, if we would like to compute the navmesh on the fly,
we will have to fix it in the pathfinder module.

## How Has This Been Tested
local test. See the result.
Screenshot from 2020-01-16 18-18-34

## Types of changes

## Checklist


msavva:
If we introduce this as a stopgap measure, I think we should add a TODO comment indicating that. 
 Otherwise, we risk this stopgap measure being forgotten.

bigbike:
Comment was added to address reviewer's concern.

Commit it in.

PR #438, opened 2020-01-21
[esp/sim] move Simulator to sim namespace

msbaines:
Goal is to collapse Simulator and SimulatorWithAgents into one class.
Moved Simulator to sim namespace an in intermediate step. Next step
is to pull up SimulatorWithAgents logic into Simulator.

The change is part of the sensor refactoring.

## Motivation and Context

Collapse Simulator and SimulatorWithAgents into one class.

## How Has This Been Tested

SimTest

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/bindings.cpp
+++ b/src/esp/bindings/bindings.cpp
@@ -69,6 +69,7 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
   esp::physics::initPhysicsBindings(m);
   esp::scene::initSceneBindings(m);
   esp::sensor::initSensorBindings(m);
+  esp::sim::initSimBindings(m);
 
   py::bind_map<std::map<std::string, std::string>>(m, "MapStringString");
 }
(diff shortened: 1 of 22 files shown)


src/esp/bindings/bindings.cpp
```diff
@@ -69,6 +69,7 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
   esp::physics::initPhysicsBindings(m);
   esp::scene::initSceneBindings(m);
   esp::sensor::initSensorBindings(m);
+  esp::sim::initSimBindings(m);
```
erikwijmans: I believe this needs to be moved before init sensor bindings for doc generation to cross-link correctly.

erikwijmans: Actually, it looks like we have a lot of things that aren't cross linking correctly anymore :/

msbaines: How does the cross-linking work? How do I verify it?

msbaines: Talked to @erikwijmans offline. Will fix cross-linking in a separate change.

PR #439, opened 2020-01-21
Allow sliding in habitat-sim to be turned off.

erikwijmans:
## Motivation and Context

In habitat-sim, agents slide when they collide.  This was shown to hurt sim2real performance.

The implementation of no sliding here differs considerably from the previous implementation.  The previous implementation was pretty hacky (although it did allow for intermediary values between on and off).  This implementation changes the core of recast and allows for sliding to be truly disabled.  See the commit on the recast side here: https://github.com/erikwijmans/recastnavigation/commit/354591e10a772b8057e42a4f9de701155b9aaa02

I also added method documentation and fixed a few little bugs here and there in the pathfinder (I can break these into separated PRs if needed, but GitHub makes that a pain..)

## How Has This Been Tested

Via the unit test!

## Types of changes


src/esp/nav/PathFinder.h
```diff
@@ -129,60 +178,202 @@ class PathFinder : public std::enable_shared_from_this<PathFinder> {
...
-
  protected:
   bool initNavQuery();
   void removeZeroAreaPolys();
-  std::vector<vec3f> prevEnds;
+  template <typename T>
+  T tryStepImpl(const T& start, const T& end, bool allowSliding);
 
-  impl::IslandSystem* islandSystem_ = nullptr;
+  IslandSystemUniquePtr islandSystem_;
```
msbaines: Why not just use a regular unique_ptr since it's just a C++ class?

src/esp/nav/PathFinder.h
```diff
@@ -129,60 +178,202 @@ class PathFinder : public std::enable_shared_from_this<PathFinder> {
...
   std::pair<vec3f, vec3f> bounds() const { return bounds_; }
 
-  friend impl::ActionSpaceGraph;
-
  protected:
   bool initNavQuery();
   void removeZeroAreaPolys();
-  std::vector<vec3f> prevEnds;
+  template <typename T>
+  T tryStepImpl(const T& start, const T& end, bool allowSliding);
```
msbaines: Should this be handled via a boolean state flag? Wouldn't allowSliding always be passed as true or false for a give PathFinder instance?

src/esp/nav/PathFinder.h
```diff
@@ -129,60 +178,202 @@ class PathFinder : public std::enable_shared_from_this<PathFinder> {
...
 
-  impl::IslandSystem* islandSystem_ = nullptr;
+  IslandSystemUniquePtr islandSystem_;
 
   NavMeshUniquePtr navMesh_;
   NavMeshQueryUniquePtr navQuery_;
-  dtQueryFilter* filter_;
+  dtQueryFilterUniquePtr filter_;
```
msbaines: Much of the patch is due to the change. Might be simpler as a separate PR.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -127,6 +127,7 @@ void initGfxBindings(py::module& m) {
       .def_readwrite("gpu_device_id", &SimulatorConfiguration::gpuDeviceId)
       .def_readwrite("compress_textures",
                      &SimulatorConfiguration::compressTextures)
+      .def_readwrite("allow_sliding", &SimulatorConfiguration::allowSliding)
```
msbaines: You will need to merge now that Simulator move has landed.

src/esp/nav/PathFinder.cpp
```diff
@@ -530,6 +535,8 @@ bool PathFinder::build(const NavMeshSettings& bs,
     }
   }
 
+  removeZeroAreaPolys();
```
msbaines: Maybe add one-line to the comment message mentioning this bug fix. It's nice have a list of bug fixes in the commit message. Can be handy later on when searching through 'git log'.

src/esp/nav/PathFinder.h
erikwijmans: It is forward declared, so it complains about not knowing how the size of the class.

src/esp/nav/PathFinder.h
erikwijmans: It seemed logical to just have both methods be implemented on the pathfinder and just choose the one to call.  May also be useful/let people play around with turning on/off sliding stochastically.

src/esp/nav/PathFinder.h
erikwijmans: It's the unique_ptr that complains, i.e. `std::unique_ptr<impl::IslandSystem>` fails.

erikwijmans: Probably the right thing to do is just pimpl pathfinder, I will back out these changes and do that in a different PR.

msbaines: PathFinder::~PathFinder() = default; should definitely fix the compile error you had with std::unique_ptr.

The issue is that default PathFinder destructor is inline and does not have access to the Impl::IslandSystem concrete type. If you move the default destructor to the PathFinder.cpp, the issue goes away.

I just pull this patch and double-checked that it would work:

diff --git a/src/esp/nav/PathFinder.cpp b/src/esp/nav/PathFinder.cpp
index 7cd67e5..9671ddc 100644
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -1062,5 +1062,7 @@ bool PathFinder::isNavigable(const vec3f& pt,
   return true;
 }

+  PathFinder::~PathFinder() = default;
+
 }  // namespace nav
 }  // namespace esp
diff --git a/src/esp/nav/PathFinder.h b/src/esp/nav/PathFinder.h
index 96ea36c..79fa961 100644
--- a/src/esp/nav/PathFinder.h
+++ b/src/esp/nav/PathFinder.h
@@ -198,6 +198,7 @@ class PathFinder : public std::enable_shared_from_this<PathFinder> {
    * @brief Constructor.
    */
   PathFinder();
+  ~PathFinder();

   bool build(const NavMeshSettings& bs,
              const float* verts,
@@ -369,7 +370,7 @@ class PathFinder : public std::enable_shared_from_this<PathFinder> {
   template <typename T>
   T tryStepImpl(const T& start, const T& end, bool allowSliding);

-  IslandSystemUniquePtr islandSystem_;
+  std::unique_ptr<impl::IslandSystem> islandSystem_;

   NavMeshUniquePtr navMesh_;
   NavMeshQueryUniquePtr navQuery_;

PR #441, opened 2020-01-22
Assert createRenderer is configured for Simulator::recomputeNavmesh

aclegg3:
## Motivation and Context

The `Simulator::recomputeNavmesh` function requires scene geometry to rebuild the recast navmesh for the scene. This geometry is not loaded when no sensors are configured. Trying to call this function without geometry loaded results in a failed assert during mesh joining which is non-obvious. 

This PR introduces a clearer assertion that `SimulatorConfiguration::createRenderer == true` is required to use this function.

May want to discuss whether this behavior is desired, or geometry loading should be decoupled from sensor initialization and what that would entail.

## How Has This Been Tested
Add the following code to the end of `test_recompute_navmesh.py` and see assertion failure.
```
    #test recompute navmesh with no sensors
    cfg_settings["color_sensor"] = False
    hab_cfg = examples.settings.make_cfg(cfg_settings)
    sim.reconfigure(hab_cfg)
    sim.recompute_navmesh(sim.pathfinder, navmesh_settings)
```

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/Simulator.cpp
+++ b/src/esp/gfx/Simulator.cpp
@@ -373,7 +373,12 @@ double Simulator::getWorldTime() {
 
 bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
                                  const nav::NavMeshSettings& navMeshSettings) {
-  CHECK(config_.createRenderer);
+  CORRADE_ASSERT(
+      config_.createRenderer,
+      "Simulator::recomputeNavMesh: SimulatorConfiguration::createRenderer is "
+      "false. Scene geometry is required to recompute navmesh. No geometry is "
+      "loaded without renderer initialization.",
+      false);
 
   assets::MeshData::uptr joinedMesh =
       resourceManager_.createJoinedCollisionMesh(config_.scene.id);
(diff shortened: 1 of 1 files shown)


src/esp/gfx/Simulator.cpp
```diff
@@ -373,6 +373,8 @@ double Simulator::getWorldTime() {
 
 bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
                                  const nav::NavMeshSettings& navMeshSettings) {
+  CHECK(config_.createRenderer);
```
erikwijmans:
```suggestion
  CORRADE_ASSERT(config_.createRenderer,
                 "Meshes were not loaded, cannot recompute navmesh.  Please "
                 "add a sensor to the agent(s).",
                 {});
```

Let's add an error message!

bigbike: Yes, we prefer to use CORRADE_ASSERT.
Also, add `Simulator::recomputeNavMesh: `  to the beginning of the message body.

Thanks!

Issue #421, opened 2020-01-09
Make habitat-sim conda installable


mathfac:
## Motivation
That involve multiple steps to clone, build and install habitat-sim and is limiting adoption of the framework by community dramatically.

## Pitch
Produce and publish pre-built binaries of habitat-sim for Ubuntu and make it conda installable (similar to [PyTorch install](https://pytorch.org/get-started/locally/#mac-prerequisites)): `conda install habitat_sim -c habitat`
<img width="833" alt="Screenshot 2020-01-09 17 01 03" src="https://user-images.githubusercontent.com/877440/72117222-be5d7080-3301-11ea-8dd1-2524da32525c.png">

That will decrease framework usage threshold significantly and make our Continues Integration setup simpler. 

## Additional context
As reference point [PyTorch night builds CI system](https://github.com/pytorch/pytorch/tree/master/.circleci) can be checked and partially replicated.

dhruvbatra:
Yes! This will be a high-impact feature!

mathfac:
Reference point to previous @erikwijmans effort to make it done: [https://github.com/facebookresearch/habitat-sim/tree/conda-build](https://github.com/facebookresearch/habitat-sim/compare/conda-build?expand=1).

erikwijmans:
I took another looked a pytorch's conda builder to see what differences are there.  I noticed 2 things:

1.  They include the path to the conda env's lib folder in their rpath manipulations: https://github.com/pytorch/builder/blob/master/conda/pytorch-nightly/build.sh#L118
2. They do the build of the binaries in a docker container.  My guess is that their choice of base container and OS is what helps make the binaries compatible with other linux variants: https://github.com/pytorch/builder/blob/master/conda/Dockerfile.  We should try building the binaries in a centos7 cudagl container (https://hub.docker.com/r/nvidia/cudagl)

dhruvbatra:
@mpiseno -- now that #445 has landed, can we add instructions in the README on how to use the conda build?

shuishida:
@dhruvbatra Hello, I was trying to install Habitat sim onto my linux but for some reason it didn't build properly, so I was looking for an alternative solution when I came across this thread. I was wondering if you could point me to the conda-installable habitat sim solution. I tried

`conda install -c aihabitat -c conda-forge habitat-sim`

which I found in the README, but it produced the following errors:

```Collecting package metadata (current_repodata.json): done
[...]

PR #445, opened 2020-01-24
Make Habitat-sim conda installable 

mpiseno:
## Motivation and Context

Made habitat-sim conda installable on macOS. Run ```conda install -c aihabitat -c conda-forge habitat-sim``` to install it. The Linux version is also on anaconda cloud, however I'm not confident it works properly. For the headless linux version, you need to first run ```conda install -c aihabitat -c conda-forge headless``` to get the feature.

Just creating an initial pr for feedback, particularly with the linux build as I continue to work on this.

Related issue: https://github.com/facebookresearch/habitat-sim/issues/421

## How Has This Been Tested

For macOS, I have run the install command described above and simply ran ```import habitat_sim``` from a python shell. I then ran a python file with one of the habitat-sim demos from https://aihabitat.org/docs/habitat-api/habitat-sim-demo.html and it worked fine. As for Linux, I tried the same thing but the demo failed (might be because I was trying to run from a linux docker container that was not headless).

EDIT: Erik Tested on the GT cluster (headless linux) and all the tests passed

## Types of changes

I used a Dockerfile from the pytorch/builder github repository to create a linux container for the build. Currently only the headless build works correctly, as the non-headless build does not compile. The README contains instructions on how to create the docker container and build from there.


conda-build/common/install_cuda.sh
```diff
@@ -0,0 +1,186 @@
...
+    ./setup --silent --no-opengl-libs --toolkit
+    rm -f setup
+
+    # patch 1
+    wget -q https://developer.nvidia.com/compute/cuda/9.2/Prod2/patches/1/cuda_9.2.148.1_linux -O setup
+    chmod +x setup
+    ./setup -s --accept-eula
+    rm -f setup
+
+    # install CUDA 9.2 CuDNN
```
erikwijmans: We can probably remove all the cudnn stuff, I doubt it matters for us.

conda-build/habitat-sim/build.sh
```diff
@@ -0,0 +1,80 @@
...
+  export PATH=/public/apps/cuda/${CUDA_VER}/bin:${PATH}
+  build_args+=(--cmake-args="-DCUDA_TOOLKIT_ROOT_DIR=${CUDA_HOME}")
+fi
+
+if [ ${WITH_BULLET} == "1" ]; then
+  build_args+=("--with-bullet")
+fi
+
+if [ $(uname) == "Linux" ]; then
+  cp -r /usr/include/EGL ${PREFIX}/include/.
```
erikwijmans: We may be able to remove the hacks in the if block here now

erikwijmans:
Works on the GT cluster

erikwijmans:
Confirmed on Oregon state's cluster.

erikwijmans:
@mpiseno -- We should add somewhere to make sure to do `git submodule update --init --recursive` before building the binaries, update the submodules won't work correctly in the conda-build env.

conda-build/habitat-sim/build.sh
```diff
@@ -0,0 +1,80 @@
...
+  if [ -f 'bin/habitat-viewer' ]; then
+    patchelf --set-rpath "\$ORIGIN/../${ext_folder}:\$ORIGIN/../lib" --force-rpath bin/habitat-viewer
+  fi
+
+  find $(dirname ${hsim_bindings}) -name "*Corrade*so" | xargs -I {} patchelf --set-rpath "\$ORIGIN:\$ORIGIN/../../../.." --force-rpath {}
+
+  pushd $(dirname ${corrade_bindings})/corrade
+    find . -name "*so" | xargs -I {} patchelf --set-rpath "\$ORIGIN/../habitat_sim/_ext:\$ORIGIN/../../.." --force-rpath {}
+  popd
+fi
```
mosra: I think these two RPATH-related blocks could be done in a clearer way directly from within CMake (which would make them available not just for Conda but potential other packaging systems as well). 

I was fiddling with RPATH few years ago for a dynamically-loadable module on iOS, unfortunately forgot most of the particulars since then :sweat_smile: This could help: https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling

erikwijmans: Noted!  For now this set of wonkiness seems to be working and debugging these rpaths was a pain as a full conda build + install takes about 15 minutes (really the fault of conda here, ccache does its thing and habitat takes only a couple minutes).

mathfac:
Looks like you picked up unexpected changes for submodule recastnavigatio. Try to do submodule update or reset hard to master deps.

PR #446, opened 2020-01-24
Refactor RigidBody to Feature

aclegg3:
## Motivation and Context

RigidBody objects should be Features attached to a SceneNode rather than derived from them. 

This change also:
 - refactors raw pointers to unique pointers
 - adds `override` specifier to some functions to address warnings
 - makes `SceneNode` deletion optional for `PhysicsManager::removeObject()`
 - replaces `RigidBody::removeObject()` with Destructor

## How Has This Been Tested

Existing tests pass.

## Types of changes

## Checklist


src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -272,6 +270,7 @@ bool BulletRigidObject::removeObject() {
   }
   bWorld_.reset();  // release shared ownership of the world
   rigidObjectType_ = RigidObjectType::NONE;
+  delete &node();
```
msbaines: After node() is deleted, 'this' no longer exists. That's confusing. This function seems like it's the destructor. Why can't teardown be implemented with unique_ptrs and destructors instead of calls to removeObject from the BulletPhysicsManager destructor?

mosra: Yeah, the semantics of `removeObject()` is a bit confusing now ... since this is a feature, isn't it enough to remove the feature and keep the object as-is (so it loses its rigid body capabilities)?

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -212,7 +210,7 @@ bool BulletRigidObject::initializeObject(
   }
 
   //! Bullet rigid body setup
-  bObjectMotionState_ = new Magnum::BulletIntegration::MotionState(*this);
+  bObjectMotionState_ = new Magnum::BulletIntegration::MotionState(node());
```
mosra: What about deriving `BulletRigidObject` from `MotionState`? Since those two are tightly coupled anyway.

As it is now, creating a `BulletRigidObject` and then destructing it again would leave the (dangling) MotionState feature still attached to the original object. I think both should be removed, and the easiest way to achieve that is to derive from it.

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: Good point, I've refactored the deconstruction/removal process.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -181,46 +182,46 @@ void BulletPhysicsManager::stepPhysics(double dt) {
...
-  static_cast<BulletRigidObject*>(sceneNode_)
-      ->setFrictionCoefficient(frictionCoefficient);
+  sceneNode_->setFrictionCoefficient(frictionCoefficient);
 }
 
 void BulletPhysicsManager::setSceneRestitutionCoefficient(
     const double restitutionCoefficient) {
-  static_cast<BulletRigidObject*>(sceneNode_)
-      ->setRestitutionCoefficient(restitutionCoefficient);
+  sceneNode_->setRestitutionCoefficient(restitutionCoefficient);
```
mosra: The `sceneNode_` name is confusing now, as this looks like these are all methods on a SceneNode, while it's a Feature instead.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -181,46 +182,46 @@ void BulletPhysicsManager::stepPhysics(double dt) {
...
 }
 
 double BulletPhysicsManager::getMargin(const int physObjectID) const {
   assertIDValidity(physObjectID);
-  return static_cast<BulletRigidObject*>(existingObjects_.at(physObjectID))
+  return dynamic_cast<BulletRigidObject*>(
+             existingObjects_.at(physObjectID).get())
```
mosra: Does the `assertIdValidity` check the type? If so, I'd keep the `static_cast`.. or at least `dynamic_cast` a reference so you don't do an `->` on a null pointer.

src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -141,13 +142,13 @@ class BulletRigidObject : public RigidObject {
...
    */
-  bool isActive();
+  bool isActive() override;
 
   /**
    * @brief Set an object as being actively simulated rather than sleeping.
    * See @ref btCollisionObject::activate.
    */
-  void setActive();
+  void setActive() override;
```
mosra: This added `override` is unrelated to this now being a subclass of MotionState, right?

aclegg3: Correct, this just just another function which is virtual in PhysicsManager but was not explicitly marked as override in BulletPhysicsManager.

src/esp/physics/bullet/BulletPhysicsManager.cpp
aclegg3: `assertIDValidity` only checks that this ID is allocated, it does not check type. However, it should be the case that the type must be `BulletRigidObject` if we get to this line due to the construction process, so we can stick to `static_cast`.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -92,12 +93,14 @@ int PhysicsManager::addObject(const std::string& configFile,
   return physObjectID;
 }
 
-int PhysicsManager::removeObject(const int physObjectID) {
+int PhysicsManager::removeObject(const int physObjectID, bool deleteSceneNode) {
```
msbaines: Why is deleteSceneNode needed? It seems to always be true.

mosra: Also, why is the `pysObjectID` returned? I don't see any purpose in that.

src/esp/physics/bullet/BulletPhysicsManager.cpp
mosra:
```suggestion
             existingObjects_.at(physObjectID))
```

Then this change could be discarded completely I guess, it's unnecessary.

src/esp/physics/PhysicsManager.cpp
mosra: I tried to figure out a better variant of the API (like returning the `objectNode` so the caller can delete it or not), but it's not ideal either, as it changes the default to "not delete".

aclegg3: I agree the ID return is unnecessary, I'll change this. 

I provide the option for SceneNode delete so that the object can be detached from physics while preserving the SceneNode for rendering if desired. Most of the time you'll want to remove it, but this provides an alternative (maybe you are holding a reference to the node for some purpose).

PR #447, opened 2020-01-24
Fix: Handle exit code 1 if there is no changes to doc

mathfac:
## Motivation and Context
Automatic documentation build [fails when there is no changes to push](https://app.circleci.com/jobs/github/facebookresearch/habitat-sim/5795). Added the check if there are changes to commit first.

## How Has This Been Tested
Switched running for current branch to check success.


mathfac:
[Tested properly,](https://app.circleci.com/jobs/github/facebookresearch/habitat-sim/5843) should be good to go.

PR #449, opened 2020-01-25
Linear Frustum Culling

bigbike:
## Motivation and Context

Improve the rendering speed.

Only render drawables that can pass the frustum culling test. We did it by testing the absolute AABB of each mesh component, pre-computed in the loading stage, against the current frustum.

Do linear test through all the drawables. No optimization is added.

## How Has This Been Tested
Local test.

## Types of changes

## Checklist


bigbike:
Try to get early feedback while working on the tests.

bigbike:
Designed a test case, and visually verified it in the viewer.
The result from the test matches the visual result in viewer.
(5 boxes: box 3 is invisible, and box 0, 1, 2, 4 are visible)

image

src/tests/ResourceManagerTest.cpp
```diff
@@ -77,71 +77,3 @@ TEST(ResourceManagerTest, createJoinedCollisionMesh) {
     ASSERT_EQ(indexGroundTruth[iix], joinedBox->ibo[iix]);
   }
 }
-
```
bigbike: This test has been moved to the newly created CullingTest.cpp

bigbike:
Dear reviwers,

I just created the test for the frustum culling.
The test case and test results have been visually verified in utility viewer. See the picture in the previous comments.

Now, this diff is ready to be reviewed again. 

Thank you!

src/esp/bindings/GfxBindings.cpp
```diff
@@ -46,6 +46,8 @@ void initGfxBindings(py::module& m) {
         Set this `Camera`'s projection matrix.
       )",
            "width"_a, "height"_a, "znear"_a, "zfar"_a, "hfov"_a)
+      .def("enable_frustum_culling", &RenderCamera::enableFrustumCulling,
```
mosra: `enableFrustumCulling(false)` feels weird ... what about `setFrustumCullingEnabled()` instead?

src/esp/gfx/RenderCamera.cpp
```diff
@@ -39,10 +43,46 @@ RenderCamera& RenderCamera::setProjectionMatrix(int width,
...
 
-RenderCamera& RenderCamera::draw(MagnumDrawableGroup& drawables) {
-  // TODO: visibility culling
-  MagnumCamera::draw(drawables);
-  return *this;
+uint32_t RenderCamera::draw(MagnumDrawableGroup& drawables) {
+  uint32_t numDrawnObjects = drawables.size();
+  if (!frustumCulling) {
+    MagnumCamera::draw(drawables);
+  } else {
```
mosra:
```cpp
if (!frustumCulling) {
  MagnumCamera::draw(drawables);
  return drawables.size();
}
```

then the (longer) rest of the function doesn't need to be indented and the branching flow is clearer (to me at least).

src/esp/gfx/RenderCamera.cpp
```diff
@@ -39,10 +43,46 @@ RenderCamera& RenderCamera::setProjectionMatrix(int width,
...
+    // camera frustum relative to world origin
+    const Mn::Math::Frustum<float> frustum =
+        Mn::Math::Frustum<float>::fromMatrix(projectionMatrix() *
```
mosra:
```suggestion
        Mn::Frustum::fromMatrix(projectionMatrix() *
```

Same elsewhere, and in the test. *All* `Math` classes have convenience float/double typedefs in the root namespace.

src/tests/CullingTest.cpp
```diff
@@ -0,0 +1,170 @@
...
+  // The camera to be set:
+  // pos: {7.3589f, -6.9258f,4.9583f}
+  // rotation: 77.4 deg, around {0.773, 0.334, 0.539}
+  // fov = 39.6 deg
+  // resolution: 800 x 600
+  // clip planes (near: 0.1m, far: 100m)
+  // with such a camera, the box 3 should be invisible, box 0, 1, 2, 4 should be
+  // visible.
+
+  // NOTE: the following test reults have been visually verified in utility
```
mosra:
```suggestion
  // NOTE: the following test results have been visually verified in utility
```

src/tests/CullingTest.cpp
```diff
@@ -0,0 +1,170 @@
...
+        dynamic_cast<esp::scene::SceneNode&>(drawables[iDrawable].object())
+            .getAbsoluteAABB();
+    if (aabb) {
+      if (Mn::Math::Intersection::rangeFrustum(*aabb, frustum)) {
+        visible++;
+      } else {
+        CHECK_EQ(iDrawable, invisibleBoxGroundTruth);
+      }
+    }
+  }
```
mosra: This is kinda brittle with a false sense of security because it's basically the same code as in the culling routine, but **doesn't actually call the culling routine at all**, unless I overlooked something? It's also hard to extend further.

I am not sure if testing a code by rewriting it again in a test is a good idea, because those two will inevitably get out of sync. One thing that could work is:

1. add a culling helper (`std::size_t RenderCamera::cull(std::vector<std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>, Mn::Matrix4>>&)`) that does just the `rename_if()` part, returning how many elements at front are not culled away. This one will be used from `draw()` as well as from this test.
2. get the vector of pairs from the drawable group and call `cull()` on these in the test, remember the returned value
3. set up a [SampleQuery](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1SampleQuery.html) and draw everything in the second part of the vector, expecting 0 samples to be reported by the query (to check for things that shouldn't be culled but were).
4. for each object in the first part, set up another query, clear and draw it, expect >0 samples to be drawn. Important is that each drawn object has a new query, to check for things that should be culled but weren't.
5. and as a final check, `renderCamera.draw()` should report the same value as what `cull()` returned above.

This shouldn't be too hard to set up I think, but still be pretty robust and easy to extend for new culling methods.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -39,10 +43,46 @@ RenderCamera& RenderCamera::setProjectionMatrix(int width,
...
+              // obtain the absolute aabb
+              Corrade::Containers::Optional<Mn::Range3D> aabb =
+                  dynamic_cast<scene::SceneNode&>(a.first.get().object())
+                      .getAbsoluteAABB();
+              if (aabb) {
+                // if it has an absolute aabb, it is a static mesh
+                return !Mn::Math::Intersection::rangeFrustum(*aabb, frustum);
+              } else {
+                // keep the drawable if its node does not have an absolute AABB
+                return true;
```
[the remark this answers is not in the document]
bigbike: This is a bug. Good catch.
Thank you very much.

src/tests/CullingTest.cpp
bigbike: Typo. Good catch. Thank you.

src/tests/CullingTest.cpp
bigbike: OK. This is sharp. I did have a bit concern when I wrote it. Adding a new helper function seems a good idea. Will update immediately.

src/esp/gfx/RenderCamera.cpp
bigbike: Good to know. Will change.

src/esp/bindings/GfxBindings.cpp
bigbike: OK. Will do.

src/tests/CullingTest.cpp
```diff
@@ -0,0 +1,210 @@
...
+    for_each(
+        newEndIter, drawableTransforms.end(),
+        [&](const std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>,
+                            Mn::Matrix4>& a) { objects.emplace_back(a); });
+    Mn::GL::SampleQuery q{Mn::GL::SampleQuery::Target::AnySamplesPassed};
+    q.begin();
+    renderCamera.MagnumCamera::draw(objects);
+    q.end();
+
+    EXPECT_EQ(q.result<bool>(), false);
```
bigbike: It is funny the check would fail here indicating that there is something visible. So I hacked the viewer. 
It turns out the "newEndIter" contains 1 actually visible object. I am really wondering  if "remove_if" indeed moved the visible objects to the first half. I am still debugging it.

bigbike: Update: It is solved. See the current code.
The invisibles are not stored in the 2nd part of the vector because std::remove_if only move elements, not swap elements.

src/tests/CullingTest.cpp
```diff
@@ -0,0 +1,231 @@
...
+            objects;
+        objects.emplace_back(a);
+
+        Mn::GL::SampleQuery q{Mn::GL::SampleQuery::Target::AnySamplesPassed};
+        q.begin();
+        renderCamera.MagnumCamera::draw(objects);
+        q.end();
+
+        // check if it a genuine visible drawable
+        EXPECT_EQ(q.result<bool>(), true);
```
bigbike: @mosra : After I fixed Test 1, now the Test 2 will fail. 

I copied the this piece of code (`renderOneDrawable` and the `for_each`) to viewer. It can render these objects on the screen, which means they are visible.

Not sure why it reports false. Did I setup anything wrong?

bigbike: When fixing test 1, I was working on a Fedora system (linux). I am pretty sure if something is rendered, the sampleQuery can capture it.

After I went home, I was using may Macbook to fix Test 2, which gave me the above error (sampleQuery told me some objects (supposed to be visible) were invisible.) I hope it would again a OS-specific issue. (Hope it is not as I can see the CI test also failed.)

I will try it tomorrow in the office with a linux machine.

src/esp/gfx/RenderCamera.h
```diff
@@ -41,9 +41,41 @@ class RenderCamera : public MagnumCamera {
...
+   * absolute transformation
+   * @return an iterator to the new end of the range, which contains visible
+   * pairs
+   *
+   * NOTE: user are not encouraged to call this function directly.
+   * The preferred way is to enable the frustum culling by calling @ref
+   * setFrustumCullingEnabled and then call @ref draw
+   */
+  std::vector<std::pair<std::reference_wrapper<Magnum::SceneGraph::Drawable3D>,
+                        Magnum::Matrix4>>::iterator
```
mosra: I would really just return a `std::size_t` (`return std::remove_if(...) - drawableTransforms.begin()`), it's shorter, easier to read/document, doesn't get invalidated, and doesn't depend on the container type :)

src/tests/CullingTest.cpp
bigbike: I just remotely connected to a linux box and tested it. It failed. So it should not be an OS issue.

src/esp/gfx/RenderCamera.h
```diff
@@ -41,9 +41,41 @@ class RenderCamera : public MagnumCamera {
...
+  cull(std::vector<
+       std::pair<std::reference_wrapper<Magnum::SceneGraph::Drawable3D>,
+                 Magnum::Matrix4>>& drawableTransforms);
 
  protected:
+  bool frustumCullingEnabled = false;
```
mosra:
```suggestion
  bool frustumCullingEnabled_ = false;
```

or so I guess based on the naming scheme elsewhere?

src/tests/CullingTest.cpp
mosra: Argh, sorry, for some reason I thought `std::remove_if()` non-destructively partitions the elements. [std::partition()](https://en.cppreference.com/w/cpp/algorithm/partition) is what should be used there, and then it returns both parts correctly. With that I think the `object.erase()` block above isn't needed anymore either, right?

Another option is `std::stable_partition()` but I don't think we need to preserve the order.

src/tests/CullingTest.cpp
mosra: Random shoot -- I think you'll need a framebuffer clear before each, otherwise the depth buffer will prevent things from being rendered in subsequent runs (which iteration fails? the first one? or the subsequent ones?).

bigbike: All of them failed.
Actually, if I disable the Test 1, it is the same. (If this is the dirty framebuffer issue, at least it should succeed at the 1st one.)
I hope you can download and try this branch on your linux machine.

Also, how can I clean the framebuffer at this point? Any docs can help me? Thanks a lot!

mosra: Depending on the OS, the depth buffer might be by default all 0s (in which case the first render would work) or all 1s (in which case it wouldn't). Ran into this exact issue on a different project last week.

How? Not sure, I mean there has to be some framebuffer the simulator had set up, no? :)

src/tests/CullingTest.cpp
bigbike: if remove_if() is used, then object.erase() is needed if I understand the docs correctly.
https://en.cppreference.com/w/cpp/algorithm/remove

src/tests/CullingTest.cpp
bigbike: I cannot remember where we set the framebuffer. Are we using the default framebuffer?
Anyway, I have to go. It is 2am. Hope you can try this test.

Thank you very much for all the feedbacks. I will continue tomorrow.

src/tests/CullingTest.cpp
mosra: Sure, sure, I meant regarding the 

> std::remove_if only move elements, not swap elements

Then, with `std::partition()` the second call to `drawableTransforms()` and the complex block after in the test with `std::addressof()` inside is no longer needed and it could be just:

```cpp
std::vector objectsVisible{drawableTransforms.begin(), newEndIter};
std::vector objectsCulled{newEndIter, drawableTransforms.end()};
```

and then continuing on each of those.

src/tests/CullingTest.cpp
mosra: I'll try to test, but not sure if today.

The default framebuffer can't really be relied on, since some platforms (macOS) don't even have it and EGL has only a 16x16 one (making all sample queries return 0 anyway).

If getting the framebuffer out of the simulator would be too painful, it's always possible to create a temporary one, with just a color buffer for example.

erikwijmans: @bigbike This is where we get the framebuffer: https://aihabitat.org/docs/habitat-cpp/classesp_1_1gfx_1_1RenderTarget.html

Our windowless context does not create a framebuffer, so IDK what you are even rendering too currently :)

bigbike: Thank you very much. That makes a lot of sense.
On Fedora, I can build in GUI mode. So all the tests work perfectly. Just tried.

But on CI, and devserver, they must be in headless mode, which causes failures. I need this render target.
Thank you very much, @erikwijmans !

src/tests/CullingTest.cpp
bigbike: Hi @mosra,
Yes, `std::partition` can divide the set into two sub-sets, but it is slower than `std::remove_if`. 
There is just no "free lunch". This piece of code is performance-critical. So I would like to keep `std::remove_if`.
By using `std::remove_if`, the entries after the new end will still have their old values. It is fine, as we are only bothered by this fact in the tests.

src/tests/CullingTest.cpp
bigbike: All right. It is indeed related to the frame buffer, and OS-specific.
After I added the RenderTarget, it works on mac OSX, Ubuntu now. (Fedora does not have this issue since the beginning.)

@mosra , @erikwijmans , your insights really help here. Thank you!

src/esp/gfx/RenderCamera.h
bigbike: @mosra :
 This is a great suggestion. I agree with you that returning such a long, tedious iterator is not cool at all.
However,
- It is convenient since STL functions such as `std::erase` are defined on the iterators, and I will use the return values in such functions. It is very practical.
- I will not expect user to call this function directly as I stated clearly in the comment. It is not encouraged. "Cull()" is just a helper function. I would not even bother to extract and create it if it is not for the testing purpose. And for the same reason, I have to make it "public" instead of "protected". The point is, the only user perhaps is just me, the developer, not the user.

So if you are not strongly against it, I prefer to keep it in this way.

bigbike:
Dear reviewers,

CI tests (including the new tests from this PR) are passed. This diff now is ready to go.
Please take another look. See if you have further suggestions or comments.

Thanks.

src/esp/gfx/RenderCamera.h
bigbike: Update:
I looked at this function again tonight. Yes, it is too long. I changed my mind.
Now it will return the number of visible drawables, as you suggested.
The changes afterwards were not that hard actually, which was great.

src/tests/CullingTest.cpp
mosra: Is the difference actually measurable? Don't think so. AFAICT the difference is that one does `std::swap(a, b)` while the other `a = std::move(b)`, and I'd say the amount of operations is roughly the same, as `std::partition` doesn't need to keep the order.

bigbike: I wrote a test as follows. Hope it is correct.

```cpp
#include <algorithm>
#include <iostream>
#include <sys/time.h>
#include <time.h>
#include <vector>

double cpuSecond() {
  struct timeval tp;
  gettimeofday(&tp, NULL);
  return ((double)tp.tv_sec + (double)tp.tv_usec * 1.e-6);
}

struct matrix4 {
  double e[16];
  matrix4() {
    for (int i = 0; i < 16; ++i) {
      e[i] = (double)(rand() / 10000) / 100.00;
    }
  }
};

const int range = 10000;

bool isInFirstHalf(const std::pair<int, matrix4> &a) {
  if (a.first < range / 2) {
    return true;
  }
  return false;
};

int main() {
  int nElem = 1 << 26;
  srand(time(NULL));

  auto testFunc =
      [&](std::vector<std::pair<int, matrix4>>::iterator (*func)(
              std::vector<std::pair<int, matrix4>>::iterator beginIter,
              std::vector<std::pair<int, matrix4>>::iterator endIter,
              bool (*criterion)(const std::pair<int, matrix4> &a)),
          const std::string &funcNameStr, int rounds) {
        double totalTime = 0.0;
        std::cerr << std::endl
                  << " ====== Testing: " << funcNameStr
                  << " ======" << std::endl;
        for (int iRound = 0; iRound < rounds; ++iRound) {
          std::cerr << "Round " << iRound << " : "
                    << "Preparing data... ";
          std::vector<std::pair<int, matrix4>> testVector;
          for (int i = 0; i < nElem; ++i) {
            testVector.emplace_back(rand() / range, matrix4{});
          }
          std::cerr << "DONE. ";
          double start = cpuSecond();
          (*func)(testVector.begin(), testVector.end(), isInFirstHalf);
          double elaps = cpuSecond() - start;
          std::cerr << " Time used by " << funcNameStr << ": " << elaps
                    << " sec." << std::endl;

          totalTime += elaps;
        }
        std::cerr << "Total time used by " << funcNameStr << ": " << totalTime
                  << " sec." << std::endl;
        std::cerr
            << "NOTE: Time to prepare data is excluded from above results."
            << std::endl;
      };


  const int rounds = 5;

  testFunc(&std::remove_if, "std::remove_if", rounds);
  testFunc(&std::partition, "std::partition", rounds);

  return 0;
}
```

Result is as follows:
```
 ====== Testing: std::remove_if ======
Round 0 : Preparing data... DONE.  Time used by std::remove_if: 1.78626 sec.
Round 1 : Preparing data... DONE.  Time used by std::remove_if: 1.78558 sec.
Round 2 : Preparing data... DONE.  Time used by std::remove_if: 1.78178 sec.
Round 3 : Preparing data... DONE.  Time used by std::remove_if: 1.77711 sec.
Round 4 : Preparing data... DONE.  Time used by std::remove_if: 1.78206 sec.
Total time used by std::remove_if: 8.91279 sec.
NOTE: Time to prepare data is excluded from above results.

 ====== Testing: std::partition ======
Round 0 : Preparing data... DONE.  Time used by std::partition: 1.84218 sec.
Round 1 : Preparing data... DONE.  Time used by std::partition: 1.84741 sec.
Round 2 : Preparing data... DONE.  Time used by std::partition: 1.85215 sec.
Round 3 : Preparing data... DONE.  Time used by std::partition: 1.85704 sec.
Round 4 : Preparing data... DONE.  Time used by std::partition: 1.85889 sec.
Total time used by std::partition: 9.25767 sec.
NOTE: Time to prepare data is excluded from above results.

```

I also switched the testing sequence (testing std::partition first).
Result is as follows: 
```
 ====== Testing: std::partition ======                                                                                
Round 0 : Preparing data... DONE.  Time used by std::partition: 1.86608 sec.                                          
Round 1 : Preparing data... DONE.  Time used by std::partition: 1.86733 sec.                                          
Round 2 : Preparing data... DONE.  Time used by std::partition: 1.87532 sec.                                          
Round 3 : Preparing data... DONE.  Time used by std::partition: 1.88619 sec.                                          
Round 4 : Preparing data... DONE.  Time used by std::partition: 1.8821 sec.                                           
Total time used by std::partition: 9.37702 sec.                                                                       
NOTE: Time to prepare data is excluded from above results.                                                            
                                                                                                                      
 ====== Testing: std::remove_if ======                                                                                
Round 0 : Preparing data... DONE.  Time used by std::remove_if: 1.82084 sec.                                          
Round 1 : Preparing data... DONE.  Time used by std::remove_if: 1.80141 sec.                                          
Round 2 : Preparing data... DONE.  Time used by std::remove_if: 1.80309 sec.                                          
Round 3 : Preparing data... DONE.  Time used by std::remove_if: 1.79911 sec.                                          
Round 4 : Preparing data... DONE.  Time used by std::remove_if: 1.79908 sec.                                          
Total time used by std::remove_if: 9.02353 sec.                                                                       
NOTE: Time to prepare data is excluded from above results.   
```

Conclusion: `std::remove_if` is slightly faster than `std::partition`.

bigbike: Intuitively, `swap(a, b)` involves 3 move assignments while `a = std::move(b)` is just one.

bigbike: I wrote a test. Hope it is correct and helps.
```cpp
#include <algorithm>
#include <iostream>
#include <sys/time.h>
#include <time.h>
#include <vector>

double cpuSecond() {
  struct timeval tp;
  gettimeofday(&tp, NULL);
  return ((double)tp.tv_sec + (double)tp.tv_usec * 1.e-6);
}

struct matrix4 {
  double e[16];
  matrix4() {
    for (int i = 0; i < 16; ++i) {
      e[i] = (double)(rand() / 10000) / 100.00;
    }
  }
};

const int range = 10000;

bool isInFirstHalf(const std::pair<int, matrix4> &a) {
  if (a.first < range / 2) {
    return true;
  }
  return false;
};

int main() {
  int nElem = 1 << 26;
  srand(time(NULL));

  auto testFunc =
      [&](std::vector<std::pair<int, matrix4>>::iterator (*func)(
              std::vector<std::pair<int, matrix4>>::iterator beginIter,
              std::vector<std::pair<int, matrix4>>::iterator endIter,
              bool (*criterion)(const std::pair<int, matrix4> &a)),
          const std::string &funcNameStr, int rounds) {
        double totalTime = 0.0;
        std::cerr << std::endl
                  << " ====== Testing: " << funcNameStr
                  << " ======" << std::endl;
        for (int iRound = 0; iRound < rounds; ++iRound) {
          std::cerr << "Round " << iRound << " : "
                    << "Preparing data... ";
          std::vector<std::pair<int, matrix4>> testVector;
          for (int i = 0; i < nElem; ++i) {
            testVector.emplace_back(rand() % range, matrix4{});
          }
          std::cerr << "DONE. ";
          double start = cpuSecond();
          (*func)(testVector.begin(), testVector.end(), isInFirstHalf);
          double elaps = cpuSecond() - start;
          std::cerr << " Time used by " << funcNameStr << ": " << elaps
                    << " sec." << std::endl;

          totalTime += elaps;
        }
        std::cerr << "Total time used by " << funcNameStr << ": " << totalTime
                  << " sec." << std::endl;
        std::cerr
            << "NOTE: Time to prepare data is excluded from above results."
            << std::endl;
      };


  const int rounds = 5;

  testFunc(&std::remove_if, "std::remove_if", rounds);
  testFunc(&std::partition, "std::partition", rounds);

  return 0;
}
```

The result is as follows:
```
 ====== Testing: std::remove_if ======
Round 0 : Preparing data... DONE.  Time used by std::remove_if: 1.74562 sec.
Round 1 : Preparing data... DONE.  Time used by std::remove_if: 1.73391 sec.
Round 2 : Preparing data... DONE.  Time used by std::remove_if: 1.73497 sec.
Round 3 : Preparing data... DONE.  Time used by std::remove_if: 1.73419 sec.
Round 4 : Preparing data... DONE.  Time used by std::remove_if: 1.73396 sec.
Total time used by std::remove_if: 8.68264 sec.
NOTE: Time to prepare data is excluded from above results.

 ====== Testing: std::partition ======
Round 0 : Preparing data... DONE.  Time used by std::partition: 1.75068 sec.
Round 1 : Preparing data... DONE.  Time used by std::partition: 1.76395 sec.
Round 2 : Preparing data... DONE.  Time used by std::partition: 1.76781 sec.
Round 3 : Preparing data... DONE.  Time used by std::partition: 1.80863 sec.
Round 4 : Preparing data... DONE.  Time used by std::partition: 1.75734 sec.
Total time used by std::partition: 8.84842 sec.
NOTE: Time to prepare data is excluded from above results.

```

I also changed the testing sequence (test `std::partition` first).
The result is as follows:
```
 ====== Testing: std::partition ======
Round 0 : Preparing data... DONE.  Time used by std::partition: 1.77457 sec.
Round 1 : Preparing data... DONE.  Time used by std::partition: 1.77498 sec.
Round 2 : Preparing data... DONE.  Time used by std::partition: 1.77569 sec.
Round 3 : Preparing data... DONE.  Time used by std::partition: 1.77508 sec.
Round 4 : Preparing data... DONE.  Time used by std::partition: 1.77671 sec.
Total time used by std::partition: 8.87702 sec.
NOTE: Time to prepare data is excluded from above results.

 ====== Testing: std::remove_if ======
Round 0 : Preparing data... DONE.  Time used by std::remove_if: 1.73139 sec.
Round 1 : Preparing data... DONE.  Time used by std::remove_if: 1.73032 sec.
Round 2 : Preparing data... DONE.  Time used by std::remove_if: 1.7295 sec.
Round 3 : Preparing data... DONE.  Time used by std::remove_if: 1.72571 sec.
Round 4 : Preparing data... DONE.  Time used by std::remove_if: 1.73221 sec.
Total time used by std::remove_if: 8.64913 sec.
NOTE: Time to prepare data is excluded from above results.
```

Conclusion: `std::remove_if` is slightly faster than `std::partition`.

bigbike:
Let me know if you have more concerns about this PR.
I am planning to commit it in on Fri.

bigbike:
Dear reviewers,

I just uploaded a new version.
Please be aware that I introduced an important change -- removed `frustumCullingEnabled_` from the `RenderCamera` class.

This afternoon, I am changing the code to introduce the culling to the front-end (python) so people can really adopt this new functionality in training and benchmark. I found that, `frustumCullingEnabled_` would make the `RenderCamera` a state machine at the **lowest level** of the chain (`simulator --> renderer --> renderCamera`). However, when `sim` starts, the argument (e.g., `enable_frustum_culling`) would pass from top to the bottom. Maintaining such state (as well as its consistency) at the bottom level is just too hard (because we can have many copies of render cameras) and unnecessary. It would be much easier if it is done at higher level (e.g., in `simulator` or `renderer`).

Based on this reason, now, I just passed it as an argument to the `Draw(...)` function within the `RenderCamera` class.

erikwijmans:
> Please be aware that I introduced an important change -- removed frustumCullingEnabled_ from the RenderCamera class.

This change makes sense to me -- good catch!

src/tests/CullingTest.cpp
bigbike: Update: I improved the above program tonight in order to do apple-by-apple comparison, namely, both methods work on exactly the same testing data.

The conclusion did not change. `std::remove_if` is a bit faster.

mosra: Fair enough. Pick whatever you prefer, I'd say the difference is not so huge so i'd still choose convenience in the test.

bigbike:
@mosra :

Thank you very much for your invaluable advices. I like them and I need them.
If possible, hope you can take a look at #400, which served as the foundation of the culling.
It was committed in a while ago but we missed your reviews.

Thank you!

PR #452, opened 2020-01-28
Constrained LookUp and LookDown

erikwijmans:
## Motivation and Context

Our current LookUp and LookDown actions allow you to spin a full 360 degrees.  This is very unrealistic.  This PR introduces a constraint to that.

## How Has This Been Tested

Via the test

## Types of changes

### DIFF
--- a/habitat_sim/agent/controls/default_controls.py
+++ b/habitat_sim/agent/controls/default_controls.py
@@ -4,11 +4,18 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-import magnum as mn
+from typing import Union
 
-import habitat_sim.bindings as hsim
-from habitat_sim.agent.controls.controls import ActuationSpec, SceneNodeControl
+import magnum as mn
+import numpy as np
+
+from habitat_sim.agent.controls.controls import (
+    ActuationSpec,
+    ConstrainedActuationSpec,
+    SceneNodeControl,
+)
 from habitat_sim.registry import registry
+from habitat_sim.scene import SceneNode
 
 __all__ = []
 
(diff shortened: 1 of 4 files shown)


habitat_sim/agent/controls/default_controls.py
```diff
@@ -18,67 +25,79 @@
...
     scene_node.translate_local(ax * distance)
 
 
-def _rotate_local(scene_node: hsim.SceneNode, theta: float, axis: int):
+def _rotate_local(scene_node: SceneNode, theta: float, axis: int):
     _rotate_local_fns[axis](scene_node, mn.Deg(theta))
     scene_node.rotation = scene_node.rotation.normalized()
 
 
+def _apply_look_constraint(scene_node: SceneNode, constraint: float):
```
msbaines: This implementation here seems to be specific to lookUp and lookDown.

habitat_sim/agent/controls/default_controls.py
```diff
@@ -18,67 +25,79 @@
 _Z_AXIS = 2
 
 _rotate_local_fns = [
-    hsim.SceneNode.rotate_x_local,
-    hsim.SceneNode.rotate_y_local,
-    hsim.SceneNode.rotate_z_local,
+    SceneNode.rotate_x_local,
```
msbaines: The refactoring is a large part of the patch which makes it harder to review. As I'm scrolling up and down, I find myself working extra hard to ignore the refactoring bits so that I can focus on the change.

habitat_sim/agent/controls/default_controls.py
```diff
@@ -18,67 +25,79 @@
...
     scene_node.rotation = scene_node.rotation.normalized()
 
 
+def _apply_look_constraint(scene_node: SceneNode, constraint: float):
+    constraint = mn.Deg(constraint)
+    rotation = scene_node.rotation
+    look_axis = rotation.transform_vector(mn.Vector3(0, 0, -1))
+    look_angle = mn.Rad(np.arctan2(look_axis[1], -look_axis[2]))
+
+    if look_angle > constraint:
```
msbaines: Might be simpler if _apply_look_constraint took constraint and actuation_spec.amount as input and returned the final rotation amount use for _rotate_local resulting in only one call to _rotate_local vs two.

habitat_sim/agent/controls/default_controls.py
erikwijmans: Yeah.. I really wish git/github did a better job at supporting things like this -- making another PR and then dealing with the resulting merge conflicts is painful, which is why I tend to do it this way.

msbaines: One idea would be to put the refactoring bits into a separate commit but part of the same PR.

habitat_sim/agent/controls/default_controls.py
```diff
@@ -36,21 +32,33 @@ def _move_along(scene_node: SceneNode, distance: float, axis: int):
...
 
 
-def _rotate_local(scene_node: SceneNode, theta: float, axis: int):
-    _rotate_local_fns[axis](scene_node, mn.Deg(theta))
-    scene_node.rotation = scene_node.rotation.normalized()
+def _rotate_local(
+    scene_node: SceneNode, theta: float, axis: int, constraint: Optional[float] = None
+):
+    theta = mn.Deg(theta)
+    if constraint is not None:
```
msbaines: Not sure about python idioms but this might be simpler as:

`if constraint:`

Double negatives are challenging for me to parse.

erikwijmans: `if <x> is not None` is very idiomatic in python, also, because it is duck typed, `if <x>` when you just want to unwrap an optional is bad as many things besides None are false (including 0!).

Issue #319, opened 2019-10-22
Improve shader support and configuration


aclegg3:
## Shader support:
Habitat currently supports flat shading and vertex colored rendering of scenes and objects assuming that lighting is baked into object textures. Many scene and object datasets do not include lighting information in their color, texture, and/or material and are therefore not rendered realistically under this assumption. 

Supporting shaders with illumination models will allow such assets to be rendered as intended and greatly increase overall asset compatibility and user experience.

## Shader Configuration:
Habitat sim presently relies on automatic parsing of asset features (such as filetype and the presence of materials/textures) to select shaders for rendering scenes and objects. With improved shader support, automation becomes more likely to fail as the number of diverging use cases increases. In addition to shader type, users may prefer to manually configure shader parameters such as light position, intensity, color, relative coordinate frame, etc... 

Supporting programmatically configurable shader setups would alleviate issues caused by incorrect shader configurations computed automatically. It would also provide greater support for use cases involving domain randomization, dynamic scene generation, and/or unknown/untested datasets.

## Short list of desired features:
- Support for Magnum Phong illumination shader
- Programatic configuration of shader parameters
- Configurable assignment of scenes/objects to shader groups (to override automatic assignment)

PR #454, opened 2020-01-28
Multiple drawable groups

matthewjmay:
## Motivation and Context

As part of the configurable Shader project, we need to have multiple DrawableGroups instead of using only 1. Each Drawable also needs a pointer to their drawable group in order to retrieve its shader at render time. This PR lays the groundwork for this to happen, while minimizing the number of changes by still using a single default group.

Part of #319 

## How Has This Been Tested

Need to add tests before merge (will update).

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/Drawable.h
+++ b/src/esp/gfx/Drawable.h
@@ -23,8 +23,6 @@ class DrawableGroupClient;
  * itself with the program.
  */
 class Drawable : public Magnum::SceneGraph::Drawable3D {
-  friend DrawableGroupClient;
-
  public:
   /**
    * @brief Constructor
@@ -40,6 +38,14 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
   virtual ~Drawable() {}
 
   virtual scene::SceneNode& getSceneNode() { return node_; }
+
+  /**
+   * @brief Get the @ref DrawableGroup this drawable is in.
+   *
+   * This overrides Magnum::SceneGraph::Drawable so that the derived @ref
+   * DrawableGroup can be used
+   */
+  DrawableGroup* drawables();
 
  protected:
   /**
@@ -57,18 +63,6 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
   scene::SceneNode& node_;
   Magnum::GL::AbstractShaderProgram& shader_;
   Magnum::GL::Mesh& mesh_;
-
-  DrawableGroup* group_ = nullptr;
-};
-
-/**
- * @brief Expose @Drawable group membership to @ref DrawableGroup
- */
-class DrawableGroupClient {
-  DrawableGroupClient() = delete;
-  static DrawableGroup* getGroup(Drawable& d) { return d.group_; }
-  static void setGroup(Drawable& d, DrawableGroup* g) { d.group_ = g; }
-  friend class DrawableGroup;
 };
 
 }  // namespace gfx
--- a/src/esp/gfx/Drawable.h
+++ b/src/esp/gfx/Drawable.h
@@ -14,13 +14,11 @@ class SceneNode;
 }
 namespace gfx {
 
-class DrawableGroupClient;
-
 /**
  * @brief Drawable for use with @ref DrawableGroup.
  *
- * Drawable will retrieve its shader program from its group, and draw
- * itself with the program.
+ * Drawable will retrieve its shader from its group, and draw
+ * itself with the shader.
  */
 class Drawable : public Magnum::SceneGraph::Drawable3D {
  public:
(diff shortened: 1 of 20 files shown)


src/esp/gfx/Drawable.h
```diff
@@ -13,25 +14,61 @@ class SceneNode;
...
+   * nothing more than setting up shader parameters and drawing the mesh.
+   */
   virtual void draw(const Magnum::Matrix4& transformationMatrix,
                     Magnum::SceneGraph::Camera3D& camera) = 0;
 
   scene::SceneNode& node_;
   Magnum::GL::AbstractShaderProgram& shader_;
   Magnum::GL::Mesh& mesh_;
+
+  DrawableGroup* group_ = nullptr;
```
mosra: This pointer is already available through [group()](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1AbstractGroupedFeature.html#a670e9a5ed8e81ed9ea5b819e3af01ee6), no need to store it again separately :)

matthewjmay: The issue here is that we need to store the shader and other settings common to all drawables in `DrawableGroup`, since a requirement is for `Drawables` to change groups at runtime. In other words, we need to extend `DrawableGroup` to have other methods (such as `prepareForDraw()`, `getShader()`, etc.).  We could downcast the result of `group()` to this inherited class, but we would either need a slow dynamic cast, or assume the Drawable is always used with this extended class, which I would rather be explicit about.

mosra: I see.

> We could downcast the result of group() to this inherited class

That's I think still cleaner than duplicating this state and then having to ensure the two states never get out of sync -- even with the above it's still possible to add the drawable to a different group than is expected, and then `parent::group() and `group_` become out of sync.

> slow dynamic cast

It could be only in an assert that gets compiled away for release builds, and `static_cast` otherwise. Also, with all the slow dynamic RTTI-based dispatch pybind is doing, I don't think this one particular occurence would be a problem .. and the compilers which implemented dynamic cast using slow string comparisons are fortunately long gone :)

matthewjmay: Haven't looked into it much, but would it be possible to add a 4th template parameter to `AbstractGroupedFeature`, which is by default `class Group = FeatureGroup<Dimensions, Derived, T>`. Haven't fully thought about the repercussions of this though.

matthewjmay: The other solution I see is to not make `DrawableGroup` inherit from `Magnum::DrawableGroup`, but instead have a Magnum group as a member. This would make it easier to ensure the states don't get out of sync. However, this makes rendering more difficult as `Camera::draw` needs a `DrawableGroup`, but I still think it's possible.

matthewjmay: I created a gist where I added a 4th template parameter to `AbstractGroupedFeature`. Only thing left to do is allow camera to use other group types: https://gist.github.com/matthewjmay/70b4e94354d25af47c2133105062b839

mosra: Why so complicated, tho? This significantly expands the API surface and would force many things (such as the whole camera class) to be inlined in headers now, slowing down compile times... The SceneGraph is already extremely over-templated, and I'm not a fan of making the problem even worse. What's wrong with just the following?

```cpp
DrawableGroup* drawables() {
    CORRADE_ASSERT(dynamic_cast<DrawableGroup*>(SceneGraph::DrawableGroup::drawables()),
        "someone messed up!", {});
    return static_cast<DrawableGroup*>(SceneGraph::DrawableGroup::drawables());
}
```

Are there realistically going to be multiple mutually incompatible drawable group subclasses or just one? If just one, why would anybody try to use the vanilla magnum one anyway? Nobody's using vanilla magnum Object either because there is habitat's SceneNode already. If there would be more than one, why would you need multiple of them?

matthewjmay: Sure, I'm fine with either approach. I was just thinking that this was a safe general and extensible solution for any user of Magnum who wanted to create their own group without casts in all of their derived drawables. Also, would it be possible to just template the `draw` functions of Camera? I also like it because it catches errors in adding Drawables to incompatible groups at compile rather than runtime.

I agree though, in Habitat there will likely not even be drawable group subclasses. For now if @bigbike and @aclegg3 agree we will do the `dynamic_cast` approach.

aclegg3: Having caught up on this thread, I think the casting approach sounds fine. As @mosra mentioned, we are already using Habitat specific Magnum wrapper subclasses for many other purposes. We'll want multiple `Drawable` and `Shader` subclasses, but we should only need one `DrawableGroup` subclass to implement the shader setup and compatibility checking. 

I think runtime validation is fine also, because we want dynamic shader configuration (e.g. for python users). Having checks when state changes (add/assign Drawable or Shader) rather than compile time is ideal in this case.

PR #456, opened 2020-01-28
[esp/bindings] fix a bunch of pybind11 doc reference issues

msbaines:
This does not fix all pybind11 doc errors. There are still a few
unresolved due to cross-references between some namespaces.

The cross-references can be avoided and I think the architecture
will benefit as a result. I hope to address this as part of the
sensor refactor.

Remaining errors:

WARNING:root:cannot parse pybind11 function signature bind_render_target(self: habitat_sim._ext.habitat_sim_bindings.Renderer, arg0: esp::sensor::VisualSensor) -> None
WARNING:root:cannot parse pybind11 function signature draw(self: habitat_sim._ext.habitat_sim_bindings.Renderer, visualSensor: esp::sensor::VisualSensor, scene: habitat_sim._ext.habitat_sim_bindings.SceneGraph) -> None
WARNING:root:cannot parse pybind11 function signature get_default_render_camera(self: habitat_sim._ext.habitat_sim_bindings.SceneGraph) -> esp::gfx::RenderCamera
WARNING:root:cannot parse pybind11 function signature set_default_render_camera_parameters(self: habitat_sim._ext.habitat_sim_bindings.SceneGraph, targetSceneNode: esp::sensor::VisualSensor) -> None
WARNING:root:cannot parse pybind11 function signature get_observation(self: habitat_sim._ext.habitat_sim_bindings.Sensor, arg0: esp::sim::Simulator, arg1: habitat_sim._ext.habitat_sim_bindings.Observation) -> bool
WARNING:root:cannot parse pybind11 function signature get_observation(self: habitat_sim._ext.habitat_sim_bindings.Sensor, arg0: esp::sim::Simulator, arg1: habitat_sim._ext.habitat_sim_bindings.Observation) -> bool
WARNING:root:cannot parse pybind11 function signature get_observation(self: habitat_sim._ext.habitat_sim_bindings.Sensor, arg0: esp::sim::Simulator, arg1: habitat_sim._ext.habitat_sim_bindings.Observation) -> bool

## Motivation and Context

Bring pybind11 doc errors down to zero.

## How Has This Been Tested

doc build

## Types of changes

### DIFF
--- a/src/esp/bindings/bindings.cpp
+++ b/src/esp/bindings/bindings.cpp
@@ -61,15 +61,18 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
 
   m.import("magnum.scenegraph");
 
+  py::bind_map<std::map<std::string, std::string>>(m, "MapStringString");
+
+  // NOTE(msb) These need to be run in dependency order.
+  // TODO(msb) gfx, scene, and sensor should not cross-depend
+  // TODO(msb) sim and sensor should not cross-depend
   esp::initEspBindings(m);
   esp::core::initCoreBindings(m);
   esp::geo::initGeoBindings(m);
-  esp::gfx::initGfxBindings(m);
-  esp::nav::initShortestPathBindings(m);
   esp::physics::initPhysicsBindings(m);
   esp::scene::initSceneBindings(m);
+  esp::gfx::initGfxBindings(m);
   esp::sensor::initSensorBindings(m);
+  esp::nav::initShortestPathBindings(m);
   esp::sim::initSimBindings(m);
-
-  py::bind_map<std::map<std::string, std::string>>(m, "MapStringString");
 }
--- a/src/esp/bindings/SceneBindings.cpp
+++ b/src/esp/bindings/SceneBindings.cpp
@@ -45,6 +45,39 @@ void initSceneBindings(py::module& m) {
                -> bool { return self != other; });
 
   // ==== SceneGraph ====
+
+  // !!Warning!!
+  // CANNOT apply smart pointers to "SceneNode" or ANY its descendant classes,
+  // namely, any class whose instance can be a node in the scene graph. Reason:
+  // Memory will be automatically handled in simulator (C++ backend). Using
+  // smart pointers on scene graph node from Python code, will claim the
+  // ownership and eventually free its resources, which leads to "duplicated
+  // deallocation", and thus memory corruption.
+
+  // ==== enum SceneNodeType ====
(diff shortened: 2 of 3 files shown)


src/esp/bindings/bindings.cpp
```diff
@@ -61,15 +61,18 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
...
   esp::core::initCoreBindings(m);
   esp::geo::initGeoBindings(m);
-  esp::gfx::initGfxBindings(m);
-  esp::nav::initShortestPathBindings(m);
   esp::physics::initPhysicsBindings(m);
   esp::scene::initSceneBindings(m);
+  esp::gfx::initGfxBindings(m);
   esp::sensor::initSensorBindings(m);
+  esp::nav::initShortestPathBindings(m);
   esp::sim::initSimBindings(m);
```
mosra: What could help with untangling is having a separate "declare bindings" step instantiating all the types, and then filling the `py::class` instances doesn't need to be ordered in any particular way.

msbaines: I looked into this and it was Erik's original suggestion. It would mean a lot more boiler plate code. I would prefer to fix the dependency graph (seems doable based on the current set of errors) and then enforce that the dependency graph remains a DAG. I've worked on a few project where we avoided cross-references and it worked out quite well and added a nice discipline. It forced people to think about proper modularity and architecture and resulted in a nice architecture that was easy to understand and reason about.

mosra: From my experience, even when exposing the basic Magnum math, circular references were inevitable due to templates no longer being templates -- in my case I wanted to expose integer/float vectors in separate compilation units to speed up build times, but there were circular deps coming from conversion constructors (Vector2i->Vector2 and Vector2->Vector2i).

msbaines: I agree. Circular references within a module are inevitable but I think circular references across modules can and should be avoided.

Regardless of which direction we go, I think this change is harmless and independent of what we decide. It's merely shuffling a few declarations around to put things in dependency order.

Approved?

msbaines: I'm not attached to having a DAG but if we can pull it off, we should. It's a nice property to have. If it becomes too restrictive, then we can drop the constraint.

mosra: Yes, I wasn't opposing this change at all, just sharing my own experience. Sorry if it sounded like a disproval.

PR #465, opened 2020-01-30
Update multi goal path implementation

erikwijmans:
## Motivation and Context

The multigoal path implementation was relying on being able to compare path lengths on the navigation mesh before optimization, this turns out to be not correct a little too often.  This PR changes it to be simply min(A* paths).

This also has the benefit of removing on our deviations from recast main!

## How Has This Been Tested

With the test!  Also added a benchmark because Corrade's test suite is fun.

## Types of changes

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -3,7 +3,6 @@
 // LICENSE file in the root directory of this source tree.
 
 #include "PathFinder.h"
-#include <numeric>
 #include <stack>
 #include <unordered_map>
 
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -882,7 +882,7 @@ float pathLength(const std::vector<vec3f>& points) {
 bool PathFinder::Impl::findPath(ShortestPath& path) {
   MultiGoalShortestPath tmp;
   tmp.requestedStart = path.requestedStart;
-  tmp.requestedEnds = {path.requestedStart};
+  tmp.requestedEnds = {path.requestedEnd};
 
   bool status = findPath(tmp);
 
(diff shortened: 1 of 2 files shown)


src/esp/nav/test/PathFinderTest.cpp
```diff
@@ -72,6 +78,67 @@ void PathFinderTest::tryStepNoSliding() {
...
+
+void PathFinderTest::pathFindBenchmark() {
+  esp::nav::PathFinder pathFinder;
+  pathFinder.loadNavMesh(skokloster);
+  CORRADE_VERIFY(pathFinder.isLoaded());
+
+  if (testCaseInstanceId() == 0) {
+    setTestCaseDescription("Path to single point");
+    esp::nav::ShortestPath path;
+    path.requestedStart = pathFinder.getRandomNavigablePoint();
```
mathfac: Does that make sense to calculate distance to the same 100 points using single shortest path search?

src/esp/nav/PathFinder.cpp
```diff
@@ -885,78 +950,35 @@ bool PathFinder::Impl::findPath(MultiGoalShortestPath& path) {
...
-
-  if (numPolys) {
-    const vec3f& closestRequestedEnd = path.requestedEnds[goalFoundIdx];
-
-    path.points.resize(MAX_POLYS);
-    status = navQuery_->findStraightPath(
-        path.requestedStart.data(), closestRequestedEnd.data(), polys, numPolys,
-        path.points[0].data(), 0, 0, &numPoints, MAX_POLYS);
+  for (int i = 0; i < path.requestedEnds.size(); ++i) {
+    if ((path.requestedStart - path.requestedEnds[i]).norm() >
```
mathfac: As I understand this check is selecting prospect end points that can be closer that `path.geodesicDistance`

mathfac:
Merging to unblock https://github.com/facebookresearch/habitat-api/pull/290.

src/esp/nav/PathFinder.cpp
erikwijmans: Yep, exactly.

PR #472, opened 2020-02-04
[Culling] optimization: do not erase invisible objects

bigbike:
We do not have to strictly conform to the "erase-remove" idiom. After std::remove_if is applied, we simply render the visible objects, which locate at the 1st part of the vector, while ignore the invisible ones, which are in the 2nd part. It can save a little bit for us.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/RenderCamera.cpp
+++ b/src/esp/gfx/RenderCamera.cpp
@@ -77,18 +77,22 @@ uint32_t RenderCamera::draw(MagnumDrawableGroup& drawables,
     return drawables.size();
   }
 
-  std::vector<std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>,
-                        Mn::Matrix4>>
-      drawableTransforms = drawableTransformations(drawables);
+  // dt is short for drawable transformations
+  using dt =
+      std::vector<std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>,
+                            Mn::Matrix4>>;
+  dt drawableTransforms = drawableTransformations(drawables);
 
   // draw just the visible part
   size_t numVisibles = cull(drawableTransforms);
-  // erase all items that did not pass the frustum visibility test
-  drawableTransforms.erase(drawableTransforms.begin() + numVisibles,
-                           drawableTransforms.end());
 
-  MagnumCamera::draw(drawableTransforms);
-  return drawableTransforms.size();
+  dt::iterator iter = drawableTransforms.begin();
+  for (size_t iDrawable = 0; iDrawable < numVisibles; ++iDrawable) {
+    iter->first.get().draw(iter->second, *this);
+    iter++;
+  }
+
+  return numVisibles;
 }
 
 }  // namespace gfx


src/esp/gfx/RenderCamera.cpp
```diff
@@ -77,18 +77,22 @@ uint32_t RenderCamera::draw(MagnumDrawableGroup& drawables,
...
   }
 
-  std::vector<std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>,
-                        Mn::Matrix4>>
-      drawableTransforms = drawableTransformations(drawables);
+  // dt is short for drawable transformations
+  using dt =
+      std::vector<std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>,
+                            Mn::Matrix4>>;
+  dt drawableTransforms = drawableTransformations(drawables);
```
bigbike: @mosra: drawableTransformations will create the vector from scratch every time? If we pre-allocate memory for the vector, and keep it in the entire cycle of the simulator, do you think it can save us time (in malloc)?

bigbike:
I am not confident if it will buy us sometime. Because you will have to delete these "invisible" objects sooner or later at each frame since the vector is "local".

My shallow opinion is that, if the vector is "global" and the memory is pre-allocated. Then it would work.

src/esp/gfx/RenderCamera.cpp
mosra: This is on my "get rid of STL containers in all places" TODO list :)

On the other hand, from my experiments when removing usage of STL containers elsewhere, `std::allocator` usually does some pool allocation, thus the allocation overhead is not as big in subsequent (de)allocations. And since this is done just once in a frame, not a big deal at all.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -77,18 +77,22 @@ uint32_t RenderCamera::draw(MagnumDrawableGroup& drawables,
...
-  drawableTransforms.erase(drawableTransforms.begin() + numVisibles,
-                           drawableTransforms.end());
 
-  MagnumCamera::draw(drawableTransforms);
-  return drawableTransforms.size();
+  dt::iterator iter = drawableTransforms.begin();
+  for (size_t iDrawable = 0; iDrawable < numVisibles; ++iDrawable) {
+    iter->first.get().draw(iter->second, *this);
+    iter++;
+  }
```
mosra: Does *this* actually improve anything? I doubt that, since `std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>, Mn::Matrix4>` has a trivial destructor and with optimizations enabled, `erase()` collapses to just adjusting vector size.

bigbike: This is just the consequence if I would like to get rid of the "erase()". 
Because if I do not use "erase()", the end() will still point to the old end position, and you will draw all the "invisible ones" at the 2nd half of the vector.

bigbike:
> "I don't think this improves anything. "

Yes, that is my concern as well (see my comments on Feb.4) because this temporary vector will get destroyed in every frame. I can put this PR temporarily in the waiting list.

src/esp/gfx/RenderCamera.cpp
mosra: You could use `erase()` even with a preallocated vector -- it won't free any memory, `capacity()` will stay the same after.

bigbike: If I use `erase()` it will call the destructor of the to-be-deleted items, which I tried to save.
It has nothing to do with the "deallocator".

bigbike: But like I said, it will not help you much since this is a local vector, and all the elements within will be destroyed (the destructor will be called) sooner or later.

erikwijmans: I am not sure if grouping up the destructor calls really matters.  The destructor will always need to run, even if it is a global vector.

bigbike: @erikwijmans : Aha, so even if this is a global vector, when we initialize it every frame, the "old value" from last frame will get destroyed no matter what. I see.

bigbike:
Drop this PR based on the discussion.

src/esp/gfx/RenderCamera.cpp
mosra: > it will call the destructor of the to-be-deleted items

@bigbike Just to reiterate what I said earlier and avoid confusion/misunderstanding -- the `std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>, Mn::Matrix4>` is a trivially-destructible type, which means there's no code at all to run at destruction, so the destructor calls will get optimized out.

PR #474, opened 2020-02-05
Maskrcnn experiment

mpiseno:
## Motivation and Context

The new data extraction API build on top of habitat-sim is a good way to extract image data from the simulator for training models in an end-to-end fashion. This PR sets up some of the training infrastructure for running semantic/instance segmentation models using the data extractor.

## How Has This Been Tested

Created a jupyter notebook to train a simple model on skoklostercastle scene.


erikwijmans:
Please fix the cpp formatting.

mpiseno:
@erikwijmans This PR is based on the Data Extraction API PR that I merged earlier today, but it still shows a lot of those files in this PR. I thought it would only show the files that differ from master? Also after running pre-commit, a lot of files I didn't mess with changed slightly.

erikwijmans:
You'll need to merge with master for slurm to update the diff.  

For the files that you changed, please check out the master versions of them.  I am not sure why your formatting is producing different things (my clang format is `clang-format version 8.0.0 (tags/RELEASE_800/final)` on MacOS). Anyways, the pre-commit hooks will only run on the files you changed by default (on each commit if you do `pre-commit install .`), there is no need to manually run them on all files.

PR #478, opened 2020-02-07
[esp/sim] collapse SimulatorWithAgents into Simulator

msbaines:
Part of Sensor refactor.

## Motivation and Context

Simplify by having one common Simulator class.

## How Has This Been Tested

SimTest and WebGL

## Types of changes

## Checklist


src/esp/sim/Simulator.cpp
```diff
@@ -51,6 +51,20 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
     sceneFilename = cfg.scene.filepaths.at("mesh");
   }
 
+  // create pathfinder and load navmesh if available
```
aclegg3: Some of this is now duplicated on [the python side](https://github.com/facebookresearch/habitat-sim/blob/6deb47b45e4e4548378d82f733aec2f8913e2c12/habitat_sim/simulator.py#L163). Could be a separate PR to refactor if everything is still functioning.

src/esp/sim/Simulator.cpp
```diff
@@ -178,10 +192,21 @@ void Simulator::reset() {
   if (physicsManager_ != nullptr)
     physicsManager_
         ->reset();  // TODO: this does nothing yet... desired reset behavior?
+
+  for (int iAgent = 0; iAgent < agents_.size(); ++iAgent) {
+    auto& agent = agents_[iAgent];
+    agent::AgentState::ptr state = agent::AgentState::create();
+    sampleRandomAgentState(state);
```
aclegg3: This will now conflict with reset behavior in python as of #435 (reset to initial vs. random).

src/esp/sim/Simulator.cpp
```diff
@@ -389,5 +414,134 @@ bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
...
+      renderer_->bindRenderTarget(*sensor);
+    }
+  }
+
+  agents_.push_back(ag);
+  // TODO: just do this once
+  if (pathfinder_->isLoaded()) {
+    ag->getControls()->setMoveFilterFunction(
+        [&](const vec3f& start, const vec3f& end) {
+          return pathfinder_->tryStep(start, end);
```
aclegg3: This won't be synchronized with python move filter functions.

src/esp/sim/Simulator.h
```diff
@@ -326,6 +359,9 @@ class Simulator {
   core::Random random_;
   SimulatorConfiguration config_;
 
+  std::vector<agent::Agent::ptr> agents_;
```
aclegg3: This structure and related functions will not be synchronized with python agents.

src/esp/sim/Simulator.cpp
msbaines: Fixed.

src/esp/sim/Simulator.cpp
msbaines: Python code doesn't use this pathfinder. Cool if we address this later when python is modified to use this pathfinder?

src/esp/sim/Simulator.h
msbaines: Yes. Plan was to merge the class first and then merge the functionality. This vector is unused in python since it never uses the C++ addAgent API. Plan is to migrate python over to this API.

erikwijmans: Some of the functionality will either need to stay in python or will need a bridge or some sort into python.  The ability to add actions in python has been quite useful for instance.

msbaines: The public python API should not change but the internals move to C++. I think that could be accomplished with bridging. We should also support the ability to extend habitat via python. This can accomplished by supporting python sub-classing. One example use case is to create your own Sensor class. 

Overall, I think we should be able to support all current python use cases with bindings and python sub-classing support of some of the C++ classes. I'm hoping none of our users (other than extreme power users) will even notice the change.

But I think it will be extremely valuable to have the core of the simulator in one language instead of 3.

src/esp/sim/Simulator.cpp
aclegg3: Sounds good.

aclegg3:
Can we land this? Would be nice to unify these now and start the next step of refactoring.

bigbike:
It looks good to me.

@mosra: would you like to take a review pass?

PR #481, opened 2020-02-09
[Culling][Optimization] Fustum culling with temporal coherence

bigbike:
Let the simulator memorize the frustum plane index, which culls the object in the last frame.
In the current frame, it is likely the object will be rejected again by this plane. Thus, start the culling from this plane first.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/RenderCamera.cpp
+++ b/src/esp/gfx/RenderCamera.cpp
@@ -23,7 +23,7 @@ struct CullingResult {
 };
 
 /**
- * @brief do frustum culling with temporal coherency
+ * @brief do frustum culling with temporal coherence
  * @param range, the axis-aligned bounding box
  * @param frustum, the frustum
  * @param frustumPlaneIndex, the frustum plane in last frame that culled the
--- a/src/esp/gfx/RenderCamera.cpp
+++ b/src/esp/gfx/RenderCamera.cpp
@@ -22,6 +22,8 @@ namespace gfx {
  * @param frustum, the frustum
  * @param frustumPlaneIndex, the frustum plane in last frame that culled the
  * aabb (default: 0)
+ * @return NullOpt if aabb intersects the frustum, otherwise the fustum plane
+ * index that culls the aabb,
  */
 Cr::Containers::Optional<int> rangeFrustum(const Mn::Range3D& range,
                                            const Mn::Frustum& frustum,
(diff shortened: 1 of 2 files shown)


src/esp/gfx/RenderCamera.cpp
```diff
@@ -15,6 +15,41 @@ namespace Mn = Magnum;
...
+};
+
+/**
+ * @brief do frustum culling with temporal coherence
+ * @param range, the axis-aligned bounding box
+ * @param frustum, the frustum
+ * @param frustumPlaneIndex, the frustum plane in last frame that culled the
+ * aabb (default: 0)
+ */
+CullingResult rangeFrustum(const Mn::Range3D& range,
```
erikwijmans: I'd prefer using an https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1Optional.html here.

bigbike: Can I ask why?

erikwijmans: Doesn't requiring introducing an additional type.

bigbike: I see. You were actually talking the return type. I thought you were referring the arguments.

That is a good idea.

erikwijmans: Ah, apologies for lack of clarity.

bigbike: Not a problem at all. Good advice, makes the code simpler. Will change.

bigbike:
Dear reviewers,

Today, I designed an experiment to measure the performance gain from leveraging temporal coherence (denoted as TC). 

The test code can be found [here](https://github.com/facebookresearch/habitat-sim/compare/culling-test_op2...op2-eval).

To evaluate the performance, I designed 2 metrics:
- ratio between the number of culled objects that benefit from TC and the total number of culled objects;
- ratio between the number of frustum plane tests when TC is turned ON / OFF;

The result is as follows:
```
 ========================= Performance ========================
 256 x 256, total time 7.45 s, frame time 1.863 ms (536.8 FPS)
 ==============================================================
###### The efficiency of temporal coherence ######
culled objects that benefit from temporal coherency / total culled objects: 1096767 / 1115010, ratio: 98.36%
plane tests with and without temporal coherence enabled: 12560284283 / 14545762971, SAVE: 13.65%
```

**Conclusions:**
- Temporal coherence almost exists in every culled object (**~99%**);
- With temporal coherence heuristic, it roughly saves **10%** visibility tests;



The command line I used:
`python examples/example.py --scene ~/models/basis-mp3d/dhjEzFoUFzH/dhjEzFoUFzH.glb --width 256 --height 256 --max_frames 4000 --silent --seed 5`

mosra:
This is good! :+1:

PR #483, opened 2020-02-10
Dataset extraction API

mpiseno:
## Motivation and Context

Added dataset extraction API to habitat-sim that allows users to specify a scene file and get back a bunch of images that can easily be fed into a PyTorch model or saved for later use. The API supports RGBA, depth, and semantic images as well as allows the user to specify image size.

## How Has This Been Tested

Created two tests in tests/test_dataset_extraction.py that:

1. Create a topdown view
2. input a scene file to the data extractor and feed it through a pytorch model

## Types of changes

## Example Output

Screen Shot 2020-02-13 at 3 18 02 PM
Screen Shot 2020-02-13 at 3 12 45 PM
Screen Shot 2020-02-13 at 3 12 54 PM
Screen Shot 2020-02-13 at 3 13 08 PM
Screen Shot 2020-02-13 at 3 13 19 PM


src/esp/nav/PathFinder.cpp
```diff
@@ -229,6 +231,8 @@ struct PathFinder::Impl {
 
   std::pair<vec3f, vec3f> bounds() const { return bounds_; };
 
+  std::vector<std::vector<bool>> getTopDownView(const float res);
```
erikwijmans: This should be an `Eigen::Matrix<bool, Eigen::Dynamic, Eigen::Dynamic>` matrix -- that'll get sent over to python as a numpy array with zero copy :)

src/esp/nav/PathFinder.cpp
```diff
@@ -1128,6 +1132,42 @@ bool PathFinder::Impl::isNavigable(const vec3f& pt,
   return true;
 }
 
+std::vector<std::vector<bool>> PathFinder::Impl::getTopDownView(
+    const float res) {
```
erikwijmans: The caller should be responsible for providing a height.

src/esp/nav/PathFinder.cpp
erikwijmans: Let's give `res` a more intelligible name.  Seems like it is being used as `metersPerPixel`?

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,283 @@
...
+        else:
+            plt.imshow(self.topdown_view, cmap=cmap)
+
+        plt.show()
+
+    def _get_pathfinder_reference_point(self):
+        bound1, bound2 = self.pathfinder.get_bounds()
+        startw = min(bound1[0], bound2[0])
+        starth = min(bound1[2], bound2[2])
+        starty = self.pathfinder.get_random_navigable_point()[
```
erikwijmans: If you have the caller send in the height for generating the top-down map, then you can get this exactly.

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,283 @@
...
+import os
+
+import matplotlib.pyplot as plt
+import numpy as np
+import torch
+from torch.utils.data import Dataset
+
+import habitat_sim
+import habitat_sim.bindings as hsim
+from examples.settings import make_cfg
```
erikwijmans: We should not be pulling something in from examples into the core of habitat-sim.

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,283 @@
...
+                if i < len(self.poses)
+            ]
+
+        pos, rot, label = self.poses[idx]
+        new_state = AgentState()
+        new_state.position = pos
+        new_state.rotation = rot
+        self.sim.agents[0].set_state(new_state)
+        obs = self.sim.get_sensor_observations()
+        sample = {
```
erikwijmans: Color is output as RGB-A, so you'll want to either chop-off the alpha channel or change the name.

src/esp/nav/PathFinder.cpp
```diff
@@ -1128,6 +1132,42 @@ bool PathFinder::Impl::isNavigable(const vec3f& pt,
...
+  float navigableHeight = getRandomNavigablePoint()[1];
+  std::vector<std::vector<bool>> topdownMap;
+
+  float curHeight = starty;
+  float curWidth = startx;
+  for (int h = 0; h < heightResolution; h++) {
+    std::vector<bool> curRow;
+    for (int w = 0; w < widthResolution; w++) {
+      vec3f point = vec3f(curWidth, navigableHeight, curHeight);
+      if (isNavigable(point, 0.5)) {
```
erikwijmans: This will need to change for the Eigen matrix, but anyways, you can do `curRow.push_back(isNavigable(point, 0.5))` here.

mpiseno:
I had to force-push to fix some difficult errors that caused api tests to fail. Will address the comments above @erikwijmans thanks for the feedback.

requirements.txt
```diff
@@ -5,3 +5,5 @@ numpy-quaternion
 pillow
 scipy>=1.3.0
 tqdm
+torch
```
dhruvbatra: I think we've avoided adding torch as a dependency so far. @erikwijmans -- thoughts?

mpiseno:
Fixed everything you gave feedback for @erikwijmans thank you for the help!

requirements.txt
erikwijmans: Yeah, we should continue to avoid it as a hard dependency.  @mpiseno we can have the dataset extractor class be API-compatible and then mix-in a torch dataset in the training examples or something.   It could also be a soft dependency (i.e. you will only need it if you want to use this code).

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,371 @@
...
+        bound1, bound2 = pf.get_bounds()
+        startw = min(bound1[0], bound2[0])
+        starth = min(bound1[2], bound2[2])
+        starty = pf.get_random_navigable_point()[
+            1
+        ]  # Can't think of a better way to get a valid y-axis value
+        return (startw, starty, starth)  # width, y, height
+
+
+class TopdownView(object):
```
erikwijmans: What does this class do?  Seems like it's just a helper function?

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,371 @@
...
+    def __init__(self, sim, height, pixels_per_meter=0.1):
+        self.topdown_view = np.array(
```
erikwijmans: Should no longer need the `np.array` here.

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,371 @@
...
+            "scene": scene_filepath,  # Scene path
+            "default_agent": 0,
+            "sensor_height": 1.5,  # Height of sensors in meters
+            "color_sensor": True,  # RGBA sensor
+            "semantic_sensor": True,  # Semantic sensor
+            "depth_sensor": True,  # Depth sensor
+            "silent": True,
+        }
+
+        return make_cfg(sim_settings)
```
erikwijmans: Using the dictionary and helper function doesn't really seem necessary here.  Just making the config in this method would be cleaner IMO.

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,371 @@
...
+        startw = min(bound1[0], bound2[0])
+        starth = min(bound1[2], bound2[2])
+        starty = self.pathfinder.get_random_navigable_point()[
+            1
+        ]  # Can't think of a better way to get a valid y-axis value
+        return (startw, starty, starth)  # width, y, height
+
+    def _compute_quat(self, cam_normal):
+        """Rotations start from -z axis"""
+        return quat_from_two_vectors(np.array([0, 0, -1]), cam_normal)
```
erikwijmans:
```suggestion
        return quat_from_two_vectors(habitat_sim.geo.FRONT, cam_normal)
```

requirements.txt
mpiseno: The only think I actually use pytorch for is in the tests in tests/test_dataset_extraction. I could remove this without any changes to the ImageExtractor class, but I think @mathfac wanted to run the data through a pytorch model in testing?

habitat_sim/utils/data/dataextractor.py
mpiseno: Right now it is essentially a helper function. @msbaines thought the topdown functionality could be useful in other contexts in the future, which is why I made it separate.

habitat_sim/utils/data/dataextractor.py
mpiseno: For the next PR im working on that depends on this one, I need to call this method from multiple classes, but for this branch it does seem cleaner to do what you suggested.

requirements.txt
erikwijmans: That's fine, we have a soft dependency on pytorch for some tests already (--with-cuda relies on torch in places), but that soft dependency doesn't need to be in requirements.txt

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,181 @@
...
+            for out_name in self.output
+        }
+        sample["label"] = self.label_map[label]
+
+        return sample
+
+    def close(self):
+        r"""Deletes the instance of the simulator. Necessary for instatiating a different ImageExtractor.
+        """
+        self.sim.close()
```
erikwijmans: Add an `self.sim = None` bellow and then guard all this with `if self.sim is not None:` that way you can call close twice.

requirements.txt
mpiseno: Is it sufficient to simply remove torch from requirements.txt? Will the soft dependency during testing be handled appropriately even if i import torch in my tests/test_dataset_extraction.py file?

erikwijmans: Yeah, just removing it from requirements.txt is sufficient.  All of a "soft" dependency is in these terms is something you don't list in requirements.txt and the user will only need to install if they actually want to run certain code (as opposed to just importing the function).

src/esp/nav/PathFinder.cpp
```diff
@@ -1222,5 +1258,17 @@ std::pair<vec3f, vec3f> PathFinder::bounds() const {
   return pimpl_->bounds();
 }
 
+// std::vector<std::vector<bool>> PathFinder::getTopDownView(
```
erikwijmans: Remove.

habitat_sim/utils/data/poseextractor.py
```diff
@@ -0,0 +1,192 @@
+import collections
```
erikwijmans: snake_case filename to match the rest of the codebase.

habitat_sim/utils/data/dataextractor.py
```diff
@@ -0,0 +1,183 @@
+import math
```
erikwijmans: Same request for filename.

examples/Image-Data-Extraction-API.ipynb
```diff
@@ -0,0 +1,234 @@
+{
```
erikwijmans: We banished notebooks a while ago as they don't play nice with git, CC @abhiskk

dhruvbatra: I had asked Michael to make one so that it serves as a tutorial. What would be better if not a notebook?

erikwijmans: We banished them from git, they still exist, just not in git.

mpiseno: Oleksandr said I should make a tutorial on the habitat website that reflects the same information

mpiseno:
I believe @mathfac said the 1 test that is currently failing is a known issue and if everything else is passing then this PR should be good. If that's the case, I'm just waiting on PR approval to merge.

PR #486, opened 2020-02-11
RigidObject velocity control

aclegg3:
## Motivation and Context

This change introduces the `VelocityControl` class for setting constant linear|angular control velocities for `RigidObject`s. The `PhysicsManager::stepPhysics()` and derived variants consume this class to either set initial simulation velocities (MotionType::DYNAMIC), or integrate velocities to set new object states (MotionType::KINEMATIC). 

The `VelocityControl::linVelIsLocal` variable allows the user to define local linear velocities in order to implement commands such as "forward" (e.g. `[0.0,0.0,-1.0]`) and "back" easily. In this case the linear velocity vector is rotated by the object's orientation transform before application.

Kinematic integration is done with `VelocityControl::integrateTransform()` which implements explicit Euler integration native to Habitat (i.e. does not require Bullet). This can be overrode in the future to introduce new integrators if necessary. Note: this form of velocity control is not constrained by collision.

## How Has This Been Tested

New C++ unit tests and viewer demo testing (not included in this change):

Example with `MotionType::DYNAMIC`:
velcontrol_dynamic 2020-02-10 14_10_24

Example with `MotionType::KINEMATIC`:
velcontrol_kinematic 2020-02-10 14_12_59

Example with random angular velocity (to demonstrate 3D flight):
velcontrol_kin_flying 2020-02-10 14_14_24

## Types of changes

## Checklist


src/esp/physics/RigidObject.cpp
```diff
@@ -213,5 +213,39 @@ Magnum::Matrix3 RigidObject::getInertiaMatrix() {
...
+  Magnum::Matrix3 newRotationScaling = objectTransform.rotationScaling();
+  // then angular
+  if (controllingAngVel) {
+    Magnum::Quaternion q;
+    Magnum::Vector3 ha = angVel * dt * 0.5;  // vector of half angle
+    float l = ha.length();                   // magnitude
+    if (l > 0) {
+      float ss = std::sin(l) / l;
+      q = Magnum::Quaternion(
+          Magnum::Vector3{ha.x() * ss, ha.y() * ss, ha.z() * ss}, std::cos(l));
```
mosra:
```suggestion
          ha*ss, std::cos(l));
```

Same below.

Though ... to me this looks like an unfinished derivation, but can't quite point the finger on where exactly. Isn't it possible to get the quat out of it in a more direct way, without having to do the square root and trig operations on something that already was a scaled direction vector?

erikwijmans: Seems like this is just converting an angle-axis into a quaternion, and thus you can use: https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Quaternion.html#accba70d55affe884d8086110a130475a ?

mosra: I wanted to suggest that at first, but even that didn't feel right... how's the angle actually represented in `angVel`? Is it the direction or the magnitude? (Sorry if my question is dumb :sweat_smile:)

erikwijmans: Yeah, angle-axis representation is another way to specify a log quaternion.  Looks like `angVel * dt` is a log quaternion, and this code is take the exponential of it to get the correspond unit quaternion (aka versor).  `Mn::Quaternion::rotation` also computes the exponential of a log quaternion, it just takes it in angle-axis format -- aka magnitude and direction format.

aclegg3: Thanks for pointing this out. Much more concise. 😄

aclegg3:
> Haha, looking at the animations at first I thought you were hooking up some easing curves from the [Easing namespace](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Animation_1_1Easing.html) :)

Does seem that way, haha. Thanks for pointing those out, I'll have to take a look. We've been considering some curve utilities. 👍

src/esp/physics/PhysicsManager.cpp
```diff
@@ -189,6 +189,16 @@ void PhysicsManager::stepPhysics(double dt) {
   double targetTime = worldTime_ + dt;
   while (worldTime_ < targetTime) {
     // per fixed-step operations can be added here
+
+    // kinematic velocity control intergration
+    for (auto& object : existingObjects_) {
+      VelocityControl& velControl = object.second->getVelocityControl();
```
bigbike: Usually `std::pair` will only be used in a small area where the definition and application are not too "far away" to each other. This is for readability consideration. Better solution here is to use a `struct`.
Not necessary to change in this PR as it is beyond the scope of this PR.

src/esp/physics/RigidObject.h
```diff
@@ -80,6 +81,47 @@ enum class RigidObjectType {
 
 };
 
+/**@brief Convenience class for applying constant velocity control to a rigid
+ * body. */
+class VelocityControl {
+ public:
+  /**@brief Constant linear velocity. */
+  Magnum::Vector3 linVel;
```
bigbike: linVel_

src/esp/physics/RigidObject.h
```diff
@@ -80,6 +81,47 @@ enum class RigidObjectType {
...
+  /**@brief Constant angular velocity. */
+  Magnum::Vector3 angVel;
+  /**@brief Whether or not to set linear control velocity before stepping. */
+  bool controllingLinVel = false;
```
bigbike: underscore should be added everywhere.

src/esp/physics/RigidObject.h
```diff
@@ -80,6 +81,47 @@ enum class RigidObjectType {
 
 };
 
+/**@brief Convenience class for applying constant velocity control to a rigid
+ * body. */
+class VelocityControl {
```
bigbike: use `struct`, since all the variables are public.

src/esp/physics/RigidObject.cpp
```diff
@@ -213,5 +213,35 @@ Magnum::Matrix3 RigidObject::getInertiaMatrix() {
   return inertia;
 }
 
+//////////////////
+// VelocityControl
+Magnum::Matrix4 VelocityControl::integrateTransform(
```
bigbike: @mosra: have you read this functions? 
Any advanced matrix manipulation from Magnum library can kick in to improve it?

src/esp/physics/PhysicsManager.cpp
aclegg3: This `second` is actually coming from a std::map iterator for `existingObjects_`, not a std::pair.

src/esp/physics/RigidObject.h
mosra: Those are all public members, isn't the guideline to have only privates underscored?

src/esp/physics/RigidObject.cpp
mosra: [See my earlier comments.](https://github.com/facebookresearch/habitat-sim/pull/486#discussion_r378325455)

src/esp/physics/RigidObject.h
bigbike: Well, this is a good point here. Good catch.

My understanding is that by default all the variables in the class should not be public, and thus have the underscore.

Since I suggested using the struct, the variables here are public. That means the underscore is not needed.

src/esp/physics/RigidObject.cpp
```diff
@@ -213,5 +213,35 @@ Magnum::Matrix3 RigidObject::getInertiaMatrix() {
...
+  Magnum::Matrix3 newRotationScaling = objectTransform.rotationScaling();
+  // then angular
+  if (controllingAngVel) {
+    Magnum::Vector3 globalAngVel = angVel;
+    if (angVelIsLocal) {
+      globalAngVel = objectTransform.rotation() * angVel;
+    }
+    Magnum::Quaternion q = Magnum::Quaternion::rotation(
+        Magnum::Rad{(globalAngVel * dt).length()}, globalAngVel.normalized());
+    newRotationScaling = q.toMatrix() * newRotationScaling;
```
bigbike: Just be careful.
It may cause significant numerical drift here (even with "double"). Not so sure if it would an issue in the future.

mosra: One option to consider in the future might be switching to [TranslationRotationScalingTransformation3D](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1SceneGraph.html#a5af7f3fa56af8a3e906490e63562c8a9) that stores rotation as a quat and thus could allow for faster and more precise calculation without having to go through matrices. However extra care has to be taken as relative order of the T/R/S transformations is different than with `MatrixTransformation3D`.

Another option might be [DualQuaternionTransformation](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1SceneGraph.html#ace6339f6ce9e8983c8d02c7219419f3d) but while it reduces the amount of ALU ops even further, it is more restrictive as there's no shear or scaling.

bigbike: Storing and time stepping rotation using a unit quaternion, or exponential map is the typical way we do in rigid body simulation.

Applying DualQuaternion method to handle rigid-body dynamics is rare. It is initially introduced by Kavan to computer animation to avoid the artifacts shown in linear skinning.

mosra: > rare

Yes, I know it got first widely used for skinning, but that's not relevant in this case. Dual quats are, at its core, an efficient way to store translation+rotation together, so why not use it like that. The only thing related to animation/skinning is the DLB operation, and that's not used here.

aclegg3: Thanks for the good discussion on state representation. We should continue to keep an eye on our use cases moving forward. If we find that our current representation is lacking in speed or accuracy we can revisit the alternatives. 👍

PR #489, opened 2020-02-12
Enable phong shading and lighting configurations

matthewjmay:
## Motivation and Context

This PR enables phong shading by default for added objects (scene remains flat shaded), and sets up the framework for multiple light configurations to be used, another step toward #319.

### Without Phong (before)
no_phong_shading

### With Phong
with_phong

The default light configuration for dynamically added objects consists of lights at the corners of the scene bounding box.

In later PRs, the following functionality will be exposed through the simulator
- selecting light configurations to use for added/existing objects
- updating existing light configurations

## How Has This Been Tested

- End to end test in `SimTest.GetPinholeCameraRGBAObservation` shows no changes to images rendered with no lighting. Once lighting selection is exposed through simulator, I will add tests for different phong lighting configurations 
- Visual test using viewer shows phong lighting works!

#### Scene Only

Tested locally with:
`python examples/benchmark.py --scene ~/habitat-sim/data/scene_datasets/habitat-test-scenes/skokloster-castle.glb  --max_frames 4000 --resolution=256 --num_procs=1`

On branch master:
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution      rgb             rgbd            depth_only
256 x 256       438.2           249.0           462.8
 ==============================================================================
```

On this branch:
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution      rgb             rgbd            depth_only
256 x 256       442.8           225.7           425.5
 ==============================================================================
```

Should have no impact since scene is rendered with no lights

#### With Phong Shaded Physics objects

Tested locally with:
`python examples/benchmark.py --enable_physics --scene ~/habitat-sim/data/scene_datasets/habitat-test-scenes/skokloster-castle.glb  --max_frames 4000 --resolution=256 --num_procs=1`

On branch master:
```
 ================ Performance (FPS) NPROC=1 ===================================
Resolution      rgb             rgbd            depth_only      phys_rgb        phys_rgbd
256 x 256       484.9           232.9           454.0           122.0           116.7
 ==============================================================================
 ================ Performance (step time: milliseconds) NPROC=1 ===================================
Resolution      rgb             rgbd            depth_only      phys_rgb        phys_rgbd
256 x 256       0.00            0.00            0.00            0.57            0.50
 ==============================================================================
```

On this branch:
```
 ====== FPS (256 x 256, phys_rgbd): 119.0 ======
 ================ Performance (FPS) NPROC=1 ===================================
Resolution      rgb             rgbd            depth_only      phys_rgb        phys_rgbd
256 x 256       441.8           231.4           452.3           106.8           119.0
 ==============================================================================
 ================ Performance (step time: milliseconds) NPROC=1 ===================================
Resolution      rgb             rgbd            depth_only      phys_rgb        phys_rgbd
256 x 256       0.00            0.00            0.00            0.82            0.49
 ==============================================================================
```

Negligible impact. 

## Types of changes

## Checklist


dhruvbatra:
Great job on the benchmarks.

src/esp/gfx/ShaderManager.h
```diff
@@ -0,0 +1,21 @@
...
+#include <Magnum/GL/AbstractShaderProgram.h>
+#include <Magnum/ResourceManager.h>
+
+#include "esp/gfx/LightSetup.h"
+#include "esp/gfx/MaterialData.h"
+
+namespace esp {
+namespace gfx {
+
+using ShaderManager = Magnum::ResourceManager<Magnum::GL::AbstractShaderProgram,
```
erikwijmans: `Magnum::ResourceManager` is a true singleton (lifetime tied to the program itself, not an instance of the simulator). We currently make assumptions that a true singleton violates -- that all opengl resources are cleared when the simulator is killed. This is why that test is failing.

(We should add an explicit test for being able to construct/destruct the simulator multiple times as that one is checking something specific and won't catch everything involved).

matthewjmay: @mosra Any thoughts? Really would not like to have to reimplement the `ResourceManager` logic if we could avoid it. Perhaps an option could be added on the Magnum side?

matthewjmay: After diving into the Magnum code, the singleton functionality of `ResourceManager` is actually deprecated. I turned off the `BUILD_DEPRECATED` build flag (and minor related build issues) and it fixes the issue.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -4,43 +4,106 @@
...
+  if (materialData_->diffuseTexture)
+    shader_->bindDiffuseTexture(*(materialData_->diffuseTexture));
+  if (materialData_->specularTexture)
+    shader_->bindSpecularTexture(*(materialData_->specularTexture));
+
+  const Magnum::Matrix4 cameraMatrix = camera.cameraMatrix();
+  for (Magnum::UnsignedInt i = 0; i < lightSetup_->size(); ++i) {
+    shader_->setLightPosition(
+        i, getLightPositionRelativeToCamera(
+               (*lightSetup_)[i], transformationMatrix, cameraMatrix));
```
mosra: To reduce the number of GL calls it might be better to first calculate all positions to a pre-sized array and then call `setLightPositions()` / `setLightColors()` for all lights at once.

src/esp/gfx/LightSetup.cpp
```diff
@@ -0,0 +1,44 @@
...
+    const LightInfo& light,
+    const Magnum::Matrix4& transformationMatrix,
+    const Magnum::Matrix4& cameraMatrix) {
+  if (light.model == LightPositionModel::OBJECT)
+    return transformationMatrix.transformPoint(light.position);
+
+  if (light.model == LightPositionModel::GLOBAL)
+    return cameraMatrix.transformPoint(light.position);
+
+  // LightPositionModel::CAMERA
```
mosra: `CORRADE_INTERNAL_ASSERT()` this? Would ensure this function doesn't get forgotten when adding new light position models.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1197,12 +1166,86 @@ void ResourceManager::loadMaterials(Importer& importer,
...
+}
+
+gfx::PhongMaterialData::uptr ResourceManager::getPhongShadedMaterialData(
+    const Mn::Trade::PhongMaterialData& material,
+    int textureBaseIndex) {
+  // NOLINTNEXTLINE(google-build-using-namespace)
+  using namespace Mn::Math::Literals;
+
+  auto finalMaterial = gfx::PhongMaterialData::create_unique();
+  // TODO: figure out why materials are being loaded with shininess == 1
```
mosra: Which file format? I know Assimp was doing some shady business for certain formats :)

src/esp/gfx/ShaderManager.h
mosra: @erikwijmans hmm, no, the ResourceManager isn't a *true singleton* -- as with all other globals in Magnum, the application has the full control over its lifetime. I deprecated it for slightly different reasons -- it's hard/impossible to have a truly unique global instance pointer for a template class and because all attempts to mitigate this problems in a generic way only made it worse, I decided to move this responsibility on the users instead.

I'd say the true cause of abort was due to two instances of the resource manager existing at the same time -- is that the case in tests? (It should print that on the error output, if you run `pytest` in a certain way I think.) And if so, could that be avoided, for the time being at least? That check isn't present anymore when you disable `BUILD_DEPRECATED`, but I'd like to keep that flag on since it makes upgrades *a lot* smoother. I'd like to remove this deprecated feature soon, but unfortunately various users still rely on this feature so I can't get rid of it just yet.

src/esp/geo/geo.cpp
```diff
@@ -2,6 +2,9 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+#include <Corrade/Containers/ArrayViewStl.h>
+#include <Magnum/Math/FunctionsBatch.h>
+
```
aclegg3: What are these additions for? They don't seem to be needed.

src/esp/gfx/ShaderManager.h
aclegg3: Agreed, we should be avoiding multiple instances in one process. This should be the case while `Magnum::ResourceManager` is a member of `esp::assets::ResourceManager`. Is it true that we try to instantiate this twice in some existing test?

I'm fine with adding an explicit test for closing and re-initializing the simulator multiple times, this has been an issue in the past for other changes.

src/esp/gfx/LightSetup.cpp
```diff
@@ -0,0 +1,44 @@
...
+  return light.position;
+}
+
+LightSetup getLightsAtSceneCorners(scene::SceneGraph& sceneGraph,
+                                   Magnum::Float lightIntensity) {
+  const Magnum::Range3D& sceneBB =
+      sceneGraph.getRootNode().computeCumulativeBB();
```
aclegg3: If we intend to expose this to the user or call it again after initialization we may want to pre-compute this bounding box after initial load and pass it in instead. `computeCumulativeBB()` on the root will recurse the full SceneGraph. This is fine to do once, but could be trouble in some edge cases (e.g. an object falls through a hole and stretches the root bounding box on recompute or a user calls this function every other frame for some reason.)

This could be `getLightsAtBoxCorners` or similar and take a `Magnum::Range3D` instead. Would let users place lights for arbitrary bounding regions. Could include an optional (default Identity) transformation parameter also to provide non-axis aligned box support.

src/esp/geo/geo.cpp
mosra: This is I think related to the disabled `BUILD_DEPRECATED` flag -- I got rid of some inter-header dependencies for faster compile times but kept the includes for backwards compatibility when the flag is disabled.

src/esp/gfx/ShaderManager.h
matthewjmay: We are instantiating multiple instances of `Simulator` in the **same process**, which in turn causes two instances of `ResourceManager` in this test:

https://github.com/facebookresearch/habitat-sim/blob/d281a086d7cc24a4269276c21d6c55e3aa6c989f/tests/test_sensors.py#L130-L143

We need to decide if this a requirement or not @erikwijmans @aclegg3.

erikwijmans: > hmm, no, the ResourceManager isn't a true singleton -- as with all other globals in Magnum, the application has the full control over its lifetime.

@mosra That's great!  I poked around it only a little bit.

I am fine with only allowing one instance of the simulator at a time (this is already a requirement when you have any OpenGL resources).  @matthewjmay can you put a skip on the current test that is failing and some kind of comment/message to take that skip away once @mosra deprecates that officially?

I also added an explicit test for multiple construct/destructs here: https://github.com/facebookresearch/habitat-sim/pull/490

mosra: Is the simulator instantiating the GLContext as well? Because that one has a similar assert check (allowing only one instance to be "current" at a time), so I wonder why that one isn't triggering also in this case. Or is the GLContext external to the simulator (and then maybe ResourceManager could follow suit) or some extra `makeCurrent()` calls done to it?

erikwijmans: In the specific case of that test, it isn't.  That test is instating the simulator for a "blind" agent (which shouldn't load opengl things since its blind and we don't need to render for it).

matthewjmay: I think we have an explicit `try_create` for the context:
https://github.com/facebookresearch/habitat-sim/blob/d281a086d7cc24a4269276c21d6c55e3aa6c989f/src/esp/gfx/WindowlessContext.cpp#L256-L261

src/esp/gfx/LightSetup.cpp
matthewjmay: left out the transformation parameter to keep it simple, if user wants they can transform range before.

src/esp/assets/ResourceManager.h
```diff
@@ -338,6 +340,32 @@ class ResourceManager {
...
+      const Magnum::ResourceKey& key);
+
+  //! @brief The @ref ShaderManager key for @ref LightInfo which has no lights
+  static constexpr char NO_LIGHT_KEY[] = "no_lights";
+
+  //! @brief The @ref ShaderManager key for the default @ref LightInfo
+  static constexpr char DEFAULT_LIGHTING_KEY[] = "default_lighting";
+
+  //! @brief The @ref ShaderManager key for the default @ref MaterialInfo
+  static constexpr char DEFAULT_MATERIAL_KEY[] = "default_material";
```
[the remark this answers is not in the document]
matthewjmay: I want these to be exported so the defaults can be updated without hardcoding the "default" string everywhere that does this. eg. `Simulator`, and `Viewer` do this. For the `SHADER_KEY`, they don't really need to be in the headers, but personally I like that since it makes a collision down the line easier to debug without looking through the implementation files.

src/esp/assets/GltfMeshData.cpp
```diff
@@ -19,7 +19,8 @@ void GltfMeshData::uploadBuffersToGPU(bool forceReload) {
   renderingBuffer_.reset();
   renderingBuffer_ = std::make_unique<GltfMeshData::RenderingBuffer>();
   // position, normals, uv, colors are bound to corresponding attributes
-  renderingBuffer_->mesh = Magnum::MeshTools::compile(*meshData_);
+  renderingBuffer_->mesh = Magnum::MeshTools::compile(
+      *meshData_, Magnum::MeshTools::CompileFlag::GenerateSmoothNormals);
```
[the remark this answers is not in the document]
matthewjmay: Thanks! For the "sphere", it's actually not supposed to be smooth! I verified this with other rendering engines.

src/esp/assets/ResourceManager.cpp
matthewjmay: actually, every single asset I've tried to load has shininess == 1, ie. `data/objects/banana.glb`, `data/objects/cheezit.glb`

src/esp/assets/ResourceManager.h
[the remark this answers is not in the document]
aclegg3: I think it's fine to leave these explicit for now.

src/esp/assets/ResourceManager.cpp
aclegg3: Can you try the other test objects? All the test objects we provide are converted obj -> glb YCB objects.

mosra: It's actually done by Magnum's TinyGltfImporter sorry. It attempts to translate the PBR materials to a basic Phong shading and leaves shininess at 1 unless the model defines also a Blinn/Phong shading model.

I'll patch this to a better default and remove this TODO once we upgrade to latest magnum.

aclegg3: Thanks for identifying the cause, Mosra. 

In the meantime it makes sense to provide an easy handle for manipulating the material so users can edit this (follow-up PR?). 

It may also be better to assume low/no shininess for most objects (furniture/cardboard/etc). Let's iterate on the default based on a set of test objects.

erikwijmans:
@matthewjmay Looks like we have functionality in habitat-api that gets broken by a singleton also.  We have a threaded vector env (instead of the normal subprocess vector env) that is nice for debugging.  This is fine with OpenGL contexts as those are local to a thread, but a singleton will get shared across threads.  @mathfac do you know if its fine to limit the threaded vector env to just 1 sub-thread for a little bit?  AFAIK, that functionality is only really useful for debugging where 1 process would likely be sufficient?

mathfac:
@erikwijmans, great suggestion about making `num_ens=1`. Yes, Threaded Vec envs are useful for sure and we want to keep it, @mosra what is ETA to remove the singleton constrain and update us to new version of Magnum?

src/esp/assets/ResourceManager.cpp
mosra: The default is fixed in #496 :)

mosra:
> what is ETA to remove the singleton constrain and update us to new version of Magnum?

ETA is now -- see #496 :) I came to a conclusion that removing the deprecated functionality is the lesser evil, even though it would break things for existing users who didn't have a chance to upgrade yet.

Issue #332, opened 2019-11-01
Render the Agent!


dhruvbatra:
## 🚀 Feature
We should render an actual agent/virtual-robot as opposed to it being an idealized cylinder. Something like in the video here: 
http://crlab.cs.columbia.edu/learning_your_way/

Fetch files: 
http://wiki.ros.org/Robots/fetch

Locobot files: 
https://github.com/facebookresearch/pyrobot/tree/master/robots/LoCoBot/locobot_description

## Motivation

Right now, all observations in habitat-sim are "first-person" egocentric perspective. We should enable "third-person" (camera above/behind the agent's head) videos. However, in order to be intuitive, they should include a rendering of the agent. 

## Details

We don't need full URDF support of articulated parts for this. Initially, we could simply combine all the part meshes into a single hierarchical mesh of the robot and render that. 

I could be mistaken, but I believe it will require providing agent's class with a mesh to be rendered for the agent: 
https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/agent/Agent.cpp

We will need to be careful that the agent's camera doesn't get self-occluded via the agent mesh. 

Also, we should add an overhead camera to the agent so that we can easily create visualization videos like this: 
![IMG_0862](https://user-images.githubusercontent.com/2941091/68060627-7ab81a80-fd44-11e9-8f9e-7c14a4b567eb.jpg)

Issue contains feedback from @erikwijmans and @msavva

aclegg3:
Addressed by #492.

PR #492, opened 2020-02-13
Attach SceneNode to simulation

aclegg3:
## Motivation and Context

This change enables an existing SceneNode to be attached to a newly instanced physics object. Also enables removal of objects from simulation while deleting either the object's full SceneNode structure, the visual component only, or nothing. 

This enables rendering of embodied agents (#332) and continuous control of an Agent's state via `PhysicsManager` functions. 

Added python bindings for this functionality.

## How Has This Been Tested
 
C++ and python unit tests ~~and new viewer demo option to attach/detach the agent to/from physics with `'p'`. While attached to an object, `'w','a','s','d'` command dynamic linear velocity of the attached object (instead of navmesh discrete control of the agent) and `LEFT|RIGHT` command local angular velocity (+-Y axis).~~ **_Edit: viewer demo shown below removed to keep things clean, but functionality is the same._**

Example - agent attachment to a can with user controlled rolling:
attach_agent_can_small

Example - agent attachment to a locobot with user control:
attach_agent_locobot_small

## Types of changes

## Checklist


src/esp/physics/PhysicsManager.cpp
```diff
@@ -48,7 +48,8 @@ bool PhysicsManager::addScene(
 }
 
 int PhysicsManager::addObject(const int objectLibIndex,
-                              DrawableGroup* drawables) {
+                              DrawableGroup* drawables,
```
matthewjmay: We should probably change these parameters to a `ObjectConfig` of some type, especially since this will soon also include light setup, material setup, etc., some of which we'll want defaults for, others required, and so on... Don't want to have massive parameter lists like we do in `ResourceManager` create drawable pipeline.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -68,7 +70,8 @@ int PhysicsManager::addObject(const int objectLibIndex,
   //! Draw object via resource manager
   //! Render node as child of physics node
   resourceManager_->loadObject(
-      configFile, &existingObjects_.at(nextObjectID_)->node(), drawables);
+      configFile, existingObjects_.at(nextObjectID_)->visualNode_, drawables);
+  existingObjects_.at(nextObjectID_)->node().computeCumulativeBB();
```
matthewjmay: how do you feel about moving https://github.com/facebookresearch/habitat-sim/blob/01a8e36ba8fba7192e7a336a0b26b3b35df95603/src/esp/assets/ResourceManager.cpp#L448 here too? This is more a responsibility of physics manager anyways

src/esp/physics/PhysicsManager.cpp
```diff
@@ -133,11 +141,15 @@ int PhysicsManager::deallocateObjectID(int physObjectID) {
 //! Create and initialize rigid object
 int PhysicsManager::makeRigidObject(
     const std::vector<assets::CollisionMeshData>& meshGroup,
-    assets::PhysicsObjectAttributes physicsObjectAttributes) {
+    assets::PhysicsObjectAttributes physicsObjectAttributes,
+    scene::SceneNode* attachmentNode /* = nullptr */) {
   int newObjectID = allocateObjectID();
-  scene::SceneNode& newNode = staticSceneObject_->node().createChild();
+  scene::SceneNode* newNode = attachmentNode;
```
matthewjmay: small thing: can you change the name of `newNode` since it might now be new anymore

src/esp/physics/RigidObject.cpp
```diff
@@ -9,7 +9,9 @@ namespace esp {
 namespace physics {
 
 RigidObject::RigidObject(scene::SceneNode* rigidBodyNode)
-    : Magnum::SceneGraph::AbstractFeature3D(*rigidBodyNode) {}
+    : Magnum::SceneGraph::AbstractFeature3D(*rigidBodyNode) {
+  visualNode_ = &rigidBodyNode->createChild();
```
matthewjmay: if you use the initializer list, this could be a reference, unless we plan to have a rigidobject without a visual part though. Either way could you use the initializer list?

src/esp/physics/RigidObject.cpp
```diff
@@ -64,10 +66,8 @@ bool RigidObject::setMotionType(MotionType mt) {
 }
 
 void RigidObject::shiftOrigin(const Magnum::Vector3& shift) {
-  // shift each child node
-  for (auto& child : node().children()) {
-    child.translate(shift);
-  }
+  // shift visual components
+  visualNode_->translate(shift);
```
matthewjmay: nice :+1:

src/esp/physics/RigidObject.h
```diff
@@ -507,6 +507,11 @@ class RigidObject : public Magnum::SceneGraph::AbstractFeature3D {
   //! drawing is off. See @ref toggleBBDraw().
   scene::SceneNode* BBNode_ = nullptr;
 
+  /**
+   * @brief All Drawable components are children of this node.
+   */
+  scene::SceneNode* visualNode_ = nullptr;
```
matthewjmay: any reason this is public versus having a getter (like `node()`)?

src/esp/physics/PhysicsManager.cpp
aclegg3: 👍 Good point. We'll continue to discuss this for a later refactor.

src/esp/physics/PhysicsManager.cpp
aclegg3: Thanks for point this out. Actually, that call is redundant after this addition.

src/esp/physics/RigidObject.h
aclegg3: Reasoning: Currently,  `PhysicsManager` needs access to these pointers and they can be null, so references aren't ideal. The getters would then be for raw pointer anyway. Users should not be getting direct access to a RigidObject to manipulate these, so I left them at public.

src/tests/PhysicsTest.cpp
```diff
@@ -408,3 +408,52 @@ TEST_F(PhysicsManagerTest, TestVelocityControl) {
     ASSERT_EQ(physicsManager_->getAngularVelocity(objectId), Magnum::Vector3{});
   }
 }
+
+TEST_F(PhysicsManagerTest, TestSceneNodeAttachment) {
```
bigbike: From now on, suggest using Corrade::TestSuite.

src/tests/PhysicsTest.cpp
```diff
@@ -408,3 +408,52 @@ TEST_F(PhysicsManagerTest, TestVelocityControl) {
...
+      dataDir, "test_assets/objects/transform_box.glb");
+
+  std::string sceneFile =
+      Cr::Utility::Directory::join(dataDir, "test_assets/scenes/plane.glb");
+
+  initScene(sceneFile);
+
+  esp::assets::PhysicsObjectAttributes physicsObjectAttributes;
+  physicsObjectAttributes.setString("renderMeshHandle", objectFile);
+  resourceManager_.loadObject(physicsObjectAttributes, objectFile);
```
bigbike: Underscore is not needed.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -88,21 +91,27 @@ int PhysicsManager::addObject(const int objectLibIndex,
 
 int PhysicsManager::addObject(const std::string& configFile,
                               DrawableGroup* drawables,
+                              scene::SceneNode* attachmentNode,
```
bigbike: May I understand why this argument is a must have?
Can we just create a scene object within addObject(...), and thus hide this detail to the user?
I did not see any usage case in this PR that leverage this argument. The only caller (in Simulator.cpp) just passed a nullptr.

Did I miss the big picture here?

src/tests/PhysicsTest.cpp
mosra: Yes but definitely in a separate PR :) 

The new test doesn't do anything that would *strongly* benefit from switching to TestSuite, so I'm fine with keeping it in gtest and rewriting with TestSuite later.

src/utils/viewer/viewer.cpp
```diff
@@ -510,17 +530,35 @@ void Viewer::mouseMoveEvent(MouseMoveEvent& event) {
   event.setAccepted();
 }
 
+void Viewer::setAgentObjectLocalVelocity(Magnum::Vector3 vel, bool angular) {
```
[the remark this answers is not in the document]
mosra: Yep, definitely, functions that do two things depending on a bool parameter are never a good API idea. Also:

```suggestion
void Viewer::setAgentObjectLocalVelocity(const Magnum::Vector3& vel, bool angular) {
```

src/tests/PhysicsTest.cpp
aclegg3: This is a call to `PhysicsManagerTest::resourceManager_`, so underscore is needed. Did I miss your meaning?

src/esp/physics/PhysicsManager.cpp
aclegg3: I pushed this into python and wrote a test case in test_physics.py to demonstrate the usage. This PR is really to address the visual/simulated agent problems. An Agent `SceneNode` can be passed to `addObject` to use that node instead of generating a new one.

src/tests/PhysicsTest.cpp
bigbike: Oh, I see. This is the member variable of the class PhysicsManagerTest.

My bad.

PR #493, opened 2020-02-13
Extend bindings for OBB and misc fixes.

Skylion007:
## Motivation and Context

Added some functionality needed for properly using OBBs in the Habitat-API. 
1). Add new python bindings to OBB
2). Fix bugs computeMOBB and expose to python.
3). Have sampleNavigablePoint return a point that will always not be navigable if it fails. That ensure that we get a proper error state that we can check against pt(inf,inf,inf).
## How Has This Been Tested

It has been tested by being used for generating the ObjectNav datasets in the Habitat-API.
## Types of changes

## Checklist

### DIFF
--- a/src/esp/geo/OBB.cpp
+++ b/src/esp/geo/OBB.cpp
@@ -107,7 +107,7 @@ OBB computeGravityAlignedMOBB(const vec3f& gravity,
     const float dd = d0[0] * d1[1] - d0[1] * d1[0];
 
     const float dx = s1[0] - s0[0];
-    const float dy = s1[1] - s0[0];
+    const float dy = s1[1] - s0[1];
     const float t = (dx * d1[1] - dy * d1[0]) / dd;
 
     return s0 + t * d0;
@@ -240,7 +240,7 @@ OBB computeGravityAlignedMOBB(const vec3f& gravity,
     aabb.extend(T_w2b * pt);
   }
 
-  return OBB{aabb.center(), aabb.sizes(), T_w2b.inverse()};
+  return OBB{T_w2b.inverse() * aabb.center(), aabb.sizes(), T_w2b.inverse()};
 }
 
 }  // namespace geo
--- a/src/esp/geo/geo.cpp
+++ b/src/esp/geo/geo.cpp
@@ -39,13 +39,12 @@ std::vector<vec2f> convexHull2D(const std::vector<vec2f>& points) {
   }
 
   // Build upper hull
-  for (size_t i = idx.size() - 2, t = k + 1; i >= 0; i--) {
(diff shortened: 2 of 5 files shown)


src/esp/nav/PathFinder.cpp
```diff
@@ -854,7 +854,8 @@ static float frand() {
 
 vec3f PathFinder::Impl::getRandomNavigablePoint() {
   dtPolyRef ref;
-  vec3f pt;
+  float inf = std::numeric_limits<float>::infinity();
```
erikwijmans:
```suggestion
  constexpr float inf = std::numeric_limits<float>::infinity();
```

src/esp/geo/OBB.cpp
```diff
@@ -107,7 +107,7 @@ OBB computeGravityAlignedMOBB(const vec3f& gravity,
     const float dd = d0[0] * d1[1] - d0[1] * d1[0];
 
     const float dx = s1[0] - s0[0];
-    const float dy = s1[1] - s0[0];
+    const float dy = s1[1] - s0[1];
```
mathfac: @Skylion007, can you remind me what was the bug here?

src/esp/geo/OBB.cpp
```diff
@@ -240,7 +240,7 @@ OBB computeGravityAlignedMOBB(const vec3f& gravity,
     aabb.extend(T_w2b * pt);
   }
 
-  return OBB{aabb.center(), aabb.sizes(), T_w2b.inverse()};
+  return OBB{T_w2b.inverse() * aabb.center(), aabb.sizes(), T_w2b.inverse()};
```
mathfac: And here.

src/esp/geo/geo.cpp
```diff
@@ -39,13 +39,12 @@ std::vector<vec2f> convexHull2D(const std::vector<vec2f>& points) {
   }
 
   // Build upper hull
-  for (size_t i = idx.size() - 2, t = k + 1; i >= 0; i--) {
```
mathfac: And here, just to have a history record for all this deep changes.

src/esp/geo/OBB.cpp
Skylion007: dy was not actually calculating a delta y.

src/esp/geo/OBB.cpp
Skylion007: The center was not in world coordinates

src/esp/geo/geo.cpp
Skylion007: Resolves a segfault that occurs on Linux due to undefined behavior of unsigned integer underflows.

PR #495, opened 2020-02-14
fix physics object file and directory loading bug

matthewjmay:
## Motivation and Context

Ran into a small bug where the physics rigid object path was also the name of another directory so we were skipping it. This fixes that by still checking for file even if its also a directory. Cleaned up the logic while I fixed it.

## How Has This Been Tested

- try to import an object `AAA.phys_properties.json` with another directory `AAA` existing. Object is correctly loaded and directory is parsed

## Types of changes

## Checklist


src/esp/assets/ResourceManager.cpp
```diff
@@ -332,53 +332,55 @@ PhysicsManagerAttributes ResourceManager::loadPhysicsConfig(
...
+    namespace Directory = Cr::Utility::Directory;
+    std::string absolutePath =
+        Directory::join(configDirectory, paths[i].GetString());
+    std::string objPhysPropertiesFilename =
+        absolutePath + ".phys_properties.json";
+    const bool dirExists = Directory::isDirectory(absolutePath);
+    const bool fileExists = Directory::exists(objPhysPropertiesFilename);
+
+    if (!dirExists && !fileExists) {
+      LOG(WARNING) << "The specified file/dir does not exist. Aborting parse.";
```
aclegg3: Name the problem path in this warning.

Issue #361, opened 2019-11-25
Domain Randomization Methods?


BryanBo-Cao:
## ❓ Questions and Help

I was wondering if there are domain randomization methods in habitat(changing color, shading, texture, etc.) that we can use directly? Not just a random initial position/orientation for an agent.

Thanks!

aclegg3:
Thanks for the feature request! 👍 

At this time, we are working on a shader re-design which will allow more control over these sort of features (Issue #319). Note however that many datasets provide illumination baked into textures. We do not provide any methods for dynamic modification of these textures at this time. For assets which do not have baked in illumination, we are considering domain randomization approaches for parameters such as light positions, colors, intensities, specularity, etc... 

In short, we absolutely have this use case in mind and will keep you posted about updates.

Issue #315, opened 2019-10-18
PLY import fails to load texture.


tlalexander:
## 🐛 Bug

When loading a PLY file as a scene, I expect the texture from the PLY file to load. Instead I get a view of only black or white pixels with no texture. The file shows texture in Meshlab and CloudCompare.

## Command

./build/viewer /path/to/scene.ply

## To Reproduce

Download the file I am using:
https://drive.google.com/open?id=1vEeRPTJunJ_ck0W1aFvzfh4tHoWP4Ecr

Run it in the viewer with the command above. You should see a viewer window with only black and white pixels and no texture.

(Note about the file: This file looks weird. It is a hastily resampled version of a larger mesh I am using. However the larger version and this resampled version exhibit the same behavior, and its easier to share the small version.)

## Expected behavior

I expect the mesh to load with the texture.

## Additional context

I originally generated these PLY files using photogrammetry with COLMAP. Fixing this bug means enabling photogrammetry scene capture to simulation. :)

Also the command line shows some warnings which I've found originate in Magnum. See the full terminal output below:

```
taylor@taylor-desktop:~/Software/habitat-sim$ ./build/viewer examples/colmap/depth/mesh_colmap.ply 
Renderer: Quadro P5000/PCIe/SSE2 by NVIDIA Corporation
OpenGL version: 4.6.0 NVIDIA 430.26
Using optional features:
    GL_ARB_ES2_compatibility
    GL_ARB_direct_state_access
    GL_ARB_get_texture_sub_image
    GL_ARB_invalidate_subdata
    GL_ARB_multi_bind
    GL_ARB_robustness
    GL_ARB_separate_shader_objects
    GL_ARB_texture_filter_anisotropic
    GL_ARB_texture_storage
    GL_ARB_texture_storage_multisample
    GL_ARB_vertex_array_object
    GL_KHR_debug
Using driver workarounds:
    no-forward-compatible-core-context
    no-layout-qualifiers-on-old-glsl
    nv-zero-context-profile-mask
    nv-implementation-color-read-format-dsa-broken
    nv-cubemap-inconsistent-compressed-image-size
    nv-cubemap-broken-full-compressed-image-query
    nv-compressed-block-size-in-bits
Trade::AssimpImporter::material(): white ambient detected, forcing back to black
```

aclegg3:
Hey, I don't have permissions for the shared file. I've send a request. 😄

aclegg3:
[...]

PR #500, opened 2020-02-19
Expose lighting setup and individual object lighting manipulation

### LINKED ISSUE #302: Question about texture of inserted objects
I have a model of a chair, and I am using the object insertion capability to add the model to a test scene.  The texture of the synthetic object does not look very good.  Basically, there is almost no texture at all.  I am wondering if I am doing something wrong.  Maybe I need to add a lighting setting or something?  Attached is a screenshot of the glb file in an online viewer and the image generated by Habitat for the same object.

screenshot_04
scene_and_object_color_001

matthewjmay:
## Motivation and Context

This PR builds on #489 and allows manipulation of light setups through the simulator.

Fixes #302.
Contributes to #332, #361, #319, #315.

Through the simulator, the user can now:
* select a light setup to use when adding an object
* change the light setup a single object uses
* create new light setups
* change an existing light setup, updating all current objects using that setup

For changing the setup for an individual entity, we currently have to traverse the subtree of that object and dynamic cast to the `Drawable` feature. There are ways to avoid this (ie. store a set of `Drawable` features) which requires larger code changes and changes to the physics objects memory management, which I can revisit later if wanted.

## How Has This Been Tested
* many C++ screenshot tests which compare different lighting configurations, manipulation of lighting configurations, and more!
* Python sanity tests which update lighting configurations
* Visual tests using viewer trying out different lighting setups

## Types of changes

## Checklist


matthewjmay:
There's an annoying issue where a `const char[]` can't be used as a default parameter to a `Magnum::ResourceKey` since default parameters undergo array to pointer decay, and `ResourceKey` ctor requires array size info. To mitigate this I've changed default parameter types to strings.

mosra:
> There's an annoying issue where a `const char[]` can't be used as a default parameter to a `Magnum::ResourceKey` on certain compilers

`ResourceKey keyParam = ResourceKey{"bla"}` should work I think (used that back in GCC 4.4 days), no need to go back to `std::string`. I'll check this, I think the constructor could accept a plain pointer also.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -111,6 +113,32 @@ void initGfxBindings(py::module& m) {
...
+
+  py::class_<LightInfo>(m, "LightInfo")
+      .def(py::init())
+      .def(py::init<Magnum::Vector3, Magnum::Color4, LightPositionModel>(),
+           "position"_a, "color"_a = Magnum::Color4{1},
+           "model"_a = LightPositionModel::GLOBAL)
+      .def_readwrite("position", &LightInfo::position)
+      .def_readwrite("color", &LightInfo::color)
+      .def_readwrite("model", &LightInfo::model)
+      .def("__eq__",
```
mosra:
```suggestion
      .def(py::self == py::self)
```

Same below. [Docs.](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#operator-overloading). Tho you might need to [suppress some warnings](https://github.com/mosra/magnum-bindings/blob/15d0714d90ecb66957e8ab518c62ccf195b3945d/src/python/magnum/math.vector.h#L97-L100).

matthewjmay:
> `ResourceKey keyParam = ResourceKey{"bla"}` should work I think (used that back in GCC 4.4 days), no need to go back to `std::string`. I'll check this, I think the constructor could accept a plain pointer also.

Right, the issue is that I want to be able to pass in `ResourceManager::DEFAULT_LIGHTING_KEY` instead of the literal "bla". I could always use your suggestion to just use an empty string to do this though.

mosra:
Yes, I mean `ResourceKey param = ResourceKey{ResourceManager::DEFAULT_LIGHTING_KEY}` should work too, just that the implicit conversion decays the array for some reason.

src/esp/sim/Simulator.h
```diff
@@ -27,6 +27,9 @@ class SemanticScene;
 namespace gfx {
 class Renderer;
 }  // namespace gfx
+namespace physics {
+enum class MotionType;
```
mosra:
```suggestion
enum class MotionType: int;
```

I wonder why compilers don't warn here? I'd always specify the type as well to be safe, otherwise if the actual `MotionType` definition has 64-bit values, this forward declaration would be incorrect.

src/esp/gfx/ShaderManager.cpp
```diff
@@ -0,0 +1,23 @@
...
+namespace esp {
+namespace gfx {
+
+void setLightSetupForSubTree(scene::SceneNode& root,
+                             const Magnum::ResourceKey& lightSetup) {
+  scene::preOrderFeatureTraversalWithCallback<Drawable>(
+      root, [&lightSetup](Drawable& drawable) {
+        drawable.setLightSetup(lightSetup);
+      });
+}
```
mosra: Is this a "breadth-first" traversal? For this particular operation I think a simple recursion without any explicit `std::stack` could be enough (please correct me if I'm wrong, of course).

In any case, various tree traversals would be useful to have directly in Magnum itself, I'll think about integrating them.

matthewjmay: This is a pre-order, depth first traversal (ie. callback executed on parent, then move to children from left to right). Recursion would be fine, but I don't think we gain anything from it since the code is already quite readable, and technically the iterative approach is slightly faster, although I doubt there would be any real difference unless the tree was huge.

src/tests/SimTest.cpp
```diff
@@ -124,8 +164,116 @@ void SimTest::getPinholeCameraRGBAObservation() {
...
+}
+
+void SimTest::getSceneRGBAObservation() {
+  setTestCaseName(CORRADE_FUNCTION);
+  auto simulator = getSimulator(vangogh);
+  checkPinholeCameraRGBAObservation(*simulator, "SimTestExpectedScene.png");
+}
+
+void SimTest::getDefaultLightingRGBAObservation() {
+  setTestCaseName(CORRADE_FUNCTION);
```
mosra: `setTestCaseName()` is strictly necessary only in the `getSceneRGBAobservation()` above, in the others the `CORRADE_VERIFY()` macros are early enough to take care of extracting the correct function name.

I tried to find a way around having to explicitly call this, but it looks like the approach you went with here was the best possible :) The test cases are not similar enough to make use of [instanced tests](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-instanced) and OTOH putting the common sensor & agent setup into a [setup / teardown function](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-setup-teardown) would unnecessarily complicate things.

src/tests/SimTest.cpp
```diff
@@ -124,8 +164,116 @@ void SimTest::getPinholeCameraRGBAObservation() {
           Mn::PixelFormat::RGBA8Unorm,
           {pinholeCameraSpec->resolution[0], pinholeCameraSpec->resolution[1]},
           observation.buffer->data}),
-      Cr::Utility::Directory::join(TEST_DIR, "SimTestExpected.png"),
-      (Mn::DebugTools::CompareImageToFile{15.0f, 0.17f}));
+      Cr::Utility::Directory::join(screenshotDir, groundTruthImageFile),
+      (Mn::DebugTools::CompareImageToFile{30.f, 0.5f}));
```
mosra: Not sure if having a global comparison threshold for all tests is a good idea in the long run -- I can imagine some particular test scenes might exhibit larger differences across drivers than others, and then globally increasing the thresholds would give a false sense of security.

Maybe have the min/mean threshold as a parameter to `checkPinholeCameraRGBAObservation()` instead?

src/tests/SimTest.cpp
```diff
@@ -124,8 +164,116 @@ void SimTest::getPinholeCameraRGBAObservation() {
           Mn::PixelFormat::RGBA8Unorm,
           {pinholeCameraSpec->resolution[0], pinholeCameraSpec->resolution[1]},
           observation.buffer->data}),
-      Cr::Utility::Directory::join(TEST_DIR, "SimTestExpected.png"),
-      (Mn::DebugTools::CompareImageToFile{15.0f, 0.17f}));
+      Cr::Utility::Directory::join(screenshotDir, groundTruthImageFile),
```
mosra: I added the `TEST_DIR` variable to `src/tests/configure.h.cmake` in #496, so if it's not needed anymore, can you remove it again? Thanks!

src/tests/SimTest.cpp
matthewjmay: looking at the test failures there seem to be massive differences across drivers (a few pixels vary by over 100). The good news is that mean differences are still very low. Do you think we should just ignore individual pixel differences and only test mean?

matthewjmay: @aclegg3 @mosra @bigbike the other option would be to only run these visual tests in CI, but this would be annoying for debugging build failures. Is anyone opposed to only checking the mean? Lets try to get this solved soon since this seems to be the only blocker.

erikwijmans: I am not opposed to just checking the mean.  I view these tests as more as smoke tests than anything (are the pipes all still hooked up correctly) -- to that end, if you try to break things, does the mean catch it? i.e. turn off phong shading for one that test that uses it, change the light color dramatically, etc.

mosra: Those massive differences in few pixels are expected, and are due to differences in how rasterizers work on various GPUs. The mean will still catch it I think.

That said, I'm open to ideas how to handle this better in the image comparison tool. Some options I was thinking of:

1. removing (some configurable percentage of) largest outliers, checking max/mean only in the rest
2. applying a slight gaussian blur before doing the comparison which should not cause false positives since it's IIRC a reversible operation, but on the other hand should account better for slight spatial differences and differences in edge rasterization
3. something else? (how do browser test frameworks do it? any obvious tricks I'm missing?)

erikwijmans: I think most frameworks use software rendering for exact image comparison tests.  I do like the idea of outlier removal.  Unlikely that a few pixels being very wrong is actually a problem.

mosra: Great. I'll return with a PR that adapts the test when the outlier removal is implemented.

PR #501, opened 2020-02-20
[optimization] compute AABB from OBB

bigbike:
For algorithm details, see my comments before the function in cpp file.

**Old** method:
do 8 matrix4x4-vertex multiplications and 8 vector min+max operations

**New** method:
convert aabb from (min, max) to (center, extent), create a matrix3x3, do 1 matrix4x4-vector
multiplication, 1 matrix3x3-vector multiplication, and convert (center,
    extent) to (min, max), which should be faster.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/geo/geo.cpp
+++ b/src/esp/geo/geo.cpp
@@ -7,6 +7,8 @@
 #include <Magnum/Math/FunctionsBatch.h>
 #include <numeric>
 
+namespace Mn = Magnum;
+
 namespace esp {
 namespace geo {
 
@@ -58,23 +60,63 @@ std::vector<vec2f> convexHull2D(const std::vector<vec2f>& points) {
 
   return hull;
 }
+/**
+ * Assume the aabb is expressed as the center 'c' and the extent 'e'.
+ * each corner X is nothing but a combination from
+ * (c_x +/- e_x, c_y +/- e_y, c_z +/- e_z)
+ *
+ * Denote y = (+/- e_x, +/- e_y, +/- e_z)
+ *
+ * The corner is transformed by:
+ * x = R * x0 + t,
+ * where x_0, x are the coordinates before and after transformation, t is the
+ * translation x = R * (c0 + y0) + t  = (Rc0 + t) + Ry0    eq(1)
+ *
+ * Our Goal is to find the x_max and x_min after the transformation.
+ *
+ * First, determing the x_max:
+ *
+ * In eq(1), Rc0 + t is a constant, which means max{x} iff max{Ry0}
+ *
+ * R looks like:
+ * [R0, R1, R2]
+ * [R3, R4, R5]
+ * [R6, R7, R8]
+ *
+ * We focus on the 1st entry (the 'x' entry) of Ry0:
+ * y_x =<(R0, R1, R2), (+/- e_x, +/- e_y, +/- e_z)>
+ *     =<(+/- R0, +/- R1, +/- R2), (e_x, e_y, e_z)>
+ *
+ * Now, note that e_x, e_y, e_z are all positive values for any non-degenerate
+ * aabb.
+ *
+ * So y_x reaches MAX when +/- R0, +/- R1, +/- R2 are all >=0
+ * that means max{y_x} = <(|R0|, |R1|, |R2|), (e_x, e_y, e_z)>
+ * (|R0| means the absolute value of R0)
+ *
+ * and likewise for y_y, y_z
+ *
+ * The derivation for x_min is similar since same logic applies.
+ */
+Mn::Range3D getTransformedBB(const Mn::Range3D& range,
+                             const Mn::Matrix4& xform) {
+  Mn::Matrix3x3 absRotationScaling = xform.rotationScaling();
+  // compute the absolute value of the rotationScaling part of the original
+  // transformation matrix
+  for (int iRow = 0; iRow < 3; ++iRow) {
+    for (int jCol = 0; jCol < 3; ++jCol) {
+      absRotationScaling[iRow][jCol] = abs(absRotationScaling[iRow][jCol]);
+    }
+  }
+  const Mn::Vector3 center = (range.min() + range.max()) / 2.0;
+  const Mn::Vector3 extent = (range.max() - range.min()) / 2.0;
 
-Magnum::Range3D getTransformedBB(const Magnum::Range3D& range,
-                                 const Magnum::Matrix4& T) {
-  std::vector<Magnum::Vector3> corners;
-  corners.push_back(T.transformPoint(range.frontBottomLeft()));
-  corners.push_back(T.transformPoint(range.frontBottomRight()));
-  corners.push_back(T.transformPoint(range.frontTopLeft()));
-  corners.push_back(T.transformPoint(range.frontTopRight()));
+  // compute Rc0 + t
+  Mn::Vector3 newCenter = xform.transformPoint(center);
+  // compute max{Ry0}
+  Mn::Vector3 newExtent = absRotationScaling * extent;
 
-  corners.push_back(T.transformPoint(range.backTopLeft()));
-  corners.push_back(T.transformPoint(range.backTopRight()));
-  corners.push_back(T.transformPoint(range.backBottomLeft()));
-  corners.push_back(T.transformPoint(range.backBottomRight()));
-
-  Magnum::Range3D transformedBB{Magnum::Math::minmax<Magnum::Vector3>(corners)};
-
-  return transformedBB;
+  return Mn::Range3D::fromCenter(newCenter, newExtent);
 }
 
 }  // namespace geo
--- a/src/tests/GeoTest.cpp
+++ b/src/tests/GeoTest.cpp
@@ -2,6 +2,7 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
+#include <Magnum/Math/FunctionsBatch.h>
 #include <gtest/gtest.h>
 #include "esp/geo/CoordinateFrame.h"
 #include "esp/geo/OBB.h"
@@ -97,4 +98,69 @@ TEST(GeoTest, CoordinateFrame) {
   CoordinateFrame c4;
   c4.fromJson(j);
   EXPECT_EQ(c3, c4);
+}
+
+// standard method
+// transform the 8 corners, and extract the min and max
+Magnum::Range3D getTransformedBB_standard(const Magnum::Range3D& range,
+                                          const Magnum::Matrix4& xform) {
+  std::vector<Magnum::Vector3> corners;
+  corners.push_back(xform.transformPoint(range.frontBottomLeft()));
+  corners.push_back(xform.transformPoint(range.frontBottomRight()));
+  corners.push_back(xform.transformPoint(range.frontTopLeft()));
+  corners.push_back(xform.transformPoint(range.frontTopRight()));
+
+  corners.push_back(xform.transformPoint(range.backTopLeft()));
+  corners.push_back(xform.transformPoint(range.backTopRight()));
+  corners.push_back(xform.transformPoint(range.backBottomLeft()));
+  corners.push_back(xform.transformPoint(range.backBottomRight()));
+
+  Magnum::Range3D transformedBB{Magnum::Math::minmax<Magnum::Vector3>(corners)};
+
+  return transformedBB;
+}
+
+TEST(GeoTest, Aabb) {
+  // TODO: Generate N boxes in random position and orientation
+  // compute aabb for each box using standard method and library method
+  // respectively compare the results, which should be identical
+
+  Magnum::Range3D box = Magnum::Range3D{Magnum::Vector3{-10.0, -10.0, -10.0},
+                                        Magnum::Vector3{10.0, 10.0, 10.0}};
+
+  int numBoxes = 100;
+  for (int iBox = 0; iBox < numBoxes; ++iBox) {
+    // generate random translation
+    const Magnum::Vector3 translation{
+        static_cast<float>(rand() % 1000) + (rand() % 1000) / 1000.0f,
+        static_cast<float>(rand() % 1000) + (rand() % 1000) / 1000.0f,
+        static_cast<float>(rand() % 1000) + (rand() % 1000) / 1000.0f};
+
+    // generate random rotation using:
+    // http://planning.cs.uiuc.edu/node198.html
+    double u1 = (rand() % 1000) / 1000.0;
+    double u2 = (rand() % 1000) / 1000.0;
+    double u3 = (rand() % 1000) / 1000.0;
+
+    Magnum::Vector3 qAxis(sqrt(1 - u1) * cos(2 * M_PI * u2),
+                          sqrt(u1) * sin(2 * M_PI * u3),
+                          sqrt(u1) * cos(2 * M_PI * u3));
+
+    // assemble the transformation matrix
+    Magnum::Matrix4 xform = Magnum::Matrix4::from(
+        Magnum::Quaternion(qAxis, sqrt(1 - u1) * sin(2 * M_PI * u2)).toMatrix(),
+        translation);
+
+    Magnum::Range3D aabbControl = getTransformedBB_standard(box, xform);
+    Magnum::Range3D aabbTest = esp::geo::getTransformedBB(box, xform);
+
+    auto diff = [](Magnum::Vector3& control, Magnum::Vector3& test) -> float {
+      Magnum::Vector3 a = control - test;
+      return dot(a, a);
+    };
+
+    float eps = 1e-8;
+    EXPECT_LE(diff(aabbControl.min(), aabbTest.min()), eps);
+    EXPECT_LE(diff(aabbControl.max(), aabbTest.max()), eps);
+  }  // for iBox
 }
--- a/src/esp/geo/geo.h
+++ b/src/esp/geo/geo.h
@@ -29,7 +29,7 @@ typedef Eigen::Transform<float, 3, Eigen::Affine, Eigen::DontAlign> Transform;
 std::vector<vec2f> convexHull2D(const std::vector<vec2f>& points);
 
 Magnum::Range3D getTransformedBB(const Magnum::Range3D& range,
-                                 const Magnum::Matrix4& T);
+                                 const Magnum::Matrix4& xform);
 
 template <typename T>
 T clamp(const T& n, const T& low, const T& high) {


bigbike:
Dear reviewers,

Thanks.

src/esp/geo/geo.cpp
```diff
@@ -58,23 +61,66 @@ std::vector<vec2f> convexHull2D(const std::vector<vec2f>& points) {
...
+                             const Mn::Matrix4& xform) {
+  Mn::Matrix3x3 absRotationScaling = xform.rotationScaling();
+  // compute the absolute value of the rotationScaling part of the original
+  // transformation matrix
+  for (int iRow = 0; iRow < 3; ++iRow) {
+    for (int jCol = 0; jCol < 3; ++jCol) {
+      absRotationScaling[iRow][jCol] =
+          std::fabs(absRotationScaling[iRow][jCol]);
+    }
+  }
```
mosra: The above can be 

```cpp
auto absRotationScaling = Mn::Matrix3x3::fromVector(Mn::Math::abs(xform.rotationScaling().toVector()));
```

Explanation: the Math lib has abs()/min()/max()... functions for vectors, but not matrices (to keep it maintainable), so the above is the "canonical way" to do it. 

In any case, when you find yourself looping over matrix/vector elements to do some calculation, there's a high chance such operation is already in the API ;)

src/esp/geo/geo.cpp
```diff
@@ -58,23 +61,66 @@ std::vector<vec2f> convexHull2D(const std::vector<vec2f>& points) {
...
+  const Mn::Vector3 center = (range.min() + range.max()) / 2.0;
```
mosra:
```suggestion
  const Mn::Vector3 center = range.center();
```

src/esp/geo/geo.cpp
```diff
@@ -58,23 +61,66 @@ std::vector<vec2f> convexHull2D(const std::vector<vec2f>& points) {
...
+  const Mn::Vector3 extent = (range.max() - range.min()) / 2.0;
```
mosra:
```suggestion
  const Mn::Vector3 extent = range.size()/2.0f;
```

Hmm, I think I could add an `extent()` accessor there too... oh and probably related: for some `Math::Intersection` routines we went with having a special overload that takes center + extent directly, to save the subtraction and division from the above calculation. I planto add a variant of the `Range` type that stores center + extent instead of min/max to have a builtin type for these use cases, and could prioritize that if you think storing everything as center + extent here would speed things up.

src/tests/GeoTest.cpp
```diff
@@ -98,3 +99,64 @@ TEST(GeoTest, CoordinateFrame) {
...
+    double u2 = (rand() % 1000) / 1000.0;
+    double u3 = (rand() % 1000) / 1000.0;
+
+    Magnum::Vector3 qAxis(sqrt(1 - u1) * cos(2 * M_PI * u2),
+                          sqrt(u1) * sin(2 * M_PI * u3),
+                          sqrt(u1) * cos(2 * M_PI * u3));
+
+    // assemble the transformation matrix
+    Magnum::Matrix4 xform = Magnum::Matrix4::from(
+        Magnum::Quaternion(qAxis, sqrt(1 - u1) * sin(2 * M_PI * u2)).toMatrix(),
```
mosra: I remember @aclegg3 added a similar piece of code here: https://github.com/facebookresearch/habitat-sim/blob/f2845b58217883517586feaad18fdcb46180f511/src/utils/viewer/viewer.cpp#L278-L289 ... maybe turn it into an utility function?

This would be great to have in `Magnum::Math` of course, I'll think about adding it there.

src/tests/GeoTest.cpp
```diff
@@ -98,3 +99,64 @@ TEST(GeoTest, CoordinateFrame) {
...
+        translation);
+
+    Magnum::Range3D aabbControl = getTransformedBB_standard(box, xform);
+    Magnum::Range3D aabbTest = esp::geo::getTransformedBB(box, xform);
+
+    float eps = 1e-12;
+    EXPECT_LE((aabbControl.min() - aabbTest.min()).dot(), eps);
+    EXPECT_LE((aabbControl.max() - aabbTest.max()).dot(), eps);
```
mosra: Like I did with `SimTest` in #496, maybe consider rewriting this test using `Corrade::TestSuite` as well? It should be rather easy I think. The above could then be turned into what's below, printing a much better diagnostic in case of comparison failure:

```cpp
CORRADE_COMPARE_AS(aabbTest, aabbControl, 
    Cr::TestSuite::Compare::around(1e-13f));
```

bigbike: OK, will do.

bigbike:
> Do you have any benchmark / numbers showing the execution time before and after?

I really wish I could, but the overhead to set up the experiment is too big. Basically I need to setup a physics scene, add timing functions, set the switch between the old/new functions, hack into examples.py and related pybindings to make it run. It needs too much time.

This is a known trick I learned many years ago. I am almost certain it would be faster, though missing actual data to support this claim. Let me search if someone did the measurement before.

By the way, how would you measure the performance usually? Any utility functions in Magnum? Please let me know.

src/esp/geo/geo.cpp
bigbike: Thank you. I really need it (your advice)! 

You just touched a pain point I am suffering everyday. I am not familiar with the Magnum library like you are. (I am not complaining, just describing a fact.) 

As you can see, the code is not long, but it cost me a couple of hours this afternoon. Most of the time was spent on searching and reading the Magnum docs, understanding your build-in functions, trying to simply the code. But still, you caught me in so many places that I can improve. :-)

mosra:
> By the way, how would you measure the performance usually?

Since you're porting away from gtest anyway, the `TestSuite` has benchmarking capabilities also (search for `addBenchmark()`) -- so if you can extract the routine to a function, you could benchmark it against the "standard" variant :) I did this in the [DepthUnprojectionTest](https://github.com/facebookresearch/habitat-sim/blob/175ea235d9a12cc0a7bb6903c474c705b48438d4/src/esp/gfx/test/DepthUnprojectionTest.cpp#L328), if you want an example.

src/esp/geo/geo.cpp
bigbike: Good, thank you. Will change.

src/esp/geo/geo.cpp
mosra: Anytime, If I'm around and you're thinking "there has to be some easier way", just ping me on Slack, and I can tell you the API to use ;)

bigbike:
> > By the way, how would you measure the performance usually?
> 
> Since you're porting away from gtest anyway, the `TestSuite` has benchmarking capabilities also (search for `addBenchmark()`) -- so if you can extract the routine to a function, you could benchmark it against the "standard" variant :) I did this in the [DepthUnprojectionTest](https://github.com/facebookresearch/habitat-sim/blob/175ea235d9a12cc0a7bb6903c474c705b48438d4/src/esp/gfx/test/DepthUnprojectionTest.cpp#L328), if you want an example.

Oh! Good idea! The tests is not only used to guard the code, but also used to measure the performance. Nice!

src/tests/GeoTest.cpp
bigbike: Yes, that was where I copied ("duplicated") the code. Sorry, I am lazy. Definitely we need to put it into a utility file. :-)

src/esp/geo/geo.cpp
bigbike: Interesting. Let me think which one is preferable. <min, max> or <center, extent>

src/esp/geo/geo.cpp
bigbike: > Anytime, If I'm around and you're thinking "there has to be some easier way", just ping me on Slack, and I can tell you the API to use ;)

Thanks again. It is very thoughtful.
The only issue is that we have big time zone difference. :-)

src/esp/geo/geo.cpp
mosra: Okay, adding to my TODO list :)

src/tests/GeoTest.cpp
bigbike: @mosra:

Cannot use what you suggested. The error message reads:
```
n file included from /home/user/habitat-sim/src/tests/AabbTest.cpp:5:                                                 
/home/user/habitat-sim/src/tests/AabbTest.cpp: In member function ‘void {anonymous}::GeoTest::aabb()’:                
/home/user/habitat-sim/src/tests/AabbTest.cpp:78:54: error: temporary of non-literal type ‘Corrade::TestSuite::Compare
::Around<float>’ in a constant expression                                                                             
   78 |                        Cr::TestSuite::Compare::around(eps));                                                  
      |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~                                                    
/home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1377:27: note: in definition of macro ‘CORRADE_
COMPARE_AS’                                                                                                           
 1377 |         Tester::compareAs<__VA_ARGS__>(#actual, actual, #expected, expected); \                               
      |                           ^~~~~~~~~~~                                                                         
In file included from /home/user/habitat-sim/src/tests/AabbTest.cpp:6:                                                
/home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Compare/Numeric.h:108:25: note: ‘Corrade::TestSuite::Com
pare::Around<float>’ is not literal because:                                                                          
  108 | template<class T> class Around {                                                                              
      |                         ^~~~~~                                                                                
/home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Compare/Numeric.h:108:25: note:   ‘Corrade::TestSuite::C
ompare::Around<float>’ is not an aggregate, does not have a trivial default constructor, and has no ‘constexpr’ constr
uctor that is not a copy or move constructor                                                                          
In file included from /home/user/habitat-sim/src/tests/AabbTest.cpp:5:                                                
/home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1377:76: error: no matching function for call t
o ‘{anonymous}::GeoTest::compareAs<Corrade::TestSuite::Compare::around<float>(eps)>(const char [9], Magnum::Range3D&, 
const char [12], Magnum::Range3D&)’                                                                                   
 1377 |         Tester::compareAs<__VA_ARGS__>(#actual, actual, #expected, expected); \                               
      |                                                                            ^                                  
/home/user/habitat-sim/src/tests/AabbTest.cpp:77:5: note: in expansion of macro ‘CORRADE_COMPARE_AS’                  
   77 |     CORRADE_COMPARE_AS(aabbTest, aabbControl,                                                                       |     ^~~~~~~~~~~~~~~~~~                                                                                        
/home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1123:78: note: candidate: ‘template<template<cl
ass> class T, class Actual, class Expected> void Corrade::TestSuite::Tester::compareAs(const char*, const Actual&, const char*, const Expected&)’                                                                                           
 1123 |         template<template<class> class T, class Actual, class Expected> void compareAs(const char* actual, const Actual& actualValue, const char* expected, const Expected& expectedValue) {                                        
      |                                                                              ^~~~~~~~~                        /home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1123:78: note:   template argument deduction/su
bstitution failed:                                                                                                    /home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1128:50: note: candidate: ‘template<class T, cl
ass U, class V> void Corrade::TestSuite::Tester::compareAs(const char*, const U&, const char*, const V&)’             
 1128 |         template<class T, class U, class V> void compareAs(const char* actual, const U& actualValue, const char* expected, const V& expectedValue) {                                                                                
      |                                                  ^~~~~~~~~                                                    /home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1128:50: note:   template argument deduction/substitution failed:    
```

bigbike:
@mosra :

About using addBenchmarks() found here:
https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#a424aeb9a3c2180464c155d982ae08389
It involves too much guesswork without any examples. For example, how to construct the benchmark list, what should be a benchmark? I did not figure them out until I read some actual examples in the source code of Magnum)
And when I read the docs, it mentioned, "batch size..." and I do not know what it is. No argument is "batchSize".

src/tests/GeoTest.cpp
mosra: Argh, sorry, should have been `WITH`, not `AS`, and comparing the two vectors instead of the whole Range:

```cpp
CORRADE_COMPARE__WITH(aabbTest.min(), aabbControl.min(), 
    Cr::TestSuite::Compare::around(1e-12f));
CORRADE_COMPARE__WITH(aabbTest.max(), aabbControl.max(), 
    Cr::TestSuite::Compare::around(1e-12f));
```

Btw., I think `1e-12f` is too little and could lead to failures depending on compiler / optimization level, `1e-7f` or `1e-6f` could be not overly strict but still precise enough.

mosra:
> It involves too much guesswork without any examples.

@bigbike err ... did you actually read the `Tester` class docs, at all? Everything is explained there in a high-level manner, with a full example for each, and all parameters explained: https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-benchmark 

The functions and macros then only explain the dry low-level details. If that helps you in any way, I'll add links back to the class docs from each of those.

bigbike:
> > It involves too much guesswork without any examples.
> 
> @bigbike err ... did you actually read the `Tester` class docs, at all? Everything is explained there in ...

Oh, I see the gap here. Good to point it out. No, I did not read the entire docs from scratch. I just directly went to the functions I would like to understand. Apologize for that.

src/tests/GeoTest.cpp
bigbike: OK, It depends if I compare the distance or the square distance. Usually I use 1e-6f for the distance and 1e-10f ~1e-12f for the square distance.

Your comments make sense. I can change.

bigbike: hmmm.... Still not correct.

```
/home/user/habitat-sim/src/tests/AabbTest.cpp:77:5:   required from here                                              
/home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1516:129: error: invalid initialization of reference of type ‘const ActualType&’ {aka ‘const float&’} from expression of type ‘const Magnum::Math::Vector3<float>’   
 1516 |     const typename Implementation::ComparatorTraits<T, typename std::decay<U>::type, typename std::decay<V>::type>::ActualType& actualValueInExpectedActualType = actualValue;                                                      
            |                                                                                                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                               
              /home/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Tester.h:1517:131: error: invalid initialization of reference of type ‘const ExpectedType&’ {aka ‘const float&’} from expression of type ‘const Magnum::Math::Vector3<float>’                                                                                                         
               1517 |     const typename Implementation::ComparatorTraits<T, typename std::decay<U>::type, typename std::decay<V>::type>::ExpectedType& expectedValueInExpectedExpectedType = expectedValue;                                
                  |                                                                                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                   
                    gmake[2]: *** [tests/CMakeFiles/AabbTest.dir/build.make:63: tests/CMakeFiles/AabbTest.dir/AabbTest.cpp.o] Error 1                                                                                                       
                    gmake[1]: *** [CMakeFiles/Makefile2:6263: tests/CMakeFiles/AabbTest.dir/all] Error 2 
```

bigbike:
Well, if I am reading it correctly, please enjoy **13x** speedup.
obb-optimization

bigbike:
@mosra , @erikwijmans :

I believe this is the 1st time that we introduce "benchmarks" to the system using `Corrade::TestSuite`. (After I figured out how to benchmark using the `TestSuite`, I really think it is an amazing, powerful utility. Thanks a lot. @mosra)
Questions:
-) I put the test and the benchmark together in this class, will it affect the CI tests later?
-) Should I separate them into GeoTest and GeoBenchmarks?

src/tests/AabbTest.cpp
```diff
@@ -0,0 +1,112 @@
...
+void GeoTest::aabb() {
+  // compute aabb for each box using standard method and library method
+  // respectively.
+  // compare the results, which should be identical
+  for (auto& xform : xforms_) {
+    Mn::Range3D aabbControl = Test::getTransformedBB_standard(box_, xform);
+    Mn::Range3D aabbTest = esp::geo::getTransformedBB(box_, xform);
+
+    float eps = 1e-8f;
+    CORRADE_COMPARE_WITH((aabbTest.min() - aabbControl.min()).dot(), 0.0,
```
bigbike: @mosra: This is what I did now with the `CORRADE_COMPARE_WITH(...)`. Finally it is the same as the 1st version I did in GTEST.

bigbike:
Dear reviewers,
This diff is ready to be reviewed. Thanks.

src/tests/GeoTest.cpp
mosra: Okay, then this:

```cpp
CORRADE_COMPARE__WITH(aabbTest.min(), aabbControl.min(), 
    Cr::TestSuite::Compare::around(Mn::Vector3{1e-12f}));
CORRADE_COMPARE__WITH(aabbTest.max(), aabbControl.max(), 
    Cr::TestSuite::Compare::around(Mn::Vector3{1e-12f}));
```

src/tests/AabbTest.cpp
mosra: Should work like this I think:

```cpp
CORRADE_COMPARE__WITH(aabbTest.min(), aabbControl.min(), 
    Cr::TestSuite::Compare::around(Mn::Vector3{1e-12f}));
CORRADE_COMPARE__WITH(aabbTest.max(), aabbControl.max(), 
    Cr::TestSuite::Compare::around(Mn::Vector3{1e-12f}));
```

Apologies that this took so many iterations, heh :)

mosra:
> I believe this is the 1st time that we introduce "benchmarks" to the system

Nope ;) See [my comment above](https://github.com/facebookresearch/habitat-sim/pull/501#issuecomment-588772645), the `DepthUnprojectionTest` and the `PathFinderTest` (all of them visible in the CI output).

> will it affect the CI tests later?

What do you mean?

> Should I separate them?

Not necessary. If it proves to be too slow for the CI at some point, it's possible to selectively run just the tests / just the benchmarks.

src/tests/AabbTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include <Corrade/TestSuite/Compare/Numeric.h>
+#include <Corrade/TestSuite/Tester.h>
+#include <Magnum/Math/FunctionsBatch.h>
+#include "esp/core/Utility.h"
+#include "esp/geo/CoordinateFrame.h"
+#include "esp/geo/OBB.h"
+#include "esp/geo/geo.h"
+
+// TODO: refactor the GeoTest to use Cr::TestSuite
```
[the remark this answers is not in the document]
bigbike: I am testing a function in the geo module. Ideally, it should have one GeoTest, not 2 files. 

I am planning to migrate the current GeoTest to here but it would be a lot of work and beyond the scope of this diff. So will do it later.

After it is done, will rename this file to “GeoTest” and deprecate the old one.

src/esp/core/Utility.h
```diff
@@ -0,0 +1,23 @@
+#include <Magnum/Magnum.h>
+#include <Magnum/Math/Quaternion.h>
+#include <Magnum/Math/Vector3.h>
+#include <cmath>
+#include <cstdlib>
+
+namespace esp {
+namespace core {
+Magnum::Quaternion randomRotation() {
```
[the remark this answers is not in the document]
bigbike: Will do.

src/tests/AabbTest.cpp
bigbike: Sorry, it will not work.

Should be only one underscore before WITH.

bigbike:
By the way, does the warning matter? Would it affect the confidence of the measurement? How can I get rid of it?

obb-optimization

src/tests/AabbTest.cpp
mosra: Yeah of course (edited). The main suggestion was about `around(Mn::Vector3{1e-12f})` instead of `around(1e-12f)`. Did *that* work?

mosra:
@bigbike re the `WARN` about CPU scaling -- if you run with `--verbose`, it'll tell you why and what can be done to fix that. Tho in your case the measurements weren't noisy anyway (otherwise the +- numbers would be much larger and highlighted in yellow/red), so this is fine.

src/tests/AabbTest.cpp
bigbike: Just tested. Yes, it works now. Thank you.

src/tests/AabbTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void GeoTest::aabb() {
+  // Generate N boxes in random position and orientation.
+  // compute aabb for each box using standard method and library method
+  // respectively.
+  // compare the results, which should be identical
+
+  Magnum::Range3D box = Magnum::Range3D{Magnum::Vector3{-10.0, -10.0, -10.0},
+                                        Magnum::Vector3{10.0, 10.0, 10.0}};
```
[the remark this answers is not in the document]
bigbike: changed.

bigbike:
It is ready to be reviewed again. Thank you.

bigbike:
@mosra: do you have further concerns or suggestions on this diff? Thanks.

PR #503, opened 2020-02-22
Replica OBB

erikwijmans:
## Motivation and Context

Add Replica OBB parsing.

## How Has This Been Tested

Via the test

### DIFF
--- a/src/esp/assets/GenericInstanceMeshData.h
+++ b/src/esp/assets/GenericInstanceMeshData.h
@@ -48,6 +48,10 @@ class GenericInstanceMeshData : public BaseMesh {
     return cpu_ibo_;
   }
 
+  const std::vector<uint16_t> getObjectIdsBufferObjectCPU() const {
+    return objectIds_;
+  }
+
  protected:
   void updateCollisionMeshData();
 
--- a/src/esp/assets/GenericInstanceMeshData.h
+++ b/src/esp/assets/GenericInstanceMeshData.h
@@ -48,7 +48,7 @@ class GenericInstanceMeshData : public BaseMesh {
     return cpu_ibo_;
   }
 
-  const std::vector<uint16_t> getObjectIdsBufferObjectCPU() const {
+  const std::vector<uint16_t>& getObjectIdsBufferObjectCPU() const {
     return objectIds_;
   }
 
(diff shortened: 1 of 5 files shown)


src/esp/scene/ReplicaSemanticScene.cpp
```diff
@@ -5,6 +5,7 @@
 #include "ReplicaSemanticScene.h"
 #include "SemanticScene.h"
 
+#include <Sophus/se3.hpp>
```
Skylion007: fatal error: Sophus/se3.hpp: No such file or directory

src/esp/assets/GenericInstanceMeshData.h
```diff
@@ -48,6 +48,10 @@ class GenericInstanceMeshData : public BaseMesh {
     return cpu_ibo_;
   }
 
+  const std::vector<uint16_t> getObjectIdsBufferObjectCPU() const {
```
matthewjmay: did you mean to return a ref here?

src/esp/scene/test/ReplicaSceneTest.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    for (uint64_t fid = 0; fid < ibo.size(); fid += 6) {
+      if (objectIds[fid] == id) {
+        esp::vec3f quadCenter = esp::vec3f::Zero();
+        // Mesh was converted from quads to tris
+        for (int i = 0; i < 6; ++i) {
+          quadCenter += vbo[ibo[fid + i]] / 6;
+        }
+
+        CORRADE_VERIFY(obj->obb().contains(quadCenter, 5e-2));
```
mosra: With #505 merged, you can now use [`CORRADE_ITERATION()`](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-iteration-annotations) to have an actually usable diagnostic when something fails. For example:

```cpp
  for (const auto& obj : scene.objects()) {
    if (obj == nullptr)
      continue;

    const auto& stringId = obj->id();
    const int id = std::stoi(stringId.substr(stringId.find_last_of("_") + 1));
    CORRADE_ITERATION(stringId);

    for (uint64_t fid = 0; fid < ibo.size(); fid += 6) {
      CORRADE_ITERATION(fid);
      …
```

erikwijmans: Ooo yay.  I wanting to use that but wasn't in our repo yet.

src/esp/assets/GenericInstanceMeshData.h
erikwijmans: Yes I did, thank you!

PR #508, opened 2020-02-26
Tests: fix double narrowing errors on GCC 9

mosra:
## Motivation and Context

The `CullingTest` was not compiling on my machine due to `sqrt()` returning a double, which got passed to `Vector3` initializer list. Using builtin Magnum constants for that instead, also adding `f` to all literals to avoid warnings/errors on random other compilers and initializing the `std::vector` in a single step.

## How Has This Been Tested

CIs pass, I hope.

## Types of changes

### DIFF
--- a/src/tests/CullingTest.cpp
+++ b/src/tests/CullingTest.cpp
@@ -64,25 +64,28 @@ TEST(CullingTest, computeAbsoluteAABB) {
    *  a) a cube, with edge length 2.0
    *
    */
-  std::vector<Mn::Range3D> aabbsGroundTruth;
-  // Box 0: root (parent: null), object "a", centered at origin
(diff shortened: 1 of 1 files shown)


bigbike:
Sorry about this issue. It is weird since I used Fedora with GCC 9.2 and it seemed no problem on my end.
Anyway, thanks for fixing it.

mosra:
Interesting :) Not completely sure what's to blame, somehow in my case only a `double sqrt(double)` overload was present, causing a narrowing error, and in your case it picked the expected `float sqrt(float)`. This is one of the reasons why I've gotten used to writing everything with `std::` in front -- that ensures there's always the correct overload to pick from.

PR #512, opened 2020-02-28
Fix most compiler warnings

matthewjmay:
## Motivation and Context

Compiler warnings make it hard to find actual errors and make people think its okay to have warnings as they add code. This PR removes most of them.
Side note: we should consider turning on `-Werror` to treat warnings as errors.

## How Has This Been Tested

* tests pass

## Types of changes

## Checklist


src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -46,13 +46,13 @@ struct SimulationContactResultCallback
...
-                           const btCollisionObjectWrapper* colObj1Wrap,
-                           int partId1,
-                           int index1) {
+  btScalar addSingleResult(btManifoldPoint&,
+                           const btCollisionObjectWrapper*,
+                           int,
+                           int,
+                           const btCollisionObjectWrapper*,
+                           int,
+                           int) {
```
aclegg3: You can use `CORRADE_UNUSED` to keep the variable names for semantic understanding while suppressing warnings.

mosra:
As @aclegg3 pointed out as well, I would keep parameter names in the `*.h` files for documentation purposes, and use either `CORRADE_UNUSED` or `static_cast<void>(foo)` to suppress warnings.

> we should consider turning on `-Werror` to treat warnings as errors

I don't recommend that, often a minor compiler / dependency upgrade causes warnings to appear, and `-Werror` is then blocking further progress until the warning is circumvented. (Also Magnum is issuing deprecation warnings on API upgrades, and treating those as errors would basically go against my effort of providing backward compatibility.)

PR #514, opened 2020-03-02
Refactor bullet scene construction

aclegg3:
## Motivation and Context

Collision scene construction implicitly assumed single collision mesh and did not properly create the static collision objects for scenes composed of many meshes.

This change refactors `initializeScene` to fix this problem and breaks construction of collision objects from compound assets into two distinct functions: `constructBulletSceneFromMeshes`
for static, concave, scene collision meshes and `constructBulletCompoundFromMeshes` for kinematic and dynamic rigid object convex hulls.

Also cleaned up some comments and leftover references to `btDiscreteDynamicsWorld`. Modified collision AABB query for scenes to accommodate the refactor.

## How Has This Been Tested

Added a new test scene with multiple objects meshes. Consumed this scene in an existing C++ test. Also ran viewer testing:

Before:
testscene_bug

After:
testscene_fixed

Removed test scene textures to reduce asset size from 1.5Mb to 11Kb. Use `./build/viewer --enable-physics --scene-requires-lighting -- "data/test_assets/scenes/simple_room.glb"` to view correctly:
image

## Checklist


src/tests/PhysicsTest.cpp
```diff
@@ -109,7 +109,7 @@ TEST_F(PhysicsManagerTest, JoinCompound) {
             Magnum::Matrix4::rotationX(Magnum::Math::Rad<float>(-1.56)) *
             Magnum::Matrix4::rotationY(Magnum::Math::Rad<float>(-0.25))};
         float boxHeight = 2.0 + (o * 2);
-        Magnum::Vector3 initialPosition{0.0, boxHeight, 0.0};
+        Magnum::Vector3 initialPosition{0.0, float(boxHeight + 1.25), 0.0};
```
mosra:
```suggestion
        Magnum::Vector3 initialPosition{0.0, boxHeight + 1.25f, 0.0};
```

achieves the same without temporarily going through doubles ;)

erikwijmans:
If it is textures, basis to the rescue?!

mosra:
A test file is small enough when there's nothing left to remove -- does the test need any textures at all?

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -36,34 +36,83 @@ bool BulletRigidObject::initializeScene(
...
+    bulletMesh.m_indexType = PHY_INTEGER;
+    bulletMesh.m_vertexType = PHY_FLOAT;
+    bSceneArrays_.push_back(std::make_unique<btTriangleIndexVertexArray>());
+    bSceneArrays_.back()->addIndexedMesh(bulletMesh,
+                                         PHY_INTEGER);  // exact shape
+
+    //! Embed 3D mesh into bullet shape
+    //! btBvhTriangleMeshShape is the most generic/slow choice
+    //! which allows concavity if the object is static
+    bSceneShapes_.emplace_back(std::make_unique<btBvhTriangleMeshShape>(
```
matthewjmay: all of the `.back/.get` calls in this PR make it harder to read. How about setting up the unique_ptr first, and then at the very end adding it using `bSceneShapes_.emplace_back(std::move(ptr))`. This pattern (set up first, add to vector last) is also good to follow because if there is an error/exception during setup, then the invalid object is never added to the vector.

aclegg3:
> A test file is small enough when there's nothing left to remove -- does the test need any textures at all?

With recent scene lighting options, textures are no longer required to view the scene mesh. I've stripped them. File size down to 11Kb. 😆

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: Good suggestion 👍

PR #516, opened 2020-03-03
Higher mesh coverage pose extraction for ImageExtractor

mpiseno:
## Motivation and Context

The data extraction in habitat sim is currently limited because it extracts 1 camera pose per camera position. In this PR I created another method for pose extraction, named panorama, in which for each camera position, we can extract multiple different poses by doing a simple 360 degree turn at each point. The purpose is to obtain a more complete mesh coverage and have more image data per scene to train models with.

## How Has This Been Tested

Added a simple test that uses the "panorama" method of pose extraction and asserts that the extractor extracted some poses using this method.

## Types of changes

## Example Output

Below are a series of image that were taken from the same camera position within Replica Apartment 1. With the other method of pose extraction, this camera position would only produce one image.

pan1
pan2
pan3
pan4
pan5
pan6
pan7


dhruvbatra:
Can you please post pictures of the kinds of frames you are able to extract? 

Also (just curious, don't know the right answer) -- would this be better captured as a new camera model in habitat-sim RGB sensor?

mpiseno:
> Can you please post pictures of the kinds of frames you are able to extract?
> 
> Also (just curious, don't know the right answer) -- would this be better captured as a new camera model in habitat-sim RGB sensor?

Sure no problem! As for the new camera model, that's something I can bring up, but the way I have it implemented currently is simple and works. Perhaps we could get a more complete coverage of the mesh with a new camera model. This will become more clear when I implement a metric to track how much of the mesh has actually been covered.

mpiseno:
> Can you please post pictures of the kinds of frames you are able to extract?
> 
> Also (just curious, don't know the right answer) -- would this be better captured as a new camera model in habitat-sim RGB sensor?

Updated the PR with some example output.

habitat_sim/utils/data/pose_extractor.py
```diff
@@ -28,8 +29,10 @@ def __init__(self, topdown_views, sim, pixels_per_meter=0.1):
         self.sim = sim
         self.pixels_per_meter = pixels_per_meter
 
-    def extract_poses(self, labels):
-        r"""Returns a numpy array of camera poses.
+    def extract_poses(self, labels, extraction_method):
+        r"""Returns a numpy array of camera poses. If extraction method is "closest", this method will
+        use bfs to find the closest point of interest to each camera position. If the extraction method
+        is "panorama", it will extract a full panorama of camera poses for each camera position
```
mathfac: When you mention panorama here, does that mean one image or a number of images taking turns? Let's make it explicit what to expect.

habitat_sim/utils/data/pose_extractor.py
```diff
@@ -28,8 +29,10 @@ def __init__(self, topdown_views, sim, pixels_per_meter=0.1):
         self.sim = sim
         self.pixels_per_meter = pixels_per_meter
 
-    def extract_poses(self, labels):
-        r"""Returns a numpy array of camera poses.
+    def extract_poses(self, labels, extraction_method):
```
mathfac: This is part of public API, let's add typing to the signature.

habitat_sim/utils/data/pose_extractor.py
```diff
@@ -174,3 +188,29 @@ def get_neighbors(p):
                     q.append((n, layer + step))
 
         return None, None
+
+    def _panorama_extraction(self, point, view, dist):
+        in_bounds_of_tdv = lambda row, col: 0 <= row < len(view) and 0 <= col < len(
+            view[0]
+        )
+        cpi_label_pairs = []
```
mathfac: What is `tdv`, `cpi` here? Let's make it verbose.

habitat_sim/utils/filesystem.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+
+def search_dir_tree_for_ext(mydir, ext):
```
mathfac: Let's add typing to the function signature.

mathfac:
If you have related unit test that would be great to add `panorama` options to testing.

PR #518, opened 2020-03-05
Tutorial for image data extraction

mpiseno:
## Motivation and Context

Added a tutorial to the doc/pages/ directory that explains how to use the ImageExtractor class in habitat sim to generate image data.

## Types of changes

### DIFF
--- a/docs/pages/image-extractor.rst
+++ b/docs/pages/image-extractor.rst
@@ -11,22 +11,29 @@ This notebook will go over how to use the Image Extraction API in Habitat Sim an
 
     import os
 
-    import numpy as np
     import matplotlib.pyplot as plt
+    import numpy as np
+    import torch
+    import torch.nn as nn
+    import torch.nn.functional as F
+    from torch.utils.data import DataLoader, Dataset
+
+    from habitat_sim.utils.data.data_extractor import ImageExtractor
+
 
     # Helper functions
     def display_sample(sample):
-        img = sample['rgba']
-        depth = sample['depth']
-        semantic = sample['semantic']
-        label = sample['label']
+        img = sample["rgba"]
+        depth = sample["depth"]
+        semantic = sample["semantic"]
+        label = sample["label"]
 
         arr = [img, depth, semantic]
-        titles = ['rgba', 'depth', 'semantic']
-        plt.figure(figsize=(12 ,8))
+        titles = ["rgba", "depth", "semantic"]
+        plt.figure(figsize=(12, 8))
         for i, data in enumerate(arr):
-            ax = plt.subplot(1, 3, i+1)
-            ax.axis('off')
+            ax = plt.subplot(1, 3, i + 1)
+            ax.axis("off")
             ax.set_title(titles[i])
             plt.imshow(data)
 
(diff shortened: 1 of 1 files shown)


docs/pages/image-extractor.rst
```diff
@@ -0,0 +1,158 @@
...
+        def __init__(self, extractor):
+            self.extractor = extractor
+
+        def __len__(self):
+            return len(self.extractor)
+
+        def __getitem__(self, idx):
+            sample = self.extractor[idx]
+            output = {
+                'rgba': sample['rgba'].astype(np.float32), # dataloader requires certain types
```
erikwijmans:
```suggestion
                'rgb': sample['rgba'].astype(np.float32)[..., :-1], # dataloader requires certain types
```
Most people will probably want rgb, not rgba

mpiseno: right but the color sensor outputs rgba so wouldn't it been more accurate to keep it as rgba? That was the same argument for changing the output name from "rgb" to "rgba" in data_extractor.py a while back.

erikwijmans: Naming it `rgba` was needed as this is rgba (it would have also been fine to convert rgba to rgb and then name it rgb).  Most people will want to train models to operate on rgb not rgba (that is the standard thing output by cameras) so it is useful to who them how to convert from rgb to rgba (it isn't strictly true that you can just chop off alpha to go from rgba to rgb, but it works in this case).

docs/pages/image-extractor.rst
```diff
@@ -0,0 +1,158 @@
...
+        def __init__(self):
+            super(TrivialNet, self).__init__()
+
+        def forward(self, x):
+            x = F.relu(x)
+            return x
+
+
+    dataset = HabitatDataset(extractor)
+    dataloader = DataLoader(dataset, batch_size=2)
```
erikwijmans: Worth mentioning that `num_workers` must be set to `0` currently.

docs/pages/image-extractor.rst
```diff
@@ -0,0 +1,158 @@
...
+"label" which is the class label (specified by the user in the constructor) of the image. Note:
+for the scene in this example there is no semantic data which is why the semantic output
+below is not represented.
+
+.. code:: py
+
+    from habitat_sim.utils.data.data_extractor import ImageExtractor
+
+    # Give the extractor a path to the scene
+    scene_filepath = "../data/scene_datasets/habitat-test-scenes/skokloster-castle.glb"
```
erikwijmans:
```suggestion
    scene_filepath = "data/scene_datasets/habitat-test-scenes/skokloster-castle.glb"
```

docs/pages/image-extractor.rst
```diff
@@ -0,0 +1,158 @@
...
+        def __init__(self, extractor):
+            self.extractor = extractor
+
+        def __len__(self):
+            return len(self.extractor)
+
+        def __getitem__(self, idx):
+            sample = self.extractor[idx]
+            output = {
+                'rgb': sample['rgba'].astype(np.float32)[..., :-1], # dataloader requires certain types
```
erikwijmans: Since you converting to float, also a good idea to divide by `255.0` (people expect float images to have pixel values in [0, 1])

PR #519, opened 2020-03-05
[refactor] GeoTest in Corrade TestSuite

bigbike:
## Motivation and Context
Keep my promise to refactor the GeoTest, migrating it to Corrade::TestSuite.
If I am not going to do it now, perhaps nobody will do it in the future.

## Types of changes

## Checklist


src/tests/AabbTest.cpp
```diff
@@ -107,6 +113,97 @@ void GeoTest::aabb() {
...
+    CORRADE_VERIFY(c1.gravity().isApprox(-up));
+    CORRADE_VERIFY(c1.front().isApprox(front));
+    CORRADE_VERIFY(c1.back().isApprox(-front));
+    CORRADE_VERIFY(c1.up().isApprox(rotation * ESP_UP));
+    CORRADE_VERIFY(c1.front().isApprox(rotation * ESP_FRONT));
+    CORRADE_VERIFY(c1.origin().isApprox(origin));
+    CORRADE_VERIFY(c1.rotationWorldToFrame().isApprox(rotation));
+
+    CoordinateFrame c2(rotation, origin);
+    CORRADE_COMPARE(c1, c2);
```
bigbike: @mosra: Your doc says item must be printable via Debug, which is NOT the case here. How can I fix this problem at this point?

bigbike: Part of the error message reads:
`habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Comparator.h:270:7: error: invalid operands to binar
y expression ('Corrade::Utility::Debug' and 'const std::__1::basic_string<char>')
      << *actualValue << Utility::Debug::newline << "        but expected\n       " << *expectedValue;
      ^  ~~~~~~~~~~~~
/Users/user/habitat-sim/src/deps/corrade/src/Corrade/TestSuite/Comparator.h:326:16: note: in instantiation of membe
r function 'Corrade::TestSuite::Comparator<std::__1::basic_string<char> >::printMessage' requested here
    comparator.printMessage(flags, out, actual, expected);`

mosra: [`#include <Corrade/Utility/DebugStl.h>`](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html#Utility-Debug-stl)

src/tests/AabbTest.cpp
```diff
@@ -107,6 +113,97 @@ void GeoTest::aabb() {
...
+  CORRADE_VERIFY(!obb2.contains(vec3f(5, 0, 2)));
+  CORRADE_VERIFY(!obb2.contains(vec3f(-10, 0.5, -2)));
+
+  const box3f aabb = obb2.toAABB();
+  CORRADE_COMPARE_AS(aabb.min().x(), -10, float);
+  CORRADE_COMPARE_AS(aabb.min().y(), -5, float);
+  CORRADE_COMPARE_AS(aabb.min().z(), -1, float);
+  CORRADE_COMPARE_AS(aabb.max().x(), 10, float);
+  CORRADE_COMPARE_AS(aabb.max().y(), 5, float);
+  CORRADE_COMPARE_AS(aabb.max().z(), 1, float);
```
mosra:
```suggestion
  CORRADE_COMPARE(Mn::Vector3{aabb.min()}, (Mn::Vector3{-10.0f, -5.0f, -1.0f}));
  CORRADE_COMPARE(Mn::Vector3{aabb.max()}, (Mn::Vector3{10.0f, 5.0f, 1.0f}));
```

I.e., converting Eigen type to Magnum first and then using the convenient builtin comparisons (don't forget to [`#include <Magnum/EigenIntegration/Integration.h>`](https://doc.magnum.graphics/magnum/integration_2src_2Magnum_2EigenIntegration_2Integration_8h.html) for those). Alternatively

```cpp
  CORRADE_VERIFY(aabb.min().isApprox(vec3f{-10.0f, -5.0f, -1.0f}));
  CORRADE_VERIFY(aabb.max().isApprox(vec3f{10.0f, 5.0f, 1.0f}));
```

This is a bit of a clash between Magnum and Eigen types and I'm not sure what's better -- would the OBB class benefit from directly using Magnum types instead of Eigen? Or would that make it harder to use in certain contexts?

src/tests/AabbTest.cpp
mosra: For the actual `CoordinateFrame` comparison above, unless you'd want to write an `operator<<(Mn::Debug&, const CoordinateFrame&)` for it (which might be only useful for this particular test and nowhere else), the easiest would be to do just

```suggestion
    CORRADE_VERIFY(c1 == c2);
```

That way you don't get any useful diagnostic when the two aren't equal, but you didn't get that with gtest either.

bigbike:
It is ready to be reviewed again. Thanks.

src/tests/AabbTest.cpp
erikwijmans: I think we expose enough of OBB/AABB to python to keep it Eigen as that becomes numpy on the other end.  

My opinion is that we should keep things as eigen/numpy as much as we can as python users will expect numpy as the vector type.  Magnum types do play very well with numpy, so when there is a performance concern due to needing to transfer between numpy and magnum, we should use magnum types, but I don't think that applies here.

src/tests/GeoTest.cpp
```diff
@@ -2,64 +2,165 @@
...
+#include <Corrade/TestSuite/Tester.h>
+#include <Corrade/Utility/DebugStl.h>
+#include <Magnum/EigenIntegration/Integration.h>
+#include <Magnum/Math/FunctionsBatch.h>
+#include "esp/core/Utility.h"
 #include "esp/geo/CoordinateFrame.h"
 #include "esp/geo/OBB.h"
 #include "esp/geo/geo.h"
 
+// TODO: refactor the GeoTest to use Cr::TestSuite
```
mosra: This TODO is now done, I guess? :)

bigbike:
ready to be merged in.

bigbike:
Also do you think I need to disable the benchmark in the CI test? Is so, how?

mosra:
> do you think I need to disable the benchmark in the CI test

Why do *you* think it would need to be disabled? ;) As I said in https://github.com/facebookresearch/habitat-sim/pull/501#issuecomment-592422053 before, unless the benchmarks take tens of seconds to the point of significantly slowing down the CI builds, I would keep them there. Having benchmark numbers preserved "forever" in the CI logs helps tracking down accidental performance regressions, among other things.

OTOH, the CI log is quite noisy and somehow it doesn't include colors anymore, so digging out the actual numbers isn't so easy :/ I'm pretty sure it did have colors back when I introduced my first tests/benchmarks, for both TestSuite and gtest (see "Run sim tests" [in this build](https://circleci.com/gh/facebookresearch/habitat-sim/1424)) -- @mathfac @aclegg3 do you remember changing anything that would break it? Or did circleci change to not include the colors anymore? Also, could we try to increase the signal/noise ratio in the tests? Some tests have quite excessive unnecessary logging.

dhruvbatra:
> As I said in #501 (comment) before, unless the benchmarks take tens of seconds to the point of significantly slowing down the CI builds, I would keep them there. Having benchmark numbers preserved "forever" in the CI logs helps tracking down accidental performance regressions, among other things.

Completely agree.

bigbike:
I am totally fine to keep the benchmark in the CI tests. Actually I prefer to do so.
The reason I table it again before committing it to the master is that I found it is really a bit "slow" compared to the other tests. So I was not sure at that time.

PR #521, opened 2020-03-10
Semantic mesh optimization - split by objectID and cull

matthewjmay:
## Motivation and Context

Previously, semantic meshes were loaded in as a single mesh, making them impossible to cull. This PR leverages the spatial locality of ObjectIDs to split the single mesh into multiple different meshes by object ID, calculate AAbbs, and enables culling on this semantic mesh. See below for a visualization of how this works:
test_image_save

Full benchmarks are currently being ran.
Some example scene A/B tests benchmarking culling on semantic only, and overall rgbd + semantic:

```
Scene 1:

 ================ Performance (FPS) NPROC=1 ===================================
Resolution      semantic_only                   rgbd_semantic
128 x 128       286.1 /803.2  ( 180.8%) 212.9 /440.6  ( 106.9%)
256 x 256       279.3 /763.5  ( 173.4%) 210.8 /399.2  (  89.4%)
512 x 512       257.0 /613.7  ( 138.8%) 178.9 /291.2  (  62.8%)
 ==============================================================================
 ================ Performance (FPS) NPROC=3 ===================================
Resolution      semantic_only                   rgbd_semantic
128 x 128       374.0 /1479.2 ( 295.5%) 292.2 /786.5  ( 169.1%)
256 x 256       369.6 /1439.2 ( 289.4%) 285.9 /739.8  ( 158.8%)
512 x 512       344.6 /1355.1 ( 293.2%) 270.9 /649.2  ( 139.7%)
 ==============================================================================
 ================ Performance (FPS) NPROC=5 ===================================
Resolution      semantic_only                   rgbd_semantic
128 x 128       379.3 /1575.9 ( 315.5%) 310.5 /909.9  ( 193.0%)
256 x 256       376.5 /1534.8 ( 307.6%) 305.4 /860.3  ( 181.7%)
512 x 512       376.2 /1535.6 ( 308.2%) 297.5 /802.2  ( 169.7%)
 ==============================================================================

Scene 2:

 ================ Performance (FPS) NPROC=1 ===================================
Resolution      semantic_only                   rgbd_semantic
128 x 128       1664.5/2583.4 (  55.2%) 609.7 /769.7  (  26.2%)
256 x 256       1496.5/2296.9 (  53.5%) 564.4 /627.5  (  11.2%)
512 x 512       1022.6/1395.7 (  36.5%) 375.1 /409.7  (   9.2%)
 ==============================================================================
 ================ Performance (FPS) NPROC=3 ===================================
[...]

PR #523, opened 2020-03-10
csv output for ab test

matthewjmay:
## Motivation and Context

Optionally output csv file during ab test for easier processing later.

## How Has This Been Tested

Eg of output:

```
num_procs,resolution,sensor_types,fps_test,fps_control
1,128,rgb,1220.180686527348,433.6937190315082
1,128,rgbd,646.3853434963924,225.71452249824296
1,128,depth_only,1536.040313558034,529.8223253269938
1,128,semantic_only,1071.0588643870651,113.05412282733961
1,128,rgbd_semantic,635.9399827759786,97.00690401337553
1,256,rgb,2033.1829804767501,1308.7303209339552
1,256,rgbd,1091.2095574079278,639.8652627467972
1,256,depth_only,1095.6735265382465,436.0366373137765
1,256,semantic_only,925.8233911287207,113.08794255527332
1,256,rgbd_semantic,485.2895269984924,91.28888760233752
```

## Types of changes

## Checklist

### DIFF
--- a/examples/ab_test.py
+++ b/examples/ab_test.py
@@ -95,10 +95,17 @@ parser.add_argument(
 )
 
 parser.add_argument("--csv", type=str, help="csv output file")
+parser.add_argument(
+    "--speedup", action="store_true", help="display speedup instead of percent change"
+)
 
 
-def get_percent_diff(test_val, control_val):
-    return ((test_val - control_val) / control_val) * 100.0
+def get_percent_diff_str(test_val, control_val):
+    return f"{((test_val - control_val) / control_val) * 100.0:.1f}%"
+
+
+def get_speedup_str(test_val, control_val):
+    return f"{test_val / control_val:.1f}x"
 
 
 def seconds_to_ms(seconds):
@@ -106,7 +113,12 @@ def seconds_to_ms(seconds):
 
 
 def print_metric(
-    performance_data, resolutions, title_list, metric="fps", metric_transformer=None
+    performance_data,
+    resolutions,
+    title_list,
+    metric="fps",
+    comparison_label_generator=None,
+    metric_transformer=None,
 ):
     for nproc, performance in performance_all.items():
         header = f" Performance ({metric}) NPROC={nproc} "
@@ -121,14 +133,12 @@ def print_metric(
             for t in title_list:
                 control_metric = perf[t][dr.ABTestGroup.CONTROL][metric]
                 test_metric = perf[t][dr.ABTestGroup.TEST][metric]
-                percent_diff = get_percent_diff(test_metric, control_metric)
+                comparison_str = comparison_label_generator(test_metric, control_metric)
                 if metric_transformer:
                     control_metric = metric_transformer(control_metric)
                     test_metric = metric_transformer(test_metric)
 
-                row += (
-                    f"{control_metric:6.1f}/{test_metric:6.1f} ({percent_diff:>6.1f}%)"
-                )
+                row += f"{control_metric:6.1f}/{test_metric:6.1f} ({comparison_str:>6})"
             print(row)
         print(f"{' END ':=^100}")
 
@@ -261,13 +271,21 @@ for nprocs in nprocs_tests:
 
     performance_all[nprocs] = performance
 
-print_metric(performance_all, resolutions, title_list, metric="fps")
+comparison_label_generator = get_speedup_str if args.speedup else get_percent_diff_str
+print_metric(
+    performance_all,
+    resolutions,
+    title_list,
+    metric="fps",
+    comparison_label_generator=comparison_label_generator,
+)
 print_metric(
     performance_all,
     resolutions,
     title_list,
     metric="frame_time",
     metric_transformer=seconds_to_ms,
+    comparison_label_generator=comparison_label_generator,
 )
 if args.enable_physics:
     print_metric(performance_all, resolutions, title_list, metric="avg_sim_step_time")
--- a/examples/ab_test.py
+++ b/examples/ab_test.py
@@ -6,6 +6,7 @@
 
 
 import argparse
+import csv
 import distutils
 from distutils import util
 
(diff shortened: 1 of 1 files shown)


mosra:
I just noticed this in #521 -- can the output have frame time as well? Or, just the frame time alone? Because measuring just FPS alone has several disadvantages, one of them is that going from 20 to 30 FPS isn't the same as going from 80 to 90, and comparing such numbers is *hard*. 

I did a similar change for a different code quite a while ago in #149, but it looks like the know-how from there got forgotten again :)

dhruvbatra:
>  can the output have frame time as well? Or, just the frame time alone? Because measuring just FPS alone has several disadvantages, one of them is that going from 20 to 30 FPS isn't the same as going from 80 to 90, and comparing such numbers is hard.

Agree with Mosra.

bigbike:
My opinion:
-) We will need to keep the FPS. From the audience perspective, It is the most intuitive, straightforward, easy to understand metrics for rendering speed. (you always hear e.g., XXX game can run at 4K 60 FPS on XXX Graphics card. you never hear XXX game can run at 4K with 0.016667 sec per frame.)

>one of them is that going from 20 to 30 FPS isn't the same as going from 80 to 90, and comparing such numbers is hard.

-) When looking at FPS, we never ever simply look at the absolute FPS increase. We compute the increase rate (or speedup) in percentage. 20-->30 it means the FPS is improved by 50%, while in 80 to 90 case,  it is improved by 12.5%. So the comparison is easy. (in AB_test.py, such percentage will be displayed to the user).

-) FPS is just the reciprocal of the frame time, if I understand it correctly. So I am not sure if there is obvious advantage to use frame time over FPS.

dhruvbatra:
Agree with @bigbike 's points. To be clear, I support frame time in addition to FPS, not instead of.

bigbike:
For this diff:
By exporting the results in csv format, it is easy to be loaded to e.g., excel. Correct? Smart idea.

examples/ab_test.py
```diff
@@ -93,6 +94,70 @@
...
+        for idx in range(len(performance)):
+            row = "%d x %d" % (resolutions[idx], resolutions[idx])
+            # break down by benchmark items
+            for t in title_list:
+                control_fps = performance[idx][t + "_" + str(dr.ABTestGroup.CONTROL)][
+                    "fps"
+                ]
+                test_fps = performance[idx][t + "_" + str(dr.ABTestGroup.TEST)]["fps"]
+                ratio = (
+                    (float(test_fps) - float(control_fps)) / float(control_fps) * 100.0
```
bigbike: I know it is from my original code. But I am now thinking it twice.

To everyone:
Do you think it is better to just compute the speedup instead of the increase rate?
namely:
`speedup = (float(test_fps) / float(control_fps) * 100.0`
According to my current experience, `speedup` is more intuitive (since we always report it as e.g., the speedup is 3.6x)

matthewjmay:
@bigbike, Excel yes, but even better is we can load the data into [pandas](https://pandas.pydata.org/pandas-docs/stable/index.html). This easily allows us to do things like compute averages, combine outputs of **multiple** benchmarks programmatically (my current use case) and do things like filter specific resolutions/nprocs/etc, get [nsmallest](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.nsmallest.html), [nlargest](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.nlargest.html), and more. This also helps the fps vs. frame_time, and % difference vs. speedup debates, since if the raw data is exported as csv, we can easily compute these derived metrics

examples/ab_test.py
matthewjmay: See my below comment. The important thing is we output as an intermediate format. Here's how to compute all metrics everyone wants very easily once this is merged:

```
>>> import pandas as pd
>>> results = pd.read_csv("~/habitat-sim/semantic_bench.csv")
>>> print(results)
   num_procs  resolution   sensor_types    fps_test  fps_control
0          1         128  semantic_only  703.088740   108.107696
1          1         128  rgbd_semantic  173.315447    89.380839
2          1         256  semantic_only  546.703172   104.385620
3          1         256  rgbd_semantic  358.973824    93.043743
>>> results["frame_time_test_ms"] = 1000 / results["fps_test"]
>>> results["frame_time_control"] = 1000 / results["fps_control"]
>>> results["speedup"] = results["fps_test"] / results["fps_control"]
>>> print(results)
   num_procs  resolution   sensor_types    fps_test  fps_control  frame_time_test_ms  frame_time_control   speedup
0          1         128  semantic_only  703.088740   108.107696            1.422296            9.250035  6.503596
1          1         128  rgbd_semantic  173.315447    89.380839            5.769826           11.188080  1.939067
2          1         256  semantic_only  546.703172   104.385620            1.829146            9.579864  5.237342
3          1         256  rgbd_semantic  358.973824    93.043743            2.785718           10.747633  3.858119
>>> filtered_256_px = results[results["resolution"]==256]
>>> print(filtered_256_px)
   num_procs  resolution   sensor_types    fps_test  fps_control  frame_time_test_ms  frame_time_control   speedup
2          1         256  semantic_only  546.703172   104.385620            1.829146            9.579864  5.237342
3          1         256  rgbd_semantic  358.973824    93.043743            2.785718           10.747633  3.858119
>>> avg_256_speedup = filtered_256_px["speedup"].mean()
>>> print(f"avg_256_speedup={avg_256_speedup}")
avg_256_speedup=4.54773034562938
```

bigbike:
Awesome. I would advise that you upload your python code as well to somewhere in the repository.

examples/ab_test.py
```diff
@@ -93,6 +94,73 @@
...
+            for t in title_list:
+                control_metric = perf[t][dr.ABTestGroup.CONTROL][metric]
+                test_metric = perf[t][dr.ABTestGroup.TEST][metric]
+                percent_diff = get_percent_diff(test_metric, control_metric)
+                if metric_transformer:
+                    control_metric = metric_transformer(control_metric)
+                    test_metric = metric_transformer(test_metric)
+
+                row += (
+                    f"{control_metric:6.1f}/{test_metric:6.1f} ({percent_diff:>6.1f}%)"
```
bigbike: Like I mentioned in the previous message. Do you think it is more reasonable (useful) to report speedup instead of percent_diff?

bigbike: Let us add an option to the program, "--speedup".
By default, we use percent_diff, otherwise display "speedup" (e.g., 3.6x) if "--speedup" is specified.

PR #525, opened 2020-03-12
New sensor that outputs unique ID for each triangle in mesh

mpiseno:
## Motivation and Context

Created a new sensor that outputs a unique ID for each triangle in the mesh. Each pixel of the output contains the value of the triangle ID that the pixel is aligned with. This sensor will become useful for creating a metric that reports how much of the total mesh a set of images covers.

## How Has This Been Tested

I'm just creating this PR for code feedback right now. I'm open to testing suggestions.

## Types of changes

# Example Output

In these visuals it's perhaps a bit hard to see, but each triangle in the mesh is given a distinct color based on its unique id.

Screen Shot 2020-03-16 at 8 39 27 PM
Screen Shot 2020-03-16 at 8 39 50 PM
Screen Shot 2020-03-16 at 8 40 08 PM
Screen Shot 2020-03-16 at 8 40 49 PM


mathfac:
@mpiseno, that's great! is it possible to visualize with lut colors  and attach picture to the PR?

mathfac:
Looks like `viewer_demo.visual.test.js` should be updated.

erikwijmans:
IIRC, WebGL does not support `gl_PrimitiveID` -- @mpiseno you will need to IFDEF around all this to disable it for the webgl build.  

The better option is to make enabling this a specific build flag (and then make sure it and webgl are mutually exclusive).  The performance impact of adding this extra output isn't huge, but it isn't zero either and I expect the use-case for it is likely quite niche, so no need to enable it by default.  It could also be a runtime flag instead of a build flag if you'd prefer to do that (the OpenGL shaders get compiled at runtime, so those will still be built with or without this regardless).

mosra:
@erikwijmans It generally does, except for SwiftShader (software renderer) and really old ANGLE (not in use anymore IIRC) 

@mpiseno  you can check for `gl_VertexID` with the [`MAGNUM_shader_vertex_id`](https://doc.magnum.graphics/magnum/opengl-support.html#opengl-support-webgl20-extensions) pseudo-extension and print a warning / abort in case it's not supported (or alternatively provide it via an additional vertex attribute).

erikwijmans:
@mosra IIRC, we changed the semantic shading logic in this PR: https://github.com/facebookresearch/habitat-sim/pull/191 to resolve the `gl_PrimitiveID` for WebGL.

mosra:
Yes, that was it. Sorry for the noise again.

erikwijmans:
> Yes, that was it. Sorry for the noise again.

No problem, always worth making me actually check to see if I am correct if your recollection is different.  I thought I was correct, but didn't bother to actually check and you made me actually check :)

tests/test_data_extraction.py
```diff
@@ -82,3 +82,10 @@ def test_data_extractor_all_modes(sim):
         extraction_method="panorama",
     )
     assert len(extractor) > 0
+
+
+def test_triangle_id(sim):
+    ## test that empty frames can be rendered without a scene mesh
+    obs = sim.step("move_forward")
+    triangle = obs["triangle_sensor"]
```
mathfac: you can freeze average value of `obs["triangle_sensor"]`

msbaines:
> IIRC, WebGL does not support `gl_PrimitiveID` -- @mpiseno you will need to IFDEF around all this to disable it for the webgl build.
> 
> The better option is to make enabling this a specific build flag (and then make sure it and webgl are mutually exclusive). The performance impact of adding this extra output isn't huge, but it isn't zero either and I expect the use-case for it is likely quite niche, so no need to enable it by default. It could also be a runtime flag instead of a build flag if you'd prefer to do that (the OpenGL shaders get compiled at runtime, so those will still be built with or without this regardless).

mosra:
Hmm, this could get supported on WebGL by deindexing the mesh and then using `gl_VertexID/3` instead. Latest Magnum has a [convenient tool for that](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#ace8c9f5b375324c3c6ba38032d3c1c9f) so when I'm done porting to it, this could be done.

*(I had an evil master plan to completely remove the PrimitiveID shader in process of upgrading to the new `MeshData` APIs in Magnum but this PR breaks that plan as the builtin Magnum shaders don't support emitting triangle IDs ... yet. So maybe I'll add that there too.)*

habitat_sim/simulator.py
```diff
@@ -389,6 +393,14 @@ def __init__(self, sim, agent, sensor_id):
                     (self._spec.resolution[0], self._spec.resolution[1]),
                     dtype=np.float32,
                 )
+            elif self._spec.sensor_type == hsim.SensorType.TRIANGLE:
+                # self._buffer = np.empty(
+                #     (self._spec.resolution[0], self._spec.resolution[1]),
+                #     dtype=np.uint32,
+                # )
+                self._buffer = np.ones(
```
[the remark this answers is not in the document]
mpiseno: It does, I just forgot to switch it back. Fixing this now.

habitat_sim/simulator.py
```diff
@@ -422,7 +434,10 @@ def draw_observation(self):
             )
 
         # get the correct scene graph based on application
-        if self._spec.sensor_type == hsim.SensorType.SEMANTIC:
+        if self._spec.sensor_type in (
+            hsim.SensorType.SEMANTIC,
+            hsim.SensorType.TRIANGLE,
+        ):
```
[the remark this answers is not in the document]
mpiseno: The triangle sensor uses the same shaders as the semantic sensors, so all the plumbing I did for build the triangle sensor is based on the semantic mesh.

tests/test_data_extraction.py
mpiseno: Do you mean save the average value of the sensor at some state s0, then during testing set the state to s0 again and see if we get the same average value of the sensor?

mpiseno:
> IIRC, WebGL does not support `gl_PrimitiveID` -- @mpiseno you will need to IFDEF around all this to disable it for the webgl build.
> 
> The better option is to make enabling this a specific build flag (and then make sure it and webgl are mutually exclusive). The performance impact of adding this extra output isn't huge, but it isn't zero either and I expect the use-case for it is likely quite niche, so no need to enable it by default. It could also be a runtime flag instead of a build flag if you'd prefer to do that (the OpenGL shaders get compiled at runtime, so those will still be built with or without this regardless).

I just made it a build flag, but one thing I just thought about was how can I write a test for the sensor now? If I build with the triangle sensor we have the webgl issue, but if I build without the triangle sensor I can't really test it.

tests/test_data_extraction.py
mathfac: Just do `np.avg(triangle) == TEST_AVG_TRIANGLE_VALUE` in the test. Then test will fail output of the sensor will change because of some code changes. Then we will know that all functionality depend on triangles sensor will be affected.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -111,6 +130,7 @@ struct RenderTarget::Impl {
     framebuffer_.clearDepth(1.0);
     framebuffer_.clearColor(0, Mn::Color4{0, 0, 0, 1});
     framebuffer_.clearColor(1, Mn::Vector4ui{});
+    framebuffer_.clearColor(2, Mn::Vector4i{});
```
msbaines: #ifdef here

src/esp/gfx/RenderTarget.cpp
```diff
@@ -234,6 +286,7 @@ struct RenderTarget::Impl {
   Mn::GL::Renderbuffer objectIdBuffer_;
   Mn::GL::Texture2D depthRenderTexture_;
   Mn::GL::Framebuffer framebuffer_;
+  Mn::GL::Renderbuffer triangleIdBuffer_;
```
msbaines: #ifdef here

src/esp/gfx/RenderTarget.cpp
```diff
@@ -244,6 +297,7 @@ struct RenderTarget::Impl {
 #ifdef ESP_BUILD_WITH_CUDA
   cudaGraphicsResource_t colorBufferCugl_ = nullptr;
   cudaGraphicsResource_t objecIdBufferCugl_ = nullptr;
+  cudaGraphicsResource_t triangleIdBufferCugl_ = nullptr;
```
msbaines: #ifdef here

src/shaders/primitive-id-textured-gl410.frag
```diff
@@ -1,8 +1,10 @@
 in mediump vec3 v_color;
 flat in mediump uint v_objectId;
+in int gl_PrimitiveID;
```
msbaines: We need an #ifdef in the shared or two separate shaders with an #ifdef choosing which shader to use.

erikwijmans:
@mpiseno 
>  how can I write a test for the sensor now

Same way we do the CUDA tests.  Issue another build command (should be quick because of ccache), then run just that test.

mpiseno:
> @mpiseno
> 
> > how can I write a test for the sensor now
> 
> Same way we do the CUDA tests. Issue another build command (should be quick because of ccache), then run just that test.

Got that working fine. The triangle sensor requires a scene with a semantic mesh right now. Are there any mp3d scenes on the machines that run the circleci tests? Looking at tests/test_semantic_scene.py it looks like it references some mp3d scenes but it skips these tests because the navmesh file doesn't exist.

eundersander:
Closing this stale PR.

PR #528, opened 2020-03-16
Lighting tutorial

matthewjmay:
## Motivation and Context

Users need to know how to use new lighting abilities in habitat.

Added `examples/tutorials` directory to hold stand-alone tutorials.

Supporting changes:
 - `stereo_agent.py` and `new_actions.py` moved into examples/tutorials directory.
 - added a new chair test asset to remove external dataset download dependency from tutorial.

## How Has This Been Tested

Example code run in CI `test_examples.py`

Build docs with `./build.sh` in the `habitat_sim/docs` directory.
After build, view the new tutorial at: `/habitat-sim/build/docs/habitat-sim/lighting-setups.html`

## Types of changes

## Checklist


bigbike:
Looks good to me.
@aclegg3: Would you please take a look? See if I missed anything. Thanks.

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,180 @@
...
+
+By default, added objects will be phong shaded with lights at the corners of the scene.
+
+.. code:: py
+
+    id_1 = sim.add_object(1)
+    sim.set_translation(agent_transform.transform_point([0.3, 0.9, -1.8]), id_1)
+
+    show_obs(sim)
+.. image:: default_object_lighting.png
```
mosra: There needs to be a newline before, otherwise the page won't compile. Same in the four cases below.

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,180 @@
...
+
+To use this a light setup, pass in the name as a parameter to `Simulator.add_object`.
+
+.. code:: py
+
+    remove_all_objects(sim)
+
+    id_1 = sim.add_object(2, light_setup_key="my_custom_lighting")
+    sim.set_rotation(mn.Quaternion.rotation(mn.Rad(mn.Deg(90)), mn.Vector3.x_axis()) *
+                     mn.Quaternion.rotation(mn.Rad(mn.Deg(-115)), mn.Vector3.z_axis()), id_1)
```
mosra:
```suggestion
    sim.set_rotation(mn.Quaternion.rotation(mn.Deg(90), mn.Vector3.x_axis()) *
                     mn.Quaternion.rotation(mn.Deg(-115), mn.Vector3.z_axis()), id_1)
```

just this should work, same below and in one other case above

.circleci/config.yml
```diff
@@ -341,6 +341,10 @@ jobs:
               # run tests with code coverage
               CORRADE_TEST_COLOR=ON GTEST_COLOR=yes ./build.sh --headless --bullet --with-cuda --run-tests --cmake-args='-DCMAKE_CXX_FLAGS="--coverage"'
 
+              #with bullet installed, run tutorial validation
+              #TODO: loop for all tutorials in the directory?
+              python examples/tutorials/lighting_tutorial.py
```
erikwijmans, line 346: Please see https://github.com/facebookresearch/habitat-sim/blob/master/tests/test_examples.py for how to test things like this.

aclegg3: What do you think about moving `stereo_agent.py` and `new_actions.py` into `examples/tutorials` to separate stand-alone tutorial scripts from the example demo_runner?

erikwijmans: That is fine by me :)

aclegg3:
Major development is done here. Check it out and let me know what you think. 😀

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
+Working with Lights
```
dhruvbatra, line 1: See comment on #habitat-dev about ipynb files.

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+###################
+
+:summary: This tutorial demonstrates the creation and manipulation of LightSetups in Habitat-sim.
```
dhruvbatra, line 4: We should make this sentence broadly accessible. 

This tutorial demonstrates lighting of scenes and objects in Habitat -- creation, placement, configuration, manipulation of light sources and shaders.

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+
+.. contents::
+    :class: m-block m-default
+
+A **LightSetup** consists of a set of LightInfo structures defining the common configuration
```
dhruvbatra, line 9: If LightSetup is a class, should it link to it's class description page?

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+Each **LightInfo** structure in a LightSetup defines the `color`, `position`, and `LightPositionModel` of a single point light source.
+The LightPositionModel defines the coordinate frame of the light and can be set to any of:
+
+.. code:: python
+
+    habitat_sim.gfx.LightPositionModel.CAMERA
+    habitat_sim.gfx.LightPositionModel.GLOBAL
+    habitat_sim.gfx.LightPositionModel.OBJECT
+
+Each LightSetup is registered in the simulator via a unique key. Two default lighting setups are pre-defined:
```
dhruvbatra, line 22: I feel like this entire block just started diving into details of an object/class without telling me what I should use this for. 

I like the sections below much better. They start with a use case -- suppose you want to light a scene, here's what you should do. Suppose you want to light an object, here's what you should do. Could we re-order things to start with use-cases?

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+
+
+`Working with Lights Example:`_
+===============================
+
+The example code below demonstrates the usage of default and custom LightSetups and is runnable via:
+
+.. code:: shell-session
+
+    $ python examples/tutorials/lighting_tutorial.py
```
dhruvbatra, line 54: Oh wow, I didn't realize things were so "clean" that we can just call an existing python file. Nice setup. I wish the integration with ipynb was just as clean.

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+    sim.set_light_setup(new_light_setup, "my_custom_lighting_key")
+
+Any existing LightSetup can be queried with **simulator.get_light_setup**:
+
+.. code:: python
+
+    custom_light_setup = sim.set_light_setup(n"my_custom_lighting_key")
+
+
+`Working with Lights Example:`_
```
dhruvbatra, line 47: Similar comment as above -- could we state what goal is this lights example trying to accomplish?

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+.. code:: shell-session
+
+    $ python examples/tutorials/lighting_tutorial.py
+
+First, we import modules we will need, define some convenience functions, and initialize the Simulator, Agent, and objects.
+
+.. include:: ../../examples/tutorials/lighting_tutorial.py
+    :code: py
+    :start-after: # [setup]
+    :end-before: # [/setup]
```
dhruvbatra, line 61: It gets better! This is awesome! I wish we could similarly refer to cells in an ipynb. 

@mosra -- wish list :-).

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+
+`Scene Lighting`_
+=================
+
+By default, the scene will be shaded with no lights using the **habitat_sim.gfx.NO_LIGHT_KEY** LightSetup.
+This configuration is ideal for scene assets with illumination baked into textures (e.g. building scans such as MP3D).
```
dhruvbatra, line 67: Awesome!

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+
+.. include:: ../../examples/tutorials/lighting_tutorial.py
+    :code: py
+    :start-after: # [default scene lighting]
+    :end-before: # [scene swap shader]
+
+.. image:: images/lighting-setups-images/0.jpg
+    :width: 45%
+
+To use a custom light setup for the scene, edit the **habitat_sim.SimulatorConfiguration.scene_light_setup**
```
dhruvbatra, line 77: This is great! This right here is a use-case. 

Could you expand on this and say why someone might want to use a different light-setup? (Presumably because not all 3D scans have lighting baked in; anything else?)

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+option when creating/reconfiguring your Simulator. Note that
+while you can modify the scene's LightSetup dynamically during runtime, you will need to
+reconfigure the simulator to switch the scene's LightSetup key. Also, due to asset loading specifics
+**you must close and re-initialize the Simulator to swap between Flat and Phong shading setups**.
```
dhruvbatra, line 81: Ouch :-(, but beyond the scope of tutorial.

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+    :start-after: # [scene swap shader]
+    :end-before: # [/scene]
+
+
+
+.. |im1| image:: images/lighting-setups-images/1.jpg
+   :width: 100%
+
+.. |im2| image:: images/lighting-setups-images/2.jpg
+   :width: 100%
```
dhruvbatra, line 94: Great example images!

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
++---------+-----------+
+|  |im1|  +   |im2|   +
+|  |cp1|  +   |cp2|   +
++---------+-----------+
+
+
+`Object Lighting`_
+==================
+
+By default, object assets are loaded with Phong shading compatability. This is not ideal for assets with illumination baked into textures.
```
dhruvbatra, line 109: This right here is the beginning of a use-case. Could you expand this to say why someone might be introducing a new object asset (that it could be a robot mesh or a target object to navigate to or avoid)?

docs/pages/lighting-setups.rst
mosra: Since I have the whole stack under my control, nothing is impossible here (*evil laughter* :smiling_imp:).

I imagine Jupyter has some Python API for accessing notebook contents, so all that's needed to be done is hooking that up. I only have a "drive by" experience with Jupyter, can someone point me to APIs that could enable something like this?

```rst
.. jupyter:: bla.ipynb
    :cell: setup
```

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+    :end-before: # [/example 7]
+
+Updates to existing light setups will update all objects using that setup
+
+.. include:: ../../examples/tutorials/lighting_tutorial.py
+    :code: py
+    :start-after: # [example 8]
+    :end-before: # [/example 8]
+
+.. image:: images/lighting-setups-images/7.jpg
```
mosra, line 186: Why do all the images have an explicit white border? Padding can be done much more flexibly on the markup side. (I wanted to experiment with layouts a bit, but the white border made it all ugly :sweat_smile:)

The style also applies slight rounding to all image corners, but the white border makes that invisible.

docs/pages/lighting-setups.rst
mosra, line 109:
```suggestion
By default, object assets are loaded with Phong shading compatibility. This is not ideal for assets with illumination baked into textures.
```

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+:summary: This tutorial demonstrates the creation and manipulation of LightSetups in Habitat-sim.
+
+.. contents::
+    :class: m-block m-default
+
+A **LightSetup** consists of a set of LightInfo structures defining the common configuration
+of a set of point lights used to render objects in a scene. Once defined and registered,
+a LightSetup can be assigned to any subset of objects in the scene, including the scene asset itself.
+
+Each **LightInfo** structure in a LightSetup defines the `color`, `position`, and `LightPositionModel` of a single point light source.
```
mosra, line 13:
```suggestion
Each **LightInfo** structure in a LightSetup defines the ``color``, ``position``, and ``LightPositionModel`` of a single point light source.
```

reST needs two backticks for inline code. Here it might have accidentally done the correct thing, but only because it interpreted those as `:ref:`s (which is the default here), failed to find the target and rendered as monospace:

```
WARNING:root:Sphinx symbol `color` not found, rendering as monospace
WARNING:root:Sphinx symbol `position` not found, rendering as monospace
WARNING:root:Sphinx symbol `LightPositionModel` not found, rendering as monospace
```

examples/tutorials/lighting_tutorial.py
```diff
@@ -0,0 +1,226 @@
...
+    # [/example 3]
+
+    # [example 4]
+    id_2 = sim.add_object(chair_template_id)
+    sim.set_rotation(
+        mn.Quaternion.rotation(mn.Deg(90), mn.Vector3.x_axis())
+        * mn.Quaternion.rotation(mn.Deg(-115), mn.Vector3.z_axis()),
+        id_2,
+    )
+    sim.set_translation(np.array([3.06, 0.47, 1.15]), id_2)
```
mosra, line 160: Why not just this?

```suggestion
    sim.set_translation([3.06, 0.47, 1.15], id_2)
```

The API accepts a `mn.Vector3`, so while the above works because numpy arrays are implicitly convertible to magnum vector types, it's unnecessary. By the way, it would be good to have type annotations for the [Simulator](https://aihabitat.org/docs/habitat-sim/habitat_sim.simulator.Simulator.html) class, there's currently none (only the [SimulatorBackend](https://aihabitat.org/docs/habitat-sim/habitat_sim.sim.SimulatorBackend.html) has them because those are extracted from the native code, but that one is never used directly as far as I can see).

Same in all other cases (I see four calls to `set_translation()` with `np.array`).

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+
+
+`Object Lighting`_
+==================
+
+By default, object assets are loaded with Phong shading compatability. This is not ideal for assets with illumination baked into textures.
+Objects can be loaded for Flat shading by using the asset template's **PhysicsObjectAttributes.setRequiresLighting** function *before* loading the asset.
+Alternatively, this option can be set in the object template's configuration file:
+
+E.g. in *my_object.phys_properties.json*
```
mosra, line 113:
```suggestion
E.g. in ``my_object.phys_properties.json``
```

maybe?

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+By default, the scene will be shaded with no lights using the **habitat_sim.gfx.NO_LIGHT_KEY** LightSetup.
+This configuration is ideal for scene assets with illumination baked into textures (e.g. building scans such as MP3D).
+
+.. include:: ../../examples/tutorials/lighting_tutorial.py
+    :code: py
+    :start-after: # [default scene lighting]
+    :end-before: # [scene swap shader]
+
+.. image:: images/lighting-setups-images/0.jpg
+    :width: 45%
```
mosra, line 75:
```suggestion
    :width: 20em
```

This way the images have acceptable size also on small screen widths (a phone, e.g.). Adjust as you wish ;)

Same for all other images.

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,197 @@
...
+   :width: 100%
+
+.. |cp1| replace:: After loading with DEFAULT_LIGHTING_KEY.
+
+.. |cp2| replace:: After swapping to custom "my_scene_lighting".
+
++---------+-----------+
+|  |im1|  +   |im2|   +
+|  |cp1|  +   |cp2|   +
++---------+-----------+
```
mosra, line 103: *(Interesting, it never occured to me to use the || replacements for a table.)*

A table is however problematic for responsive layout -- on small screen sizes it makes the images really tiny and text wrapped badly. The following will place the images next to each other if the screen is wide enough, falling back to placing the images one after another (and limiting them to reasonable width again) if there's not enough space:

```rst
.. container:: m-row

    .. container:: m-col-m-5 m-push-m-1 m-col-s-6

        .. figure:: images/lighting-setups-images/1.jpg
            :figclass: m-flat
            :width: 20em

            ..

            After loading with :ref:`DEFAULT_LIGHTING_KEY <habitat_sim.gfx.DEFAULT_LIGHTING_KEY>`.

    .. container:: m-col-m-5 m-push-m-1 m-col-s-6

        .. figure:: images/lighting-setups-images/2.jpg
            :figclass: m-flat
            :width: 20em

            ..

            After swapping to custom ``my_scene_lighting``.
```

This is how it looks:

image

I get that this is probably not an area of expertise any of you would want to specialize in (mine wasn't either, until I *had to*, haha), but getting familiar with various `m-col-*` things from [Grid Layout](https://mcss.mosra.cz/css/grid/) will empower you to do various tricks to make the tutorials as well as the Habitat website look great and be readable on any screen size (which in my opinion matters a lot and is a big differentiating factor in documentation usability).

docs/pages/lighting-setups.rst
mosra:
```suggestion
`Working with Lights Example`_
```

Typographically I don't like the colon at the end (no other heading has it), but I feel the title doesn't really work either (maybe `Basic Setup` or something like that, at least?).

examples/tutorials/lighting_tutorial.py
```diff
@@ -0,0 +1,226 @@
...
+        sim.remove_object(id)
+
+
+def place_agent(sim):
+    # place our agent in the scene
+    agent_state = habitat_sim.AgentState()
+    agent_state.position = [5.0, 0.0, 1.0]
+    agent_state.rotation = quat_from_angle_axis(
+        math.radians(70), y_axis
+    ) * quat_from_angle_axis(math.radians(-20), x_axis)
```
mosra, line 52: This is most probably not possible to solve in this PR alone, but -- I remember we wanted to get rid of the `quaternion` package and those utils and replace them with something else (`mn.Quaternion` or 3x3 matrices?), but it kinda got forgotten since. Here it makes the example look kinda bad because there's two incompatible sets of APIs used for specifying rotations and one has to guess which one is used where.

docs/pages/lighting-setups.rst
aclegg3: Yes, but we are choosing whether to compile normals and how to load the material data at asset load time. I agree this needs to be addressed. I wrote this section to avoid future issues opening without a place to point to. 😅

mosra: Right, normals, forgot about those.... tho for models that are designed to be shadeless/unlit this could be forced to be always Flat, no? So then you could easily flip between Phong and Flat for models that *can* be shaded, while having Flat always for those which can't.

But yes, this is out of scope of this tutorial. Let's revisit that later.

examples/tutorials/lighting_tutorial.py
aclegg3: Yeah, not ideal. We should think about unifying the python api.

docs/pages/lighting-setups.rst
aclegg3: Thanks! I looked at a couple methods and didn't immediately find one that worked well. This works way better. 😁

docs/pages/lighting-setups.rst
aclegg3: This is a result of pyplot saving images with default padding. Should be better now.

docs/pages/lighting-setups.rst
aclegg3: I did another iteration on all the text. Should be more use-case driven now 👍  There will also be a complete tutorial specifically on objects (templates, instancing, manipulation, etc...). Once that is together we can cross link.

docs/pages/lighting-setups.rst
aclegg3: I like the idea of stand-alone, versioned python scripts which can be run in CI tests and directly linked to the tutorials. Keeps things up-to-date and does not require any external system to execute.

aclegg3:
Thanks @mosra and @dhruvbatra for your excellent comments and suggestions! New iteration is ready for review. 👍

docs/pages/lighting-setups.rst
```diff
@@ -0,0 +1,214 @@
...
+The :ref:`LightPositionModel` defines the coordinate frame of the light.
+
+Each light setup is registered in the simulator via a unique key. Two default lighting setups are pre-defined:
+
+
+- :ref:`DEFAULT_LIGHTING_KEY`
+
+  8 white, GLOBAL lights at scene bounding box corners.
+
+- :ref:`NO_LIGHT_KEY`
```
mosra, line 193: (I will do a more detailed pass over this tomorrow, but this caught my attention) I *think* it won't work like this (will produce a warning on the output and render as monospace, not a link). Instead you'll need to say the whole path, you can still make the title shorter tho):

```suggestion
- :ref:`NO_LIGHT_KEY <habitat_sim.gfx.NO_LIGHT_KEY>`
```

Same I guess for all others -- if a `:ref:` doesn't render a link, there's something off.

aclegg3: I added ref prefixes to address this. It works on my build 🤔 
```
:ref-prefix:
    habitat_sim.gfx
    habitat_sim.simulator
    habitat_sim.sim
    habitat_sim.agent
    habitat_sim.attributes
```

mosra: Ah, sorry about that, missed this part (it's too late here :sweat_smile:).

Tho to avoid confusing users I would use those in moderation, similarly to best practices with `using namespace`. One rule could be to use `:ref-prefix:` only if the tutorial code in question does the equivalent (`from habitat_sim.sim import Simulator` for example). I'm trying to follow this rule closely in Magnum docs -- always specifying the fully qualified path in all links (`Shaders::Phong` for example) unless it's undoubtedly clear from the context.

docs/pages/lighting-setups.rst
matthewjmay: Hi everyone :) thought I'd check in on my pending PRs to make sure things were going smoothly. Just wanted to clarify here that you **can** switch between flat/phong by requiring normals at load time and **correctly** specifying ambient color/texture. The issue is many of our models use a material model which isn't supported (PBR), so ambient color/texture isn't specified. Therefore, this issue is fixed without any changes to the lighting code, but instead

- supporting more material models (PBR should cover most cases)
- material manipulation, so users can swap diffuse <-> ambient textures if they really don't want ambient present during phong shading

aclegg3: Thanks for your input here, @matthewjmay. 

I agree with your proposed solutions. Material editing would give us some options for unblocking this use-case, and we are actively considering how to move forward on this. PBR support is also on our radar. 

For now, this documented work-around is functional, but not ideal.

docs/pages/lighting-setups.rst
aclegg3: I also don't have tons of experience here. Maybe this? https://nbsphinx.readthedocs.io/en/0.6.0/index.html

mosra: Oh, thanks! While I'm not running on top of Sphinx, I'm sure I could take this apart and extract valuable parts. Goes to my TODO list :)

aclegg3:
Any more comments, or can I merge this? @dhruvbatra @mosra

dhruvbatra:
LGTM 👍

docs/pages/lighting-setups.rst
mosra: This is still unresolved :P

examples/tutorials/lighting_tutorial.py
mosra: @aclegg3 And here, even though you said the math types need a general overhaul for better consistency, I nevertheless think removing the unneccessary `np.array` is a good thing to do.

aclegg3: Missed this somehow, good point. My comment really was targeted at the use of numpy vs magnum types across the stack. `Simulator` is mostly using magnum types, but `AgentState`, for example, is using numpy.

PR #531, opened 2020-03-17
Visualize NavMesh

aclegg3:
## Motivation and Context

Would be nice to see the NavMesh for debugging purposes with dynamic NavMesh re-computation. This change enables the NavMesh to be loaded into the `primitive_meshes_` list and added to the SceneGraph. From there it can be moved around to serve visualization uses.

## How Has This Been Tested
C++ unit test and viewer.cpp utility features.

Example - skokloster castle:

image
image

Example - decimated replica apartment_1:

navmesh_visualization

## Types of changes

## Checklist


src/esp/nav/PathFinder.cpp
```diff
@@ -1165,6 +1174,56 @@ PathFinder::Impl::getTopDownView(const float pixelsPerMeter,
...
+        // Get the polygon reference from the tile and polygon id
+        dtPolyRef polyRef = navMesh_->encodePolyId(iTile, tile->salt, jPoly);
+        const dtPoly* poly = nullptr;
+        const dtMeshTile* tmp = nullptr;
+        navMesh_->getTileAndPolyByRefUnsafe(polyRef, &tmp, &poly);
+
+        CORRADE_INTERNAL_ASSERT(poly != nullptr);
+        CORRADE_INTERNAL_ASSERT(tmp != nullptr);
+
+        // Code to iterate over triangles from here:
```
erikwijmans: Let's add a `getPolygonTriangles` helper method or something.  This code is kinda weird/complicated and we should avoid duplicating it.

src/esp/assets/ResourceManager.h
```diff
@@ -716,7 +735,7 @@ class ResourceManager {
    * @brief Primitive meshes available for instancing via @ref
    * addPrimitiveToDrawables for debugging or visualization purposes.
    */
-  std::vector<Magnum::GL::Mesh> primitive_meshes_;
+  std::vector<std::shared_ptr<Magnum::GL::Mesh>> primitive_meshes_;
```
erikwijmans:
```suggestion
  std::vector<std::unique_ptr<Magnum::GL::Mesh>> primitive_meshes_;
```
Perhaps?

src/esp/nav/PathFinder.cpp
aclegg3: Good idea. Makes the code look much cleaner and supports ease of future efforts.

PR #542, opened 2020-03-23
Upgrade to Magnum's new MeshData API

mosra:
meshdata

Depends on #540.

## What's this about?

In a classic import pipeline implemented in many 3D asset libraries, Assimp included, the process of importing a mesh and uploading it to a GPU looks roughly like shown the left side of the image above. There's nothing wrong with that for many file formats — especially textual ones such as OBJ — as various amounts of parsing and processing needs to be done anyway. 

Thanks to the new [`Trade::MeshData`](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1MeshData.html) API coming in this PR, *if* the file format is designed in a way that allows this, the process can be reduced to just metadata extraction and a direct copy of the data.
Compared to the classic case it also means that if the source files already has the data packed in some way (for example storing 24bit vertex colors instead of four floats), we can simply reuse that instead of first unpacking to a canonical type (and then possibly packing again to optimize GPU memory use). You may remember this plot from before — the majority of speed gains shown is exactly due to all extra processing being gone.

StanfordImporter

## Why this took so long?

The changes needed to allow describing arbitrary generic interleaved data on Magnum side amounted to more than 35kLOC of changes. Most of those is in the completely rewritten [`MeshTools`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html) library, which also has several new convenience utilities for operating directly on `Trade::MeshData` instances. The second major source of changes is in the importer plugins, to make them take advantage of all new possibilites. Among the changes included in this PR is:

-   Ability to describe, pack and unpack vertex attributes stored in (normalized/unsigned) 8- and 16-bit integer formats as well as half floats -- see the [`VertexFormat`](https://doc.magnum.graphics/magnum/namespaceMagnum.html#a174f1c4f12e98745bee3db77239aa53f) enum for a full list of builtin formats. It's of course possible to have custom implementation-specific formats as well, in case the builtin ones aren't enough.
-   Builtin support for tangent and bitangent attributes that are used for normal mapping and PBR workflows. Builtin primitives can now have tangents generated as well.
-   Normal map and texture transformation support in material import. *A PR enabling normal maps for glTF models will come later this week.*
-   Initial scaffolding for vertex weight and bone ID attributes, for mesh skinning and anmation.
-   Builtin support for per-vertex object IDs, which will finally allow replacing the custom PLY import code and obsoleting related use of `PrimitiveIDShader`. *PR with this is coming as well.*
-   Ability to describe custom per-vertex, per-edge, per-face and per-instance attributes (such as various semantic annotations), with transparent support in all `MeshTools` APIs and importers.
-   [`MeshTools::combineIndexedAttributes()`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a089320dd987fcc076550108019085609) and [`MeshTools::concatenate()`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#afff1c4db363a7c563ce8a48b03c4975d) for joining multiple meshes together
-   [`MeshTools::duplicate()`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#ad0789f7fc50abc956c1c69b7c6662bdc) or [`MeshTools::compressIndices()`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a82ad128b35a71350f5515a81afc995ef) operating directly on `Trade::MeshData`
-   A variant of [`MeshTools::removeDuplicates()`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a584e1fd1f4128ef2acec389337a3f248) that works on discrete / integer data (instead of doing fuzzy compare)
-   [`MeshTools::combineFaceAttributes()`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#abf0628e2a92c23d2d2f3aa03c01f4843) that'll turn per-face information into per-vertex and keep only unique combinations
-   Initial bits for GPU instancing, opening new optimization opportunities (for example scattering furniture across the scene, physics with many objects etc.)
-   Groundwork for mesh shaders, providing an ability to fully describe meshlet layout in a generic way. These got much more important lately, with AMD announcing support and D3D12 Ultimate having them builtin.

Additional things are in a work-in-progress Magnum branch and will be introduced later — among them is an efficient mesh (de)serialization format and interfaces for asset conversion plugins, bringing useful things like mesh decimation, vertex cache / overdraw optimization, index buffer compression etc.

## What file formats can take advantage of such optimized loading?

Crucially, the importer needs to allow the data to be imported that way. That's unfortunately not the case for Assimp, which decodes and copies all data into contiguous arrays that have to be interleaved back again.

-   **glTF** is designed to allow uploading its index / vertex buffers directly to the GPU, and is the go-to format suited for this case.
    -   With the new [`KHR_mesh_quantization`](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_mesh_quantization) extension that's being implemented in Magnum right now, large meshes can be packed into a much more compact representation that's better for both storage and rendering performance
    -   Additionally, a proposed [`KHR_mesh_instancing`](https://github.com/KhronosGroup/glTF/pull/1691) extension allows direct usage of GPU instancing on huge scenes, without the engine or application having to do extra guesswork or processing
-   **PLY**, to some extent. Magnum's own [`StanfordImporter`](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StanfordImporter.html) received a lot of improvements, copying all vertex data directly and index data only with minimal processing (as shown in the plot above). Additionally it gained support for per-face attributes (and their automatic conversion to per-vertex), normals and texture coordinates, which make it a full replacement to the 3rd party `tinyply` library
-   **STL**. This is a very simple format supporting only vertex positions and colors. Currently handled by Assimp, but since this format is used quite often nowadays, I'm writing a `StlImporter` plugin that'll allow us to get rid of Assimp for this format as well.

## What this PR does

-   All Magnum repositories were updated to latest master. Without any further code changes, it was verified that (apart from deprecation warnings) current code still compiles and all CI tests pass with the new version.
-   All usage of `Trade::MeshData3D` was replaced by `Trade::MeshData`, switching to a more appropriate containers (such as strided views instead of contiguous views) where needed, but I tried to minimize the amount of changes to a minimum to make this PR easy to review and verify, with further changes directed to future PRs. More on that below.
-   Additionally, Magnum changed the workflow from `mesh.draw(shader)` to (arguably more intuitive) `shader.draw(mesh)`. All code was updated to follow this.

## ... and what it doesn't do yet

-   Current Habitat code has a lot of places where mesh data are resident in separate `std::vector` instances, which are then interleaved and sent to the GPU — the major example of this is the `esp::assets::MeshData` structure. Ideally, this structure would get gradually replaced by Magnum's
`MeshData`. With this first PR I however wanted to ensure the massive dependency upgrade continues to work correctly before starting with larger codebase modifications.
-   Various utilities for mesh joining or duplicate removal could be replaced with the new functions in `MeshTools` (@aclegg3 you have a bunch of these in the physics code I think)
-   :warning: There's a TODO for PLY loading, and that's also the only place where deprecated functionality is still used. Upcoming PRs will clean it up, switch to `StanfordImporter` as well as builtin Magnum shaders for per-vertex object ID rendering. Again I didn't want to do this here to minimize the chance of introducing a breakage.
-   I want Habitat to work with the `KHR_mesh_quantization` glTF extension (and optimize all our datasets with it), but current Habitat code has a lot of assumptions about position / color / index data types that this extension will break. Not sure how to solve this yet (and don't really want to unconditionally unpack all imported data) — any ideas?

## How Has This Been Tested

My tests went well and CI is green, but again this is 35kLOC of new code, so it would be great if you can play with this branch before going further :)

## Types of changes

- [x] Docs change / refactoring / dependency upgrade (so far)

### DIFF
--- a/src/esp/gfx/test/DepthUnprojectionTest.cpp
+++ b/src/esp/gfx/test/DepthUnprojectionTest.cpp
@@ -202,8 +202,8 @@ void DepthUnprojectionTest::testGpuDirect() {
     Mn::GL::Mesh mesh = Mn::MeshTools::compile(Mn::Primitives::planeSolid());
     DepthShader shader;
     shader.setTransformationMatrix(transformation)
-        .setProjectionMatrix(data.projection);
-    mesh.draw(shader);
+        .setProjectionMatrix(data.projection)
+        .draw(mesh);
   }
 
   MAGNUM_VERIFY_NO_GL_ERROR();
@@ -246,8 +246,9 @@ void DepthUnprojectionTest::testGpuUnprojectExisting() {
       .clear(Mn::GL::FramebufferClear::Color);
 
   DepthShader shader{DepthShader::Flag::UnprojectExistingDepth};
-  shader.setProjectionMatrix(data.projection).bindDepthTexture(depth);
-  Mn::GL::Mesh{}.setCount(3).draw(shader);
+  shader.setProjectionMatrix(data.projection)
+      .bindDepthTexture(depth)
+      .draw(Mn::GL::Mesh{}.setCount(3));
 
   MAGNUM_VERIFY_NO_GL_ERROR();
 
@@ -325,7 +326,7 @@ void DepthUnprojectionTest::benchmarkGpuDirect() {
       .setProjectionMatrix(
           Mn::Matrix4::perspectiveProjection(60.0_degf, 1.0f, 0.001f, 100.0f));
 
-  CORRADE_BENCHMARK(10) { mesh.draw(shader); }
+  CORRADE_BENCHMARK(10) { shader.draw(mesh); }
 }
 
 void DepthUnprojectionTest::benchmarkGpuUnprojectExisting() {
@@ -365,7 +366,7 @@ void DepthUnprojectionTest::benchmarkGpuUnprojectExisting() {
   Mn::GL::Mesh mesh;
   mesh.setCount(3);
 
-  CORRADE_BENCHMARK(10) { mesh.draw(shader); }
+  CORRADE_BENCHMARK(10) { shader.draw(mesh); }
 }
 
 }  // namespace
--- a/src/esp/gfx/test/DepthUnprojectionTest.cpp
+++ b/src/esp/gfx/test/DepthUnprojectionTest.cpp
@@ -18,7 +18,7 @@
 #include <Magnum/PixelFormat.h>
 #include <Magnum/Primitives/Plane.h>
 #include <Magnum/Shaders/Flat.h>
-#include <Magnum/Trade/MeshData3D.h>
+#include <Magnum/Trade/MeshData.h>
 
 #include "esp/gfx/DepthUnprojection.h"
 
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -12,4 +12,10 @@ google-*,
 -google-runtime-references
 '
 
-WarningsAsErrors: '*'
+# Let deprecation warnings through. From time to time, Magnum annotates APIs
+# with deprecation warnings, suggesting people to upgrade to newer / better
+# designed / more flexible APIs. Treating such warnings as error is
+# counterproductive, since the first thing you want to do after an upgrade is
+# compiling existing *unchanged* code and ensuring all tests pass, and only
+# then start updating the code.
+WarningsAsErrors: '*,-clang-diagnostic-deprecated-declarations'
--- a/src/esp/assets/BaseMesh.h
+++ b/src/esp/assets/BaseMesh.h
@@ -16,7 +16,7 @@
 #include <Magnum/Math/Color.h>
 #include <Magnum/Math/Range.h>
 #include <Magnum/Mesh.h>
-#include <Magnum/Trade/MeshData3D.h>
+#include <Magnum/Trade/MeshData.h>
 #include "CollisionMeshData.h"
 #include "MeshData.h"
 #include "esp/core/esp.h"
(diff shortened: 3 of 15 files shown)


aclegg3:
Exciting changes. We'll have to migrate some of these mesh processing tools. 

Thanks for the comments on `CollisionMeshData`. We were planning to unify MeshData and CollisionMeshData at some point, but maybe it makes sense to keep them distinct and spread both over all internal asset types now. We can continue to think about this.

Everything is compiling and running as expected on my side.

src/esp/gfx/test/DepthUnprojectionTest.cpp
```diff
@@ -325,7 +326,7 @@ void DepthUnprojectionTest::benchmarkGpuDirect() {
       .setProjectionMatrix(
           Mn::Matrix4::perspectiveProjection(60.0_degf, 1.0f, 0.001f, 100.0f));
 
-  CORRADE_BENCHMARK(10) { mesh.draw(shader); }
+  CORRADE_BENCHMARK(10) { shader.draw(mesh); }
```
bigbike: This is great. `shader.draw(mesh)` is more intuitive than the old one.
I had this thought ever the beginning when I used Magnum library.

mosra: You're not alone -- and you should have told me that, I'm blind to past mistakes ;) 

And as far as I asked everyone else who thought the same didn't complain because they thought "there has to be a reason why it's like that" (there wasn't, it happened to be this way because of a sequence of unrelated events). Fortunately one user finally had the courage to point this out, and I happily changed for the better.

PR #544, opened 2020-03-26
Fix triangle sensor to work on gltf meshes

mpiseno:
## Motivation and Context

In one of my previous PRs (https://github.com/facebookresearch/habitat-sim/pull/525), I created a sensor that gets the unique triangle ID for each triangle in the mesh, however it did not work on .glb files, only .ply. I'm opening this PR for feedback because I can't get it to work properly yet.

I've verified that I can manually populate the TriangleIDBuffer in RenderTarget.cpp with some value and see that output in the sensor on the python side. However, the problem arises when trying to connect the TriangleShader to The logic in RenderTarget.cpp. I either get:

1. All 0s populating my triangleidBuffer
2. A bunch of unique numbers populating my triangleidbuffer, which shouldn't happen because right now I have triangle.frag setting its output to a constant value just to see if the shader is working properly.

Any graphics people please let me know if you see anything incorrect.

## Types of changes


src/esp/gfx/RenderTarget.cpp
```diff
@@ -128,14 +128,46 @@ struct RenderTarget::Impl {
...
+      //     .mapForDraw({{0, TriangleIdBuffer}});
+
+      // triangleFrameBuffer_.clearColor(1, Mn::Vector4i{27, 27, 27, 27});
+
+      CORRADE_INTERNAL_ASSERT(
+          framebuffer_.checkStatus(Mn::GL::FramebufferTarget::Draw) ==
+          Mn::GL::Framebuffer::Status::Complete);
+
+      triangleMesh_ = Mn::GL::Mesh{};
+      triangleMesh_.setCount(3);
```
erikwijmans: This is just going to segfault (or fail silently) as there isn't a vertex buffer (or index buffer) getting uploaded to the GPU.

This really shouldn't be done in the RenderTarget (the depth unprojection shader is a special beast -- it is a "compute shader" -- which is why it's in the render target).  I believe this can be done with the ShaderManager from https://github.com/facebookresearch/habitat-sim/pull/489 and just draw to the existing ObjectID attachment on the main frame buffer (cc @bigbike @matthewjmay)?

bigbike: @aclegg3 : What do you think?

mosra: > This is just going to segfault (or fail silently) as there isn't a vertex buffer (or index buffer) getting uploaded to the GPU.

Just FYI, attribute-less index-less meshes are fine :) That's how a [fullscreen triangle](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#aba8238805e0e572ea2edb8b2bca8c35c) is usually implemented, with an input-less shader that generates vertices from `gl_VertexID`.

erikwijmans: Ooo fancy -- I should stop doubting that magnum won't make OpenGL just do the right thing.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -76,22 +81,17 @@ struct RenderTarget::Impl {
...
-#endif
         .attachTexture(Mn::GL::Framebuffer::BufferAttachment::Depth,
                        depthRenderTexture_, 0)
-        .mapForDraw({{0, RgbaBuffer},
-                     {1, ObjectIdBuffer},
-#ifdef ESP_BUILD_WITH_TRIANGLE_SENSOR
-                     {2, TriangleIdBuffer}
-#endif
-        });
+        .mapForDraw({{0, RgbaBuffer}, {1, ObjectIdBuffer}});
```
mosra: I didn't read the full code yet, but this immediately stands out. (Sorry, Magnum API is a bit messy in this case.) The `mapForDraw()` affects global state, which basically means the second call wipes out the result of the first. The original code before this change was more correct I think.

src/shaders/triangle.vert
```diff
@@ -0,0 +1,6 @@
+layout(location = 0) in highp vec4 position;
+out int dummyvar;
+
+void main() {
+    dummyvar = 4;
```
mosra: The shader doesn't write to `gl_Position` so the GPU discards everything, and that could explain why you get nothing on the output. It needs to multiply the position with transformation and projection matrix, like the other shaders do.

mpiseno: Is it possible to just have the fragment shader? I only need gl_PrimitiveID, but I added a vertex shader because I didn't know if both were necessary

src/esp/gfx/RenderTarget.cpp
mpiseno: I updated the PR by modifying the vert and frag shaders, creating a TriangleIDDrawable class, and adding it to ResourceManager.cpp. What I don't understand is the connection between the new shader I wrote, the drawable, and TriangleIdBuffer inside Rendertarget.cpp.

eundersander:
Closing this stale PR.

PR #547, opened 2020-03-27
Getter/setters for derived Attribute defaults

aclegg3:
## Motivation and Context

This change refactors derived classes of `assets::Attributes` to expose getter/setter methods for expected default fields. This change is preparation for python bindings and tutorials to enable user friendly manipulation of object, scene, and simulator templates (e.g. `PhysicsObjectAttributes`) for domain randomization. 

## How Has This Been Tested

Existing code (including tests) modified to use getter/setters.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/Attributes.cpp
+++ b/src/esp/assets/Attributes.cpp
@@ -260,41 +260,218 @@ std::string Attributes::listAttributes() {
   return attributes;
 }
 
+//----------------------------------------//
+//  Derived attribute implementations
+//----------------------------------------//
+
 PhysicsObjectAttributes::PhysicsObjectAttributes() {
   // fill necessary attribute defaults
-  setDouble("mass", 1.0);
-  setDouble("margin", 0.01);
-  setMagnumVec3("scale", Magnum::Vector3(1.0, 1.0, 1.0));
-  setMagnumVec3("COM", Magnum::Vector3(0));
-  setMagnumVec3("inertia", Magnum::Vector3(0., 0., 0.));
-  setDouble("frictionCoefficient", 0.5);
-  setDouble("restitutionCoefficient", 0.1);
-  setDouble("linDamping", 0.2);
-  setDouble("angDamping", 0.2);
-  setString("originHandle", "");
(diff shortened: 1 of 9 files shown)


mosra:
Great work, everything is much cleaner now.

Though -- possibly an idea for a later refactor -- I wonder why there's this `Attributes` class and then [core::Configuration](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/core/Configuration.h), which is built on top of [Utility::Configuration](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Configuration.html) and *seems* to be capable of handling the same, without needing to have several `std::map`s for each type and thus being restricted to only the statically defined types. Or are there inherent limitations in its API that prevent it from being used here? I'm aware it lacks an ability to iterate over keys (which might be a showstopper, but I plan to add it for a feature I need soon). I also have to admit its API is very C++03y (it was built back then), but it's not like it would be completely unusable :)

aclegg3:
> Great work, everything is much cleaner now.
> 
> Though -- possibly an idea for a later refactor -- I wonder why there's this `Attributes` class and then [core::Configuration](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/core/Configuration.h), which is built on top of [Utility::Configuration](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Configuration.html) and _seems_ to be capable of handling the same, without needing to have several `std::map`s for each type and thus being restricted to only the statically defined types. Or are there inherent limitations in its API that prevent it from being used here? I'm aware it lacks an ability to iterate over keys (which might be a showstopper, but I plan to add it for a feature I need soon). I also have to admit its API is very C++03y (it was built back then), but it's not like it would be completely unusable :)

Thanks for the suggestion. I actually ran into this while looking into python bindings for Attributes. I'll dig into this a bit more. Thanks for letting me know the development plans. Key iteration is a desire-able feature, but I don't think we explicitly use it yet, so it may be ok to convert at this point and wait.

PR #548, opened 2020-03-28
Faster RGB & Depth noise models

mathfac:
## Motivation and Context
We noticed that RGB noise models are 7x times slower than whole Habitat step call.
@erikwijmans with @Skylion007  made our RGB noise models 50x faster with following PR.
- One minor change is to make GreedyFollower compatible with PyRobot actions. The change makes it possible to build shortest path on the fly with noise actuations.

## How Has This Been Tested
Existing [HSIM noise tests](https://github.com/facebookresearch/habitat-sim/blob/master/tests/test_sensors.py#L152-L207), [HAPI noise tests](https://github.com/facebookresearch/habitat-api/blob/master/test/test_sensors.py#L329).

## Types of changes

### DIFF
--- a/habitat_sim/nav/greedy_geodesic_follower.py
+++ b/habitat_sim/nav/greedy_geodesic_follower.py
@@ -65,7 +65,7 @@ class GreedyGeodesicFollower(object):
     def _find_action(self, name):
         candidates = list(
             filter(
-                lambda v: v[1].name == name,
+                lambda v: name in v[1].name,
                 self.agent.agent_config.action_space.items(),
             )
         )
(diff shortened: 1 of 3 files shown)


habitat_sim/nav/greedy_geodesic_follower.py
```diff
@@ -65,7 +65,7 @@ def __attrs_post_init__(self):
     def _find_action(self, name):
         candidates = list(
             filter(
-                lambda v: v[1].name == name,
+                lambda v: name in v[1].name,
```
Skylion007: Why is this change necessary?

erikwijmans: This is a little dangerous as things are assumed to be deterministic in the follower, but #546 will subsume this, so fine.

mathfac:
Removed greedy_geodesic_follower change.

PR #555, opened 2020-03-31
Cache pathfinding setup

erikwijmans:
## Motivation and Context

I did some additional benchmarking of the multigoal path finding, and the setup for projecting all the requested ends onto the navmesh is actually the bottleneck in a lot of cases.  This PR enables that setup to be cached so that subsequent called with the same path object/end points will not have to re-project them.

This PR also speeds up multigoal path finding by leveraging temporal consistency.  It is very likely that two concurrent calls with the same end points will have very similar starting points, thus we can heavily restrict the set of locations we need to search through by using their previously known distance and updating that to provide a minimum distance the point could possibly be.  We then explore candidate points based on this minimum distance.

## How Has This Been Tested

Existing tests pass, new benchmark shows this is faster!

## Types of changes

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -1060,6 +1060,8 @@ bool PathFinder::Impl::findPath(MultiGoalShortestPath& path) {
           path.pimpl_->minTheoreticalDist[i] - movedAmount,
           (path.pimpl_->requestedEnds[i] - path.requestedStart).norm());
     }
+
+    path.pimpl_->prevRequestedStart = path.requestedStart;
   }
 
   // Explore possible goal points by their minimum theoretical distance.
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -264,6 +264,10 @@ struct PathFinder::Impl {
                    const vec3f& end,
                    dtPolyRef endRef,
                    const vec3f& pathEnd);
+
+  bool findPathSetup(MultiGoalShortestPath& path,
+                     dtPolyRef& startRef,
+                     vec3f& pathStart);
 };
 
 namespace {
@@ -928,12 +932,12 @@ PathFinder::Impl::findPathInternal(const vec3f& start,
                                    const vec3f& pathEnd) {
   // check if trivial path (start is same as end) and early return
   if (pathStart.isApprox(pathEnd)) {
-    return std::make_tuple(0.0f, std::vector<vec3f>{});
+    return std::make_tuple(0.0f, std::vector<vec3f>{pathStart, pathEnd});
   }
 
   // Check if there is a path between the start and any of the ends
   if (!islandSystem_->hasConnection(startRef, endRef)) {
-    return Corrade::Containers::NullOpt;
+    return Cr::Containers::NullOpt;
   }
 
   static const int MAX_POLYS = 256;
@@ -944,7 +948,7 @@ PathFinder::Impl::findPathInternal(const vec3f& start,
       navQuery_->findPath(startRef, endRef, pathStart.data(), pathEnd.data(),
                           filter_.get(), polys, &numPolys, MAX_POLYS);
   if (status != DT_SUCCESS || numPolys == 0) {
-    return Corrade::Containers::NullOpt;
+    return Cr::Containers::NullOpt;
   }
 
   int numPoints = 0;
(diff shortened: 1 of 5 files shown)


erikwijmans:
Makes a fairly significant boost in training speed for ObjectNav

src/esp/nav/PathFinder.cpp
```diff
@@ -34,6 +35,41 @@ namespace Cr = Corrade;
...
+    : pimpl_{spimpl::make_unique_impl<Impl>()} {};
+
+void MultiGoalShortestPath::setRequestedEnds(
+    const std::vector<vec3f>& newEnds) {
+  pimpl_->setRequestedEnds(newEnds);
+}
+
+const std::vector<vec3f>& MultiGoalShortestPath::getRequestedEnds() const {
+  return pimpl_->getRequestedEnds();
+}
```
mosra, line 71: Probably more of a style thing, but since `Impl` is a struct with everything public, why not just this, and getting rid of all methods on the struct? You're accessing the members directly elsewhere anyway. (This is how I do PIMPLs myself.)

```suggestion
void MultiGoalShortestPath::setRequestedEnds(
    const std::vector<vec3f>& newEnds) {
    pimpl_->endRefs.clear();
    pimpl_->pathEnds.clear();
    pimpl_->requestedEnds = newEnds;

    pimpl_->minTheoreticalDist.assign(newEnds.size(), 0);
}

const std::vector<vec3f>& MultiGoalShortestPath::getRequestedEnds() const {
  return pimpl_->requestedEnds;
}
```

src/esp/nav/test/PathFinderTest.cpp
```diff
@@ -120,21 +152,47 @@ void PathFinderTest::pathFindBenchmark() {
...
+    for (int i = 0; i < 1000; ++i) {
+      rqEnds.emplace_back(pathFinder.getRandomNavigablePoint());
+    }
+    path.setRequestedEnds(rqEnds);
+
+    pathFinder.findPath(path);
+
+    bool status;
+    CORRADE_BENCHMARK(1) { status = pathFinder.findPath(path); };
     CORRADE_VERIFY(status);
```
mosra, line 196: Since this is three instances and every instance has a dedicated branch, I think it would be cleaner to just have three separate benchmarks :) Alternatively, you could make this data-driven ([canonical example in the docs](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-instanced)), putting the varying parameters in a struct and fetching them from the test:

```cpp
constexpr struct {
  const char* name;
  bool cache;
  int points;
} PathFindBenchmarkData[]{
  {"path to a single point", false, 1},
  {"path to closest of 1000", false, 1000},
  {"cached path to closest of 1000", true, 1000}
};

...

  addInstancedBenchmarks({&PathFinderTest::pathFindBenchmark},
    100, Cr::Containers::arraySize(PathFindBenchmarkData));
```

and then something like

```cpp
void PathFinderTest::pathFindBenchmark() {
  auto&& data = PathFindBenchmarkData[testCaseInstanceId()];
  setTestCaseDescription(data.name);

  ...

  std::vector<esp::vec3f> rqEnds;
  for (int i = 0; i < data.points; ++i) {
    rqEnds.emplace_back(pathFinder.getRandomNavigablePoint());
  }

  if(data.cached) pathFinder.findPath(path);

  ..
```

src/esp/nav/PathFinder.cpp
erikwijmans: Yeah, this makes more sense.  I am using PIMPL kinda strangely here as I wanted to hide the thing that needs a setter/getter, but not the other things.

src/esp/nav/test/PathFinderTest.cpp
erikwijmans: Did it a little differently as I still want to use the single-goal path object to make sure that code-path is still fast.

src/esp/nav/test/PathFinderTest.cpp
```diff
@@ -1,3 +1,4 @@
+#include <Corrade/Containers/Array.h>
```
mosra, line 1: `ArrayView` is enough if you use just `arraySize()`

src/esp/nav/PathFinder.cpp
```diff
@@ -34,6 +35,35 @@ namespace Cr = Corrade;
 namespace esp {
 namespace nav {
 
+struct MultiGoalShortestPath::Impl {
+  Impl() = default;
+  ~Impl() = default;
```
mosra, line 40: These aren't needed either I think ([rule of zero](https://en.cppreference.com/w/cpp/language/rule_of_three)).

PR #563, opened 2020-04-03
Simulator object template bindings

aclegg3:
## Motivation and Context

This change enables the python Simulator class to programmatically add/load new object templates and leverage template domain randomization utilities.

Related change: template reference by string key is being deprecated in favor of unique ids and the unused template getter by string key function has been removed from C++ Simulator.

## How Has This Been Tested

New python test `test_simulator.py::test_object_template_editing`.

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -283,8 +283,8 @@ class Simulator:
     def load_object_configs(self, path):
         return self._sim.load_object_configs(path)
 
-    def load_object_template(self, object_template_handle):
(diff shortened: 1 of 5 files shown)


aclegg3:
> However, following up on @jturner65 's comment above, what's the rationale for moving away from string keys/handles? I would think that string handles should result in more user-friendly asset management APIs.

Primarily aiming to prevent name collisions and provide a unified API for template interactions. If you feel string key accessors are important we can continue to provide dual access via ids and string keys.

PR #564, opened 2020-04-03
Port to StanfordImporter and Magnum's builtin per-vertex object ID shader

mosra:
## Motivation and Context

This:

- removes the `tinyply` submodule
- parses the PLY using [StanfordImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StanfordImporter.html), which does things like conversion of per-face to per-vertex attributes implicitly and thus we need much less of this hard-to-test logic on Habitat side
- removes the `PrimitiveIDShader` and `PrimitiveIDDrawable` because [Shaders::Phong can now do that itself](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1Phong.html#Shaders-Phong-usage-object-id)
- removes stale `texture.{vert,frag}` which isn't even compiled into resources anymore and thus clearly unused

:sparkles: The updated `magnum` submodule now has a [DebugTools::ColorMap](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1DebugTools_1_1ColorMap.html) namespace and [Shaders::MeshVisualizer*D](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1MeshVisualizer3D.html) that has an ability to visualize per-vertex object ID and primitive ID, among other things. Those are not getting used anywhere in this PR, but since the now-removed primitive ID shader contained a colormap (added in https://github.com/facebookresearch/habitat-sim/pull/521 but unused) and there's things like [habitat_sim.utils.common.d3_40_colors_hex](https://aihabitat.org/docs/habitat-sim/habitat_sim.utils.common.html#d3_40_colors_hex), I wonder if those new Magnum APIs could be useful here :)

image

:construction: To ensure I don't break too many things at once, the porting is done half-way -- the `InstancePlyData` full of de-interleaved `std::vector` instances is still kept. This is left as-is to keep the splitting by object ID from #521 intact. Future work could keep the data in the imported `MeshData` instance without any further copies / processing and only partitioning its index buffer so it can be rendered in parts, instead of each chunk being a completely separate mesh. This would likely need some updates on the resource manager side (buffers shared by multiple meshes), which is why it's not done in this PR.

:information_source: Also, based on a feedback by @erikwijmans, the `Mp3dInstanceData` (which does its own variant of PLY parsing) is left without changes, since it was an one-time-use tool and there are no regression tests for it in place, making accidental breakage very possible.

## How Has This Been Tested

Tested locally with `Mp3dTest`, `ReplicaSceneTest` and `test_sensors.py` against Replica and MP3D data. The CI doesn't have those, so it would be great if you could confirm that I'm indeed deleting the code :)

plot

:chart_with_upwards_trend: Since this moves from `tinyply` to Magnum's `StanfordImporter`, there may be differences in loading times. *Ideally* the loading time would drop down to a few hundred of milliseconds -- yes, it's the same plot again, but this time Habitat actually *gets to use* the advantages it presents -- however because of the not-yet-removed `InstancePlyData` copies and additional overhead from conversion of per-face to per-vertex data it'll likely be not so significant. I'd like to hear your feedback on that, if loading times visibly improve or not.

## Types of changes


mosra:
Any further comments on this? I feel unsure merging with just one approval, and the submodule update blocks another thing I want to submit.

PR #566, opened 2020-04-04
Add e2e test for frustum culling

erikwijmans:
## Motivation and Context

Finding a bug in #564 brought up that we don't have an e2e test for frustum culling ATM.  This PR adds frustum culling to the list of things we test in the existing e2e sensor test.

## How Has This Been Tested

Via  itself!

## Types of changes

### DIFF
--- a/tests/test_sensors.py
+++ b/tests/test_sensors.py
@@ -130,7 +130,6 @@ def test_sensors(
     # thus we need to force a reload when frustum culling gets swapped
     if cfg.sim_cfg.frustum_culling != sim.config.sim_cfg.frustum_culling:
         sim.close()
-        print("Swapping")
 
     sim.reconfigure(cfg)
 
--- a/tests/test_sensors.py
+++ b/tests/test_sensors.py
@@ -86,22 +86,31 @@ _test_scenes = [
 
 @pytest.mark.gfxtest
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


tests/test_sensors.py
```diff
@@ -111,11 +120,17 @@ def test_sensors(scene, has_sem, sensor_type, gpu2gpu, sim, make_cfg_settings):
     make_cfg_settings = {k: v for k, v in make_cfg_settings.items()}
     make_cfg_settings["semantic_sensor"] = has_sem
     make_cfg_settings["scene"] = scene
+    make_cfg_settings["frustum_culling"] = frustum_culling
 
     cfg = make_cfg(make_cfg_settings)
     for sensor_spec in cfg.agents[0].sensor_specifications:
         sensor_spec.gpu2gpu_transfer = gpu2gpu
 
+    # The scene loading may be done differently with/without frustum culling,
```
bigbike, line 129: Scene loading should be the same with/without FC, no special operations when it is on. :)
 If testing speed is not an issue, just keep it like this for precaution.

erikwijmans: The semantic mesh splitting is currently one done if frustum culling is enabled: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/sim/Simulator.cpp#L149

PR #569, opened 2020-04-06
Move attributes functionality to header; Modify Attributes object to be wrapped in smart pointer;

jturner65:
## Motivation and Context
This pull request has two major functions:  
1) all Attributes objects' functionality has been moved to header file, allowing getters/setters to be inlined and limit function overhead
2) all  PhysicsObjectAttributes objects are now wrapped in esp-managed smart pointers.  This is to facilitate more elegant management and access to POA objects.

This pull req also has some stub functionality added to viewer to facilitate future physical prims to be added to scene.

## How Has This Been Tested
run_tests and pytest both passed with these changes
PhysicsTest.cpp and SimTest.cpp have been modified to appropriately access PhysicsObjectAttributes objects and test their functionality. 

## Types of changes

## Checklist


src/esp/physics/PhysicsManager.cpp
```diff
@@ -481,24 +482,26 @@ void PhysicsManager::setObjectBBDraw(int physObjectID,
...
+        &existingObjects_.at(physObjectID)->visualNode_->createChild();
+    existingObjects_.at(physObjectID)->BBNode_->MagnumObject::setScaling(scale);
+    existingObjects_.at(physObjectID)
+        ->BBNode_->MagnumObject::setTranslation(
+            existingObjects_[physObjectID]
+                ->visualNode_->getCumulativeBB()
+                .center());
     resourceManager_->addPrimitiveToDrawables(
-        0, *existingObjects_[physObjectID]->BBNode_, drawables);
+        0, *existingObjects_.at(physObjectID)->BBNode_, drawables);
```
mosra, line 504: The move from `[]` to `at()` is unrelated to the Attributes-related changes, maybe do it in a separate PR to have this easier to review? 

Just a suggestion, I'm not the one to enforce any rules here :)

src/utils/viewer/viewer.cpp
```diff
@@ -41,18 +41,17 @@
...
 constexpr float moveSensitivity = 0.1f;
 constexpr float lookSensitivity = 11.25f;
 constexpr float rgbSensorHeight = 1.5f;
 
 namespace {
 
+// for ease of access
+namespace MnGL = ::Magnum::GL;
+namespace CorrUtil = ::Corrade::Utility;
+namespace MathLits = ::Magnum::Math::Literals;
```
mosra, line 53: Um... can we not, please? :) There's a `Mn` / `Cr` shorthand that is quite well-known at this point, but these will only cause confusion to unaware readers. So instead, like in other files

```cpp
namespace Cr = Corrade;
namespace Mn = Magnum;
```

and keeping those in the global namespace shouldn't hurt either.

`Mn::GL`, `Cr::Utility` are not much longer than `MnGL` or `CorrUtil`, and you can do `using Mn::Math::Literals` directly here -- it doesn't (and will not) bring any literals potentially conflicting with the standard ones.

jturner65: Thanks for suggestions, I was unaware of the prevalence of existing aliases.  Normally i would be uncomfortable putting this kind of thing (or much of anything, really) in the global namespace, but if this is acceptable, then I'll certainly do it that way :).

mosra: To provide more context -- on python side it's similar (`import magnum as mn`), and we refrain from putting those in headers as it confuses Doxygen (it can't then properly link the shorthand types to upstream docs). Aliases in global namespaces in cpp files are fine tho, having them always in the global namespace also makes it easier for the IDE autocompletion engine.

PR #573, opened 2020-04-07
--Better inheritance structure

jturner65:
RigidObject base class and Bullet-based child class objects now have better inheritance pathways and more obvious overloading.

## Motivation and Context

This pull request contains code to more clearly execute functionality with RigidObjects and BulletRigidObjects which inherit from them without code duplication or excess casting.  The base class functionality in rigidObject for initializing a scene object or an actual physical object is universal, while method hooks have been added for instancing/child class functionality to be added.  This will be easier to maintain (without code duplication) and easier to expand to support other physics engines.

## How Has This Been Tested
run_tests and pytest execute without issues.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletRigidObject.cpp
+++ b/src/esp/physics/bullet/BulletRigidObject.cpp
@@ -28,34 +28,109 @@
 namespace esp {
 namespace physics {
 
-BulletRigidObject::BulletRigidObject(scene::SceneNode* rigidBodyNode)
-    : RigidObject{rigidBodyNode}, MotionState(*rigidBodyNode){};
+BulletRigidObject::BulletRigidObject(
+    scene::SceneNode* rigidBodyNode,
+    std::shared_ptr<btMultiBodyDynamicsWorld> bWorld)
+    : RigidObject{rigidBodyNode},
+      MotionState(*rigidBodyNode),
+      bWorld_(bWorld) {}
 
-bool BulletRigidObject::initializeScene(
+bool BulletRigidObject::initializeSceneFinalize(
+    const assets::ResourceManager* resMgr,
     const assets::PhysicsSceneAttributes::ptr physicsSceneAttributes,
-    const assets::MeshMetaData& metaData,
-    const std::vector<assets::CollisionMeshData>& meshGroup,
-    std::shared_ptr<btMultiBodyDynamicsWorld> bWorld) {
-  if (rigidObjectType_ != RigidObjectType::NONE) {
-    LOG(ERROR) << "Cannot initialize a RigidObject more than once";
-    return false;
-  }
-
-  //! Turn on scene flag
-  rigidObjectType_ = RigidObjectType::SCENE;
-  objectMotionType_ = MotionType::STATIC;
+    const std::vector<assets::CollisionMeshData>& meshGroup) {
+  const assets::MeshMetaData& metaData =
+      resMgr->getMeshMetaData(physicsSceneAttributes->getCollisionMeshHandle());
 
   constructBulletSceneFromMeshes(Magnum::Matrix4{}, meshGroup, metaData.root);
   for (auto& object : bSceneCollisionObjects_) {
     object->setFriction(physicsSceneAttributes->getFrictionCoefficient());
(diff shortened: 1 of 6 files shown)


src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -28,34 +28,109 @@
...
-bool BulletRigidObject::initializeScene(
+BulletRigidObject::BulletRigidObject(
+    scene::SceneNode* rigidBodyNode,
+    std::shared_ptr<btMultiBodyDynamicsWorld> bWorld)
+    : RigidObject{rigidBodyNode},
+      MotionState(*rigidBodyNode),
+      bWorld_(bWorld) {}
+
+bool BulletRigidObject::initializeSceneFinalize(
+    const assets::ResourceManager* resMgr,
```
mosra, line 39:
```suggestion
    const assets::ResourceManager& resMgr,
```

instead (and same for other functions)? This function accesses the pointer without checking it's non-null and the documentation doesn't explicitly say that it's allowed to be null, so I assume it's not, and thus I think it would be clearer to enforce this by accepting a reference.

src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -442,6 +408,38 @@ class BulletRigidObject : public RigidObject,
...
+   * See @ref PhysicsManager::staticSceneObject_. Sets @ref rigidObjectType_ to
+   * @ref RigidObjectType::SCENE. See @ref btCollisionObject.
+   * @param physicsSceneAttributes The template structure defining relevant
+   * phyiscal parameters for the physical scene.
+   * @param meshGroup The collision mesh data for the scene.
+   * @param bWorld The @ref btMultiBodyDynamicsWorld to which the scene should
+   * belong.
+   * @return true if initialized successfully, false otherwise.
+   */
+  bool initializeSceneFinalize(
```
mosra, line 422: Assuming those are meant to be ever called from `initializeScene()`, this (and the "parent" implementation as well) could be `private` I think.

I'm doing a similar thing for plugin interfaces in magnum, the actual implementations being private to prevent them from being called by accident from outside (see e.g. [Trade::AbstractImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1AbstractImporter.html#pri-methods)).

src/esp/physics/bullet/BulletRigidObject.cpp
jturner65: the reason i am using a pointer is that the calling manager has it as a pointer already.  just passing what was there already.

src/esp/physics/bullet/BulletRigidObject.h
jturner65: Thanks for suggestion here - I was thinking access was restricted from base class unless they were protected (Which is what i made them).

src/esp/physics/bullet/BulletRigidObject.cpp
jturner65: Perhaps they should be friended and then made private?

mosra: In my case I'm always trying to move the responsibility of null checking to the bottom of the stack as much as possible, and since here it clearly always needs a non-null pointer, I would make these (and `initializeScene()`) accept references. Is there a piece of code clearly responsible for creating the pointer? Or can it be references all the way down?

aclegg3: We could refactor to `ResourceManager` reference in `PhysicsManager` and pass that along the full path. I don't see any problem with that. 👍

aclegg3: Let's call this a follow-up refactor PR.

src/esp/physics/bullet/BulletRigidObject.h
mosra: Just FYI I still see them `protected`, not `private`. Did you actually make the change? :)

jturner65: oops.  will address this in next pr, along with ResourceManager ptr->ref conversion

PR #577, opened 2020-04-09
Prevent Assimp from appending `d` to all our binaries

mosra:
## Motivation and Context

Reported by @dexter1691 -- it breaks things when doing `./build.sh --debug`, in particular it causes
the Python bindings to have a different name, which are then not found properly by Python.

## How Has This Been Tested

There should be no `d` suffix anymore in bindings produced in the Debug build.

:warning: Unfortunately the `INJECT_DEBUG_POSTFIX` option that's meant to fix this behavior wasn't working properly in the submodule commit that was in the repository, so I had to upgrade the `assimp` submodule to the `v5.0.1` tag. While I know about several things that got fixed for this version, there are also known regressions -- so please if you could check with random OBJ, STL or COLLADA files and check for suspicious behavior, it'd be great.

## Types of changes

### DIFF
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -74,6 +74,11 @@ if(BUILD_ASSIMP_SUPPORT)
     set(ASSIMP_BUILD_ASSIMP_TOOLS OFF CACHE BOOL "ASSIMP_BUILD_ASSIMP_TOOLS" FORCE)
     set(ASSIMP_BUILD_TESTS OFF CACHE BOOL "ASSIMP_BUILD_TESTS" FORCE)
     set(BUILD_SHARED_LIBS OFF CACHE BOOL "ASSIMP_BUILD_TESTS" FORCE)
+    # The following is important to avoid Assimp appending `d` to all our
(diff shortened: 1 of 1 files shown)


erikwijmans:
> check with random OBJ, STL or COLLADA files

I don't think we have any of these :)

mosra:
We don't? Not even OBJs? In that case we might as well start phasing out Assimp completely, hehe :)

Tho I think COLLADA will get used for URDF models at least (@aclegg3?)

aclegg3:
> We don't? Not even OBJs? In that case we might as well start phasing out Assimp completely, hehe :)
> 
> Tho I think COLLADA will get used for URDF models at least (@aclegg3?)

We do support obj for objects, so we should check these.

 We should support stl and collada in the future. I'm not too worried about regression on these at this time.

mosra:
STL can be now handled by Magnum itself (see the [StlImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StlImporter.html) plugin, arrived in `master` yesterday). Magnum used to have its own COLLADA importer but the file format is so horrible that it was a maintenance nightmare I happily removed since.

PR #578, opened 2020-04-09
Update GLFW, various GLFW-related fixes

mosra:
## Motivation and Context

Partially reported by @dexter1691, partially from my "known issues" list that I'm finally fixing.

 * `GlfwApplication` was aborting on a pure virtual method call on macOS when instantiated from Python; not reproducible when running natively or when using Magnum Python bindings from Homebrew
 * Better error reporting in case `draw_event()` is not provided in an `Application` subclass in Python (again, it would fail on a pure virtual method call, now it raises a descriptive exception)
 * GLFW applications were reported to run only with a 16-bit color depth on macOS, that's fixed now as well.

:sparkle: Also updating GLFW to 3.3.2. It contains a bunch of very useful fixes for all platforms (and no known regressions), no reason to not upgrade.

## How Has This Been Tested

I expect the CI to pass. In case it fails, there *might* be some missing X11 dependencies that need to get added, but I know which ones.

## Types of changes


mosra:
Since this is a blocking issue for @dexter1691, can we get some reviews / approvals here? :) Thanks!

Issue #581, opened 2020-04-09
Viewer crashing when trying to visualize the navmesh 


dexter1691:
## 🐛 Bug
The latest master is crashing when trying to visualize the navmesh while running the Viewer. 

## Command
`./build/viewer --enable-physics data/scene_datasets/habitat-test-scenes/skokloster-castle.glb`

## To Reproduce
- Run the above command, and then press N to visualize the navmesh

## System Info
Platform: Darwin-19.3.0-x86_64-i386-64bit
Machine: x86_64
Processor: i386
Libc version:
Mac version: 10.15.3
Python version: 3.6.10
Architecture: 64bit
Win version:
System OS: Darwin
Release: 19.3.0
Version: Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64
Operational System: darwin
GCC version: Apple clang version 11.0.0 (clang-1100.0.33.17)
CMAKE version: cmake version 3.14.0
NVIDIA-SMI:
Pip packages versions:

Conda packages versions:

jturner65:
Please see if this has been addressed by patch.  Thanks!

dexter1691:
Yes. This fixes it! Thanks

PR #582, opened 2020-04-09
--Address Issue #581

jturner65:
navmeshVisNode_ needed to be verified not null before deletion

## Motivation and Context  

Addressing issue #581 
Bug Fix

## How Has This Been Tested
run_test and pytest passed

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -128,7 +128,7 @@ class Viewer : public Mn::Platform::Application {
   esp::scene::SceneGraph* sceneGraph_;
   esp::scene::SceneNode* rootNode_;
 
-  esp::scene::SceneNode* navmeshVisNode_;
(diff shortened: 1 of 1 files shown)


src/utils/viewer/viewer.cpp
```diff
@@ -445,7 +445,7 @@ void Viewer::toggleNavMeshVisualization() {
     int nevMeshVisPrimID = resourceManager_.loadNavMeshVisualization(
         *pathfinder_, navmeshVisNode_, &sceneGraph_->getDrawables());
     navmeshVisNode_->translate({0, 0.1, 0});
-  } else {
+  } else if (navmeshVisNode_ != nullptr) {
```
mosra, line 448: Not that it matters much in this case (no followup PR needed), but `delete` (and `free()` as well) is a no-op when called on a null pointer, so this additional check isn't strictly needed. 

What makes the fix work is initializing the variable to `nullptr` above.

PR #583, opened 2020-04-10
Create README.md

dhruvbatra:
Instructions for building documentation

## Motivation and Context

Because friends don't let friends hunt in CI config for how to build docs. 

## How Has This Been Tested

🤷‍♂️ 

## Types of changes

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,13 @@
+## Building Docs Locally
+
+```bash
+python setup.py install (#or ./build.sh) 
+cd docs
+mkdir -p ../build/docs
+git submodule update --init
+./build-public.sh
+```
+
+## Viewing Docs Locally
+../build/docs-public/habitat-cpp/index.html
+../build/docs-public/habitat-sim/index.html


docs/README.md
```diff
@@ -0,0 +1,13 @@
+## Building Docs Locally
+
+```bash
+python setup.py install (#or ./build.sh) 
+cd docs
+mkdir -p ../build/docs
```
mosra, line 6: The shell script, if it completes, [shows a clickable link](https://github.com/facebookresearch/habitat-sim/blob/28a24acfd10c1c262c88a795a5c5b62bae2195c8/docs/build.sh#L37-L41) with output location (which I thought could be intuitive enough), but having to do this explicit step makes it work less than ideal. I'll follow up with a PR that removes the need to explicitly create an output directory.

dhruvbatra: @mosra -- is this ready to merge?

mosra: You're right, this fell through cracks and I completely forgot about it. The README isn't wrong on its own, so let's merge this.

> I'll follow up with a PR that removes the need to explicitly create an output directory.

This still holds, of course.

PR #585, opened 2020-04-10
[WIP] Laying groundwork for generating Physics primitives

jturner65:
## Motivation and Context
We want to be able to instantiate physically modeled primitives on demand that do not rely upon a template file for their instantiation.  This pull request contains some groundwork code to facilitate this.

Some minor refactoring is also included to bring PhysicsManagerAttributes access into line with recent changes to Attributes accessors.

## How Has This Been Tested
run_test and pytest conducted successfully.  New tests will need to be constructed to support this functionality 

## Types of changes

## Checklist


src/utils/viewer/viewer.cpp
```diff
@@ -321,21 +322,25 @@ void Viewer::addObject(std::string configFile) {
...
-  int physObjectID = physicsManager_->addObject(configFile, &drawables);
+  int physObjectID = physicsManager_->addObject(ID, &drawables);
   physicsManager_->setTranslation(physObjectID, new_pos);
 
   physicsManager_->setRotation(physObjectID, esp::core::randomRotation());
 
   objectIDs_.push_back(physObjectID);
 }  // addObject
 
+void Viewer::addObject(std::string configFile) {
```
mosra, line 333:
```suggestion
void Viewer::addObject(const std::string& configFile) {
```

since you're changing this function anyway :)

jturner65: Thanks!

src/esp/assets/Attributes.h
```diff
@@ -182,6 +182,25 @@ class PhysicsManagerAttributes : public esp::core::Configuration {
...
+    return getDouble("frictionCoefficient");
+  }
+
+  void setRestitutionCoefficient(double restitutionCoefficient) {
+    setDouble("restitutionCoefficient", restitutionCoefficient);
+  }
+  double getRestitutionCoefficient() const {
+    return getDouble("restitutionCoefficient");
+  }
+
```
aclegg3, line 203: This is information is already present in the `PhysicsSceneAttributes`. This distinction is not functional currently, but philosophically a `PhysicsManager` steps the world, while a scene has gravity, friction, and restitution.

Ideally, a single PhysicsManager could hold multiple scenes with different worlds (TODO).

src/esp/assets/ResourceManager.cpp
```diff
@@ -170,6 +170,57 @@ bool ResourceManager::loadScene(
...
 
+void ResourceManager::initPhysicsManager(
+    std::shared_ptr<physics::PhysicsManager>& physicsManager,
+    PhysicsManagerAttributes::ptr physicsManagerAttributes) {
+  //! PHYSICS INIT: Use the above config to initialize physics engine
+  bool defaultToNoneSimulator = true;
+  if (physicsManagerAttributes->getSimulator().compare("bullet") == 0) {
+#ifdef ESP_BUILD_WITH_BULLET
+    physicsManager.reset(
+        new physics::BulletPhysicsManager(*this, physicsManagerAttributes));
```
aclegg3, line 181: Maybe pass in the PhysicsSceneAttributes also. Maybe add keys for the scene attributes to physicsManagerAttributes? We can refactor later for multi-scene support.

src/esp/assets/ResourceManager.cpp
```diff
@@ -485,15 +503,29 @@ void ResourceManager::addObjectToDrawables(int objTemplateLibID,
...
-PhysicsObjectAttributes::ptr ResourceManager::getPhysicsObjectAttributes(
-    const std::string& objectName) {
-  return physicsObjTemplateLibrary_.at(objectName);
-}
 PhysicsObjectAttributes::ptr ResourceManager::getPhysicsObjectAttributes(
     const int objectTemplateID) {
   return physicsObjTemplateLibrary_.at(getObjectConfig(objectTemplateID));
 }
 
+bool ResourceManager::FileMeshLoad(AssetInfo& meshInfo,
```
aclegg3, line 511: Maybe `loadObjectMeshDataFromFile`?

src/esp/assets/Attributes.h
jturner65: Should i remove these then?  I only added them because the equivalent fields were being set in PhysicsManagerAttributes via string keys, but i much prefer the scenario where a scene completely manages these world-related quantities and the PhysicsManager only steps the world.

src/esp/assets/ResourceManager.cpp
jturner65: How about we pass a reference to the PhysicsSceneAttributes map?  We can then specify which scene via passing the key to that scene.  This will support future multi-scenes more easily with minimal modifications.

src/esp/assets/Attributes.h
jturner65: On the other hand, having these set in PhysicsManagerAttributes makes them able to be easily broadcast to the scenes, since the physics manager attributes are read first.

PR #592, opened 2020-04-15
Fix glb object miss-classification as MP3D and corresponding rotation.

aclegg3:
## Motivation and Context

Object asset loading was using the `assets::AssetInfo::fromPath()` used to load and classify scenes. This mis-classified all glTF assets as `AssetType::MP3D` and assigned them a 90 degree rotation. This causes confusion and difficulty converting assets from external scene design sources like Blender or Unity 3D.

This is a breaking change in that all glTF assets previously loaded in user side scripts, tests, and tutorials will be off by a 90 degree rotation after this change. Better now that later. 😅 

## How Has This Been Tested

Most existing tests pass, some will need to be adjusted. Most test assets are symmetric, so should not be affected.

TODO: fix broken tests.

## Types of changes

## Checklist


erikwijmans:
Looks like we need to re-generate the golden data for some of tests?

mosra:
I checked this out locally and updated the expected images. Found this, tho (magnified):

image

in particular, the pillow is rendered on top of the cube. Is that expected? Just asking, my depth sensing fails on that image so maybe it's actually correct :)

@aclegg3 for future reference -- if you run the test locally and some of the image comparisons fail, it'll print something like this:

image

Then you can use for example `--save-diagnostic .` to save the new output to current working directory, and then replace the original ones with them. [Docs.](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-save-diagnostic) This is how Matthew generated the original images I think, so it should work for you as well; if it's complaining about a `PngImageConverter` not found, you might need to enable, find & link the test to [MagnumPlugins::StbImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StbImageConverter.html#Trade-StbImageConverter-usage) -- in my case it just picked up a system-installed image converter plugin, which is how it's designed to work by default.

aclegg3:
> in particular, the pillow is rendered on top of the cube. Is that expected? Just asking, my depth sensing fails on that image so maybe it's actually correct :) I also fail to understand the connection between the 90° rotation and the change in the test images -- to me it seems that the cubes moved rather than rotated.

Actually this shape is not a cube, but a rectangular prism with 3 different dimensions. By rotating, I think a shorter side is now facing up, so it appears as though the object moved.

However, these tests are still failing locally for me after pulling this branch.

mosra:
That one failed for me too (are you running on an Intel GPU?). I assumed this was local to my machine/driver and didn't update anything.

To make it pass, update the mean threshold to `0.71` from `0.65` at the line where it complains. This seems to be just some rounding error, nothing critical.

src/tests/SimTest.cpp
```diff
@@ -282,17 +282,17 @@ void SimTest::updateObjectLightSetupRGBAObservation() {
...
   simulator->setTranslation({1.0f, 0.5f, -0.5f}, objectID);
   checkPinholeCameraRGBAObservation(
-      *simulator, "SimTestExpectedDefaultLighting.png", maxThreshold, 0.65f);
+      *simulator, "SimTestExpectedDefaultLighting.png", maxThreshold, 0.71f);
 
   // change from default lighting to custom
   simulator->setObjectLightSetup(objectID, "custom_lighting_1");
   checkPinholeCameraRGBAObservation(
-      *simulator, "SimTestExpectedCustomLighting.png", maxThreshold, 0.65f);
+      *simulator, "SimTestExpectedCustomLighting.png", maxThreshold, 0.71f);
```
mosra, line 290: Um, these don't necessarily need to be same everywhere ;) For some scenes the difference might be higher, for some not, and making the thresholds unnecessarily large might hide errors.

aclegg3: My tests were failing on all of these scenes with the same error ~0.702.

mosra: oh, interesting -- in my case it was just one of these.

Fair enough, then :)

PR #593, opened 2020-04-17
Don't double apply rotation

erikwijmans:
## Motivation and Context

Little bug in OBB loading for mp3d.  The radius of the obb is getting double corrected by the worldRotation.

## How Has This Been Tested

User has a visualization script, used that.

## Types of changes

### DIFF
--- a/src/esp/scene/Mp3dSemanticScene.cpp
+++ b/src/esp/scene/Mp3dSemanticScene.cpp
@@ -94,12 +94,13 @@ bool SemanticScene::loadMp3dHouse(
 
   const bool hasWorldRotation = !rotation.isApprox(quatf::Identity());
 
-  auto getVec3f = [&](const std::vector<std::string>& tokens, int offset) {
+  auto getVec3f = [&](const std::vector<std::string>& tokens, int offset,
+                      bool applyRotation = true) {
     const float x = std::stof(tokens[offset]);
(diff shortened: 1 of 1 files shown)


erikwijmans:
> Maybe better to re-factor with a boolean (default true) to apply world rotation instead of undoing?

Yeah, that does seem cleaner.  Change made.

src/esp/scene/Mp3dSemanticScene.cpp
```diff
@@ -94,29 +94,36 @@ bool SemanticScene::loadMp3dHouse(
...
+    if (applyRotation && hasWorldRotation) {
       p = rotation * p;
     }
     return p;
   };
 
   auto getBBox = [&](const std::vector<std::string>& tokens, int offset) {
-    return box3f(getVec3f(tokens, offset), getVec3f(tokens, offset + 3));
+    return box3f(getVec3f(tokens, offset),
+                 getVec3f(tokens, offset + 3).array().abs().matrix());
```
aclegg3, line 111: This seems unrelated to the rotation. Different bug?

erikwijmans: Oh yeah, just saw that we still hadn't fixed this -- after rotation, the sizes may be negative, which is weird -- so just did it.

PR #594, opened 2020-04-20
Only load semantic scene if needed

erikwijmans:
## Motivation and Context

Loading the semantic mesh takes time and we don't always need it.  This PR makes the loading of the semantic mesh conditioned on having a sensor of type SEMANTIC.

Also includes some relatively minor house keeping of no longer using the bindings directly and instead using the broken out modules.

## How Has This Been Tested

Updated the test to not load the semantic scene on the color and depth sensor renders.  All passes and it's faster!

## Types of changes


tfy14esa:
Hi Erik!
Does this mean that I can run the habitat simulator on .ply files that don't have semantic information or is the semantic information still needed? I understand that I still need a navmesh to the .ply file.

The reason for my question is that I have made some replica .ply files watertight (using poisson reconstruction), but the semantic information got lost along the way and I can no longer run the habitat simulator on these watertight .ply files. Perhaps your update solves my problem without having to add semantic information as a post-processing step?

Cheers,

erikwijmans:
No, this update does not solve that.  This update made it so the semantic mesh for mp3d won't get loaded at all if they aren't needed (mp3d has two meshes, one for rgb and one for semantics).

In terms of doing what you want to do, putting the semantic information back would likely be the simplest way to get it to work.  A add a code-path for colored plys without semantics could be added, but I am not sure how much work that would be.

tfy14esa:
Thanks for your quick response. I added fake object_ids to each face by writing a script. That worked and now I can run habitat with my watertight meshes.

If anyone is interested in the script - just comment below.

PR #596, opened 2020-04-22
Better default material

aclegg3:
## Motivation and Context

The `DEFAULT_MATERIAL` was initialized with full white ambient and diffuse light, causing objects to appear flat shaded. These values have been modified for a subjectively better appearance.

## How Has This Been Tested

Original:
image

New:
image

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1750,8 +1750,11 @@ void ResourceManager::initDefaultLightSetups() {
 }
 
 void ResourceManager::initDefaultMaterials() {
-  shaderManager_.set<gfx::MaterialData>(DEFAULT_MATERIAL_KEY,
-                                        new gfx::PhongMaterialData{});
+  auto defaultMaterial = new gfx::PhongMaterialData{};
+  defaultMaterial->ambientColor = Magnum::Color4{0.1};
+  defaultMaterial->diffuseColor = Magnum::Color4{0.7};
+  shaderManager_.set<gfx::MaterialData>(DEFAULT_MATERIAL_KEY, defaultMaterial);
+
   auto perVertexObjectId = new gfx::PhongMaterialData{};
   perVertexObjectId->perVertexObjectId = true;
   shaderManager_.set<gfx::MaterialData>(PER_VERTEX_OBJECT_ID_MATERIAL_KEY,
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1752,7 +1752,6 @@ void ResourceManager::initDefaultLightSetups() {
 void ResourceManager::initDefaultMaterials() {
   shaderManager_.set<gfx::MaterialData>(DEFAULT_MATERIAL_KEY,
                                         new gfx::PhongMaterialData{});
-
   auto perVertexObjectId = new gfx::PhongMaterialData{};
   perVertexObjectId->perVertexObjectId = true;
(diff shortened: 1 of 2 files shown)


aclegg3:
> Sort of tangent to the PR, but in line with what was stated above : Since a PhongMaterialData has pointers to textures, and only a default destructor, is this not a possible source of memory leaking?

These textures live in the `textures_` map as shared pointers. The pointers stored in the material structure are pointing here (assigned in `getPhongShadedMaterialData`), so no issue with memory leaks. We also don't allow textures to be freed during runtime, so none of these pointers should get stale.

aclegg3:
> A comment unrelated to the content of the changes: the use of naked `new` makes loud alarm bells ring in my head whenever I see it outside the innards of class construction/destruction logic that is neatly paired up. 😅I assume this is safe because the `ShaderManager` takes responsibility to free the memory eventually?

That is a valid concern. In this case we don't remove any materials once added and as a member of `esp::assets::ResourceManager` the `shaderManager_` outlives a `Simulator` so there is no potential for missed deletion within the current assumptions. 

@mosra I see the reference counting and freeing functionality in `Magnum::ResourceManager`. For these `gfx::MaterialData` pointers, `free` handles deletion, right?

src/esp/assets/ResourceManager.cpp
```diff
@@ -1750,8 +1750,11 @@ void ResourceManager::initDefaultLightSetups() {
 }
 
 void ResourceManager::initDefaultMaterials() {
-  shaderManager_.set<gfx::MaterialData>(DEFAULT_MATERIAL_KEY,
-                                        new gfx::PhongMaterialData{});
+  auto defaultMaterial = new gfx::PhongMaterialData{};
+  defaultMaterial->ambientColor = Magnum::Color4{0.1};
+  defaultMaterial->diffuseColor = Magnum::Color4{0.7};
```
mosra, line 1755: Wouldn't it make more sense to assign reasonable defaults in the PhongMaterialData itself instead? Having a white ambient in there by default doesn't make sense either.

https://github.com/facebookresearch/habitat-sim/blob/d34c3452c52e88336f67b4811856adea5d5ac3b6/src/esp/gfx/MaterialData.h#L19-L30

That way one can rely on them to be set to correct values and only modify what's needed.

aclegg3: Good point. Done. 👍

mosra:
Sorry, missed the comments above.

> I assume this is safe because the ShaderManager takes responsibility to free the memory eventually?

> For these gfx::MaterialData pointers, free handles deletion, right?

Yes to both. This API originates in early C++11 times and it was only updated to take r-value references and smart pointers quite recently (it was waiting for a lighter-weight unique pointer implementation), and most of existing code, docs and examples still uses the discouraged "naked new" way. The PR doesn't touch the relevant part anymore, but what could be done to hide the raw `new` is for example

```cpp
shaderManager_.set<gfx::MaterialData>(DEFAULT_MATERIAL_KEY,
                                      gfx::PhongMaterialData{});
```

```cpp
shaderManager_.set<gfx::MaterialData>(DEFAULT_MATERIAL_KEY,
                                      Cr::Containers::pointer<gfx::PhongMaterialData>());
```

or, using the standard unique pointers

```cpp
#include <Corrade/Containers/PointerStl.h>

shaderManager_.set<gfx::MaterialData>(DEFAULT_MATERIAL_KEY,
                                      std::make_unique<gfx::PhongMaterialData>());
```

PR #601, opened 2020-04-24
Refactor VelocityControl::IntegrateTransform

aclegg3:
## Motivation and Context

Numerical error was triggering assertions. ~This patch orthogonalizes the transforms if necessary.~ Also adds explicit docs about treatment of scaling for `RigidObject` transformations.

Edit: Introduced a RigidState struct to hold translation/rotation pairs. VelocityControl::IntegrateTransform now consumes/produces these objects and operates on quaternions.

## How Has This Been Tested

New test added.

## Types of changes

## Checklist


src/esp/physics/RigidObject.cpp
```diff
@@ -236,32 +237,36 @@ Magnum::Matrix3 RigidObject::getInertiaMatrix() {
...
 // VelocityControl
+
 Magnum::Matrix4 VelocityControl::integrateTransform(
     const float dt,
-    const Magnum::Matrix4& objectTransform) {
+    const Magnum::Matrix4& objectRotationTranslation) {
+  Magnum::Matrix3 newRotation{objectRotationTranslation.rotationScaling()};
+  if (!newRotation.isOrthogonal()) {
+    Magnum::Math::Algorithms::gramSchmidtOrthonormalizeInPlace(newRotation);
+  }
```
jturner65, line 247: Is it necessary to check orthogonality every integration step?  Seems potentially expensive, although non-ortho transforms are definitely bad.  Could we not do the check every 'x' integration steps?

aclegg3: We could provide a boolean to make the check optional, but I don't think that we will really know when not to check in practice.

erikwijmans: I'd not even do `newRotation.isOrthogonal()`.  `Magnum::Math::Algorithms::gramSchmidtOrthonormalizeInPlace` is really fast.  We use it unconditionally on every single render call: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/sensor/PinholeCamera.cpp#L46  When I added this, there was no measure-able hit to perf.

aclegg3: I ran 60 seconds of simulation and tracked the frequency of non-orthonormal rotation transforms. Looks like 98% of the time (in my experimental setup), this transform is already orthonormal. May be worth checking since we could end up integrating many times for each object between each render call.
image

@mosra thoughts? Is the check fast enough to be worth it?

erikwijmans: Asymptotically, the check is O(n^3) and so is GramSchmidt.

jturner65: The problem with numerical drift is causing assertions in the quaternion construction, right?  Why not manually construct the quaternion and then normalize it?

erikwijmans: We currently keep everything in Magnum's scene graph as a 4x4 matrix, so we'd have to go 3x3 Matrix -> Quat -> Normalize -> 3x3 Matrix, which is likely slower than Gram Schmidt.  If we wanted to, we could switch to keeping things as https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1BasicDualQuaternionTransformation.html but we'd need to add support for scaling (I think we use scaling somewhere?)

erikwijmans: This is what we currently use: https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1BasicTranslationRotationScalingTransformation3D.html

mosra: > I wouldn't even do newRotation.isOrthogonal().

:+1: to that from me as well, the check involves a lot of special casing due to the nature of floating point comparisons and the orthonormalization alone should be fast enough to not need this.

> Why not manually construct the quaternion and then normalize it?

I'd rather add some `Quaternion::fromMatrixYesIKnowItsNotAPureRotationTrustMe()` than ending up with the (relatively complex) quat-to-matrix code unnecessarily duplicated here in habitat :) There's `Math::Implementation::quaternionFromMatrix()` that's just the algorithm without the assertion, but it's a private API that might disappear / get changed at any time. Other ideas? Some time ago I was wondering if we could switch to a [translation/rotation/scaling](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1BasicTranslationRotationScalingTransformation3D.html) (or maybe even dual quaternion) transformation implementation instead of a pure matrix-based one, that could solve this as well, plus make various other operations faster.

mosra: Argh, comment clash, sorry :)

@erikwijmans huh, I thought we're still at pure matrices. If we use a TRS transformation then we could operate directly with quats, no? Basically the `integrateTransform()` could take a (scaling +) quat + translation and operate on that instead of a 4x4 matrix?

erikwijmans: Is TRS already using a quat internally? For some reason I thought it was 4x4, but I could definitely be wrong and a quat would make may more sense :)

erikwijmans: We do use TRS for the scene graph: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/gfx/magnum.h#L21

jturner65: > 
> 
> We currently keep everything in Magnum's scene graph as a 4x4 matrix, so we'd have to go 3x3 Matrix -> Quat -> Normalize -> 3x3 Matrix, which is likely slower than Gram Schmidt. If we wanted to, we could switch to keeping things as https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1BasicDualQuaternionTransformation.html but we'd need to add support for scaling (I think we use scaling somewhere?)

I think the complexity is much less than gramm schmidt.  The code already isolates the rotational component, which should be the only component causing loss of orthogonality (otherwise we'd deform, right?)  
3x3 Mat -> Quat is the following : 

qw= √(1 + m00 + m11 + m22) /2
qx = (m21 - m12)/( 4 *qw)
qy = (m02 - m20)/( 4 *qw)
qz = (m10 - m01)/( 4 *qw)

normalizing the quat is trivial, and converting the quat back to a 3x3 is :

1 - 2*qy2 - 2*qz2 | 2*qx*qy - 2*qz*qw | 2*qx*qz + 2*qy*qw
2*qx*qy + 2*qz*qw | 1 - 2*qx2 - 2*qz2 | 2*qy*qz - 2*qx*qw
2*qx*qz - 2*qy*qw | 2*qy*qz + 2*qx*qw | 1 - 2*qx2 - 2*qy2

mosra: Operating directly on a quat would be the fastest, and since the TRS transformation implementation uses a quat internally, we don't really need to go back and forth between a 3x3 matrix and a quat here I think. The reorthonormalization needed to account for FP drifts then simplifies to just normalizing a quat.

erikwijmans: Let's see, GramSchmidt on a 3x3 is 3 vector normalizes, 3 dot products, and 3 vector-vector subtractions.  So (9 + 9) + 9 + 0 multiplies/divides, 6 + 6 + 9 add/subtract, and 3 square roots, so it's probably quicker to do 3x3 mat -> quat -> 3x3 mat, but very close!

aclegg3: Thanks for the great conversation here. From this I decided to refactor from 4x4 Matrices to translation vector and rotation quaternion. This skips several conversions and should be more stable. Let me know what you think.

PR #602, opened 2020-04-25
Add python bindings for getObjectInitializationTemplate

aclegg3:
## Motivation and Context

When an object is initialized, a copy of its construction parameters template is saved. Querying this allows for static properties of the object (unchanging after instancing) to be queried including mass, scale, friction, and restitution.

## How Has This Been Tested

Consumed this functionality in a test. If queried mass is wrong, the test will fail.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/sim/Simulator.cpp
+++ b/src/esp/sim/Simulator.cpp
@@ -309,6 +309,14 @@ int Simulator::loadObjectTemplate(
   return resourceManager_.loadObjectTemplate(objTmplPtr, objectTemplateHandle);
 }
 
+const assets::PhysicsObjectAttributes::ptr
+Simulator::getObjectInitializationTemplate(int objectId, const int sceneID) {
+  if (sceneHasPhysics(sceneID)) {
+    return physicsManager_->getInitializationAttributes(objectId);
+  }
+  return nullptr;
+}
+
 // return a list of existing objected IDs in a physical scene
 std::vector<int> Simulator::getExistingObjectIDs(const int sceneID) {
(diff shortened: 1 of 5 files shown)


src/esp/sim/Simulator.cpp
```diff
@@ -309,6 +309,14 @@ int Simulator::loadObjectTemplate(
   return resourceManager_.loadObjectTemplate(objTmplPtr, objectTemplateHandle);
 }
 
+const assets::PhysicsObjectAttributes::ptr
+Simulator::getObjectInitializationTemplate(int objectId, const int sceneID) {
+  if (sceneHasPhysics(sceneID)) {
+    return physicsManager_->getInitializationAttributes(objectId);
+  }
+  return nullptr;
+}
```
jturner65, line 318: Perhaps make getter const?

tests/test_physics.py
```diff
@@ -150,8 +152,9 @@ def test_dynamics(sim):
 
             # TODO: expose object properties (such as mass) to python
             # Counter the force of gravity on the object (it should not translate)
-            obj_mass = 0.41
-            sim.apply_force(np.array([0, obj_mass * 9.8, 0]), np.zeros(3), object_id)
+            sim.apply_force(
+                np.array([0, object1_mass * 9.8, 0]), np.zeros(3), object_id
```
Skylion007, line 156: 'G' (9.8) should really be defined as a constant somewhere, shouldn't it?

aclegg3: Thanks, some of this was written with early bindings.

PR #603, opened 2020-04-26
Fix vertex colored

erikwijmans:
## Motivation and Context

The vertex colored shader for ply meshes currently doesn't work, we rely on this for Replica meshes for webgl.

## How Has This Been Tested

Visually (I need to add a replica scene to the list of sensors).

## Types of changes

### DIFF
--- a/src/esp/gfx/LightSetup.cpp
+++ b/src/esp/gfx/LightSetup.cpp
@@ -28,7 +28,7 @@ Magnum::Vector3 getLightPositionRelativeToCamera(
       return light.position;
   }
 
-  CORRADE_ASSERT_UNREACHABLE();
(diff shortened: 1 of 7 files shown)


src/esp/gfx/LightSetup.cpp
```diff
@@ -28,7 +28,7 @@ Magnum::Vector3 getLightPositionRelativeToCamera(
       return light.position;
   }
 
-  CORRADE_ASSERT_UNREACHABLE();
+  CORRADE_INTERNAL_ASSERT_UNREACHABLE();
   return {};
```
mosra, line 32: Since you're touching this anyway, the return statement isn't needed here -- that's actually why this unreachable macro exists :)

PR #605, opened 2020-04-28
Attributes string search

jturner65:
## Motivation and Context
This PR adds the ability to search by substring within the existing library of object templates in resource manager.  The user provides a substring to search for and a vector/list of all template handles that contain that substring are returned.  An empty substring returns all available handles. This functionality was extended to python bindings as well, and requisite testing was added.

Const correctness was also enforced in pure getters in Resource Manager.

## How Has This Been Tested
C++ tests were added to verify the empty substring will return all handles and a partial string built from a known handle will return the full handle.  Python tests were modified so that the handle corresponding to the desired object ID was retrieved and used to instantiate the desired object.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -73,10 +73,6 @@ void initSimBindings(py::module& m) {
            "object_id"_a)
       .def("get_template_handles", &Simulator::getObjectTemplateHandles,
            "search_str"_a = "")
-      .def("get_file_template_handles",
-           &Simulator::getFileBasedObjectTemplateHandles, "search_str"_a = "")
-      .def("get_prim_template_handles", &Simulator::getPrimitiveTemplateHandles,
-           "search_str"_a = "")
       .def("add_object", &Simulator::addObject, "object_lib_index"_a,
            "attachment_node"_a, "light_setup_key"_a, "scene_id"_a = 0)
       .def("add_object_by_handle", &Simulator::addObjectByHandle,
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -68,8 +68,17 @@ void initSimBindings(py::module& m) {
                     &Simulator::setFrustumCullingEnabled,
                     R"(Enable or disable the frustum culling)")
       /* --- Physics functions --- */
+      .def("get_template_handles", &Simulator::getObjectTemplateHandles,
+           "search_str"_a = "")
+      .def("get_file_template_handles",
+           &Simulator::getFileBasedObjectTemplateHandles, "search_str"_a = "")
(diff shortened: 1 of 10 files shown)


tests/test_physics.py
```diff
@@ -77,7 +80,9 @@ def test_kinematics(sim):
     old_object_id = sim.remove_object(object_id)
     assert len(sim.get_existing_object_ids()) == 0
 
-    object_id = sim.add_object(0)
+    obj_handle = sim.get_template_handle_by_ID(0)
```
aclegg3, line 83: Let's unify all python tests for a known object using the "data/objects/cheezit" substring instead of ID->handle.

src/utils/viewer/viewer.cpp
```diff
@@ -627,6 +627,26 @@ void Viewer::keyPressEvent(KeyEvent& event) {
...
+      strdat.str("");
+      strdat.clear();
+      for (auto s : loadedHandles) {
+        strdat << s << " | ";
+      }
+
+      LOG(INFO) << "Loaded Object Templates matching 'chee' : " << strdat.str();
+
+      break;
+    }
```
aclegg3, line 649: Good for testing, but seems bulky. Is this necessary for viewer demo? Can we move to SimTest.cpp?

src/esp/bindings/SimBindings.cpp
```diff
@@ -68,8 +68,20 @@ void initSimBindings(py::module& m) {
...
       /* --- Physics functions --- */
+
+      .def("get_template_handle_by_ID", &Simulator::getObjectTemplateHandleByID,
+           "object_id"_a)
+      .def("get_template_handles", &Simulator::getObjectTemplateHandles,
+           "search_str"_a = "")
+      .def("get_file_template_handles",
+           &Simulator::getFileBasedObjectTemplateHandles, "search_str"_a = "")
+      .def("get_prim_template_handles", &Simulator::getPrimitiveTemplateHandles,
+           "search_str"_a = "")
```
aclegg3, line 79: Since these will be unified soon do we need this distinction in Simulator?

src/utils/viewer/viewer.cpp
jturner65: Just a debugging/development tool.  I can get rid of it entirely.

src/esp/bindings/SimBindings.cpp
jturner65: Well, as I mentioned earlier, I felt it might have worth getting templates to the non-file-based objects (perhaps instead of calling them "prim_templates" they could be "generated_templates" ?).  Again, not sure if folks would care from a usage perspective, but if so, i wanted to make it available.

tests/test_physics.py
jturner65: are we sure it is always cheezit?  If so, we can just search on "chee" :).

PR #607, opened 2020-04-29
Physics Primitives : PrimitiveImporter and PrimitiveAssetAttributes

jturner65:
## Motivation and Context
To instantiate primitive assets, we wish to use a configuration object to hold desired parameters and feed this object to a PrimitiveImporter, which provides all the primitive object instantiation functionality we require.  This PR sets the groundwork for this functionality : 

1. Attributes : Primitive asset template attribute classes with appropriately named functions are added so that the underlying configuration object will be usable by PrimitiveImporter to instantiate an appropriate object.  These templates will synthesize their own handles internally based on their parameters when any parameter changes.

2.  PrimitiveImporter : cmake files have been modified appropriately to reference necessary PrimitiveImporter functionality.  ResourceManager has been modified to instance a PrimitiveImporter via PluginManager.

3. All references to "AbstractPhysicsAttributes::XXXXMeshHandle" have been changed to "AbstractPhysicsAttributes::XXXXAssetHandle" to more accurately reflect the role these handles are expected to play once primitive assets become generally available.

## How Has This Been Tested
run_tests and pytest executed successfully

## Types of changes

## Checklist


mosra:
And I'm too late to this, again. General question -- can we *please* have at least a 24-hour window before opening a PR and merging it, unless it's straightforward enough? This one got opened at 8 PM in my timezone (at which time I didn't have enough brain cells anymore to absorb all the new code) and because I was finishing up a leftover task from yesterday I didn't have time to look at this until now.

> `getBoolDispStr()`, `CYLINDER_WF`, `primitiveAssetTmpltLibByID_`, `physicsSynthObjTmpltLibByID_`, ...

I have to admit I don't like the direction in which these names are going ... to me the last one could as well be just `b` and it wouldn't make it any less meaningful. Is `primitiveAssetTemplateLibraryByID_` really too long to type? Abbreviation overuse makes the code harder to read than it could be. (Of course if I'm the only one who has a problem with this feel free to ignore me.)

Skimming over the rest of the code, I'm a bit surprised that what were originally just [pure functions](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Primitives.html#a3293ada161344543b8341f486b3f62c6) in the `Primitives` library is now a complex class hierarchy with virtual calls, severals getters and setters and a caching scheme. In the beginning I suggested to use either the `Primitives` library directly or the [PrimitiveImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PrimitiveImporter.html), whichever would result in less code. I thought, since we already have a lot of code operating with generic `Importer` plugins, the `PrimitiveImporter` would be easier to integrate as it's just switching from `AnySceneImporter` to `PrimitiveImporter` and adding a possibility to extract just a particular mesh. (That's precisely why it exists, so I have a possibility to inject primitives into apps that can only open files.) I didn't expect it to become *this* :sweat_smile: The `Primitives` library itself is also [exposed in full](https://doc.magnum.graphics/python/magnum/primitives/) to Python, while all the complex code here needs completely fresh bindings. 

What's probably the most complex part of this is the `buildOriginHandleIndiv()` (what's `Origin` and `Indiv` btw?) -- do we really need to expose all of the flexibility and then cache every single variant under a unique handle? Why not cache just one base variant (that's smooth enough and has all important attributes) and then provide a flexible-but-uncached API when someone really needs a circle with exactly 76 vertices (in which case they could just supply the `MeshData` directly from the `Primitive` library anyway)?

I wasn't fully involved in the discussion leading to this and I have only very little understanding of the concept behind physics attributes and template object libraries etc., so if there's some high-level architecture I'm missing which warrants existence of all this code, I'd be happy to know more about it. However from a pure GPU developer perspective, if our goal is to reduce memory consumption, overhead and move towards immutable memory-mapped assets shared among several instances, this new code makes it harder, not easier.

msavva:
I agree with many of the points raised by @mosra here.  When it comes to changes such as in this PR that have a sizable footprint on the codebase, it would be good to go through a more deliberate process.  I propose that we clarify the design goals surrounding templates, have an API sketch and iterate on that once before we go into PRs.  @jturner65 @aclegg3 @mosra let's followup regarding the above on slack.

PR #612, opened 2020-05-02
Material-editing for render assets

eundersander:
##  Motivation and Context

This is a python API and C++ backend to change some material properties for a "render asset" (mesh) at runtime. Specifically, we let you change colors and shininess but not textures or other properties. Once you change the render asset's material, any current objects using that render asset are immediately affected. Any future objects that use that render asset will also be affected.

This PR also provides `overrideMaterialForSubTree` and `clonePhongMaterial`, which will be used in a later PR for material-editing for Drawables (object instances).

I'm reviving this old PR. I will post some earlier feedback as inline comments.

material_tutorial2

## How Has This Been Tested

I added test_material.py and modified lighting_tutorial.py to exercise the material-editing API.

## Types of changes

## Checklist


examples/tutorials/material_tutorial.py
```diff
@@ -0,0 +1,190 @@
+# [setup]
```
undersan, line 1: much of this file is copy-pasted from `lighting_tutorial.py`.

examples/tutorials/material_tutorial.py
```diff
@@ -0,0 +1,190 @@
...
+    agent_transform = place_agent(sim)
+
+    # [example 1]
+
+    # load some object templates from configuration files
+    sphere_template_id = sim.load_object_configs(
+        str(os.path.join(data_path, "test_assets/objects/sphere"))
+    )[0]
+    engine_template_id = sim.load_object_configs(
+        str(os.path.join(data_path, "test_assets/objects/2CylinderEngine"))
```
undersan, line 94: See zip file linked in the PR description.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1584,12 +1598,16 @@ gfx::PhongMaterialData::uptr ResourceManager::buildFlatShadedMaterialData(
   } else {
     finalMaterial->ambientColor = material.ambientColor();
   }
+
+  finalMaterial->importName = importName;
```
undersan, line 1602: We want to give users access to the "import name"--the material name in the source gltf file.

src/esp/gfx/Drawable.h
```diff
@@ -46,6 +48,9 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
 
   virtual void setLightSetup(const Magnum::ResourceKey& lightSetup){};
 
+  virtual Magnum::ResourceKey getOriginalMaterial() { return Magnum::ResourceKey{}; }
```
undersan, line 51: These are needed to support the "find-and-replace" logic in `overrideMaterialForSubTree`.

src/esp/gfx/MaterialData.h
```diff
@@ -25,9 +25,21 @@ struct PhongMaterialData : public MaterialData {
   Magnum::GL::Texture2D *ambientTexture = nullptr, *diffuseTexture = nullptr,
                         *specularTexture = nullptr, *normalTexture = nullptr;
   bool perVertexObjectId = false, vertexColored = false;
+  std::string importName;
 
   ESP_SMART_POINTERS(PhongMaterialData)
 };
 
+struct PythonMaterial {
```
erikwijmans, line 33: Instead of making a port for python, let's have the phong material data have two parts, the material assets (the textures) and the material parameters (these things).  Material assets don't have a reason to go to python but then the material parameters can.

src/esp/gfx/MaterialData.h
```diff
@@ -16,11 +16,19 @@ namespace gfx {
 
 struct MaterialData {};
 
-struct PhongMaterialData : public MaterialData {
+// This struct contains a subset of phong material fields (with the rest in
+// PhongMaterialData). We expose this struct to python so that scripts can
+// modify these fields at runtime.
+struct PhongMaterialInfo {
```
undersan, line 22: I named this `PhongMaterialInfo` to be consistent with `LightInfo` which we also expose to python.

examples/tutorials/nb_python/material_tutorial.py
```diff
@@ -0,0 +1,198 @@
...
+            )
+            sim.override_object_render_asset_material(
+                id, material_index, randomized_material
+            )
+
+        show_obs(sim)
+# %%
+    # place another test object, a torus stack. We don't modify any materials yet.
+    engine_template_id = obj_templates_mgr.load_object_configs(
+        str(os.path.join(data_path, "test_assets/objects/torus_stack"))
```
eundersander, line 164: I need to get this model added to `http://dl.fbaipublicfiles.com/habitat/objects_v0.1.zip` @aclegg3 . (We needed a test model that has multiple materials.)

examples/tutorials/colabs/material_tutorial.ipynb
```diff
@@ -0,0 +1,261 @@
+{
```
eundersander, line 1: I guess there is no easy way to test this Colab until this PR merges and my C++ changes get into the Hab-sim conda build.

eundersander, line 1: Note this file is auto-generated from `material_tutorial.py`.

examples/tutorials/nb_python/material_tutorial.py
```diff
@@ -0,0 +1,198 @@
...
+
+    # reduce the specular highlights (make them darker)
+    material.specular_color *= 0.3
+
+    # set diffuse to dark green
+    material.diffuse_color = mn.Color4(r=0.2, g=0.4, b=0.2)
+
+    # change material for all existing and future uses of this render asset, including
+    # the existing sphere.
+    sim.set_render_asset_material(render_asset_handle, material_index, material)
```
eundersander, line 128: From the earlier PR, there was an idea/suggestion to expose the object you're modifying to python, so instead of calling get_render_asset_material and then set_render_asset_material, you just call get and modify the returned object.

eundersander: I'm not a fan of this. I don't particularly like the idea that every modification to a python PhongMaterialInfo object immediately translates to a call into C++ to modify the underlying C++ object. It seems more efficient to set up all the fields of your PhongMaterialInfo in python and then send them all down to C++ in one call to `set_render_asset_material`.

In addition, I don't see this object-expose paradigm used anywhere else in hab-sim. Meanwhile, setters and getters are common, see `set_transformation`, `get_transformation`, `set_rigid_state`, `get_rigid_state`, `set_light_setup`, `get_light_setup`. I'm sticking with that convention.

examples/tutorials/nb_python/material_tutorial.py
```diff
@@ -0,0 +1,198 @@
...
+    # could also use sim.get_object_template(sphere_template_id) instead of
+    # get_object_initialization_template
+    render_asset_handle = sim.get_object_initialization_template(
+        sphere_ids[0]
+    ).render_asset_handle
+    num_materials = sim.get_num_render_asset_materials(render_asset_handle)
+    assert num_materials == 1
+    material_index = 0
+
+    material = sim.get_render_asset_material(render_asset_handle, material_index)
```
eundersander, line 115: A concern from the earlier PR: I'm not exposing our C++ material IDs to python. Instead, the python can only refer to a material by `(render_asset_handle, material_index)`. This will be a problem if we eventually get materials from other sources, such as created from scratch or loaded from something other than a render asset. More broadly, we'd like a consistent way of referencing assets in general (materials, meshes, textures, other things?) at the python level.

eundersander: After discussion with the infra team, I'm punting for now. I've mostly left the API as-is. I've removed the python API for overriding a material for an object because that was probably the strangest API here and most likely to change. I'm open to changes here before merging this PR.

tests/test_material.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    assert new_material != existing_material
+
+    # change one of the render asset's materials
+    sim.set_render_asset_material(render_asset_handle, material_index, new_material)
+
+    updated_material = sim.get_render_asset_material(
+        render_asset_handle, material_index
+    )
+    assert updated_material == new_material
```
eundersander, line 55: I'm reminded of how difficult it is to unit-test or integration-test graphics features. We might consider building a graphical test framework that lets us output RGB images and then automatically checks that those images don't change over time. This would be easier than trying to write asserts on pixel values.

examples/tutorials/nb_python/material_tutorial.py
```diff
@@ -0,0 +1,198 @@
...
+        random.seed(5)
+        for id in sphere_ids:
+            randomized_material = PhongMaterialInfo(material)
+            randomized_material.specular_color *= random.uniform(0, 2)
+            randomized_material.diffuse_color = mn.Color4(
+                r=material.diffuse_color.r + random.uniform(-0.1, 0.1),
+                g=material.diffuse_color.g + random.uniform(-0.1, 0.1),
+                b=material.diffuse_color.b + random.uniform(-0.1, 0.1),
+            )
+            sim.override_object_render_asset_material(
```
eundersander, line 156: This is probably the strangest API here and most likely to change, so I just removed it from this PR. See this slack discussion: https://cvmlp.slack.com/archives/G0131KVLBLL/p1597770680011800 .

This PR still has the C++ backend, `Simulator::overrideObjectRenderAssetMaterial`, which is unfortunately not tested at all in this PR.

examples/tutorials/nb_python/material_tutorial.py
jturner65: The way all the attributes templates are handled is that the user gets a copy of what is held in the library.  The various editable quantities are treated as a properties of the python-bound user's copy of the object, and when editing is completed, the user's copy of the template is registered, adding/overwriting the library object.  

Now, these are templates that are used to construct things, not actual production constructions, so this paradigm might not work for materials.

src/esp/gfx/MaterialUtil.h
```diff
@@ -0,0 +1,27 @@
...
+
+// todo: pass resourceKey by value?
+gfx::PhongMaterialInfo getPhongMaterialInfo(ShaderManager& shaderManager,
+                                            Magnum::ResourceKey key);
+
+void updatePhongMaterialInfo(ShaderManager& shaderManager,
+                             Magnum::ResourceKey key,
+                             const gfx::PhongMaterialInfo& materialInfo);
+
+void overrideMaterialForSubTree(scene::SceneNode& root,
```
jturner65, line 23: Please comment what is going on :).

tests/test_material.py
jturner65: There is support for that kind of thing now : checkPinholeCameraRGBAObservation in SimTest.cpp.

Skylion007:
You should add an entry to test_examples to test the entire colab script.

examples/tutorials/nb_python/material_tutorial.py
aclegg3: Can we use material keys here for this? Would be almost exactly the same API as [Light Setups](https://aihabitat.org/docs/habitat-sim/lighting-setups.html)
Something like:
```
material_key = sim.get_render_asset_material_key(render_asset_handle, material_index)
material = sim.get_material(material_key) #this returns a copy of the material info
#or to create a new material
material = habitat_sim.gfx.PhongMaterialInfo()
#then edit the material properties here before setting
sim.set_material(material, material_key) #override or register a key
sim.set_render_asset_material_key(render_asset_handle, material_index, material_key)
```

This will require changes to how we define material keys for assets from ids:
in [MeshMetaData](https://github.com/facebookresearch/habitat-sim/blob/378c6f4287116ff90aae67dd5a4be12bdf884b2e/src/esp/assets/MeshMetaData.h#L80) and [MeshTransformNode
](https://github.com/facebookresearch/habitat-sim/blob/378c6f4287116ff90aae67dd5a4be12bdf884b2e/src/esp/assets/MeshMetaData.h#L31) in [loadMaterial](https://github.com/facebookresearch/habitat-sim/blob/378c6f4287116ff90aae67dd5a4be12bdf884b2e/src/esp/assets/ResourceManager.cpp#L1122)
in [addComponent](https://github.com/facebookresearch/habitat-sim/blob/378c6f4287116ff90aae67dd5a4be12bdf884b2e/src/esp/assets/ResourceManager.cpp#L1528)
and elsewhere in this pipeline. Seems like all of these could be generic string keys now.

docs/pages/lighting-setups.rst
```diff
@@ -212,3 +212,17 @@ An existing object's light setup can be modified at any time with :ref:`Simulato
 .. code:: python
 
     sim.set_object_light_setup(my_object_id, "my_custom_lighting_key")
+
```
eundersander, line 215: I didn't flesh out this RST content with new screenshots or otherwise because this is all throwaway. We're going to replace this RST tutorial with a Colab-based tutorial soon.

examples/tutorials/nb_python/material_tutorial.py
eundersander: Cool, I've used `transform_box`. This entailed adding `transform_box.phys_properties.json`.

examples/tutorials/nb_python/material_tutorial.py
eundersander: Ok, I'll take a stab at this.

tests/test_material.py
mosra: ... and mainly the [DebugTools::CompareImage](https://doc.magnum.graphics/magnum/classMagnum_1_1DebugTools_1_1CompareImage.html) magic in Magnum :)

examples/tutorials/lighting_tutorial.py
```diff
@@ -218,6 +221,78 @@ def main(show_imgs=True, save_imgs=False):
...
+    # get the render asset for the box template
+    render_asset_handle = sim.get_object_initialization_template(
+        id_1
+    ).render_asset_handle
+
+    # create a new material: purple with a yellow specular highlight.
+    new_material = PhongMaterialInfo(
+        ambient_color=mn.Color4(0.15, 0.0, 0.15, 1),
+        diffuse_color=mn.Color4(0.15, 0.0, 0.15, 1),
+        specular_color=mn.Color4(1.0, 1.0, 0.0, 1),
```
mosra, line 256: These could be all just lists / tuples I think, the actual type is not important:

```suggestion
        ambient_color=(0.15, 0.0, 0.15, 1),
        diffuse_color=(0.15, 0.0, 0.15, 1),
        specular_color=(1.0, 1.0, 0.0, 1),
```

... and maybe even with the alpha omitted, but I'm not 100% sure if pybind handles such a deep conversion correctly. Same elsewhere if there's more similar cases.

Skylion007: ^Pretty sure it can actually, just might need some massaging to get it to work. You can definitely construct mn.Color4 for all the values in the tuples within the pybind lambda function. At the very least, PyBind can convert tuple to std::tuple and then you can to do the conversion in C++. You could also create a constructor fro mn.Color4 takes in a std::tuple etc.. Finally, you can just implement the cast yourself if needed: https://pybind11.readthedocs.io/en/stable/advanced/cast/custom.html

mosra:
> Instead of making a port for python, let's have the phong material data have two parts, the material assets (the textures) and the material parameters (these things). Material assets don't have a reason to go to python but then the material parameters can.

@erikwijmans @eundersander Based on my recently gained experience with how PBR models are usually structured, the textures are in most cases the *only* material inputs, so I'm afraid we can't just ignore those. If I remember correctly, both the image import API and the `GL::Texture` classes are fully exposed to python, so providing those fields in the python bindings shouldn't be a problem. And if you need equality comparison for those, comparing the GL `id()` could be a good enough tradeoff I think.

examples/tutorials/lighting_tutorial.py
mosra: Yeah I mean all things for this conversion (tuple constructors, implicit conversion marks etc.) should be in place, so I'm 90% sure it'll work -- but not 100% because pybind is impenetrable magic :))

eundersander:
Closing this old PR. The latest progress here is this open design question in a comment: https://github.com/facebookresearch/habitat-sim/pull/612#discussion_r478775517

PR #614, opened 2020-05-02
Update agent state set state to be less confusing

erikwijmans:
## Motivation and Context

The agent state setting is kinda confusing currently as the sensors don't follow the agent body by default, this leads to the anti-pattern of

```python
state = agent.state
state.rotation = something
agent.state = state
```
seemingly doing nothing!  This PR makes the more intuitive behavior of the sensors following the agent base the default.

## How Has This Been Tested

The example above now moves the sensors.

## Types of changes

- Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -202,21 +202,20 @@ class Agent(object):
         :param state: The state to set the agent to
         :param reset_sensors: Whether or not to reset the sensors to their
             default intrinsic/extrinsic parameters before setting their extrinsic state.
-
-            Setting this to `False` allows the agent base state to be moved and the new
-            sensor locations inferred without changing the configuration of the sensors
-            with respect to the base state of the agent.
-
-            Default: `True`
         :param infer_sensor_states: Whether or not to infer the location of sensors based on
             the new location of the agent base state.
-
-            Setting this to `False` is useful if you'd like to directly control
-            the state of a sensor instead of moving the agent.
-
-            Default: `True`
         :param is_initial: Whether this state is the initial state of the
             agent in the scene. Used for resetting the agent at a later time
+
+        Setting ``reset_sensors`` to :py:`False`
+        allows the agent base state to be moved and the new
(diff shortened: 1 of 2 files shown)


habitat_sim/agent/agent.py
```diff
@@ -191,14 +191,30 @@ def get_state(self) -> AgentState:
...
         :param reset_sensors: Whether or not to reset the sensors to their
-            default intrinsic/extrinsic parameters before setting their
-            extrinsic state
+            default intrinsic/extrinsic parameters before setting their extrinsic state.
+
+            Setting this to `False` allows the agent base state to be moved and the new
+            sensor locations inferred without changing the configuration of the sensors
+            with respect to the base state of the agent.
+
+            Default: `True`
```
mosra, line 210: The default is [shown in the docs already](https://aihabitat.org/docs/habitat-sim/habitat_sim.agent.Agent.html#set_state) so I don't think it's needed to be repeated here. Same below.

habitat_sim/agent/agent.py
```diff
@@ -216,20 +232,21 @@ def set_state(
...
+                    quat_rotate_vector(
+                        state.rotation.inverse(), v.position - state.position
+                    )
                 )
-            )
-            s.node.rotation = quat_to_magnum(state.rotation.inverse() * v.rotation)
+                s.node.rotation = quat_to_magnum(state.rotation.inverse() * v.rotation)
 
         if is_initial:
             self.initial_state = state
```
mosra, line 252: Looking at the snippet in the description I first thought the `state` property assignment wasn't working correctly and thus one had to use `set_state()` instead. There's [no docs for the property](https://aihabitat.org/docs/habitat-sim/habitat_sim.agent.Agent.html#state), so maybe it could be worth adding a docstring that explains relation to `set_state()`?

erikwijmans: Makes sense!  Added

PR #616, opened 2020-05-03
Improvements to Conda build

dhruvbatra:
Assorted improvements to README, build functions to keep track of version numbers, etc. 

CC: @mpiseno 

## Types of changes

### DIFF
--- a/conda-build/README.md
+++ b/conda-build/README.md
@@ -11,11 +11,11 @@ To then download the package, run ```conda install -c aihabitat -c conda-forge h
 
 The process is almost the same for linux; there is a corresponding python script for starting things off, however we use a docker container to do the builds. There is a dockerfile in this directory that you can use to create a container.
 
-```docker build -t mycontainer -f Dockerfile .```
+```docker build -t hsim_condabuild_dcontainer -f Dockerfile .```
 
 That will create your docker container. Now run
 
-```docker run -it --ipc=host --rm -v $(pwd)/../:/remote mycontainer bash```
+```docker run -it --ipc=host --rm -v $(pwd)/../:/remote hsim_condabuild_dcontainer bash```
 
 From there you will have a shell within your linux container. Now, navigate to ```cd /remote/conda-build``` where habitat-sim has been mounted. Create a conda environment within the linux container with python>=3.6 (identical to that needed by habitat-sim build). And then run ```python linux_matrix_build.py```, which will kick off the build process. After this has finished, upload it to anaconda cloud in the same way described in the macOS section.
 
--- a/conda-build/README.md
+++ b/conda-build/README.md
@@ -15,7 +15,7 @@ The process is almost the same for linux; there is a corresponding python script
 
 That will create your docker container. Now run
 
-```docker run -it --ipc=host --rm -v $(pwd)../:/remote mycontainer bash```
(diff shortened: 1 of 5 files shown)


conda-build/linux_matrix_builder.py
```diff
@@ -4,6 +4,12 @@
 import os.path as osp
 import itertools
 import os
+import time
+
+import sys
```
erikwijmans, line 9: I think these also need an `import builtins`

conda-build/linux_matrix_builder.py
```diff
@@ -38,6 +44,7 @@ def main():
         py_vers, bullet_modes, headless_modes, cuda_vers
     ):
         env = os.environ.copy()
+        env["VERSION"] = habitat_sim.__version__ + time.strftime(".%Y.%m.%d") # including a timestamp in anticipation of nightly builds
```
Skylion007, line 47: Why not just do a has of the current commit? No reason to add builds that are identical.

conda-build/macos_matrix_builder.py
```diff
@@ -34,6 +40,7 @@ def main():
 
     for py_ver, use_bullet in itertools.product(py_vers, bullet_modes):
         env = os.environ.copy()
+        env["VERSION"] = habitat_sim.__version__ + time.strftime(".%Y.%m.%d") # adding time-stamp in anticipation of nightly builds
```
Skylion007, line 43: Likewise for above

conda-build/README.md
```diff
@@ -11,11 +11,17 @@ To then download the package, run ```conda install -c aihabitat -c conda-forge h
 
 The process is almost the same for linux; there is a corresponding python script for starting things off, however we use a docker container to do the builds. There is a dockerfile in this directory that you can use to create a container.
 
-```docker build -t myContainer -f conda-build/Dockerfile .```
+```docker build -t mycontainer -f Dockerfile .```
```
Skylion007, line 14: We should probably actually give it an intelligent name like: habitat_sim_conda_build

conda-build/linux_matrix_builder.py
Skylion007: Requires gitpython library.
```python
import git
repo = git.Repo(search_parent_directories=True)
sha = repo.head.object.hexsha
```

dhruvbatra:
@erikwijmans @Skylion007 -- addressed. Let me know if this is ready to merge.

Skylion007:
Sorry to go back on my previous comment, but upon reading more of the conda
doc, it suggests that the commit hash and such should be stored in the
build string, not the version number:
https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html

I am not sure what the actual best way to version nightly releases in conda
is though.

Skylion007:
This is how torch does it (with the devprefix):
https://github.com/pytorch/pytorch/issues/33103#issuecomment-583805093

Skylion007:
Looks like we may have to increment the BUILD_NUMBER variable to sort the
versions properly.

dhruvbatra:
@Skylion007 -- I added date back to version and git SHA to build string. This is similar but not identical to PyTorch. I don't think we need to overoptimize this. It's a nightly build. We don't plan on having multiple nightly builds with the same date. This seems good enough to me.

PR #617, opened 2020-05-04
--Renamed GltfMeshData -> GenericMeshData;

jturner65:
## Motivation and Context
This PR renames the GltfMeshData class to be GenericMeshData, to more accurately reflect this class's applicability to many different mesh formats.  A method was added to instance a mesh object by the importer using a string name, in order to support mesh instancing using PrimitiveImporter.

## How Has This Been Tested
All existing c++ and python tests 

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/GenericMeshData.cpp
+++ b/src/esp/assets/GenericMeshData.cpp
@@ -2,20 +2,20 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-#include "GltfMeshData.h"
+#include "GenericMeshData.h"
 
 #include <Corrade/Containers/Array.h>
 #include <Corrade/Containers/ArrayViewStl.h>
+#include <Corrade/Utility/DebugStl.h>
 #include <Magnum/MeshTools/Compile.h>
 #include <Magnum/MeshTools/Interleave.h>
-
 namespace Cr = Corrade;
 namespace Mn = Magnum;
 
 namespace esp {
 namespace assets {
 
-void GltfMeshData::uploadBuffersToGPU(bool forceReload) {
+void GenericMeshData::uploadBuffersToGPU(bool forceReload) {
   if (forceReload) {
     buffersOnGPU_ = false;
   }
@@ -24,7 +24,7 @@ void GltfMeshData::uploadBuffersToGPU(bool forceReload) {
   }
 
   renderingBuffer_.reset();
-  renderingBuffer_ = std::make_unique<GltfMeshData::RenderingBuffer>();
+  renderingBuffer_ = std::make_unique<GenericMeshData::RenderingBuffer>();
   Magnum::MeshTools::CompileFlags compileFlags{};
   if (needsNormals_ &&
       !meshData_->hasAttribute(Mn::Trade::MeshAttribute::Normal)) {
@@ -36,7 +36,7 @@ void GltfMeshData::uploadBuffersToGPU(bool forceReload) {
   buffersOnGPU_ = true;
 }
 
-Magnum::GL::Mesh* GltfMeshData::getMagnumGLMesh() {
+Magnum::GL::Mesh* GenericMeshData::getMagnumGLMesh() {
   if (renderingBuffer_ == nullptr) {
     return nullptr;
   }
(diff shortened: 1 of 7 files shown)


src/esp/assets/ResourceManager.h
```diff
@@ -105,7 +105,7 @@ enum class PrimObjTypes : uint32_t {
    * DOES NOT EXIST
    * TODO: can/should this be added?
    */
```
mosra, line 107: The comment needs updating also I think?

src/esp/assets/GenericMeshData.cpp
```diff
@@ -36,26 +36,26 @@ void GltfMeshData::uploadBuffersToGPU(bool forceReload) {
...
-      importer.mesh(meshID);
-  if (meshData)
-    meshData_ = Mn::MeshTools::interleave(*std::move(meshData));
-  else
-    meshData_ = Cr::Containers::NullOpt;
+
+  CORRADE_ASSERT(
+      meshData_->primitive() == Mn::MeshPrimitive::Triangles,
+      "Cannot instance collisionMeshData_; Triangle Mesh expected.", );
+  meshData_ = Mn::MeshTools::interleave(*std::move(meshData_));
```
mosra, line 58: This question got kinda lost in our conversation and still needs to be clarified -- how are the wireframe primitives going to work with this assert in place?

Random thought: since the primitives are, well, primitive, maybe Bullet could get some builtin box/sphere/capsule collision shape in this case instead of a generic "bag of triangles"? That would solve the problem of wireframe primitives also. @aclegg3 is that feasible (also regarding other physic/collision APIs we use)?

src/esp/assets/GenericMeshData.cpp
```diff
@@ -36,26 +36,26 @@ void GltfMeshData::uploadBuffersToGPU(bool forceReload) {
...
   return &(renderingBuffer_->mesh);
 }
 
-void GltfMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
-                               int meshID) {
-  ASSERT(0 <= meshID && meshID < importer.meshCount());
+void GenericMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
+                                  int meshID) {
+  /* Guarantee mesh instance success */
+  CORRADE_INTERNAL_ASSERT_OUTPUT(meshData_ = importer.mesh(meshID));
```
mosra, line 50:
```suggestion
  CORRADE_INTERNAL_ASSERT(0 <= meshID && meshID < importer.meshCount());
  
  /* Guarantee mesh instance success */
  CORRADE_INTERNAL_ASSERT_OUTPUT(meshData_ = importer.mesh(meshID));
```

Somehow the original assert got lost? Tho Magnum will provide a helpful assertion in this case also, so it's probably redundant to check for meshID bounds here as well.

src/esp/assets/GenericMeshData.cpp
```diff
@@ -74,5 +74,13 @@ void GltfMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
     collisionMeshData_.indices = indexData_ = meshData_->indicesAsArray();
 }
 
+void GenericMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
+                                  const std::string& meshName) {
+  // make sure name is appropriate for importer
+  int meshID = importer.meshForName(meshName);
+  CORRADE_ASSERT(meshID != -1, "Unknown meshName : " << meshName, );
```
mosra, line 81:
```suggestion
  CORRADE_ASSERT(meshID != -1, "Unknown meshName :" << meshName, );
```

(spaces are inserted automatically in between)

src/esp/assets/GenericMeshData.cpp
jturner65: Oh, I specifically removed it thinking that it addressed all 3 of the original lines of code.  Should i replace the range check?

src/esp/assets/GenericMeshData.cpp
aclegg3: Yes, we will consume the implicit primitives and built-in shapes when possible. Also, it is not necessary to have triangle meshes to create convex collision shapes (just need a point cloud), so we could relax that requirement a bit for all but the concave scene meshes if necessary.

mosra: Oh, just a point cloud? In that case the index buffer (or the primitive type) would be unnecessary.

> concave

All primitives currently are convex, but *if* a torus will get ever added, that property won't hold anymore.

jturner65: The default collision handling is going to be implicit for primitives.  Perhaps we can just force all collision handling to be implicit (or at least all wireframe handling) until this can be addressed here?

aclegg3: I am fine with forcing built-in collision primitives for the time being. 👍

jturner65: Wow, weird.  I missed a conversation here, sorry about that :)

PR #623, opened 2020-05-07
Direct Navmesh/Meshdata handling; PluginManager instance variable

jturner65:
## Motivation and Context
This PR provides direct setting of MeshData  in GenericMeshData objects, to facilitate programmatically composed MeshData objects, such as the navMesh, in an ongoing effort to move away from custom "PrimitiveMesh" handling. A future PR will actually consume this functionality for NavMeshes.

The PluginManager object is also now treated as a protected instance variable for ResourceManager, instantiated in constructor, so that it does not need to be reinstanced everywhere an importer is required.  A primitive importer is provided as an instance variable, to be used whenever primitive meshes are needed, so it also does not need to be reinstanced.

Lastly, some existing functions were renamed/reworked to reflect recent changes with regard to attributes and how certain resources are stored/accessed.  Superfluous/redundant methods and variables were removed, and some methods were moved within ResourceManager to be closer to other similar methods.

## How Has This Been Tested
All existing tests (c++ and python) passed

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/GenericMeshData.cpp
+++ b/src/esp/assets/GenericMeshData.cpp
@@ -44,8 +44,7 @@ Magnum::GL::Mesh* GenericMeshData::getMagnumGLMesh() {
   return &(renderingBuffer_->mesh);
 }
 
-void GenericMeshData::setMeshData(
-    Corrade::Containers::Optional<Magnum::Trade::MeshData>& meshData) {
+void GenericMeshData::setMeshData(Magnum::Trade::MeshData&& meshData) {
   /* Interleave the mesh, if not already. This makes the GPU happier (better
      cache locality for vertex fetching) and is a no-op if the source data is
      already interleaved, so doesn't hurt to have it there always. */
@@ -53,7 +52,7 @@ void GenericMeshData::setMeshData(
   /* TODO: Address that non-triangle meshes will have their collisionMeshData_
    * incorrectly calculated */
 
-  meshData_ = Mn::MeshTools::interleave(*std::move(meshData));
+  meshData_ = Mn::MeshTools::interleave(std::move(meshData));
 
   collisionMeshData_.primitive = meshData_->primitive();
 
@@ -76,16 +75,18 @@ void GenericMeshData::importAndSetMeshData(
     Magnum::Trade::AbstractImporter& importer,
     int meshID) {
   /* Guarantee mesh instance success */
-  CORRADE_INTERNAL_ASSERT_OUTPUT(meshData_ = importer.mesh(meshID));
-  setMeshData(meshData_);
+  Cr::Containers::Optional<Mn::Trade::MeshData> mesh = importer.mesh(meshID);
+  CORRADE_INTERNAL_ASSERT(mesh);
+  setMeshData(*std::move(mesh));
 }  // importAndSetMeshData
 
 void GenericMeshData::importAndSetMeshData(
     Magnum::Trade::AbstractImporter& importer,
     const std::string& meshName) {
   /* Guarantee mesh instance success */
-  CORRADE_INTERNAL_ASSERT_OUTPUT(meshData_ = importer.mesh(meshName));
-  setMeshData(meshData_);
+  Cr::Containers::Optional<Mn::Trade::MeshData> mesh = importer.mesh(meshName);
+  CORRADE_INTERNAL_ASSERT(mesh);
+  setMeshData(*std::move(mesh));
 }  // importAndSetMeshData
 
 }  // namespace assets
--- a/src/esp/assets/GenericMeshData.cpp
+++ b/src/esp/assets/GenericMeshData.cpp
@@ -44,10 +44,8 @@ Magnum::GL::Mesh* GenericMeshData::getMagnumGLMesh() {
   return &(renderingBuffer_->mesh);
 }
 
-void GenericMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
-                                  int meshID) {
-  /* Guarantee mesh instance success */
-  CORRADE_INTERNAL_ASSERT_OUTPUT(meshData_ = importer.mesh(meshID));
+void GenericMeshData::setMeshData(
+    Corrade::Containers::Optional<Magnum::Trade::MeshData>& meshData) {
   /* Interleave the mesh, if not already. This makes the GPU happier (better
      cache locality for vertex fetching) and is a no-op if the source data is
      already interleaved, so doesn't hurt to have it there always. */
@@ -55,7 +53,7 @@ void GenericMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
   /* TODO: Address that non-triangle meshes will have their collisionMeshData_
    * incorrectly calculated */
 
-  meshData_ = Mn::MeshTools::interleave(*std::move(meshData_));
+  meshData_ = Mn::MeshTools::interleave(*std::move(meshData));
 
   collisionMeshData_.primitive = meshData_->primitive();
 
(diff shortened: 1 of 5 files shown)


src/esp/assets/GenericMeshData.cpp
```diff
@@ -44,18 +44,16 @@ Magnum::GL::Mesh* GenericMeshData::getMagnumGLMesh() {
   return &(renderingBuffer_->mesh);
 }
 
-void GenericMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
-                                  int meshID) {
-  /* Guarantee mesh instance success */
-  CORRADE_INTERNAL_ASSERT_OUTPUT(meshData_ = importer.mesh(meshID));
+void GenericMeshData::setMeshData(
+    Corrade::Containers::Optional<Magnum::Trade::MeshData>& meshData) {
```
mosra, line 48: Because the function takes ownership and expects the mesh data is valid, I would change the signature to

```suggestion
void GenericMeshData::setMeshData(Magnum::Trade::MeshData&& meshData) {
```

That (in my eyes at least) conveys better what is the function going to do with the argument.

The private `meshData_` member could still stay an optional (because the class needs an empty state I guess), but could also be just a plain `Trade::MeshData` with no index buffer, no attributes and 0 vertices by default.

src/esp/assets/GenericMeshData.cpp
```diff
@@ -72,15 +70,23 @@ void GenericMeshData::setMeshData(Magnum::Trade::AbstractImporter& importer,
...
-  int meshID = importer.meshForName(meshName);
-  CORRADE_ASSERT(meshID != -1, "Unknown meshName :" << meshName, );
-  setMeshData(importer, meshID);
-}
+void GenericMeshData::importAndSetMeshData(
+    Magnum::Trade::AbstractImporter& importer,
+    int meshID) {
+  /* Guarantee mesh instance success */
+  CORRADE_INTERNAL_ASSERT_OUTPUT(meshData_ = importer.mesh(meshID));
+  setMeshData(meshData_);
```
mosra, line 80: Not sure about this. It sets the `meshData_` member and then it calls `setMeshData()` which takes a reference to `meshData_` and sets it to `meshData_` again. If there wouldn't be the `interleave()` in between, that would basically be a self-assignment ;) What about this instead (together with the `setMeshData()` signature change mentioned above)?

```suggestion
  Cr::Containers::Optional<Mn::Trade::MeshData> mesh = importer.mesh(meshID));
  CORRADE_INTERNAL_ASSERT(mesh);
  setMeshData(*std::move(mesh));
```

Same in the function below.

src/esp/assets/GenericMeshData.h
```diff
@@ -52,23 +52,35 @@ class GenericMeshData : public BaseMesh {
...
-   * component. Sets the @ref collisionMeshData_ references.
-   * @param importer The importer pre-loaded with asset data from file.
+   * component ID. Sets the @ref collisionMeshData_ references.
+   * @param importer The importer pre-loaded with asset data from file, or a
+   * Primitive Importer.
    * @param meshID The local identifier of a specific mesh component of the
    * asset.
    */
-  void setMeshData(Magnum::Trade::AbstractImporter& importer, int meshID);
+  void importAndSetMeshData(Magnum::Trade::AbstractImporter& importer,
```
mosra, line 72:
```suggestion
  void setImportedMeshData(Magnum::Trade::AbstractImporter& importer,
```

Maybe? The "and set" seems superfluous to me. Same below. (It's a matter of personal preference I guess, feel free to ignore.)

src/esp/assets/ResourceManager.h
```diff
@@ -1110,6 +1077,18 @@ class ResourceManager {
   gfx::ShaderManager shaderManager_;
 
   // ======== Physical parameter data ========
+  /**
+   * @brief Plugin Manager used to instantiate importers which in turn are used
+   * to load asset data
+   */
+  Magnum::PluginManager::Manager<Importer> importManager;
```
mosra, line 1084:
```suggestion
  Corrade::PluginManager::Manager<Importer> importerManager_;
```

Keeping the PluginManager in namespace where it originated from to avoid confusion, and `importerManager` is commonly used in a lot of Magnum docs and examples so keeping the name the same. 

I had to pause on the `Importer` for a bit, took me a while to realize it's a typedef for `Trade::AbstractImporter`. Again to reduce confusion, maybe remove it and use the original name instead? It's not too long I think ;)

src/esp/assets/ResourceManager.cpp
```diff
@@ -72,12 +72,23 @@ constexpr char ResourceManager::DEFAULT_LIGHTING_KEY[];
...
+#endif
+{
   initDefaultLightSetups();
   initDefaultMaterials();
-}
+}  // namespace assets
 
 void ResourceManager::initDefaultPrimAttributes() {
+  // instantiate a primitive importer
+  CORRADE_INTERNAL_ASSERT(
```
mosra, line 88:
```suggestion
  CORRADE_INTERNAL_ASSERT_OUTPUT(
```

;)

src/esp/assets/GenericMeshData.cpp
jturner65: I was thinking about that, but I figured it was nice to have things in place. -shrug-. 

how about this, so that the import is enclosed (or is this not necessary i.e. will importer never fail?):

`  Cr::Containers::Optional<Mn::Trade::MeshData> mesh;
  CORRADE_INTERNAL_ASSERT_OUTPUT(mesh = importer.mesh(meshID));
`

src/esp/assets/GenericMeshData.h
jturner65: Wished to specify that "setMeshData" was also called.

src/esp/assets/GenericMeshData.cpp
mosra: Um... not sure about your point, your code and the first two lines in mine do practically the same, it's just that the assignment happens later.

src/esp/assets/GenericMeshData.cpp
```diff
@@ -44,16 +44,15 @@ Magnum::GL::Mesh* GenericMeshData::getMagnumGLMesh() {
   return &(renderingBuffer_->mesh);
 }
 
-void GenericMeshData::setMeshData(
-    Corrade::Containers::Optional<Magnum::Trade::MeshData>& meshData) {
+void GenericMeshData::setMeshData(Magnum::Trade::MeshData& meshData) {
```
mosra, line 47:
```suggestion
void GenericMeshData::setMeshData(Magnum::Trade::MeshData&& meshData) {
```

I really meant accepting a r-value here, to convey that there's an ownership transfer happening inside.

src/esp/assets/ResourceManager.h
```diff
@@ -1110,6 +1077,18 @@ class ResourceManager {
...
+   * to load asset data
+   */
+  Corrade::PluginManager::Manager<Importer> importerManager_;
+
+  /**
+   * @brief Importer used to synthesize Magnum Primitives.  This object allows
+   * for similar usage to File-based importers, but requires no file to be
+   * available/read.
+   */
+  Corrade::Containers::Pointer<Importer> primImporter;
```
mosra, line 1091:
```suggestion
  Corrade::Containers::Pointer<Importer> primImporter_;
```

Underscore here also? IIRC all privates + protected vars have them.

src/esp/assets/GenericMeshData.cpp
jturner65: Wasn't sure if we wanted to wrap the access/assignment in the assertion.  Guess that's not necessary.

src/esp/assets/GenericMeshData.cpp
jturner65: Oops.

src/esp/assets/GenericMeshData.cpp
mosra: Ultimately the assertion needs to check that `mesh` is not `NullOpt`, and it doesn't matter if it's done on the result of the assignment expression or separately. Apart from `mesh` being a null optional there's no other case where something could fail.

PR #625, opened 2020-05-08
Agent position random seeding

erikwijmans:
## Motivation and Context

Add random seed to the simulators config and expose core::Random to python so that the initial agent position/orientation is a) independent of np.random and b) consistent.

@mathfac I believe this won't effect habitat-api at all.  Habitat-api already calls `.seed` in habitat sim, so the random seed in the config here is irrelevant.

## How Has This Been Tested

Took the random seeding out of my little viewer script and the agent always spawns at the same location!

## Types of changes


dhruvbatra:
CC: @soumith

PR #626, opened 2020-05-09
Sharing of rendering results

erikwijmans:
## Motivation and Context

Habitat-sim currently has the ability to draw RGB, Depth, and Semantic observations all on the same render pass, however, this isn't currently leveraged!  This PR adds the ability for sensor to "borrow" eachother's render results so that if you draw for an RGB sensor and have a Depth sensor that is configured identically, the Depth sensor won't invoke another draw but will instead just read the Depth Buffer from the RGB draw. For Replica scenes, we can share between RGB, Depth, and Semantic!

One question for reviewers:  I also tried an implementation that copies between RenderTargets instead of sharing them.  While this seems safer, that copy obviously has a cost and grows with resolution.  We currently don't make any assumptions that would be violated by this direct sharing so maybe it is fine?  If we do want this copy, my testing says that the cost of doing the copy in `draw_observation` is pretty high (performance with GPU2GPU drops to bellow without for 256x256 for instance).  Doing the copy in `get_observation` is less costly.

@mosra I left the code to copy between framebuffers in the PR to see if you knew if there was a way to do it faster.

## How Has This Been Tested

New tests.

I re-ran the benchmark on the normal MP3D test scene and got

```
 ====== FPS (512 x 512, depth_only): 5624.3 =============
 ================ Performance (FPS) NPROC=1 ========
Resolution      rgb             rgbd            depth_only
128 x 128       3151.7          1510.4          2361.9
256 x 256       2750.8          1137.1          1893.8
512 x 512       1607.2          697.9           1315.9
 ====================================================
 ================ Performance (FPS) NPROC=3 =============
Resolution      rgb             rgbd            depth_only
128 x 128       7684.1          3423.9          5935.4
256 x 256       6339.5          2885.4          4878.6
512 x 512       3817.1          1929.5          3610.2
 ======================================================
 ================ Performance (FPS) NPROC=5 ===============
Resolution      rgb             rgbd            depth_only
128 x 128       10651.7         4819.9          9146.4
256 x 256       8940.7          4056.9          7848.4
512 x 512       6674.5          2939.8          5624.3
 =====================================================
```

Comparing to the existing results, this maybe matter for higher resolutions, but otherwise it suprisingly doesn't.  However, `17DRP5sb8fy` is a reasonably lightweight scene.  

`1LXtFkjw3qL` on the otherhand is much heavier and here it matters.  Before the PR:
```
 ================ Performance (FPS) NPROC=1 ==========
Resolution      rgb             rgbd            depth_only
128 x 128       1227.5          557.1           1129.8
256 x 256       1128.7          501.7           918.3
512 x 512       833.3           415.0           802.9
 =================================================
```

After:
```
 ================ Performance (FPS) NPROC=1 ========
Resolution      rgb             rgbd            depth_only
128 x 128       1289.9          868.7           1073.2
256 x 256       1270.8          855.0           1044.9
512 x 512       994.9           566.6           843.0
 ===============================================
```

Then with GPU2GPU transfer enabled we get very close to zero-cost RGB-D

```
 ================ Performance (FPS) NPROC=1 ======
Resolution      rgb             rgbd            depth_only
128 x 128       1319.1          909.5           1219.3
256 x 256       1426.3          924.4           1207.0
512 x 512       1364.3          959.6           1125.2
 ==========================================
```

## Types of changes

### DIFF
--- a/habitat_sim/sensors/sensor_suite.py
+++ b/habitat_sim/sensors/sensor_suite.py
@@ -1,10 +1,18 @@
-import habitat_sim.bindings as hsim
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from collections import OrderedDict
+
+from habitat_sim.sensor import Sensor
 
 
-class SensorSuite(dict):
+class SensorSuite(OrderedDict):
     r"""Holds all the agents sensors. Simply a dictionary with an extra method
     to lookup the name of a sensor as the key
     """
 
-    def add(self, sensor: hsim.Sensor):
+    def add(self, sensor: Sensor):
         self[sensor.specification().uuid] = sensor
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -4,8 +4,11 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+from __future__ import annotations
+
 import os.path as osp
 import time
+from collections import OrderedDict
 from typing import Any, Dict, List, Optional
 
 import attr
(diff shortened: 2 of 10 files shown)


Skylion007:
This looks like it will help speed up the habitat-challenge repos quite a bit! Well done.

Skylion007:
@erikwijmans Any reason this still hasn't been merged? Are we blocked by something?

erikwijmans:
The speedups aren't high enough to justify the ugliness that is this solution IMO.  @mosra and I had a chat when I originally made this and the result of that (after some initial resistance from me) is that a better design would be to have a VisualSensor that inherently supports rendering multiple modalities (this would also reduce the ugliness that is making an agent with RGB-D + Sem agent currently), but that is a larger change that I haven't had time to make yet.

PR #627, opened 2020-05-11
Updated m.css submodule.

mosra:
## Motivation and Context

Second part of the fix for https://github.com/facebookresearch/habitat-api/issues/385.

For the Doxygen 1.8.18 issue, the script now detects if Doxygen crashes while processing and reports that clearly. I have bisected the crash to a particular Doxygen commit, but still need to figure out why it happens and provide them with a minimal repro case so they can fix it (and then the 1.8.19 release might take several months). So we'll simply stay on 1.8.17 for as long as needed.

## How Has This Been Tested

Localized and reproduced the Py3.6-specific assertion, fixing its cause.

## Types of changes


dhruvbatra:
Plan sounds good, thank you!

Issue #552, opened 2020-03-30
Queries about using Matterport


MADHAVAN001:
## ❓ Questions and Help

Is the mesh provided in matterport habitat tasks, complete?

I downloaded the dataset using `python download_mp.py --task habitat -o path/to/download/`

I am using habitat_sim.utils.data.data_extractor's Image extractor to generate free poses, images and the number of images obtained from 17DRP5sb8fy is only 21.

I want to load the entire  house1, house2 mesh in habitat, how do I proceed about this?

dhruvbatra:
Could you please say more -- what are you trying to achieve?

CC: @mpiseno

MADHAVAN001:
I am trying to extract free poses in the environment where the agent can move around. For all such places, I want to extract the image that the agent sees.

```python
from habitat_sim.utils.data.data_extractor import ImageExtractor

scene_filepath = "/path/to/mp3d/17DRP5sb8fy/17DRP5sb8fy.glb"

extractor = ImageExtractor(
    scene_filepath,
    labels=[0.0],
    img_size=(512, 512),
    output=["rgba", "depth", "semantic"],
)

print(len(extractor[:1000]))

```

dhruvbatra:
I'm sure @mpiseno can let you know how to increase that number (21). 

However, please keep in mind that the number of places the agent can "stand" is infinite, so you can't get all such images.

MADHAVAN001:
I don't want an infinite number of images, but I want to control the number for my experiment.

Right now, I am not so sure about how to do that

mpiseno:
The way the image extractor works (by default) is by creating a grid of camera positions, and for each camera position, the camera angle is determined by finding the closest non-navigable point to that position. For some scenes this results in a low number of images (e.g. 21). Creating rules for extracting more images from each scene is an active area of work and is non-trivial, but there are currently 2 ways you can get more images from the extractor:

1. Set the extraction_method parameter to "panorama". This spins the camera around to get more camera poses rather than finding the closest non-navigable point.
2. Instead of 1 scene, you can pass in a list of ".glb" or ".ply" files instead of just 1. Just note that doing this can be very memory intensive because you'll be switching scenes a lot and it might slow down any training pipelines you have. 

Overall the image extractor is still a work in progress so please feel free to give feedback on existing features or features that would be helpful for the future.

MADHAVAN001:
This makes sense. 

[...]

PR #629, opened 2020-05-11
Modular pose extraction

mpiseno:
## Motivation and Context

One piece of feedback I got from several people was that the image extractor does not extract enough images for a single scene. One direction we can take to resolve this is to allow people to create their own custom rules for pose extraction. In this PR I have modularized the PoseExtractor class to subclass different types of pose extraction and support the creating of custom pose extractors. I will also add a section on to the tutorial about how to create a custom pose extractor in a future PR.

relevant issue: https://github.com/facebookresearch/habitat-sim/issues/552#issuecomment-606197026

## How Has This Been Tested

The current tests cover the two main ways of pose extraction already (closest and panorama).

## Types of changes

## Checklist


dhruvbatra:
Motivation makes sense to me. I know @arjunmajum already found this useful. 

Let me know when there's a tutorial writeup. I'd like to play around with this then.

mpiseno:
> Motivation makes sense to me. I know @arjunmajum already found this useful.
> 
> Let me know when there's a tutorial writeup. I'd like to play around with this then.

Will do, I'll likely have time to create a detailed tutorial within the next week.

habitat_sim/utils/data/data_extractor.py
```diff
@@ -12,7 +12,7 @@
 from habitat_sim.agent import AgentState
 from habitat_sim.utils.common import quat_from_two_vectors
 from habitat_sim.utils.data.data_structures import ExtractorLRUCache
-from habitat_sim.utils.data.pose_extractor import PoseExtractor
+from habitat_sim.utils.data.pose_extractor import *
```
erikwijmans, line 15: No `import *`

habitat_sim/utils/data/data_extractor.py
```diff
@@ -245,6 +249,18 @@ def _generate_label_map(self, scene, verbose=False):
 
         return instance_id_to_name
 
+    def _register_pose_extractor(self, method, args):
```
erikwijmans, line 252: We currently have a registry pattern implemented by `habitat_sim/registry.py`.  I suggest using that, that way users can easily ad their own pose extractor.

mpiseno: Wasn't 100% sure how the registry system worked, but I did my best to figure it out. Let me know if there's anything that jumps out that needs changing.

habitat_sim/utils/data/data_extractor.py
```diff
@@ -9,12 +9,28 @@
...
 import habitat_sim
 import habitat_sim.bindings as hsim
+import habitat_sim.registry as registry
 from habitat_sim.agent import AgentState
 from habitat_sim.utils.common import quat_from_two_vectors
 from habitat_sim.utils.data.data_structures import ExtractorLRUCache
 from habitat_sim.utils.data.pose_extractor import PoseExtractor
 
 
+def make_pose_extractor(name: str) -> PoseExtractor:
```
erikwijmans, line 19: This doesn't construct the class, just gets it it seems.

habitat_sim/utils/data/data_extractor.py
```diff
@@ -63,7 +79,7 @@ def __init__(
         if sum(split) != 100:
             raise Exception("Train/test split must sum to 100.")
 
-        assert extraction_method in ["closest", "panorama"]
+        # assert extraction_method in ["closest", "panorama"]
```
erikwijmans, line 82: Remove

PR #624, opened 2020-05-07
Documentation generator updates

mosra:
## Motivation and Context

Assorted updated to the documentation generator.

- Handles issues like in https://github.com/facebookresearch/habitat-api/issues/385 more gracefully. Unfortunately I wasn't able to reproduce locally (maybe it's something specific to a particular Python version, I'm on 3.8 here), so if this happens again it should print a descriptive warning and continue.
- Fixes non-deterministic output for default arguments containing function pointers, which means useless changes like in https://github.com/facebookmicrosites/habitat-website/commit/9cbead7f85aad0c30d86514e0bd790fe4dda8c1c won't happen anymore
- Creates output directory for Doxygen because Doxygen is unable to do that on its own.
- Adjusts itself for new bugs and regression in Doxygen 1.8.17 and 1.8.18.
- Properly includes CONTRIBUTING.md in the Doxyfile
- updates Doxyfile to latest Doxygen (it only removes a bunch of outdated options, should still keep working on older versions)

## How Has This Been Tested

Docs build correctly locally

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)

PR #630, opened 2020-05-12
Hopefully the final fix to yet another Doxygen failure

mosra:
## Motivation and Context

Followup to https://github.com/facebookresearch/habitat-sim/pull/624 based on a comment by @dhruvbatra. 

Yet another case of Doxygen unable to create a directory, however this time it fails silently. Sigh.
    
When generating public docs and the `build/docs/` directory didn't exist, Doxygen fails to create `build/docs/habitat-cpp.tag`, without any error, without any warning, nothing. To aid Doxygen in another case, m.css script creates `build/docs-public/` in that case, so tell Doxygen to put everything there, including the tagfile.

Also updated the Doxyfile to what 1.8.18 expects, getting rid of one warning in the output. (Tangential: the C++ docs are a *total mess*, not even my 5000 line terminal history scrollback was enough to fit all the warnings, and it seems to be getting progressively worse. Unless people actively try to minimize the warnings by building docs locally and checking the output, this will never improve.)

## How Has This Been Tested

Doc generation using `./build-public.sh` passes now even with the `build` directory not existing before.

## Types of changes

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -263,12 +263,6 @@ TAB_SIZE               = 4
 
 ALIASES                =
 
-# This tag can be used to specify a number of word-keyword mappings (TCL only).
-# A mapping has the form "name=value". For example adding "class=itcl::class"
-# will allow you to use the command class in the itcl::class meaning.
-
-TCL_SUBST              =
-
 # Set the OPTIMIZE_OUTPUT_FOR_C tag to YES if your project consists of C sources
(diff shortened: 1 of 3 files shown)


dhruvbatra:
Thank you for this! I tested this branch. Works. 

Re: C++ docs mess. Agreed. CC: @bigbike @jturner65 @aclegg3 @eundersander as the people working most closely on the C++ codebase.

PR #632, opened 2020-05-13
removing unnecessary mkdirs

dhruvbatra:
Following: https://github.com/facebookresearch/habitat-sim/pull/630


dhruvbatra:
It indeed passes CI tests!

PR #633, opened 2020-05-14
--Enable construction of PrimitiveAttributes using string name

jturner65:
Enable the construction of PrimitiveAttributes objects by accessing a map of preconfigured template function pointers keyed by string name of Magnum Primitive Class.

## Motivation and Context
As users wish to instantiate modified primitives to be used for rendering or collisions, we will want to instantiate new attributes on demand.  Instead of having a different instantiation method for each type of attributes, it would be ideal if we could reuse a single function, passing the desired type as a template parameter.  This functionality premaps calls to a template function that instantiates the appropriate shared pointer for each type of primitive currently supported.  Now, all that is required is knowledge of the name of the primitive desired to access the appropriate function to instantiate the desired primitive attributes.

## How Has This Been Tested
python and c++ tests have passed

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.h
+++ b/src/esp/assets/ResourceManager.h
@@ -126,6 +126,24 @@ constexpr const char* PrimitiveNames3D[]{
     "icosphereWireframe", "uvSphereSolid",      "uvSphereWireframe"};
 
 /**
+ * @brief Build a shared pointer to the appropriate attributes for passed object
+ * type
+ */
+template <typename T, bool isWF, int idx>
+std::shared_ptr<AbstractPrimitiveAttributes> createPrimitiveAttributes() {
+  return T::create(isWF, idx, PrimitiveNames3D[idx]);
+}
+
+/**
+ * @brief Define a map referencing function pointers to @ref
+ * createPrimitiveAttributes() keyed by string names of classes being instanced
+ */
+typedef std::map<
+    std::string,
+    std::shared_ptr<esp::assets::AbstractPrimitiveAttributes> (*)()>
+    map_of_primTypes;
+
+/**
  * @brief Singleton class responsible for
  * loading and managing common simulator assets such as meshes, textures, and
  * materials.
@@ -313,6 +331,14 @@ class ResourceManager {
    * which the key, synthesized from primAssetHandle, refers.
    */
   int buildAndRegisterPrimPhysObjTemplate(const std::string& primAssetHandle);
+
+  /**
+   * @brief This function will assign the appropriately configured function
+   * pointers for @ref createPrimitiveAttributes calls for each type of
+   * supported primitive to the @ref primTypeConstructorMap, keyed by type of
+   * primtive
+   */
+  void buildMapOfPrimTypeConstructors();
 
   /**
    * @brief Add a @ref PhysicsObjectAttributes object to the @ref
@@ -754,11 +780,6 @@ class ResourceManager {
                                   const bool requiresLighting);
 
   /**
-   * @brief put primitive asset template attributes in appropriate library
-   *
-   */
-
-  /**
    * @brief Put primitive-based synthesized object template attributes in
    * template map. Calls @ref addObjTemplateToLibrary with primTemplate's set
    * origin handle and prim map.  Convenience method.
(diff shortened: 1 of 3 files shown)


src/esp/assets/ResourceManager.h
```diff
@@ -125,6 +125,24 @@ constexpr const char* PrimitiveNames3D[]{
     "cylinderSolid",      "cylinderWireframe",  "icosphereSolid",
     "icosphereWireframe", "uvSphereSolid",      "uvSphereWireframe"};
 
+/**
+ * @brief Build a shared pointer to the appropriate attributes for passed object
+ * type
+ */
+template <typename T, bool isWF, int idx>
+std::shared_ptr<AbstractPrimitiveAttributes> createPrimitiveAttributes() {
```
bigbike, line 133: the `idx` is hard coded in the `void ResourceManager::buildMapOfPrimTypeConstructors()`, and it was actually **inferred** by the position of primitive names defined in `PrimitiveNames3D`. And it took me a while to figure it out (though may be it is obvious to the code author).

I do have concerns on such design.
- It is error-prone. Programmer in the future cannot change the order of this name list, cannot add names in the middle, but to the end. cannot deprecate any existing names etc. When changes are made, she has to remember to upgrade `buildMapOfPrimTypeConstructors`.  Too much burden would be added.
- Such `idx` cannot not easily figured out immediately especially when the list becomes long in the future. (For example, can you immediately tell what is correct index for `cylinderWireframe`?)

jturner65: Thanks for checking this out.  The limitation on the index is is by design.  The only primitives that are supported are those that are supported in magnum.  The index is not necessary to be known by the user, since it is known at compile time.  All that the user needs to use this functionality is the string name of the class.  Adding new primitives to this functionality should -not- be done lightly or arbitrarily.

jturner65: The template function itself should not be consumed directly, but rather through the function pointer mechanism.  I had originally put it within ResourceManager as a private method, but accessing it as a function pointer proved challenging.  Perhaps you could assist me with this?

bigbike: I am from the developer's perspective, not the user's. The primitives in Magnum might be extended, so is the name list here. 

What I can think of, for example, to use a `std::unordered_map<std::string, int> primitiveNameIndex` to map the primitive name to the `idx`. So `primitiveNameIndex["cylinderWireframe"]` would give you the index directly.

At least, alternatively I would do:
```cpp
constexpr const char* PrimitiveNames3D[]{
"capsule3DSolid",     // 0
"capsule3DWireframe", // 1
"coneSolid",          // 2 
...
};
```
Add also add a comment above the 
`template <typename T, bool isWF, int idx>`
/**
 * @brief Build a shared pointer to the appropriate attributes for passed object
 * type
 * The index should match the one listed in @ref PrimitiveNames3D
 */

bigbike: > The template function itself should not be consumed directly, but rather through the function pointer mechanism. I had originally put it within ResourceManager as a private method, but accessing it as a function pointer proved challenging. Perhaps you could assist me with this?

Not sure if I understand you correctly. Is it what you want as the following?
```cpp
#include <iostream>
class A {
    public:
    static void print(bool isWF, int idx) {
        std::cout << "from A: " << (isWF ? "TRUE, " : "FALSE, ") << idx <<std::endl;
    }
};

class B {
    public:
    static void print(bool isWF, int idx) {
        std::cout << "from B: " << (isWF ? "TRUE, " : "FALSE, ") << idx <<std::endl;
    }
};

class C
{
public:
 void callIt() { pointerToTemplateMemberFunction(); }
 
 protected:
 template <typename T, bool isWF, int idx>
 void method()
 {
     T::print(isWF, idx);
 }
 void pointerToTemplateMemberFunction();
};

void C::pointerToTemplateMemberFunction()
{
 C demo;
 void (C::*A_true_0)() = &C::method<A, true, 0>;
 void (C::*B_false_1)() = &C::method<B, false, 1>;
 
 (*this.*A_true_0)();
 (*this.*B_false_1)();
}

using namespace std;

int main()
{
    C c;
    c.callIt();

    return 0;
}
```

jturner65: Thanks Yili!  Great point in first post regarding the ambiguity of the in idx.  How about I use the enum I defined in the header?  That is, after all, why I put it there :D.  Thanks for the suggestions regarding the method pointer implementation, this helps.  I will adjust and update pressently :).

src/esp/assets/ResourceManager.h
```diff
@@ -686,7 +708,69 @@ class ResourceManager {
...
+                       static_cast<PrimObjTypes>(primTypeVal))])();
+  }  // buildPrimitiveAttributes
+
  private:
+  /**
+   * @brief Build a shared pointer to the appropriate attributes for passed
+   * object type as defined in @ref PrimObjTypes, where each entry except @ref
+   * END_PRIM_OBJ_TYPES corresponds to a Magnum Primitive type
+   */
+  template <typename T, bool isWF, PrimObjTypes primType>
```
bigbike, line 763: isWF --> isWireFrame

PR #635, opened 2020-05-19
--Make physical object asset loading/instantiation on demand

jturner65:
## Motivation and Context
This PR changes how rendering and collision assets for physical objects (both file-based and primitive-based) are loaded.  Now, on program start, the assets requested are verified to exist but not loaded/instantiated.  Instead the required assets for an object are instantiated when the first object using a particular asset is instanced.  This will allow many physical objects to be made available without loading all their assets immediately, nor filling memory with object assets that may not even be used.  The program also starts much faster now, since the load time for the assets has been deferred until they are actually used.

Also, checks are put in place in case a user decides to change a template's render or collision asset handle and forgets to register this modified template.  Now, whenever values are changed that would require re-registering (verifying existence of a mesh file, for instance) a flag is set within the template invisibly, which forces re-registration whenever that template is consumed.

Lastly, some big fixes and improvements (such as replacing pragma's for named header guards) are also added.

## How Has This Been Tested
All existing cpp and python tests have passed.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/Attributes.cpp
+++ b/src/esp/assets/Attributes.cpp
@@ -40,10 +40,12 @@ PhysicsObjectAttributes::PhysicsObjectAttributes(
   setInertia({0, 0, 0});
   setLinearDamping(0.2);
   setAngularDamping(0.2);
-  // default collisions will be mesh for physics objects
+  // default rendering and collisions will be mesh for physics objects
   // primitive-based objects do not currently support mesh collisions, however,
   // due to issues with how non-triangle meshes (i.e. wireframes) are handled in
   // @ref GenericMeshData::setMeshData
+  setRenderAssetIsPrimitive(false);
+  setCollisionAssetIsPrimitive(false);
   setUseMeshCollision(true);
 
   setBoundingBoxCollisions(false);
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -150,7 +150,7 @@ void ResourceManager::initDefaultPrimAttributes() {
       std::make_unique<Magnum::GL::Mesh>(Magnum::MeshTools::compile(*wfCube)));
 
   // build default primtive object templates corresponding to given default
-  // assets
+  // asset templates
   for (auto primAsset : primitiveAssetsTemplateLibrary_) {
     buildAndRegisterPrimPhysObjTemplate(primAsset.first);
   }
@@ -619,7 +619,8 @@ int ResourceManager::addPrimAssetTemplateToLibrary(
   return primAssetTemplateID;
 }  // addPrimAssetTemplateToLibrary
 
-int ResourceManager::loadObjectTemplate(
+// registerObjectTemplate
+int ResourceManager::registerObjectTemplate(
     PhysicsObjectAttributes::ptr objectTemplate,
     const std::string& objectTemplateHandle) {
   if (physicsObjTemplateLibrary_.count(objectTemplateHandle) != 0) {
@@ -636,6 +637,7 @@ int ResourceManager::loadObjectTemplate(
   //! Get render and collision mesh names
   // AssetInfo renderMeshinfo;
   std::string renderMeshFilename = objectTemplate->getRenderAssetHandle();
+
   bool renderMeshSuccess = loadObjectMeshDataFromFile(
       renderMeshFilename, objectTemplateHandle, "render", requiresLighting);
 
@@ -643,6 +645,7 @@ int ResourceManager::loadObjectTemplate(
   // mesh since we will use it to render
   // AssetInfo collisionMeshinfo;
   std::string collisionMeshFilename = objectTemplate->getCollisionAssetHandle();
+
   bool collisionMeshSuccess = loadObjectMeshDataFromFile(
       collisionMeshFilename, objectTemplateHandle, "collision",
       !renderMeshSuccess && requiresLighting);
@@ -685,7 +688,7 @@ int ResourceManager::loadObjectTemplate(
       objectTemplate, objectTemplateHandle, physicsFileObjTmpltLibByID_);
 
   return objectTemplateID;
-}  // loadObjectTemplate
+}  // registerObjectTemplate
 
 std::string ResourceManager::getRandomTemplateHandlePerType(
     const std::map<int, std::string>& mapOfHandles,
@@ -928,7 +931,7 @@ int ResourceManager::parseAndLoadPhysObjTemplate(
   physicsObjectAttributes->setCollisionAssetHandle(collisionMeshFilename);
 
   // 5. load the parsed file into the library
-  return loadObjectTemplate(physicsObjectAttributes, objPhysConfigFilename);
+  return registerObjectTemplate(physicsObjectAttributes, objPhysConfigFilename);
 }  // parseAndLoadPhysObjTemplate
 
 int ResourceManager::buildAndRegisterPrimPhysObjTemplate(
@@ -960,7 +963,6 @@ int ResourceManager::buildAndRegisterPrimPhysObjTemplate(
   // CollisionMesh
 
   // set margin to be 0
-
   physicsObjectAttributes->setMargin(0.0);
   // make smaller
   const Magnum::Vector3 scale(0.1, 0.1, 0.1);
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -131,19 +131,24 @@ void ResourceManager::buildMapOfPrimTypeConstructors() {
     auto attr = buildPrimitiveAttributes(elem.second);
     addPrimAssetTemplateToLibrary(attr);
   }
-}  // buildMapOfPrimTypeConstructors
 
-void ResourceManager::initDefaultPrimAttributes() {
   // instantiate a primitive importer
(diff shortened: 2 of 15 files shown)


src/esp/assets/Attributes.cpp
```diff
@@ -40,10 +40,12 @@ PhysicsObjectAttributes::PhysicsObjectAttributes(
   setInertia({0, 0, 0});
   setLinearDamping(0.2);
   setAngularDamping(0.2);
-  // default collisions will be mesh for physics objects
+  // default rendering and collisions will be mesh for physics objects
   // primitive-based objects do not currently support mesh collisions, however,
   // due to issues with how non-triangle meshes (i.e. wireframes) are handled in
   // @ref GenericMeshData::setMeshData
+  setRenderAssetIsPrimitive(false);
+  setCollisionAssetIsPrimitive(false);
```
jturner65, line 48: As this is currently configured, these flags should only be set contextually based on asset handles set by the user, and therefore should not be exposed to bindings (since user set values would be ignored anyway).

src/esp/assets/ResourceManager.cpp
```diff
@@ -1346,9 +1347,9 @@ bool ResourceManager::loadGeneralMeshData(
   const bool fileIsLoaded = resourceDict_.count(filename) > 0;
   const bool drawData = parent != nullptr && drawables != nullptr;
 
-  Cr::Containers::Pointer<Importer> importer;
-  CORRADE_INTERNAL_ASSERT_OUTPUT(
-      importer = importerManager_.loadAndInstantiate("AnySceneImporter"));
+  // Cr::Containers::Pointer<Importer> importer;
+  // CORRADE_INTERNAL_ASSERT_OUTPUT(
+  //     importer = importerManager_.loadAndInstantiate("AnySceneImporter"));
```
mosra, line 1352: Remove this?

src/esp/assets/ResourceManager.cpp
```diff
@@ -1811,41 +1813,106 @@ void ResourceManager::loadTextures(Importer& importer,
...
+  // Meta data
+  PhysicsObjectAttributes::ptr physicsObjectAttributes =
+      physicsObjTemplateLibrary_.at(objectTemplateHandle);
+
+  // if attributes are "dirty" (values have changed since last registered)
+  // re-register
+  if (physicsObjectAttributes->getIsDirty()) {
+    int ID =
+        registerObjectTemplate(physicsObjectAttributes, objectTemplateHandle);
+    if (ID == ID_UNDEFINED) {
```
mosra, line 1829:
```suggestion
    if (registerObjectTemplate(physicsObjectAttributes, objectTemplateHandle) == ID_UNDEFINED) {
```

maybe?

src/esp/assets/ResourceManager.h
```diff
@@ -366,86 +359,99 @@ class ResourceManager {
...
-   * @return The key referencing the asset in @ref physicsObjTemplateLibrary_.
+   * @return The key referencing the asset in @ref physicsObjTemplateLibrary_,
+   * or an empty string if does not exist.
    */
-  std::string getObjectTemplateHandle(const int objectTemplateID) const;
+  std::string getObjectTemplateHandle(const int objectTemplateID) const {
+    CORRADE_ASSERT(physicsTemplatesLibByID_.count(objectTemplateID) > 0,
+                   "ResourceManager::getObjectTemplateHandle : No loaded or "
+                   "primitive template with index "
+                       << objectTemplateID << " exists.",
```
mosra, line 375:
```suggestion
    CORRADE_ASSERT(physicsTemplatesLibByID_.count(objectTemplateID) > 0,
                   "ResourceManager::getObjectTemplateHandle : No loaded or "
                   "primitive template with index"
                       << objectTemplateID << "exists.",
```

FYI, Corrade's Debug (and also all Corrade assert macros) implicitly add a space between the values, so adding one more would cause double spacing. Same below.

src/esp/assets/ResourceManager.h
```diff
@@ -366,86 +359,99 @@ class ResourceManager {
...
-    return physicsObjTemplateLibrary_.at(
-        getObjectTemplateHandle(objectTemplateID));
+    std::string key = getObjectTemplateHandle(objectTemplateID);
+    CORRADE_ASSERT((physicsObjTemplateLibrary_.count(key) > 0),
+                   "ResourceManager::getPhysicsObjectAttributes : Unknown "
+                   "object template ID:"
+                       << objectTemplateID,
+                   nullptr);
+    return physicsObjTemplateLibrary_.at(key);
   }
```
mosra, line 436: I would personally de-inline all the `getPhysics*()` functions into the cpp file.

src/esp/assets/ResourceManager.h
```diff
@@ -1216,7 +1227,25 @@ class ResourceManager {
...
+      {PrimObjTypes::CONE_WF, "coneWireframe"},
+      {PrimObjTypes::CUBE_SOLID, "cubeSolid"},
+      {PrimObjTypes::CUBE_WF, "cubeWireframe"},
+      {PrimObjTypes::CYLINDER_SOLID, "cylinderSolid"},
+      {PrimObjTypes::CYLINDER_WF, "cylinderWireframe"},
+      {PrimObjTypes::ICOSPHERE_SOLID, "icosphereSolid"},
+      {PrimObjTypes::ICOSPHERE_WF, "icosphereWireframe"},
+      {PrimObjTypes::UVSPHERE_SOLID, "uvSphereSolid"},
+      {PrimObjTypes::UVSPHERE_WF, "uvSphereWireframe"},
+      {PrimObjTypes::END_PRIM_OBJ_TYPES, "NONE DEFINED"}};
```
mosra, line 1248: Since this is `const` already, I would make it `static` and put the definition into the cpp file so it doesn't pollute the header too much.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -79,9 +79,15 @@ int PhysicsManager::addObject(const std::string& configFileHandle,
...
-      makeAndAddRigidObject(nextObjectID_, physicsObjectAttributes, objectNode);
+  // verify whether necessary assets exist, and if not, instantiate them
+  bool instanceSuccess =
+      resourceManager_.instantiateAssetsOnDemand(configFileHandle);
+
+  bool objectSuccess = false;
+  if (instanceSuccess) {  // only make object if asset instantiation succeeds
+    objectSuccess = makeAndAddRigidObject(nextObjectID_,
+                                          physicsObjectAttributes, objectNode);
+  }
```
mosra, line 90:
```suggestion
  // verify whether necessary assets exist, and if not, instantiate them;
  // only make object if asset instantiation succeeds
  bool objectSuccess = resourceManager_.instantiateAssetsOnDemand(configFileHandle) &&
    makeAndAddRigidObject(nextObjectID_,
                          physicsObjectAttributes, objectNode);
  }
```

Make use of the short-circuit evaluation here? Tho could be cryptic for people not used to such workflow.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1811,41 +1813,106 @@ void ResourceManager::loadTextures(Importer& importer,
...
+      // set collision mesh data
+      const MeshMetaData& meshMetaData = getMeshMetaData(collisionAssetHandle);
+
+      int start = meshMetaData.meshIndex.first;
+      int end = meshMetaData.meshIndex.second;
+      //! Gather mesh components for meshGroup data
+      std::vector<CollisionMeshData> meshGroup;
+      for (int mesh_i = start; mesh_i <= end; ++mesh_i) {
+        GenericMeshData* gltfMeshData =
+            dynamic_cast<GenericMeshData*>(meshes_[mesh_i].get());
```
mosra, line 1891:
```suggestion
        GenericMeshData& gltfMeshData =
            dynamic_cast<GenericMeshData&>(*meshes_[mesh_i].get());
```

(I know it's not your code and it's just moved from elsewhere.) This will :boom: if the type doesn't match. With a pointer you'd get `nullptr` and then crash on that on the line below.

src/esp/assets/ResourceManager.h
jturner65: How come? I am interested in hearing your philosophy about this.   My (limited) perspective is that it is ok to inline methods that are < 10 lines, especially if they are just glorified accessors, and I am trying to remove some of the clutter from ResourceManager.cpp so I can focus on the actual functionality.  

I'd be the first to admit, though, that I am no expert, so if you have any insights to share I would really appreciate it :).

mosra: these methods aren't too long from the visual perspective, but:

- if they use STL containers in any way (which these do), there's a lot of template onions to be parsed and instantiated every time the compiler sees the definition, slowing down compile times compared to if the method is de-inlined in the cpp
- additionally, them being defined inside the class definiton makes them implicitly inline, which inflates binary size and slows down link times (especially when LTO is enabled, which is for python bindings), as the same code is pasted into every call site

that said, this project doesn't have the best compile times anyway (and can't have, due to heavy header-only dependencies like eigen and pybind), so one change like this won't make much difference, but it adds up i'd say

jturner65: Thanks for this info.  I was unaware of the impact the stl containers would have, something to research :).

PR #642, opened 2020-06-03
[Bug Fix][Viewer] Fix inconsistency between mouse move and keyboard move

bigbike:
## Motivation and Context
Play will the viewer you will find:
-) Rendering images are discontinuous when you press left mouse button to rotate + scroll to zoom in and out, and use keyboard to control the agent and the sensors;
-) Rotation on left mouse button suffers significant numerical drifts, a bit difficult to manipulate the scene.

Root cause:
Different from keyboard event, when handling mouse press and scroll event, we convert such motions directly to the render camera, not to the agent or the attached sensors.

We are suffering such artifacts for quite a long time. This is the 1st diff to fix this problem.

## Types of changes

## Checklist

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -65,7 +65,6 @@ if(BUILD_TEST)
 endif()
 
 # include source dirs
-message(STATUS "============ SOURCE ++++ " ${PROJECT_SOURCE_DIR})
 include_directories(${PROJECT_SOURCE_DIR})
 
 # Physics
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -65,6 +65,7 @@ if(BUILD_TEST)
 endif()
 
 # include source dirs
+message(STATUS "============ SOURCE ++++ " ${PROJECT_SOURCE_DIR})
 include_directories(${PROJECT_SOURCE_DIR})
 
 # Physics
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -135,11 +135,6 @@ class Viewer : public Mn::Platform::Application {
   esp::gfx::RenderCamera* renderCamera_ = nullptr;
   esp::nav::PathFinder::ptr pathfinder_;
   esp::scene::ObjectControls controls_;
-  // TODO: not needed anymore, will be depricated in next PR.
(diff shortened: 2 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ if(BUILD_TEST)
 endif()
 
 # include source dirs
+message(STATUS "============ SOURCE ++++ " ${PROJECT_SOURCE_DIR})
```
bigbike, line 68: Weird: I did not add this line here.

src/utils/viewer/viewer.cpp
```diff
@@ -586,20 +596,22 @@ void Viewer::mouseScrollEvent(MouseScrollEvent& event) {
 }
 
 void Viewer::mouseMoveEvent(MouseMoveEvent& event) {
+  if (mousePreviousPosition2D_ == Mn::Vector2i{-1})
+    mousePreviousPosition2D_ = event.position();
+
+  const Mn::Vector2i delta = event.position() - mousePreviousPosition2D_;
+  mousePreviousPosition2D_ = event.position();
```
mosra, line 603: Both SDL and GLFW app provide a convenience [relativePosition()](https://doc.magnum.graphics/magnum/classMagnum_1_1Platform_1_1GlfwApplication_1_1MouseMoveEvent.html#a4954fc274c486ebac749fedc730cf710) that is meant to be used in this case.

The `mousePreviousPosition2D_` member is then not needed anymore.

src/utils/viewer/viewer.cpp
```diff
@@ -134,8 +135,11 @@ class Viewer : public Mn::Platform::Application {
   esp::gfx::RenderCamera* renderCamera_ = nullptr;
   esp::nav::PathFinder::ptr pathfinder_;
   esp::scene::ObjectControls controls_;
+  // TODO: not needed anymore, will be depricated in next PR.
   Mn::Vector3 previousPosition_;
```
mosra, line 139: remove now and not later? i don't see it used anymore

bigbike: OK, as you wish.
I was planning to fix it in another diff.
This diff is to just fix exactly one problem and let everybody reach a consensus.

PR #643, opened 2020-06-04
Attribute managers

jturner65:
## Motivation and Context
As per comments in #639, the accessor methods for attributes quantities are becoming absurdly long.  This is because of the extant chaining for all of these quantities through Simulator and into ResourceManager - full context must be given to differentiate functionality with respect to Simulator.  Furthermore, in order to support flexible editing capabilities, ResourceManager is becoming intolerably large, filled with nearly duplicated methods to handle attributes, and only stands to get larger as new attribute-related functionality is implemented.

Enter AttributeManagers.  This PR introduces AttributesManagers as a mechanism for managing instantiation and interaction with all attributes.  No more going through Simulator->ResourceManager for every attribute-related interaction, now the user will just request the appropriate attribute manager and consume it to directly work on the attributes themselves.

This PR includes a first pass of all 4 AttributeManager implementations, and uses SceneAttributeManager and PhysicsAttributeManager within ResourceManager directly, to provide a proof of concept.  The implementations for ObjectAttributeManager and AssetAttributeManager are included for review but are not yet consumed.

Also, this PR has some small bug fixes and comment cleanup.

  

## How Has This Been Tested
c++ and python tests passed.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/Attributes.h
+++ b/src/esp/assets/Attributes.h
@@ -383,7 +383,7 @@ class AbstractPrimitiveAttributes : public AbstractAttributes {
    * - will be either 2 or 4 for primitives value checking
    * @return whether check passes
    */
-  bool isValMultOfDiv(int val, int div) {
+  bool isValueMultipleOfDivisor(int val, int div) {
     return (val >= div) && (val % div == 0);
   }
 
@@ -433,8 +433,9 @@ class CapsulePrimitiveAttributes : public AbstractPrimitiveAttributes {
    * type
    */
   virtual bool isValidTemplate() override {
-    bool wfCheck = ((getIsWireframe() && isValMultOfDiv(getNumSegments(), 4)) ||
-                    (!getIsWireframe() && getNumSegments() > 2));
+    bool wfCheck =
+        ((getIsWireframe() && isValueMultipleOfDivisor(getNumSegments(), 4)) ||
+         (!getIsWireframe() && getNumSegments() > 2));
 
     return (getCylinderRings() > 0 && getHemisphereRings() > 0 && wfCheck &&
             getHalfLength() > 0);
@@ -477,7 +478,7 @@ class ConePrimitiveAttributes : public AbstractPrimitiveAttributes {
    */
   virtual bool isValidTemplate() override {
     bool wfCheck =
-        ((getIsWireframe() && isValMultOfDiv(getNumSegments(), 4)) ||
+        ((getIsWireframe() && isValueMultipleOfDivisor(getNumSegments(), 4)) ||
          (!getIsWireframe() && getNumSegments() > 2 && getNumRings() > 0));
 
     return (getHalfLength() > 0 && wfCheck);
@@ -543,8 +544,9 @@ class CylinderPrimitiveAttributes : public AbstractPrimitiveAttributes {
    * type
    */
   virtual bool isValidTemplate() override {
-    bool wfCheck = ((getIsWireframe() && isValMultOfDiv(getNumSegments(), 4)) ||
-                    (!getIsWireframe() && getNumSegments() > 2));
+    bool wfCheck =
+        ((getIsWireframe() && isValueMultipleOfDivisor(getNumSegments(), 4)) ||
+         (!getIsWireframe() && getNumSegments() > 2));
     return getNumRings() > 0 && getHalfLength() > 0 && wfCheck;
   }
 
@@ -622,8 +624,9 @@ class UVSpherePrimitiveAttributes : public AbstractPrimitiveAttributes {
    * type
    */
   virtual bool isValidTemplate() override {
-    return ((getIsWireframe() && isValMultOfDiv(getNumSegments(), 4) &&
-             isValMultOfDiv(getNumRings(), 2)) ||
+    return ((getIsWireframe() &&
+             isValueMultipleOfDivisor(getNumSegments(), 4) &&
+             isValueMultipleOfDivisor(getNumRings(), 2)) ||
             (!getIsWireframe() && getNumRings() > 1 && getNumSegments() > 2));
   }
 
--- a/src/esp/assets/Attributes.h
+++ b/src/esp/assets/Attributes.h
@@ -20,15 +20,20 @@ namespace esp {
 namespace assets {
 
 /**
- * @brief base attributes object holding attributes shared by all
- * PhysicsXXXAttributes objects; Is abstract - should never be instanced
+ * @brief Base class for all implemented attributes.
  */
-class AbstractPhysicsAttributes : public esp::core::Configuration {
+class AbstractAttributes : public esp::core::Configuration {
  public:
-  AbstractPhysicsAttributes(const std::string& originHandle = "");
-  // forcing this class to be abstract - note still needs definition
-  // can't do this because of pybind issues, currently
-  // virtual ~AbstractPhysAttributes() = 0;
+  AbstractAttributes(const std::string& originHandle = "") : Configuration() {
+    setOriginHandle(originHandle);
+  }
+
+  /**
+   * @brief Set this attributes name/origin.  Some attributes derive their own
+   * names based on their state, such as @ref AbstractPrimitiveAttributes;  in
+   * such cases this should be overridden with NOP.
+   * @param originHandle the handle to set.
+   */
   void setOriginHandle(const std::string& originHandle) {
     setString("originHandle", originHandle);
   }
@@ -37,6 +42,22 @@ class AbstractPhysicsAttributes : public esp::core::Configuration {
     setInt("objectTemplateID", objectTemplateID);
   }
   int getObjectTemplateID() const { return getInt("objectTemplateID"); }
+
+ public:
+  ESP_SMART_POINTERS(AbstractAttributes)
+};  // class AbstractAttributes
+
+/**
+ * @brief base attributes object holding attributes shared by all
+ * PhysicsObjectAttributes and PhysicsSceneAttributes objects; Should be treated
+ * as if is abstract - should never be instanced directly
+ */
+class AbstractPhysicsAttributes : public AbstractAttributes {
+ public:
+  AbstractPhysicsAttributes(const std::string& originHandle = "");
+  // forcing this class to be abstract - note still needs definition
+  // can't do this because of pybind issues, currently
+  // virtual ~AbstractPhysAttributes() = 0;
 
   void setScale(const Magnum::Vector3& scale) { setVec3("scale", scale); }
   Magnum::Vector3 getScale() const { return getVec3("scale"); }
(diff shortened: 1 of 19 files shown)


src/esp/assets/ResourceManager.h
```diff
@@ -374,9 +344,35 @@ class ResourceManager {
...
+   * @brief Return manager for construction and access to physics world
+   * attributes.
+   */
+  managers::PhysicsAttributesManager::ptr getPhysicsAttributesManager() {
+    return physicsAttributesManager_;
+  }
+  /**
+   * @brief Return manager for construction and access to scene attributes.
+   */
+  managers::SceneAttributesManager::ptr getsceneAttributesManager() {
```
aclegg3, line 369:
```suggestion
  managers::SceneAttributesManager::ptr getSceneAttributesManager() {
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -337,9 +326,13 @@ bool ResourceManager::loadScene(
...
   // In-memory representation of scene meta data
+
   PhysicsManagerAttributes::ptr physicsManagerAttributes =
-      loadPhysicsConfig(physicsFilename);
-  physicsManagerLibrary_.emplace(physicsFilename, physicsManagerAttributes);
+      physicsAttributesManager_->createAttributesTemplate(physicsFilename,
+                                                          true);
+
+  // loadPhysicsConfig(physicsFilename);
+  // physicsManagerLibrary_.emplace(physicsFilename, physicsManagerAttributes);
```
aclegg3, line 335: remove?

src/esp/assets/ResourceManager.h
jturner65: Oh for crying out loud.  my camel case lost a hump :D

src/esp/assets/Attributes.h
```diff
@@ -273,12 +296,8 @@ class AbstractPrimitiveAttributes : public esp::core::Configuration {
...
 
   // originHandle is set internally based on attributes configuration
-  std::string getOriginHandle() const { return getString("originHandle"); }
-
-  void setAssetTemplateID(int assetTemplateID) {
-    setInt("assetTemplateID", assetTemplateID);
-  }
-  int getAssetTemplateID() const { return getInt("assetTemplateID"); }
+  // setting externally is prohibited
+  void setOriginHandle(const std::string&) {}
```
bigbike, line 300: It is empty, and it is not a virtual function. 
If user cannot set it externally, can we remove this function?

src/esp/assets/Attributes.h
```diff
@@ -347,6 +375,18 @@ class AbstractPrimitiveAttributes : public esp::core::Configuration {
...
  protected:
+  /**
+   * @brief Verifies that @ref val is larger than, and a multiple of, divisor
+   * div
+   * @param val the value to check
+   * @param div the divsior (value to verify is greater than and a multiple of)
+   * - will be either 2 or 4 for primitives value checking
+   * @return whether check passes
+   */
+  bool isValMultOfDiv(int val, int div) {
```
bigbike, line 386: `isValueMultipleOfDivisor(int val, int div)`

I think it is easier to parse. Up to you.

src/esp/assets/ResourceManager.h
```diff
@@ -35,6 +35,11 @@
 #include "esp/gfx/configure.h"
 #include "esp/scene/SceneNode.h"
 
+// #include "managers/AssetAttributesManager.h"
```
bigbike, line 38: If you do not need these headers, would you consider removing them?

src/esp/assets/ResourceManager.h
```diff
@@ -1299,6 +1295,24 @@ class ResourceManager {
...
+  //  */
+  // managers::AssetAttributesManager::ptr assetAttributesManager_ = nullptr;
+  // /**
+  //  * @brief Manages all construction and access to object attributes.
+  //  */
+  // managers::ObjectAttributesManager::ptr objectAttributesManager_ = nullptr;
+  /**
+   * @brief Manages all construction and access to physics world attributes.
+   */
+  managers::PhysicsAttributesManager::ptr physicsAttributesManager_ = nullptr;
```
bigbike, line 1310: Is there any case that we must share the pointer?
If not, I strongly recommend to use unique pointer, which should be our default choice for smart pointers.

src/esp/assets/ResourceManager.h
```diff
@@ -1299,6 +1295,24 @@ class ResourceManager {
...
+  /**
+   * @brief Manages all construction and access to scene attributes.
+   */
+  managers::SceneAttributesManager::ptr sceneAttributesManager_ = nullptr;
```
bigbike, line 1314: same as above

src/esp/assets/managers/AssetAttributesManager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // build default AbstractPrimitiveAttributes objects
+  for (const std::pair<PrimObjTypes, const char*>& elem : PrimitiveNames3DMap) {
+    if (elem.first == PrimObjTypes::END_PRIM_OBJ_TYPES) {
+      continue;
+    }
+    createAttributesTemplate(elem.second, true);
+  }
+}  // buildMapOfPrimTypeConstructors
+
+std::shared_ptr<AbstractPrimitiveAttributes>
```
bigbike, line 128: Same here.
Think about if the returned pointers must be shared.

bigbike:
Can we also add Eric as a reviewer so that he can familiarize him with the code?

src/esp/assets/Attributes.h
jturner65: This is correct.  It will become virtual once the appropriate pybinds are developed.  For pybind to support virtual, seems you have to get "trampoline" classes implemented in c++ and that was beyond the scope of this PR.

src/esp/assets/ResourceManager.h
jturner65: These are for the next PR.  The code for the asset and object attribute managers is done, these are just stubs for them.

src/esp/assets/ResourceManager.h
jturner65: yes, we share it by giving it to the user, so they can access it directly.  that is the purpose of the PR, instead of having them go through Resource Manager, they will get the object and manipulate the attributes themselves.

erikwijmans: Do they get ownership?  I am not sure if this has any OpenGL resources, but if it does, then the user shouldn't get ownership (we assume all ownership of OpenGL).

jturner65: These are just attributes managers - they only manage the objects that describe the configurations of various constructs, like objects, primitive assets, scenes and the physics world.  They do not actually touch on assets, that is still the sole purview of ResourceManager.

jturner65:
> 
> 
> Can we also add Eric as a reviewer so that he can familiarize him with the code?

I can't find him on the list.  I'll contact him directly.

src/esp/assets/Attributes.h
jturner65: To elaborate more on this and other similar issues in the Attributes : 

All of the various attributes classes have the primary purpose of providing the user with easily accessible editing capabilities for the objects they describe, and of course the key mechanism for this access is pybind.  The integration of pybind requires, in some cases, to build c++ "trampoline" classes that act as adapters between the underlying c++ code and pybind's python interface/code gen.  These classes are required by pybind, for example, to implement proper support of virtual functions.  

The inheritance hierarchy of the attributes is intended to minimize code duplication and provide code expectations (much like an interface) to future attributes class development.  As an aside, ideally, all the attributes classes that are named "AbstractXXX" would actually -be- abstract, but since they inherit from a concrete class (configuration), it is not immediately obvious how to properly accomplish this in pybind (I am still researching how to do it, but I have not yet figured out the mechanism).  I understand that this kind of inheritance hierarchy is uncommon, but since configuration is in an external library/used many places elsewhere, I am stuck with using it as an instantiable class.  As far as the entire "Attributes" subtree of functionality, however, the current hierarchy satisfies Liskov substitution. 

However, pybind's seemingly arcane mechanisms to accomplish this inheritance hierarchy has put me in the undesirable position of having to slightly modify the c++ in suboptimal ways to accommodate pybind itself, at least temporarily.  Thankfully, these modifications have been slight, but the code above is an example of one of them.

The next PR will introduce the necessary trampoline classes required to support the virtual functions in the various attributes classes.  The purpose of this PR was primarily to introduce the concept of the AttributesManagers, and as you said, it is a pretty large PR as it is, and so I felt that making suboptimal (but temporary) coding decisions as above was preferable to substantially increasing the size of the PR itself with the trampolines :).

src/esp/assets/ResourceManager.h
erikwijmans: You may need to bind the subclass into python for things to work correctly.  We use virtual methods here: https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/bindings/SceneBindings.cpp#L143

src/esp/assets/Attributes.h
jturner65: I'll check that out, thanks.

src/esp/assets/Attributes.h
mosra: This could also be a free function since it has no relation to the class (or at least a `static` one).

src/esp/assets/managers/AttributesManagerBase.h
```diff
@@ -0,0 +1,334 @@
...
+                   [](unsigned char c) { return std::tolower(c); });
+    bool found = (std::string::npos != key.find(strToLookFor));
+    if (found == contains) {
+      // if found and searching for contains, or not found and searching for not
+      // contains
+      res.push_back(iter->second);
+    }
+  }
+  return res;
+}  // getTemplateHandlesBySubStringPerType
```
mosra, line 329: I remember we were discussing putting the function definitions into the cpp file, since it's quite a huge bulk of code and the set of types is limited and known in advance.

Did that not work, after all? Or what's the reason it's all in the header?

src/esp/assets/CMakeLists.txt
```diff
@@ -14,8 +14,19 @@ set(assets_SOURCES
   MeshMetaData.h
   Mp3dInstanceMeshData.cpp
   Mp3dInstanceMeshData.h
+  managers/AttributesManagerBase.h
+  # managers/AssetAttributesManager.h
+  # managers/AssetAttributesManager.cpp
+  # managers/ObjectAttributesManager.h
+  # managers/ObjectAttributesManager.cpp
```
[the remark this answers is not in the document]
jturner65: This is coming in next PR.  Had implemented them all but in the interest of keeping this PR somewhat reasonable in size, commented out the not-yet-used file refs.

src/esp/assets/managers/AttributesManagerBase.h
jturner65: These were put here to avoid future linker errors.  Otherwise I would have to put all possible typename (attributes types) declarations at the bottom, and this would then break when more attributes were added to the project.

PR #644, opened 2020-06-05
Refactor init to not call reconfigure

erikwijmans:
## Motivation and Context

I went to do the habitat-api version of #620 and realized that our current `__init__` method doesn't allow for inheritance as it calls `reconfigure`.  This PR refactors `__init__` to not call reconfigure (which is better design regardless).

## How Has This Been Tested

With the tests!

## Types of changes


habitat_sim/simulator.py
```diff
@@ -167,34 +185,16 @@ def _config_pathfinder(self, config: Configuration):
...
-
-        # NB: Configure backend last as this gives more time for python's GC
-        # to delete any previous instances of the simulator
-        # TODO: can't do the above, sorry -- the Agent constructor needs access
-        # to self._sim.get_active_scene_graph()
+    def __set_from_config(self, config: Configuration):
         self._config_backend(config)
         self._config_agents(config)
         self._config_pathfinder(config)
-        self.frustum_culling = config.sim_cfg.frustum_culling
```
aclegg3, line 197: Does this get set anywhere now? Looks like the config setting is used directly in reconfigure to decide whether or not to split the mesh for the semantic SceneGraph in C++ [here](https://github.com/facebookresearch/habitat-sim/blob/e31c39fbf09655c434d32500395b5c1e46db7120/src/esp/sim/Simulator.cpp#L184). However, it seems that we don't consume the config setting for RGBD culling now.

erikwijmans: I looked over the code base and nothing was using that value, may have just been there from testing and didn't get properly removed before landing.

aclegg3: I see. 👍  Maybe this is now tangential to the current change, but should this have been used [here](https://github.com/facebookresearch/habitat-sim/blob/d28b051eb72196365a2d03630f20b7d86599a268/habitat_sim/simulator.py#L405)? @bigbike

bigbike: I initially introduced in this PR:
https://github.com/facebookresearch/habitat-sim/commit/9a1fcb2c0b80d2d65a7b4f78cb476d719d71b34c

Yes, it is used in the place as you pointed out. @aclegg3 

I did it because I need to enable the frustum culling when the reconfigure() is called. It is not something just for testing and is left there accidentally.

If you removed it, where would it (enabling the FC or not) get reset?

erikwijmans: I miss-remember why I removed it.  I left it there accidentally in #620.  We have a property from the cpp bindings called `frustum_culling`, so `self.frustum_culling` just calls that property after #620.  I should have removed that call in that PR and moved it to cpp (makes sense to set that in the cpp `reconfigure` since its part of the backend/cpp config).

erikwijmans: Scratch that, I am just miss-remembering my line of thought entirely here.  I am just going to leave this the way it was.  Maybe doing in cpp would be better, but not worth changing.

PR #645, opened 2020-06-07
Windowless context updates

erikwijmans:
## Motivation and Context

Add some additional logging statements to aid in EGL debugging.  This will let us give us additional information about the EGL devices found to help figure out why none of them are valid CUDA devices (in the case that no EGL device can be found for a given CUDA device).

Also removes the `isNvidiaGpuReadable` check.  This was there because it was there in House3D's windowless context code and it seemed like it couldn't hurt (i.e. the CUDA device check is strictly stronger than that check). However, this doesn't play nice with Docker.  Docker started with say, `NV_GPU=5`, results in only 1 GPU (GPU 5) being visible in the docker container, however, the device file will still be `/dev/nvidia5`, so this check will fail when we go to check if `/dev/nvidia0` is readable.  The only times I have seen or head of this check failing is this, so I think it is safe to remove it and just rely on the CUDA device ID check.

### Update!

After chatting with @mosra and landing a PR in magnum, we have moved all this functionality over to magnum.  There is nicer debug logging and this also gives us the ability to support a software rendering if we want.

## How Has This Been Tested

It logs!

## Types of changes


src/esp/gfx/WindowlessContext.cpp
```diff
@@ -60,16 +60,6 @@ const int MAX_DEVICES = 128;
...
 
-bool isNvidiaGpuReadable(int device) {
-  const std::string dev = "/dev/nvidia" + std::to_string(device);
-  const int retval = open(dev.c_str(), O_RDONLY);
-  if (retval == -1) {
-    return false;
-  }
-  close(retval);
-  return true;
-}
```
mosra, line 71: Great, with this removed the implementation moves closer to what I have in `WindowlessEglApplication` :D At this point I think the vanilla one is only missing a `--magnum-cuda-device N` counterpart to `--magnum-device N` and it could fully replace this code.

erikwijmans:
@mosra is there any difference besides CUDA devices?  

The only thing is see is differing `configAttribs`/`attribs`, but I don't think that matters because that's for the EGL frame buffer but we don't use that.

mosra:
It should be only the CUDA devices, yes. The config attribs are needed for SwiftShader I think (it crashes without some pbuffer attached), but that's not relevant here.

erikwijmans:
Yeah, if you would like to integrate CUDA device selection in to Magnum's EGL context, I am happy to switch to that :)

src/esp/gfx/WindowlessContext.cpp
```diff
@@ -62,120 +62,33 @@ const int MAX_DEVICES = 128;
```
mosra, line 62: Hmm, I think the `CHECK_EGL_ERROR`, `MAX_DEVICES`, the `<fcntl.h>` includes and the `// Based on code from:` comments aren't needed anymore either :)

src/esp/gfx/WindowlessContext.cpp
```diff
@@ -213,8 +117,8 @@ struct ESPGLXContext : ESPContext {
   int gpuDevice() const { return 0; }
 
  private:
-  Mn::Platform::WindowlessGlxContext glxCtx_;
   Mn::Platform::GLContext magnumGlContext_;
+  Mn::Platform::WindowlessGlxContext glxCtx_;
```
mosra, line 121: And I think this could be collapsed together with the CGL variant for Apple (use `WindowlessGLContext` for both) ... actually maybe even together with EGL, with the same code for all platforms, only have the 

```cpp
    CHECK_EQ(device, 0)
        << "glX context does not support multiple GPUs. Please compile with "
           "BUILD_GUI_VIEWERS=0 for multi-gpu support via EGL";
    CHECK(std::getenv("DISPLAY") != nullptr)
        << "DISPLAY not detected. For headless systems, compile with "
           "--headless for EGL support";
```

checks wrapped in `#ifndef ESP_BUILD_EGL_SUPPORT`? Ultimately, the `virtual`s shouldn't be needed either, I think those would only make sense if there was runtime switching between GLX and EGL -- which there isn't. 

(I'm pretty certain magnum would fail to link if you would attempt to link to both GLX and EGL, although runtime switching between GLX and EGL is still my roadmap. I secretly hope that GLX becomes obsolete before I get to implement that, haha.)

PR #653, opened 2020-06-15
[object picking] Step 1: remove object id from GenericDrawable

bigbike:
## Motivation and Context
I am working on a new functionality in viewer -- object picking. This is a rather useful function that we would like to have for quite a long time.

This is 1st diff that is to address this problem.
We did not use the object id anywhere in our simulator, and it confuses the developers and users. The solution is to deprecate it, and introduce a drawableID in the Drawable class (parent class of GenericaDrawable) to address this problem (in the coming diff). So each drawable (NOT just the GenericDrawable) would have a unique id to identify itself.

## Types of changes

## Checklist


dhruvbatra:
CC: @dexter1691 -- you may want to chat with @bigbike about your current functionality in Sokoban so appropriate parts can be sent as PRs.

PR #654, opened 2020-06-16
[object picking] Step 2: Add a drawable id to each Drawable object

bigbike:
## Motivation and Context

This diff is to introduce a unique ID for each Drawable object in the simulator that can be used to identify itself, e.g., in the object picking in the editor (viewer).

## Types of changes

## Checklist


src/esp/gfx/DrawableGroup.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "DrawableGroup.h"
+
+namespace esp {
+namespace gfx {
+
+bool DrawableGroup::hasDrawable(uint64_t id) {
+  return (idToDrawable_.find(id) != idToDrawable_.end());
+}
+
+Drawable* DrawableGroup::getDrawable(uint64_t id) {
```
jturner65, line 14:
```suggestion
Drawable const* DrawableGroup::getDrawable(uint64_t id) const {
```
How about this?  So that the address being pointed to isn't changed by the consumer accidentally, and the getter itself can be const too, right?

bigbike: Once we get the drawable object, we will not just simply query its data, but make changes to the drawable object. For example, in the coming diff, I will set this drawable as "selected" (a boolean member variable to be introduced to Drawable class) if it is picked by the mouse.

jturner65: Does that mean you may wish for the return value to point to another location in memory (i.e. like with a new assignment)?  <type> const* will just make sure the pointer always points to the same location, but what it is pointing at can change.  :)

jturner65: -typename- const * just means the memory address being pointed to by the pointer won't change.  The data at that memory can change.  Was thinking that we might not want folks to be able to reassign things externally

bigbike: Oh, now I understand what you really would like to do, which is a good point.

However, `typename const*` is the same as `const typename*`, and that means the object (which has type `typename`) is constant and you cannot change the value within. Certainly this is not the case we would like to have here.

So let me guess, you were trying to say, `typename* const`, which means a constant pointer to a (mutable) object. But I do not think we can use `typename* const` to the return value of a function because the `const` is totally redundant in this case. It is meaningless to add a type qualifier (const) on the return type itself (a pointer). It is ONLY meaningful to add `const` to the object the raw pointer is pointing to.

So the correct way to do what you really suggested is in the viewer:
we will have something like:
`esp::gfx::Drawable *const object = ...`

jturner65: Thanks for correcting me, Today has been a long day :).  I was thinking of the case where folks call things with auto&, and thinking that perhaps that could be dangerous.

src/esp/gfx/Drawable.cpp
```diff
@@ -10,11 +10,25 @@
...
+      node_(node),
+      mesh_(mesh),
+      drawableId_(drawableIdCounter++) {
+  if (group) {
+    group->idToDrawable_.insert({drawableId_, this});
+  }
+}
+
+Drawable::~Drawable() {
+  DrawableGroup* group = drawables();
```
bigbike, line 27: I found a serious problem here.
In the dtor, It seems the usage of function `drawables()` has limitations. It cannot be used in this destructor.
The reason is that this group is deconstructed BEFORE this drawable is deconstructed. So when you call the `drawables()`, it would trigger the `CORRADE_ASSERT` within the `drawable()`

Thoughts?

bigbike: All right. I updated the code.
Current solution is:
When DrawableGroup get deconstructed earlier than the Drawables stored within, it will send a message to inform all the drawables that "no more query or operation on this group any more".

src/esp/gfx/DrawableGroup.cpp
```diff
@@ -0,0 +1,33 @@
...
+namespace esp {
+namespace gfx {
+
+class Drawable;
+
+DrawableGroup::~DrawableGroup() {
+  // has to inform any currently existing drawables, do not query anything
+  // regarding this group
+  for (int iDrawable = 0; iDrawable < this->size(); ++iDrawable) {
+    static_cast<Drawable&>((*this)[iDrawable]).groupExists_ = false;
```
bigbike, line 16: I do not like this solution actually.
Look at this static_cast here, it means the DrawableGroup can only work with our Drawable class, not any magnum original drawables. 
(I think it is OK, just like our scene graph will only work with our own scene node. Do you agree?)

bigbike:
Dear reviewers,

This diff is now ready to be reviewed. 
Thanks!

src/esp/gfx/DrawableGroup.h
```diff
@@ -3,22 +3,57 @@
 // LICENSE file in the root directory of this source tree.
 
 #pragma once
-
+#include <Magnum/SceneGraph/Drawable.h>
 #include <Magnum/SceneGraph/FeatureGroup.h>
+#include <Magnum/SceneGraph/SceneGraph.h>
+#include <functional>
```
mosra, line 9: unneeded? this include is [one of the heavier ones](https://artificial-mind.net/projects/compile-health/), so the less is used the better :)

src/esp/gfx/DrawableGroup.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+DrawableGroup& DrawableGroup::unregisterDrawable(Drawable& drawable) {
+  uint64_t id = drawable.getDrawableId();
+  // if this id is not in this group, return directly
+  if (!hasDrawable(id)) {
+    return *this;
+  }
+
+  idToDrawable_.erase(id);
+  drawable.attachedToGroup_ = false;
```
mosra, line 61: this is unnecessarily doing two map searches, could be just one:

```suggestion
  // if this id is not in this group, erase() erases nothing
  if(idToDrawable_.erase(id) == 0)
    drawable.attachedToGroup_ = false;
```

src/esp/gfx/DrawableGroup.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+DrawableGroup& DrawableGroup::registerDrawable(Drawable& drawable) {
+  uint64_t id = drawable.getDrawableId();
+  // it is already registered, return directly
+  if (hasDrawable(id)) {
+    return *this;
+  }
+
+  idToDrawable_.insert({id, &drawable});
+  drawable.attachedToGroup_ = true;
```
mosra, line 50: same, unnecessary two searches, could be just one. also, `emplace():

```suggestion
  // if it is already registered, emplace() does nothing
  if(idToDrawable_.emplace(id, &drawable).second) 
    drawable.attachedToGroup_ = true;
```

src/esp/gfx/DrawableGroup.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+bool DrawableGroup::hasDrawable(uint64_t id) const {
+  return (idToDrawable_.find(id) != idToDrawable_.end());
+}
+
+Drawable* DrawableGroup::getDrawable(uint64_t id) const {
+  if (hasDrawable(id)) {
+    return idToDrawable_.at(id);
+  }
```
mosra, line 37: Unnecessary two map searches, could be one. I'm lazy to write another snippet -- use `find()` ;)

src/esp/gfx/Drawable.h
```diff
@@ -53,14 +59,25 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
...
   scene::SceneNode& node_;
   Magnum::GL::Mesh& mesh_;
+
+  /**
+   * Why a friend class?
+   * because DrawableGroup will have to update the state, attachedToGroup
+   * directly, and ONLY this class can do it
+   */
+  friend class DrawableGroup;
+  bool attachedToGroup_ = false;
```
mosra, line 80: hm actually isn't this state already present in magnum's feature class? there's [group()](https://doc.magnum.graphics/magnum/classMagnum_1_1SceneGraph_1_1AbstractGroupedFeature.html#a670e9a5ed8e81ed9ea5b819e3af01ee6) (or `drawable()`), which returns a pointer -- if it's null, the feature is not in any group

src/esp/gfx/DrawableGroup.cpp
```diff
@@ -0,0 +1,66 @@
...
+  unregisterDrawable(drawable);
+  this->Magnum::SceneGraph::DrawableGroup3D::remove(drawable);
+  return *this;
+}
+
+DrawableGroup::~DrawableGroup() {
+  // has to unregister any currently existing drawables
+  for (int iDrawable = 0; iDrawable < this->size(); ++iDrawable) {
+    unregisterDrawable(static_cast<Drawable&>((*this)[iDrawable]));
+  }
```
mosra, line 27: With `attachedToGroup_` gone as I suggested above, this loop is no longer needed -- it would only empty the unordered_map, which we don't really care about since we're destructing anyway

bigbike: We actually care. I met the CI failure yesterday, due to this. The DrawableGroup (NOT your magnum lib) is deconstructed earlier than our Drawable (not your drawable), and thus I cannot call the `drawables()` function in the destructor of the Drawable! The corrade_assert fails!

In the dtor of the Drawable, I MUST call the drawables() to get the drawableGroup this drawable is currently in. So it can remove itself from that idToDrawable_ lookup table. Because there is a chance, we just delete a drawable during the runtime.

Simply put: I must handle both cases: DrawableGroup is deconstructed first, and a within Drawable is deconstructed first.

src/esp/gfx/Drawable.h
bigbike: We need it. we need to make sure it is attaching to our DrawableGroup (sub-class of a magnum drawable group), not the magnum base drawable group.

src/esp/gfx/DrawableGroup.cpp
bigbike: This is an unordered_map not a map. So the search depends on the hash of the key, which is super fast. 

But I will change anyway.

src/esp/gfx/DrawableGroup.cpp
bigbike: This is great. Thank you!

src/esp/gfx/DrawableGroup.cpp
mosra: In theory it is fast, in practice `std::unordered_map` is [one of the worst performing implementations of a hashmap](https://tessil.github.io/2016/08/29/benchmark-hopscotch-map.html), due to various unfortunate design decisions mandated by the standard .. and in some cases the constant overhead of the O(1) lookup can be larger than a O(log n) search in a map.

doesn't matter if map or unordered_map, the change will make it 2x faster ;)

src/esp/gfx/DrawableGroup.cpp
erikwijmans: This should like a bug we need to fix in our `DrawableGroup` then.  Looking at our code:

https://github.com/facebookresearch/habitat-sim/blob/fcb1a595b6c151b59a5c537738cac354cc7e91f8/src/esp/gfx/Drawable.cpp#L19-L25

If we made that
```cpp
DrawableGroup* Drawable::drawables() {
  auto drawables = Magnum::SceneGraph::Drawable3D::drawables();
  if (! drawables)
    return nullptr;
  CORRADE_ASSERT(
      dynamic_cast<DrawableGroup*>(drawables),
      "Drawable must only be used with esp::gfx::DrawableGroup!", {});
  return static_cast<DrawableGroup*>(drawables);
}
```

Then this would correctly return a nullptr, alleviating this issue.

bigbike: This is a smart idea! Will do it in this way.

mosra: Ok, then if this assert is in place, `attachedToGroup_` ultimately *is* redundant, right?

bigbike: @mosra: Yes, correct. It was removed already.

bigbike:
added a test. Ready to be reviewed again.
Thanks.

PR #656, opened 2020-06-17
Attributes managers remove and tests

jturner65:
## Motivation and Context
This continues the implementation of the Attributes Managers by introducing the capability of removing existing attributes, as well as introducing tests for creating, modifying, and removing attributes from PhysicsAttributesManager, SceneAttributesManager and ObjectAttributesManager.

The internal structure of how attributes themselves would be managed has also been modified.  Now, the python user will always be working with a copy of an attributes.  The user will query the manager for an attributes, modify it to taste, and register it with a desired name.  An attributes can be removed by either ID or handle.

The next PR will introduce python-based tutorial and python bindings for these objects.

## How Has This Been Tested
CPP and pytest passed without complaint.

## Types of changes

## Checklist


src/utils/viewer/viewer.cpp
```diff
@@ -321,8 +332,8 @@ void Viewer::addObject(const std::string& configFile) {
   }
 
   Mn::Matrix4 T =
-      agentBodyNode_
-          ->MagnumObject::transformationMatrix();  // Relative to agent bodynode
+      agentBodyNode_->MagnumObject::transformationMatrix();  // Relative to
+                                                             // agent bodynode
```
aclegg3, line 336: Looks like the linter made some unnecessary style changes several times in this file.

src/esp/assets/ResourceManager.h
```diff
@@ -431,12 +396,10 @@ class ResourceManager {
    * @brief Instantiate, or reinstatiate, PhysicsManager defined by passed
    * attributes
    * @param physicsManager The currently defined @ref physics::PhysicsManager.
-   * Will be reseted to the configured physics implementation.
-   * @param physicsFilename The physics configuration file from which to
-   * re-instatiate the @ref physics::PhysicsManager and parse object templates
-   * for the
-   * @ref physicsObjTemplateLibrary_. Defaults to the file location @ref
-   * ESP_DEFAULT_PHYS_SCENE_CONFIG set by cmake.*/
+   * Will be reseated to the configured physics implementation.
```
eundersander, line 399: "reseated"? typo?

jturner65: Actually it is intended to be "reseat".  Apparently it has a meaning with regard to smart pointers.   Surprised me too :).

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r35-take-a-shared_ptrwidget-parameter-to-express-that-a-function-might-reseat-the-shared-pointer

src/esp/assets/managers/AssetAttributesManager.h
```diff
@@ -163,46 +149,321 @@ class AssetAttributesManager
...
-    if (primTypeConstructorMap_.count(primTypeName) == 0) {
+  AbstractPrimitiveAttributes::ptr buildPrimAttributes(
+      const std::string& primClassName) {
+    if (primTypeConstructorMap_.count(primClassName) == 0) {
       LOG(ERROR) << "AssetAttributesManager::buildPrimAttributes : No "
-                    "primitive of type"
-                 << primTypeName << "exists.  Aborting.";
+                    "primitive class"
+                 << primClassName << "exists in Magnum::Primitives.  Aborting.";
       return nullptr;
```
eundersander, line 441: Maybe off-topic for this PR, but I want to ask about this error handling. Would we consider this a "fatal" error? If yes, then shouldn't we exit? If no, it doesn't seem like returning nullptr will be safe--I saw some calling code that seems to assume the return value will be non-null.

src/esp/sim/Simulator.cpp
```diff
@@ -148,9 +148,21 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
...
     bool loadSuccess = false;
     if (config_.enablePhysics) {
-      loadSuccess = resourceManager_->loadScene(
-          sceneInfo, physicsManager_, &rootNode, &drawables,
-          cfg.sceneLightSetup, cfg.physicsConfigFile);
+      // use physics world attributes manager to get physics manager attributes
+      // described by config file
+      auto physicsManagerAttributes =
+          resourceManager_->getPhysicsAttributesManager()
+              ->createAttributesTemplate(cfg.physicsConfigFile, true);
```
eundersander, line 155: It looks like, deep inside `createAttributesTemplate` (specifically, inside `addTemplateToLibrary`), we construct a copy of the new attributes object. Can you explain what's going on here and why we need to make that copy?

src/esp/assets/managers/ObjectAttributesManager.cpp
```diff
@@ -21,60 +21,85 @@ namespace esp {
...
-    // if attributesTemplateHandle != some existing primitive attributes, then
+  } else if (Corrade::Utility::Directory::exists(attributesTemplateHandle)) {
+    // if attributesTemplateHandle == some existing file then
     // assume this is a file-based object template we are building.
     return createFileBasedAttributesTemplate(attributesTemplateHandle,
                                              registerTemplate);
+  } else {
+    // if neither of these is true, then build an empty template and assign the
+    // passed handle to its origin handle and its render asset handle
+    return createEmptyAttributesTemplate(attributesTemplateHandle,
```
eundersander, line 40: Can you talk about the use cases for this function? Will the user ever intentionally create an empty template, or is that strictly an error-condition fallback? 

This behavior of looking for a file and then falling back to creating an empty template... it seems confusing and error-prone. Doesn't the user want an error/warning if the file she was expecting is missing?

src/esp/assets/managers/AssetAttributesManager.h
jturner65: There might be some situations that are not checking for nullptr, but all callers of this should be checking. I'll investigate.  We don't want to kill everything if this fails, though, since it only means they got the class name wrong.

src/esp/assets/managers/ObjectAttributesManager.cpp
jturner65: Current tests rely on the ability to create a blank object attributes template and then configure it as desired.  Also, it is conceivable that a user would wish to have a blank attributes they could consume to create objects manually, without having to configure physics/scenes to load object json.  Perhaps feedback to the effect of "<type> attributes created" might be helpful, but in this PR we are assuming the user knows what they are doing.

This will be clearly described in tutorials consuming these functions, of course. :)

src/esp/sim/Simulator.cpp
jturner65: yes, all attributes are passed to the user as copies of what is kept in the library. No changes to the library occur unless attributes are specifically registered with the appropriate manager.  There are numerous reasons for this, such as allowing the user to reset the simulator while retaining copies of desired attributes, or allowing asset attribute edits to be validated before they are written to the library.

jturner65: Note : in this PR not all attributes getters send copies.  This is to support older functionality before this design decision was made to give users copies.  The full change-over to this mechanism will be completed in the next few PRs.

src/esp/assets/managers/AssetAttributesManager.h
jturner65: Found a few callers who were not checking.  Thanks for finding this.

src/esp/assets/managers/ObjectAttributesManager.cpp
jturner65: Feedback added.

eundersander: >in this PR we are assuming the user knows what they are doing
I'm really just asking for consistency with the error-handling in the rest of this file. I like the LOG(INFO) you added because it's better than silent failure. But it would be more consistent with other code in this file to issue a LOG(ERROR) in the case of a missing file. And I understand the current interface makes it difficult to distinguish the error-case from the making-a-blank-template case, so I suspect the interface needs to change.

jturner65: Originally I had an error trap here as you describe, but then I realized i needed to be able to support the creation of a blank template - passing an unknown template handle could denote the user wanting a blank one with the given name.  I considered setting some kind of sentinel value for the name to denote that it was intended to be a new template (like "newTemplate") but this seemed suboptimal because it too could fail if it was misspelled or unknown, and it would be too easy for the user to forget to rename the template, causing later template creations to overwrite earlier ones.  

Also, as far as the user goes, I do not think there's going to be an expectation bias toward any type of attributes to be created here.  The file based attributes in a particular scene are usually specified in the scene json, and will be created when the scene loads, so the user won't even access this.  She is just as likely (in fact, i would expect more likely) to be wanting to manually create a prim-based object attributes here as a file based one, and there are certainly use cases for blank attributes as well.

This is why I chose the method I did, the least of evils as I saw it, but i am certainly interested in revisiting this at a later date once folks have consumed it and can give some feedback.

PR #659, opened 2020-06-24
Attributes managers object sim tests

jturner65:
## Motivation and Context
This PR presents the final C++ code to support and test the attributes managers implementations.  Included in this PR are : 
1. Added testing to SimTest to test the creation of objects based on primitive assets.
2. Removed simulator access to individual attribute templates for most cases.  All attributes should be accessed via attribute managers.
3. Most attributes access has been changed to provide a copy of the attributes being stored by the manager, instead of the actual attributes.  The goal is to have all such access be via copies, but this will be a breaking change, and will be introduced with the next PR, which will also introduce a tutorial on the proper usage of the attributes.
4. Cleaned up rigid object and rigid scene attributes access, and made object creation more efficient.
5. Overrides of virtual methods are virtual by default, and so do not need to be specified as such (it is recommended not to do so, in fact) so these keywords have been removed where appropriate.

The next, and final PR, will introduce python tests and tutorials about consuming the attributes managers and creating primitive-asset-based objects.  

## How Has This Been Tested
All cpp and python tests pass, including added test for primitive asset-based object creation.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/Attributes.h
+++ b/src/esp/assets/Attributes.h
@@ -190,10 +190,10 @@ class PhysicsObjectAttributes : public AbstractPhysicsAttributes {
   Magnum::Vector3 getCOM() const { return getVec3("COM"); }
 
   // whether com is provided or not
-  void setCOMIsProvided(bool COMIsProvided) {
-    setBool("COMIsProvided", COMIsProvided);
+  void setComputeCOMFromShape(bool computeCOMFromShape) {
+    setBool("computeCOMFromShape", computeCOMFromShape);
   }
-  bool getCOMIsProvided() const { return getBool("COMIsProvided"); }
+  bool getComputeCOMFromShape() const { return getBool("computeCOMFromShape"); }
 
   // collision shape inflation margin
   void setMargin(double margin) { setDouble("margin", margin); }
--- a/src/esp/assets/Attributes.h
+++ b/src/esp/assets/Attributes.h
@@ -189,6 +189,12 @@ class PhysicsObjectAttributes : public AbstractPhysicsAttributes {
   void setCOM(const Magnum::Vector3& com) { setVec3("COM", com); }
   Magnum::Vector3 getCOM() const { return getVec3("COM"); }
 
+  // whether com is provided or not
+  void setCOMIsProvided(bool COMIsProvided) {
+    setBool("COMIsProvided", COMIsProvided);
+  }
+  bool getCOMIsProvided() const { return getBool("COMIsProvided"); }
+
   // collision shape inflation margin
   void setMargin(double margin) { setDouble("margin", margin); }
   double getMargin() const { return getDouble("margin"); }
--- a/src/esp/assets/Attributes.h
+++ b/src/esp/assets/Attributes.h
@@ -30,6 +30,8 @@ class AbstractAttributes : public esp::core::Configuration {
     setAttributesClassKey(attributesClassKey);
     setOriginHandle(originHandle);
   }
+
+  virtual ~AbstractAttributes() = default;
   /**
    * @brief Get this attributes' class.  Should only be set from constructor.
    * Used as key in constructor function pointer maps in AttributesManagers.
@@ -42,7 +44,7 @@ class AbstractAttributes : public esp::core::Configuration {
    * such cases this should be overridden with NOP.
    * @param originHandle the handle to set.
    */
-  void setOriginHandle(const std::string& originHandle) {
+  virtual void setOriginHandle(const std::string& originHandle) {
     setString("originHandle", originHandle);
   }
   std::string getOriginHandle() const { return getString("originHandle"); }
@@ -51,6 +53,18 @@ class AbstractAttributes : public esp::core::Configuration {
     setInt("objectTemplateID", objectTemplateID);
   }
   int getObjectTemplateID() const { return getInt("objectTemplateID"); }
+
+  /**
+   * @brief Returns configuration to be used with PrimitiveImporter to
+   * instantiate Primitives.  Names in getter/setters chosen to match parameter
+   * name expectations in PrimitiveImporter.
+   *
+   * @return a reference to the underlying configuration group for this
+   * attributes object
+   */
+  const Corrade::Utility::ConfigurationGroup& getConfigGroup() const {
+    return cfg;
+  }
 
  protected:
   /**
@@ -76,9 +90,8 @@ class AbstractPhysicsAttributes : public AbstractAttributes {
  public:
   AbstractPhysicsAttributes(const std::string& classKey,
                             const std::string& originHandle);
-  // forcing this class to be abstract - note still needs definition
-  // can't do this because of pybind issues, currently
-  // virtual ~AbstractPhysAttributes() = 0;
+
+  virtual ~AbstractPhysicsAttributes() = default;
 
   void setScale(const Magnum::Vector3& scale) { setVec3("scale", scale); }
   Magnum::Vector3 getScale() const { return getVec3("scale"); }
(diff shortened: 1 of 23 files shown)


src/esp/physics/RigidBase.h
```diff
@@ -89,14 +89,20 @@ class RigidBase : public Magnum::SceneGraph::AbstractFeature3D {
...
-      const assets::ResourceManager& resMgr,
-      const assets::AbstractPhysicsAttributes::ptr physicsAttributes) = 0;
+  virtual bool initialize(const assets::ResourceManager& resMgr,
+                          const std::string& handle) = 0;
+
+  /**
+   * @brief Finalize the creation of @ref RigidObject or @ref RigidScene that
+   * inherits from this class.
+   */
+  virtual void finalizeObject() = 0;
```
aclegg3, line 105: Are we assuming that nothing in `finalizeObject` or `finalizeObject_LibSpecific` can fail? Should this be a boolean return with destruction/de-allocation guard in `PhysicsManager`?

src/esp/physics/RigidBase.h
```diff
@@ -106,14 +112,15 @@ class RigidBase : public Magnum::SceneGraph::AbstractFeature3D {
...
       const assets::ResourceManager& resMgr) = 0;
-
- public:
   /**
-   * @brief Finalize this object with any necessary post-creation processes.
+   * @brief any physics-lib-specific finalization code that needs to be run
+   * after @ref RigidObject or @ref RigidScene is created.
    */
-  virtual void finalizeObject() {}
+  virtual void finalizeObject_LibSpecifc() = 0;
```
aclegg3, line 121: Typo
```suggestion
  virtual void finalizeObject_LibSpecific() = 0;
```

src/esp/physics/RigidObject.h
```diff
@@ -101,30 +101,58 @@ class RigidObject : public RigidBase {
...
    * @brief Virtual destructor for a @ref RigidObject.
    */
-  virtual ~RigidObject(){};
+  virtual ~RigidObject() {}
 
   /**
-   * @brief Initializes the @ref RigidObject or @ref RigidScene that inherits
-   * from this class
-   * @param physicsAttributes The template structure defining relevant
+   * @brief Initializes the @ref RigidObject that inheritsfrom this class
```
aclegg3, line 107:
```suggestion
   * @brief Initializes the @ref RigidObject that inherits from this class
```

src/esp/physics/RigidObject.h
```diff
@@ -101,30 +101,58 @@ class RigidObject : public RigidBase {
...
    * @return true if initialized successfully, false otherwise.
    */
-  virtual bool initializationFinalize(
+  bool initialization_LibSpecific(
       const assets::ResourceManager& resMgr) override;
 
+  /**
+   * @brief any physics-lib-specific finalization code that needs to be run
+   * after @ref RigidObject is created.Overridden
+   * by inheriting class specific to certain physics libraries.Necessary to
```
aclegg3, line 151:
```suggestion
   * after @ref RigidObject is created. Overridden
   * by inheriting class specific to certain physics libraries. Necessary to
```

src/esp/physics/RigidObject.h
```diff
@@ -152,7 +180,7 @@ class RigidObject : public RigidBase {
 
  public:
   ESP_SMART_POINTERS(RigidObject)
-};
+};  // namespace physics
```
aclegg3, line 183:
```suggestion
};  // class RigidObject
```

src/esp/physics/RigidScene.h
```diff
@@ -19,31 +19,58 @@ class RigidScene : public RigidBase {
...
+  bool initialize(const assets::ResourceManager& resMgr,
+                  const std::string& handle) override;
+
+  /**
+   * @brief Get the template used to initialize this object or scene.
+   *
+   * AbstractPhysicsAttributes templates are able to be changed between
+   * instances of objects.
+   * @return The initialization settings of this object instance, casted to the
+   * appropriate value.
```
aclegg3, line 41: Just an opinion, feel free to disregard.
```suggestion
   * @brief Get the template used to initialize this scene.
   *
   * See @ref RigidBase::getInitializationAttributes.
   * @return The initialization settings of this scene instance.
```

src/esp/physics/RigidScene.h
```diff
@@ -19,31 +19,58 @@ class RigidScene : public RigidBase {
...
-  virtual bool initializationFinalize(
+  bool initialization_LibSpecific(
       CORRADE_UNUSED const assets::ResourceManager& resMgr) override {
     return true;
   }
+  /**
+   * @brief any physics-lib-specific finalization code that needs to be run
+   * after@ref RigidScene is created.  Called from finalizeObject.  Overridden
+   * by inheriting class specific to certain physics libraries. Necessary to
+   * support kinematic objects without any dynamics support.
```
aclegg3, line 71:
```suggestion
   * after @ref RigidScene is created.  Called from finalizeObject.  Overridden
   * by inheriting class specific to certain physics libraries.
```

src/esp/physics/RigidScene.h
```diff
@@ -19,31 +19,58 @@ class RigidScene : public RigidBase {
...
+   */
+  void finalizeObject() override { finalizeObject_LibSpecifc(); }
 
  private:
   /**
    * @brief Finalize the initialization of this @ref RigidScene
-   * geometry.  This is overridden by inheriting objects
+   * geometry.  This is overridden by inheriting class specific to certain
+   * physics libraries.Necessary to support kinematic objects without any
+   * dynamics support.
```
aclegg3, line 58:
```suggestion
   * physics libraries.
```

src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -56,17 +54,20 @@ class BulletRigidObject : public BulletBase,
...
-   * desired object
+   * @param primTypeVal int value corresponding to assets::PrimObjTypes enum
+   * describing primitive collision shape.
+   * @param halfLength half length of object, for primitives using this value
    * @return a unique pointer to the bullet primitive object
    */
   std::unique_ptr<btCollisionShape> buildPrimitiveCollisionObject(
-      const assets::AbstractPrimitiveAttributes& primAttributes);
+      int primTypeVal,
+      double halfLength);
```
aclegg3, line 69: Why are we pulling back from directly consuming `AbstractPrimitiveAttributes` here? Seems like this change introduces unnecessary complexity (e.g. need to pass `halfLength` even for types which don't consume this).

src/tests/SimTest.cpp
```diff
@@ -419,30 +435,175 @@ void SimTest::loadingObjectTemplates() {
...
+    CORRADE_VERIFY(primAttr->getString("test") != primAttr2->getString("test"));
+    // remove modified attributes
+    esp::assets::AbstractPrimitiveAttributes::ptr primAttr3 =
+        assetAttribsMgr->removeTemplateByHandle(newHandle);
+    CORRADE_VERIFY(nullptr != primAttr3);
+  }
+  // empty vector of handles
+  primObjAssetHandles.clear();
+  {
+    // test creation off new object, using edited attributes
```
aclegg3, line 579: ?
```suggestion
    // test creation of new object using edited attributes.
```

src/esp/assets/Attributes.h
```diff
@@ -176,6 +189,12 @@ class PhysicsObjectAttributes : public AbstractPhysicsAttributes {
   void setCOM(const Magnum::Vector3& com) { setVec3("COM", com); }
   Magnum::Vector3 getCOM() const { return getVec3("COM"); }
 
+  // whether com is provided or not
+  void setCOMIsProvided(bool COMIsProvided) {
```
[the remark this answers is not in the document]
jturner65: Didn't want to add anything that wasn't there to this - seems to reference something that is not completely supported in physicsManager.

If you think this should be auto-set, I'll do it :).

src/esp/physics/RigidBase.h
jturner65: good idea.  I'll update it to support that.

src/esp/physics/RigidScene.h
jturner65: Don't we want that added info?  Some future maintainer might not understand what's going on.

src/esp/physics/bullet/BulletRigidObject.h
jturner65: Actually gets rid of a copy of the attributes.  Isn't needed until after we actually make a copy in the rigid object initialize

src/esp/assets/Attributes.h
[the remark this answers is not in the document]
jturner65: Changed as discussed - xComputeCOMFromShape (x==get/set) directive for com derivation on object creation, with the expectation that users will set the flag appropriately should they modify the com manually.

PR #661, opened 2020-06-25
[object picking] Step 3: select object in the viewer

bigbike:
## Motivation and Context
This diff introduces a new functionality that allows user to select the object (drawables) from the RGB mesh in the viewer

## Types of changes

## Checklist


src/esp/gfx/GenericDrawable.cpp
```diff
@@ -54,14 +56,18 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
     lightColors.emplace_back((*lightSetup_)[i].color);
   }
 
+  Mn::Color3 highLightColor = 0xa5c9ea_rgbf * 0.5;
```
aclegg3, line 59: Just an opinion. Maybe a yellow highlight color to stand out more? `#f5dc45`?

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -54,14 +56,18 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
...
       .setDiffuseColor(materialData_->diffuseColor)
       .setSpecularColor(materialData_->specularColor)
       .setShininess(materialData_->shininess)
       .setLightPositions(lightPositions)
       .setLightColors(lightColors)
-      .setObjectId(materialData_->perVertexObjectId ? 0 : node_.getId())
+      // e.g., semantic mesh has its own per vertex annotation, which has been
+      // uploaded to GPU so simply pass 0 to the uniform "objectId" in the
+      // fragment shader
+      .setObjectId(materialData_->perVertexObjectId ? 0 : drawableId_)
```
aclegg3, line 70: Possible un-intended consequences of this change. This will cause all `SEMANTIC` sensor output for objects added through the physics interface (or without `perVertexObjectId`) to be the unique DrawableId. 


**Replica before** (note the scene disappears because replaced node_.getId() for chairs was default -1)


**Replica after** (now the chairs are rendered with DrawableId here)


This will also cause scenes without semantic annotations to render DrawableId through the `SEMANTIC` sensor.

**Van-gogh-room before** (was all -1 since no semantics are configured)


**Van-gogh-room after** (now all rendered as DrawableId)

src/utils/viewer/viewer.cpp
```diff
@@ -552,11 +588,90 @@ void Viewer::viewportEvent(ViewportEvent& event) {
   renderCamera_->setViewport(event.windowSize());
   imgui_.relayout(Mn::Vector2{event.windowSize()} / event.dpiScaling(),
                   event.windowSize(), event.framebufferSize());
+
+  /* Recreate object ID reading renderbuffers that depend on viewport size */
+  selectionDepth_ = Mn::GL::Renderbuffer{};
```
eundersander, line 593: Some duplicate code here from around L333. It'd be nice to see a helper function instead.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -54,14 +56,18 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
     lightColors.emplace_back((*lightSetup_)[i].color);
   }
 
+  Mn::Color3 highLightColor = 0xa5c9ea_rgbf * 0.5;
   (*shader_)
-      .setAmbientColor(materialData_->ambientColor)
+      .setAmbientColor(selected_ ? highLightColor : materialData_->ambientColor)
```
eundersander, line 61: It feels a little weird for such application-specific selection logic to live in this GenericDrawable class, namely the choice of highlight color and the choice to override ambient color when highlighted. It feels more like the viewer class would like to make those decisions.

Maybe this is good enough for now and we can revisit it when we finalize our material API. Maybe the viewer will be able to easily swap in an alternate highlight material at that point.

src/utils/viewer/viewer.cpp
```diff
@@ -552,11 +588,90 @@ void Viewer::viewportEvent(ViewportEvent& event) {
...
+  selectionFramebuffer_
+      .attachRenderbuffer(Mn::GL::Framebuffer::BufferAttachment::Depth,
+                          selectionDepth_)
+      .attachRenderbuffer(Mn::GL::Framebuffer::ColorAttachment{1},
+                          selectionDrawableId_)
+      .setViewport({{}, event.framebufferSize()});
 }
 
 void Viewer::mousePressEvent(MouseEvent& event) {
+  if (event.button() == MouseEvent::Button::Right && objectSelectionOn_) {
```
eundersander, line 608: This is a cool feature and there's a fair amount of new code in viewer.cpp to support it. It'd be nice to see all this code encapsulated in an ObjectPickerHelper class which the viewer uses.

eundersander:
Nice feature. I left some comments. None are blocking but I'd appreciate replies to the comments.

src/esp/gfx/Drawable.h
```diff
@@ -50,6 +50,20 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
    */
   uint64_t getDrawableId() { return drawableId_; }
 
+  /**
+   * @brief Return if this drawable object is selected in the editor (e.g.,
+   * viewer)
+   * @return the drawable object state, if it is selected
+   */
+  bool selected() { return selected_; }
```
mosra, line 58: `getSelected`? `isSelected`? not exactly sure what's the naming scheme for bool getters here. Also

```suggestion
  bool selected() const { return selected_; }
```

src/utils/viewer/viewer.cpp
```diff
@@ -552,11 +588,90 @@ void Viewer::viewportEvent(ViewportEvent& event) {
...
+        position.x(),
+        selectionFramebuffer_.viewport().sizeY() - position.y() - 1};
+    const Mn::Range2Di area =
+        Mn::Range2Di::fromSize(fbPosition, Mn::Vector2i{1});
+
+    const Mn::UnsignedInt selectedObject =
+        selectionFramebuffer_.read(area, {Mn::PixelFormat::R32UI})
+            .pixels<Mn::UnsignedInt>()[0][0];
+
+    for (std::pair<std::string, esp::gfx::DrawableGroup>&& it :
```
mosra, line 663:
```suggestion
    for (std::pair<const std::string, esp::gfx::DrawableGroup>&& it :
```

or this is at least what latest Clang is always pointing out, that you'd get a copy otherwise (oh, wait, that means `DrawableGroup` is *copyable*? :scream:)

Same in all cases above.

src/utils/viewer/viewer.cpp
```diff
@@ -552,11 +588,90 @@ void Viewer::viewportEvent(ViewportEvent& event) {
...
+         sceneGraph.getDrawableGroups()) {
+      if (it.second.hasDrawable(selectedObject)) {
+        selectedDrawableId_ = selectedObject;
+        it.second.getDrawable(selectedDrawableId_)->setSelected(true);
+        break;
+      }
+    }  // for
+  }    // drawable selection
```
mosra, line 671:
```suggestion
  } else return;
```

we surely don't want to redraw & unconditionally accept all events, right?

src/utils/viewer/viewer.cpp
```diff
@@ -552,11 +588,90 @@ void Viewer::viewportEvent(ViewportEvent& event) {
...
   event.setAccepted();
-}
+  redraw();
+}  // namespace
```
mosra, line 674:
```suggestion
}
```

i don't think this closes a namespace

src/utils/viewer/viewer.cpp
```diff
@@ -552,11 +588,90 @@ void Viewer::viewportEvent(ViewportEvent& event) {
...
+      for (std::pair<std::string, esp::gfx::DrawableGroup>&& it :
+           sceneGraph.getDrawableGroups()) {
+        esp::gfx::Drawable* d = it.second.getDrawable(selectedDrawableId_);
+        if (d != nullptr) {
+          d->setSelected(false);
+          break;  // the selected object is found and processed
+        }
+      }
+      selectedDrawableId_ = -1;
+    }
```
mosra, line 635: why not store a drawable pointer instead to avoid the lookup?

src/utils/viewer/viewer.cpp
```diff
@@ -604,6 +719,9 @@ void Viewer::keyPressEvent(KeyEvent& event) {
     case KeyEvent::Key::Esc:
       std::exit(0);
       break;
+    case KeyEvent::Key::LeftShift:
+      objectSelectionOn_ = true;
+      break;
```
mosra, line 724: Instead of explicitly maintaining this state, you can change the check in `mousePressEvent()` to 

```cpp
if (event.button() == MouseEvent::Button::Right  && (event.modifiers() & MouseEvent::Modifier::Shift))
```

That's the designated way to do this. Then you also don't need `keyReleaseEvent()` at all.

src/esp/gfx/GenericDrawable.cpp
mosra: Feels quite weird for me as well. I'd say the main goal for most of the code is to be optimized for the training use case, and since object picking is not part of that use case, it's extra overhead that doesn't belong there.

In the magnum-player app I'm adding a temporary second drawable for the highlighted object, which I think is a better way. Again, the overhead of drawing the same object twice doesn't matter as much in this case because it's the viewer and we don't need it to run 1000 FPS.

Additionally, the extra drawable would then allows you to use a totally different shader and thus open the possibilities for wireframe visualization, normal visualization, topology/material inspection, overdraw / texture LOD visualization etc etc., like the latest magnum-player has.

mosra:
So, about the conflict with #668 and what we discussed yesterday:

After going through the code of both, I don't see a need for adding a *third* object ID input to magnum's builtin Phong shader. The shader simply [does a sum](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1Phong.html#aa1647bfd654e1ad7af3c711f712e20c8a92d8291a3d82dc28cca9fb0f9e7753ce) of the per-vertex and uniform ID, so both of those can be set at the same time, you only need to ensure the two don't stomp on each other's bits.

About fitting both drawable ID and semantic ID into a single 32-bit attribute -- well, let's look at the data. I just added a `--bounds` option to `magnum-sceneconverter` for that and, in case of Replica, the largest ID range was in `apartment_0`, about 500 IDs (which fits easily into 10 bits). In case of MP3D the largest file I'm aware of (`B6ByNegPMKs`) had about 2.2k (which fits easily into 12 bits), and I doubt having more than 4 000 *different* semantic IDs is a realistic scenario.

Even going with just 12 bits, that leaves 20 bits left for drawable IDs -- which is a million drawables. I'd argue that there would be something *terribly wrong* if we want to issue a million draw calls / draw a million meshes each frame. GPUs already struggle drawing millions of *triangles*, let alone separate meshes, and even with some particle simulation in place I don't think you'd want to pick individual particles there :) Going into the extremes, in case you even would have a scene with millions of different objects, you would want to cull most of it away before drawing -- and then you could cull away the excess IDs as well and map the rest to a smaller contiguous range.

- - -

Actually, I just realized, what are we even talking about here? Unless I'm missing something there's two different and completely orthogonal cases:

1. Drawing semantic IDs into the object ID buffer, which is the case *only* in the offscreen rendering for the training use case. Object picking has no use here, so there's no conflict.
2. Drawing drawble ID into the object ID buffer, which is the case only for the viewer. *Integer* semantic ID has no place in the viewer as displays can show only RGB, so there's no conflict either -- and if you want to visualize the semantic ID in the player, you'd use [Shaders::MeshVisualizer3D](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1MeshVisualizer3D.html) or some such, which outputs the ID mapped to some color, which ultimately doesn't touch the object ID buffer either.

src/utils/viewer/viewer.cpp
bigbike: I agree with you.
Long story behind this non-pointer choice. 
Previously people have concerns when using raw pointers. They thought it was unsafe and leading to potential memory leaks or seg-fault, which I did not agree, but I chose to follow such redundant solution to address their concerns.

bigbike:
@mosra :
I agree with your 1st half of the comment (embedding drawableID, and semanticID into 1 unsigned int 32bits), but I have different opinions on your 2nd half of the comment (drawableID and semanticID were in different use case).

Yes, drawableID and semanticID are in different user case for now. But it is likely we need them both in headless mode in the future for training purpose. I can think of certain cases where drawableID can be useful, e.g., object recognition and detection. This is a door we should not shut down for future application.

I think I will go with the solution to "embed drawableID, and semanticID into 1 unsigned int 32bits".

Thank you very much for your very thorough review of this PR.

bigbike:
Any more thoughts on handling drawableID and semanticID?
@aclegg3 @mosra

src/utils/viewer/viewer.cpp
bigbike: I will use the raw pointers from now on whenever it is proper.

erikwijmans:
In my view, what the user really wants is to be able to set the ID of an _object_ -- be it if the object is defined via a node, via a drawable, a collection of nodes, a collection of drawables, etc.

bigbike:
> In my view, what the user really wants is to be able to set the ID of an _object_ -- be it if the object is defined via a node, via a drawable, a collection of nodes, a collection of drawables, etc.

That is what the semanticID does in #668.
Here, the drawableID is used to identify each individual object within the system. It is useful in object picking and per object material editing.

src/utils/viewer/viewer.cpp
bigbike: @mosra: Oh, I see. I introduced a bug here. If I use `auto&&` directly, it should be prevented.

The container for a STL `map` or `unordered_map` is actually `std::pair<const K, V>` NOT `std::pair<K, V>`.

Am I right? Wow, good catch.

bigbike:
I was working on it, addressing all the concerns from the reviewers.
Will inform the reviewers when it is done.
Currently, I am focusing on some other higher-prior tasks, and will switch back to this PR as soon as possible.

Thanks for checking it though. @jturner65

jturner65:
OK great. I just wanted to make sure I wasn't missing anything :)

bigbike:
Latest result using a MeshVisualizerDrawable. Picked Object is highlighted in white and wireframe.

Screenshot from 2020-07-24 16-22-38

bigbike:
Dear reviewers, this PR is now ready to be reviewed.

I eventually figured out the blending. The latest result is as follows:
Screenshot from 2020-07-28 16-19-51

jturner65:
> Dear reviewers, this PR is now ready to be reviewed.
> 
> I eventually figured out the blending. The latest result is as follows:
> Screenshot from 2020-07-28 16-19-51

Hi Yili

This all looks really cool.  It looks like, in busy scenes, seeing the selection/highlight might sometimes be difficult.  In the past, I've had success dealing with this kind of thing by dilating the wireframe and changing its color cyclically.  An example can be seen here, at about 17:00 (very near the end).  
[SOM Geometry From Pointcloud](
https://www.dropbox.com/home/Public/SOM?preview=SOM_GeomDemo1.mov)

This kind of thing might help the selection show up better in busy scenes, if you think that might be a problem, and shouldn't be that hard to implement (should it?)

src/esp/gfx/RenderCamera.cpp
```diff
@@ -152,6 +156,11 @@ uint32_t RenderCamera::draw(MagnumDrawableGroup& drawables, Flags flags) {
   }
 
   MagnumCamera::draw(drawableTransforms);
+
+  // reset
+  if (renderingForObjectPicking_) {
+    renderingForObjectPicking_ = false;
+  }
```
jturner65, line 163: why not just set renderingForObjectPicking_ to false always?  1 less check.

bigbike: Thanks. Because most of the time, it does not need to be reset.

bigbike:
@jturner65 :
"color cyclically" -- I cannot open that dropbox link. It says the folder does not exist. Sounds like a very cool feature by your description. But maybe too fancy for my user case though. :)

I am actually wondering if there is a way to increase line width of the wireframe. @mosra .

aclegg3:
> ... dilating the wireframe and changing its color cyclically. 
> This kind of thing might help the selection show up better in busy scenes, if you think that might be a problem, and shouldn't be that hard to implement (should it?)

Part of this sounds like a material editing feature (cycling colors). As it depends on time, it would be an animation feature of the viewer also, not inherent to the Simulator, Drawable, or Sensor implementation. Simply modify the SceneNode scale and material color for the attached Drawable. Might be a fun add-on later once material editing is un-blocked if we find it necessary.

src/esp/gfx/DrawableGroup.h
```diff
@@ -3,10 +3,14 @@
...
 
 #pragma once
+
 #include <Magnum/SceneGraph/Drawable.h>
 #include <Magnum/SceneGraph/FeatureGroup.h>
+#include <Magnum/SceneGraph/SceneGraph.h>
 #include <unordered_map>
 
+#include <functional>
+#include <unordered_map>
```
mosra, line 13: isn't clang tidy complaining here? the include is used twice

src/esp/gfx/MeshVisualizerDrawable.h
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * @brief Constructor
+   *
+   * @param node   Node, to which the drawable is attached
+   * @param shader Shader for the mesh visualizer
+   * @param mesh   Mesh to draw when on render.
+   * @param group  Drawable group this drawable will be added to.
+   */
+  explicit MeshVisualizerDrawable(scene::SceneNode& node,
+                                  Magnum::Shaders::MeshVisualizer3D* shader,
```
mosra, line 25:
```suggestion
                                  Magnum::Shaders::MeshVisualizer3D& shader,
```

Why not a reference? The pointer isn't tested for null anywhere inside so I don't think it can be null.

src/utils/viewer/ObjectPickingHelper.cpp
```diff
@@ -0,0 +1,132 @@
...
+  recreateFramebuffer(viewportSize);
+  selectionFramebuffer_.setViewport({{}, viewportSize});
+
+  shader_->setViewportSize(Mn::Vector2{viewportSize});
+  return *this;
+}
+
+unsigned int ObjectPickingHelper::getObjectId(
+    Magnum::Vector2i mouseEventPosition,
+    Magnum::Vector2i windowSize) {
```
mosra, line 88:
```suggestion
    const Magnum::Vector2i& mouseEventPosition,
    const Magnum::Vector2i& windowSize) {
```

I think it's like this everywhere else, so why not be consistent. Same in other cases.

src/utils/viewer/ObjectPickingHelper.cpp
```diff
@@ -0,0 +1,132 @@
...
+  if (meshVisualizerDrawable_) {
+    delete meshVisualizerDrawable_;
+    meshVisualizerDrawable_ = nullptr;
+  }
+
+  if (!pickedObject) {
+    return *this;
+  }
+
+  // magnum scene graph will handle the recycling even we did not recycle it by
```
mosra, line 125: > recycling

you mean ... "garbage collection"? :) i never saw this term used in this context

src/utils/viewer/ObjectPickingHelper.h
```diff
@@ -0,0 +1,67 @@
...
+   * @brief create a mesh visualizer (a drawable) for the picked object
+   * @param pickedObject, the picked object (drawable)
+   */
+  ObjectPickingHelper& createPickedObjectVisualizer(
+      esp::gfx::Drawable* pickedObject);
+
+  /**
+   * @brief return true if an object is picked at the moment
+   */
+  bool objectPicked() { return meshVisualizerDrawable_ != nullptr; }
```
mosra, line 53:
```suggestion
  bool isObjectPicked() const { return meshVisualizerDrawable_ != nullptr; }
```

Not sure about the naming (`getIs`?), but if everything has `get`, this should have something also. Oh, and `const`.

src/utils/viewer/viewer.cpp
```diff
@@ -492,6 +506,33 @@ void Viewer::toggleNavMeshVisualization() {
...
+  if (!objectPickingHelper_->objectPicked()) {
+    // nothing is picked, return directly
+    return;
+  }
+  // setup blending function
+  Mn::GL::Renderer::enable(Mn::GL::Renderer::Feature::Blending);
+  Mn::GL::Renderer::setBlendColor(Magnum::Color4{0.0, 0.0, 0.0, 0.5});
+  Mn::GL::Renderer::setBlendFunction(
+      Mn::GL::Renderer::BlendFunction::ConstantAlpha,
+      Mn::GL::Renderer::BlendFunction::OneMinusConstantAlpha);
```
mosra, line 519: Why not have the selection / wireframe color with alpha set to `0.5f` instead? That then wouldn't need a special blend function, and thus two state switches less.

src/utils/viewer/viewer.cpp
```diff
@@ -570,10 +610,47 @@ void Viewer::viewportEvent(ViewportEvent& event) {
...
+      auto* pickedDrawable = it.second.getDrawable(objectId);
+      objectPickingHelper_->createPickedObjectVisualizer(pickedDrawable);
+      break;
+    }
+  }
 }
 
 void Viewer::mousePressEvent(MouseEvent& event) {
+  if (event.button() == MouseEvent::Button::Right &&
+      (event.modifiers() & MouseEvent::Modifier::Shift)) {
```
mosra, line 629: Is there some existing documentation for the viewer utility (such as [magnum-player](https://doc.magnum.graphics/magnum/magnum-player.html#magnum-player-controls) has) where this new key control could get listed?

Otherwise <kbd>Shift</kbd> + <kbd>right click</kbd> is quite hard to discover I fear.

src/utils/viewer/ObjectPickingHelper.h
```diff
@@ -0,0 +1,67 @@
...
+   * @brief get the object id, aka, the drawable id defined in Drawable class
+   * @param eventPosition, mouse event position
+   * @param windowSize, the size of the GUI window
+   */
+  unsigned int getObjectId(Magnum::Vector2i mouseEventPosition,
+                           Magnum::Vector2i windowSize);
+
+  /**
+   * @brief create a mesh visualizer (a drawable) for the picked object
+   * @param pickedObject, the picked object (drawable)
```
mosra, line 45: Can the `pickedObject` be null? It can, but the doc doesn't mention that.

src/utils/viewer/ObjectPickingHelper.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <Magnum/Shaders/Generic.h>
+#include <Magnum/Shaders/MeshVisualizer.h>
+
+namespace Cr = Corrade;
+namespace Mn = Magnum;
+using Mn::Math::Literals::operator""_rgbf;
+
+ObjectPickingHelper::ObjectPickingHelper(Mn::Vector2i viewportSize) {
+  // create the framebuffer and set the color attachment
+  recreateFramebuffer(viewportSize).mapForDraw();
```
mosra, line 19:
```suggestion
  recreateFramebuffer(viewportSize);
  mapForDraw();
```

less confusing this way I think (I would even go further and just make all methods return `void` because I don't see method chaining having any use in this context and it's just noise)

src/utils/viewer/ObjectPickingHelper.cpp
```diff
@@ -0,0 +1,132 @@
...
+  CORRADE_INTERNAL_ASSERT(
+      selectionFramebuffer_.checkStatus(Mn::GL::FramebufferTarget::Draw) ==
+      Mn::GL::Framebuffer::Status::Complete);
+
+  // create a shader
+  Mn::Shaders::MeshVisualizer3D::Flags flags =
+      Mn::Shaders::MeshVisualizer3D::Flag::Wireframe;
+  shader_ = std::make_unique<Mn::Shaders::MeshVisualizer3D>(flags);
```
mosra, line 27: The `shader_` wouldn't need to be a pointer at all, and you could initialize it directly in the class definition in the header:

```cpp
Mn::Shaders::MeshVisualizer3D shader_{Mn::Shaders::MeshVisualizer3D::Flag::Wireframe};
```

This is only a lot of unnecessarily verbose code.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -54,14 +55,26 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
     lightColors.emplace_back((*lightSetup_)[i].color);
   }
 
+  bool usingDrawableId = false;
+  {
+    RenderCamera& cam = static_cast<RenderCamera&>(camera);
+    usingDrawableId = cam.isRenderingForObjectPicking();
+  }
```
mosra, line 62: why not just a single expression for such a thing?

```suggestion
  bool usingDrawableId = static_cast<RenderCamera&>(camera).isRenderingForObjectPicking();
```

src/utils/viewer/viewer.cpp
```diff
@@ -529,6 +566,8 @@ void Viewer::drawEvent() {
     physicsManager_->debugDraw(projM * camM);
   }
 
+  drawPickedObject();
```
mosra, line 569: I would personally just inline contents of the function here, because:

- it can only be called from this exact spot as it prepares blend state for imgui
- it's not called from anywhere else
- it does the exact same `RenderCamera::Flags` setup as above, thus that part could get deduplicated

that way the `drawEvent()` will get longer, yes, but it'll become much clearer what state is being set and restored, and thus also any potentially redundant operations

mosra:
> I am actually wondering if there is a way to increase line width of the wireframe. @mosra .

yes, you're right, there is :)

For the visualization, apart from a cyclic color I could imagine some "breathing" effect where the alpha of the visualization gets cyclically changed, maybe using some of the [easing functions](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Animation_1_1Easing.html). Cyclic color could be then implemented by picking from an animated offset in one of the [colormaps](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1DebugTools_1_1ColorMap.html) for example. I don't think animations are used anywhere yet, but I think it shouldn't be hard to set something up -- actually I wanted to code up a simple animation example for fun, so you could then steal from there.

However I'm not really a fan of animations as I find them distracting, so another alternative approach could be using some of the [advanced blend modes](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Renderer.html#aef456700b0c3416041ce33f03ccc456c) like for example pushing up the saturation / brightness for a particular object. Those extensions are pretty common nowadays so I don't think that would be a compatibility problem (well, except Mac, but ... :trollface:)

src/esp/gfx/DrawableGroup.h
bigbike: I do not have clang tidy installed locally on a linux machine. Will do it.

src/utils/viewer/ObjectPickingHelper.cpp
bigbike: re-creating the frame buffers and putting them immediately in use is quite natural to me. May I ask why you think method chaining here does not make sense to you?

src/utils/viewer/ObjectPickingHelper.cpp
bigbike: I did not realize the flag of the shader is actually fixed before coding up everything. Will change.

src/utils/viewer/ObjectPickingHelper.cpp
bigbike: because I thought Vector2i should be very light, almost like an "int", where I thought "passing by value" was OK.

src/utils/viewer/ObjectPickingHelper.cpp
bigbike: Yes, garbage collection. The scene graph will destroy all the features attaching a node. 
That is what I meant.

src/utils/viewer/ObjectPickingHelper.h
bigbike: Let just use isObjectPicked(), one exception for the getter here.
Otherwise it is wrong in terms of grammar.

Yes, const will be added.

src/utils/viewer/ObjectPickingHelper.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+  shader_->setViewportSize(Mn::Vector2{viewportSize});
+
+  if (flags & Mn::Shaders::MeshVisualizer3D::Flag::Wireframe) {
+    (*shader_).setColor(0x2f83cc_rgbf).setWireframeColor(0xdcdcdc_rgbf);
```
bigbike, line 32: @mosra: Should I set the alpha for the color of the MeshVisualizer3D here??

src/utils/viewer/viewer.cpp
bigbike: Very good point. In the previous version, I put it in the keyPressEvent function. So, somehow people can know this functionality.
Then I simplified the code in this way as you suggested.

No, we do not have one to describe the functionality. Usually we go and read the keyPressEvent().

src/utils/viewer/ObjectPickingHelper.cpp
mosra: Because at first it looked like `recreateFramebuffer()` returns some other object (instead of `*this`), so I had to check what was the actual return type of it. 

You don't save any keystrokes with that and it only adds congitive load.

src/utils/viewer/ObjectPickingHelper.cpp
mosra: Yep, `0x2f83cc7f_rgbaf` for example.

bigbike:
Screenshot from 2020-08-02 14-36-42

Improved the code a lot.
@mosra : would you please take a look? Thanks!

src/utils/viewer/viewer.cpp
mosra: @bigbike i see you marked this function as `inline`, but that's not what I meant -- I meant pasting its contents directly into the `drawEvent()`, for the reasons above

bigbike: I see your point.
I made it an independent function, even though it was only called here, simply because it can hide details about picked object drawing. People may not be interested.

Since what is inside the function now is not too long, I can move it here directly.

PR #668, opened 2020-07-01
Configurable Semantic id

aclegg3:
## Motivation and Context

Many use cases require the ability to set semantic id for objects introduced to the scene via the physics pipeline. This PR aims to address that need.

Currently, `SEMANTIC` sensor observations include per-vertex object ids imported from some datasets (e.g. MP3D and Replica). All other objects are rendered through this sensor by `SceneNode::id_` which defaults to -1 (not ideal as it is converted to unsigned int). This PR introduces a new `SceneNode::semanticId_` variable for this purpose with default 0 and necessary getters/setters + python property. 

A single object will have multiple SceneNodes which can have Drawables attached. This PR adds a function to assist users with setting the semantic id for all object nodes, an object template field for default semantic id, and a function to query visual SceneNodes cached on object creation.

A tutorial script is added for this feature (tutorial page pending).

Previous, outdated iteration of this feature: #341 

## How Has This Been Tested

New tutorial script (also run with pytest): `python <path-to-habitat>/examples/tutorials/semantic_id_tutorial.py`

**Note: no per-vertex semantic scenes exist for testing yet. This would be an ideal use case for that.**
(Tested locally on Replica)

**Example output:**
Empty room (no per-vertex ids, so default semanticId_ == 0)

Chairs added with default semanticId_ == 0.

Set semanticId_ = 20 for both chairs.

Set semanticId_ = 10 for one chair.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1272,10 +1272,12 @@ bool ResourceManager::instantiateAssetsOnDemand(
   return true;
 }  // ResourceManager::instantiateAssetsOnDemand
 
-void ResourceManager::addObjectToDrawables(const std::string& objTemplateHandle,
-                                           scene::SceneNode* parent,
-                                           DrawableGroup* drawables,
-                                           const Mn::ResourceKey& lightSetup) {
+std::vector<scene::SceneNode*> ResourceManager::addObjectToDrawables(
+    const std::string& objTemplateHandle,
+    scene::SceneNode* parent,
+    DrawableGroup* drawables,
+    const Mn::ResourceKey& lightSetup) {
+  std::vector<scene::SceneNode*> newNodes;
   if (parent != nullptr and drawables != nullptr) {
     //! Add mesh to rendering stack
 
@@ -1297,22 +1299,32 @@ void ResourceManager::addObjectToDrawables(const std::string& objTemplateHandle,
     // need a new node for scaling because motion state will override scale
     // set at the physical node
     scene::SceneNode& scalingNode = parent->createChild();
+    newNodes.push_back(&scalingNode);
     Magnum::Vector3 objectScaling = physicsObjectAttributes->getScale();
     scalingNode.setScaling(objectScaling);
 
-    addComponent(loadedAssetData.meshMetaData, scalingNode, lightSetup,
-                 drawables, loadedAssetData.meshMetaData.root);
+    std::vector<scene::SceneNode*> componentNodes =
+        addComponent(loadedAssetData.meshMetaData, scalingNode, lightSetup,
+                     drawables, loadedAssetData.meshMetaData.root);
+
+    for (auto node : componentNodes) {
+      newNodes.push_back(node);
+    }
   }  // should always be specified, otherwise won't do anything
+  return newNodes;
 }  // addObjectToDrawables
 
 //! Add component to rendering stack, based on importer loading
-void ResourceManager::addComponent(const MeshMetaData& metaData,
-                                   scene::SceneNode& parent,
-                                   const Mn::ResourceKey& lightSetup,
-                                   DrawableGroup* drawables,
-                                   const MeshTransformNode& meshTransformNode) {
+std::vector<scene::SceneNode*> ResourceManager::addComponent(
+    const MeshMetaData& metaData,
+    scene::SceneNode& parent,
+    const Mn::ResourceKey& lightSetup,
+    DrawableGroup* drawables,
+    const MeshTransformNode& meshTransformNode) {
+  std::vector<scene::SceneNode*> newNodes;
   // Add the object to the scene and set its transformation
   scene::SceneNode& node = parent.createChild();
+  newNodes.push_back(&node);
   node.MagnumObject::setTransformation(
       meshTransformNode.transformFromLocalToParent);
 
@@ -1332,8 +1344,13 @@ void ResourceManager::addComponent(const MeshMetaData& metaData,
 
   // Recursively add children
   for (auto& child : meshTransformNode.children) {
-    addComponent(metaData, node, lightSetup, drawables, child);
+    std::vector<scene::SceneNode*> childNodes =
+        addComponent(metaData, node, lightSetup, drawables, child);
+    for (auto node : childNodes) {
+      newNodes.push_back(node);
+    }
   }
+  return newNodes;
 }  // addComponent
 
 void ResourceManager::addMeshToDrawables(const MeshMetaData& metaData,
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1307,9 +1307,8 @@ std::vector<scene::SceneNode*> ResourceManager::addObjectToDrawables(
         addComponent(loadedAssetData.meshMetaData, scalingNode, lightSetup,
                      drawables, loadedAssetData.meshMetaData.root);
 
-    for (auto node : componentNodes) {
-      newNodes.push_back(node);
-    }
+    newNodes.insert(newNodes.end(), componentNodes.begin(),
+                    componentNodes.end());
   }  // should always be specified, otherwise won't do anything
   return newNodes;
 }  // addObjectToDrawables
@@ -1346,9 +1345,7 @@ std::vector<scene::SceneNode*> ResourceManager::addComponent(
   for (auto& child : meshTransformNode.children) {
     std::vector<scene::SceneNode*> childNodes =
         addComponent(metaData, node, lightSetup, drawables, child);
-    for (auto node : childNodes) {
-      newNodes.push_back(node);
-    }
+    newNodes.insert(newNodes.end(), childNodes.begin(), childNodes.end());
   }
   return newNodes;
 }  // addComponent
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -876,7 +876,10 @@ bool ResourceManager::loadGeneralMeshData(
     }
   }  // forceReload
 
-  addComponent(meshMetaData, newNode, lightSetup, drawables, meshMetaData.root);
+  std::vector<scene::SceneNode*>
+      visNodeCache;  // TODO: cache visual nodes added by this process
+  addComponent(meshMetaData, newNode, lightSetup, drawables, meshMetaData.root,
+               visNodeCache);
   return true;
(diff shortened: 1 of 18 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -1297,22 +1299,32 @@ void ResourceManager::addObjectToDrawables(const std::string& objTemplateHandle,
...
-                 drawables, loadedAssetData.meshMetaData.root);
+    std::vector<scene::SceneNode*> componentNodes =
+        addComponent(loadedAssetData.meshMetaData, scalingNode, lightSetup,
+                     drawables, loadedAssetData.meshMetaData.root);
+
+    for (auto node : componentNodes) {
+      newNodes.push_back(node);
+    }
   }  // should always be specified, otherwise won't do anything
+  return newNodes;
```
jturner65, line 1314: How about something like this? 
```suggestion
    componentNodes.push_back(&scalingNode);
    newNodes = componentNodes;
    }
  }  // should always be specified, otherwise won't do anything
  return newNodes;
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -1332,8 +1344,13 @@ void ResourceManager::addComponent(const MeshMetaData& metaData,
 
   // Recursively add children
   for (auto& child : meshTransformNode.children) {
-    addComponent(metaData, node, lightSetup, drawables, child);
+    std::vector<scene::SceneNode*> childNodes =
+        addComponent(metaData, node, lightSetup, drawables, child);
+    for (auto node : childNodes) {
+      newNodes.push_back(node);
+    }
```
jturner65, line 1351: How about this?
```suggestion
    newNodes.insert(newNodes.end(), childNodes.begin(), childNodes.end());
```

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -61,7 +61,8 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
       .setShininess(materialData_->shininess)
       .setLightPositions(lightPositions)
       .setLightColors(lightColors)
-      .setObjectId(materialData_->perVertexObjectId ? 0 : node_.getId())
+      //.setObjectId(materialData_->perVertexObjectId ? 0 : node_.getId())
```
jturner65, line 64: Can this be removed?

src/esp/physics/PhysicsManager.cpp
```diff
@@ -103,9 +103,16 @@ int PhysicsManager::addObject(const std::string& configFileHandle,
...
+  std::vector<scene::SceneNode*> newNodes =
+      resourceManager_.addObjectToDrawables(
+          configFileHandle, existingObjects_.at(nextObjectID_)->visualNode_,
+          drawables, lightSetup);
+
+  existingObjects_.at(nextObjectID_)
+      ->visualNodes_.push_back(existingObjects_.at(nextObjectID_)->visualNode_);
+  for (auto node : newNodes) {
+    existingObjects_.at(nextObjectID_)->visualNodes_.push_back(node);
+  }
```
jturner65, line 115: Maybe this? 

```suggestion
  existingObjects_.at(nextObjectID_)->visualNodes_.insert( existingObjects_.at(nextObjectID_)->visualNodes_.end(), newNodes.begin(), newNodes.end());
```

src/esp/assets/ResourceManager.cpp
aclegg3: 🤔  This will break the depth-first ordering of the nodes. I don't think this matters now, just throwing it out there.

jturner65: Oh, I was not aware we were hoping to retain ordering.  Then I would still suggest we do the insert method like i have below, since it is more efficient.

aclegg3: Agreed. I don't think there is a reason to maintain order, but we may as well if it's easy. Nice property to have 🤷‍♂️

aclegg3:
> It seems like this PR and #661 are in conflict with each other. Both seem to rely on the `setObjectId` field in the shader and the functionalities seem contradictory.

Yes, we already spoke about this and resolved the conflict. 👍

src/esp/assets/ResourceManager.cpp
```diff
@@ -1332,8 +1343,11 @@ void ResourceManager::addComponent(const MeshMetaData& metaData,
 
   // Recursively add children
   for (auto& child : meshTransformNode.children) {
-    addComponent(metaData, node, lightSetup, drawables, child);
+    std::vector<scene::SceneNode*> childNodes =
+        addComponent(metaData, node, lightSetup, drawables, child);
```
mosra, line 1347: (Probably an unnecessary optimization given that there's a lot more allocations elsewhere anyway, but since I already went through the train of though I'll leave it here -- it's not a big change after all.)

- - -

I could imagine an `addComponentInternal(...., std::vector<scene::SceneNode*>& newNodes)` that avoids the extra copies/allocations on recursion:

```suggestion
    addComponentInternal(metaData, node, lightSetup, drawables, newNodes);
```

and the public `addComponent()` would then look simply like this:

```cpp
std::vector<scene::SceneNode*> ResourceManager::addComponent(
    const MeshMetaData& metaData,
    scene::SceneNode& parent,
    const Mn::ResourceKey& lightSetup,
    DrawableGroup* drawables,
    const MeshTransformNode& meshTransformNode) {
  std::vector<scene::SceneNode*> newNodes;
  addComponentInternal(metaData, parent, lightSetup, drawables, meshTransformNode, newNodes);
  return newNodes;
}
```

src/esp/physics/PhysicsManager.cpp
```diff
@@ -103,9 +103,17 @@ int PhysicsManager::addObject(const std::string& configFileHandle,
...
+      resourceManager_.addObjectToDrawables(
+          configFileHandle, existingObjects_.at(nextObjectID_)->visualNode_,
+          drawables, lightSetup);
+
+  existingObjects_.at(nextObjectID_)
+      ->visualNodes_.push_back(existingObjects_.at(nextObjectID_)->visualNode_);
+  existingObjects_.at(nextObjectID_)
+      ->visualNodes_.insert(
+          existingObjects_.at(nextObjectID_)->visualNodes_.end(),
+          newNodes.begin(), newNodes.end());
```
mosra, line 116: Hmm, actually here the `*Internal` could be used too:

```suggestion
  // done first to have the order the same as before
  existingObjects_.at(nextObjectID_)
      ->visualNodes_.push_back(existingObjects_.at(nextObjectID_)->visualNode_);

  resourceManager_.addObjectToDrawablesInternal(
      configFileHandle, existingObjects_.at(nextObjectID_)->visualNode_,
      drawables, lightSetup, existingObjects_.at(nextObjectID_)->visualNodes_);
```

Of course now it needs a better name than `*Internal` ... `*Into`? `*Out`?

src/esp/physics/PhysicsManager.h
```diff
@@ -811,6 +811,16 @@ class PhysicsManager {
...
 
+  /**
+   * @brief Set the @ref esp::scene:SceneNode::semanticId_ for all visual nodes
+   * belonging to an object.
+   *
+   * @param objectID The object ID and key identifying the object in @ref
+   * existingObjects_.
+   * @param semanticId The desired semantic id for the object.
+   */
+  void setSemanticId(const int physObjectID, uint32_t semanticId);
```
mosra, line 822:
```suggestion
  void setSemanticId(int physObjectID, uint32_t semanticId);
```

src/esp/physics/RigidBase.h
```diff
@@ -598,6 +610,10 @@ class RigidBase : public Magnum::SceneGraph::AbstractFeature3D {
    */
   scene::SceneNode* visualNode_ = nullptr;
 
+  //! all nodes created when this object's render asset was added to the
+  //! SceneGraph
+  std::vector<esp::scene::SceneNode*> visualNodes_;
```
mosra, line 615: Wait, is this whole thing actually needed? Contents of this vector is the same as if you would recursively collect all children of `visualNode_`, no?

I'm just trying to minimize the amount of redundant state that needs to be maintained.

src/esp/sim/Simulator.h
```diff
@@ -555,6 +555,20 @@ class Simulator {
...
+   *
+   * @param semanticId The desired semantic id for the object.
+   * @param objectID The object ID and key identifying the object in @ref
+   * esp::physics::PhysicsManager::existingObjects_.
+   * @param sceneID !! Not used currently !! Specifies which physical scene of
+   * the object.
+   */
+  void setObjectSemanticId(uint32_t semanticId,
+                           const int objectID,
+                           const int sceneID);
```
mosra, line 570:
```suggestion
                           int objectID,
                           int sceneID);
```

Some shitty IDEs insist on adding the unnecessary `const` to the declaration if you have it in the definition, I hope it's not the case of yours :D

mosra:
Regarding the conflict between this and #661 -- I think this one continues in the right direction, expanding the existing functionality of per-vertex ID to per-object, and #661 needs to be reworked. I'll share my ideas there.

src/esp/physics/RigidBase.h
aclegg3: True for now. 👍  However, we are planning SceneGraph re-structuring in the near future which will invalidate this assumption, so planning for that.

src/esp/assets/ResourceManager.h
```diff
@@ -314,18 +314,21 @@ class ResourceManager {
    * gfx::Drawable will be rendered.
    * @param lightSetup The @ref LightSetup key that will be used
    * for the added component.
+   * @param visNodeCache Cache for pointers to all nodes created as the result
```
mosra, line 317:
```suggestion
   * @param[out] visNodeCache Cache for pointers to all nodes created as the result
```

will mark the parameter as output-only in the generated docs. Same in the other cases below.

bigbike:
> Regarding the conflict between this and #661 -- I think this one continues in the right direction, expanding the existing functionality of per-vertex ID to per-object, and #661 needs to be reworked. I'll share my ideas there.

This is because Alex and I talked about this issue a lot in the past week. We spent hours in the meetings to discuss the solution.
In the nutshell,  it is all about whether drawableID or SemanticID should occupy the uniform "objectID" in the shader. Finally I gave my way to the "semanticID" as it has more urgent needs currently.

The main idea and the direction in #661 is not wrong either.

PR #670, opened 2020-07-02
Compute/query navigable area in Pathfinder

aclegg3:
## Motivation and Context

In response to [this issue](https://github.com/facebookresearch/habitat-api/issues/420) this PR adds NavMesh area computation/query to Pathfinder. This computes the sum of all NavMesh polygon areas including all floors and isolated islands.

**Query the navigable area in python with:** `pathfinder.navigable_area`

## How Has This Been Tested

Validated area locally on a 1.1m radius cylinder and 0.1m radius agent resulting in ~`π` area.
Added golden number test to pytest for `skokloster-castle` and `van-gogh-room`.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -254,6 +254,8 @@ struct PathFinder::Impl {
 
   bool isNavigable(const vec3f& pt, const float maxYDelta = 0.5) const;
 
+  float computeNavigableArea();
+
   std::pair<vec3f, vec3f> bounds() const { return bounds_; };
 
   Eigen::Matrix<bool, Eigen::Dynamic, Eigen::Dynamic> getTopDownView(
@@ -282,6 +284,7 @@ struct PathFinder::Impl {
   std::pair<vec3f, vec3f> bounds_;
 
   void removeZeroAreaPolys();
+
   bool initNavQuery();
 
   Cr::Containers::Optional<std::tuple<float, std::vector<vec3f>>>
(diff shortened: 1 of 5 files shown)


tests/test_recompute_navmesh.py
```diff
@@ -69,31 +72,43 @@ def test_recompute_navmesh(test_scene, sim):
     assert sim.recompute_navmesh(sim.pathfinder, navmesh_settings)
     assert sim.pathfinder.is_loaded
 
+    # get the re-computed navmesh area
+    recomputedNavMeshArea1 = sim.pathfinder.navigable_area
+    assert loadedNavMeshArea == recomputedNavMeshArea1
```
aclegg3, line 77: This fails for skokloster castle scene because recomputed navmesh with default params includes the roof as navigable. We must generated that NavMesh with special considerations. 

Thoughts on resolution for this? @erikwijmans 

We could test this differently.

src/esp/nav/PathFinder.cpp
```diff
@@ -771,8 +780,11 @@ void PathFinder::Impl::removeZeroAreaPolys() {
       CORRADE_INTERNAL_ASSERT(poly != nullptr);
       CORRADE_INTERNAL_ASSERT(tmp != nullptr);
 
-      if (polyArea(poly, tile) < 1e-5) {
+      float polygonArea = polyArea(poly, tile);
+      if (polygonArea < 1e-5) {
         navMesh_->setPolyFlags(polyRef, POLYFLAGS_DISABLED);
+      } else {
```
erikwijmans, line 786:
```suggestion
      } else if (poly->flags & POLYFLAGS_WALK) {
```

That will make it so that is only navigable area.

tests/test_recompute_navmesh.py
erikwijmans: In terms of testing, this is really testing the recompute logic, not the area logic (i.e. as long as the area logic is deterministic, it can be broken and still pass this test).  Comparing to golden values is really the only thing I can think of to test regressions.  In terms of testing correctness in the first place, that would either need manual inspection to see if the area values are sensible or testing against some mesh where we can derive the area manually.

src/esp/nav/PathFinder.cpp
aclegg3: Thanks, this makes sense to accommodate future extensions. I don't see any other poly types being used in the codebase currently, is this correct or am I missing something?

erikwijmans: I believe recast sets the flags to all zeros by default, so anything with all zeros wouldn't be navigable, but maybe we don't have any of those?

tests/test_recompute_navmesh.py
aclegg3: Good point, still curious what we did to remove the ceiling here. This is relevant for analogous runtime re-compute during training.

I'll run some validation tests and then include golden values for the test scenes. 👍

aclegg3: Updated testing sections of description for this.

tests/test_recompute_navmesh.py
```diff
@@ -43,6 +43,13 @@ def test_recompute_navmesh(test_scene, sim):
     hab_cfg = examples.settings.make_cfg(cfg_settings)
     sim.reconfigure(hab_cfg)
 
+    # get the initial navmesh area. This test assumes default navmesh assets.
```
erikwijmans, line 46: One nit:  Can we move this to their own test?  It'll be easier to localize failures that way.

tests/test_navmesh.py
```diff
@@ -105,3 +109,33 @@ def test_recompute_navmesh(test_scene, sim):
...
+
+    cfg_settings = examples.settings.default_sim_settings.copy()
+    cfg_settings["scene"] = test_scene
+    hab_cfg = examples.settings.make_cfg(cfg_settings)
+    sim.reconfigure(hab_cfg)
+
+    # get the initial navmesh area. This test assumes default navmesh assets.
+    loadedNavMeshArea = sim.pathfinder.navigable_area
+    if test_scene.endswith("skokloster-castle.glb"):
+        assert loadedNavMeshArea == 226.65673828125
```
Skylion007, line 127: These asserts are going to be rather brittle, perhaps we should check if the are equal within an epsilon.

PR #663, opened 2020-06-26
Conda build for head/headless x with bullet/without bullet, CI test, issues fixed

mathfac:
#  Motivation and Context
- Made Linux binaries build for head / headless x with bullet / without bullet
- Improved Conda build readme
- Use Frozen `conda-build` lib version that works with current setup
- Added conda binaries build step to CI without upload to Anaconda upload for testing purposes. 
- Next Conda workflows are supported now:
Screenshot 2020-07-01 22 57 29

## How Has This Been Tested
- Built v0.1.5 binaries.
- Added CI build step

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #671, opened 2020-07-02
Move Conda Nightly builds to Docker

mathfac:
## Motivation and Context
https://github.com/facebookresearch/habitat-sim/pull/663 uses Ubuntu 16.04 as base OS to build conda builds. That limits conda build compatibility compare to using Centos7 as base OS.
- Added install Docker step to Nightly builds job
- Added "mesa-libGL-devel" install to conda Docker to improver build with Display
- Moved build conda env creation to docker build phase

## How Has This Been Tested
- Checked that nightly builds are working good on Colab

## Types of changes

## Checklist


.circleci/config.yml
```diff
@@ -463,17 +463,27 @@ jobs:
...
       - run:
           name: Build conda Linux packages
+          no_output_timeout: 240m
           command: |
               export PATH=$HOME/miniconda/bin:$PATH
               . activate habitat;
               cd habitat-sim
 
+              # install Docker
+              curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
Skylion007, line 473: Any reason you aren't using the magic `- setup_remote_docker` per the [documentation](https://circleci.com/docs/2.0/building-docker-images/)?

mathfac: `setup_remote_docker: step is not compatible with machine executor. You can use Docker without a special step if the build image has it installed.` Looks, like using `setup_remote_docker ` you should use different executor other than we use now.

Skylion007: @mathfac "It is possible to specify a different executor type for every job in your ‘.circleci/config.yml’ by specifying the executor type and an appropriate image. An image is a packaged system that has the instructions for creating a running environment. A container or virtual machine is the term used for a running instance of an image. For example, you could specify an executor type and an image for every job:" https://circleci.com/docs/2.0/executor-types/

Skylion007:
Building the conda viewer still doesn't work. :( If just building the headless conda binaries unblock us, that's fine with me. We can add the viewer back in a separate PR.

mathfac:
We should be good to go here.
Screenshot 2020-07-14 14 10 35

Issue #675, opened 2020-07-08
New objects not rendered with SEMANTIC sensor in MP3D


aclegg3:
## 🐛 Bug

New objects imported into an MP3D scene are not rendered with SEMANTIC sensor. 

In this case the separate semantic SceneGraph is being rendered. Imported objects are not currently being added to this separate SceneGraph and therefore are not rendered by the sensor.

## To Reproduce

1. Add an MP3D scene to the `test_scenes` list in `<path-to-habitat>/examples/tutorials/semantic_id_tutorial.py`. 
2. Run the script `python <path-to-habitat>/examples/tutorials/semantic_id_tutorial.py`

Example output:

![5](https://user-images.githubusercontent.com/1445143/86858516-a62e9a00-c075-11ea-8c92-7194c3c6ab93.jpg)

## Expected behavior

The chair and box should be rendered in the SEMANTIC image.

## Additional context

Related to new functionality in #668

aclegg3:
This is addressed by #679.

PR #679, opened 2020-07-08
Render OBJECT nodes in SEMANTIC sensor

aclegg3:
## Motivation and Context

This PR addresses issue #675. The purpose of a separate semantic scenegraph for datasets like MP3D is to handle two distinct render assets for the static scene geometry in RGB-D and SEMANTIC observation modes. Objects imported by the physics pipeline (i.e., `add_object`) have SceneNodes and Drawables attached to the RGB-D SceneGraph which are therefore missing in the SEMANTIC render which traverses only the separate semantic SceneGraph. 

This PR introduces:
- `SceneNodeType::OBJECT` and assigns this type to any SceneNodes created in the object instantiation pipeline and cached in #668. 
- A culling flag and corresponding function to render only `SceneNode`s with `SceneNodeType::OBJECT`.
- A 2nd render pass with non-OBJECT culling over the active SceneGraph when a separate semantic SceneGraph is in use.

Minor fix to `semantic_id_tutorial.py`:
- fixed some accidentally commented get_obs lines

## How Has This Been Tested

Locally with an MP3D asset as described in #675 with `semantic_id_tutorial.py`.

**Blocked on better testing by lack of available testing asset with separate semantic SceneGraph.**

Example Output:

## Types of changes

## Checklist


src/esp/sensor/PinholeCamera.cpp
```diff
@@ -110,6 +110,10 @@ void PinholeCamera::drawObservation(sim::Simulator& sim) {
     // TODO: check sim has semantic scene graph
     renderer->draw(*this, sim.getActiveSemanticSceneGraph(),
                    sim.isFrustumCullingEnabled());
+    if (&sim.getActiveSemanticSceneGraph() != &sim.getActiveSceneGraph()) {
+      renderer->draw(*this, sim.getActiveSceneGraph(),
+                     sim.isFrustumCullingEnabled(), true);
+    }
```
aclegg3, line 116: Note that this code is actually duplicated in python which calls directly into `render->draw` for sensors.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -57,15 +57,15 @@ void initGfxBindings(py::module& m) {
   py::class_<Renderer, Renderer::ptr>(m, "Renderer")
       .def(py::init(&Renderer::create<>))
       .def("draw",
-           py::overload_cast<sensor::VisualSensor&, scene::SceneGraph&, bool>(
-               &Renderer::draw),
+           py::overload_cast<sensor::VisualSensor&, scene::SceneGraph&, bool,
+                             bool>(&Renderer::draw),
            R"(Draw given scene using the visual sensor)", "visualSensor"_a,
-           "scene"_a, "frustumCulling"_a = true)
+           "scene"_a, "frustumCulling"_a = true, "objectsOnly"_a = false)
```
jturner65, line 63: As we talked about yesterday, instead of passing individual booleans (which may increase in number and impact a large amount of code) I would suggest we approach these draw state-related boolean arguments as a single int.  We define an enum of possible flag settings, pybind the enum and perhaps a 2 line helper function for taking a list of enum values and or'ing them together to produce the state int encoding, and then use the enum values anded with the state int encoding when we wish to consume them.  

This way, we can grow to use up to 32 booleans without requiring any refactoring of method headers or bindings.  As a user, too, i would not want to always have to enter all the flags on method call if i did not need them.

erikwijmans: On @jturner65's note here, we should definitely use kwargs for call the python side if we don't do the ENUM thing.
```suggestion
           "scene"_a, "frustum_culling"_a = true, "objects_only"_a = false)
```

Then we can do
```python
self._sim.renderer.draw(
                 self._sensor_object,
                 self._sim.get_active_scene_graph(),
                 frustum_culling=self._sim.frustum_culling,
                 objects_only=True,
             )
```

I am also a fan of doing kwargs in cpp via comments, i.e.
```cpp
/*frustumCulling=*/sim.isFrustumCullingEnabled(), /*objectsOnly=*/true
```

mosra: You can also use [Containers::EnumSet](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html) and bind it to a `py::enum_`, which will make it more typesafe. Check Magnum Python bindings sources for how this could be done (e.g. the shader flags), there are also a bunch of utils for adding bitwise operators to enum types.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -107,9 +109,30 @@ size_t RenderCamera::cull(
...
+    std::vector<std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>,
+                          Mn::Matrix4>>& drawableTransforms) {
+  auto newEndIter = std::remove_if(
+      drawableTransforms.begin(), drawableTransforms.end(),
+      [&](const std::pair<std::reference_wrapper<Mn::SceneGraph::Drawable3D>,
+                          Mn::Matrix4>& a) {
+        // obtain the absolute aabb
+        auto& node = static_cast<scene::SceneNode&>(a.first.get().object());
+        Corrade::Containers::Optional<Mn::Range3D> aabb =
+            node.getAbsoluteAABB();
```
mosra, line 122: `aabb` doesn't seem to be used here?

src/esp/sensor/PinholeCamera.cpp
```diff
@@ -110,6 +110,10 @@ void PinholeCamera::drawObservation(sim::Simulator& sim) {
     // TODO: check sim has semantic scene graph
     renderer->draw(*this, sim.getActiveSemanticSceneGraph(),
                    sim.isFrustumCullingEnabled());
+    if (&sim.getActiveSemanticSceneGraph() != &sim.getActiveSceneGraph()) {
```
mosra, line 113: I probably just missed something in the previous PRs -- there are two independent scenegraphs now? :)

src/esp/bindings/GfxBindings.cpp
mosra: > kwargs

Does anybody know if it's possible to *force* pybind arguments to be keyword-only? Something equivalent to the `*` in `draw(object, scene, *, frustum_culling, objects_only)`. I know only that if one doesn't annotate the arguments with `_a`, those are treated as position-only (equivalent to a `draw(object, scene, /)` signature).

src/esp/sensor/PinholeCamera.cpp
aclegg3: Unfortunately yes in some cases only which this PR is addressing. When we load MP3D for example we load an RGB-D asset in to one SceneGraph (where objects also go) and a separate semantic asset into another SceneGraph. This PR enables SEMANTIC sensor to render only the objects in the RGB-D SceneGraph as a 2nd pass after rendering the semantic SceneGraph.

src/esp/bindings/GfxBindings.cpp
aclegg3: I took a pass adding EnumSet following the example of the phong flags. Builds fine and C++ tests pass, but getting a python module issue. Anything glaringly wrong here @mosra?

src/esp/sensor/PinholeCamera.cpp
bigbike: @mosra: This (two scene graphs) was not newly introduced to the system. They were here in the system since day 1, when the semantic mesh was added.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -118,11 +137,21 @@ uint32_t RenderCamera::draw(MagnumDrawableGroup& drawables,
...
 
-  // draw just the visible part
-  size_t numVisibles = cull(drawableTransforms);
-  // erase all items that did not pass the frustum visibility test
-  drawableTransforms.erase(drawableTransforms.begin() + numVisibles,
-                           drawableTransforms.end());
+  if (flags & Flag::ObjectsOnly) {
+    // draw just the OBJECTS
+    size_t numObjects = cullNonObjects(drawableTransforms);
+    // erase all items that did not pass the frustum visibility test
```
bigbike, line 143: This comments must be removed.

src/esp/gfx/RenderCamera.h
```diff
@@ -62,6 +82,14 @@ class RenderCamera : public MagnumCamera {
               std::pair<std::reference_wrapper<Magnum::SceneGraph::Drawable3D>,
                         Magnum::Matrix4>>& drawableTransforms);
 
+  /**
+   * @brief Cull Drawables for SceneNodes which are not OBJECT type.
+   */
+  size_t cullNonObjects(
```
bigbike, line 88: I would call this function: removeNonObjects(...)

src/esp/gfx/RenderCamera.h
```diff
@@ -14,6 +14,25 @@ namespace gfx {
...
+   * @brief Rendering Flags
+   */
+  enum class Flag : unsigned int {
+    /**
+     * Cull Drawables with bounding boxes not intersecting the camera frustum.
+     */
+    FrustumCulling = 1 << 0,
+
+    /**
+     * Cull Drawables not attached to @ref SceneNodes with @ref
```
bigbike, line 27: Remove drawables that are NOT @ref scene::SceneNodeType::OBJECT.

src/esp/bindings/GfxBindings.cpp
erikwijmans: > force pybind arguments to be keyword-only

I don't think so -- I did a little looking as I _love_ keyword-only args and didn't find anything

src/esp/bindings/GfxBindings.cpp
```diff
@@ -57,15 +69,16 @@ void initGfxBindings(py::module& m) {
...
+           "scene"_a,
+           "flags"_a = RenderCamera::Flags{RenderCamera::Flag::FrustumCulling})
       .def("draw",
-           py::overload_cast<RenderCamera&, scene::SceneGraph&, bool>(
-               &Renderer::draw),
+           py::overload_cast<RenderCamera&, scene::SceneGraph&,
+                             RenderCamera::Flags>(&Renderer::draw),
            R"(Draw given scene using the camera)", "camera"_a, "scene"_a,
-           "frustumCulling"_a = true)
+           "flags"_a = RenderCamera::Flags{RenderCamera::Flag::FrustumCulling})
```
erikwijmans, line 81: This looks like what needs to be done:
```cpp
     .def(
          "draw",
          [](Renderer& self, sensor::VisualSensor& visualSensor,
             scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
            self.draw(visualSensor, sceneGraph, RenderCamera::Flags{flags});
          },
          R"(Draw given scene using the visual sensor)", "visualSensor"_a,
          "scene"_a,
          "flags"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
      .def(
          "draw",
          [](Renderer& self, RenderCamera& camera,
             scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
            self.draw(camera, sceneGraph, RenderCamera::Flags{flags});
          },
          R"(Draw given scene using the camera)", "camera"_a, "scene"_a,
          "flags"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
```

It's just the Flag enum that gets bound on the python side, not the whole flag set, so you have to write a wrapper lambda (Idk why @mosra didn't need to do a similar thing for the constructor on say the Phong shader but 🤷)

aclegg3: Thanks, @erikwijmans this works. Appreciate the suggestion!

src/esp/bindings/GfxBindings.cpp
```diff
@@ -46,7 +46,7 @@ void initGfxBindings(py::module& m) {
           R"(RenderCamera: The object of this class is a camera attached
       to the scene node for rendering.)");
 
-  py::enum_<RenderCamera::Flag> flags{render_camera, "Flags", "Flags"};
+  py::enum_<RenderCamera::Flag> flags{m, "Flags", "Flags"};
```
erikwijmans, line 49: This should still go on the render camera IMO.  You can access it from python as `Camera.Flags`

src/esp/bindings/GfxBindings.cpp
mosra: I *did* need that -- the `py::init<>()` has `Flag`, not `Flags`. Since `Flag` is convertible to `Flags` this worked: https://github.com/mosra/magnum-bindings/blob/24b18aff931131f62ae635f1f5db6f1668d7a731/src/python/magnum/shaders.cpp#L59-L60

However for everything else except a constructor you need to write a wrapper lambda, as you did here.

PR #677, opened 2020-07-08
Google Colab Tutorial Infrastructure

Skylion007:
## Motivation and Context

It allows us to begin to come up with some unified visualization utils and the ability to generate notebooks from existing tutorial scripts for our upcoming ECCV tutorial.
Features
* Pre-commit hook that generates colab compatible notebooks from python scripts and vice versa
* Script to run that pre-commit
* Automated formatting of all notebook code within JuPyText
* Pre-commit hook to ensure that no notebooks with output are uploaded to the Github
* Automatic installation of jupytext through the pre-commit hook
* New testing frameworks that allows you to run any python script with CLI args with pytest and have it count toward coverage. This allows us to test and maintain all example code.
* Updated exclusion code to black and such. (Black currently destroys the format for notebook scripts, but I hacked JuPyText to fix it.
* Unified/Optimized Colab Installation Script for all tutorials. 
* Early version of tutorial utilities file

## How Has This Been Tested
The functions have been run and optimized in Collab.
All pre-commit hooks have been tested.
The RigidObjects Notebook Tutorial has been converted to notebook form.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #683, opened 2020-07-10
Simulator NavMesh visualization

aclegg3:
## Motivation and Context

Users may want to visualize the NavMesh loaded/modified in python. Also useful for tutorial/demo visualizations. The PR adds a `toggle_navmesh_visualization` function to the Simulator class to visualize a wireframe of the current `Simulator::pathfinder_` object.

Supporting changes:
 - Added new default material: `WHITE_MATERIAL_KEY = "white_ambient"` for wiremesh primitive visualizations.
 - `ResourceManager::primtive_meshes_` refactored to map and added `ResourceManager::removePrimitiveMesh(id)` to avoid stale asset build-up.
 - Refactored `Viewer` utility to better reflect these changes.

## How Has This Been Tested

 - Called in pytest (no verification).
 - Demo of feature in preliminary tutorial file `examples/tutorials/navmesh_tutorial.py` (This can be removed before merge until a better tutorial can be generated after colab/utils process is stable).

Example output:

## Types of changes

## Checklist

- [x I have completed my CLA (see **CONTRIBUTING**)


src/esp/sim/Simulator.cpp
```diff
@@ -638,10 +638,37 @@ bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
...
+    return;
+  if (navMeshVisNode_ == nullptr && pathfinder_->isLoaded()) {
+    auto& sceneGraph = sceneManager_->getSceneGraph(activeSceneID_);
+    auto& rootNode = sceneGraph.getRootNode();
+    auto& drawables = sceneGraph.getDrawables();
+    navMeshVisNode_ = &rootNode.createChild();
+    navMeshVisPrimID_ = resourceManager_->loadNavMeshVisualization(
+        *pathfinder_, navMeshVisNode_, &drawables);
+    // navMeshVisNode_->translate({0, 0.1, 0}); //hover above the floor
+  } else if (navMeshVisNode_ != nullptr) {
```
erikwijmans, line 664: This clean-up logic also needs to be in `close()`

src/esp/sim/Simulator.cpp
```diff
@@ -638,10 +640,37 @@ bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
...
 
+void Simulator::toggleNavMeshVisualization() {
+  if (pathfinder_ == nullptr)
+    return;
+  if (navMeshVisNode_ == nullptr && pathfinder_->isLoaded()) {
+    auto& sceneGraph = sceneManager_->getSceneGraph(activeSceneID_);
+    auto& rootNode = sceneGraph.getRootNode();
+    auto& drawables = sceneGraph.getDrawables();
+    navMeshVisNode_ = &rootNode.createChild();
+    navMeshVisPrimID_ = resourceManager_->loadNavMeshVisualization(
```
eundersander, line 663: I see that loadNavMeshVisualization can return ID_UNDEFINED. I guess removePrimitiveMesh will fail a CHECK on the next toggle. Does that sound like the behavior we want?

src/utils/viewer/viewer.cpp
```diff
@@ -465,15 +473,18 @@ void Viewer::wiggleLastObject() {
 }
 
 void Viewer::toggleNavMeshVisualization() {
```
eundersander, line 475: Would be nice if we didn't have to duplicate navMeshVisPrimID_, navMeshVidNode_, and toggleNavMeshVisualization in Viewer and Simulator.

examples/tutorials/navmesh_tutorial.py
```diff
@@ -0,0 +1,230 @@
...
+from habitat_sim.utils import common as ut
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+data_path = os.path.join(dir_path, "../../data")
+output_path = os.path.join(dir_path, "navmesh_tutorial_output/")
+
+save_index = 0
+
+
+def make_video_cv2(observations, prefix="", open_vid=True, multi_obs=False):
```
eundersander, line 21: As you mentioned, it might be wise to omit this tutorial from this PR since Aaron is refactoring here.

Skylion007:
Now that #677 is merged, maybe now is the time to bite the bullet and convert it to notebook form.

examples/tutorials/navmesh_tutorial.py
Skylion007, line 21: We should extract this and some other code out and put it in viz_utils. Also a good time to start turning them into notebooks.

src/esp/sim/Simulator.cpp
aclegg3: Good point, thanks. 👍

src/utils/viewer/viewer.cpp
aclegg3: Agreed. We need to refactor Viewer to use the Simulator class and this kind of duplication won't be a problem anymore. 🤷‍♂️

examples/tutorials/navmesh_tutorial.py
aclegg3: Great point, let's cache this for a conversion PR.

src/utils/viewer/viewer.cpp
dhruvbatra: I know I'm late to this conversation, but completely agree.

PR #691, opened 2020-07-17
Add recompute_navmesh to demo_runner and examples

Skylion007:
## Motivation and Context

It allows benchmark.py and other scripts to run on scenes that don't come with a precomputed navmesh. I also added args to examples/example.py. It automatically runs if a navmesh isn't loaded.

## How Has This Been Tested

Added additional tests to tests/test_example.py

## Types of changes

## Checklist

### DIFF
--- a/examples/demo_runner.py
+++ b/examples/demo_runner.py
@@ -7,6 +7,7 @@ import math
 import multiprocessing
 import os
 import random
+import sys
 import time
 from enum import Enum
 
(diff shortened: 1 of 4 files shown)


examples/demo_runner.py
```diff
@@ -338,6 +339,14 @@ def init_common(self):
         random.seed(self._sim_settings["seed"])
         self._sim.seed(self._sim_settings["seed"])
 
+        recompute_navmesh = self._sim_settings.get("recompute_navmesh")
+        if recompute_navmesh or not self._sim.pathfinder.is_loaded:
```
aclegg3, line 343: Minor comment: this does not allow turning recompute off explicitly if desired. 🤷‍♂️

examples/demo_runner.py
```diff
@@ -338,6 +339,14 @@ def init_common(self):
...
+            if not self._sim_settings["silent"]:
+                print("Recomputing navmesh")
+            navmesh_settings = habitat_sim.NavMeshSettings()
+            navmesh_settings.set_defaults()
```
aclegg3, line 347: This will not adjust to agent params, but we don't include those in settings anyway.

Skylion007: Exactly my thoughts.

examples/demo_runner.py
Skylion007: @aclegg3 I was thinking about adding an explict if === false check so that would have errored out instead, but decided against it. Thoughts?

Issue #527, opened 2020-03-14
Error in WebGL Habitat-Sim while running the webserver


rutadesai:
## 🐛 Bug

I followed the instruction on https://github.com/facebookresearch/habitat-sim#webgl to run the webgl demo. However, in step 5., when I run: `python -m http.server 8000 --bind 127.0.0.1` , I get the following error: 

```
127.0.0.1 - - [14/Mar/2020 14:53:19] "GET /build_js/esp/bindings_js/bindings.html HTTP/1.1" 200 -
127.0.0.1 - - [14/Mar/2020 14:53:20] "GET /build_js/esp/bindings_js/WebApplication.css HTTP/1.1" 200 -
127.0.0.1 - - [14/Mar/2020 14:53:20] "GET /build_js/esp/bindings_js/WindowlessEmscriptenApplication.js HTTP/1.1" 200 -
127.0.0.1 - - [14/Mar/2020 14:53:20] "GET /build_js/esp/bindings_js/bundle.js?4c536547307b8acf0c58 HTTP/1.1" 200 -
127.0.0.1 - - [14/Mar/2020 14:53:20] code 404, message File not found
127.0.0.1 - - [14/Mar/2020 14:53:20] "GET /build_js/esp/bindings_js/hsim_bindings.js HTTP/1.1" 404 -
127.0.0.1 - - [14/Mar/2020 14:53:20] code 404, message File not found
```

Corresponding output in the browser: 

![image](https://user-images.githubusercontent.com/7421733/76691198-ab1e9800-6605-11ea-83f9-bb4946220a27.png)

Detailed view from the terminal:

```
python -m http.server 8000 --bind 127.0.0.1
Serving HTTP on 127.0.0.1 port 8000 (http://127.0.0.1:8000/) ...
127.0.0.1 - - [14/Mar/2020 14:53:19] "GET /build_js/esp/bindings_js/bindings.html HTTP/1.1" 200 -
127.0.0.1 - - [14/Mar/2020 14:53:20] "GET /build_js/esp/bindings_js/WebApplication.css HTTP/1.1" 200 -
127.0.0.1 - - [14/Mar/2020 14:53:20] "GET /build_js/esp/bindings_js/WindowlessEmscriptenApplication.js HTTP/1.1" 200 -
127.0.0.1 - - [14/Mar/2020 14:53:20] "GET /build_js/esp/bindings_js/bundle.js?4c536547307b8acf0c58 HTTP/1.1" 200 -
[...]

PR #699, opened 2020-07-24
[WIP] Extend Habitat-Sim WebGL support

Ram81:
## Motivation and Context

Adding support for following actions in Habitat-Sim WebGL setup:

1. Add object to scene
2. Remove object from scene
3. Grab object from the scene
4. Release object in the scene

Adding support for these actions allows us to use the WebGL set up for different tasks for data collection on AMT, etc.

## How Has This Been Tested

By running WebGL setup on local http server and validating execution of every action.

## Types of changes

## Checklist


Skylion007:
@Ram81 Make sure you have installed pre-commit and run it on your files so that they are properly formatted.

src/esp/bindings_js/bindings_js.cpp
```diff
@@ -216,6 +237,16 @@ EMSCRIPTEN_BINDINGS(habitat_sim_bindings_js) {
       .property("categories", &SemanticScene::categories)
       .property("objects", &SemanticScene::objects);
 
+  em::class_<SceneNode>("SceneNode")
+      .function("getId", &SceneNode::getId)
```
Skylion007, line 241: Any reason these are functions instead of read only properties?

src/esp/sim/Simulator.cpp
```diff
@@ -809,5 +821,168 @@ void Simulator::setObjectLightSetup(int objectID,
...
+  } else {
+    return;
+  }
+
+  esp::nav::NavMeshSettings navMeshSettings;
+  navMeshSettings.setDefaults();
+  navMeshSettings.agentRadius = 0.3f;
+
+  recomputeNavMesh(*getPathFinder().get(), navMeshSettings, true);
+  toggleNavMeshVisualization();
```
Skylion007, line 934: Why is this toggled twice?

src/esp/sim/Simulator.h
```diff
@@ -608,6 +608,25 @@ class Simulator {
   void setObjectLightSetup(int objectID,
                            const std::string& lightSetupKey,
                            int sceneID = 0);
+  
+  int findNearestObject(int refObjectID, float distance=1.0);
```
Skylion007, line 612: Please consider updating the Python bindings for these new functions too

src/esp/physics/bullet/BulletPhysicsManager.h
```diff
@@ -166,6 +166,19 @@ class BulletPhysicsManager : public PhysicsManager {
    */
   bool contactTest(const int physObjectID) override;
 
+  /**
```
Skylion007, line 169: Actual doc strings needed

build_js.sh
```diff
@@ -53,6 +53,8 @@ cmake ../src \
 cmake --build . -- -j 4
 cmake --build . --target install -- -j 4
 
+cp -v Release/bin/hsim_bindings.* esp/bindings_js/
```
Skylion007, line 56: We shouldn't need to do this ideally, something incorrect is configured in CMake. Can someone else tell me what cmake variable we don't have properly configured? @mosra @aclegg3 ?

src/utils/viewer/viewer.cpp
```diff
@@ -370,6 +392,268 @@ void Viewer::removeLastObject() {
   objectIDs_.pop_back();
 }
 
+// float Viewer::depthAt(const Mn::Vector2i& windowPosition) {
```
Skylion007, line 395: Should probably be removed

src/esp/bindings_js/modules/simenv_embind.js
```diff
@@ -59,7 +103,89 @@ class SimEnv {
...
+    this.setObjectMotionType(Module.MotionType.STATIC, objectId, 0);
+    return objectId;
+  }
+
+  /**
+   * Add a random primitive object to the environment.
+   * @returns {number} object ID or -1 if object was unable to be added
+   */
+  removeLastObject() {
+    let exsitingObjectIds = this.getExistingObjectIDs();
```
Skylion007, line 179: typo in variable name

src/esp/sim/Simulator.cpp
Ram81: this was test code. I will remove this

src/esp/sim/Simulator.h
Ram81: I am using the code that @dexter1691 created for actions grab/release. He is planning to submit these changes with python bindings in a separate PR.

Skylion007: Fair, but we should finalize the bindings so that they agree. Either please submit a micro-PR with those bindings ASAP @dexter1691, or just add the bindings for him. If they are the same, it shouldn't affect his PR.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -218,5 +218,28 @@ bool BulletPhysicsManager::contactTest(const int physObjectID) {
       ->contactTest();
 }
 
+btCollisionWorld::AllHitsRayResultCallback BulletPhysicsManager::castRay(
```
aclegg3, line 221: We are planning to get Bullet collision object ray-casting into master as a stand-alone feature in the next week or so. 👍

src/esp/sim/Simulator.cpp
```diff
@@ -809,5 +821,168 @@ void Simulator::setObjectLightSetup(int objectID,
...
+  toggleNavMeshVisualization();
+}
+
+Magnum::Vector3 Simulator::unproject(const Magnum::Vector2i& crosshairPos,
```
aclegg3, line 933: Maybe we want to support this in the VisualSensor or RenderCamera instead of through Simulator? Currently, there can be multiple sensors in the simulator with different transformations, so there is some ambiguity here.

src/esp/sim/Simulator.cpp
```diff
@@ -809,5 +821,168 @@ void Simulator::setObjectLightSetup(int objectID,
...
+  }
+
+  if ((hitPoint - refPoint).length() > distance) {
+    return -1;
+  }
+
+  return nearestObjId;
+}
+
+void Simulator::grabReleaseObjectUsingCrossHair(Magnum::Vector2i windowSize) {
```
aclegg3, line 883: We probably want to think more about an extensible Simulator API for gripping objects like this. I imagine this should live on the application/task side rather than in Simulator for now.

src/esp/sim/Simulator.cpp
```diff
@@ -809,5 +821,168 @@ void Simulator::setObjectLightSetup(int objectID,
...
+    Magnum::Vector3 refPosition = 
+        physicsManager_->getTranslation(refObjectID);
+    Magnum::Vector2 agentPos_xz(refPosition.x(), refPosition.z());
+
+    for(int objectID : physicsManager_->getExistingObjectIDs()) {
+      if (objectID != refObjectID) {
+        Magnum::Vector3 objectPos = physicsManager_->getTranslation(objectID);
+        Magnum::Vector2 objectPosXZ(objectPos.x(), objectPos.z());
+        
+        float d = Magnum::Vector2(agentPos_xz - objectPosXZ).length();
```
Skylion007, line 836: Wait, so the distance does not include y? This is bad considering we have datasets with multi-floor environments.

aclegg3: If we assume object COMs are within 0.5m in Y of the NavMesh, geodesic distance could be better 🤷‍♂️ . May be some caveats to this approach also.

Ram81: @Skylion007  one of the `findNearestObject` is not being used. It was only used for testing purpose. I'll remove the code for that.

build_js.sh
mosra: Related: https://github.com/facebookresearch/habitat-sim/issues/527

I'd say this would fix it but unfortunately have no way to check locally at the moment -- @Ram81 @Skylion007 can you try and confirm?

```patch
diff --git a/src/esp/bindings_js/CMakeLists.txt b/src/esp/bindings_js/CMakeLists.txt
index d9df462d..3da748a6 100644
--- a/src/esp/bindings_js/CMakeLists.txt
+++ b/src/esp/bindings_js/CMakeLists.txt
@@ -12,7 +12,12 @@ target_link_libraries(hsim_bindings
     sim
 )
 
-set_target_properties(hsim_bindings PROPERTIES LINK_FLAGS "--bind")
+set_target_properties(hsim_bindings PROPERTIES
+  LINK_FLAGS "--bind"
+  # Override the Magnum-set default which is in $<CONFIG>/bin or $<CONFIG>/lib
+  RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
 # copy JS/HTML/CSS resources for WebGL build
 set(resources

```

src/esp/bindings_js/bindings_js.cpp
```diff
@@ -92,10 +96,45 @@ EMSCRIPTEN_BINDINGS(habitat_sim_bindings_js) {
       .element(em::index<2>())
       .element(em::index<3>());
 
+  em::value_array<Magnum::Vector2i>("Vector2i")
+      .element(em::index<0>())
+      .element(em::index<1>());
+
+  em::class_<Magnum::Matrix4>("Matrix4")
```
Skylion007, line 103: @mosra Is this really the best way to bind Magnum's Matrices and Vectors? Surely there is a better way.

mosra: What do you mean, exactly? Can you elaborate? To me this looks like bindings done on an as-needed basis, and for that it looks fine.

For a general case of "expose everything possible" this could get templated by size and type, similarly to how Python bindings are done. But JavaScript has no operator overloads nor anything similar so any custom math code will always be extremely miserable.

Skylion007: I guess I was thinking of the expose everything possible template, but if you say we don't need it yet, I suppose we don't have to go down the root just yet.

build_js.sh
Ram81: @mosra tested on my local. `hsim_bindings` are getting updated in the `build_js/esp/bindings_js/`

src/esp/sim/Simulator.cpp
Ram81: @aclegg3 for moving `unproject` to RenderCamera I am facing issues using it on JS side in `grabReleaseObject` action. When returned object is a reference emscripten uses a copy constructor to instantiate the returned value. `SceneGraph` `SceneNode` and `RenderCamera` don't have a copy constructor and I am not able to expose bindings for those. What would be a better way to implement with the above constraints? Should I expose methods in Simulator to call `unproject`, that way I don't have to bind `RenderCamera` and `SceneGraph` on JS side.

build_js.sh
```diff
@@ -62,3 +62,7 @@ echo "Then open in browser:"
 echo "http://0.0.0.0:8000/build_js/utils/viewer/viewer.html?scene=skokloster-castle.glb"
 echo "Or:"
 echo "http://0.0.0.0:8000/build_js/esp/bindings_js/bindings.html?scene=skokloster-castle.glb"
+echo "To enable physics use:"
```
Skylion007, line 65: This should on be printed if `${BULLET}` is set. Otherwise, you can't enable physics, right?

src/esp/bindings_js/bindings_js.cpp
Skylion007: Once again, any reason this is a function instead of read_only property?

src/esp/bindings_js/index.js
```diff
@@ -26,13 +31,63 @@ function preload(url) {
...
+
 Module.preRun.push(() => {
   let config = {};
   config.scene = defaultScene;
   buildConfigFromURLParameters(config);
   window.config = config;
   const scene = config.scene;
   Module.scene = preload(scene);
+
+  var physicsConfigFile = window.config.defaultPhysConfig;
```
Skylion007, line 83: Do we reference the window explicitly anywhere else? Best to avoid referencing it since it can limit where the JS can be deployed (like to node).

build_js.sh
Ram81: Yes, the grab/release actions don't work when Bullet is not enabled

src/esp/sim/Simulator.cpp
```diff
@@ -18,6 +24,7 @@
 #include "esp/io/io.h"
 #include "esp/nav/PathFinder.h"
 #include "esp/physics/PhysicsManager.h"
+#include "esp/physics/bullet/BulletPhysicsManager.h"
```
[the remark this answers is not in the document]
Ram81: actually `BulletPhysicsManager.h` was not required explicitly. I added this when I was experimenting initially.

.circleci/config.yml
```diff
@@ -313,8 +313,6 @@ jobs:
               export EMSCRIPTEN=$HOME/emsdk/fastcomp/emscripten
               nvm use v11.9.0
               ./build_js.sh
```
Skylion007, line 315:
```suggestion
              ./build_js.sh --bullet
```

build_js.sh
Skylion007: @Ram81 In fact, it should probably print a warning to the effect it's not built with bullet.

Skylion007:
@Ram81 Once you add fix those last few issues, it would be great if you could add Javascript unit tests:

Example can be found here:
https://github.com/facebookresearch/habitat-sim/blob/eb68bc36a8c568ec0708fc6680b6e4d8fedc6782/src/esp/bindings_js/tests/viewer_demo.visual.test.js#L14

You should also test that these don't break when built with --bullet.

src/esp/bindings_js/index.js
Skylion007: AH nvm, that's our bad I just realized as we do set the window in the line before, disregard that comment. That should be added as a todo on our side.

Ram81:
Sure @Skylion007 I will add the tests

.circleci/config.yml
Ram81: Should we add steps for both `./build_js.sh` and `./build_js.sh --bullet` in circleci?

build_js.sh
Ram81: User will still be able to use the webGL setup in browser even when build without bullet. It will onlt disable grab release actions when not built with bullet. Should it still be a warning?

.circleci/config.yml
Skylion007: Absolutely @Ram81 as long we have tests that will run on both of them.

build_js.sh
Skylion007: I guess the lack of this message sort of acts as a warning...

Ram81:
@aclegg3 @Skylion007 @jturner65 I was wondering if the actions like grab and release from this PR is required or not? These action implementation don't seem generic. So do we want to extend the current webgl setup in habitat-sim to be same as cpp viewer? If yes, maybe I can remove grab/release actions and just add the add/remove objects actions and other js bindings as a part of this PR.

src/esp/bindings_js/bindings_js.cpp
```diff
@@ -237,5 +302,32 @@ EMSCRIPTEN_BINDINGS(habitat_sim_bindings_js) {
...
+      .function("syncGrippedObject", &Simulator::syncGrippedObject)
+      .function("initOrUpdateCrossHairNode",
+                &Simulator::initOrUpdateCrossHairNode)
+      .function("recomputeNavMesh", &Simulator::recomputeNavMesh)
+      .function("findNearestObjectUnderCrosshair",
+                &Simulator::findNearestObjectUnderCrosshair)
+      .function("unproject", &Simulator::unproject)
+      .function("getAgentTransformation", &Simulator::getAgentTransformation)
+      .function("getAgentAbsoluteTranslation",
+                &Simulator::getAgentAbsoluteTranslation);
```
jturner65, line 332: All the getter/setters in this section really ought to be properties of the format 

`.property("x", &MyClass::getX, &MyClass::setX)`

jturner65 (commented):
Comment re:bindings.

src/esp/bindings_js/bindings_js.cpp
Ram81: @jturner65 we can only use properties for getter setters which are operating on a variable in simulator class right? `setRotation` `getRotation`,etc can't be changed to properties right?

erikwijmans: Nope, you can use them for anything.  They are just syntactic sugar.

eundersander:
Closing this old PR. We can always re-open if there's renewed interest.

PR #706, opened 2020-07-28
--[WIP] Expand python utility functions for tutorials

jturner65:
## Motivation and Context
This PR moves some common functions to the vis_utils library to minimize the footprint of non-tutorial-specific functionality in the colabs themselves.  Once this file has been approved, remove_all_objects, simulate, make_cfg and make_video_cv2 can be removed from the colabs and accessed through the libraries.

The protocol for calling make_video_cv2 has changed slightly, to make multi-image videos more customizable in configuration, and the protocol for make_cfg has changed slightly to enable more easily configured sensor types, locations and orientations relative to the agent.

Existing colabs can experimentally use this protocol before migrating to this mechanism, so this is not going to break anything, but some of the functionality (with make_video_cv2) needs to be tested.

This is a WIP because it has not yet been thoroughly tested. 

## How Has This Been Tested
Existing python tests pass, but the modified functionality has not yet been tested (hence WIP)

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/utils/viz_utils.py
+++ b/habitat_sim/utils/viz_utils.py
@@ -112,7 +112,7 @@ def make_cfg(settings):
     return habitat_sim.Configuration(backend_cfg, [agent_cfg])
 
 
-def make_video_cv2(
+def make_video(
     observations,
     primary_pov,
     output_path,
--- a/habitat_sim/utils/viz_utils.py
+++ b/habitat_sim/utils/viz_utils.py
@@ -1,10 +1,11 @@
 import base64
-import cv2
 import io
-import numpy as np
 import os
 import subprocess
 import sys
+
+import cv2
+import numpy as np
 
 import habitat_sim
 
--- a/habitat_sim/utils/viz_utils.py
+++ b/habitat_sim/utils/viz_utils.py
@@ -1,8 +1,12 @@
 import base64
+import cv2
 import io
+import numpy as np
 import os
 import subprocess
 import sys
+
+import habitat_sim
 
 if "google.colab" in sys.modules:
     os.environ["IMAGEIO_FFMPEG_EXE"] = "/usr/bin/ffmpeg"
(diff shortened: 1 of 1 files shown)


habitat_sim/utils/viz_utils.py
```diff
@@ -26,6 +31,260 @@ def is_notebook():
...
+        sensor_specs.append(sensor_spec)
+
+    # agent configurations
+    agent_cfg = habitat_sim.agent.AgentConfiguration()
+    agent_cfg.sensor_specifications = sensor_specs
+
+    return habitat_sim.Configuration(backend_cfg, [agent_cfg])
+
+
+def make_video_cv2(
```
Skylion007, line 115: This function should be renamed to be more descriptive. Also, might want to see if we can drop the OpenCV requirement since it's not really necessary anymore. Something like embed_observations_into_video?

jturner65: Do we not need to use cv2 to resize and all that?

habitat_sim/utils/viz_utils.py
```diff
@@ -26,6 +31,260 @@ def is_notebook():
...
+
+    Returns
+    -------
+    rgb_d_im : 2d numpy array
+        Clipped depth image data
+
+    """
+    d_im = np.clip(depth_image, 0, clip_max)
+    d_im /= clip_max
+    rgb_d_im = cv2.cvtColor((d_im * 255).astype(np.uint8), cv2.COLOR_GRAY2RGB)
```
Skylion007, line 284: This should be pure OpenCV or pure Numpy. Also you are combing the color conversion and RGB generation into the same function, but it's not obvious except for the return type.

habitat_sim/utils/viz_utils.py
Skylion007, line 115: Type Hints should be used more aggressively in this file.

habitat_sim/utils/viz_utils.py
```diff
@@ -94,3 +353,83 @@ def display_video(video_file: str, height: int = 400):
...
+
+    """
+    agent_state = habitat_sim.AgentState()
+    agent_state.position = pos
+    agent_state.rotation = rot
+    agent = sim.initialize_agent(0, agent_state)
+    return agent.scene_node.transformation_matrix()
+
+
+def remove_all_objects(sim):
```
Skylion007, line 383: This is general enough we may want to add a helper to Simulator.cpp, thoughts @aclegg3 ?

habitat_sim/utils/viz_utils.py
```diff
@@ -94,3 +353,83 @@ def display_video(video_file: str, height: int = 400):
...
+    -------
+    None.
+
+    """
+
+    for id in sim.get_existing_object_ids():
+        sim.remove_object(id)
+
+
+def simulate(sim, dt=1.0, get_frames=True):
```
Skylion007, line 402: Function name should be more descriptive, simulate_physics at the very least or visualize_physics?

habitat_sim/utils/viz_utils.py
```diff
@@ -94,3 +353,83 @@ def display_video(video_file: str, height: int = 400):
...
+
+    Returns
+    -------
+    observations : list
+        list of 2d numpy array of ints or floats representing observation 
+        per sensor per step.
+    """
+
+    # simulate dt seconds at 60Hz to the nearest fixed timestep
+    print("Simulating " + str(dt) + " world seconds.")
```
Skylion007, line 427: There should be a silent boolean to avoid printing output.

habitat_sim/utils/viz_utils.py
```diff
@@ -94,3 +353,83 @@ def display_video(video_file: str, height: int = 400):
...
+    observations = []
+    start_time = sim.get_world_time()
+    while sim.get_world_time() < start_time + dt:
+        sim.step_physics(1.0 / 60.0)
+        if get_frames:
+            observations.append(sim.get_sensor_observations())
```
Skylion007, line 433: What's the impact of actually storing all the images in a lazy list like this, I imagine it's negligible. You may want to construct an np.array of observations and place them in there so they are stored contiguously for downstream tasks.

habitat_sim/utils/viz_utils.py
jturner65: I tried to keep naming the same as the methods that have been used in existing tutorials, so that migration could occur more easily.  Of course they can be renamed, just the reason why they haven't been :).

habitat_sim/utils/viz_utils.py
jturner65: This is how things are currently -shrug-.  Wanted to migrate things out of the colabs before I went and refactored everything :).

habitat_sim/utils/viz_utils.py
```diff
@@ -26,6 +31,260 @@ def is_notebook():
         return True
 
 
+def make_cfg(settings):
```
Skylion007, line 34: Can the logic from https://github.com/facebookresearch/habitat-sim/blob/9c96776cf6a783d036f7d051d648864dd6f20a13/examples/settings.py#L38 . Maybe those utility functions should be merged into Habitat-Sim proper.

Skylion007: Does this function really deserve to be in viz_utils? Should it be renamed to something else?

habitat_sim/utils/viz_utils.py
Skylion007: ImageIO probably has some resize functionality and PIL definitely does.

habitat_sim/utils/viz_utils.py
Skylion007: For all the tutorials that have been committed, you can easily refactor and test.

habitat_sim/utils/viz_utils.py
jturner65: Are the docstrings insufficient in this regard, or should the hints be used as well?

Skylion007: IDEs can actually use type hints so yes both should be used when possible and helpful.

Skylion007, line 115:
```suggestion
def make_video(
```

habitat_sim/utils/viz_utils.py
```diff
@@ -94,3 +353,83 @@ def display_video(video_file: str, height: int = 400):
         else:
             opener = "open" if sys.platform == "darwin" else "xdg-open"
             subprocess.call([opener, video_file])
+
+
+def place_agent(sim, pos=[0.0, 0.0, 0.0], rot=np.quaternion(-1, 0, 0, 0)):
```
Skylion007, line 358: These last two don't feel like they belong in viz_utils. Otherwise looks good!

Skylion007:
Minus that minor complaint about place_agents and remove_objects. Looks good!

habitat_sim/utils/viz_utils.py
```diff
@@ -26,6 +31,260 @@ def is_notebook():
...
+        Height by Width of camera for image. Default is [540, 720]
+
+    open_vid : boolean, optional
+        Whether to open video upon creation.  Default is True
+
+    multi_obs : boolean, optional
+        Whether to embed images from POVs listed in pov_list.
+        Defaults to False
+
+    pov_list : List of tuples, optional
```
Skylion007, line 152: I personally really like this function call, but @aclegg3 suggested maybe allowing more args to be passed to the sub function. Particularly the video border size and video size.

habitat_sim/utils/viz_utils.py
```diff
@@ -26,6 +31,260 @@ def is_notebook():
...
+            embed_image_data_list = [
+                (ob[embed_image_list[i][0]], embed_image_list[i][1])
+                for i in range(0, len(embed_image_list))
+            ]
+
+            res_image = build_multi_obs_image(
+                ob[primary_pov], embed_image_data_list, thumb_size, outline_frame,
+            )
+            if res_image.shape[:2] != videodims:
+                res_image = cv2.resize(
```
Skylion007, line 190: I know we are running a bit late on this, so it's low pri, but we can refactor this to PIL/numpy and remove the CV2 dependency entirely.

jturner65: Sounds good.  I'll address these issues as soon as I can.

jturner65:
This PR is mostly obviated by #734 so I'll close it.  We can revisit the configuration mechanism in this at a later date if necessary.

PR #718, opened 2020-08-05
Topdown map matplotlib display utility

aclegg3:
## Motivation and Context
Provide a convenience utility to display a topdown map via matplotlib.

## How Has This Been Tested

Local testing in Colab. Example output:

Default:
image

With key-points and title:
image

With trajectory and title:
image

## Types of changes

## Checklist


habitat_sim/utils/viz_utils.py
```diff
@@ -94,3 +96,55 @@ def display_video(video_file: str, height: int = 400):
         else:
             opener = "open" if sys.platform == "darwin" else "xdg-open"
             subprocess.call([opener, video_file])
+
+
+def display_topdown_map(
+    topdown_map, plot_title=None, trajectory=None, key_points=None, marker_alpha=0.8
```
Skylion007, line 102: May want some type annotations here. Otherwise looks good!

erikwijmans:
We should make sure to not duplicate the functionality here: https://github.com/facebookresearch/habitat-api/blob/master/habitat/utils/visualizations/maps.py too much

aclegg3:
> We should make sure to not duplicate the functionality here: https://github.com/facebookresearch/habitat-api/blob/master/habitat/utils/visualizations/maps.py too much

Thanks for pointing this out. We've been wanting to display topdown map visualizations on the sim side in a simple way using the existing C++ Pathfinder topdown map implementation. This module looks much more configurable, but would need to be converted for use in Habitat-sim.

Skylion007:
> > We should make sure to not duplicate the functionality here: https://github.com/facebookresearch/habitat-api/blob/master/habitat/utils/visualizations/maps.py too much
> 
> Thanks for pointing this out. We've been wanting to display topdown map visualizations on the sim side in a simple way using the existing C++ Pathfinder topdown map implementation. This module looks much more configurable, but would need to be converted for use in Habitat-sim.

This is a common theme and will lead to duplicate in Habitat-API that will only grow over time, no? We can move things out of habitat-sim to Habitat-API, but then we may need those utilities later without wanting to wrap around Habitat-API. We should keep this in mind for future Habitat-API design.

aclegg3:
Using the API utilities.

PR #720, opened 2020-08-05
--Plug possible loophole in mesh handle/type setting in json;

jturner65:
Reading json handles and json meshtypes improved -if a mesh handle is changed in json, but no mesh type is specified, rudimentary mesh type setting is performed; Unused methods are removed;

## Motivation and Context
Possibility of changing mesh handle to different type of mesh in json, without specifying new type.  This puts hooks into process that allows for semi-intelligent mesh type setting if needed.  Also a few methods have been removed that were not being used from attributes handlers.

## How Has This Been Tested
All C++ and python tests have passed.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/managers/SceneAttributesManager.cpp
+++ b/src/esp/assets/managers/SceneAttributesManager.cpp
@@ -18,15 +18,6 @@ namespace assets {
 
 namespace managers {
 
-const std::map<std::string, esp::assets::AssetType>
-    SceneAttributesManager::AssetTypeNamesMap = {
-        {"mp3d", AssetType::MP3D_MESH},
-        {"navmesh", AssetType::NAVMESH},
-        {"ptex", AssetType::FRL_PTEX_MESH},
-        {"semantic", AssetType::INSTANCE_MESH},
-        {"suncg", AssetType::SUNCG_SCENE},
-};
-
 SceneAttributesManager::SceneAttributesManager(
     assets::ResourceManager& resourceManager,
     ObjectAttributesManager::ptr objectAttributesMgr,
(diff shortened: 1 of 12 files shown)


src/esp/assets/managers/SceneAttributesManager.cpp
```diff
@@ -339,6 +300,45 @@ PhysicsSceneAttributes::ptr SceneAttributesManager::initNewAttribsInternal(
...
+  fwd = fwd1;
+  if (endsWith(fileName, "_semantic.ply")) {
+    meshTypeSetter(static_cast<int>(AssetType::INSTANCE_MESH));
+  } else if (endsWith(fileName, "mesh.ply")) {
+    meshTypeSetter(static_cast<int>(AssetType::FRL_PTEX_MESH));
+    up = up2;
+    fwd = fwd2;
+  } else if (endsWith(fileName, "house.json")) {
+    meshTypeSetter(static_cast<int>(AssetType::SUNCG_SCENE));
+  } else if (endsWith(fileName, ".glb")) {
```
eundersander, line 326: Why are we associating glb files with Matterport3D here? Is this a backward-compatibility thing?

What's the recommended way for a user to specify her desired orientUp and orientFront when loading a glb?

dhruvbatra: > Why are we associating glb files with Matterport3D here? Is this a backward-compatibility thing?

My understanding is that this is a legacy thing. We wrote support for MP3D by looking for glb files, for Replica by looking for ply files, for the dataset that shall not be named by looking for house.json files. 

CC: @msavva

aclegg3: Correct, this is the backward compatibility for the AssetInfo::fromPath() function. The upcoming tutorial will demonstrate the preferred programmatic and file-based configuration options for scenes.

msavva: That's right, this branching logic was hardcoded assuming particular datasets are only loaded from particular file types.  We really ought to clean this up and start to rely on correctly populating an `AssetInfo` to define asset type, coordinate frame conventions etc.

jturner65: Just a note, this currently is hardcoded only as a fallback if the user did not specify other values in their provided json config file.

jturner65: > What's the recommended way for a user to specify her desired orientUp and orientFront when loading a glb?

The scene importer allows for these fields to be set in the scene-descriptor json.  These only consume the defaults if the user has not set anything.

PR #729, opened 2020-08-11
[WIP] [URDF] Parser and Model

aclegg3:
## Motivation and Context

First PR toward URDF and articulated object support. Adaptation of the Bullet examples URDF Parser and Model. Since the parsing exists independent of consumption by an importer, seems like a good first chunk.

Uses tinyxml for the file parsing step.

## How Has This Been Tested

Added an iiwa URDF test asset and an IOTest entry for parsing the model.

## Types of changes

## Checklist


aclegg3:
Looks like clang-tidy hates tinyxml. Could refactor to clean this up or migrate to another parser.

Skylion007:
Isn't there a .clang-tidy-ignore file that exists exactly for this reason @aclegg3 ?

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,314 @@
...
+////////////////////////////////////
+
+namespace esp {
+namespace io {
+
+struct UrdfMaterialColor {
+  Magnum::Color4 m_rgbaColor;
+  Magnum::Color3 m_specularColor;
+  UrdfMaterialColor()
+      : m_rgbaColor(0.8, 0.8, 0.8, 1), m_specularColor(0.4, 0.4, 0.4) {}
```
eundersander, line 31: I don't have much context with URDF so I'm phrasing this as a question: Is an unspecified material or color very common, or is it more of an error case?

If it's more of an error case, it might be better to use an obnoxious color like (1, 0, 1) magenta, so when users complain about their model being pink, we know right away the problem.

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,314 @@
...
+
+struct UrdfMaterial {
+  std::string m_name;
+  std::string m_textureFilename;
+  UrdfMaterialColor m_matColor;
+
+  UrdfMaterial() {}
+};
+
+enum UrdfJointTypes {
```
eundersander, line 42: Consider `enum class` for better type safety and so that you don't have to prefix "URDF" to all these.

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,314 @@
...
+  URDF_GEOM_MESH,
+  URDF_GEOM_PLANE,
+  URDF_GEOM_CAPSULE,  // non-standard URDF
+  URDF_GEOM_UNKNOWN,
+};
+
+struct UrdfGeometry {
+  UrdfGeomTypes m_type;
+
+  double m_sphereRadius;
```
eundersander, line 66: I'm too lazy to scan this PR to check that this gets properly initialized everywhere :), but in general, PODs like doubles that don't get default initialized make me nervous.

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,314 @@
...
+    if (m_linkIndicesToNames.count(linkIndex)) {
+      return getLink(m_linkIndicesToNames.at(linkIndex))->m_parentJoint;
+    }
+    return nullptr;
+  }
+
+  UrdfModel() : m_overrideFixedBase(false) {}
+};
+
+class URDFParser {
```
eundersander, line 267: How about a `URDF` namespace, so you don't have to prepend `URDF`/`Urdf` to all the types in this file?

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,314 @@
...
+  // datastructures
+  UrdfModel m_urdfModel;
```
eundersander, line 269: It seems weird that URDFParser owns the model object. This means the parser must persist for as long as you want to use the object. I would expect to throw away a parser object after parsing is finished.

Why not just a static parse method that returns the model?

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,314 @@
...
+  bool validateMeshFile(std::string& filename);
+
+ public:
+  URDFParser(){};
+
+  // parse a loaded URDF string into relevant general data structures
+  // return false if the string is not a valid urdf or other error causes abort
+  bool parseURDF(const std::string& meshFilename);
+
+  void setGlobalScaling(float scaling) { m_urdfScaling = scaling; }
```
eundersander, line 305: This API is a little confusing. If I call `parseURDF`, then `setGlobalScaling`, then `getModel`, do I get a scaled model?

erikwijmans:
> Looks like clang-tidy hates tinyxml. Could refactor to clean this up or migrate to another parser.

If you move tinyxml into src/deps, clang-tidy won't look at it

aclegg3:
More recent version OTW.

PR #730, opened 2020-08-12
[WIP] Attributes RST Tutorial

jturner65:
## Motivation and Context
This document is to serve the purpose of exposing the json tags that are supported in the various physics manager, scene and object configuration files.  It should also be expanded to present tutorial python code for attributes modification and consumption.

EDIT(JT) : Note, this is still a WIP, and won't be ready to merge until Scene configurations have been implemented.

## How Has This Been Tested
Compiles locally within docs 

## Types of changes

## Checklist

### DIFF
--- a/docs/pages/attributes.rst
+++ b/docs/pages/attributes.rst
@@ -1,0 +1,273 @@
+:ref-prefix:
+    habitat_sim.sim
+    habitat_sim.attributes
+    habitat_sim.attributes_managers
+
+Using JSON Files to configure Attributes
+########################################
+
+:summary: Attributes templates are used in Habitat to configure the simulation world and specify characteristics for scenes and objects. This document describes the appropriate tags to be used when authoring JSON files to properly customize these various attributes templates.
+
+.. contents::
+    :class: m-block m-default
+
+Attributes templates provide a mechanism by which the various constructions in Habitat can be customized and built with user-specified characteristics, either at program start or on the fly.
+
+`Physics Manager Attributes`_
+=============================
+Physics Manager Attributes templates describe quantities pertinent to building the simulation world.  Any source configuration JSON files used to build these attributes should be named in the following format : 
+
+ 	<worldname>.physics_config.json
+
+
+Below are the supported JSON tags for Physics Manager Attributes templates, and their meanings.
+
+"physics simulator"
+	- string
+	- What physics engine should be used for dynamics simulation.  Currently supports "bullet" for Bullet physics simulation, and "none", meaning kinematic motion is to be used.	
+"gravity"  
+	- 3-vector
+	- Gravity to use for physical modeling. 
+"timestep" 
+	- double
+	- The timestep to use for forward simulation. 
+"friction coefficient"
+	- double
+	- The coefficient of friction. 
+"restitution coefficient"
+	- double
+	- The coefficient of restitution. 
+"rigid object paths"
+	- list of strings
+	- A list of locations to query for supported object files that should be available to be loaded into the world.
+
+`Scene Attributes`_
+===================
+A scene in Habitat-Sim contains a stage and objects.  Scene Attributes templates hold relelvant information describing the names of the stages and objects contained within the scene, their locations (when applicable) and physics quantities that should override defaults.  Any source configuration files used to build these attributes should be named in the format : 
+
+ 	<scenename>.scene_config.json
+	
+Scene Stage and Object Configurations
+-------------------------------------
+
+Below are the stage and object
+
+"stage config"
+	- JSON object
+	- JSON construct for the stage to use for this scene.  Each element has the following configuration : 
+		- "filename"
+			- string
+			- name of stage configuration file describing the stage to use for this scene.
+		- "location"
+			- 3-vector
+			- location within scene to place the origin of the stage.  If not provided, this defaults to [0,0,0]
+
+"object configs"
+	- list of JSON objects
+	- A list of JSON constructs for the objects placed in the scene.  Each element of the list has the following configuration :
+		- "filename" 
+			- string
+			- name of object configuration file describing object to be placed within the scene.
+		- "location"
+			- 3-vector
+			- location within the scene to place the object. If not provided, this defaults to [0,0,0]
+		- "orientation"
+			- 3-vector (euler angles) or 4-vector (quaternion)
+			- orientation of object within scene. If not provided, uses object's default orientation.
+		- "relative to agent"
+			- boolean
+			- whether location is relative to agent location.  Defaults to false.	
+"nav mesh"
+	- string
+	- The name of the file describing the NavMesh for this scene's stage and objects.
+
+Scene Physics and Object-related Parameters
+-------------------------------------------
+
+Below are scene-specific physical and object-related quantities.  These values will override similarly-named values specified in the Physics Manager Attributes, and in turn will be overridden, where appropriate, by Stage and Object attributes files.
+
+"scale"
+	- 3-vector
+	- The default scale to be used for the scene.
+"gravity"
+	- 3-vector
+	- Gravity to use in scene for physical modeling.
+"margin"
+	- double
+	- Distance margin for collision calculations.
+"friction coefficient"
+	- double
+	- The coefficient of friction.
+"restitution coefficient"
+	- double
+	- The coefficient of restitution.
+"units to meters"
+	- double
+	- The conversion of given units to meters.
+"requires lighting"
+	- boolean
+	- Whether or not the scene should use lighting.
+"lighting setup"
+	- string
+	- The Habitat lighting setup to use if lighting is required.
+
+
+`Stage Attributes`_
+===================
+A stage in Habitat-Sim is a static object consisting of background scenery wherein an agent acts.  Stage Attributes templates hold relevant information describing a stage's mesh, geometry and physical properties.  Any source configuration files used to build these attributes should be named in the format : 
+
+ 	<stagename>.stage_config.json
+	
+Stage Mesh Handles And Types
+----------------------------	
+
+Below are the handles and descriptors for various mesh assets used by a stage.
+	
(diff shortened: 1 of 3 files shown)


docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+	
+"gravity"  
+	- 3-vector
+	- Gravity to use in scene for physical modeling.
+
+"timestep" 
+	- double
+	- The timestep to use for forward simulation.
+ 
+"friction coefficient"
```
eundersander, line 36: I'm surprised to see this at the physics-world level. Is this just a default in case a physics object doesn't specify it?

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+Scene Mesh Handles And Types
+----------------------------	
+
+Below are the handles and descriptors for various mesh assets used by a scene.
+	
+Mesh types are included to support future format flexibility. In the JSON they are limited to string keys used by :ref:`AbstractPhysicsAttributes::AssetTypeNamesMap`, which map to values in :ref:`AssetType`.  Any mesh type specification used in JSON but not found in :ref:`AbstractPhysicsAttributes::AssetTypeNamesMap` results in that mesh type being set to :ref:`AssetType::UNKNOWN`.  JSON-Specified mesh types will override any default settings, such as those that might be derived from file name extensions.
+
+"render mesh" 
+	- string
+	- The name of the file describing the render mesh to be used by the scene.
```
eundersander, line 64: We talked about terminology around "scene" and "background/static" yesterday. I'd still like us to move to terminology where the "scene" is everything we simulate and render, and then we have a separate term like "background" to refer to this static object/mesh.

eundersander: It might be as simple as renaming this field to "background render mesh" or "static render mesh".

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+Below are the handles and descriptors for various mesh assets used by a scene.
+	
+Mesh types are included to support future format flexibility. In the JSON they are limited to string keys used by :ref:`AbstractPhysicsAttributes::AssetTypeNamesMap`, which map to values in :ref:`AssetType`.  Any mesh type specification used in JSON but not found in :ref:`AbstractPhysicsAttributes::AssetTypeNamesMap` results in that mesh type being set to :ref:`AssetType::UNKNOWN`.  JSON-Specified mesh types will override any default settings, such as those that might be derived from file name extensions.
+
+"render mesh" 
+	- string
+	- The name of the file describing the render mesh to be used by the scene. 
+"render mesh type"
+	- string
+	- The type of the render mesh. If not specified, value will be derived based on given **"render mesh"** file extension.
```
eundersander, line 67: What are my options here? What are the types?

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+"collision mesh type"
+	- string
+	- The type of the collision mesh. If not specified, value will be derived based on given **"collision mesh"** file extension.  Currently only "mp3d" is supported, other values default to :ref:`AssetType::UNKNOWN` 
+"semantic mesh"
+	- string
+	- The name of the file describing the scene's semantic mesh. 
+"semantic mesh type"
+	- string
+	- The type of the render mesh. If not specified, value will be derived based on given **"semantic mesh"** file extension.  Currently only "mp3d" is supported, other values default to :ref:`AssetType::INSTANCE_MESH`
+"house filename"
```
eundersander, line 80: "house" seems oddly specific. Is there a more general term?

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+	- The name of the file containing semantic type maps and hierarchy. 
+"nav mesh"
+	- string
+	- The name of the file describing the NavMesh. 
+
+
+Scene Frame and Origin
+----------------------
+
+The tags below are used to build a coordinate frame for the scene, and will override any default values set based on render mesh file name/extension.  If either **"up"** or **"front"** are specified, both must be provided and they must be orthogonal.
```
eundersander, line 91: Is the user allowed to choose the handedness, or do we want to assert on a certain handedness here?

Beware, naively changing handedness can sometimes break things like backface-culling.

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+	- Describes the **origin** of the scene, for alignment purposes.
+	 
+Scene Physics and Object-related Parameters
+-------------------------------------------
+
+Below are scene-specific physical and object-related quantities.  These values will override similarly-named values specified in the Physics Manager Attributes.
+
+"scale"
+	- 3-vector
+	- The default scale to be used for the scene.
```
eundersander, line 110: What does this mean? Does is the scale the background render mesh based on the value here?

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+"gravity"
+	- 3-vector
+	- Gravity to use in scene for physical modeling.
+"margin"
+	- double
+	- Distance margin for collision calculations.
```
eundersander, line 116: How should a user choose this value? This is a runtime-performance/accuracy tradeoff in bullet, right?

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+	- The name of the file describing the collision mesh to be used by the object. 
+"collision mesh type"
+	- string
+	- The type of the collision mesh. If not specified, defaults to :ref:`AssetType::UNKNOWN`.
+	
+
+Object Frame and Origin
+-----------------------
+
+The tags below are used to build a coordinate frame for the object, and will override any default values set based on render mesh file name/extension.  If either **"up"** or **"front"** are specified, both must be provided and they must be orthogonal.  The object's COM is used as its origin.
```
eundersander, line 169: Beware, we're advising Appen to deliver objects with the origin placed at the COM projected down to ground height. In other words, we're going to have some objects soon where the origin is not the COM.

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+	- The coefficient of friction.
+"restitution coefficient"
+	- double
+	- The coefficient of restitution.
+"units to meters"
+	- double
+	- The conversion of given units to meters.
+"requires lighting"
+	- boolean
+	- Whether or not the scene should use lighting.
```
eundersander, line 201: "scene"? typo?

docs/pages/attributes.rst
```diff
@@ -0,0 +1,215 @@
...
+  - The values of the diagonal of the inertia matrix for the object.
+"COM"
+  - 3-vector
+  - The center of mass for the object.  If this is not specified in JSON, it will be derived from the object's bounding box in Habitat.
+"use bounding box for collision"
+  - boolean
+  - Whether collision calculations should be based on object's bounding box geometry.
+"join collision meshes"
+  - boolean
+  - Whether collision mesh assets should be joined into a single unified collision object.
```
eundersander, line 213: How should a user choose this? Is there any reason *not* to join collision meshes?

docs/pages/attributes.rst
jturner65: As i refactored the existing code to fit this paradigm, I kept the existing labels and names, with the expectation that the names might need to be changed eventually. Hence the odd label - I wanted the folks who were familiar with the old configuration to suggest further refactors/renaming.

docs/pages/attributes.rst
jturner65: Handedness seems to be defaulted in the system to be right-handed, and the esp::geo::CoordinateFrame that consumes these bases does not include a 3rd basis vector, so it doesn't seem that it can be changed.

docs/pages/attributes.rst
jturner65: I am pretty sure that is correct. @aclegg3 ?

docs/pages/attributes.rst
jturner65: > I left a bunch of minor comments just as feedback from a user perspective. Don't feel obligated to respond or make changes. Thanks for writing good documentation!

Thank you for being thorough Eric, I am glad you looked things over, and sorry I didn't include you as a reviewer. :)

jturner65:
> I left a bunch of minor comments just as feedback from a user perspective. Don't feel obligated to respond or make changes. Thanks for writing good documentation!

Thank you for being thorough Eric, I am glad you looked things over, and sorry I didn't include you as a reviewer. :)

docs/pages/attributes.rst
eundersander: I'm curious if it will also scale all the objects you add to the scene, or just the background.

jturner65: Currently each object's scale is specified independently of the scene.  

I think it would be a reasonable addition to have a boolean "override scene scale" set at the object level (coupled with the default behavior being that every object inherits the scene's scale), or else "all objects inherit scene scale", set at the static scene level.  

Depends on the desired default/"99.4%" behavior we wish to have.

docs/pages/attributes.rst
jturner65: The COM is used as the origin to build the object's Coordinate frame (frame requires "up", "front" and "origin" for rotation calcs) - this origin field is specified above this section in the static scene discussion, so I mention this here to address the potential question "where is the origin?". 

The object COM can differ from the mesh COM, if the user wishes it - it is either specified by the user in json, or else computed when an object is built.

docs/pages/attributes.rst
jturner65: This is a field that I inherited and so I am not certain what it's purpose is, but, just shooting in the dark if there are multiple meshes used to construct an object then joining them _may_ be less efficient for collision calcs - many small bounding boxes vs a single large one.

jturner65:
This has been pre-empted by [this PR](https://github.com/facebookresearch/habitat-sim/pull/769)

PR #738, opened 2020-08-14
To fix broken links in docs excluded std lib to from linking pipeline

mathfac:
## Motivation and Context
We have around 100+ broken links from the doc page: https://aihabitat.org/docs/habitat-cpp/namespacestd.html.
To fix broken links in docs excluded std lib to from linking pipeline.

## How Has This Been Tested
Check docs build in CI.

## Types of changes


mathfac:
Checked, std docs aren't generated anymore.
Screen Shot 2020-08-14 at 6 12 17 PM

PR #740, opened 2020-08-16
Update to the new MaterialData APIs in Magnum

mosra:
## Motivation and Context

Updates to the new [Trade::MaterialData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1MaterialData.html) APIs in Magnum, which provide:

- generic key/value store for arbitrary material properties including strings and pointers
- generic material layer support for PBR clearcoat/thinfilm/sheen... properties as well as custom material layering
- convenience APIs for common Flat, Phong and PBR material properties
- import of glTF PBR specular/glossiness properties via [Trade::PbrSpecularGlossinessMaterialData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PbrSpecularGlossinessMaterialData.html)
- import of glTF PBR metallic/roughness properties via [Trade::PbrMetallicRoughnessMaterialData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PbrMetallicRoughnessMaterialData.html)
- import of unlit glTF materials via [Trade::FlatMaterialData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1FlatMaterialData.html)
- import of glTF PBR clearcoat layer properties via [Trade::PbrClearCoatMaterialData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PbrClearCoatMaterialData.html)

For Assimp, the behavior is currently the same as before, as PBR support in that library is very ad-hoc, inconsistent and rather buggy (compared to `TinyGltfImporter` basically none of the glTF PBR properties are imported as they should and each will need a workaround to make the files open correctly). This will be done later, as I don't think it's *that* necessary -- you'll be most probably using glTF files for PBR anyway.

I did just the minimal port to make things compile and avoid using deprecated APIs. I especially didn't touch the existing `MaterialData` structure in Habitat at all -- since @bigbike is going to work on the PBR shader anyway, I didn't think any temporary throwaway changes for the Phong workflow would make sense.

## How Has This Been Tested

Through extensive tests in Magnum itself. There *might* be slight rendering differences due to minor changes in Assimp material import (different default shininess, possibly). I'll check those if the CI complains.

## Types of changes

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1113,10 +1113,10 @@ void ResourceManager::loadMaterials(Importer& importer,
     // TODO:
     // it seems we have a way to just load the material once in this case,
     // as long as the materialName includes the full path to the material
-    std::unique_ptr<Magnum::Trade::AbstractMaterialData> materialData =
+    Cr::Containers::Optional<Mn::Trade::MaterialData> materialData =
         importer.material(iMaterial);
     if (!materialData ||
-        materialData->type() != Magnum::Trade::MaterialType::Phong) {
+        !(materialData->types() & Magnum::Trade::MaterialType::Phong)) {
       LOG(ERROR) << "Cannot load material, skipping";
       continue;
     }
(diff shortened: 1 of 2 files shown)


mosra:
Yep, checked on a bunch of OBJ and glTF files (Habitat scenes or not) and everything looks as expected.

bigbike:
In description, the links are misplaced.

import of glTF PBR metallic/roughness properties via Trade::PbrSpecularGlossinessMaterialData
import of glTF PBR specular/glossiness properties via Trade::PbrMetallicRoughnessMaterialData

bigbike:
@aclegg3 : Would you test it with the physical scene enabled? See if the dynamic objects can be loaded and rendered correctly? I do not have the dynamic objects. 

Also, the bounding boxes, the primitives. Simply hope the flat shading is still working as expected.

mosra:
> Simply hope the flat shading is still working as expected.

Yes, it is. The recommended way for new code is to retrieve flat material properties is using the `FlatMaterialData`, and until everyone switches to that there's [backwards compatibility in place](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1TinyGltfImporter.html#Trade-TinyGltfImporter-behavior-materials).

PR #744, opened 2020-08-18
-Allow for expanded 'read-only' setting for Attributes Templates.

jturner65:
Certain attributes templates are expected to always be around (such as default primitive asset attribute templates and the object templates built from them.  Furthermore, users may wish to set templates to be locked to prevent accidental deletion, or to unlock them once they decide to delete them.  This PR supports this functionality.

## How Has This Been Tested
Existing c++ and python tests passed.
EDIT (JT) : Tests have been added in c++ and python to verify functionality.

## Types of changes

## Checklist

### DIFF
--- a/tests/test_attributes_managers.py
+++ b/tests/test_attributes_managers.py
@@ -56,20 +56,66 @@ def perform_general_tests(attr_mgr, search_string):
     # same object
     assert template0.get_string("test_key") != template2.get_string("test_key")
 
-    # add new template
-    attr_mgr.register_template(template0, "new_template_0")
+    # add new template with specified handle
+    new_template_handle = "new_template_0"
+
+    attr_mgr.register_template(template0, new_template_handle)
 
     # register new template and verify size is greater than original
     curr_num_templates = attr_mgr.get_num_templates()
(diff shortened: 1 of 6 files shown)


tests/test_attributes_managers.py
```diff
@@ -56,20 +56,66 @@ def perform_general_tests(attr_mgr, search_string):
...
+        new_iter_handle = new_handle_stub + str(i)
+        tmplt_id = attr_mgr.register_template(template3, new_iter_handle)
+        assert tmplt_id != -1
+
+    # lock all added templates
+    locked_template_handles = attr_mgr.set_lock_by_substring(
+        True, new_handle_stub, True
+    )
+    # verify that the number of added and locked templates are equal
+    assert num_to_add == len(locked_template_handles)
```
aclegg3, line 101: Also test the `get_user_locked_handles ` here? Would be a good spot.

jturner65: That would not be appropriate, since locked_template_handles are only those that were locked in that particular call to set_lock_by_substring, while get_user_locked_handles are -all- the locked handles.

PR #745, opened 2020-08-19
Object picking for replica mesh

bigbike:
## Motivation and Context
As titled.
A very useful functionality for the future development.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/PTexMeshData.cpp
+++ b/src/esp/assets/PTexMeshData.cpp
@@ -91,6 +91,22 @@ const std::vector<PTexMeshData::MeshData>& PTexMeshData::meshes() const {
 std::string PTexMeshData::atlasFolder() const {
   return atlasFolder_;
 }
+// this is to break the quad into 2 triangles
+// we need this triangle mesh to do object picking
+void computeTriangleMeshIndices(uint64_t numFaces,
+                                PTexMeshData::MeshData& currentSubMesh) {
+  for (size_t jFace = 0; jFace < numFaces; ++jFace) {
+    size_t offset = jFace * 4;
+    // 1st triangle is (0, 1, 2)
+    currentSubMesh.ibo_tri.push_back(currentSubMesh.ibo[offset + 0]);
+    currentSubMesh.ibo_tri.push_back(currentSubMesh.ibo[offset + 1]);
+    currentSubMesh.ibo_tri.push_back(currentSubMesh.ibo[offset + 2]);
+    // 2nd triangle is (0, 2, 3)
+    currentSubMesh.ibo_tri.push_back(currentSubMesh.ibo[offset + 0]);
+    currentSubMesh.ibo_tri.push_back(currentSubMesh.ibo[offset + 2]);
+    currentSubMesh.ibo_tri.push_back(currentSubMesh.ibo[offset + 3]);
+  }
+}
 
 // split the original ptex mesh into sub-meshes.
 //
@@ -240,6 +256,8 @@ std::vector<PTexMeshData::MeshData> splitMesh(
       }
     }
 
+    computeTriangleMeshIndices(chunkSize, subMeshes[i]);
+
     // add referenced vertices to submesh
     subMeshes[i].vbo.resize(refdVerts.size());
     subMeshes[i].nbo.resize(refdVerts.size());
@@ -333,6 +351,10 @@ std::vector<PTexMeshData::MeshData> loadSubMeshes(
         ibo[idx++] = local;
       }
     }  // for jFace
+
+    // this is to break the quad into 2 triangles
+    // we need this triangle mesh to do object picking
+    computeTriangleMeshIndices(numFaces, subMeshes[iMesh]);
 
     // compute the vbo, nbo for the current sub-mesh
     uint64_t numVertices = localToGlobal.size();
@@ -798,6 +820,8 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
                                       Magnum::GL::BufferUsage::StaticDraw);
     currentMesh->indexBuffer.setData(submeshes_[iMesh].ibo,
                                      Magnum::GL::BufferUsage::StaticDraw);
+    currentMesh->triangleMeshIndexBuffer.setData(
+        submeshes_[iMesh].ibo_tri, Magnum::GL::BufferUsage::StaticDraw);
   }
 #ifndef CORRADE_TARGET_APPLE
   LOG(INFO) << "Calculating mesh adjacency... ";
@@ -823,13 +847,26 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
     currentMesh->mesh
         .setPrimitive(Magnum::GL::MeshPrimitive::LinesAdjacency)
         // Warning:
-        // CANNOT use currentMesh.ibo.size() when calling
+        // CANNOT use currentMesh->indexBuffer.size() when calling
         // setCount because that returns the number of bytes of the buffer,
         // NOT the index counts
         .setCount(submeshes_[iMesh].ibo.size())
         .addVertexBuffer(currentMesh->vertexBuffer, offset,
                          gfx::PTexMeshShader::Position{})
         .setIndexBuffer(currentMesh->indexBuffer, offset,
+                        Magnum::GL::MeshIndexType::UnsignedInt);
+
+    // this triangle mesh will be used for object picking
+    currentMesh->triangleMesh
+        .setPrimitive(Magnum::GL::MeshPrimitive::Triangles)
+        // Warning:
+        // CANNOT use currentMesh->indexBuffer.size() when calling
+        // setCount because that returns the number of bytes of the buffer,
+        // NOT the index counts
+        .setCount(submeshes_[iMesh].ibo_tri.size())
+        .addVertexBuffer(currentMesh->vertexBuffer, offset,
+                         gfx::PTexMeshShader::Position{})
+        .setIndexBuffer(currentMesh->triangleMeshIndexBuffer, offset,
                         Magnum::GL::MeshIndexType::UnsignedInt);
   }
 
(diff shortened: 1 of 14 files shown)


bigbike:
This PR is being blocked from testing.
The error message says, "Collision mesh load failed. Aborting scene initialization.".
I am communicating with my folks to solve this issue first.

bigbike:
Well, it turns out a bit more difficult to enable object picking for the replica mesh.
Check the result:

picking

(By the way, I noticed that there is NO frustum culling now for the replica mesh. Why? I am investigating it.)

src/esp/assets/PTexMeshData.cpp
```diff
@@ -91,6 +91,22 @@ const std::vector<PTexMeshData::MeshData>& PTexMeshData::meshes() const {
 std::string PTexMeshData::atlasFolder() const {
   return atlasFolder_;
 }
+// this is to break the quad into 2 triangles
+// we need this triangle mesh to do object picking
+void computeTriangleMeshIndices(uint64_t numFaces,
```
jturner65, line 96: This method looks really useful.  Is there some centralized place where we could put a method like this so that it might be used by other processes/classes?

src/utils/viewer/viewer.cpp
```diff
@@ -630,6 +630,9 @@ void Viewer::mousePressEvent(MouseEvent& event) {
     unsigned int pickedObject =
         objectPickingHelper_->getObjectId(event.position(), windowSize());
 
+    // XXX
+    // std::cout << "picked object = " << pickedObject << std::endl;
+
```
jturner65, line 635: Remove this?

src/esp/assets/PTexMeshData.cpp
mosra: Oh! Can I steal this? :grin: 

I wanted to put something similar directly into Magnum's `MeshTools`, with the added feature of [properly triangulating non-convex quads](https://stackoverflow.com/questions/12239876/fastest-way-of-converting-a-quad-to-a-triangle-strip), and then use it in all places where needed (STL and OBJ meshes, mainly). Extremely hasty ASCII art to visualize what I mean:

```
+                              +
|\                             |\
| \                            | \  
|  +---+           this        |  +---+
|     /                        |  .  /
|    /              --->       | .  /
|   /                          | . /
|  /                           |. /
| /                            |./
|/                             |/
+                              +

                               +
                               |\ .
                               | \  .
                  instead      |  +---+
                  of this      |     /
                               |    /
                   -x->        |   /
                               |  /
                               | /
                               |/
                               + 
```

If you say yes, I can append this to #740 (or open a new PR, whichever you prefer) by tomorrow to avoid blocking you.

bigbike: Be my guest. 
Use it anywhere you want. :)

src/esp/assets/PTexMeshData.cpp
```diff
@@ -798,6 +820,8 @@ void PTexMeshData::uploadBuffersToGPU(bool forceReload) {
                                       Magnum::GL::BufferUsage::StaticDraw);
     currentMesh->indexBuffer.setData(submeshes_[iMesh].ibo,
                                      Magnum::GL::BufferUsage::StaticDraw);
+    currentMesh->triangleMeshIndexBuffer.setData(
```
eundersander, line 823: Is it true that we're using additional CPU and GPU memory for this triangle index buffer, for all PTexMesh's, regardless of whether the picking feature is being used? Does this increase our memory usage during RL training?

src/esp/gfx/Drawable.h
```diff
@@ -54,6 +54,16 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
 
   Magnum::GL::Mesh& getMesh() { return mesh_; }
 
+  /**
+   * @brief Get the Magnum GL mesh for visualization (e.g., used in object
```
eundersander, line 58: Can you explain this a little better? I would say that Drawables are entirely for visualization. That's the whole point of rendering. Do you mean debug visualization specifically?

src/esp/gfx/PTexMeshDrawable.cpp
```diff
@@ -41,12 +43,18 @@ PTexMeshDrawable::PTexMeshDrawable(scene::SceneNode& node,
 
 void PTexMeshDrawable::draw(const Magnum::Matrix4& transformationMatrix,
                             Magnum::SceneGraph::Camera3D& camera) {
+  bool usingDrawableId =
+      static_cast<RenderCamera&>(camera).isRenderingForObjectPicking();
```
eundersander, line 47: Nit: it feels a little weird that this Drawable class knows about object picking and knows to use its drawable ID in that case. Consider renaming the Camera method from `isRenderingForObjectPicking` to `useDrawableIds`.

We might find another scenarios to render using drawable Ids, and we won't have to maintain that logic in this class; it can live in the camera class or elsewhere.

bigbike: isRenderingForObjectPicking --> useDrawableIds
Good. I think it is better. Will change.

> we won't have to maintain that logic in this class; 

May I understand what logic you are mentioning?
"objectId" is a uniform in the shader that must be set dynamically on the fly. and it happens in every "drawable" before being drawn. It cannot be set in the camera class. camera class just received the signal -- "using the drawable id in the following rendering" and conveys it to the actual drawables.

src/esp/assets/PTexMeshData.cpp
bigbike: Plus it is just the index here. Yes, it will increase the memory usage, but compared to the HDR textures used in the replica mesh, it is not significant for sure. 
So I would leave it like this to keep the code simple.

eundersander: Can you create a Task and link it here in the code in a comment? Or some other way to remember this as a candidate for optimization.

From Brennan's recent presentation on batched rendering, it sounded like GPU memory is the bottleneck limiting further batching, so I'd at least like to remember this is here to be optimized later.

src/esp/gfx/PTexMeshDrawable.cpp
eundersander: >we won't have to maintain that logic in this class
I was just making another argument for my proposed name change. Like, maybe we want to flicker between drawable ID and semantic ID every other frame, for some debug purpose. Or we want to render with drawable ID for some reason other than object picking. All that new logic could live in the Camera class or elsewhere. This class just needs to know to check `RenderCamera::useDrawableIds`.

src/esp/gfx/Drawable.h
eundersander: I see. It sounds like "highlight" is the word we want.

src/esp/assets/PTexMeshData.cpp
bigbike: @eundersander :

I do not think it is necessary.
For the entire replica dataset, I printed out the memory usage of this index buffer "ibo_tri" for each model, list it side-by-side with the texture size in the following table:

NAME   ====     SIZE         ====        TEXTURE SIZE                                                         
apartment_0 ==== 104MB (0.10GB)  ====      7GB                                                                
apartment_1====  43MB (0.04GB)   ====       2.9GB                                                              
apartment_2 ==== 48MB (0.05GB)   ====       3.3GB                                                              
frl_apartment_0 ==== 40MB (0.04GB) ====     2.7GB                                                              
frl_apartment_1 ==== 40MB (0.04GB) ====     2.7GB                                                              
frl_apartment_2 ==== 41MB (0.04GB) ====     2.8GB                                                              
frl_apartment_3 ==== 40MB (0.04GB) ====     2.7GB                                                              
frl_apartment_4 ==== 41MB (0.04GB)   ====   2.8GB                                                              
frl_apartment_5====  40MB (0.04GB) ====     2.7GB                                                              
hotel_0    ====      25MB (0.02GB) ====     1.7GB                                                              
office_0   ====      13MB (0.01GB) ====     906.9MB                                                            
office_1   ====      9MB (0.009GB)  ====    650MB                                                              
office_2    ====     19MB (0.02GB)  ====    1.3GB                                                              
office_3      ====   27MB (0.03GB) ====     1.8GB                                                              
office_4    ====     22MB (0.02GB)   ====   1.5GB                                                              
room_0     ====      21MB (0.02GB)   ====   1.5GB                                                              
room_1     ====      14MB (0.01GB)  ====    993.8MB                                                            
room_2      ====     16MB (0.02GB) ====      1.1GB 

The size of the texture is roughly **70x** of the new index buffer size.
"GPU memory is the bottleneck limiting further batching" -- it makes sense. 
However, the index buffer is certainly not the "bottleneck" here.

eundersander: This is great data. I'm not asking you to make any code changes. I'm just asking for a code comment or something, to help the future engineer working on memory optimization. "beware this ibo_tri buffer consumes ~20 MB GPU memory per mesh and is only used for object-picking". If this is in the ballpark of 1% of total GPU mem usage, it might be worth pursuing in the future.

bigbike: Sure, I can add comments here. This is good suggestion.
In terms of optimization, this would be the very last place I will look into since the space complexity for the index buffer and the texture are not at the same level at all.

src/esp/assets/PTexMeshData.cpp
mosra: Ok, if you want to use a general-purpose utility, it's available in Magnum as [MeshTools::generateQuadIndices()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#aeba1cfbb6e33040f3025872a0b340db0) since mosra/magnum@b77eafb9deba8228214cd7759d12bbdff555eba0. Simply pull the magnum submodule, nothing else needs to be updated as there was no breaking change since last time.

image

PR #751, opened 2020-08-20
Added Habitat-Lab link as experiment framework to Readme

mathfac:
# Motivation and Context
There is no mentions of Habitat-Lab from the Readme. That can lead to users re-inventing the wheel for embodied AI experiments. Added Habitat-Lab link as experiment framework to Readme.

## Types of changes


Skylion007:
Should we maybe make this more visible? Like with a Badge? Otherwise, looks good to me.

mathfac:
Badge would be great, can be separate PR as require some design skills.

Issue #507, opened 2020-02-26
Make Viewer utility use Simulator class


matthewjmay:
## 🚀 Feature
Update the Viewer utility to internally use Simulator class

## Motivation

Right now there is lots of duplicate logic between the Viewer and the Simulator, and every time new features are added logic gets duplicated.

This will also be a nice test to make sure the Simulator API is flexible enough.

## Pitch

Change the implementation of Viewer to use Simulator.

PR #727, opened 2020-08-09
[WIP] Python Viewer

dexter1691:
## Motivation and Context
Adds a python based viewer to test the habitat_sim functionalities but with the flexibility of Python. 

## How Has This Been Tested

By running the python viewer

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #676, opened 2020-07-08
Added Rotation around Z-axis to the defaults control.

qureshinomaan:
## Motivation and Context
Gives the agent 6 degrees of freedom.

PR #752, opened 2020-08-21
Simulator C++ Viewer

aclegg3:
## Motivation and Context

The C++ Viewer application is used for rapid feature development and testing in Habitat-sim C++ codebase. 
Currently this application directly uses internal Habitat-sim structures (e.g. ResourceManager, SceneGraph, Pathfinder) and is therefore not representative of the functionality contained in the primary API target: Simulator.

This PR is a refactor of the Viewer application to directly use the Simulator class reducing code duplication and improving synchronization with the primary python simulator API.

Notes:
- Thanks to #727 for Simulator application GL context handling. 👍 
- Now caching number of Drawables visible after culling in RenderCamera
- Added `physicsDebugDraw` and `applyImpulse` to Simulator
- Future PR should add support for non-color sensor display options
- Agent state is no longer logged every time it is modified, but on keypress 'Q'
- Removed mouse wheel zooming in preparation from upcoming "mouse mode" feature

## How Has This Been Tested

Local testing.

## Types of changes

## Checklist


src/utils/viewer/simviewer.cpp
```diff
@@ -0,0 +1,724 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include <stdlib.h>
+
+#include <Magnum/configure.h>
+#include <Magnum/ImGuiIntegration/Context.hpp>
+#ifdef MAGNUM_TARGET_WEBGL
+#include <Magnum/Platform/EmscriptenApplication.h>
```
Skylion007, line 10: Looks like one of the other includes you added broke the emscripten build. Hmm...

Skylion007:
@mosra Looks like "flextGL" is one of your symbols. Any idea how it's leaking into the emscripten build and how to solve that issue?

Skylion007:
Fixes #507

src/utils/viewer/simviewer.cpp
```diff
@@ -0,0 +1,668 @@
...
+  for (auto entry : sensors.getSensors()) {
+    auto visualSensor =
+        dynamic_cast<esp::sensor::VisualSensor*>(entry.second.get());
+    if (visualSensor != nullptr) {
+      visualSensor->specification()->resolution = {event.windowSize()[1],
+                                                   event.windowSize()[0]};
+      simulator_->getRenderer()->bindRenderTarget(*visualSensor);
+      Cr::Utility::Debug() << visualSensor->framebufferSize();
+    }
+  }
```
aclegg3, line 435: Image is still stretched on canvas resize. Thoughts?

Skylion007: Can we not preserve aspect ratio somehow?

src/esp/sim/Simulator.cpp
```diff
@@ -145,7 +146,7 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
   sceneID_.push_back(activeSceneID_);
 
   if (config_.createRenderer) {
-    if (!context_) {
+    if (!context_ && !Magnum::GL::Context::hasCurrent()) {
       context_ = gfx::WindowlessContext::create_unique(config_.gpuDeviceId);
     }
```
Skylion007, line 152: Will we need some ifdef for that gl context code for the Python Viewer I wonder...

src/utils/viewer/simviewer.cpp
bigbike: Take a look at #322 and #331. 
I believe you will have to set the aspectRatioPolicy::Extend for the renderCamera.

aclegg3: Thanks, this worked @bigbike!

src/utils/viewer/simviewer.cpp
```diff
@@ -0,0 +1,669 @@
...
+  // configure and initialize default Agent and Sensor
+  auto agentConfig = esp::agent::AgentConfiguration();
+  // TODO: add extra, non-default actions
+  agentConfig.sensorSpecifications[0]->resolution =
+      esp::vec2i(viewportSize[1], viewportSize[0]);
+  // add selects a random initial state and sets up the default controls and
+  // step filter
+  simulator_->addAgent(agentConfig);
+
+  // Set up camera
```
bigbike, line 220: I will consider adding more details in the comments. "Why do we need to setup the render camera here since we are using sensor to obtain the observation?"
"Because eventually our visual sensor will render the observations using the renderCamera."

(It let me wonder if it is better to provide an API in the sensor class to set the aspectRatioPolicy for the renderCamera, making the renderCamera completely transparent to the developer. Ideally the user should never, ever know we have a default renderCamera attached in the SG.)

Skylion007:
@aclegg3 You should pull from master to integrate the pre-commit checks.

src/utils/viewer/simviewer.cpp
aclegg3: Could be part of the VisualSensor API.

src/utils/viewer/viewer.cpp
```diff
@@ -184,10 +164,6 @@ Viewer::Viewer(const Arguments& arguments)
       .setHelp("debug-bullet", "render Bullet physics debug wireframes")
       .addOption("physics-config", ESP_DEFAULT_PHYS_SCENE_CONFIG_REL_PATH)
       .setHelp("physics-config", "physics scene config file")
-      .addBooleanOption("disable-navmesh")
```
Skylion007, line 187: Don't we want to keep this option?

src/utils/viewer/viewer.cpp
```diff
@@ -213,130 +189,75 @@ Viewer::Viewer(const Arguments& arguments)
...
+      {"turnLeft",
+       esp::agent::ActionSpec::create(
+           "turnLeft", esp::agent::ActuationMap{{"amount", lookSensitivity}})},
+      {"turnRight",
+       esp::agent::ActionSpec::create(
+           "turnRight", esp::agent::ActuationMap{{"amount", lookSensitivity}})},
+      {"lookUp",
+       esp::agent::ActionSpec::create(
+           "lookUp", esp::agent::ActuationMap{{"amount", lookSensitivity}})},
+      {"lookDown",
```
Skylion007, line 245: We should probably also add the default actions added in #676

src/utils/viewer/viewer.cpp
```diff
@@ -381,79 +294,59 @@ void Viewer::addTemplateObject() {
...
 }
 
 void Viewer::torqueLastObject() {
-  if (objectIDs_.size() == 0)
+  auto existingObjectIDs = simulator_->getExistingObjectIDs();
+  if (existingObjectIDs.size() == 0)
     return;
   Mn::Vector3 torque = randomDirection() * 30;
-  physicsManager_->applyTorque(objectIDs_.back(), torque);
+  simulator_->applyTorque(torque, existingObjectIDs.back());
```
Skylion007, line 349: Interesting convention we have of torque, object_id, would expect it the other way around.

src/utils/viewer/viewer.cpp
```diff
@@ -737,62 +587,53 @@ void Viewer::mouseMoveEvent(MouseMoveEvent& event) {
 // NOTE: Mouse + shift is to select object on the screen!!
 void Viewer::keyPressEvent(KeyEvent& event) {
```
Skylion007, line 588: Do we have a help button to display keyPress events? Or print them out to terminal somehow?

src/esp/gfx/RenderTarget.h
```diff
@@ -48,7 +48,7 @@ class RenderTarget {
                const Magnum::Vector2& depthUnprojection)
       : RenderTarget{size, depthUnprojection, nullptr} {};
 
-  ~RenderTarget() { LOG(INFO) << "Deconstructing RenderTarget"; }
+  ~RenderTarget() {}
```
Skylion007, line 51: Any reason this LOG got nuked?

src/utils/viewer/viewer.cpp
```diff
@@ -118,56 +108,46 @@ class Viewer : public Mn::Platform::Application {
...
-  esp::scene::SceneNode* rgbSensorNode_ = nullptr;
-
-  std::string stageFileName;
-  esp::scene::SceneGraph* sceneGraph_;
-  esp::scene::SceneNode* rootNode_;
-
-  int navMeshVisPrimID_ = esp::ID_UNDEFINED;
-  esp::scene::SceneNode* navMeshVisNode_ = nullptr;
 
+  std::string sceneFileName;
```
Skylion007, line 129: Is this stageFileName or sceneFileName? Maybe add a comment it can also be stage?

src/utils/viewer/viewer.cpp
aclegg3: We do it this way to later support set(thing, objectId, sceneId) with sceneId given a default value.

src/esp/gfx/RenderTarget.h
aclegg3: We need to rebuild the RenderTarget every time the screen is resized. This spams the terminal output.

Skylion007: We don't have a DEBUG level? Fair enough.

src/utils/viewer/viewer.cpp
aclegg3: Thanks for bringing this up. Realized several args were not handled.

src/utils/viewer/viewer.cpp
aclegg3: Could bind this to keys, but haven't done so for viewer before.

src/utils/viewer/viewer.cpp
aclegg3: Added some help text.

src/utils/viewer/viewer.cpp
aclegg3: These are not defined in C++. If we add them we can expose in Viewer then.

src/esp/sim/Simulator.cpp
```diff
@@ -401,6 +402,15 @@ void Simulator::applyForce(const Magnum::Vector3& force,
   }
 }
 
+void Simulator::applyImpulse(const Magnum::Vector3& impulse,
+                             const Magnum::Vector3& relPos,
+                             const int objectID,
+                             const int sceneID) {
+  if (sceneHasPhysics(sceneID)) {
+    physicsManager_->applyForce(objectID, impulse, relPos);
+  }
```
jturner65, line 411: make this physicsManager_->applyImpulse(objectID, impulse, relPos);

bigbike:
@mosra:

Would you join us to take a pass on this PR? Thank you!

src/esp/sim/Simulator.cpp
```diff
@@ -148,7 +149,7 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
   sceneID_.push_back(activeSceneID_);
 
   if (config_.createRenderer) {
-    if (!context_) {
+    if (!context_ && !Magnum::GL::Context::hasCurrent()) {
```
mosra, line 152: I'd put some comment here (because even if I knew what was this for back then, I definitely don't remember now).

src/esp/sim/Simulator.h
```diff
@@ -297,6 +312,22 @@ class Simulator {
...
+   * to apply the impulse.
+   * @param objectID The ID of the object identifying it in @ref
+   * esp::physics::PhysicsManager::existingObjects_.
+   * @param sceneID !! Not used currently !! Specifies which physical scene of
+   * the object.
+   */
+  void applyImpulse(const Magnum::Vector3& impulse,
+                    const Magnum::Vector3& relPos,
+                    const int objectID,
+                    const int sceneID = 0);
```
mosra, line 329:
```suggestion
                    int objectID,
                    int sceneID = 0);
```

in the declaration it's just noise :) Same above (and elsewhere) I think?

src/utils/viewer/viewer.cpp
```diff
@@ -213,139 +199,115 @@ Viewer::Viewer(const Arguments& arguments)
...
-    LOG(FATAL)
-        << physicsConfigFilename
-        << " was not found, specify an existing file in --physics-config";
+  // configure and intialize Simulator
+  auto simConfig = esp::sim::SimulatorConfiguration();
+  simConfig.scene.id = sceneFileName;
+  simConfig.enablePhysics = useBullet;
+  simConfig.frustumCulling = true;
+  if (args.isSet("stage-requires-lighting")) {
+    Cr::Utility::Debug() << "Stage using DEFAULT_LIGHTING_KEY";
```
mosra, line 214:
```suggestion
    Mn::Debug{} << "Stage using DEFAULT_LIGHTING_KEY";
```

is shorter and alias to the same thing, if you prefer. Same elsewhere

src/utils/viewer/viewer.cpp
```diff
@@ -820,33 +705,89 @@ void Viewer::keyPressEvent(KeyEvent& event) {
...
+  Cr::Utility::Debug() << "==================================================";
+  Cr::Utility::Debug() << "Welcome to the Habitat-sim C++ Viewer application!";
+  Cr::Utility::Debug() << "==================================================";
+  Cr::Utility::Debug() << " Mouse Functions:";
+  Cr::Utility::Debug() << " ----------------";
+  Cr::Utility::Debug() << "   LEFT: ";
+  Cr::Utility::Debug()
+      << "     Click and drag to rotate the agent and look up/down.";
+  Cr::Utility::Debug() << "   RIGHT: ";
+  Cr::Utility::Debug()
```
mosra, line 746: I'd instead put this into a huge `R"()"` literal and print that, it's less work and easier to update ;)

It could also be put to Utility::Arguments via `setGlobalHelp()` (as the current text there is rather useless) ... but that's shown only if the user explicitly asks for `--help` so that's probably not so useful.

src/utils/viewer/viewer.cpp
Skylion007: Ah, my bad. Don't bother then.

src/utils/viewer/viewer.cpp
```diff
@@ -639,57 +548,48 @@ void Viewer::mousePressEvent(MouseEvent& event) {
...
+
+      esp::physics::RaycastResults raycastResults = simulator_->castRay(ray);
+
+      for (auto& hit : raycastResults.hits) {
+        Corrade::Utility::Debug() << "Hit: ";
+        Corrade::Utility::Debug() << "  distance: " << hit.rayDistance;
+        Corrade::Utility::Debug() << "  object: " << hit.objectId;
+        Corrade::Utility::Debug() << "  point: " << hit.point;
+        Corrade::Utility::Debug() << "  normal: " << hit.normal;
+      }
```
mosra, line 568: Alternative ideas: 

1. print this via imgui the same as you do with fps? (I'm not sure what's the intended use case -- if it's meant to be selected in the terminal then that's useless)
2. enable this printing only if some `--debug-raycast` option is enabled on the command-line? (would help with console noise if there's a lot of output, but probably makes no sense if all you need when clicking is seeing this info)

mosra (commented):
Totally random comments, hope any of it is useful :)

src/esp/sim/Simulator.cpp
Skylion007: It had something to do with the changes needed for the Python viewer on #727

mosra: Yeah that's what I remember, but I don't remember *why* :sweat_smile: @dexter1691 do *you* remember? :)

dexter1691: I think this was because a GL context already exists and it crashes since two magnum context cannot exist. When creating a viewer based app, there is no need to create a WindowlessContext since another context already exists. This would be required both for python viewer and this PR.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -149,10 +149,11 @@ uint32_t RenderCamera::draw(MagnumDrawableGroup& drawables, Flags flags) {
 
   if (flags & Flag::FrustumCulling) {
     // draw just the visible part
-    size_t numVisibles = cull(drawableTransforms);
+    previousNumVisibleDrawables_ = cull(drawableTransforms);
```
bigbike, line 152: Now this value "number of visible drawables" is a state stored in the class now. 

You will have to reset it at the beginning of this draw() function. Otherwise, if user turn off the FC during the run-time, this value is not updated.

`previousNumVisibleDrawables_ = 0;`

src/esp/sim/Simulator.h
```diff
@@ -112,7 +112,12 @@ class Simulator {
    * simulator.  This will only be nonzero if the simulator is built in
    * --headless mode on linux
    */
-  int gpuDevice() const { return context_->gpuDevice(); }
+  int gpuDevice() const {
+    if (context_ == nullptr) {
+      return 0;
```
bigbike, line 117: Is device 0 the default GPU device in UI mode?
Then if context_ is NULL, does it mean the context is not built (which means it is illegal to call this function), and we should return -1 maybe?

In my opinion:
Do we need to just quit and prompt an error (using CORRADE_ASSERT)? It does not make any sense to tolerate this error.

mosra: +1 for the assert :)

src/esp/sim/Simulator.h
mosra: @aclegg3 just to clarify after seeing the diff -- I meant just the declaration, it could stay as-is in the definition (`const` is good, after all). 

However if your IDE (or some linter) forces you to have definition and declaration match, then you can't do much :/ (I hope that's not the case?)

src/utils/viewer/viewer.cpp
aclegg3: Just removed this output. Originally meant for debugging anyway.

src/utils/viewer/viewer.cpp
```diff
@@ -90,14 +86,57 @@ class Viewer : public Mn::Platform::Application {
...
+  'f': (physics) Push the most recently added object.
+  't': (physics) Torque the most recently added object.
+  'v': (physics) Invert gravity.
+==================================================
+  )";
+
+  //! Print viewer help text to terminal output.
+  void printHelpText() { Mn::Debug{} << helpText; };
+
+  void printHelpText2();
```
mosra, line 139: This declaration is new and I don't see it used / defined anywhere? :)

src/esp/gfx/RenderCamera.cpp
aclegg3: I now reset this with the number of drawables in the draw call. Should be more accurate than 0, since all will be visible with no culling.

src/esp/sim/Simulator.cpp
aclegg3: Yeah, creating a Simulator instance will try to create another context.

src/utils/viewer/viewer.cpp
```diff
@@ -213,139 +243,115 @@ Viewer::Viewer(const Arguments& arguments)
...
-                  << pathfinder_->distanceToClosestObstacle(currentPosition);
-
-        return currentPosition;
-      });
-    }
-  }
-
-  renderCamera_->node().setTransformation(
-      rgbSensorNode_->absoluteTransformation());
+  rootNode_ = &simulator_->getActiveSceneGraph().getRootNode();
```
bigbike, line 343: Actually, I do not think you need the rootNode_ anymore? You may double confirm it.

src/esp/sim/Simulator.h
```diff
@@ -294,22 +305,36 @@ class Simulator {
...
+   * esp::physics::PhysicsManager::applyImpulse.
+   * @param impulse The desired linear impulse to apply.
+   * @param relPos The desired location relative to the object origin at which
+   * to apply the impulse.
+   * @param objectID The ID of the object identifying it in @ref
+   * esp::physics::PhysicsManager::existingObjects_.
+   * @param sceneID !! Not used currently !! Specifies which physical scene of
+   * the object.
+   */
+  void applyImpulse(const Magnum::Vector3& impulse,
```
bigbike, line 322: What is the difference between `applyForce` and `applyImpulse`. Does it apply `f * dt` instead of `f`?

Can we add comment to tell the user how to choose in between?

src/utils/viewer/viewer.cpp
```diff
@@ -90,14 +86,55 @@ class Viewer : public Mn::Platform::Application {
...
   void wiggleLastObject();
-
-  void recomputeNavMesh(const std::string& sceneFilename,
-                        esp::nav::NavMeshSettings& navMeshSettings);
-
   void invertGravity();
   Mn::Vector3 randomDirection();
 
-  void toggleNavMeshVisualization();
+  std::string helpText = R"(
```
bigbike, line 92: Thank you very much for doing this. It must be very time-consuming.

src/utils/viewer/viewer.cpp
```diff
@@ -118,80 +155,73 @@ class Viewer : public Mn::Platform::Application {
...
+              .setColorBufferSize(Mn::Vector4i(8, 8, 8, 8))
+              .setSampleCount(4)} {
   Cr::Utility::Arguments args;
 #ifdef CORRADE_TARGET_EMSCRIPTEN
   args.addNamedArgument("scene")
 #else
   args.addArgument("scene")
 #endif
-      .setHelp("scene", "scene file to load")
+      .setHelp("scene", "scene/stage file to load")
```
bigbike, line 205: Nit: maybe it is better to display `scene (stage)`

src/utils/viewer/viewer.cpp
```diff
@@ -737,62 +663,53 @@ void Viewer::mouseMoveEvent(MouseMoveEvent& event) {
...
   const auto key = event.key();
-  bool agentMoved = false;
   switch (key) {
     case KeyEvent::Key::Esc:
       std::exit(0);
       break;
     case KeyEvent::Key::Left:
-      controls_(*agentBodyNode_, "turnLeft", lookSensitivity);
-      agentMoved = true;
+      simulator_->getAgent(0)->act("turnLeft");
```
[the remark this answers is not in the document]
bigbike: Again, you can cache the `defaultAgent_` so that you do not have to call the `getAgent(...)` multiple times.

src/utils/viewer/viewer.cpp
```diff
@@ -118,80 +157,73 @@ class Viewer : public Mn::Platform::Application {
...
+      assetAttrManager_ = nullptr;
+  std::shared_ptr<esp::assets::managers::StageAttributesManager>
+      stageAttrManager_ = nullptr;
+  std::shared_ptr<esp::assets::managers::PhysicsAttributesManager>
+      physAttrManager_ = nullptr;
 
   bool debugBullet_ = false;
 
-  std::vector<int> sceneID_;
+  esp::scene::SceneNode* rootNode_ = nullptr;
```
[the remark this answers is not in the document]
aclegg3: Yeah, I agree this will probably happen again in the future. Just removing it because it has no use currently.

src/esp/sim/Simulator.h
aclegg3: Applying impulse modifies object velocity immediately, rather than being integrated as a force during dynamics compute phase. Copied some doc text from PhysicsManager to clarify this in Simulator.

src/utils/viewer/viewer.cpp
aclegg3: From an API perspective we treat these the same way for a Simulator. 
Once dev is complete you'll be able to provide a stage asset, stage config, or scene config to initialize the Simulator. If you only provide the stage (current implementation), then we'll interpret that as a scene which is empty except for the stage you provide.

aclegg3: maybe `.setHelp("scene", "scene or stage file to load.")` ?

bigbike: > From an API perspective we treat these the same way for a Simulator.
Exactly. I agree.

> .setHelp("scene", "scene or stage file to load.")
I see. They are actually different. When I say `scene (stage)`. I thought they were the same.
Then you do not have to change. 
The question is do we need to add comments somewhere so the user knows the difference in between?

src/utils/viewer/viewer.cpp
```diff
@@ -737,62 +669,53 @@ void Viewer::mouseMoveEvent(MouseMoveEvent& event) {
...
   const auto key = event.key();
-  bool agentMoved = false;
   switch (key) {
     case KeyEvent::Key::Esc:
       std::exit(0);
       break;
     case KeyEvent::Key::Left:
-      controls_(*agentBodyNode_, "turnLeft", lookSensitivity);
-      agentMoved = true;
+      simulator_->getAgent(defaultAgentId_)->act("turnLeft");
```
bigbike, line 677: Sorry, my bad. In case you did not see my previous comments, post it again:
You can consider caching the defaultAgent_ instead of the defaultAgentId_. because you need to call it every time below.

src/utils/viewer/viewer.cpp
aclegg3: This should probably be documented across the stack when the scene import feature arrives. We're planning a docs page to cover this. May also want to put up a terminology docs page to support naming conventions.

src/esp/sim/Simulator.h
dexter1691: AFAIR, there is some issue with the order in which these things get called. gpuDevice() is probably getting called before the context is built? I remember my python viewer crashing without this null check.

bigbike: @dexter1691 : If you read the code (next line) here, `context_->gpuDevice()`, how could `gpuDevice()` get called before the context is constructed?

src/utils/viewer/viewer.cpp
```diff
@@ -90,14 +87,55 @@ class Viewer : public Mn::Platform::Application {
...
+==================================================
+Welcome to the Habitat-sim C++ Viewer application!
+==================================================
+Mouse Functions:
+----------------
+  LEFT:
+    Click and drag to rotate the agent and look up/down.
+  RIGHT:
+    (With 'enable-physics') Click a surface to instance a random primitive object at that location.
+  SHIFT-RIGHT:
```
bigbike, line 103: @aclegg3 : FYI. The object picking function does not work any more.

src/utils/viewer/viewer.cpp
```diff
@@ -90,14 +87,55 @@ class Viewer : public Mn::Platform::Application {
...
+
+  Agent Controls:
+  'wasd': Move the agent's body forward/backward, left/right.
+  'zx': Move the agent's body up/down.
+  arrow keys: Turn the agent's body left/right and camera look up/down.
+  '9': Randomly place agent on NavMesh (if loaded).
+  'q': Query the agent's state and print to terminal.
+
+  Utilities:
+  'e' enable/disable frustum culling.
```
bigbike, line 119: Also, "imgui" seems not work as well. I did not see the realtime FPS.

src/utils/viewer/viewer.cpp
aclegg3: It is working for me 🤔

aclegg3:
> I am testing this local branch.
> I found "imgui" and "object picking" currently are not working. Can we make sure we do not break any existing functions to the viewer before we merge it in? Thanks!

These are both working locally for me. Can you recompile and check again?

bigbike:
> These are both working locally for me. Can you recompile and check again?

I deleted the entire `build/` folder, pulled the latest branch "sim-viewer", and did a clean build from scratch. Unfortunately, both functions are still broken. I was testing it in a linux machine. See the attached screenshot.



Now I am going to test it on Mac.

bigbike:
Confirmed it works for the Mac OSX, but it is broken on Linux.

src/utils/viewer/viewer.cpp
```diff
@@ -540,21 +498,11 @@ void Viewer::drawEvent() {
 
     // rendering
     esp::gfx::RenderCamera::Flags flags;
-    if (frustumCullingEnabled_) {
+    if (simulator_->isFrustumCullingEnabled()) {
       flags |= esp::gfx::RenderCamera::Flag::FrustumCulling;
     }
     renderCamera_->draw(objectPickingHelper_->getDrawables(), flags);
 
-    // Neither the blend equation, nor the blend function is changed,
```
bigbike, line 548: Suggest to keep this comments. I intend to put it here, since this is some valuable lesson I learned when debugging the "blending" issues. It cost me a couple of hours to figure it out even though it is not used eventually (since I found some other simpler solutions).

bigbike:
@aclegg3 : Not so sure if you have a linux box with you.
Just FYI: I am also looking into this problem on my side as well.

bigbike:
Wrote a PR tonight to fix the aforementioned issue:
https://github.com/facebookresearch/habitat-sim/pull/775

bigbike:
> Confirmed it works for the Mac OSX, but it is broken on Linux.

Actually, the rendering of the picking object was also wrong on Mac OSX due to missing depth info. See pictures in #775.

aclegg3:
All open issues resolved. Ready for review.

bigbike:
@aclegg3 :
One more thing in the Simulator.h
Check the inline comments.
Would you use CORRADE_ASSERT?

Skylion007 (approved):

bigbike (approved):

PR #768, opened 2020-08-27
Make circleci hook for pre-commit and add flake8-bugbear and flake8-comprehensions

Skylion007:
## Motivation and Context

* Its another python flake8 style checker.
* Helps prevent difficult to detect bugs
* Functions a plugin of flake8 and is used by PyTorch and others
* Now runs our pre-commit hooks on .circleci!

## Types of changes
* I ran flake8 with the config and everything worked well. I removed some bugs and unnecessary code from the repo.

## Checklist

- [] My code follows the code style of this project.
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.
- [] All new and existing tests passed.


.circleci/config.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - run:
           name: setup
           command: |
-              sudo pip install -U black flake8 "isort[pyproject]" numpy pytest sphinx pillow tqdm hypothesis
+              sudo pip install -U black flake8 flake8-bugbear flake8-comprehensions "isort[pyproject]" numpy pytest sphinx pillow tqdm hypothesis
```
erikwijmans, line 18: `flake8-comprehensions` looks a little too opinionated and annoying IMO.  For instance, I personally prefer `set(...)` instead of `{...}` as the latter is too visually similar to a list or dict comprehension despite having very different implications, which makes scanning code hard.

My vote is to keep the linter to looking for things that are actual bugs, not small performance optimizations -- especially given that for anything we actually cared about the performance of, we'd write in cpp.

Skylion007: @erikwijmans I can add that specific rule to the ignore, but some of the other ones are valuable.

Skylion007: @erikwijmans Yeah, PyTorch has C408 on its global ignore list so I'll add it too.

Skylion007: @erikwijmans Also I agree about disabling bits that affect readability, but a lot of flake8-comprehensions are performance optimizations that make a lot of sense and don't compromise readability (like not converting a generator to a list before its fed to a builtin that takes an iterator). 

Since we do care about speed, I would vote on keeping it, but adding any of the opinionated errors codes to the flake8 ignore config.

examples/instance_segmentation/common.py
```diff
@@ -7,13 +8,11 @@
...
-            instance_id
-            for instance_id, name in instance_id_to_name.items()
-            if name in labels
-        ]
-    )
+    labels_we_care_about = {
+        instance_id
+        for instance_id, name in instance_id_to_name.items()
+        if name in labels
+    }
```
erikwijmans, line 15:
```suggestion
    labels_we_care_about = set(
        instance_id
        for instance_id, name in instance_id_to_name.items()
        if name in labels
    )
```

tests/test_pyrobot_noisy_controls.py
```diff
@@ -92,9 +92,7 @@ def test_pyrobot_noisy_actions(noise_multiplier, robot, controller):
     )
     agent = habitat_sim.Agent(scene_graph.get_root_node().create_child(), agent_config)
 
-    for base_action in set(
-        act.replace("noisy_", "") for act in agent_config.action_space
-    ):
+    for base_action in {act.replace("noisy_", "") for act in agent_config.action_space}:
```
erikwijmans, line 95:
```suggestion
    for base_action in set(act.replace("noisy_", "") for act in agent_config.action_space):
```

PR #771, opened 2020-08-28
Add CMake-format pre-commit hooks and linters

Skylion007:
## Motivation and Context

It ensures we enforce a consistent style across the cmake files in our repo. For the first round, I am only having it handle the code, and not handle any of the formatting of comments. This will also add .circleci check that ensures cmake edits confirm thanks to .circleci. It would also automatically format cmakefiles during commit. Let me know if I should further customize the style file any further.

## How Has This Been Tested
I reformatted the code and all the build configs seemed to build just fine.

## Types of changes
It adds a new linter. I am also going to do a big spellchecker PR that should fix all the misspellings in comments and code.

## Checklist


Skylion007:
Also wasn't sure if I should standardize to tabs of space 2 or tabs space size 4, thoughts?

Skylion007:
Since I expect this to be controversial, we can set up a configuration like so. Suggestions welcome: https://cmake-format.readthedocs.io/en/latest/index.html

bigbike:
This is a very interesting PR. I like it. It makes the CMake related files nice and clean.

Any thoughts? @other reviewers.

src/cmake/dependencies.cmake
```diff
@@ -7,10 +7,12 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}")
...
 if(NOT USE_SYSTEM_MAGNUM)
-  # These are enabled by default but we don't need them right now -- disabling
-  # for slightly faster builds. If you need any of these, simply delete a line.
-  set(WITH_INTERCONNECT OFF CACHE BOOL "" FORCE)
-  add_subdirectory("${DEPS_DIR}/corrade")
+    # These are enabled by default but we don't need them right now -- disabling
+    # for slightly faster builds. If you need any of these, simply delete a line.
+    set(WITH_INTERCONNECT
+        OFF
+        CACHE BOOL "" FORCE)
```
eundersander, line 14: I wonder what's the logic behind these aggressive line breaks for `set` arguments.

src/esp/sensor/CMakeLists.txt
```diff
@@ -1,51 +1,31 @@
...
-
 add_library(sensor STATIC ${sensor_SOURCES})
 
-target_link_libraries(sensor
-  PUBLIC
-    core
-    gfx
-    scene
-)
+target_link_libraries(sensor PUBLIC core gfx scene)
```
eundersander, line 18: I'm not a fan of packing these onto one line. It's easier to notice additions/deletions in diffs when each list item is its own line.

Skylion007: They only get packed onto one line if they are three or less arguments.

src/cmake/dependencies.cmake
Skylion007: @eundersander explained in the docs: https://cmake-format.readthedocs.io/en/latest/format-case_studies.html

eundersander: It looks like this style is to accommodate comments on each argument, which makes sense.

Skylion007: I could also give it control over formatting the cmake comments, but have opted not to for the first run.

src/cmake/dependencies.cmake
```diff
@@ -130,78 +173,176 @@ endif()
...
-  set(WITH_ANYIMAGECONVERTER ON CACHE BOOL "WITH_ANYIMAGECONVERTER" FORCE)
-  set(WITH_PRIMITIVEIMPORTER ON CACHE BOOL "" FORCE)
-  set(WITH_STANFORDIMPORTER ON CACHE BOOL "" FORCE)
-  set(WITH_STBIMAGEIMPORTER ON CACHE BOOL "WITH_STBIMAGEIMPORTER" FORCE)
-  set(WITH_STBIMAGECONVERTER ON CACHE BOOL "WITH_STBIMAGECONVERTER" FORCE)
-  set(WITH_EMSCRIPTENAPPLICATION OFF CACHE BOOL "WITH_EMSCRIPTENAPPLICATION" FORCE)
-  set(WITH_GLFWAPPLICATION OFF CACHE BOOL "WITH_GLFWAPPLICATION" FORCE)
-  set(WITH_EIGEN ON CACHE BOOL "WITH_EIGEN" FORCE) # Eigen integration
-  set(WITH_IMGUI ON CACHE BOOL "WITH_IMGUI" FORCE) # ImGui integration
+  set(WITH_TINYGLTFIMPORTER
```
Skylion007, line 211: @erikwijmans It's expressions like this that make prefer the default limit of 6 args per line instead of 3.

src/esp/sensor/CMakeLists.txt
mosra: Is there some middle ground between "aggressively packing everything with < 4 arguments onto a single line" and "aggressively expanding everything with >= 4 arguments onto several lines"? Something like "always expand `target_link_libraries()`" or "preserve whatever format was there for certain commands"?

Like, in this case I would prefer to have the args separated on multiple lines because it's *very likely* more libs will get added later, however with the `set(thing CACHE BOOL "bla" OFF)` it'll *never happen* that there will be more arguments and spreading it out like that just adds unnecessary congitive load as the `WITH_` options etc are no longer nicely aligned under each other.

mosra (commented):
General comment: I personally **do not like** autoformatters because the tool is never smart enough to actually preserve readability instead of just applying rigid rules (and making things worse in many cases).

But this is not my code, so please do not take my opinion too seriously ;)

Skylion007:
@mosra It can be disabled at any point if it really impacts readability. :)

src/esp/sensor/CMakeLists.txt
Skylion007: @mosra There is, is there anything we should expand besides target_link_libraries?

mosra: I thought about doing something similar to the `set(bla_SOURCES many things here)` commands but most of those lists could be put directly into `add_library()` / `add_executable()` ... and conditional sources later added with `target_sources()`. 

So maybe add `add_executable()` and `target_sources()` to the list as well, even though the latter is probably not used anywhere yet.

Skylion007: We can just make a note to add those later once we use those features.

PR #774, opened 2020-08-31
-- Expand Attributes' JSON-loading tests

jturner65:
## Motivation and Context
We needed more tests for attributes setting via JSON.  This PR adds the following : 
1. The ability to parse a JSON doc directly into an attributes template.  The JSON did not have to originate from a file.
2. Example JSON strings for each type attributes that can be built from a JSON file.  These illustrate tags being used, with non-default values so that they can be used to test the JSON parsing functionality.
3. Test JSON loading and parsing for PhysicsManager, Object and Stage Attributes
4.  JSON tests were also moved from core to IO. 

## How Has This Been Tested
All existing c++ and python tests have passed; Test added to verify all expected JSON fields are correctly populated. 

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/attributes/ObjectAttributes.cpp
+++ b/src/esp/assets/attributes/ObjectAttributes.cpp
@@ -17,7 +17,7 @@ const std::map<std::string, esp::assets::AssetType>
         {"suncg", AssetType::SUNCG_SCENE},
 };
 
-const std::string AbstractObjectAttributes::JSONConfigExample =
+const std::string AbstractObjectAttributes::JSONConfigTestString =
     R"({
       "scale":[2,3,4],
       "margin": 0.9,
@@ -30,8 +30,8 @@ const std::string AbstractObjectAttributes::JSONConfigExample =
       "render mesh": "testJSONRenderAsset.glb",
       "collision mesh": "testJSONCollisionAsset.glb")";
 
-const std::string ObjectAttributes::JSONConfigExample =
-    AbstractObjectAttributes::JSONConfigExample +
+const std::string ObjectAttributes::JSONConfigTestString =
+    AbstractObjectAttributes::JSONConfigTestString +
     R"(,
       "mass": 9,
       "use bounding box for collision": true,
@@ -40,8 +40,8 @@ const std::string ObjectAttributes::JSONConfigExample =
       "COM": [0.1,0.2,0.3]
     })";
 
-const std::string StageAttributes::JSONConfigExample =
-    AbstractObjectAttributes::JSONConfigExample +
+const std::string StageAttributes::JSONConfigTestString =
+    AbstractObjectAttributes::JSONConfigTestString +
     R"(,
       "gravity": [9,8,7],
       "origin":[1,2,3],
--- a/src/esp/assets/attributes/ObjectAttributes.cpp
+++ b/src/esp/assets/attributes/ObjectAttributes.cpp
@@ -17,11 +17,11 @@ const std::map<std::string, esp::assets::AssetType>
         {"suncg", AssetType::SUNCG_SCENE},
 };
 
-const std::string AbstractObjectAttributes::JSONConfigTemplate =
+const std::string AbstractObjectAttributes::JSONConfigExample =
     R"({
       "scale":[2,3,4],
       "margin": 0.9,
-      "friction coefficient": 0.321,
+      "friction coefficient": 0.321, 
       "restitution coefficient": 0.456,
       "requires lighting": false,
       "units to meters": 1.1,
@@ -30,8 +30,8 @@ const std::string AbstractObjectAttributes::JSONConfigTemplate =
       "render mesh": "testJSONRenderAsset.glb",
       "collision mesh": "testJSONCollisionAsset.glb")";
 
-const std::string ObjectAttributes::JSONConfigTemplate =
-    AbstractObjectAttributes::JSONConfigTemplate +
+const std::string ObjectAttributes::JSONConfigExample =
+    AbstractObjectAttributes::JSONConfigExample +
     R"(,
       "mass": 9,
       "use bounding box for collision": true,
@@ -40,8 +40,8 @@ const std::string ObjectAttributes::JSONConfigTemplate =
       "COM": [0.1,0.2,0.3]
     })";
 
-const std::string StageAttributes::JSONConfigTemplate =
-    AbstractObjectAttributes::JSONConfigTemplate +
+const std::string StageAttributes::JSONConfigExample =
+    AbstractObjectAttributes::JSONConfigExample +
     R"(,
       "gravity": [9,8,7],
       "origin":[1,2,3],
--- a/src/esp/assets/attributes/ObjectAttributes.cpp
+++ b/src/esp/assets/attributes/ObjectAttributes.cpp
@@ -17,6 +17,38 @@ const std::map<std::string, esp::assets::AssetType>
         {"suncg", AssetType::SUNCG_SCENE},
 };
 
+const std::string AbstractObjectAttributes::JSONConfigTemplate =
+    R"({
+      "scale":[2,3,4],
+      "margin": 0.9,
+      "friction coefficient": 0.321,
+      "restitution coefficient": 0.456,
+      "requires lighting": false,
+      "units to meters": 1.1,
+      "up":[2.1,0,0],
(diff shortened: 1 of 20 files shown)


src/esp/assets/attributes/ObjectAttributes.h
```diff
@@ -18,6 +20,14 @@ namespace attributes {
  */
 class AbstractObjectAttributes : public AbstractAttributes {
  public:
+  /**
+   * @brief This defines an example JSON descriptor for @ref
+   * AbstractObjectAttributes Has values that are different than defaults so
+   * this can be used to test JSON loading. These values may break the code, so
+   * should not be used.
```
aclegg3, line 27:
```suggestion
   * this can be used to test JSON loading. These values may be purposefully invalid for testing, so
   * should not be used in practice.
```
Small suggestion on wording this warning. Same for others.

src/esp/assets/managers/ObjectAttributesManager.cpp
```diff
@@ -102,57 +102,45 @@ void ObjectAttributesManager::createDefaultPrimBasedAttributesTemplates() {
...
-  }
-
-  // Construct a ObjectAttributes and populate with any
-  // AbstractPhysicsAttributes fields found in json.
+ObjectAttributes::ptr ObjectAttributesManager::loadAttributesFromJSONDoc(
+    const std::string& templateName,
+    const io::JsonDocument& jsonConfig) {
+  // Construct a ObjectAttributes and populate with any AbstractObjectAttributes
+  // fields found in json.
   auto objAttributes = this->createPhysicsAttributesFromJson<ObjectAttributes>(
```
aclegg3, line 110: Should this be?
```suggestion
  auto objAttributes = this->createObjectAttributesFromJson<ObjectAttributes>(
```

mosra (commented):
If I'm reviewing with too much effort, let me know please ;)

src/tests/IOTest.cpp
```diff
@@ -94,3 +99,54 @@ TEST(IOTest, tokenizeTest) {
   const auto& t3 = tokenize(file, ",|", 0, true);
   EXPECT_EQ((std::vector<std::string>{"", "a", "bb", "c"}), t3);
 }
+
```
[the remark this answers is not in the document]
aclegg3: Agreed most of IO is redundant now. We should cull some of this and replace any usage with Corrade. 👍 

However, we will have some new IO coming in soon with tinyXML used to parse URDF and possibly some SDF down the line(e.g. #729). We could leave JSON testing here and add the XML, or plan to isolate JSON and XML based tests into new files.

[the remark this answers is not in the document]
mosra: Depends on how huge the JSON and XML test cases would be, I'd say ... if it grows to 10s/100s of cases then I'd vote for having them in dedicated test files / executables.

src/esp/assets/attributes/ObjectAttributes.cpp
```diff
@@ -17,6 +17,38 @@ const std::map<std::string, esp::assets::AssetType>
         {"suncg", AssetType::SUNCG_SCENE},
 };
 
+const std::string AbstractObjectAttributes::JSONConfigExample =
```
erikwijmans, line 20: Compiling this into the source code is weird IMO -- it seems like it will be hard to discover. If the goal is to provide users with an example of these, files in the examples folder makes a lot more sense.

src/esp/bindings/AttributesBindings.cpp
```diff
@@ -141,6 +141,11 @@ void initAttributesBindings(py::module& m) {
       m, "ObjectAttributes")
       .def(py::init(&ObjectAttributes::create<>))
       .def(py::init(&ObjectAttributes::create<const std::string&>))
+      .def_property_readonly_static(
+          "JSON_config_example",
```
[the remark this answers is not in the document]
erikwijmans: I agree about the capitalization here.  It should either be all caps or all lower case.  Python uses `json` everywhere, so there is no need to have it in caps.

src/esp/assets/attributes/ObjectAttributes.cpp
jturner65: I'm going to link it to the docs directly, and I did not want to introduce more files that would break tests if they were not around. :shrug:

jturner65: I was trying to kill 2 birds with one stone here, and perhaps I shouldn't be.  On the one hand, i want an easily maintained and obvious construction to use to test JSON parsing within the attributes managers (which I feel is better served with an internal static string within the attributes that can be modified when the class itself is extended). 

On the other hand, I wanted to provide a "json example" using the tags for documentation/.rst, while avoiding adding external files, to minimize the required support in the future for this functionality.

erikwijmans: Yeah, I am not sure if a string in the source code makes a good example.  You either have to load the python module and print the string or piece it together from the source to make a file to be loaded.

mosra: If it's just about showing that in the docs, you can reference this string using the `@snippet` Doxygen command (if you wrap it in some delimiters), or with more flexibility using `.. include::` in Python docs similarly as you did in Python tutorials. No need to put that in a dedicated file.

src/esp/assets/attributes/ObjectAttributes.cpp
```diff
@@ -17,6 +17,38 @@ const std::map<std::string, esp::assets::AssetType>
...
 
+const std::string AbstractObjectAttributes::JSONConfigTemplate =
+    R"({
+      "scale":[2,3,4],
+      "margin": 0.9,
+      "friction coefficient": 0.321,
+      "restitution coefficient": 0.456,
+      "requires lighting": false,
+      "units to meters": 1.1,
+      "up":[2.1,0,0],
```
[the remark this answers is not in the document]
jturner65: I was under the impression that up and front did not have to be unit vectors, just orthogonal, so I specifically picked test values that were not.  If this is incorrect, I can change it of course.

src/esp/assets/attributes/ObjectAttributes.cpp
aclegg3: I agree that files make sense for good examples, but we don't want users to accidentally find counter examples in our data folder, so it makes sense to hide those the code. We can always ref these into the doc rst files if we want to expose them.

src/esp/assets/attributes/ObjectAttributes.cpp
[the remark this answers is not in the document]
jturner65: I specifically picked non-unit vectors for json parsing testing purposes.

src/esp/assets/attributes/ObjectAttributes.cpp
jturner65: So how about I make good example files, and these weird "examples" i rename to be test-related, and remove the python bindings (since users would never really want to see these anyway)?

src/tests/IOTest.cpp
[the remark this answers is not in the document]
aclegg3: Then let's plan to continue using this test file for incoming json and xml for now.

src/esp/bindings/AttributesBindings.cpp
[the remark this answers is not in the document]
jturner65: I've removed these bindings, there's no reason to expose this structure to the user anymore.  I will instead construct example json files for each supported type.

mosra (approved):
I ran out of things to complain about, so ... I guess it's fine? :)

src/esp/assets/attributes/ObjectAttributes.cpp
```diff
@@ -17,6 +17,38 @@ const std::map<std::string, esp::assets::AssetType>
         {"suncg", AssetType::SUNCG_SCENE},
 };
 
+const std::string AbstractObjectAttributes::JSONConfigTestString =
```
eundersander, line 20: I would expect to find this kind of test data/code in the test files, not in the feature implementation and header files.

jturner65: I completely agree.  I only put this here to act as a kind of reminder, so that any future maintainers would know to expand the tests via this construction when they expand attribute management.

PR #775, opened 2020-09-01
Fix the depth buffer issue in PR752

bigbike:
## Motivation and Context
Tested the PR752 and found a couple of issues:
- In Linux: "imgui" did not display run-time info;
- In Linux: object picking functionality is broken;
- In Mac OSX: The rendering of the picked object is wrong. Occluded triangles would be displayed suggesting the depth buffer was broken.

This PR is to fix this problem.
Root cause:

For the broken object picking:
Now we display the observation via the sensor. A sensor will draw the observation to a framebuffer other than the default main buffer, and then blit the RGB result to the default buffer (NOTE: ONLY color info, NO depth info). However, we need the depth info in rendering the picked object.

For imgui:
The problem was that we did not re-bind the default framebuffer after we displayed the observation of the sensor.

Solution:
When an object is picked, we render the scene via the sensor first, to its framebuffer, the same as before. However, we deferred the blitting. Then we render the highlighted, picked object to the SAME framebuffer, and blit the results to the default main buffer to display.

This is also an alternative solution I can think of:
render the scene via the sensor and blit the RGB result to the screen. Also blit the depth texture from the RenderTarget to the default main buffer so that we can have the depth info. Compared to the current method, this will do one extra pass to copy the depth info, and thus potentially affect the performance.
(Improvement: map the depth texture to a quad which covers the whole screen, render this quad to the depth buffer of the default main buffer. But I found this will cause a lot of code changes if I am correct.)

@mosra : Thoughts? Do you think the current trade-off solution is good to go?

Before on linux: (no imgui, cannot select anything)

Before on Mac OSX (pay attention to the sofa as part of the wall is displayed on top of the sofa)
bad

Now on linux (with the fix):
good-2

Now on Mac OSX (with the fix):
Screen Shot 2020-09-01 at 1 10 03 AM

## Types of changes

## Checklist

### DIFF
--- a/src/esp/sensor/PinholeCamera.h
+++ b/src/esp/sensor/PinholeCamera.h
@@ -53,6 +53,14 @@ class PinholeCamera : public VisualSensor {
   virtual Corrade::Containers::Optional<Magnum::Vector2> depthUnprojection()
       const override;
 
+  /**
+   * @brief Draw an observation to the frame buffer using simulator's renderer
+   * @return true if success, otherwise false (e.g., frame buffer is not set)
+   * @param[in] sim Instance of Simulator class for which the observation needs
+   *                to be drawn
+   */
+  virtual bool drawObservationToFramebuffer(sim::Simulator& sim) override;
+
  protected:
   // projection parameters
   int width_ = 640;      // canvas width
@@ -62,13 +70,6 @@ class PinholeCamera : public VisualSensor {
   float hfov_ = 35.0f;   // field of vision (in degrees)
 
   ESP_SMART_POINTERS(PinholeCamera)
-
-  /**
-   * @brief Draw an observation using simulator's renderer
-   * @param[in] sim Instance of Simulator class for which the observation needs
-   *                to be drawn
-   */
-  void drawObservation(sim::Simulator& sim);
 
   /**
    * @brief Read the observation that was rendered by the simulator
--- a/src/esp/sensor/PinholeCamera.h
+++ b/src/esp/sensor/PinholeCamera.h
@@ -59,7 +59,7 @@ class PinholeCamera : public VisualSensor {
    * @param[in] sim Instance of Simulator class for which the observation needs
    *                to be drawn
    */
-  virtual bool drawObservationToFramebuffer(sim::Simulator& sim) override;
+  virtual bool drawObservation(sim::Simulator& sim) override;
 
  protected:
   // projection parameters
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -29,6 +29,9 @@
 #include "esp/scene/SceneNode.h"
 
 #include <Corrade/Utility/Arguments.h>
+#include <Corrade/Utility/Assert.h>
+#include <Corrade/Utility/Debug.h>
+#include <Corrade/Utility/DebugStl.h>
 #include <Corrade/Utility/Directory.h>
 #include <Corrade/Utility/String.h>
 #include <Magnum/DebugTools/Screenshot.h>
@@ -190,6 +193,9 @@ Key Commands:
   // NOTE: Mouse + shift is to select object on the screen!!
   void createPickedObjectVisualizer(unsigned int objectId);
   std::unique_ptr<ObjectPickingHelper> objectPickingHelper_;
+  // returns the number of visible drawables (meshVisualizer drawables are not
+  // included)
+  uint32_t displayObservatationWithPickedObject();
 };
 
 Viewer::Viewer(const Arguments& arguments)
@@ -468,6 +474,40 @@ void Viewer::wiggleLastObject() {
                              existingObjectIDs.back());
 }
 
+uint32_t Viewer::displayObservatationWithPickedObject() {
+  // ONLY draw the content to the frame buffer but not immediately blit the
+  // result to the default main buffer
+  simulator_->drawObservationToFramebuffer(defaultAgentId_, "rgba_camera");
+  uint32_t visibles = renderCamera_->getPreviousNumVisibileDrawables();
+
+  // we need to immediately draw picked object to the SAME frame buffer
+  // so bind it first
+  esp::gfx::RenderTarget* sensorRenderTarget =
+      simulator_->getRenderTarget(defaultAgentId_, "rgba_camera");
+  CORRADE_ASSERT(sensorRenderTarget,
+                 "Error in Viewer::displayObservatationWithPickedObject: "
+                 "sensor's rendering target "
+                 "cannot be nullptr.",
+                 0);
+  Mn::GL::Framebuffer& framebuffer = sensorRenderTarget->getFramebuffer();
+  framebuffer.bind();
+
+  // setup blending function
+  Mn::GL::Renderer::enable(Mn::GL::Renderer::Feature::Blending);
+
+  // render the picked object on top of the existing contents
+  esp::gfx::RenderCamera::Flags flags;
+  if (simulator_->isFrustumCullingEnabled()) {
+    flags |= esp::gfx::RenderCamera::Flag::FrustumCulling;
+  }
+  renderCamera_->draw(objectPickingHelper_->getDrawables(), flags);
+
+  Mn::GL::Renderer::disable(Mn::GL::Renderer::Feature::Blending);
+  sensorRenderTarget->blitRgbaToDefault();
+
+  return visibles;
+}
+
 float timeSinceLastSimulation = 0.0;
 void Viewer::drawEvent() {
   Mn::GL::defaultFramebuffer.clear(Mn::GL::FramebufferClear::Color |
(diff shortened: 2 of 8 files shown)


bigbike:
NOTE: if this PR is committed in, it will be merged into PR752.

src/esp/sensor/PinholeCamera.h
```diff
@@ -53,6 +53,14 @@ class PinholeCamera : public VisualSensor {
   virtual Corrade::Containers::Optional<Magnum::Vector2> depthUnprojection()
       const override;
 
+  /**
+   * @brief Draw an observation to the frame buffer using simulator's renderer
+   * @return true if success, otherwise false (e.g., frame buffer is not set)
+   * @param[in] sim Instance of Simulator class for which the observation needs
+   *                to be drawn
+   */
+  virtual bool drawObservationToFramebuffer(sim::Simulator& sim) override;
```
mosra, line 62: I admit I don't use these APIs daily, but after looking around a bit, the distinction between `drawObservation()` and `displayObservation()` seems clear enough already, no need to append `ToFramebuffer` to that. 

This also breaks the consistency -- now there's `drawObservationToFramebuffer()`  which is public and virtual, but `readObservation()` is still protected, which begs for questions like "why is one needed outside but the other is not? is reading not done from a framebuffer?". Additionally, from experience, an API suffixed `ToFramebuffer` implies that I could pass the framebuffer via a parameter to the function, but that's not the case here.

A counterproposal -- rewrite the `Sensor::displayObservation()` virtual to something that's more flexible so we don't need to have another virtual doing mostly the same. Here's an initial patch that should give you the idea:

```patch
diff --git a/src/esp/gfx/RenderTarget.h b/src/esp/gfx/RenderTarget.h
index eb48342b..cbfb4de3 100644
--- a/src/esp/gfx/RenderTarget.h
+++ b/src/esp/gfx/RenderTarget.h
@@ -96,10 +96,13 @@ class RenderTarget {
   void readFrameObjectId(const Magnum::MutableImageView2D& view);
 
   /**
-   * @brief Blits the rgba buffer from internal FBO to default frame buffer
-   * which in case of EmscriptenApplication will be a canvas element.
+   * @brief Blit the RGBA buffer from internal FBO to given frame buffer
+   *
+   * Pass @ref Magnum::GL::defaultFramebuffer to blit to the default
+   * framebuffer. Expects that the framebuffer has the same viewport size as
+   * @ref framebufferSize().
    */
-  void blitRgbaToDefault();
+  void blitRgbaTo(Magnum::GL::AbstractFramebuffer& framebuffer);
 
   // @brief Delete copy Constructor
   RenderTarget(const RenderTarget&) = delete;
diff --git a/src/esp/sensor/PinholeCamera.cpp b/src/esp/sensor/PinholeCamera.cpp
index 1752c27a..a6abacae 100644
--- a/src/esp/sensor/PinholeCamera.cpp
+++ b/src/esp/sensor/PinholeCamera.cpp
@@ -152,13 +152,13 @@ void PinholeCamera::readObservation(Observation& obs) {
   }
 }
 
-bool PinholeCamera::displayObservation(sim::Simulator& sim) {
+bool PinholeCamera::drawObservationTo(sim::Simulator& sim, Mn::GL::AbstractFramebuffer* framebuffer) {
   if (!hasRenderTarget()) {
     return false;
   }
 
   drawObservation(sim);
-  renderTarget().blitRgbaToDefault();
+  if(framebuffer) renderTarget().blitRgbaTo(*framebuffer);
 
   return true;
 }
diff --git a/src/esp/sensor/Sensor.h b/src/esp/sensor/Sensor.h
index 00e2bf95..448b0f2a 100644
--- a/src/esp/sensor/Sensor.h
+++ b/src/esp/sensor/Sensor.h
@@ -104,12 +104,15 @@ class Sensor : public Magnum::SceneGraph::AbstractFeature3D {
   virtual bool getObservationSpace(ObservationSpace& space) = 0;
 
   /**
-   * @brief Display next observation from Simulator on default frame buffer
-   * @param[in] sim Instance of Simulator class for which the observation needs
-   *                to be displayed
+   * @brief Draw next observation from Simulator to given frame buffer
+   * @param sim   Instance of Simulator class for which the observation needs
+   *      to be displayed
+   * @param framebuffer Framebuffer to blit the observation to. If `nullptr`,
+   *    the observation is drawn only to sensor's internal
+   *    @ref gfx::RenderTarget without blitting it anywhere.
    * @return Whether the display process was successful or not
    */
-  virtual bool displayObservation(sim::Simulator& sim) = 0;
+  virtual bool drawObservationTo(sim::Simulator& sim, Magnum::GL::AbstractFramebuffer* framebuffer) = 0;
 
  protected:
   SensorSpec::ptr spec_ = nullptr;
```

This of couse means you need to change all `displayObservation()` call sites to pass the framebuffer as well, but I think that'll pay off in the future, as some platforms (VR is one of them) don't have a concept of a default framebuffer anyway.

Then you also don't need to have an additional `Simulator::drawObservationToFramebuffer()` as it would be handled by the (rewritten) `Simulator::drawObservationTo()` as well.

src/utils/viewer/viewer.cpp
```diff
@@ -481,8 +521,16 @@ void Viewer::drawEvent() {
...
+  uint32_t visibles = 0;
+  if (objectPickingHelper_->isObjectPicked()) {
+    visibles = displayObservatationWithPickedObject();
+  } else {
+    simulator_->displayObservation(defaultAgentId_, "rgba_camera");
+    visibles = renderCamera_->getPreviousNumVisibileDrawables();
+  }
+  // Immediately bind the main buffer back so that the "imgui" below can work
+  // properly
+  Mn::GL::defaultFramebuffer.bind();
```
mosra, line 533: After the changes I proposed above, the contents of `displayObservationWithPickedObject()` could be pasted directly here, deduplicating the code quite a lot:

```suggestion
  // draw, but don't blit it to the default framebuffer just yet
  simulator_->drawObservationTo(defaultAgentId_, "rgba_camera", nullptr);
  visibles = renderCamera_->getPreviousNumVisibileDrawables();

  // the render target should be always here
  esp::gfx::RenderTarget* sensorRenderTarget =
      simulator_->getRenderTarget(defaultAgentId_, "rgba_camera");
  CORRADE_INTERNAL_ASSERT(sensorRenderTarget);
  
  // draw the picked object to the same frame buffer for proper depth test
  if (objectPickingHelper_->isObjectPicked()) {
    sensorRenderTarget->renderReEnter();    
    
    Mn::GL::Renderer::enable(Mn::GL::Renderer::Feature::Blending);

    // blend the picked object on top of the existing contents
    esp::gfx::RenderCamera::Flags flags;
    if (simulator_->isFrustumCullingEnabled()) {
      flags |= esp::gfx::RenderCamera::Flag::FrustumCulling;
    }
    renderCamera_->draw(objectPickingHelper_->getDrawables(), flags);

    Mn::GL::Renderer::disable(Mn::GL::Renderer::Feature::Blending);
    
    // bind the main buffer back so the rest works properly
    sensorRenderTarget->renderExit();    
    Mn::GL::defaultFramebuffer.bind();
  }
  
  // finally, blit the observation (and optional picked object on top) to the 
  // default framebuffer
  sensorRenderTarget->blitRgbaTo(Mn::GL::defaultFramebuffer);
```

mosra (changes requested):
The code fixes the issue, yes, but I feel the same could be achieved through *simplification* of the original code, not by adding an extra layer of extremely specific APIs on top, which would be there only for the viewer and nothing else.

I hope my suggestions are clear enough.

mosra:
Re other approaches:

> render the scene via the sensor and blit the RGB result to the screen. Also blit the depth texture from the RenderTarget to the default main buffer so that we can have the depth info. Compared to the current method, this will do one extra pass to copy the depth info, and thus potentially affect the performance.

Yep, this is always more work and thus never faster. The essence of what you did here (rendering to the offscreen FB, reusing its depth buffer, and then blitting just color) makes more sense.

> (Improvement: map the depth texture to a quad which covers the whole screen, render this quad to the depth buffer of the default main buffer. But I found this will cause a lot of code changes if I am correct.)

Why do you think this would be faster than simply blitting the depth? Since both framebuffers have the same size, same viewport and (most probably) the exact same format, the blit would be just a plain memory copy. Rendering a quad / fullscreen triangle is a lot more work involving the rasterizer, scheduling shader cores, texture filtering etc. There's a reason why extensions like [NV_draw_texture](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_draw_texture.txt) exist -- see the `Issues` section on the bottom for juicy details.

bigbike:
> but I feel the same could be achieved through _simplification_ of the original code, not by adding an extra layer of extremely specific APIs on top, which would be there only for the viewer and nothing else.

True, I agree. I am not going to push back because your comments (together the inline comments) made a lot of sense to me. Here I simply explain why I did it in this way: We need to do it fast and do it in the safest way.

**Fast:**We need to fix it fast. #752 cannot be committed in due to this bug. However, we need #752 badly since it is scheduled in the next release (very soon). So I have to come to a very light, quick solution (even though this is not my style also) instead of a large refactoring PR.
**Safe:** Again we aim for a release. Stability is the top priority. So far the way you described in the PR make a lot of sense, but it requires to change a lot of existing functions, python bindings, test code (perhaps new tests are needed), and needs a lot of time to debug and test, making sure no other existing functions are broken. It is too risky.

So I decided to make a trade-off here, update the code. It only partially addressed your concerns.
Again, your comments are highly valuable. We will revisit them after this release.

Specifically what I have changed in the updated version:
- Removed the getFramebuffer, and added the renderReEnter()
- I moved the contents in `displayObservationWithPickedObject` to the render function

what I did not do, but I think your comments make sense and we can revisit:
- remove displayObservation. Use drawObservation instead.

aclegg3:
> So I decided to make a trade-off here, update the code. It only partially addressed your concerns.
> ...
> Specifically what I have changed in the updated version

Thanks for solving this, @bigbike.  I think these changes look like a good patch for the issue and a middle ground between the adhoc API and the larger refactor.

> Again, your comments are highly valuable. We will revisit them after this release.

Agreed. Thanks for the thought-out long-term solution @mosra. Let's revisit this refactor after the release. 👍

mosra:
Eh, okay -- but you know how it goes, "refactor later" is usually "refactor never" :)

I'd vote for adding this refactor to the Tasks board so it doesn't get forgotten. In my case at least it often happens that a collection of such smallish tasks over time forms a general design idea for a later refactor (it was the case with the new MeshData / MaterialData, for example).

PR #783, opened 2020-09-09
--[Bug Fix] Verify AttributesManager template type is valid

jturner65:
Previously, the attributes managers specialized the class template by specifying a smart pointer to an attributes class as the class template type parameter.  This relied on the implementer knowing what was required by the class template, method -wise, in order for the manager to properly function.  If an inappropriate class was specified, the class template functionality would fail.

This PR introduces a verification process to make sure the class template type parameter references a class that has the required methods for the manager to function.  This is accomplished by sending the actual attributes class as the template type to specialize the class template.  This allows us to verify that the specified template class is derived from the desired base class (so we can be sure that the objects we wish to manage with a particular manager implement the necessary functions for the attributes managers to work correctly).  If an illegal class is passed now as a template parameter, the code will not compile.

## How Has This Been Tested
All c++ and python tests have passed

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/managers/AttributesManagerBase.h
+++ b/src/esp/assets/managers/AttributesManagerBase.h
@@ -45,7 +45,7 @@ template <class T>
 class AttributesManager {
  public:
   static_assert(std::is_base_of<Attrs::AbstractAttributes, T>::value,
-                "Managed typedd must be derived from AbstractAttributes");
+                "Managed type must be derived from AbstractAttributes");
 
   typedef std::shared_ptr<T> AttribsPtr;
 
@@ -917,8 +917,7 @@ class AttributesManager {
 // Class Template Method Definitions
 
 template <class T>
-typename AttributesManager<T>::AttribsPtr
-AttributesManager<T>::createObjectAttributesFromJson(
+std::shared_ptr<T> AttributesManager<T>::createObjectAttributesFromJson(
     const std::string& configFilename,
     const io::JsonDocument& jsonDoc) {
   auto attributes = this->initNewAttribsInternal(configFilename);
@@ -1004,7 +1003,7 @@ AttributesManager<T>::createObjectAttributesFromJson(
 
 template <class T>
 bool AttributesManager<T>::setJSONAssetHandleAndType(
-    typename AttributesManager<T>::AttribsPtr attributes,
+    AttribsPtr attributes,
     const io::JsonDocument& jsonDoc,
     const char* jsonMeshTypeTag,
     const char* jsonMeshHandleTag,
@@ -1078,7 +1077,7 @@ bool AttributesManager<T>::setTemplateLock(const std::string& templateHandle,
 }  // AttributesManager::setTemplateLock
 
 template <class T>
-std::vector<typename AttributesManager<T>::AttribsPtr>
+std::vector<std::shared_ptr<T>>
 AttributesManager<T>::removeTemplatesBySubstring(const std::string& subStr,
                                                  bool contains) {
   std::vector<AttribsPtr> res;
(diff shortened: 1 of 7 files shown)


src/esp/assets/managers/AttributesManagerBase.h
```diff
@@ -41,9 +41,14 @@ namespace Attrs = esp::assets::attributes;
  * references
  */
 
-template <class AttribsPtr>
+template <class T>
 class AttributesManager {
  public:
+  static_assert(std::is_base_of<Attrs::AbstractAttributes, T>::value,
+                "Managed typedd must be derived from AbstractAttributes");
```
mosra, line 48:
```suggestion
                "Managed type must be derived from AbstractAttributes");
```

src/esp/assets/managers/AttributesManagerBase.h
```diff
@@ -998,7 +1004,7 @@ AttribsPtr AttributesManager<AttribsPtr>::createObjectAttributesFromJson(
 
 template <class T>
 bool AttributesManager<T>::setJSONAssetHandleAndType(
-    T attributes,
+    typename AttributesManager<T>::AttribsPtr attributes,
```
mosra, line 1007: Isn't `AttribsPtr` enough? :thinking:

mosra (commented):
*Technically* you could do the `static_assert` even in the original variant together with checking that it's actually a pointer, something like

```cpp
static_assert(
    std::is_base_of<Attrs::AbstractAttributes, typename AttribsPtr::element_type>::value &&
    std::is_same<std::shared_ptr<typename AttribsPtr::element_type>, AttribsPtr>::value, 
    ...);
```

... but using the underlying type directly instead of the wrapped pointer type makes more sense to me, as there's less type layers and things to go wrong :)

(Disclaimer: I know almost nothing about the rest of this code, so my reviewing capabilities are very narrow.)

src/esp/assets/managers/AttributesManagerBase.h
jturner65: I'll try it.

jturner65: Apparently it is, thanks.  Need to qualify all the return types, but the argument type didn't need it.

mosra: The return types in _out-of-class_ definitions would work only as trailing I think (`auto Class::foo() -> AttribsPtr`), the way you have them you need to prefix with class (and then with `typename`). For in-class declarations `AttribsPtr foo()` works, yes.

jturner65: I can just specify std::shared_ptr<T>, since that's what AttribsPtr is

mosra: or that, yep, depends how you want it to appear in the docs -- if there's a typedef, i would personally use the typedef everywhere :)

jturner65: should i put them back?  seemed a bit cumbersome, but i'll differ to your judgement :)

src/esp/assets/managers/AttributesManagerBase.h
```diff
@@ -1096,9 +1095,9 @@ AttributesManager<T>::removeTemplatesBySubstring(const std::string& subStr,
 }  // removeAllTemplates
 
 template <class T>
-typename AttributesManager<T>::AttribsPtr
-AttributesManager<T>::removeTemplateInternal(const std::string& templateHandle,
-                                             const std::string& sourceStr) {
+std::shared_ptr<T> AttributesManager<T>::removeTemplateInternal(
+    const std::string& templateHandle,
+    const std::string& sourceStr) {
```
mosra: Does this work? In my opinion it's consistent without being excessively long, but other people have a strong dislike for trailing return types, so it depends.

```cpp
auto AttributesManager<T>::removeTemplateInternal(
    const std::string& templateHandle,
    const std::string& sourceStr) -> AttribsPtr {
```

jturner65: Oh, i like that.

aclegg3 (approved):

PR #785, opened 2020-09-10
Update sim submodules

Skylion007:
## Motivation and Context

* This updates some of our submodules that don't relate to Magnum directly, and therefore haven't been updated in while. 
* This only updates submodules that can be incremented by minor versions without API changes.
* Changes include better cmake/compiler support and better performance as well as supporting Python 3.10 and beyond.
* Also changes the eigen-git-mirror location since they are moving their primary source code to Gitlab and their Github repo will soon be deleted.
* This also uncovered a clang-tidy bug found in OBB.cpp and patches it.

## How Has This Been Tested
Everything builds, tests pass.

## Types of changes

## Checklist


erikwijmans (approved):
If it builds it ships

PR #786, opened 2020-09-11
--[Bug fix] Address stage load w/"none" physicsManager

jturner65:
RigidStage objects were not being made if there was no collision asset being constructed.  This would only happen if physicsManager was nullptr (which only happens during certain tests) or if none-type physicsManager is specified.  
This makes RigidStage objects if collision assets are not requested, and will attempt to build collisionMeshGroups out of render mesh if no valid collision mesh is specified. 

A warning for an unused argument in VisualSensor is also addressed.

## How Has This Been Tested
C++ and Python tests have passed

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -195,7 +195,6 @@ bool ResourceManager::loadStage(
   // collision mesh next
   bool renderMeshSuccess =
       loadStageInternal(renderInfo,         // AssetInfo
-                        nullptr,            // physics manager
                         &rootNode,          // parent scene node
                         &drawables,         //  drawable group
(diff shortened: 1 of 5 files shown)


jturner65:
The only thing we must yet address here is if we wish to support a "none" stage with a non-kinematic physics environment.  Note :  we do not allow objects to exist with neither render nor collision assets (Which is basically what the "none" stage is) so an argument could be made for not allowing "the void", at least in a non-kinematic environment.

aclegg3 (approved):
Looks good to me, thanks for the fix! We'll handle the "none" scene by checking for nullptr initialization attributes for now. I think we need to keep this capability.

PR #790, opened 2020-09-12
Don't load textures if you don't need them

erikwijmans:
## Motivation and Context

Textures take a lot of GPU memory but we don't need them unless we are doing RGB rendering.  So don't load them if we aren't!

I also removed the compressTextures_ flag while I was poking around.  That is certainly deprecated as it was subsumed by basis and it never worked on Gibson (those textures are too big).

One thing this has revealed is that is likely time to get rid of the hack of sharing the simulator between tests to make them run faster.  That will make tests right slightly slower but at least it will put pressure on us to make spin-up time as fast as possible?

## How Has This Been Tested

Tests pass.  You can tell it is not loading textures by GPU memory usage and just pure load time :)

## Types of changes

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -991,8 +991,10 @@ bool ResourceManager::loadGeneralMeshData(
 
     // if this is a new file, load it and add it to the dictionary
     LoadedAssetData loadedAssetData{info};
-    loadTextures(*fileImporter_, loadedAssetData);
-    loadMaterials(*fileImporter_, loadedAssetData);
+    if (requiresTextures_) {
+      loadTextures(*fileImporter_, loadedAssetData);
+      loadMaterials(*fileImporter_, loadedAssetData);
+    }
     loadMeshes(*fileImporter_, loadedAssetData);
     auto inserted = resourceDict_.emplace(filename, std::move(loadedAssetData));
     MeshMetaData& meshMetaData = inserted.first->second.meshMetaData;
@@ -1283,9 +1285,13 @@ void ResourceManager::loadMeshHierarchy(Importer& importer,
   if (objectData->instanceType() == Magnum::Trade::ObjectInstanceType3D::Mesh &&
       meshIDLocal != ID_UNDEFINED) {
     parent.children.back().meshIDLocal = meshIDLocal;
-    parent.children.back().materialIDLocal =
-        static_cast<Magnum::Trade::MeshObjectData3D*>(objectData.get())
-            ->material();
+    if (requiresTextures_) {
+      parent.children.back().materialIDLocal =
+          static_cast<Magnum::Trade::MeshObjectData3D*>(objectData.get())
+              ->material();
+    } else {
+      parent.children.back().materialIDLocal = ID_UNDEFINED;
+    }
   }
 
   // Recursively add children
@@ -1338,12 +1344,6 @@ void ResourceManager::loadTextures(Importer& importer,
       Mn::GL::TextureFormat format;
       if (image->isCompressed()) {
         format = Mn::GL::textureFormat(image->compressedFormat());
-      } else if (compressTextures_ &&
-                 image->format() == Mn::PixelFormat::RGBA8Unorm) {
-        format = Mn::GL::TextureFormat::CompressedRGBAS3tcDxt1;
-      } else if (compressTextures_ &&
-                 image->format() == Mn::PixelFormat::RGB8Unorm) {
-        format = Mn::GL::TextureFormat::CompressedRGBS3tcDxt1;
       } else {
         format = Mn::GL::textureFormat(image->format());
       }
--- a/src/esp/gfx/Renderer.h
+++ b/src/esp/gfx/Renderer.h
@@ -14,8 +14,17 @@ namespace gfx {
 
 class Renderer {
  public:
-  Renderer();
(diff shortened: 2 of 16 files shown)


Skylion007:
YES!!! I was just thinking about taking a stab at this the other day since we use depth only rendering so much.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1338,12 +1344,6 @@ void ResourceManager::loadTextures(Importer& importer,
       Mn::GL::TextureFormat format;
       if (image->isCompressed()) {
         format = Mn::GL::textureFormat(image->compressedFormat());
-      } else if (compressTextures_ &&
```
Skylion007, line 1341: Why did we remove this code fragment?

Skylion007: Maybe the naming of the boolean is just bad, but we shouldn't we be able to require textures AND compress them?

tests/test_sensors.py
```diff
@@ -117,6 +117,8 @@ def test_sensors(
     if not habitat_sim.cuda_enabled and gpu2gpu:
         pytest.skip("Skipping GPU->GPU test")
 
+    sim.close()
```
Skylion007, line 120: We can do tuse `with habitat_sim as sim` to make the scope more clear here

erikwijmans: That will require re-factoring the tests to no share an instance of the simulator.  Which we should do, but best not in this PR imo.

Skylion007:
> One thing this has revealed is that is likely time to get rid of the hack of sharing the simulator between tests to make them run faster. That will make  tests right slightly slower but at least it will put pressure on us to make spin-up time as fast as possible?

Agree, doing it on my Hab-Lab PR as well. Optimizing Simulator startup is probably the next frontier on optimizing HabSim TBH.

src/esp/gfx/Renderer.h
```diff
@@ -15,6 +15,7 @@ namespace gfx {
 class Renderer {
  public:
   Renderer();
+  explicit Renderer(bool hasTextures);
```
mosra, line 18: Just a single overload? Would also implicitly document the default state.

```suggestion
  explicit Renderer(bool hasTextures = true);
```

Tho ... from an API usability standpoint, I'd prefer to not have a bool argument and instead do something like

```cpp
enum class Flag {
    NoTextures = 1 << 0,
};

typedef Corrade::Containers::EnumSet<Flag> Flags;
CORRADE_ENUMSET_FRIEND_OPERATORS(Flags)

explicit Renderer(Flags flags = {});
```

because then `Renderer(Renderer::Flag::NoTextures)` is much more descriptive than `Renderer(false)`.

src/esp/sim/Simulator.h
```diff
@@ -765,6 +773,13 @@ class Simulator {
   int navMeshVisPrimID_ = esp::ID_UNDEFINED;
   esp::scene::SceneNode* navMeshVisNode_ = nullptr;
 
+  /**
+   * @brief Tracks whether or not the simulator was initialized
+   * to load textures.  Because we cache mesh loading, this should
+   * *not* be changed without calling close() first
+   */
+  int requiresTextures_ = -1;
```
mosra, line 781: Minor suggestion: you could use `Containers::Optional<bool>` for a cleaner "tri bool" state (unknown / true / false) :)

Skylion007:
@erikwijmans The build error is that the changes to renderer construction broke the cpp tests for culling.

src/esp/sim/Simulator.h
erikwijmans: I like it!

src/esp/gfx/Renderer.h
erikwijmans: Makes sense.

Skylion007 (approved):

Skylion007:
LGTM, this is a great performance optimization. Especially since using only the Depth Sensor is such a common configuration.

jturner65 (approved):
LGTM

src/esp/assets/ResourceManager.cpp
```diff
@@ -991,8 +991,10 @@ bool ResourceManager::loadGeneralMeshData(
 
     // if this is a new file, load it and add it to the dictionary
     LoadedAssetData loadedAssetData{info};
-    loadTextures(*fileImporter_, loadedAssetData);
-    loadMaterials(*fileImporter_, loadedAssetData);
+    if (requiresTextures_) {
+      loadTextures(*fileImporter_, loadedAssetData);
+      loadMaterials(*fileImporter_, loadedAssetData);
+    }
```
aclegg3, line 997: Just a note that `Simulator.reconfigure` won't reload these after skipping them if the configuration changes without first calling `close()`.

bigbike (approved):
@mosra: 
Related but totally different topic inspired by this diff:
It would be nice if the draw(...) function can sometimes just render the geometry of the mesh, without textures (even they exist), namely, the phong shader is acting like a flat shader. It can be very useful, e.g., in the object picking, depth rendering. etc.

mosra:
> It would be nice if the draw(...) function can sometimes just render the geometry of the mesh, without textures

@bigbike sure, that's possible and easy to do on the application side. Ask the shader manager for a different instance that has no textures enabled (or has zero lights, for a totally flat look) ... and don't call any of the `bind*Texture()` functions, since those would assert now.

PR #791, opened 2020-09-13
No shared sim

erikwijmans:
## Motivation and Context

It's time to get rid of the sharing the instance of the simulator among all the tests hack.  I also added an explicit test that the current test_sensors was implicitly testing -- making sure reconfigure actually swaps the scene correctly.

This also allows us to enable parallel testing through https://github.com/browsertron/pytest-parallel so total test time is actually reduced!

## How Has This Been Tested

Via the tests

## Types of changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -470,7 +470,7 @@ jobs:
               export PATH=$HOME/miniconda/bin:$PATH
               . activate habitat; cd habitat-lab
               python setup.py develop --all
-              python setup.py test
+              pytest --workers 4 --verbose -rsxX -q
       - save_cache:
           key: habitat-lab-{{ checksum "./hablab_sha" }}
           background: true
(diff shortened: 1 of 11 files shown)


Skylion007:
@erikwijmans You need to add the pytest-plugin here as well: https://github.com/facebookresearch/habitat-sim/blob/1728d6e7d7e5583f1634945e66182769c414bb86/setup.py#L383

erikwijmans:
> @erikwijmans You need to add the pytest-plugin here as well:

I didn't add any flags that plugin for `setup.py test` so I don't it is needed.

Skylion007:
But doing it will automatically install the pytest plugins, so it should still be done for the end user.

erikwijmans:
> But doing it will automatically install the pytest plugins, so it should still be done for the end user.

But why install something they don't need?  That plugin does nothing by default, so unless they know to specify the flag (and want to specify it), it won't do anything.  I didn't add anything by default as I don't want to force everyone to run parallel testing locally.

Skylion007:
Fair enough, I guess it's not necessary like the other pytest ones.

Skylion007 (approved):

jturner65 (approved):
LGTM

PR #796, opened 2020-09-14
Object Attributes allow for primitives as asset handles

jturner65:
## Motivation and Context
This supports using primitive asset attributes tags as collision or render mesh names in object JSON configuration files.
To use this, the exact name of the primitive asset attributes template must be used in the object attributes template as either the render or collision mesh.  

So, for example, to use a sphere primitive as the collision asset, use "uvSphereSolid_rings_8_segments_16_useTexCoords_false_useTangents_false" as the "collision mesh" value in the object's configuration JSON file.

Also, another tag has been added, "collision asset size", which is a 3-vector to enable resizing the primitive to match the render asset.  So if the render asset has is a sphere of radius .1 meters, then this tag should be :
 "collision asset size" : [0.1,0.1,0.1]

## How Has This Been Tested
all c++ and python tests passed

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/attributes/ObjectAttributes.cpp
+++ b/src/esp/assets/attributes/ObjectAttributes.cpp
@@ -56,6 +56,7 @@ AbstractObjectAttributes::AbstractObjectAttributes(
   setFrictionCoefficient(0.5);
   setRestitutionCoefficient(0.1);
   setScale({1.0, 1.0, 1.0});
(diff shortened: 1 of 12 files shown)


aclegg3 (approved):
Looks like it solves the immediate problem. We'll re-visit geometry specification formats in more detail in the future. 👍

Skylion007 (approved):

PR #797, opened 2020-09-15
Disabled running version_conda_release on every master commit

mathfac:
## Motivation and Context
After latest change we noticed that Conda build step was triggered both for tagged commit and for the one merged into master:
Screen Shot 2020-09-14 at 6 50 46 PM

 Based on [CircelCI documentation](https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag) that's the right way to disabled running version_conda_release on every master commit as `filters` section is union not intersection of branch and tag filters.

## How Has This Been Tested
Hard to tests without landing.

## Types of changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -591,8 +591,6 @@ workflows:
       - install_and_test_ubuntu
       - pre-commit
   nightly:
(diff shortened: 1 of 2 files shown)


aclegg3 (approved):
Let's try it 👍

PR #800, opened 2020-09-15
Move Attributes and Attributes Managers to their own directory

jturner65:
## Motivation and Context
This PR is to move the attributes and attributes managers to their own directory, esp/metadata/, instead of crowding esp/assets/.
Along with relieving the crowding of assets, this also lays the groundwork for the metadataMediator class, which will reside in the metadata directory, and the library and scene attributes loading and management.  

## How Has This Been Tested
C++ and python tests pass

## Types of changes

## Checklist


src/esp/metadata/managers/StageAttributesManager.h
```diff
@@ -2,8 +2,8 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-#ifndef tageESP_ASSETS_MANAGERS_STAGEATTRIBUTEMANAGER_H_
-#define tageESP_ASSETS_MANAGERS_STAGEATTRIBUTEMANAGER_H_
+#ifndef ESP_METADATA_MANAGERS_STAGEATTRIBUTEMANAGER_H_
```
bigbike, line 5: I believe we just use `#pragma once` everywhere to make sure each head file to be included only once during the compilation.

bigbike:
By the way, I would like to understand a bit more about `metadataMediator` class.
What is that for?

src/esp/metadata/managers/StageAttributesManager.h
jturner65: I don't use pragma once because there are certain (admittedly rare) instances where it can behave poorly.

jturner65: > By the way, I would like to understand a bit more about `metadataMediator` class.
> What is that for?

We discussed this briefly last week.  The MetaDataMediator class will own and manage all the attributes managers.  It will in turn be owned by simulator.

Alex and I put together a preliminary design document right here : 
https://docs.google.com/document/d/1IDNj-dpk9Zn0Pi82VoihlTP1E1bvL41FMfvQt5UCfi4/edit#

jturner65:
> By the way, I would like to understand a bit more about `metadataMediator` class.
> What is that for?

We discussed this briefly last week.  The MetaDataMediator class will own and manage all the attributes managers.  It will in turn be owned by simulator.

Alex and I put together a preliminary design document right here : 
https://docs.google.com/document/d/1IDNj-dpk9Zn0Pi82VoihlTP1E1bvL41FMfvQt5UCfi4/edit#

src/esp/metadata/managers/StageAttributesManager.h
jturner65: I also noticed it in google style guide : [The #define Guard](https://google.github.io/styleguide/cppguide.html#The__define_Guard)

bigbike: I understand `pragma once` is not perfect. But so far we never meet any problems to use it. The cases I know where it would fail can be easily avoided, and should not be the concern.

But most importantly, this is about the uniformed coding style & convention within the system. Before these files, we use `pragma once` everywhere, and it seems we do not have compelling reasons to change them.

This is not a big problem, but shall we make a quick decision, put it into our own coding convention, and just follow it in the future?
@aclegg3 @eundersander @mosra @erikwijmans

bigbike:
> > By the way, I would like to understand a bit more about `metadataMediator` class.
> > What is that for?
> 
> We discussed this briefly last week during our Tuesday meeting. The MetaDataMediator class will own and manage all the attributes managers. It will in turn be owned by simulator.
> 
> Alex and I put together a preliminary design document right here :
> https://docs.google.com/document/d/1IDNj-dpk9Zn0Pi82VoihlTP1E1bvL41FMfvQt5UCfi4/edit#
> 
> Please share your thoughts :)

Oh, great. Good to know! I missed it because I was in PTO.
Just sent the doc access requirement.

src/esp/metadata/managers/StageAttributesManager.h
jturner65: For what it's worth, really I just followed the google style guide.

erikwijmans: I don't have a strong preference.  I do like `#pragma once` better, but that is because I am lazy

Skylion007 (approved):

aclegg3 (approved):
This looks good to me. Happy to see no change in the python API. 👍

PR #805, opened 2020-09-17
Enabled 3.6, 3.7, 3.8 to conda build automation, removed date from release versioning, updated badged.

mathfac:
## Motivation and Context
Added head/headless options, enabled 3.6, 3.7, 3.8 to conda build automation

Regarding badges, propose to avoid deduplication and remove some of them: 
- Removed black badge as pre-commit badge is presented and if pre-commit installed black will be picked up, badge is important for contributors only
- Removed version release badge as version is already mentioned on Anaconda release and should be the same
- Added Docs badge as we want to focus users more that Documentation is available and it's alive
- Added python languages supported.  

## Types of changes

## Checklist


dhruvbatra (approved):

conda-build/Dockerfile
```diff
@@ -35,6 +34,7 @@ RUN bash ./install_conda.sh && rm install_conda.sh
 
 RUN anaconda login --username $AIHABITAT_CONDA_CHN --password $AIHABITAT_CONDA_CHN_PWD
 RUN conda create --name py36 python=3.6 -y && conda init bash
+RUN conda create --name py37 python=3.7 -y
```
erikwijmans, line 37: I don't believe this is needed.  `condo-build` takes are of setting up a py3.7 environment when you build a py3.7 package.

mathfac: @erikwijmans, good catch. Noticed that as well.

Skylion007: Can we also remove the python=3.6 as well?

Skylion007 (changes requested):
Why not enable builds for Python 3.8 and 3.9 as well?

Also we should update the badge in the ReadMe to show we support multiple Python versions.

Skylion007:
We probably also need to update the build matrix then to do to install_and_test_ubuntu for 3.7, right? (in parallel)

conda-build/Dockerfile
mathfac: No, we can't. That's root environment that we use to run conda build that has all dependency installed.

mathfac:
Adding `3.8` is good idea based on statistics from [Pytorch Conda repo](https://anaconda.org/pytorch/pytorch/files?sort=time&sort_order=desc), `3.9` maybe too early.
Regarding running testing for 3.7. I would rather add some integration tests for conda package builded than running all over tests on one more setup. Let's make conda packages testing separate PR.

mathfac:
Regarding badges, propose to avoid deduplication and remove some of them: 
- Removed black badge as pre-commit badge is presented and if pre-commit installed black will be picked up, badge is important for contributors only
- Removed version release badge as version is already mentioned on Anaconda release and should be the same
- Added Docs badge as we want to focus users more that Documentation is available and it's alive
- Added python languages supported.

Skylion007:
>> Removed black badge as pre-commit badge is presented and if pre-commit installed black will be picked up
How I wish I could actually enforce pre-commit being installed properly, but there are contributors who don't run it on every commit.

README.md
```diff
@@ -3,10 +3,9 @@
 [![GitHub license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/facebookresearch/habitat-sim/blob/master/LICENSE)
 [![Anaconda-Server Badge](https://anaconda.org/aihabitat/habitat-sim/badges/version.svg)](https://anaconda.org/aihabitat/habitat-sim)
 [![Anaconda-Server Badge](https://anaconda.org/aihabitat/habitat-sim/badges/platforms.svg)](https://anaconda.org/aihabitat/habitat-sim)
+[![Documentation](https://img.shields.io/badge/docs-automated-green.svg)](https://aihabitat.org/docs/habitat-sim/)
 [![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit&logoColor=white)](https://github.com/pre-commit/pre-commit)
-[![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
-[![GitHub release (latest by date)](https://img.shields.io/github/v/release/facebookresearch/habitat-sim)](https://github.com/facebookresearch/habitat-sim/releases/latest)
```
Skylion007, line 8: Strongly disagree with as the Anaconda Release version looks entirely different from the semantic versioning we actually using in setup.py and elsewhere.

README.md
```diff
@@ -3,10 +3,9 @@
...
-[![Python 3.6](https://img.shields.io/badge/python-3.6-blue.svg)](https://www.python.org/downloads/release/python-360/)
+[![Python 3.6](https://img.shields.io/badge/python-3.6%20%7C%203.7%20%7C%203.8-blue.svg)](https://www.python.org/downloads/release/)
```
Skylion007, line 8: Change the badge name here as well.

README.md
```diff
@@ -3,10 +3,9 @@
 [![GitHub license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/facebookresearch/habitat-sim/blob/master/LICENSE)
 [![Anaconda-Server Badge](https://anaconda.org/aihabitat/habitat-sim/badges/version.svg)](https://anaconda.org/aihabitat/habitat-sim)
 [![Anaconda-Server Badge](https://anaconda.org/aihabitat/habitat-sim/badges/platforms.svg)](https://anaconda.org/aihabitat/habitat-sim)
+[![Documentation](https://img.shields.io/badge/docs-automated-green.svg)](https://aihabitat.org/docs/habitat-sim/)
```
Skylion007, line 6: This should actually be a circleci badge that points to our automatic docs deployment job so we know if docs suddenly start failing.

Skylion007 (changes requested):
Issues with some of the badge stuff

README.md
mathfac: @Skylion007  thank you for the suggestion: I can't find [CircleCI tags](https://circleci.com/docs/2.0/status-badges/) or Shields.io supporting specific step in workflow status represented as the badge. If our CircleCI build badge is green => Docs build is green for sure.

README.md
mathfac: Can you elaborate more, here is the screen showing the versioning, here badge from Anaconda was cached by Github and will update:
Screen Shot 2020-09-18 at 10 45 28 AM
How different the versions are?

mathfac:
> How I wish I could actually enforce pre-commit being installed properly, but there are contributors who don't run it on every commit.

Then they will get CI failed as soon as they send PR and discover that.

mathfac:
Badges now:
Screen Shot 2020-09-18 at 10 49 10 AM

mathfac:
[Slate Anaconda badges is known issue](https://github.com/ContinuumIO/anaconda-issues/issues/10231), switch to shields.io badge instead.

mathfac:
Removed the date from versioning for releases, while left it for nightly builds.

README.md
Skylion007: Yeah you are right, that's a huge oversight. It shouldn't be automated green then, it should informative blue.

Skylion007 (approved):

README.md
mathfac: Discussed with Aaron separately, removed the date from release versioning.

mathfac:
9 packages produced. PR is ready to be merged.

PR #812, opened 2020-09-28
[DO NOT ACCEPT] PBR in progress

bigbike:
## Motivation and Context
Implementing PBR for early feedback

## Types of changes

## Checklist


src/esp/gfx/PBRDrawable.cpp
```diff
@@ -0,0 +1,141 @@
...
+      .setObjectId(
+          static_cast<RenderCamera&>(camera).useDrawableIds()
+              ? drawableId_
+              : (materialData_->perVertexObjectId ? 0 : node_.getSemanticId()))
+      .setTransformationMatrix(transformationMatrix)  // modelview matrix
+      .setMVPMatrix(camera.projectionMatrix() * transformationMatrix)
+      // because we actually do not have scaling, only rotation, so that the
+      // modelview matrix mv = inv(mv^T) where mv stands for the up-left 3x3
+      // part of the modelview matrix
+      .setNormalMatrix(transformationMatrix.rotationScaling())
```
mosra, line 54: Why not use [normalMatrix()](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Matrix4.html#aae3f4e7f98d6838ef7be4f70531a7583)? It's a *relatively* new thing, so you might not be aware of it yet. 

And it looks like `GenericDrawable` should receive [the same treatment](https://github.com/facebookresearch/habitat-sim/blob/6774fc4b12a06bf79eab6585606936004a643ca0/src/esp/gfx/GenericDrawable.cpp#L98).

src/esp/gfx/MaterialData.h
```diff
@@ -15,21 +15,42 @@
...
   ESP_SMART_POINTERS(PhongMaterialData)
 };
 
+struct PBRMaterialData : public MaterialData {
+  // use the value if albedoTexture does not exist
+  Magnum::Color4 baseColor{0.7};
+  // use the value if roughnessTexture does not exist
+  Magnum::Float roughness = 0.9f;
+  // use the value if metallicTexture does not exist
+  Magnum::Float metallic = 0.1f;
```
mosra, line 44: According to glTF spec, [these values are *multiplied* together](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#metallic-roughness-material), not "one or the other".

src/esp/gfx/PBRDrawable.cpp
```diff
@@ -0,0 +1,141 @@
...
+      .bindTextures(
+          materialData_->albedoTexture, materialData_->roughnessTexture,
+          materialData_->metallicTexture, materialData_->normalTexture)
+      .setBaseColor(materialData_->baseColor)
+      .setRoughness(materialData_->roughness)
+      .setMetallic(materialData_->metallic);
```
mosra, line 60: Just FYI, in Magnum I'm using "metalness" to be similar to "roughness", but it's a minor thing -- just if you'd like to have the exact same naming scheme here.

src/esp/gfx/PBRDrawable.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+  if (materialData_->textureMatrix != Mn::Matrix3{}) {
+    // TODO
+  }
```
mosra, line 64: I'd add a temporary assert here -- because when MeshOptimizer comes to play, this will be the case always, and models that don't have their texture transform respected will look *extremely wrong* and nobody would know why :)

Same for the TODO below.

src/esp/gfx/PBRDrawable.cpp
```diff
@@ -0,0 +1,141 @@
...
+    CORRADE_INTERNAL_ASSERT(shader_ && shader_->lightCount() == lightCount &&
+                            shader_->flags() == flags);
+  }
+
+  return *this;
+}
+
+// update every light's color, intensity, range etc.
+PBRDrawable& PBRDrawable::updateShaderLightParameters() {
+  constexpr float dummyRange = 10000.0;
```
mosra, line 118: [Mn::Constants::inf()](https://doc.magnum.graphics/magnum/structMagnum_1_1Math_1_1Constants.html#a33437fd32195685dc87c2b8b44176c62) is what you want to use here.

I unfortunately overlooked that when it was originally commited by @eundersander, [can you fix that there as well](https://github.com/facebookresearch/habitat-sim/blob/364b2b793818efb588e89463487708d38ae08d8a/src/esp/gfx/GenericDrawable.cpp#L52)? Thanks :pray:

src/esp/gfx/PBRShader.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+namespace esp {
+namespace gfx {
+
+namespace {
+enum TextureBindingPointIndex : uint8_t {
+  albedo = 0,
+  roughness = 1,
+  metallic = 2,
+  normal = 3,
```
mosra, line 41: Coding style: uppercase? camel case? not sure which, but it's definitely not all-lowercase :))

src/esp/gfx/PBRShader.cpp
```diff
@@ -0,0 +1,320 @@
...
+PBRShader& PBRShader::setLightPosition(unsigned int lightIndex,
+                                       const Mn::Vector3& pos) {
+  CORRADE_ASSERT(
+      lightIndex < lightCount_,
+      "PBRShader::setLightPosition: lightIndex" << lightIndex << "is illegal.",
+      *this);
+
+  setUniform(lightDirectionsUniform_[lightIndex], Mn::Vector4{pos, 1.0});
+  return *this;
+}
```
mosra, line 264: Any chance all positions could be specified together in a single call (and colors / ranges in another), like [it's done in Phong](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1Phong.html#aa25f246ff11ce2fc5d069cd34fd1ea6a)? Because calling `setUnform()` 30 times for 10 lights instead of just three times will have a pretty large impact on API overhead, *especially* on the web.

If you do it like I do in Phong, it'll also have the nice property of all array elements having consecutive uniform indices, which means you wouldn't need to store uniform locations in a vector.

Alternatively throw all this away and go with UBOs instead -- all targets supported by Habitat have those.

src/esp/gfx/PBRShader.h
```diff
@@ -0,0 +1,348 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#ifndef ESP_GFX_PBRSHADER_H_
+#define ESP_GFX_PBRSHADER_H_
+
+#include <memory>
+#include <unordered_set>
```
mosra, line 9: These two don't seem to be needed in any way?

(Sorry, I am extra sensitive to such kind of includes, haha.)

src/esp/gfx/PBRShader.h
```diff
@@ -0,0 +1,348 @@
...
+  // And MOST importantly, the lightDirection (in camera space) depends on
+  // camera matrix so it changes all the time, but the other info does not.
+  // NOTE:
+  // In fragement shader, the "LightDirection" is a vec4.
+  // when w == 0, it means .xyz is the light direction;
+  // when w == 1, it means it is the light position, NOT the direction;
+  std::vector<int> lightDirectionsUniform_;
+};
+
+CORRADE_ENUMSET_OPERATORS(PBRShader::Flags);
```
mosra, line 343:
```suggestion
CORRADE_ENUMSET_OPERATORS(PBRShader::Flags)
```

src/esp/gfx/PBRShader.h
```diff
@@ -0,0 +1,348 @@
...
+  /** @brief Copying is not allowed */
+  PBRShader& operator=(const PBRShader&) = delete;
+
+  /** @brief Move assignment */
+  PBRShader& operator=(PBRShader&&) noexcept = default;
+
+  /**
+   * @brief Get number of lights
+   */
+  unsigned int lightCount() { return lightCount_; }
```
mosra, line 161:
```suggestion
  unsigned int lightCount() const { return lightCount_; }
```

src/esp/gfx/PBRShader.h
```diff
@@ -0,0 +1,348 @@
...
+
+  enum : Magnum::UnsignedInt {
+    /**
+     * Color shader output. @ref shaders-generic "Generic output",
+     * present always. Expects three- or four-component floating-point
+     * or normalized buffer attachment.
+     */
+    ColorOutput = 0,
+
+#ifndef MAGNUM_TARGET_GLES2
```
mosra, line 64: We don't target GLES2 at all, so this ifdef isn't needed.

src/esp/gfx/PBRShader.h
```diff
@@ -0,0 +1,348 @@
...
+    /**
+     * Object ID shader output. @ref shaders-generic "Generic output",
+     * present only if @ref Flag::ObjectId is set. Expects a
+     * single-component unsigned integral attachment. Writes the value
+     * set in @ref setObjectId() there.
+     * @requires_gl30 Extension @gl_extension{EXT,texture_integer}
+     * @requires_gles30 Object ID output requires integer support in
+     *      shaders, which is not available in OpenGL ES 2.0 or WebGL
+     *      1.0.
```
mosra, line 73: These `@requires_gl` are magnum-specific and will break doc build here, remove them.

src/esp/gfx/PBRShader.h
```diff
@@ -0,0 +1,348 @@
...
+namespace esp {
+
+namespace gfx {
+
+class PBRShader : public Magnum::GL::AbstractShaderProgram {
+ public:
+  /**
+   * @brief vertex positions
+   */
+  typedef Magnum::GL::Attribute<0, Magnum::Vector3> Position;
```
mosra, line 28: I'd instead use this, so you are always sure that the same attribute is used here and in magnum's builtin shaders. Because if you get out of sync, things like `MeshTools::compile()` won't work anymore. [Details here.](https://doc.magnum.graphics/magnum/structMagnum_1_1Shaders_1_1Generic.html#Shaders-Generic-custom)

```suggestion
  typedef Magnum::Shaders::Generic3D::Position Position;
```

Same for other attribs below.

src/shaders/pbr.frag
```diff
@@ -0,0 +1,246 @@
...
+
+// -------------- shader ------------------
+#if defined(NORMAL_TEXTURE)
+vec3 getNormalFromNormalMap() {
+  vec3 tangentNormal = texture(normalTexture, texCoord).xyz * 2.0 - 1.0;
+
+#if defined(PRECOMPUTED_TANGENT)
+  mat3 TBN = mat3(tangent, biTangent, normal);
+#else
+// TODO: some magic here!
```
mosra, line 78: `#error`? because otherwise, until this is implemented, you'll get all-black models and you wouldn't know why

src/esp/gfx/PBRShader.h
```diff
@@ -0,0 +1,348 @@
...
+#define ESP_GFX_PBRSHADER_H_
+
+#include <memory>
+#include <unordered_set>
+#include <vector>
+
+#include <Corrade/Containers/EnumSet.h>
+#include <Magnum/GL/AbstractShaderProgram.h>
+#include <Magnum/Math/Matrix4.h>
+#include <Magnum/Trade/LightData.h>
```
mosra, line 15: This doesn't seem to be used either.

src/shaders/pbr.frag
```diff
@@ -0,0 +1,246 @@
...
+
+  return (diffuse + specular) * lightRadiance * max(dot(normal, light), 0.0);
+}
+
+void main() {
+#if defined(ALBEDO_TEXTURE)
+  vec4 albedoColor = texture(albedoTexture, texCoord);
+#else
+  vec4 albedoColor = Material.baseColor;
+#endif
```
mosra, line 186: As I said above, to follow glTF spec, this needs to be multiplied together. Same for the others.

src/shaders/pbr.frag
```diff
@@ -0,0 +1,246 @@
...
+  float roughness = texture(roughnessTexture, texCoord).r;
+#else
+  float roughness = Material.roughness;
+#endif
+#if defined(NORMAL_TEXTURE)
+  // normal is now in the camera space
+  vec3 n = getNormalFromNormalMap();
+#else
+  vec3 n = normal;
+#endif
```
mosra, line 202: Here you might want to include normal map scale as well -- glTF has that, the Phong shader [recently received support for it](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1Phong.html#ac20d44948e41e8371daebe198a7b9b99) and I think @eundersander started using it for some models (instead of scaling the texture data). [Details how is it applied are here.](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Trade.html#ac20668da810bce38bddd4f4278f93662ab59103e987d5c0e6e17c6c184edf90b2)

src/esp/gfx/CMakeLists.txt
```diff
@@ -26,6 +28,8 @@ set(
   RenderTarget.h
   ShaderManager.cpp
   ShaderManager.h
+  PBRShader.cpp
+  PBRShader.h
```
mosra, line 32: Just FYI, Magnum's coding style uses `Pbr` instead of `PBR` (e.g. [Trade::PbrMetallicRoughnessMaterialData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PbrMetallicRoughnessMaterialData.html)), all acronyms are capitalized this way. I don't _demand_ you to change this, just nothing that so you are aware of the difference.

mosra (commented):
I think that's enough comments for now, heh.

Wow, you're fast -- do you have any screenshots already? I'm curious how this looks with just plain direct lighting and no IBL.

src/esp/gfx/PBRShader.h
bigbike: By the way, what kind of out-of-class operators would I have after I defined this?

bigbike:
@mosra :

FYI: I still need to address 3 things:
1) texture matrix;
2) light position uniforms;
3) normal map scale;

src/esp/gfx/MaterialData.h
```diff
@@ -15,21 +15,42 @@
 namespace esp {
 namespace gfx {
 
-struct MaterialData {};
+struct MaterialData {
+  bool perVertexObjectId = false;
+
+  // construct it using the default constructor. NO initial values, such as
+  // identity matrix
+  Magnum::Matrix3 textureMatrix;
```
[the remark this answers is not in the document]
mosra: Right, I see. Identity is a default because it makes more sense than a fully zeroed-out matrix -- if you use a default-constructed matrix in transformations, it's the same as if you didn't use it at all, whereas a zero matrix would wipe them out, which is almost never useful.

src/esp/gfx/PBRDrawable.cpp
mosra: MeshOptimizer quantizes texture coordinates to save memory. So for example, when a mesh uses a [0.5, 0.75] range of the texture, it'll map that range to 0-65535 and store that in a 16-byte integer instead of a 32-bit float, saving half of memory for that attribute. Then, in order to unpack that quantized coordinate back to [0.5, 0.75], it'll supply a texture matrix that scales and offsets the coordinates again. (Same is done for positions, but there this extra dequantizing transform is added to the node transform, so no extra feature is needed.)

A good model to test with is https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Lantern/glTF-Quantized -- unless this got broken since #550, the current viewer should load and display it correctly. With texture transform missing the textures would be significantly off.

This is not a *mandatory* feature of meshopt, but it's on by default, so I'm just saying -- better have a temporary assertion in place so you know what's to blame if such a model gets loaded ;)

src/esp/gfx/PBRShader.cpp
mosra: > means that you will have to compute the "location" precisely and put such "flexible length" arrays always in the very end

I do that so all non-array uniforms can have explicit location, yes, but since you can't use that here, it doesn't matter. It's enough to query the location of the first element and then assume all other elements will occupy the locations after, that always holds.

src/esp/gfx/PBRShader.h
mosra: The [bitwise operators from here](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html#pub-methods).

Hmm, now you got me thinking, I think there's even a way to not need this macros at all. (This is originally modelled after Qt, which also had such a macro, so I didn't question this at all... but now I think in-class-defined friends could work as well?)

bigbike:
@mosra : I believe I have addressed all of your previous comments. 
Take another round if you would like to.

Thanks!

src/esp/gfx/MaterialData.h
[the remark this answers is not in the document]
bigbike: I will leave it as this.

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -0,0 +1,162 @@
...
+  return *this;
+}
+
+// update every light's color, intensity, range etc.
+PbrDrawable& PbrDrawable::updateShaderLightParameters() {
+  constexpr float dummyRange = Mn::Constants::inf();
+  std::vector<float> ranges;
+  std::vector<Mn::Color3> colors;
+  for (unsigned int iLight = 0; iLight < lightSetup_->size(); ++iLight) {
+    ranges.push_back(dummyRange);
```
mosra, line 132: FYI, this is a default, so you don't need to supply those... yet. Once those get imported from the glTF then yes, but not right now.

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  return *this;
+}
+
+// update every light's color, intensity, range etc.
+PbrDrawable& PbrDrawable::updateShaderLightParameters() {
+  constexpr float dummyRange = Mn::Constants::inf();
+  std::vector<float> ranges;
+  std::vector<Mn::Color3> colors;
```
mosra, line 130:
```suggestion
  std::vector<Mn::Color3> colors;
  colors.reserve(lightSetup_->size());
```

Like you do in the other function below.

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,484 @@
...
+      .bindMetallicTexture(metallic)
+      .bindCombinedTexture(noneRoughnessMetallic)
+      // NO Worries to call to "bind combined texture" twice,
+      // ONLY ONE texture will be bound.
+      // see generateCorrectFlags() for details
+      .bindCombinedTexture(occlusionRoughnessMetallic)
+      .bindNormalTexture(normal);
+
+  return *this;
+}
```
mosra, line 346: Hmm, I assume you have modelled this API after Phong? There it is primarily to avoid multiple `glActiveTexture()` + `glBindTexture()` calls and do all binding in just a [single multi-bind call](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1AbstractTexture.html#a9daa1b4a53fa02937aed4eaa3720dbc5) as all those binding points are consecutive. But here you're calling the single-bind APIs and thus don't really gain anything.

To be frank I don't really like that API in Phong because the texture order is unclear and error-prone, especially when a new texture type gets added. To eventually fix that there I was thinking about making the `TextureUnit` enum a public API and then having something like

```cpp
shader.bindTextures({
    {PbrShader::TextureUnit::BaseColor, baseColor},
    {PbrShader::TextureUnit::Roughness, roughness},
    {PbrShader::TextureUnit::Metallic, metallic}
    ...
});
```

That's verbose, but clear and not prone to errors. Internally the function would put those in a fixed-size array in correct order and pass that to the list-taking `bind()`.

src/esp/gfx/PbrShader.h
```diff
@@ -0,0 +1,442 @@
...
+   */
+  typedef Magnum::Shaders::Generic3D::Tangent4 Tangent4;
+
+  enum : Magnum::UnsignedInt {
+    /**
+     * Color shader output. @ref shaders-generic "Generic output",
+     * present always. Expects three- or four-component floating-point
+     * or normalized buffer attachment.
+     */
+    ColorOutput = 0,
```
mosra, line 63: While we're at it, this could be a Generic value as well:

```suggestion
    ColorOutput = Magnum::Shaders::Generic3D::ColorOutput,
```

Same for `ObjectIdOutput` below.

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,484 @@
...
+    result &= ~Flag::NoneRoughnessMetallicTexture;
+    result &= ~Flag::RoughnessTexture;
+    result &= ~Flag::MetallicTexture;
+  } else if (result & Flag::NoneRoughnessMetallicTexture) {
+    result &= ~Flag::RoughnessTexture;
+    result &= ~Flag::MetallicTexture;
+  }
+
+  return result;
+}
```
mosra, line 73: To avoid this kind of explicit postprocessing, what about having a flag like this:

```cpp
    // Explicitly saying "packed" to suggest that it already includes both
    PackedNoneRoughnessMetallicTexture = (1 << 3)|RoughnessTexture|MetallicTexture,
```

(and later for Occlusion as well, but you don't handle that one yet). Then any combination of enum values would be implicitly valid, and you'd only have to take extra care when passing the enum values through to the shader:

```cpp
// equiv. to (flags_ & Flag::PackedNoneRoughnessMetallicTexture) == Flag::PackedNoneRoughnessMetallicTexture
if(flags_ >= Flag::PackedNoneRoughnessMetallicTexture) {
  frag.addSource("#define NON_ROUGHNESS_METALLIC_TEXTURE\n");
} else {
  frag
      .addSource(flags_ & Flag::BaseColorTexture ? "#define BASECOLOR_TEXTURE\n"
                                                 : "")
      .addSource(flags_ & Flag::RoughnessTexture ? "#define ROUGHNESS_TEXTURE\n"
                                                 : "")
      .addSource(flags_ & Flag::MetallicTexture ? "#define METALLIC_TEXTURE\n"
                                                : "");
}
```

Same change for the `setUniform(uniformLocation("metallicTexture"), ...` block after.

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,484 @@
...
+                 "PbrShader::bindBaseColorTexture: the shader was not "
+                 "created with BaseColor texture enabled",
+                 *this);
+  if (lightCount_ && texture) {
+    texture->bind(TextureUnit::BaseColor);
+  }
+  return *this;
+}
+
+PbrShader& PbrShader::bindRoughnessTexture(Mn::GL::Texture2D* texture) {
```
mosra, line 239: Hmm, why don't all those take a reference?

```suggestion
PbrShader& PbrShader::bindRoughnessTexture(Mn::GL::Texture2D& texture) {
```

mosra (commented):
Found a couple more on second pass :)

src/esp/gfx/PbrShader.cpp
bigbike: I think what I have done here is a cleaner way to handle this situation as it solves the "conflicts" once for all and it is cheap. I do not need to have any nested if clause any more anywhere, either in the `PbrShader`, or `PbrDrawable`. 
Later if you would like to change the logic, relations among the textures, or add new type of textures (such as independent occlusion), again, easy, you just need to change one place -- here, in this function.
Second, I am not a big fan of the operator like `>=`, as a user to your library, I am rather confused about this operator, what does it mean? The "flags" is a bitmask. I only need to query a bit (`&`), set a bit (`|=`) or reset a bit (`&=~`). That is it. `>=` looks really strange and not straightforward to me, honestly. Each time, I will have to think for a while, and check the "Flag" definition to get the answer.

src/esp/gfx/PbrShader.cpp
bigbike: Simply because in `MaterialData` all the textures are stored in pointers, and thus can be passed into these functions immediately (I do not even need to judge if the pointer is nullptr in advance.)

src/esp/gfx/PbrShader.cpp
bigbike: Good to know. Thanks. 
Yes, it did not gain anything in terms of performance. just a very convenient function to use, and make the code looks cleaner.

I will learn from the doc you mentioned above. See what I can do.

src/esp/gfx/PbrShader.h
bigbike: OK.

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,484 @@
...
+#endif
+
+  // this is not the file name, but the group name in the config file
+  // see Shaders.conf in the shaders folder
+  const Cr::Utility::Resource rs{"default-shaders"};
+
+  Mn::GL::Shader vert{glVersion, Mn::GL::Shader::Type::Vertex};
+  Mn::GL::Shader frag{glVersion, Mn::GL::Shader::Type::Fragment};
+
+  std::stringstream attributeLocationsStream;
```
bigbike, line 94: @morsa: I do not use the `generic.glsl`. I used another way as I wrote here. I think it is better.
In your case, you will have to sync the locations between `generic.glsl` and the `Generic3D` if I understand you correctly. I do not have this issue, no matter how you refactor and change the locations in the Generic3D in the future.
Thoughts?

bigbike (commented):
@mosra : Good to see your review. Thank you so much for your valuable comments!
See my reply.

src/esp/gfx/PbrDrawable.cpp
bigbike: OK

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+  frag.addSource(attributeLocationsStream.str())
+      .addSource(isTextured() ? "#define TEXTURED\n" : "")
+      .addSource(flags_ & Flag::BaseColorTexture ? "#define BASECOLOR_TEXTURE\n"
+                                                 : "")
+      .addSource(flags_ & Flag::RoughnessTexture ? "#define ROUGHNESS_TEXTURE\n"
+                                                 : "")
+      .addSource(flags_ & Flag::MetallicTexture ? "#define METALLIC_TEXTURE\n"
+                                                : "")
+      .addSource(flags_ & Flag::NoneRoughnessMetallicTexture
```
bigbike, line 134: Since I generated the "correct" flags, this piece of code (line 128 - 140) can be as clean as it is now. You can pretend that all textures are "independent", no control logic is needed here.

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -0,0 +1,162 @@
...
+      // uploaded to GPU so simply pass 0 to the uniform "objectId" in the
+      // fragment shader
+      .setObjectId(
+          static_cast<RenderCamera&>(camera).useDrawableIds()
+              ? drawableId_
+              : (materialData_->perVertexObjectId ? 0 : node_.getSemanticId()))
+      .setTransformationMatrix(transformationMatrix)  // modelview matrix
+      .setMVPMatrix(camera.projectionMatrix() * transformationMatrix)
+      .setNormalMatrix(transformationMatrix.normalMatrix())
+      .bindTextures(materialData_->baseColorTexture,
```
bigbike, line 52: `bindTextures` cannot improve the performance, but it improves the simplicity of the code here.

src/esp/gfx/PbrShader.cpp
mosra: Apart from the perf aspect, I'm just afraid it becomes rather messy once you have also occlusion and emissive textures (and then possibly other packing variants such as two-component normals), at that point it'd be very hard to make sure that the caller got the order right.

src/esp/gfx/PbrShader.cpp
mosra: Depends :) I'd say having the API designed in a way that makes disallowed combinations impossible (and then having that all done and checked in a single central place) is better than having to explicitly scatter "fixup" calls everywhere *including* the call sites (like the `PbrDrawable`). Actually, seeing that fixup call in the drawable prompted me to write that comment -- because *what if* a similar check will need to be done in some other place, how many of such places will get forgotten?

About the `>=`, of course. That's precisely why I put the explanatory comment there -- I don't *force* you to use that, it's just a shorthand for the rather verbose check. It's the ⊇ operation from the set theory, and if you squint a bit, it's not *that* different from ≥. (Another equivalent, which is shorter, but adds even more congitive overhead is `!(a & ~b)`.)

src/esp/gfx/PbrShader.cpp
mosra: Commented because in Magnum all pointers that *can't* be null should be references and I think that's a policy we adopted here as well -- but not 100% sure. For consistency with other shaders it should be a reference tho ;)

On the contrary, in the `bindTextures()` call (or at least in the variant of it implemented in Phong), these *can* be `nullptr` -- and if it's so, given slot will not be bound. That's useful because in most cases you always have just a subset of the textures, almost never all of them **(In your case it's however not true, and if you pass a `nullptr` there, it'll crash.)**

src/esp/gfx/PbrShader.cpp
mosra: Yep, that's what [the documentation suggests to do](https://doc.magnum.graphics/magnum/structMagnum_1_1Shaders_1_1Generic.html#Shaders-Generic-custom) as well. The `generic.glsl` is an implementation detail in Magnum, not meant to be used outside.

> you will have to sync the locations between

Yep, but I [have that covered by tests](https://github.com/mosra/magnum/blob/ba156d329a4347f2c58a2d63d71a2db645666ce9/src/Magnum/Shaders/Test/GenericTest.cpp#L54-L83). I chose this approach for the library because that means I don't need to do heavy string processing on every shader instantiation. It matters for certain use cases.

src/esp/gfx/PbrShader.cpp
bigbike: Lets say:

        enum class Flag: unsigned int {
            A = 1 << 0,
            B = 1 << 1,
            C = 1 << 2 | A,
            D = 1 << 3,
        };
`Flag foo = Flag|D`; So `foo` is 1000 which means `foo >= Flag|C` because Flag|C is 0101, but it does not necessarily include feature A and feature C. Correct? That is how I understand the operator `>=`.


 >  // Explicitly saying "packed" to suggest that it already includes both
 >   PackedNoneRoughnessMetallicTexture = (1 << 3)|RoughnessTexture|MetallicTexture,

You fully misunderstood what I meant here. Each bit in the bitmask is exclusive. This is different from your Magnum library. "RoughnessTexture" does NOT mean it has Roughness in the texture, but I do not care if it is in its own texture or a packed texture. It means roughness is stored in an independent, unique texture. And once my shader knows this fact, it will extract the roughness from R channel, not from G channel.

> because what if a similar check will need to be done in some other place, how many of such places will get forgotten?

Your method suffers the same problem. The same logic applies as well. Because you need the branching logic everywhere then, namely, you will have to repeat the nested "if clause" in your previous comments, for example, here, and in bindRoughnessTexture (because you will have to check if NONE_ROUGHNESS_METALLIC_TEXTURE or OCCLUSION_ROUGHNESS_METALLIC_TEXTURE are presented or not), bindMetallicTexture, and more places.

What I can improve is that, since shader key is, eventually, used in the shaderManager (besides `PbrShader` class). So I guess I will call such "fixup" in the getShaderKey function (and the PbrShader constructor of course.) 

> checked in a single central place

Depends on how you view this problem. on the contrary, I believe that is exactly what I did here. I checked and fixed the logic in a single central place -- generateCorrectFlags. And this a core idea I tried to convey in my first reply. After calling this function, it is worry-free, "if-clause" free.

> !(a & ~b)
This is a big cognitive overhead, again, not straightforward, not even close. So you are saying `a >= b`, it means b belongs to a. Is it correct? I did not see why it helps your claim (making the "disallowed combinations impossible") because you need the "if clause" to assist you everywhere. However, my "fixup" function did what exactly you claimed -- making the "disallowed combinations impossible".

Also, the burden for the developer is a bit heavier, she must understand that for such "combination bit", she should always remember to use ">=" instead of "&".

The fundamental difference on this problem between us is whether any bitmask should carry more than 1 meaning. I do not think so but you do.

src/esp/gfx/PbrShader.cpp
bigbike: I see. Thanks.

src/esp/gfx/PbrShader.cpp
bigbike: Why would it crash? 
If any nullptr is passed, it will be captured within the actual binding functions, such as bindRoughTextures etc.

src/esp/gfx/PbrShader.cpp
bigbike: > at that point it'd be very hard to make sure that the caller got the order right.

That is true. It is the developer's ability to make sure the order is right.

Adding comments may help in general, but not needed here, as the variable names are already self-explanatory. 
```
foo(texture1,  // A
    texture2);  // B
```

src/esp/gfx/PbrDrawable.cpp
mosra: Wait. This was a brain fart on my side, sorry (probably wrote this before having my :coffee: or something :exploding_head:).

It *would* be a default if you'd have that as a default in the shader source / constructor (which I'm not sure if you do). Somehow I mixed that up with the defaults I had in Phong and *assumed* you do the exact same thing :sweat_smile:

src/esp/gfx/PbrShader.cpp
mosra: Yes, I mean the case where someone would change that to

```cpp
foo(textureB, // B
    textureA); // A
```

because of refactoring or some other sweeping change. It still *looks* correct, but you have no way to know until you see the actual signature.

src/esp/gfx/PbrShader.cpp
mosra: Okay, fair enough :) I'm not 100% sure yet how I will handle packed textures in builtin shaders, so this was just one of the ideas I had. Another is supplying an explicit swizzle for each texture, but that has its own bag of problems.

> That is how I understand the operator >=

In case of bit sets I understand it as "is a superset of" (that's why I said ⊇), while you understand it as "is numerically larger than" like with integers.

src/esp/gfx/PbrShader.cpp
mosra: Ah, I overlooked the `if(texture)` part in there. Then it won't crash, you're right.

src/esp/gfx/PbrShader.cpp
bigbike: > "is numerically larger than" like with integers.

I believe most developers will take it for granted. Especially, the type of the Flag element is set as `UnsignedInt`.  It is a strong hint. They will assume the "numerically larger than", because it is the most common way to understand it. So with all due respect, I am not convinced `>=` is a good design in the Magnum since it goes against the common sense, in my shallow opinion.

src/esp/gfx/PbrShader.cpp
bigbike: Yes, you are right. What is you suggestion here?
How should I change the signature at this moment?

mosra: I was thinking this in the comment above:

> ```c++
> shader.bindTextures({
>     {PbrShader::TextureUnit::BaseColor, baseColor},
>     {PbrShader::TextureUnit::Roughness, roughness},
>     {PbrShader::TextureUnit::Metallic, metallic}
>     ...
> });
> ```

Or just ... for the first iteration keep only the single-texture functions, we can always add optimizations later ;)

src/esp/gfx/PbrDrawable.cpp
bigbike: I have not changed it yet. Will work on this part later.

src/esp/gfx/PbrShader.cpp
bigbike: I have not worked on this yet. Will do later.

bigbike:
@mosra : I believe the PBR stuff is connected to the system now, even though I have not debugged it yet.
The code is there now.
Take a look, if you are interested.

Will do the testing and debugging starting from tomorrow.

src/esp/gfx/MaterialData.h
```diff
@@ -35,19 +50,34 @@ struct PhongMaterialData : public MaterialData {
...
+  // channel)
   Magnum::GL::Texture2D* metallicTexture = nullptr;
+  // an independent, NOT packed texture to store "roughness" (in R channel)
   Magnum::GL::Texture2D* roughnessTexture = nullptr;
+  // a packed texture to store "roughness" (in G channel) and "metallic" (in B
+  // channel)
   Magnum::GL::Texture2D* noneRoughnessMetallicTexture = nullptr;
+  // a packed texture to store "occlusion"(in R channel), "roughness" (in G
+  // channel) and "metallic" (in B channel)
   Magnum::GL::Texture2D* occlusionRoughnessMetallicTexture = nullptr;
```
mosra, line 77: This packing is a superset of the above, so you'd need maybe just one of the two?

src/esp/assets/ResourceManager.h
```diff
@@ -112,8 +113,25 @@ class ResourceManager {
...
+  /**
+   * @brief Flag
+   *
+   * @see @ref Flags, @ref flags()
+   */
+  enum class Flag : Magnum::UnsignedShort {
+    /**
+     * build phong material from PBR material
+     */
+    buildPhongFromPbr = 1 << 0,
```
mosra, line 125:
```suggestion
    BuildPhongFromPbr = 1 << 0,
```

(also, doesn't Habitat use ALL_UPPERCASE for enums? same would be for the PbrShader flags)

src/esp/assets/ResourceManager.cpp
```diff
@@ -1267,6 +1272,61 @@ gfx::PhongMaterialData::uptr ResourceManager::buildPhongShadedMaterialData(
...
+    const Mn::Trade::PbrMetallicRoughnessMaterialData& material,
+    int textureBaseIndex) {
+  // NOLINTNEXTLINE(google-build-using-namespace)
+  using namespace Mn::Math::Literals;
+
+  auto finalMaterial = gfx::PbrMaterialData::create_unique();
+
+  finalMaterial->baseColor = material.baseColor();
+  finalMaterial->metallic = material.metalness();
+  finalMaterial->roughness = material.roughness();
```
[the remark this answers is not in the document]
bigbike: Yes, good point.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1589,23 +1649,33 @@ void ResourceManager::addPrimitiveToDrawables(int primitiveID,
...
+    case gfx::MaterialDataType::PHONG:
+      node.addFeature<gfx::GenericDrawable>(mesh, shaderManager_, lightSetup,
+                                            material, group);
+    case gfx::MaterialDataType::PBR:
+      node.addFeature<gfx::PbrDrawable>(mesh, shaderManager_, lightSetup,
                                         material, group);
+    default:
+      CORRADE_INTERNAL_ASSERT_UNREACHABLE();
+      break;
+  }
```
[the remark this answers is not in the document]
bigbike: >  missing break

Yes, silly mistakes.

>  will note you about unhandled cases:

I just would like to get rid of such unnecessary warning so I used "default".

src/esp/assets/ResourceManager.h
bigbike: OK.

src/esp/assets/ResourceManager.h
```diff
@@ -112,8 +113,25 @@ class ResourceManager {
...
+  };
+
+  /**
+   * @brief Flags
+   */
+  typedef Corrade::Containers::EnumSet<Flag> Flags;
```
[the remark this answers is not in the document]
bigbike: OK.

src/esp/assets/ResourceManager.h
bigbike: > doesn't Habitat use ALL_UPPERCASE for enums?
For example, in the `RenderCamera` class.

src/shaders/pbr.frag
```diff
@@ -0,0 +1,274 @@
...
+
+// Specular D, normal distribution function (NDF),
+// also known as ggxDistribution
+// normal: normal direction
+// light: light source direction
+// viwer: camera direction, aka light outgoing direction
+// half: half vector of light and view
+float normalDistribution(vec3 normal, vec3 half, float roughness) {
+  float a = roughness * roughness;
+  float a2 = a * a;
```
Skylion007, line 119: Shouldn't all you squaring be using the 'exp2' function? http://csciwww.etsu.edu/barrettm/4157/OpenGL%20Functions.htm

src/esp/gfx/MaterialData.h
bigbike: I do not think so.
In terms of design, I have different philosophy from the magnum library at this point.
- First, this is just a pointer, so the storage cost is tiny.
- It gives me a lot of convenience, avoiding any "if" control logic. For example, in Magnum, if roughnessTexture contains a texture. Later when you use it, you do not know if it is an independent texture, or a packed texture. You will have to call `hasOcclusionRoughnessMetallicTexture()` to tell which case it is. (In that function, there are "if clauses" for sure.) In my case, I do not need such things. If `occlusionRoughnessMetallicTexture` is not empty, it means it has a packed texture, and based on the priority among the textures, I do not even need to care what is in the `roughnessTexture`. In this manner, the code is simplified and cleaner.

bigbike:
@mosra: 
I think we have it.
First version includes baseColor, roughness, metallic, normal map.

bigbike:
Can support emission.
Also, "screen space tangent" causes the artifacts in some cases (e.g., in previous images, the edge of the cup sleeve (zarf)).
Use precomputed tangent instead.

bigbike:
> A good model to test with is https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Lantern/glTF-Quantized -- unless this got broken since #550, the current viewer should load and display it correctly. With texture transform missing the textures would be significantly off.

Tested it with this model, looks good to me.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1288,6 +1292,103 @@ gfx::PhongMaterialData::uptr ResourceManager::buildPhongShadedMaterialData(
...
+  // roughnessTexture, metallicTexture)
+  // It is OK! No worries! Such duplication (or "conflicts") will be handled in
+  // the PbrShader (see PbrMaterialData for more details)
+
+  // roughness
+  if (material.hasAttribute(Mn::Trade::MaterialAttribute::Roughness)) {
+    finalMaterial->roughness = material.roughness();
+  }
+  if (material.hasRoughnessTexture()) {
+    finalMaterial->occlusionRoughnessMetallicTexture =
```
bigbike, line 1358: It is a bug here.
should be:
```
finalMaterial->roughnessTexture = 
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -1646,34 +1747,48 @@ void ResourceManager::addPrimitiveToDrawables(int primitiveID,
...
+      node.addFeature<gfx::PbrDrawable>(
+          mesh,
+          meshAttributeFlags,  // mesh attribute flags
+          shaderManager_, lightSetupKey, materialKey, group);
+      break;
+    default:
+      CORRADE_INTERNAL_ASSERT_UNREACHABLE();
+      break;
+  }
+}  // namespace assets
```
bigbike, line 1791: this is not the end of the namespace

bigbike:
All the contents in this PR have been merged into master. (#837 #840 #863)

Issue #804, opened 2020-09-16
Pre-commit hook to format our test asset JSONS


Skylion007:
## 🚀 Feature

Add another pre-commit hook to format our JSONs in data/test_assets

## Motivation
We should standardize the style of our test asset jsons with a pre-commit hook like we do everything else in the repo.

Skylion007:
We now have a pre-commit hook that at least validates our JSONs, but I guess we would need a JSON "style guide"

PR #813, opened 2020-09-28
Add typehints and mypy static typing checks to habitat_sim

Skylion007:
## Motivation and Context

Finally went ahead and added typehints everywhere using [PyTest and monkeytype](https://pypi.org/project/pytest-monkeytype/). This will add typehints to the last parts of the code base that did not have it before. I used flake8 to ensure that we didn't double add any imports and monkeytype automatically parses the typed Python to ensure its valid.

This should bring our typing coverage pretty close to 100% so we can start using the static typehint checkers soon like mypy.

This is a nice automatic way to bootstrap our typehints with PyTest by using the tests we already have to add typehints where appropriate.

I also asked it to respect existing TypeHints where appropriate so it didn't overwrite and typehints we already had. I also was careful to try to make sure we didn't accidentally double register everything.

The only remaining annoyance was cyclic dependencies introduced by typing. Since we are on Python 3.6, I couldn't use the `from __futures__ import annotations` which solves the issue using the nextgen typehinting, so I used the old Python 3.6 workaround of using the String type annotations which evaluate lazily.

This is a pretty big change, so let's hope it didn't break the code in anyway.

Better typehinting for better documentation.
It solves the 

## How Has This Been Tested
With flake8, monkeytype and pytest.

## Types of changes

## Checklist


habitat_sim/simulator.py
```diff
@@ -119,45 +122,45 @@ def close(self):
 
         super().close()
 
-    def __enter__(self):
+    def __enter__(self) -> "Simulator":
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(self, exc_type: None, exc_val: None, exc_tb: None) -> None:
```
Skylion007, line 128: Hmm, this might be a mistake and should be Any, I think. Might want to consult the Python docs to see what the proper arg typing for this is.

habitat_sim/utils/data/pose_extractor.py
```diff
@@ -49,22 +58,34 @@ def extract_poses(self, view: np.ndarray, fp: str) -> List[tuple]:
...
-    def _compute_quat(self, cam_normal):
+    def _compute_quat(self, cam_normal: ndarray) -> quaternion:
         """Rotations start from -z axis"""
         return quat_from_two_vectors(habitat_sim.geo.FRONT, cam_normal)
 
-    def _convert_to_scene_coordinate_system(self, poses, ref_point):
+    def _convert_to_scene_coordinate_system(
+        self,
+        poses: Union[
+            List[Tuple[Tuple[int, int], Tuple[int, int], float64, str]],
```
Skylion007, line 81: Hmm, seems like this might be able to made slightly less verbose.

habitat_sim/bindings/__init__.py
```diff
@@ -7,31 +7,21 @@
...
-    "RigidState",
-]
-
-from habitat_sim._ext.habitat_sim_bindings import (  # noqa: F401 isort:skip
-    Simulator as SimulatorBackend,
-)
-
-exec(
-    "from habitat_sim._ext.habitat_sim_bindings import ({})".format(", ".join(modules))
+from habitat_sim._ext.habitat_sim_bindings import (
```
eundersander, line 10: What's going on here? Why did we use the `exec` before? Why don't we want to define `__all__` any more?

habitat_sim/sensors/noise_models/redwood_depth_noise_model.py
```diff
@@ -5,21 +5,27 @@
...
 
 from os import path as osp
+from typing import Union
 
 import attr
 import numba
 import numpy as np
+from numpy import ndarray
 
+try:
```
eundersander, line 15: Before, we'd only import Torch lazily, if/when we actually use `RedwoodDepthNoiseModel.simulate`. Now we'll always import it. Any concern here? Is there some chance some env worker that previously didn't import Torch is now importing it? Doesn't torch consume quite a lot of GPU memory on import just due to kernels, etc.?

habitat_sim/simulator.py
```diff
@@ -6,14 +6,22 @@
...
+from typing import Any, Dict, List, Optional, Union
 
 import attr
 import magnum as mn
 import numpy as np
+from magnum import Vector3
+from numpy import ndarray
+
+try:
+    import torch
```
eundersander, line 18: Same concern from the earlier file about moving away from lazy Torch import.

habitat_sim/utils/profiling_utils.py
```diff
@@ -66,13 +66,12 @@ class RangeContext(ContextDecorator):
     statement. See also range_push.
     """
 
-    def __init__(self, msg: str):
+    def __init__(self, msg: str) -> None:
         self._msg = msg
 
-    def __enter__(self):
+    def __enter__(self) -> "RangeContext":
```
eundersander, line 72: Is this an example of "old Python 3.6 workaround of using the String type annotations which evaluate lazily"?

habitat_sim/bindings/__init__.py
Skylion007: I read the documentation and __all__ is implicitly set to be all imported things without an underscore, so we actually don't need it. It wasn't useful because it wasn't parseable by any IDEs anyhow.

habitat_sim/sensors/noise_models/redwood_depth_noise_model.py
Skylion007: It shouldn't have any side effects until its placed on the GPU so it should be fine. You have to start a torch session before it starts consuming memory, right @erikwijmans ?

habitat_sim/utils/profiling_utils.py
Skylion007: Yep. In Python 3.7>= you just use `from __future__ import annotations` and it works automatically.

habitat_sim/sensors/noise_models/redwood_depth_noise_model.py
eundersander: >You have to start a torch session before it starts consuming memory
my quick experiment shows this to be correct.

eundersander (approved):
LGTM. Thanks for doing this!

Skylion007:
This now fixes #804

PR #814, opened 2020-09-29
Fix for Bullet static objects/stage and add get_num_active_contact_points, 

eundersander:
## Motivation and Context

This is a change to how we create Bullet static objects (including the stage). The previous approach caused an issue where the static object would never go to sleep, thus causing a potential runtime perf hit. The new approach is more consistent with Bullet example code and fixes the non-sleeping issue.

I've also added `get_num_active_contact_points`, which is generally useful as a proxy for Bullet collision cost/complexity and also helps diagnose/test the specific issue with static objects.

## How Has This Been Tested

`SimTest` unit test. Ad hoc testing in viewer with stage, dynamic objects, and static objects.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletBase.h
+++ b/src/esp/physics/bullet/BulletBase.h
@@ -98,10 +98,9 @@ class BulletBase {
   std::shared_ptr<btMultiBodyDynamicsWorld> bWorld_;
 
   /** @brief Static data: All components of a @ref RigidObjectType::SCENE are
-   * stored here. See @ref btCollisionObject.  Also, all objects set to STATIC
-   * are stored here.
+   * stored here. Also, all objects set to STATIC are stored here.
    */
-  std::vector<std::unique_ptr<btCollisionObject>> bStaticCollisionObjects_;
+  std::vector<std::unique_ptr<btRigidBody>> bStaticCollisionObjects_;
 
   //! keep a map of collision objects to object ids for quick lookups from
   //! Bullet collision checking.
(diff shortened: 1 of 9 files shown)


src/esp/physics/bullet/BulletBase.h
```diff
@@ -98,10 +98,9 @@ class BulletBase {
   std::shared_ptr<btMultiBodyDynamicsWorld> bWorld_;
 
   /** @brief Static data: All components of a @ref RigidObjectType::SCENE are
-   * stored here. See @ref btCollisionObject.  Also, all objects set to STATIC
-   * are stored here.
+   * stored here. Also, all objects set to STATIC are stored here.
    */
-  std::vector<std::unique_ptr<btCollisionObject>> bStaticCollisionObjects_;
+  std::vector<std::unique_ptr<btRigidBody>> bStaticCollisionObjects_;
```
eundersander, line 103: Should we change this variable name?

aclegg3:
Changing STATIC objects to also use rigidBody makes sense. 

Perhaps this should be refactored into `BulletRigidObject::constructRigidBody()`. Pass a MotionType to this function and handle the differences internally. Should simplify `setMotionType()` quite a bit.

eundersander:
>Perhaps this should be refactored into BulletRigidObject::constructRigidBody()

I have no strong opinion. Can we do that refactor as a separate PR? Can we merge this as-is?

aclegg3 (approved):

PR #823, opened 2020-10-03
fix for BulletRigidObject convex shapes

eundersander:
## Motivation and Context

We looked at some physics videos recently that showed books falling through the bookcase and objects not resting properly on the sofa. As a result, we were concerned about the long-term usability of convex shapes.

I tracked this down to a bug in how we're constructing the convex shapes. Specifically, while GLB nodes/meshes often contain *scale*, the Bullet `btTransform` class does *not* support scale/shear. And sadly, Bullet doesn't assert or warn about this when you construct btTransforms.

My fix here is to give Bullet world-space vertices (already transformed) instead of local-space.

## How Has This Been Tested

Ad hoc testing in viewer with our sofa object (composed of multiple convex parts). See screenshot.
sofa_test

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletRigidObject.cpp
+++ b/src/esp/physics/bullet/BulletRigidObject.cpp
@@ -239,13 +239,17 @@ void BulletRigidObject::constructBulletCompoundFromMeshes(
             btVector3(transformFromLocalToWorld.transformPoint(v)), false);
       }
     } else {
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

aclegg3 (approved):
Results look great! Thanks!

PR #835, opened 2020-10-07
Enable typeguard pytest plugin

Skylion007:
## Motivation and Context

We have static type guarding, now we can use pytest to do runtime typeguarding too (only when pytest is run). 
It also caught a few minor typing bugs (with lists/ndarrays) used interchangably that have now been fixed.

This will only check if any existing typehints are used incorrectly. (Ex passing a list when it expects a numpy array etc).

## How Has This Been Tested
Typeguard and pytest

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/agent/controls/object_controls.py
+++ b/habitat_sim/agent/controls/object_controls.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Callable
+from typing import Callable
 
 import attr
 import numpy as np
@@ -30,7 +30,9 @@ class ObjectControls(object):
         handle collisions
     """
 
-    move_filter_fn: Callable[..., Any] = attr.ib(default=_noop_filter)
+    move_filter_fn: Callable[[np.ndarray, np.ndarray], np.ndarray] = attr.ib(
+        default=_noop_filter
+    )
 
     @staticmethod
     def is_body_action(action_name: str):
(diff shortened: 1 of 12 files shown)


habitat_sim/sensors/noise_models/no_noise_model.py
```diff
@@ -21,7 +22,8 @@
 
 
 @registry.register_noise_model(name="None")
-class NoSensorNoiseModel(SensorNoiseModel):
+@attr.s(auto_attribs=True)
+class NoSensorNoiseModel(SensorNoiseModel):  # type: ignore
```
Skylion007, line 26: It kept complaining about the fact that the constructor required a GPU_ID due to the parent class, but I couldn't override it. 'attr.s' is deprecated anyhow, we really should be just using Python dataclasses instead.

habitat_sim/agent/controls/object_controls.py
```diff
@@ -16,9 +17,10 @@
 
 # epislon used to deal with machine precision
 EPS = 1e-5
+_3d_point = Union[np.ndarray, mn.Vector3, Tuple[float, float, float]]
 
 
-def _noop_filter(start: np.array, end: np.array) -> np.array:
+def _noop_filter(start: _3d_point, end: _3d_point) -> _3d_point:
```
eundersander, line 23: It seems like this noop filter was written to work with any type and is very general-purpose. Is there a good reason to narrow the types so specifically?

eundersander: Er, I guess you're making the type broader, not narrower here. I guess what I'm saying is, both the old typing (nd.array) and the new typing (_3d_point) seem unnecessarily narrow.

eundersander (approved):
I posted a philosophical question about Python typing, but regardless, LGTM.

PR #837, opened 2020-10-07
[bug fix][PBR 1] Ignore normal maps with a warning for meshes without tangents

bigbike:
## Motivation and Context

First, this diff fixes a bug. Now it will ignore the normal maps with a warning for meshes without tangents. This is quite similar to:
https://github.com/mosra/magnum-extras/commit/ad06e1ab741cb62eff43bf1b3de3be679d4a1610

Second, it improves the code readability  by changing 
```
lightSetup --> lightSetupKey
material --> materialKey
```
in a number of places.

Third, this is the 1st diff to implement the PBR. (I am going to split the giant PR #812 into many small PRs to make the reviewers' life a bit easier.)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1599,13 +1599,15 @@ void ResourceManager::addComponent(
 
     gfx::Drawable::Flags meshAttributeFlags{};
     const auto& meshData = meshes_[meshID]->getMeshData();
-    if (meshData != Cr::Containers::NullOpt &&
-        meshData->hasAttribute(Mn::Trade::MeshAttribute::Tangent)) {
-      meshAttributeFlags |= gfx::Drawable::Flag::HasTangent;
-    }
-    if (meshData != Cr::Containers::NullOpt &&
-        meshData->hasAttribute(Mn::Trade::MeshAttribute::Bitangent)) {
-      meshAttributeFlags |= gfx::Drawable::Flag::HasSeparateBitangent;
+    if (meshData != Cr::Containers::NullOpt) {
+      if (meshData->hasAttribute(Mn::Trade::MeshAttribute::Tangent)) {
+        meshAttributeFlags |= gfx::Drawable::Flag::HasTangent;
+
+        // if it has tangent, then check if it has bitangent
+        if (meshData->hasAttribute(Mn::Trade::MeshAttribute::Bitangent)) {
+          meshAttributeFlags |= gfx::Drawable::Flag::HasSeparateBitangent;
+        }
+      }
     }
     createGenericDrawable(mesh,                // render mesh
                           meshAttributeFlags,  // mesh attribute flags
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -426,7 +426,8 @@ bool ResourceManager::loadStageInternal(
     DrawableGroup* drawables /* = nullptr */,
     bool computeAbsoluteAABBs /*  = false */,
     bool splitSemanticMesh /* = true */,
-    const Mn::ResourceKey& lightSetup /* = Mn::ResourceKey{NO_LIGHT_KEY})*/) {
+    const Mn::ResourceKey&
+        lightSetupKey /* = Mn::ResourceKey{NO_LIGHT_KEY})*/) {
   // scene mesh loading
   const std::string& filename = info.filepath;
   bool meshSuccess = true;
@@ -446,11 +447,11 @@ bool ResourceManager::loadStageInternal(
         meshSuccess = loadSUNCGHouseFile(info, parent, drawables);
       } else if (info.type == AssetType::MP3D_MESH) {
         meshSuccess = loadGeneralMeshData(info, parent, drawables,
-                                          computeAbsoluteAABBs, lightSetup);
+                                          computeAbsoluteAABBs, lightSetupKey);
       } else {
         // Unknown type, just load general mesh data
         meshSuccess = loadGeneralMeshData(info, parent, drawables,
-                                          computeAbsoluteAABBs, lightSetup);
+                                          computeAbsoluteAABBs, lightSetupKey);
       }
     }
   } else {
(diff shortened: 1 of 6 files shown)


src/esp/gfx/Drawable.h
```diff
@@ -24,6 +26,24 @@ class DrawableGroup;
...
  public:
+  /** @brief Flag
+   * It will not be used directly in the base class "Drawable" but
+   * it will be used in a couple of sub-classes
+   */
+  enum class Flag {
+    /**
+     * indicates the mesh data has tangent attribute
+     */
+    HasTangent = 1 << 0,
```
bigbike, line 37: There are two conventions in the code base for enum. One is "all upper cases" style (e.g., sensor), and the other is "camel case" style (e.g., in DepthUnprojection, RenderCamera, Magnum).
I believe camel case is a bit better in terms of readability. So decided to use this style from now on.

src/esp/gfx/GenericDrawable.h
```diff
@@ -49,6 +50,8 @@ class GenericDrawable : public Drawable {
       shader_;
   Magnum::Resource<MaterialData, PhongMaterialData> materialData_;
   Magnum::Resource<LightSetup> lightSetup_;
+
+  Magnum::Shaders::Phong::Flags flags_;
```
bigbike, line 54: Before, the flag was computed again and again in each rendering loop, which I did not think it was necessary. So create a member variable to store it. It will be only computed once in the constructor.

src/esp/gfx/Drawable.h
jturner65: This decision is compatible with google c++ style guide on [enum naming](https://google.github.io/styleguide/cppguide.html#Enumerator_Names).  Do we want to use the style guide constant convention (i.e. prefix with "k") ?

jturner65 (approved):
Looks good.

src/esp/gfx/Drawable.h
mosra: Definitely no Hungarian Notation or `k` prefixes, **please** :scream: 

We use C++11 scoped enums extensively (Google didn't use them much last time I checked), so I don't think prefixing the values with `k` makes any sense -- you already see it's an enum in `SomeEnum::SomeValue`, no need to add noise to it by doing `SomeEnum::kSomeValue`.

About casing I can't say either way -- who's responsible for coding style here? :) (Actually, sorry about `DepthUnprojection`, there I forgot to use the `UPPER_CASE`.) In Magnum I'm using `CamelCase` for everything except macros, but in Magnum's Python bindings I'm using `UPPER_CASE` again to be consistent with Python itself.

src/esp/assets/ResourceManager.cpp
```diff
@@ -882,9 +883,24 @@ bool ResourceManager::loadInstanceMeshData(
...
-          PER_VERTEX_OBJECT_ID_MATERIAL_KEY, drawables);
+
+      gfx::Drawable::Flags meshAttributeFlags{};
+      if (meshes_[iMesh]->getMeshData()->hasAttribute(
+              Mn::Trade::MeshAttribute::Tangent)) {
+        meshAttributeFlags |= gfx::Drawable::Flag::HasTangent;
+      }
+      if (meshes_[iMesh]->getMeshData()->hasAttribute(
+              Mn::Trade::MeshAttribute::Bitangent)) {
+        meshAttributeFlags |= gfx::Drawable::Flag::HasBitangent;
```
mosra, line 894: I'd name it `HasSeparateBitangent` to avoid confusion because four component tangents *technically* also have a bitangent.

mosra (commented):
I appreciate that you're splitting this up into smaller chunks :+1:

src/esp/gfx/Drawable.h
bigbike: No specific reasons. The legacy code can be refactored to use the `PascalCase` (almost the same as `camelCase`) as well. 
Note: Facebook has its own C++ coding conventions, style guidelines internally. We are following it. About `enum`, we will use `PascalCase` from now on as the guidance suggests.

The guidance also links to this page http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps. 
It says do not use ALL_CAPS for enumerators to avoid clashes with macros.

src/esp/assets/ResourceManager.cpp
bigbike: Makes sense. Will do.

bigbike:
@Skylion007 @erikwijmans :

I may need your help.
It suffers fatal python errors and thus keeps failing on the CI tests. Not sure what happened.
Any chance you can take a look with me? Thanks in advance!

Skylion007:
I was able to replicate this locally with pytest test_sensors.py it seems that the simulator is failing initialization there.

Skylion007:
@bigbike Found the problem: it's with this test: "tests/test_sensors.py::test_sensors[True-True-./data/scene_datasets/mp3d/17DRP5sb8fy/17DRP5sb8fy.glb-semantic_sensor]"

This refactor breaks the MP3D semantic mesh.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1610,8 +1639,13 @@ void ResourceManager::addPrimitiveToDrawables(int primitiveID,
                                               scene::SceneNode& node,
                                               DrawableGroup* drawables) {
   CHECK(primitive_meshes_.count(primitiveID));
-  createGenericDrawable(*primitive_meshes_.at(primitiveID), node, NO_LIGHT_KEY,
-                        WHITE_MATERIAL_KEY, drawables);
+  gfx::Drawable::Flags meshAttributeFlags{};
```
Skylion007, line 1642: Wait, why do you empty the mesh attributes flags here? Shouldn't you repopulate them using the semantic mesh?

Skylion007 (changes requested):

src/esp/assets/ResourceManager.cpp
bigbike: This is inside the function `addPrimitiveToDrawables`, which has nothing to do with the semantic mesh.
I hope I did not misunderstand anything here.

src/esp/assets/ResourceManager.cpp
```diff
@@ -884,9 +884,24 @@ bool ResourceManager::loadInstanceMeshData(
 
     for (uint32_t iMesh = start; iMesh <= end; ++iMesh) {
       scene::SceneNode& node = parent->createChild();
-      node.addFeature<gfx::GenericDrawable>(
-          *meshes_[iMesh]->getMagnumGLMesh(), shaderManager_, NO_LIGHT_KEY,
-          PER_VERTEX_OBJECT_ID_MATERIAL_KEY, drawables);
+
+      gfx::Drawable::Flags meshAttributeFlags{};
+      if (meshes_[iMesh]->getMeshData()->hasAttribute(
```
erikwijmans, line 889: @bigbike `meshes_[iMesh]->getMeshData()` returns an optional and GenericInstanceMeshData doesn't currently set that field.  It probably should as there is a `// TODO: this could instead use Mn::Trade::MeshData directly` but just first checking the optional before unwrapping works here.

By the way, you can put python in tests in lldb/gdb by doing something like
```
lldb python
> r -m pytest tests/test_sensors.py -xs
```

(That is how I figured out where the abort was happening/why)

src/esp/assets/ResourceManager.cpp
```diff
@@ -1583,7 +1598,21 @@ void ResourceManager::addComponent(
           std::to_string(metaData.materialIndex.first + materialIDLocal);
     }
 
-    createGenericDrawable(mesh, node, lightSetupKey, materialKey, drawables);
+    gfx::Drawable::Flags meshAttributeFlags{};
+    if (meshes_[meshID]->getMeshData()->hasAttribute(
```
erikwijmans, line 1602: These aren't failing but should either do the unwrap conditionally or check with an assert.

bigbike:
> @bigbike Found the problem: it's with this test: "tests/test_sensors.py::test_sensors[True-True-./data/scene_datasets/mp3d/17DRP5sb8fy/17DRP5sb8fy.glb-semantic_sensor]"
> 
> This refactor breaks the MP3D semantic mesh.
> 
> Thoughts @mosra?

@Skylion007 : Thank you so much for locating the error in the instance mesh loading. Yes, it is where the problem is.

src/esp/assets/ResourceManager.cpp
bigbike: Oh, wow. I do not know this trick before. :)
Thank you so much @erikwijmans for looking into this problem and letting me know your method. Certainly I learned something rather useful from you today. I tried immediately. Yes, it gave me where it crashed.

The instanced mesh data is stored independently, and thus the member variable `meshData_` would not be set at this point. It is a trap here.

src/esp/assets/ResourceManager.cpp
bigbike: Yes, will do.

Skylion007 (approved):

PR #838, opened 2020-10-07
Use normalMatrix in the GenericDrawable

bigbike:
## Motivation and Context
- Use `normalMatrix()` in the GenericDrawable;
- Light range is set to `Mn::Constants::inf()`;
- Remove not used variables;

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #840, opened 2020-10-08
[PBR 2] PBR GL shaders (forward rendering)

bigbike:
## Motivation and Context
This is the 2nd diff in the PBR project.

This diff is to introduce the PBR OpenGL shaders for forward rendering.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/PbrShader.cpp
+++ b/src/esp/gfx/PbrShader.cpp
@@ -1,0 +1,486 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "PbrShader.h"
+
+#include <Corrade/Containers/Reference.h>
+#include <Corrade/Utility/Assert.h>
+#include <Corrade/Utility/Debug.h>
+#include <Corrade/Utility/DebugStl.h>
+#include <Corrade/Utility/FormatStl.h>
+#include <Corrade/Utility/Resource.h>
+#include <Magnum/GL/Context.h>
+#include <Magnum/GL/Extensions.h>
+#include <Magnum/GL/Shader.h>
+#include <Magnum/GL/Texture.h>
+#include <Magnum/GL/Version.h>
+#include <Magnum/ImageView.h>
+#include <Magnum/Math/Color.h>
+#include <Magnum/Math/Matrix3.h>
+#include <Magnum/Math/Matrix4.h>
+#include <Magnum/PixelFormat.h>
+
+#include "esp/core/esp.h"
+#include "esp/io/io.h"
+
+#include <sstream>
+
+// This is to import the "resources" at runtime. When the resource is
+// compiled into static library, it must be explicitly initialized via this
+// macro, and should be called *outside* of any namespace.
+static void importShaderResources() {
+  CORRADE_RESOURCE_INITIALIZE(ShaderResources)
+}
+
+namespace Mn = Magnum;
+namespace Cr = Corrade;
+
+namespace esp {
+namespace gfx {
+
+namespace {
+enum TextureUnit : uint8_t {
+  BASE_COLOR = 0,
+  ROUGHNESS = 1,
+  METALLIC = 2,
+  NORMAL = 3,
+  // NoneRoughnessMetallic or OcclusionRoughnessMetallic texture
+  PACKED = 4,
+  EMISSIVE = 5,
+};
+}  // namespace
+
+PbrShader::Flags PbrShader::generateCorrectFlags(Flags originalFlags) {
+  Flags result = originalFlags;
+  // NOTE:
+  // The priority of different kind of textures is as follows (priority
+  // means if two textures with different priorities exist at the same time,
+  // shader will adopt the texture with the higher priority, and ignore the
+  // other one.)
+  // 0 (highest): OcclusionRoughnessMetallicTexture
+  // 1          : NoneRoughnessMetallicTexture
+  // 2          : RoughnessTexture, MetallicTexture
+  if (result & Flag::OCCLUSION_ROUGHNESS_METALLIC_TEXTURE) {
+    result &= ~Flag::NONE_ROUGHNESS_METALLIC_TEXTURE;
+    result &= ~Flag::ROUGHNESS_TEXTURE;
+    result &= ~Flag::METALLIC_TEXTURE;
+  } else if (result & Flag::NONE_ROUGHNESS_METALLIC_TEXTURE) {
+    result &= ~Flag::ROUGHNESS_TEXTURE;
+    result &= ~Flag::METALLIC_TEXTURE;
+  }
+
+  return result;
+}
+
+PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
+    : flags_(generateCorrectFlags(originalFlags)), lightCount_(lightCount) {
+  if (!Cr::Utility::Resource::hasGroup("default-shaders")) {
+    importShaderResources();
+  }
+
+#ifdef MAGNUM_TARGET_WEBGL
+  Mn::GL::Version glVersion = Mn::GL::Version::GLES300;
+#else
+  Mn::GL::Version glVersion = Mn::GL::Version::GL410;
+#endif
+
+  // this is not the file name, but the group name in the config file
+  // see Shaders.conf in the shaders folder
+  const Cr::Utility::Resource rs{"default-shaders"};
+
+  Mn::GL::Shader vert{glVersion, Mn::GL::Shader::Type::Vertex};
+  Mn::GL::Shader frag{glVersion, Mn::GL::Shader::Type::Fragment};
+
+  std::stringstream attributeLocationsStream;
+  attributeLocationsStream << Cr::Utility::formatString(
+      "#define ATTRIBUTE_LOCATION_POSITION {}\n", Position::Location);
+  attributeLocationsStream << Cr::Utility::formatString(
+      "#define ATTRIBUTE_LOCATION_NORMAL {}\n", Normal::Location);
+  if (flags_ & (Flag::NORMAL_TEXTURE | Flag::PRECOMPUTED_TANGENT) &&
+      lightCount) {
+    attributeLocationsStream << Cr::Utility::formatString(
+        "#define ATTRIBUTE_LOCATION_TANGENT4 {}\n", Tangent4::Location);
+  }
+  auto isTextured = [&]() {
+    return (flags_ &
+            (Flag::BASE_COLOR_TEXTURE | Flag::ROUGHNESS_TEXTURE |
+             Flag::METALLIC_TEXTURE | Flag::NORMAL_TEXTURE |
+             Flag::EMISSIVE_TEXTURE | Flag::NONE_ROUGHNESS_METALLIC_TEXTURE |
+             Flag::OCCLUSION_ROUGHNESS_METALLIC_TEXTURE));
+  };
+  if (isTextured()) {
+    attributeLocationsStream
+        << Cr::Utility::formatString("#define ATTRIBUTE_LOCATION_TEXCOORD {}\n",
+                                     TextureCoordinates::Location);
+  }
+
+  // Add macros
+  vert.addSource(attributeLocationsStream.str())
+      .addSource(isTextured() ? "#define TEXTURED\n" : "")
+      .addSource(flags_ & Flag::NORMAL_TEXTURE ? "#define NORMAL_TEXTURE\n"
+                                               : "")
+      .addSource(flags_ & Flag::PRECOMPUTED_TANGENT
+                     ? "#define PRECOMPUTED_TANGENT\n"
+                     : "")
+      .addSource(flags_ & Flag::TEXTURE_TRANSFORMATION
+                     ? "#define TEXTURE_TRANSFORMATION\n"
+                     : "")
+      .addSource(rs.get("pbr.vert"));
+
+  std::stringstream outputAttributeLocationsStream;
+  outputAttributeLocationsStream << Cr::Utility::formatString(
+      "#define OUTPUT_ATTRIBUTE_LOCATION_COLOR {}\n", ColorOutput);
+  outputAttributeLocationsStream << Cr::Utility::formatString(
+      "#define OUTPUT_ATTRIBUTE_LOCATION_OBJECT_ID {}\n", ObjectIdOutput);
+
+  frag.addSource(attributeLocationsStream.str())
+      .addSource(outputAttributeLocationsStream.str())
+      .addSource(isTextured() ? "#define TEXTURED\n" : "")
+      .addSource(flags_ & Flag::BASE_COLOR_TEXTURE
+                     ? "#define BASECOLOR_TEXTURE\n"
+                     : "")
+      .addSource(flags_ & Flag::EMISSIVE_TEXTURE ? "#define EMISSIVE_TEXTURE\n"
+                                                 : "")
+      .addSource(
+          flags_ & Flag::ROUGHNESS_TEXTURE ? "#define ROUGHNESS_TEXTURE\n" : "")
+      .addSource(flags_ & Flag::METALLIC_TEXTURE ? "#define METALLIC_TEXTURE\n"
+                                                 : "")
+      .addSource(flags_ & Flag::NONE_ROUGHNESS_METALLIC_TEXTURE
+                     ? "#define NONE_ROUGHNESS_METALLIC_TEXTURE\n"
+                     : "")
+      .addSource(flags_ & Flag::OCCLUSION_ROUGHNESS_METALLIC_TEXTURE
+                     ? "#define OCCLUSION_ROUGHNESS_METALLIC_TEXTURE\n"
+                     : "")
+      .addSource(flags_ & Flag::NORMAL_TEXTURE ? "#define NORMAL_TEXTURE\n"
+                                               : "")
+      .addSource(flags_ & Flag::NORMAL_TEXTURE_SCALE
+                     ? "#define NORMAL_TEXTURE_SCALE\n"
+                     : "")
+      .addSource(flags_ & Flag::OBJECT_ID ? "#define OBJECT_ID\n" : "")
+      .addSource(flags_ & Flag::PRECOMPUTED_TANGENT
+                     ? "#define PRECOMPUTED_TANGENT\n"
+                     : "")
+      .addSource(
+          Cr::Utility::formatString("#define LIGHT_COUNT {}\n", lightCount))
+      .addSource(rs.get("pbr.frag"));
+
+  CORRADE_INTERNAL_ASSERT_OUTPUT(Mn::GL::Shader::compile({vert, frag}));
+
+  attachShaders({vert, frag});
+
+  CORRADE_INTERNAL_ASSERT_OUTPUT(link());
+
+  // bind attributes
+#ifndef MAGNUM_TARGET_GLES
+  if (!Mn::GL::Context::current()
+           .isExtensionSupported<
+               Mn::GL::Extensions::ARB::explicit_attrib_location>(glVersion))
+#endif
+  {
+    bindAttributeLocation(Position::Location, "vertexPosition");
+    if (lightCount) {
+      bindAttributeLocation(Normal::Location, "vertexNormal");
+    }
+    if (flags_ & (Flag::NORMAL_TEXTURE | Flag::PRECOMPUTED_TANGENT) &&
+        lightCount) {
+      bindAttributeLocation(Tangent4::Location, "vertexTangent");
+    }
+    if (isTextured()) {
+      bindAttributeLocation(TextureCoordinates::Location, "vertexTexCoord");
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


src/shaders/pbr.frag
```diff
@@ -0,0 +1,292 @@
...
+  // See here: http://www.thetenthplanet.de/archives/1180
+  vec3 q1 = dFdx(position);
+  vec3 q2 = dFdy(position);
+  vec2 st1 = dFdx(texCoord);
+  vec2 st2 = dFdy(texCoord);
+
+  vec3 N = normalize(normal);
+  vec3 T = normalize(q1 * st2.t - q2 * st1.t);
+  vec3 B = -normalize(cross(N, T));
+  mat3 TBN = mat3(T, B, N);
```
mosra, line 111: Didn't we agree we'd drop this and render without a normal map instead if there are no tangents (like done with `GenericDrawable` in #837) ? Since, as you said, this caused some weird artifacts in some cases.

mosra (changes requested):
Compared to #837 and #838, which were the right size, I fear there's too little in this PR -- it should have the C++ PbrShader class as well, the GLSL sources alone have no use.

src/shaders/pbr.frag
```diff
@@ -0,0 +1,292 @@
...
+// if .w == 0, it means it is a dirctional light, .xyz is the direction;
+// if .w == 1, it means it is a point light, .xyz is the light position;
+// it is NOT put in the Light Structure, simply because we may modify the code
+// so it is computed in the vertex shader.
+uniform vec4 LightDirections[LIGHT_COUNT];
+#endif
+
+// -------------- material, textures ------------------
+struct MaterialData {
+  vec4 baseColor;     // diffuse color, if baseColorTexture exists,
```
eundersander, line 42: Does the alpha component mean anything?

src/shaders/pbr.frag
```diff
@@ -0,0 +1,292 @@
...
+#if defined(NORMAL_TEXTURE) && defined(NORMAL_TEXTURE_SCALE)
+uniform mediump float normalTextureScale
+#ifndef GL_ES
+    = 1.0
+#endif
+    ;
+#endif
+
+// -------------- shader ------------------
+#if defined(NORMAL_TEXTURE)
```
eundersander, line 88: Does GLSL support #include's ? It seems this normal-mapping logic could be reused in other shaders. It's not specific to PBR.

src/shaders/pbr.frag
```diff
@@ -0,0 +1,292 @@
...
+    // radiance
+    vec3 lightRadiance = LightColors[iLight] * attenuation;
+
+    // light source direction: a vector from current position to the light
+    vec3 light = normalize(LightDirections[iLight].xyz -
+                           position * LightDirections[iLight].w);
+
+    finalColor += microfacetModel(baseColor.rgb, metallic, roughness, n, light,
+                                  view, lightRadiance);
+    finalAlpha += baseColor.w;
```
eundersander, line 282: I don't understand this line.

src/shaders/pbr.frag
```diff
@@ -0,0 +1,292 @@
...
+    // attenuation is always 1.0
+    highp float dist =
+        length(LightDirections[iLight].xyz) * LightDirections[iLight].w;
+    // If range is 0 for whatever reason, clamp it to a small value to
+    // avoid a NaN when dist is 0 as well (which is the case for
+    // directional lights).
+    highp float attenuation =
+        clamp(1.0 - pow(dist / max(LightRanges[iLight], 0.0001), 4.0), 0.0,
+              1.0) /
+        (1.0 + dist);
```
eundersander, line 270: Nit: @mosra  , I saw you did this divide by `(1.0 + dist)` in the phong shader, too. `1 / (1.0 + dist)^2` is not quite the same as `1 / dist^2`. In particular, it behaves very differently for small distances. Isn't this deviating from the standard falloff calculation?

mosra: Yeah I spent some time investigating this already ([check out the docs](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1LightData.html#Trade-LightData-attenuation)) -- If you don't add the 1, it'll explode to infinity at `dist` close to 0, which I don't think is desired :) 

For some reason the glTF spec [doesn't include that in the equation](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual#range-property), but the UE4 calculation (the glTF equation was based on) [has it](http://blog.selfshadow.com/publications/s2013-shading-course/karis/s2013_pbs_epic_notes_v2.pdf):

image

Hmm ... but it should be `(1.0 + dist^2)` not `(1.0 + dist)^2`. I have to check why I did that.

src/shaders/pbr.frag
mosra: Not really, unless the shader is preprocessed externally first (which is what I'm working on right now, actually). However you can supply multiple source files and do it that way instead.

On the other hand, I don't think it's likely Habitat will have more shaders? As in, the shader will eventually be replaced with one from Magnum instead.

src/shaders/pbr.frag
mosra: Yep, glTF supports alpha-masked and alpha-blended materials.

bigbike: It means a lot.
We need the alpha value if the object is a transparent or semi-transparent object.

src/shaders/pbr.frag
bigbike: I remembered I thought about this question and googled it a couple of years ago. And I decided not to share the code among shaders.
Take a look:
https://computergraphics.stackexchange.com/questions/100/sharing-code-between-multiple-glsl-shaders

src/shaders/pbr.frag
bigbike: @eundersander: `.w` should between 0.0 and 1.0. Correct? :)

src/shaders/pbr.frag
mosra: As magnum handles `#version` itself because of exactly the reasons mentioned in the SO answer, the only remaining valid point from there is

> 2. It's harder to control what needs to be included from within the shader (you need a separate system for this.)

Yes, that's true, but then again you have to prepend all the defines for `LIGHT_COUNT`, `NORMAL_TEXTURE` etc anyway, so having a separate external file doesn't make it *that much* worse.

src/shaders/pbr.frag
bigbike: @eundersander : You concern sounds reasonable.

About the attenuation:
Here, I followed what was used in the Phong shader in Magnum. When I coded it up, I understand the `1.0` is to avoid division by 0.
I did not dig into it further. But if the paper says it should be `1 + dist^2`, I will change it.

src/shaders/pbr.frag
bigbike: @mosra : Yes, we had that agreement.

But later I thought it might be good to leave it here as a fallback. It was correct in most cases.
I saw many systems using this trick, for example, here: 
https://github.com/SaschaWillems/Vulkan-glTF-PBR/blob/master/data/shaders/pbr_khr.frag#L125

Yes, true, there are artifacts in some corner cases. If you still feel uncomfortable, I can remove it though.

bigbike:
> Compared to #837 and #838, which were the right size, I fear there's too little in this PR -- it should have the C++ PbrShader class as well, the GLSL sources alone have no use.

I just try to make the reviewer's life easier. I was planning to add the PbrShader classes in the next PR. But since you asked, I will put it in this PR.

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,486 @@
...
+
+namespace Mn = Magnum;
+namespace Cr = Corrade;
+
+namespace esp {
+namespace gfx {
+
+namespace {
+enum TextureUnit : uint8_t {
+  BASE_COLOR = 0,
```
bigbike, line 44: Well, we just discussed. I will change them back to `PascalCase`.

src/shaders/pbr.frag
Skylion007: @mosra Adopting libshader_rc for Magnum would automatically handle the #INCLUDE preprocessing as well. :)

src/shaders/pbr.frag
bigbike: I just updated the code using the `1+dist^2` now. Take a look.

bigbike:
PbrShader class is added in this diff. It is ready to be reviewed.
I hope we can commit it in as soon as possible as other PRs are being blocked by it. Thanks a lot!

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,481 @@
...
+// it requires GL4.3 (We stick to GL4.1 for MacOS).
+PbrShader& PbrShader::bindBaseColorTexture(Mn::GL::Texture2D* texture) {
+  if ((flags_ & Flag::BaseColorTexture) && lightCount_ && texture) {
+    texture->bind(TextureUnit::BaseColor);
+  }
+  return *this;
+}
+
+PbrShader& PbrShader::bindRoughnessTexture(Mn::GL::Texture2D* texture) {
+  if ((flags_ & Flag::RoughnessTexture) && lightCount_ && texture) {
```
mosra, line 253: Are we really not going with references here? What does the coding style guide say about this? Because I remember (mutable) non-nullable references being used quite on a lot of places in Habitat, and using a reference would make it consistent with what Magnum does -- there a pointer is used *only* if it's allowed to be null.

I don't buy the argument that it makes it "easier to use" because MaterialData also use pointers. 

There's an additional concern I have with silent failures -- if the texture is `nullptr` for some reason or the shader doesn't have `RoughnessTexture` enabled for some reason, it will just silently do nothing. In my opinion it neither makes sense to call `bindRoughnessTexture(nullptr)` nor to call `bindRoughnessTexture()` when the shader doesn't have that enabled, so none of that should be allowed:

1. Accept a reference, so if the caller has a pointer for some reason, *the caller* is responsible for dereferencing it. It's not the shader's business to do error handling on behalf of the caller.
2. *Assert* if this function is called and `RoughnessTexture` is not enabled, instead of silently doing a bad render because I don't know, the shader got compiled with packed textures enabled and you're passing separate maps instead. The assert is what Magnum does as well, and from experience I can say it caught quite a lot of mistakes that would take significant time to discover and debug otherwise.

Cc: @aclegg3 @jturner65

src/esp/gfx/PbrShader.h
```diff
@@ -0,0 +1,453 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#ifndef ESP_GFX_PBRSHADER_H_
+#define ESP_GFX_PBRSHADER_H_
+
+#include <initializer_list>
+#include <vector>
```
mosra, line 9: Unneeded include.

src/esp/gfx/PbrShader.h
```diff
@@ -0,0 +1,453 @@
...
+
+#include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Containers/EnumSet.h>
+#include <Magnum/GL/AbstractShaderProgram.h>
+#include <Magnum/Math/Matrix4.h>
```
mosra, line 14: *Technically* you don't need this include either as there's a forward declaration that's enough. 

OTOH, if *do* you want to have the full definition here, you'll need `Matrix3.h` as well for `setTextureMatrix()`.

mosra (commented):
Eh, sorry, I thought I'd just approve but more things cropped up.

Will reply to the remaining threads separately.

src/shaders/pbr.frag
mosra: Eric means it's an useless operation, you could as well do just 

```glsl
fragmentColor += vec4(finalColor, baseColor.a);
```

below and don't calculate any `finalAlpha`.

src/shaders/pbr.frag
mosra: @Skylion007 Yeah no :P Includes are supported directly by glslang itself, it's just not exposed in the `glslangValidator` command-line tool for whatever reason.

Which means, at this point, I don't see any benefits of using shaderc over glslang + spirv-tools directly, it would only add another hard-to-build dependency to the pipeline and further expand the surface for potential bugs.

src/shaders/pbr.frag
mosra: It's easy to add tangents in an offline processing step (via Blender, for example) and most models already have them. The only PBR models that *didn't* have those were the ones exported from Blender without the Tangents option checked. 

This additional code creates another shader variant we need to test and maintain, so I'd vote for removing it -- the best code is code that's not written :P

src/shaders/pbr.frag
mosra: Okay, then I'll do the same change in Phong. Will get back to you if I discover any issues with that.

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,481 @@
...
+    attributeLocationsStream << Cr::Utility::formatString(
+        "#define ATTRIBUTE_LOCATION_TANGENT4 {}\n", Tangent4::Location);
+  }
+  auto isTextured = [&]() {
+    return (flags_ &
+            (Flag::BaseColorTexture | Flag::RoughnessTexture |
+             Flag::MetallicTexture | Flag::NormalTexture |
+             Flag::EmissiveTexture | Flag::NoneRoughnessMetallicTexture |
+             Flag::OcclusionRoughnessMetallicTexture));
+  };
```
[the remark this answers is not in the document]
bigbike: Yes, it makes sense.

src/esp/gfx/PbrShader.cpp
bigbike: > I don't buy the argument that it makes it "easier to use" because MaterialData also use pointers.

Yes, for the reason "you do not buy". It is much more convenient to use pointer here. In the render() in PbrDrawable (which is not in this PR), I can use the method chaining to call them, and do not need to care about if the texture exists, or if the shader is initialized with this feature.
```
      .bindBaseColorTexture(materialData_->baseColorTexture)
      .bindRoughnessTexture(materialData_->roughnessTexture)
      .bindMetallicTexture(materialData_->metallicTexture)
      .bindNoneRoughnessMetallicTexture(
          materialData_->noneRoughnessMetallicTexture)
      .bindOcclusionRoughnessMetallicTexture(
          materialData_->occlusionRoughnessMetallicTexture)
      .bindNormalTexture(materialData_->normalTexture)
      .bindEmissiveTexture(materialData_->emissiveTexture)
```
If you change it to the reference (which I tried indeed) you will have a lot of if clause here.

This issue actually reflects different view between us in terms of the robustness of the code.
- I tried to keep the system running, not being interrupted (terminated) due to developers' careless mistakes (NON-fatal errors, e.g., there is actually no roughness texture, but she bound it), and I hided the sanity checks within my APIs to make sure the system will not crash due to these "mistakes". I choose a more tolerant way, and provide easy-to-use APIs (e.g., here you can use method chaining, without any if clauses, and you do not have to remember if you activate corresponding feature or not). "Lazy" developers are going to like it. :)

- You used a more active way to encourage the developer "fail" earlier. As you mentioned, " I can say it caught quite a lot of mistakes that would take significant time to discover and debug otherwise." which I agree. 

In short, the difference is that you leave such overhead (sanity check, dereferencing, assert etc.) to the developer, but I do not. No right or wrong here. Just different style.

I just explain to let you know my view. And I do not think one is more appealing than the other one.

I can change to the way you suggested. I am OK with it.

src/esp/gfx/PbrShader.cpp
```diff
@@ -0,0 +1,481 @@
...
+  }
+  return *this;
+}
+
+PbrShader& PbrShader::setMVPMatrix(const Mn::Matrix4& matrix) {
+  setUniform(mvpMatrixUniform_, matrix);
+  return *this;
+}
+
+PbrShader& PbrShader::setNormalMatrix(const Mn::Matrix3& matrix) {
```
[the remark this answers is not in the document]
bigbike: Oh, good to know!
Not sure if you particularly mentioned it in your wiki. If you did not, I recommend you to do that.
I was thinking `Matrix3` is just a convenient `typedef` on top of the `Matrix3x3`.

bigbike (commented):
Thanks a lot @mosra. I saw many constructive feedback for you. I really appreciated it.

src/shaders/pbr.vert
```diff
@@ -0,0 +1,62 @@
...
+#endif
+
+// ------------ uniform ----------------------
+uniform highp mat4 ModelViewMatrix;
+uniform highp mat3 NormalMatrix;  // inverse transpose of 3x3 modelview matrix
+uniform mat4 MVP;
+
+#ifdef TEXTURE_TRANSFORMATION
+uniform highp mat3 textureMatrix
+#ifndef GL_ES
```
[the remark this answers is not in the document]
mosra: Okay, then this `textureMatrix` should be `TextureMatrix` instead :)

src/shaders/pbr.frag
bigbike: Oh, OK, I see.
The add should happen outside the for loop since baseColor.a never changes.
fragmentColor += vec4(finalColor, baseColor.a);

src/shaders/pbr.vert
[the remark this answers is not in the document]
bigbike: will do.

src/shaders/pbr.vert
```diff
@@ -0,0 +1,62 @@
...
+#if defined(NORMAL_TEXTURE) && defined(PRECOMPUTED_TANGENT)
+  tangent = normalize(NormalMatrix * vec3(vertexTangent));
+  // Gram–Schmidt
+  tangent = normalize(tangent - dot(tangent, normal) * normal);
+  biTangent = normalize(cross(normal, tangent) * vertexTangent.w);
+  // later in .frag, TBN will transform the normal perturbation
+  // (read from normal map) from tangent space to camera space
+#endif
+
+  gl_Position = MVP * vertexPosition;
```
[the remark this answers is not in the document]
mosra: To clarify, I have nothing against the notion of a model-view-projection matrix. It's just that I find it more logical to supply a separate `modelview` and `projection` parts than `modelview` and `modelviewprojection`, where the same information is encoded twice.

And as a bonus it'll also remove one unnecessary matrix multiplication in the Drawable implementation, if I'm not mistaken.

src/esp/gfx/PbrShader.cpp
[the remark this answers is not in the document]
mosra: Good point. [It is mentioned at one place](https://doc.magnum.graphics/magnum/namespaceMagnum.html#a7ee135be46a327f352d01287579e2acf), but I'll put that information elsewhere as well.

src/esp/gfx/PbrShader.cpp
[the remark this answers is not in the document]
bigbike: by the way, how to convert:
```
flags_ &
            (Flag::BaseColorTexture | Flag::RoughnessTexture |
             Flag::MetallicTexture | Flag::NormalTexture |
             Flag::EmissiveTexture | Flag::NoneRoughnessMetallicTexture |
             Flag::OcclusionRoughnessMetallicTexture);
```
to a `bool`. (currently I am using auto).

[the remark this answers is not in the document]
mosra: It has an [explicit operator bool](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html#aaa72f9d5d7ab6e51d66d3698206e7cc9), so either `!!(thiswholething)` or `bool(thiswholething)`.

Implicit bool conversion would make it convertible to an int, which is not desired from a type safety perspective.

src/esp/gfx/PbrShader.cpp
mosra: Funnily enough right now I'm fighting `glslang` which is built with this "fail silently" approach and it costs me an immense amount of time because a lot of the time it just does nothing and i don't know *why* :laughing: 

Yep, It's a matter of perspective. Could we get opinions from others here? I don't want to push my view if I would the only one who wants it that way.

src/esp/gfx/PbrShader.h
bigbike: I do not need either of them in the .h file.

src/esp/gfx/PbrShader.cpp
bigbike: I changed them to use the reference here, and also added assertions within the API to make it fail immediately in illegal usage cases. It is worth if "it caught quite a lot of mistakes that would take significant time to discover and debug otherwise." We can continue the conversation on this topic in the slack channel.

This PR is out for quite a long time, and I hope there is no more further delays.

mosra (approved):
:tada:

Skylion007 (approved):

Skylion007:
@bigbike LGTM lets get this merged! :)

src/esp/gfx/PbrShader.cpp
bigbike: @eundersander : Here is the discussion on the design philosophy.

PR #843, opened 2020-10-08
update doc build system

Skylion007:
Really quick PR to fix bug in mcss.

See https://github.com/facebookresearch/habitat-lab/pull/492#issuecomment-705621004 for more info


mosra:
Can you update to latest master instead of just that particular bugfix commit? There's a bunch of other tiny fixes that would be nice to have :)

Skylion007:
@mosra Anything else we should update? Magnum or such? I updated some of the other submodules if they built locally.

mosra:
I personally prefer to "don't fix what's not broken", so I'd bump just the m.css submodule, that's what the PR is for after all ;) Especially the `tiny*` libraries are, in my experience, not always stable, so even if something builds it doesn't mean it works.

I'll open a Magnum update PR once I land a larger chunk of work, there's not much new interesting stuff right now.

Skylion007:
@mosra What about IMGUI? Seems like mostly bugfixes and better cmake support?

mosra:
CMake support is not relevant to Magnum, and I'm not aware of any bugs that would affect us.

Don't fix what's not broken, again ;)

mosra (approved):

PR #846, opened 2020-10-09
Unify RigidObject construction pipeline and MotionType treatment.

aclegg3:
## Motivation and Context

Recent changes brought all MotionTypes into the `btRigidBody` API. This PR cleans up unnecessary separate treatment and unifies the construction pipeline.

Note: some changes to existing physical simulation results are expected as `STATIC` objects will now have the properly configured friction, restitution, and damping.

Velocity setting for KINEMATIC objects is allowed. These will not be integrated, but may improve Bullet internal contact response with DYNAMIC objects.

## How Has This Been Tested

All tests should still pass.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletRigidObject.cpp
+++ b/src/esp/physics/bullet/BulletRigidObject.cpp
@@ -46,16 +46,9 @@ BulletRigidObject::~BulletRigidObject() {
     activateCollisionIsland();
   }
 
-  if (objectMotionType_ != MotionType::STATIC) {
-    // remove rigid body from the world
-    bWorld_->removeRigidBody(bObjectRigidBody_.get());
-  } else if (objectMotionType_ == MotionType::STATIC) {
-    // remove collision objects from the world
-    for (auto& co : bStaticCollisionObjects_) {
-      bWorld_->removeRigidBody(co.get());
-      collisionObjToObjIds_->erase(co.get());
-    }
-  }
+  // remove rigid body from the world
+  bWorld_->removeRigidBody(bObjectRigidBody_.get());
+
   collisionObjToObjIds_->erase(bObjectRigidBody_.get());
 
 }  //~BulletRigidObject
@@ -127,12 +120,10 @@ bool BulletRigidObject::initialization_LibSpecific(
   bObjectShape_->setMargin(margin);
 
   bObjectShape_->setLocalScaling(btVector3{tmpAttr->getScale()});
-
+  bObjectShape_->recalculateLocalAabb();
   // create the bObjectRigidBody_
-  constructRigidBody(false);
+  constructAndAddRigidBody(objectMotionType_);
 
-  //! Add to world
-  bWorld_->addRigidBody(bObjectRigidBody_.get());
   return true;
 }  // initialization_LibSpecific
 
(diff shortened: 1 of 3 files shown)


jturner65 (approved):
LGTM

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -382,14 +334,19 @@ void BulletRigidObject::constructRigidBody(bool kinematic) {
   }
 
   //! Bullet rigid body setup
+  auto motionState = (mt == MotionType::STATIC) ? nullptr : &(btMotionState());
```
eundersander, line 337: Taking the address of a temporary object that immediately goes out of scope? This looks shady.

Also, why do you even want to pass a btMotionState here? Is it needed?

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -382,14 +334,19 @@ void BulletRigidObject::constructRigidBody(bool kinematic) {
...
-      btRigidBody::btRigidBodyConstructionInfo(mass, &(btMotionState()),
+      btRigidBody::btRigidBodyConstructionInfo(mass, motionState,
                                                bObjectShape_.get(), bInertia);
 
   info.m_friction = tmpAttr->getFrictionCoefficient();
   info.m_restitution = tmpAttr->getRestitutionCoefficient();
   info.m_linearDamping = tmpAttr->getLinearDamping();
   info.m_angularDamping = tmpAttr->getAngularDamping();
+  if (bObjectRigidBody_ != nullptr) {
+    info.m_startWorldTransform = bObjectRigidBody_->getWorldTransform();
```
eundersander, line 348: btTransform doesn't initialize to identify or anything else by default. It seems dangerous to not initialize it for the (bObjectRigidBody_ == nullptr) case.

src/esp/physics/bullet/BulletRigidObject.cpp
jturner65: > Taking the address of a temporary object that immediately goes out of scope? This looks shady.

`
btMotionState& btMotionState() { return *this; }

`
The BulletRigidObject inherits from Magnum::BulletIntegration::MotionState.  This is a way to pass a ref to the BulletRigidObject itself into the btRigidBodyConstructionInfo.

eundersander: Haha, I see now. I guess I've never considered that a method can be the same name as a type. A little confusing to read this line, but anyway that was Magnum's decision in MotionState. @mosra FYI

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: `info.m_startWorldTransform` is initialized to identity in the constructor: https://pybullet.org/Bullet/BulletFull/btRigidBody_8h_source.html#l00119.
Providing a `btMotionState` will override this transform also. Really this could be reduced to the STATIC case only, but this condition seemed more general.

src/esp/physics/bullet/BulletRigidObject.cpp
mosra: Hah, yeah here it looks very bad :laughing: To deconfuse you could

```suggestion
  auto motionState = (mt == MotionType::STATIC) ? nullptr : &(this->btMotionState());
```

PR #863, opened 2020-10-16
[PBR 3] PbrDrawables and Integration with the simulator

bigbike:
## Motivation and Context
In this PR:
- Developed PbrDrawables class;
- Integrate the PBR rendering framework with the simulator so that system would initialize and activate PBR rendering when PBR material is detected.

## Types of changes

## Checklist


bigbike:
Tested it with 2 different PBR models:
1) water bottle with packed textures (metallic, roughness) + normal textures + emissive textures


2) A big lantern model. This is to test the rendering with the quantized texture:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Lantern/glTF-Quantized

src/esp/assets/ResourceManager.cpp
```diff
@@ -1192,13 +1193,16 @@ void ResourceManager::loadMaterials(Importer& importer,
...
             Magnum::Trade::MaterialType::PbrMetallicRoughness) {
-      const Mn::Trade::PbrMetallicRoughnessMaterialData&
-          pbrMetallicRoughnessMaterialData =
-              static_cast<Mn::Trade::PbrMetallicRoughnessMaterialData&>(
-                  *materialData);
+      const auto& pbrMaterialData =
+          materialData->as<Mn::Trade::PbrMetallicRoughnessMaterialData>();
 
-      finalMaterial = gfx::buildPhongFromPbrMetallicRoughness(
-          pbrMetallicRoughnessMaterialData, textureBaseIndex, textures_);
```
bigbike, line 1201: @eundersander :

Is it a bug here? Is it missing a `continue`?
Here if it requires lighting and the material is PBR, then it will initialize the `finalMaterial` as Phong material using PBR material. It should skip the rest of the code and continue the next loop. Correct?
Otherwise the following code will build the `finalMaterial` as `PhongShaderMaterialData` anyway.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1646,33 +1746,50 @@ void ResourceManager::addPrimitiveToDrawables(int primitiveID,
...
+          meshAttributeFlags,  // mesh attribute flags
+          shaderManager_,      // shader manager
+          lightSetupKey,       // lightSetup key
+          materialKey,         // material key
+          group);              // drawable group
+      break;
+    default:
+      CORRADE_INTERNAL_ASSERT_UNREACHABLE();
+      break;
+  }
```
mosra, line 1792: Continuing from [the disccussion here](https://github.com/facebookresearch/habitat-sim/pull/812#discussion_r499002662) -- I want to understand your intention better:

> I just would like to get rid of such unnecessary warning so I used "default".

Why do you think it's unnecessary? If somebody adds a new material data type to the enum later (`Flat` or `PbrSpecularGlossiness` let's say), it will helpfully tell them what `switch`es need an update. What's unnecessary on that? 

And if the enum contains other values you don't want to handle, you'd just explicitly list them:

```cpp
  switch (materialDataType) {
    case gfx::MaterialDataType::Phong:
      node.addFeature(...);
      return;
    case gfx::MaterialDataType::Pbr:
      node.addFeature(...);
      return;

    case gfx::MaterialDataType::None:
}

```

src/esp/gfx/MaterialData.h
```diff
@@ -15,21 +15,79 @@
...
+  // a packed texture to store "roughness" (in G channel) and "metallic" (in B
+  // channel)
+  Magnum::GL::Texture2D* noneRoughnessMetallicTexture = nullptr;
+  // a packed texture to store "occlusion"(in R channel), "roughness" (in G
+  // channel) and "metallic" (in B channel)
+  Magnum::GL::Texture2D* occlusionRoughnessMetallicTexture = nullptr;
+  // TODO:
+  // AO texture
+  ESP_SMART_POINTERS(PbrMaterialData)
+};
```
mosra, line 89: Yeah so because this class will inevitably grow further as support for new map packing and material properties gets added, this is where I originally thought Magnum's [Trade::MaterialData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1MaterialData.html) could get used, instead of maintaining a downstream copy of it. It supports everything listed here, the colors and properties can be specified via builtin attributes and the texture pointers [via custom attributes](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1MaterialData.html#Trade-MaterialData-populating-custom).

Opinions? @aclegg3 @eundersander

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -0,0 +1,211 @@
...
+  if (materialData_->normalTexture) {
+    flags_ |= PbrShader::Flag::NormalTexture;
+    if (meshAttributeFlags & gfx::Drawable::Flag::HasTangent) {
+      flags_ |= PbrShader::Flag::PrecomputedTangent;
+    }
+    if (materialData_->normalTextureScale != 1.0f) {
+      flags_ |= PbrShader::Flag::NormalTextureScale;
+      CORRADE_ASSERT(materialData_->normalTextureScale > 0.0f,
+                     "PbrDrawable::PbrDrawable(): the normal texture scale "
+                     "cannot be negative.", );
```
mosra, line 56: I know I'm annoyingly picky here, but

```suggestion
                     "has to be positive.", );
```

:trollface:

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// update every light's color, intensity, range etc.
+PbrDrawable& PbrDrawable::updateShaderLightParameters() {
+  constexpr float dummyRange = Mn::Constants::inf();
+  std::vector<float> ranges;
+  std::vector<Mn::Color3> colors;
+  colors.reserve(lightSetup_->size());
+  for (unsigned int iLight = 0; iLight < lightSetup_->size(); ++iLight) {
+    ranges.push_back(dummyRange);
```
mosra, line 181: Remember my [brainfart comment](https://github.com/facebookresearch/habitat-sim/pull/812#discussion_r497328490)?  Just to finish that -- in Magnum's builtin shaders, the shader has reasonable defaults right after its construction. So if you do

```cpp
Shaders::Whichever shader;
// and maybe set the transformation matrices
shader.draw(mesh);
```

it will always do *something* and won't result in a black screen -- if you don't set any light parameters, it'll use reasonable defaults, it you don't set any color, it'll use white, if you don't set any transformation, it'll use identities etc.

Now, it's up to you if you want to have the same "reasonable defaults" behavior in Habitat or not (at which point, you wouldn't need to set implicit infinity ranges because those are there by default). I could imagine that it could help to whoever developer comes next into the rendering code here, or if it gets to the point where Python users are instantiating their own meshes and using shaders, this might be a lot of help, but maybe also not. Up to you :)

src/shaders/pbr.vert
```diff
@@ -38,7 +38,7 @@ uniform highp mat3 TextureMatrix
 #endif
 
 void main() {
-  vertexPositionInCamera = ModelViewMatrix * vertexPosition;
+  vec4 vertexPositionInCamera = ModelViewMatrix * vertexPosition;
```
mosra, line 41: Hah. That reminds me -- you know what this PR (and #840) is absolutely missing? *Tests.*

I don't think you need to go extra fancy compared to how massively I'm testing shaders in Magnum, but at least:

1. have a test that tries to compile the shader with all legal flag combinations and verifies all those work without any compilation error (see e.g. [PhongGLTest](https://github.com/mosra/magnum/blob/c226bc2d630f886969509e5a7b13d0a8513c6c70/src/Magnum/Shaders/Test/PhongGLTest.cpp#L596-L617) in Magnum for a quick nonverbose way how to do that)
2. comparing a render of some PBR model to *at least one* ground truth image. This could use Magnum's [CompareImage](https://doc.magnum.graphics/magnum/classMagnum_1_1DebugTools_1_1CompareImage.html) functionality, which is already used in e.g. SimTest I think -- look there for how it's used. When you write the test, run it with `--save-diagnostics some/path` to get or update the ground truth images.

Skylion007:
@bigbike Error seem to be occurring when we have PBR materials in a scene that don't have any lights. It should probably just disable PBR rendering if this is the case, not error.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1192,13 +1193,16 @@ void ResourceManager::loadMaterials(Importer& importer,
...
-      const Mn::Trade::PbrMetallicRoughnessMaterialData&
-          pbrMetallicRoughnessMaterialData =
-              static_cast<Mn::Trade::PbrMetallicRoughnessMaterialData&>(
-                  *materialData);
+      const auto& pbrMaterialData =
+          materialData->as<Mn::Trade::PbrMetallicRoughnessMaterialData>();
 
-      finalMaterial = gfx::buildPhongFromPbrMetallicRoughness(
-          pbrMetallicRoughnessMaterialData, textureBaseIndex, textures_);
+      if (flags_ & Flag::BuildPhongFromPbr) {
```
Skylion007, line 1199: | `num_lights == 0`

src/esp/assets/ResourceManager.cpp
```diff
@@ -1192,13 +1193,16 @@ void ResourceManager::loadMaterials(Importer& importer,
     if (loadedAssetData.assetInfo.requiresLighting &&
```
Skylion007, line 1193: Is requiresLighting, the same as enables lighting? Like can you load an asset that requires lighting into a dark scene? If so don't we need additional checks for the next if statement?

src/esp/assets/ResourceManager.cpp
Skylion007, line 1199: BINGO: @bigbike we make guarantees that just be cause an object requires lighting that the scene actually has it: https://github.com/facebookresearch/habitat-sim/blob/77d201a15442e8bb6e68baca5d8552e81c763803/src/esp/assets/ResourceManager.cpp#L1070

Skylion007: Aha, @bigbike figured it out. The Attribute Mangers are built BEFORE the lightSetups are actually changed in the getSimulator initialization so doing it here will always error until you call sim.setupLightKey after Simulator initialization!

Skylion007:
Your check of lightCount is called here (right?): https://github.com/facebookresearch/habitat-sim/blob/7db914d23f462e4451ea4d84be4b39fcd47a1926/src/tests/SimTest.cpp#L73 but the lights aren't initialized until here: https://github.com/facebookresearch/habitat-sim/blob/7db914d23f462e4451ea4d84be4b39fcd47a1926/src/tests/SimTest.cpp#L78 so the lightCount will be 0 until line 78

Skylion007:
^ @mosra seems like PBR may need to support 0 lights due to our light construction paradigm.

src/esp/gfx/PbrDrawable.cpp
bigbike: Not a problem at all. Happy to change it.

Skylion007:
@aclegg3 @bigbike and I agreed that the assertion that lighting > 0 is too strict for our use case.

src/esp/gfx/PbrDrawable.cpp
bigbike: Now, this time, I prefer to encourage it to fail earlier. 

> won't result in a black screen 

I would give them the black screen so they immediately know they forget to setup the lights for example.

>  it'll use reasonable defaults

Again, if the users forget to setup the lights, and it is not black (due to your default reasonable defaults), they may think the shader code has problem because the rendering results do not match what they expect (for example, they expect the object is red, but now it is white).

src/esp/gfx/LightSetup.cpp
```diff
@@ -55,7 +55,9 @@ LightSetup getLightsAtBoxCorners(const Magnum::Range3D& box,
 }
 
 LightSetup getDefaultLights() {
-  return LightSetup{{{1.0, 1.0, 0.0, 0.0}, {0.75, 0.75, 0.75}},
+  // return LightSetup{{{1.0, 1.0, 0.0, 0.0}, {0.75, 0.75, 0.75}},
+  //                  {{-0.5, 0.0, 1.0, 0.0}, {0.4, 0.4, 0.4}}};
+  return LightSetup{{{10.0, 10.0, 0.0, 0.0}, {2, 2, 2}},
```
bigbike, line 60: I will change it back before it is committed in.

src/esp/assets/ResourceManager.cpp
bigbike: changed.

Skylion007 (approved):

src/esp/assets/ResourceManager.cpp
```diff
@@ -82,9 +83,10 @@ constexpr char ResourceManager::DEFAULT_LIGHTING_KEY[];
 constexpr char ResourceManager::DEFAULT_MATERIAL_KEY[];
 constexpr char ResourceManager::WHITE_MATERIAL_KEY[];
 constexpr char ResourceManager::PER_VERTEX_OBJECT_ID_MATERIAL_KEY[];
-ResourceManager::ResourceManager()
-    :
+ResourceManager::ResourceManager(Flags flags)
+    : flags_(flags)
```
jturner65, line 87: As we discussed during standup, the data these flags represent will eventually be managed on a per object/per stage basis via the MetadataMediator.

src/esp/gfx/PbrDrawable.cpp
mosra: I see we have vastly different opinions on this topic, but I'll try one last time -- and if our opinions still diverge after this point, I'd say we stop and "agree to disagree" to save ourselves time :) 

I used to not have any defaults in builtin shaders, but since I have them, user experience improved a lot. Based on my experience with answering 100s of questions in Magnum support channels over the years:

- Having a black / empty render can have way too many reasons (wrong triangle winding, renderer state, depth buffer not cleared, model missing normals, model having flipped normals, camera clipping planes, wrong transformation matrices...) and because the problem space is so big, I'm taking every chance to make the space smaller, because debugging "my object is white but my lights are red" is much easier than "i tried everything but all i see is _still_ black".
- The more complex shaders have a ton of parameters and users jumping into it for the first time will just try to set the parameters they know (like transformation/projection matrices) and see what comes out. If they get something, it'll be a reward nudging them to continue exploring further with colors, lights, textures etc. If they get a black screen, the fun ends there.
- If a shader gets a new parameter (such as the normal map scale that recently arrived in Phong) and it doesn't have a reasonable default, existing code will be broken (normal mapping doing nothing) until someone notices (which might also be never). This is probably the strongest point for reasonable defaults as it applies even when your team is full of expert veterans who never suffered from a black screen render in their lifetime.

src/shaders/pbr.vert
bigbike: If it is not in this PR, it will happen in the following PRs.

src/esp/gfx/PbrShader.cpp
```diff
@@ -362,6 +405,8 @@ PbrShader& PbrShader::setTextureMatrix(const Mn::Matrix3& matrix) {
                  "PbrShader::setTextureMatrix(): the shader was not "
                  "created with texture transformation enabled",
                  *this);
+
+  // since emissive texture may need it, so no if (lightCount_) here
   setUniform(textureMatrixUniform_, matrix);
```
mosra, line 410: This may still result in warnings from GL, so maybe 

```suggestion
  if (flags_ & (Flag::TextureMatrix|Flag::EmissiveTexture))
    setUniform(textureMatrixUniform_, matrix);
```

src/esp/gfx/PbrShader.cpp
```diff
@@ -191,23 +189,25 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
 
   // set texture binding points in the shader;
   // see PBR vertex, fragment shader code for details
-  if (flags_ & Flag::BaseColorTexture) {
+  if ((flags_ & Flag::BaseColorTexture) && lightCount_) {
     setUniform(uniformLocation("BaseColorTexture"), TextureUnit::BaseColor);
   }
-  if (flags_ & Flag::RoughnessTexture) {
+  if ((flags_ & Flag::RoughnessTexture) && lightCount_) {
```
mosra, line 195: I'd maybe put all those ifs in a large `if(lightCount_)` block instead? So it's clearer what goes together. You're doing the same below for the defaults.

mosra (approved):
I'd personally mention the defaults in the PbrShader docs, but that's up to you.
 
The rest are totally minor things :)

Issue #865, opened 2020-10-16
Error in rendering high quality images from Replica dataset


Ainaz99:
## ❓ Questions and Help
Hi, I tried rendering high-quality RGB images from Replica dataset. I used `sorted faces` and pointed to the `mesh.ply` file in the scene. Looks like textures are loading correctly, but there's an error in loading mesh.ply : `ResourceManager::loadStage : Stage mesh.ply Collision mesh load failed. Aborting scene initialization.` Thank you so much for your help.

Here's also the full error.

```
I1016 19:13:42.846678 11286 PTexMeshData.cpp:471] Splitting mesh... 
I1016 19:13:43.435354 11286 PTexMeshData.cpp:381] The number of quads: 1908569, which equals to 3817138 triangles.
I1016 19:13:43.435405 11286 PTexMeshData.cpp:812] Loading mesh 1/12... 
I1016 19:13:43.441056 11286 PTexMeshData.cpp:812] Loading mesh 2/12... 
I1016 19:13:43.444914 11286 PTexMeshData.cpp:812] Loading mesh 3/12... 
I1016 19:13:43.449556 11286 PTexMeshData.cpp:812] Loading mesh 4/12... 
I1016 19:13:43.454962 11286 PTexMeshData.cpp:812] Loading mesh 5/12... 
I1016 19:13:43.456931 11286 PTexMeshData.cpp:812] Loading mesh 6/12... 
I1016 19:13:43.457381 11286 PTexMeshData.cpp:812] Loading mesh 7/12... 
I1016 19:13:43.462954 11286 PTexMeshData.cpp:812] Loading mesh 8/12... 
I1016 19:13:43.465827 11286 PTexMeshData.cpp:812] Loading mesh 9/12... 
I1016 19:13:43.468580 11286 PTexMeshData.cpp:812] Loading mesh 10/12... 
I1016 19:13:43.471900 11286 PTexMeshData.cpp:812] Loading mesh 11/12... 
I1016 19:13:43.476325 11286 PTexMeshData.cpp:812] Loading mesh 12/12... 
I1016 19:13:43.479084 11286 PTexMeshData.cpp:838] Calculating mesh adjacency... 
I1016 19:13:44.575686 11286 PTexMeshData.cpp:885] loading atlas textures: 
I1016 19:13:44.576220 11286 PTexMeshData.cpp:894] Loading atlas 1/12 from textures/0-color-ptex.hdr. 
[...]

PR #866, opened 2020-10-16
[bugfix] Replica Ptex without physics

aclegg3:
## Motivation and Context

Addresses issue #865. PTex collision mesh group creation is not supported, but was being called even when physics was not enabled. This is temporary patch to skip creation of the RigidStage when a stage is loaded and physics is not enabled. 

A better fix would be to support collisionMeshGroups for PTex assets. 

## How Has This Been Tested

Locally with viewer.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -244,7 +244,7 @@ bool ResourceManager::loadStage(
   // the collision mesh is empty scene
 
   if ((_physicsManager != nullptr) &&
-      (infoToUse.filepath.compare(EMPTY_SCENE) != 0)) {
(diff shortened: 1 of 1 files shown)


aclegg3:
@jturner65 This was just a quick-fix if you see a more elegant way to adjust the pipeline here.

bigbike:
Segmentation fault in the CI test now.

erikwijmans:
Looks like navmesh re-computation uses the collision mesh.  We should just populate the collision mesh data in PTexMeshData.  Should be fairly straightforward.  I can do it tomorrow.

aclegg3:
Fix in #867

PR #876, opened 2020-10-20
[Bug fix] Use flags, not materialData in order to set the uniforms, textures properly in the GenericDrawable. 

bigbike:
## Motivation and Context
This is to address a bug mentioned by @eundersander . Here I quote:
"the warning about lacking tangents is essentially setting up the user to inevitably fail this later assert.
https://github.com/facebookresearch/habitat-sim/commit/77d201a15442e8bb6e68baca5d8552e81c763803#diff-e849e837346024527e3aadac345a485c682ccfe8ba595cc2662a3e328ee707bfR44
Because, on the one hand, the model has a normal texture but not tangents, so the shader is not created with the normalmap flag. But then later in the same file, you have:
```
if (materialData_->normalTexture)
```

The correct way is to use `flags_` not the `materialData_` in the if clause.
This PR is to fix it.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/GenericDrawable.cpp
+++ b/src/esp/gfx/GenericDrawable.cpp
@@ -130,17 +130,23 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
       .setProjectionMatrix(camera.projectionMatrix())
       .setNormalMatrix(transformationMatrix.normalMatrix());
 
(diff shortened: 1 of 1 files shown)


eundersander (approved):
LGTM, thanks for quick fix!

Skylion007 (approved):

PR #885, opened 2020-10-21
--MetadataMediator - Scene Instance Configurations

jturner65:
## Motivation and Context
This PR add support for parsing "scene_instance.json" into SceneAttributes configuration files, which are then consumed to instantiate a scene.  

A "scene_instance.json" file, whose format can be found [here](https://drive.google.com/file/d/1_1fw5y93mAE1FuwQZWqgbQKtVQGOx-CA/view?usp=sharing), describes the layout of a scene, potentially specifying a stage and/or 0 or more objects, a lighting configuration, navmesh and semantic scene descriptor.  This configuration file is parsed by SceneAttributesManager into a SceneAttributes configuration object, which manages references to the various scene assets, including individual instances of SceneInstanceAttributes objects which describe which object and stage templates to use, if any, and where to put the instantiated construct within the scene.

## How Has This Been Tested
All c++ and python tests have passed.  Tests were added to verify that the configuration parsing is proceeding as expected.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/managers/SceneAttributesManager.cpp
+++ b/src/esp/metadata/managers/SceneAttributesManager.cpp
@@ -135,9 +135,10 @@ SceneAttributesManager::createInstanceAttributesFromJSON(
   if (io::jsonIntoVal<std::string>(jCell, "motion_type", tmpVal)) {
     // motion type tag was found, perform check - first convert to lowercase
     std::string strToLookFor = Cr::Utility::String::lowercase(tmpVal);
-    if (SceneObjectInstanceAttributes::MotionTypeNamesMap.count(strToLookFor)) {
-      motionTypeVal = static_cast<int>(
-          SceneObjectInstanceAttributes::MotionTypeNamesMap.at(strToLookFor));
+    auto found =
+        SceneObjectInstanceAttributes::MotionTypeNamesMap.find(strToLookFor);
+    if (found != SceneObjectInstanceAttributes::MotionTypeNamesMap.end()) {
+      motionTypeVal = static_cast<int>(found->second);
     } else {
       LOG(WARNING)
           << "SceneAttributesManager::createInstanceAttributesFromJSON : "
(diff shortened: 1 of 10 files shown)


src/esp/metadata/managers/SceneAttributesManager.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  int motionTypeVal = motionTypeVal =
+      static_cast<int>(physics::MotionType::UNDEFINED);
+  std::string tmpVal = "";
+  if (io::jsonIntoVal<std::string>(jCell, "motion_type", tmpVal)) {
+    // motion type tag was found, perform check - first convert to lowercase
+    std::string strToLookFor = Cr::Utility::String::lowercase(tmpVal);
+    if (SceneObjectInstanceAttributes::MotionTypeNamesMap.count(strToLookFor)) {
+      motionTypeVal = static_cast<int>(
+          SceneObjectInstanceAttributes::MotionTypeNamesMap.at(strToLookFor));
```
mosra, line 140: Minor but very common unneeded double lookup (sorry, another personal check triggerred, heh). If you do the following instead, it'll be just a single lookup:

```suggestion
    auto found = SceneObjectInstanceAttributes::MotionTypeNamesMap.find(strToLookFor);
    if (found != SceneObjectInstanceAttributes::MotionTypeNamesMap.end()) {
      motionTypeVal = static_cast<int>(*found);
```

Alternatively, if all you ever need is converting a string to an enum, to make it easier to use you could have for example a `Containers::Optional<MotionType> motionTypeFromString(const std::string&)` (a free function, even?) that does this lookup and then returns either a valid enum or an optional (and could be doing that just through a cascade of ifs, there's not that many values anyway). Then it'd be just

```cpp
    if (Containers::Optional<MotionType> type = motionTypeFromString(strToLookFor)) {
      motionTypeVal = static_cast<int>(*type);
    } else {
```

mosra (commented):
I'm afraid I'm not qualified enough to comment on the rest of the PR besides this one technicality, wasn't involved in any discussion about these APIs :)

src/esp/metadata/managers/SceneAttributesManager.cpp
jturner65: Thanks mosra! This stuff really helps me. :)

aclegg3 (approved):
Looks good to me. 👍

PR #887, opened 2020-10-21
Viewer features and physics bugfix

aclegg3:
## Motivation and Context

This PR packages a few minor fixes for physics simulation and Viewer application features:
- Viewer:
  - `SPACE` to toggle physics simulation on/off
  - `'.'` to take a single simulation step when continuous simulation is toggled off.
  - Object adding returns allocated object IDs
  - Added objects are now set to agent orientation (aligned -Z forward local frames) instead of random orientation.
- [bugfix] Now correctly propagating configured collision margin to BulletRigidStage.

## How Has This Been Tested

Local tests.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletPhysicsManager.cpp
+++ b/src/esp/physics/bullet/BulletPhysicsManager.cpp
@@ -161,11 +161,6 @@ void BulletPhysicsManager::stepPhysics(double dt) {
   int numSubStepsTaken =
       bWorld_->stepSimulation(dt, /*maxSubSteps*/ 10000, fixedTimeStep_);
   worldTime_ += numSubStepsTaken * fixedTimeStep_;
-
-  // manually sync the motionstates for all objects
-  for (auto& objectItr : existingObjects_) {
-    objectItr.second->syncPose(true);
-  }
 }
 
 void BulletPhysicsManager::setMargin(const int physObjectID,
--- a/src/esp/physics/bullet/BulletPhysicsManager.cpp
+++ b/src/esp/physics/bullet/BulletPhysicsManager.cpp
@@ -162,9 +162,12 @@ void BulletPhysicsManager::stepPhysics(double dt) {
       bWorld_->stepSimulation(dt, /*maxSubSteps*/ 10000, fixedTimeStep_);
   worldTime_ += numSubStepsTaken * fixedTimeStep_;
 
-  // manually sync the motionstates for all objects
+  // Manually sync the motionstates for all DYNAMIC objects. KINEMATIC and
+  // STATIC sync should always be one-way.
   for (auto& objectItr : existingObjects_) {
-    objectItr.second->syncPose(true);
+    if (objectItr.second->getMotionType() == MotionType::DYNAMIC) {
+      objectItr.second->syncPose(true);
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -161,6 +161,14 @@ void BulletPhysicsManager::stepPhysics(double dt) {
   int numSubStepsTaken =
       bWorld_->stepSimulation(dt, /*maxSubSteps*/ 10000, fixedTimeStep_);
   worldTime_ += numSubStepsTaken * fixedTimeStep_;
+
+  // Manually sync the motionstates for all DYNAMIC objects. KINEMATIC and
+  // STATIC sync should always be one-way.
+  for (auto& objectItr : existingObjects_) {
```
eundersander, line 167: Isn't Bullet supposed to do this automatically through the motion state interface we give it? Have you investigated why this isn't happening correctly in Bullet? This feels like a band-aid.

aclegg3: Fair point. Digging deeper I did find another option. I'll open another PR to iterate on this specific issue.

src/utils/viewer/viewer.cpp
```diff
@@ -88,14 +88,14 @@ class Viewer : public Mn::Platform::Application {
   void keyPressEvent(KeyEvent& event) override;
 
   // Interactive functions
-  void addObject(const std::string& configHandle);
-  void addObject(int objID);
+  int addObject(const std::string& configHandle);
+  int addObject(int objID);
 
   // add template-derived object
```
bigbike, line 94: consider specifying the return value in the comments.

src/utils/viewer/viewer.cpp
```diff
@@ -382,42 +391,46 @@ Viewer::Viewer(const Arguments& arguments)
...
+int Viewer::addObject(const std::string& objectAttrHandle) {
   // Relative to agent bodynode
   Mn::Matrix4 T = agentBodyNode_->MagnumObject::transformationMatrix();
   Mn::Vector3 new_pos = T.transformPoint({0.1f, 1.5f, -2.0f});
 
-  int physObjectID = simulator_->addObjectByHandle(configFile);
+  int physObjectID = simulator_->addObjectByHandle(objectAttrHandle);
   simulator_->setTranslation(new_pos, physObjectID);
-  simulator_->setRotation(esp::core::randomRotation(), physObjectID);
+  simulator_->setRotation(Mn::Quaternion::fromMatrix(T.rotationNormalized()),
```
bigbike, line 407: > set to agent orientation

I would like to learn a bit more about this design choice. Why not set it to align the stage orientation?

aclegg3: We could do that also. Just aiming for something consistent here. 

My reasoning: We are setting position from agent transform, so do the same for orientation. This way you always know that the object's local forward is aligned with the camera forward when it is added. 

Another choice: we could apply no transformation to added objects by default and provide another function to set an object transform (with orientation optional) relative to the agent as a follow-up operation. Do you have a preference?

jturner65 (commented):
LGTM - if it passes is ships! :)

jturner65 (approved):

PR #893, opened 2020-10-23
Update pybind to v2.6.0

Skylion007:
## Motivation and Context
* This allows us to support Python 3.9 and also has some good additional code correctness checks that are pretty useful. Also PyBind now properly supports CMAKE level LTO hints are other features.

It did also require a minor refactor of magnum-bindings to support it. Thanks so much @mosra for the help there.

- [X] Dependency upgrade

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.h
+++ b/src/esp/assets/ResourceManager.h
@@ -239,6 +239,15 @@ class ResourceManager {
     return metadataMediator_->getAssetAttributesManager();
   }
   /**
+   * @brief Return manager for construction and access to light and lighting
+   * layout attributes.
+   */
+  const metadata::managers::LightLayoutAttributesManager::ptr
+  getLightLayoutAttributesManager() const {
+    return metadataMediator_->getLightLayoutAttributesManager();
+  }
+
+  /**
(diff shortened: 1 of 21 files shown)


mosra:
I was once again amazed at how *slow* the build of pybind11 bindings was. The resulting binaries were slightly smaller with 2.6, but the build time at least *feels* slower to me. Did you experience the same thing or it's just my feeling?

Skylion007:
@mosra Yeah PyBind11 can be really slow to compile. @mosra one thing that could have slowed down build time is that a bunch of pybind specific options were changed to respect CMAKE defaults (like respecting the CMAKE `INTERPROCEDURAL_OPTIMIZATION` and CMAKE_CXX_STANDARD properties).

I didn't experience it locally because I was running an IO Bound training job on my machine so the file system was already super slow. :P We can check the circleci build times if its a concern.

mosra (approved):

mosra:
(Should have probably noted that I didn't see any regressions in any of the `magnum-bindings` tests, so I assume it'll be probably a painless upgrade here as well.)

PR #897, opened 2020-10-26
--MetadataMediator : Tests and bindings.

jturner65:
## Motivation and Context
This PR adds a c++ test to verify MetadataMediator functionality.  It also makes Sim.reconfigure maintain an appropriate current ref to the current MetadataMediator in ResourceManager, and provides some accessors.

## How Has This Been Tested
Passes all existing c++ and python tests; Adds tests

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -83,9 +83,11 @@ void initSimBindings(py::module& m) {
       .def_property("frustum_culling", &Simulator::isFrustumCullingEnabled,
                     &Simulator::setFrustumCullingEnabled,
                     R"(Enable or disable the frustum culling)")
-      .def_property("active_dataset", &Simulator::getActiveDatasetName,
-                    &Simulator::setActiveDatasetName,
-                    R"(The currently active dataset being used.)")
+      .def_property(
+          "active_dataset", &Simulator::getActiveSceneDatasetName,
+          &Simulator::setActiveSceneDatasetName,
+          R"(The currently active dataset being used.  Will attempt to load
+            configuration files specified if does not already exist.)")
       /* --- Physics functions --- */
       /* --- Template Manager accessors --- */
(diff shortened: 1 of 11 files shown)


Skylion007 (approved):

src/esp/bindings/SimBindings.cpp
```diff
@@ -83,6 +83,9 @@ void initSimBindings(py::module& m) {
       .def_property("frustum_culling", &Simulator::isFrustumCullingEnabled,
                     &Simulator::setFrustumCullingEnabled,
                     R"(Enable or disable the frustum culling)")
+      .def_property("active_dataset", &Simulator::getActiveDatasetName,
+                    &Simulator::setActiveDatasetName,
+                    R"(The currently active dataset being used.)")
```
aclegg3, line 88: Maybe add a note that setting this will trigger a load of all linked configuration files if a new dataset is requested? Already specified in C++ docs, but again in python API would probably be helpful.

src/esp/sim/Simulator.h
```diff
@@ -127,6 +127,22 @@ class Simulator {
     return metadataMediator_->getStageAttributesManager();
   }
 
+  /**
+   * @brief Get current active dataset name from @ref metadataMediator_.
+   */
+  std::string getActiveDatasetName() {
```
aclegg3, line 133: Should these always be "SceneDataset" to reduce confusion?

aclegg3 (approved):
Looks good. Minor comments on doc-strings and naming. Do we want to always refer to Habitat-sim `SceneDatasetAttributes` as `SceneDatasets` when setting/querying? Should make it less ambiguous when the API is called from Habitat-lab.

src/esp/sim/Simulator.h
Skylion007: Exactly!

PR #898, opened 2020-10-27
Simply the Pbr Shading framework based on glTF 2.0

bigbike:
## Motivation and Context
I revisited the glTF 2.0 Specification found here:
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-pbrmetallicroughness

The code can be simplified and optimized accordingly.
- Eliminate the independent (roughness, metallic) textures since it is always a packed texture in the glTF 2.0.
- Correctly handle the case where either roughness or metallic texture exist, not both.

## Types of changes

## Checklist


src/esp/gfx/PbrDrawable.cpp
```diff
@@ -24,20 +24,21 @@ PbrDrawable::PbrDrawable(scene::SceneNode& node,
...
       materialData_{
           shaderManager.get<MaterialData, PbrMaterialData>(materialDataKey)} {
-  flags_ = PbrShader::Flag::ObjectId;
+  if (materialData_->metallicTexture && materialData_->roughnessTexture) {
+    CORRADE_ASSERT(
+        materialData_->metallicTexture == materialData_->roughnessTexture,
+        "PbrDrawable::PbrDrawable(): if both the metallic and roughness "
+        "texture exist, they must be packed in the same texture based on glTF "
+        "2.0 Spec.", );
+  }
```
mosra, line 33: This check is not enough, especially in case the original material wouldn't have the maps in the expected channels. If you'd use Magnum's material APIs directly, it would be just

```suggestion
CORRADE_ASSERT(materialData_->hasNoneRoughnessMetallicTexture(), "...", );
```

Alternatively you could put this assert directly into the `ResourceManager`, because there you still have access to the `Trade::PbrMetallicRoughnessMaterialData` instance.

src/esp/gfx/PbrShader.cpp
```diff
@@ -43,39 +43,14 @@ namespace gfx {
...
-  // 1          : NoneRoughnessMetallicTexture
-  // 2          : RoughnessTexture, MetallicTexture
-  if (result & Flag::OcclusionRoughnessMetallicTexture) {
-    result &= ~Flag::NoneRoughnessMetallicTexture;
-    result &= ~Flag::RoughnessTexture;
-    result &= ~Flag::MetallicTexture;
-  } else if (result & Flag::NoneRoughnessMetallicTexture) {
-    result &= ~Flag::RoughnessTexture;
-    result &= ~Flag::MetallicTexture;
-  }
```
mosra, line 72: Just in case you forgot all our previous private or public discussions -- the MaterialData APIs in Magnum were not built in a generic way because I like to be an architecture astronaut designing overly generic APIs or because I did not read the glTF spec carefully enough. It's because, even if we ignore other file formats, glTF allows for much more flexibility beyond what the core spec says:

- [MSFT_packing_occlusionRoughnessMetallic](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/MSFT_packing_occlusionRoughnessMetallic/README.md) (three new packing modes)
- [MSFT_packing_normalRoughnessMetallic](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/MSFT_packing_normalRoughnessMetallic/README.md) (another packing variant)
- And [several other proposals](https://github.com/KhronosGroup/glTF/issues/1456) appearing, mainly useful for furniture models where you often have tiled textures where all maps don't necessarily have the same transformation.

I mean, there's nothing wrong with supporting just the core glTF files and failing import for everything else, but why do that **now**, after we both spent _so much time_ discussing and implementing all the generic code you just removed?

mosra (commented):
If I remember correctly, after a long discussion both privately on Slack and on the original PRs, we agreed on doing the PBR material import in a way that works not with just core glTF 2.0, but also various texture packing extensions (which then also open the door for potential other file formats such as FBX or USD, as those may or may not use the same packing). 

At first I wanted to save your time and suggest you do just the minimal viable implementation that works with just metallic/roughness packed in a single texture and ignore the rest "for now", but you opted to implement it in a generic way. 

Good then ... but why are you now removing the generic way again, now that we both spent a considerable amount of time discussing and implementing it?

Skylion007:
>> Eliminate the independent (roughness, metallic) textures since it is always a packed texture in the glTF 2.0. 

Pretty sure there are extensions that allow it to be unpacked, I've definitely seen code to handle them.

bigbike:
> > > Eliminate the independent (roughness, metallic) textures since it is always a packed texture in the glTF 2.0.
> 
> Pretty sure there are extensions that allow it to be unpacked, I've definitely seen code to handle them.

@Skylion007 : If you have it, I would like to learn it. Within glTF 2.0 Spec (https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-pbrmetallicroughness)
I do not think e.g., roughness and metallic can be specified in separated files.

src/esp/gfx/PbrShader.cpp
bigbike: - [MSFT_packing_occlusionRoughnessMetallic](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/MSFT_packing_occlusionRoughnessMetallic/README.md) (three new packing modes)
Not a problem with this PR. The occlusion will be in the `MetallicRoughness` texture. I already considered this situation see the `Flag` in the `PbrShader`. I have `PackedOcclusionTexture` to handle this case and `SeparateOcclusionTexture` to handle the other one.
- [MSFT_packing_normalRoughnessMetallic](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/MSFT_packing_normalRoughnessMetallic/README.md) (another packing variant)
Good to know. Even without this PR, the current code cannot handle this extension. So anything regarding this extension is totally a different topic, has nothing to do with the changes of this PR.

- [several other proposals](https://github.com/KhronosGroup/glTF/issues/1456)
This is a proposal from 2018 that is not finished yet. This proposal aims to apply the transforms per texture (per channel, because roughness, metallic are in different channels within the same image), not per image. Again, this is a totally separate topic, has nothing to do with this PR, has no effects on current glTF Spec yet (and may not happen in near future.)

> I mean, there's nothing wrong with supporting just the core glTF files and failing import for everything else.

To clarify, as long as we are talking about glTF, including the extensions you mentioned, this PR will NOT fail except that [MSFT_packing_normalRoughnessMetallic](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/MSFT_packing_normalRoughnessMetallic/README.md) (another packing variant), which cannot be handled by the original code before this PR as well.

bigbike:
> open the door for potential other file formats such as FBX or USD, as those may or may not use the same packing

Makes sense. This would be the only reason if I drop this PR later.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1369,25 +1362,10 @@ gfx::PbrMaterialData::uptr ResourceManager::buildPbrShadedMaterialData(
     finalMaterial->metallicTexture =
         textures_[textureBaseIndex + material.metalnessTexture()].get();
     if (!material.hasAttribute(Mn::Trade::MaterialAttribute::Metalness)) {
-      finalMaterial->metallic = 1.0f;
+      finalMaterial->metallic = 1.0f;  // based on GlTF 2.0 Spec
     }
   }
```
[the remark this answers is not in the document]
bigbike: To clarify:
I did not "invent" this `PbrMaterialData`. I just followed the `PhongMaterialData` which was there before, and created it. 
I am not against refactoring the code so that we can migrate the code to adopt the Magnum built-in `Trade::MaterialData`.

bigbike:
The purpose of this PR is to simplify and thus optimize the code, removing some burdens from the developer's shoulders in the future. That is it.
None of the extensions mentioned the reviewers' comments is a deal breaker of this PR.

bigbike:
Did a little investigation on FBX format:

Based on [Material mapping for model formats](https://docs.microsoft.com/en-us/azure/remote-rendering/reference/material-mapping#:~:text=The%20FBX%20format%20is%20closed,the%20Azure%20Remote%20Rendering%20pipeline.)
Here I quote:
"The FBX format is closed-source and FBX materials are not compatible with PBR materials in general. FBX uses a complex description of surfaces with many unique parameters and properties..."

Skylion007:
> > > > Eliminate the independent (roughness, metallic) textures since it is always a packed texture in the glTF 2.0.
> > 
> > 
> > Pretty sure there are extensions that allow it to be unpacked, I've definitely seen code to handle them.
> 
> @Skylion007 : If you have it, I would like to learn it. Within glTF 2.0 Spec (https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-pbrmetallicroughness)
> I do not think e.g., roughness and metallic can be specified in separated files.

Ah, nvm. Was confusing it with another extension.

src/esp/gfx/PbrDrawable.cpp
bigbike: This sanity check is to only guarantee that when both metallic and roughness texture are specified, they are pointing to the same image. It has to be as long as it was from the glTF file. That is it.

For even stronger check as you suggested, yes, we may put it in the ResourceManager.

bigbike:
Dear reviewers,

Would you please take a look at this PR. See if any concerns need to be resolved before committing it in.

Thanks.

src/esp/gfx/PbrShader.cpp
mosra: (Marking as resolved based on the previous discussions, either private or via the weekly.)

mosra:
To sum up -- 3/4 of my last review are still relevant I think, I marked the remaining comment as resolved:

- This assert on ensuring the right packing is used: https://github.com/facebookresearch/habitat-sim/pull/898#discussion_r512604784
- Default metalness to `0.0f` instead of something in-between: https://github.com/facebookresearch/habitat-sim/pull/898#discussion_r512585306
- Simplify the attribute-presence-checking blocks based on the comment above: https://github.com/facebookresearch/habitat-sim/pull/898#discussion_r512584820

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
bigbike: This is DONE.

src/esp/gfx/MaterialData.h
```diff
@@ -58,7 +58,7 @@ struct PbrMaterialData : public MaterialData {
   // roughness texture) are NOT presented, use these default magic values;
   Magnum::Color4 baseColor{0.7f};
   Magnum::Float roughness = 0.9f;
-  Magnum::Float metallic = 0.1f;
+  Magnum::Float metallic = 0.01f;
```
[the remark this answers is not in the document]
bigbike: Both are using 1.0f as the default value based on glTF 2.0.

src/esp/gfx/PbrDrawable.cpp
bigbike: DONE. Added to the ResourceManager.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1577,55 +1574,36 @@ gfx::PbrMaterialData::uptr ResourceManager::buildPbrShadedMaterialData(
...
-
-  if (material.hasNoneRoughnessMetallicTexture()) {
-    // roughnessTexture, metalnessTexture are pointing to the
-    // same texture ID, so just roughnessTexture
-    finalMaterial->noneRoughnessMetallicTexture =
-        textures_.at(textureBaseIndex + material.roughnessTexture()).get();
+  // sanity check when both metallic and roughness materials are presented
+  if (material.hasMetalnessTexture() && material.hasRoughnessTexture()) {
+    CORRADE_ASSERT(
+        finalMaterial->metallicTexture == finalMaterial->roughnessTexture,
```
bigbike, line 1598: @mosra: I still use: `finalMaterial->metallicTexture == finalMaterial->roughnessTexture` instead of 
`material.hasNoneRoughnessMetallicTexture() | material.hasOcclusionRoughnessMetallicTexture() | material.hasNormalRoughnessMetallicTexture()` for code simplicity.

bigbike:
@mosra: Would you take a look when you have time in Jan.? Thanks.

src/esp/gfx/MaterialData.h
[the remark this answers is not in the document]
bigbike: By the way, the previous "magic" numbers were set like this simply because they made the debugging easier.

src/esp/assets/ResourceManager.cpp
mosra: Code simplicity? No, the main issue with comparing just IDs is that it's *not enough*. It doesn't check that:

- both use the same texture coordinate attribute,
- both have the same texture transformation, and
- the metalness is in B and roughness is in G

While the 1st and 3rd case are rather unlikely with models that are currently circulating around the web, the second case sometimes happens, especially with models that were processed with meshoptimizer / gltfpack. So *please* can you use `hasNoneRoughnessMetallicTexture()` to have *everything* checked? Thanks.

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
mosra: Come on, it's not :sweat_smile: You need to use the `hasAttribute()` / `has*Texture()` only for textures, because there's no "default texture ID" that could be used. For everything else you can just use the `metalness()` etc. getter. Let me repeat once again -- instead of:

```cpp
  if (material.hasAttribute(Mn::Trade::MaterialAttribute::Metalness)) {
    finalMaterial->metallic = material.metalness();
  }
```

it can be just:

```cpp
  finalMaterial->metallic = material.metalness();
```

And the same for, as I said above:

- base color
- normal texture scale
- emissive color
- roughness

While we're deleting code, let's do that completely :wink:

mosra:
Done,  two unresolved discussions left:

- code that could use some further simplification: https://github.com/facebookresearch/habitat-sim/pull/898#discussion_r549819208
- code that's simple a bit too much: https://github.com/facebookresearch/habitat-sim/pull/898#discussion_r549817578

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
bigbike: OK. Make sense. Will do.

src/esp/assets/ResourceManager.cpp
bigbike: I see. So it means in the `hasNoneRoughnessMetallicTexture`, it actually does more sanity checks than it literally suggests. I do not know this before.

Question: other than `hasNoneRoughnessMetallicTexture`, do I need to also check:

`material.hasNoneRoughnessMetallicTexture() | material.hasOcclusionRoughnessMetallicTexture() | material.hasNormalRoughnessMetallicTexture()`?

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
bigbike: DONE.

mosra (approved):
:tada:

src/esp/assets/ResourceManager.cpp
```diff
@@ -1577,56 +1568,55 @@ gfx::PbrMaterialData::uptr ResourceManager::buildPbrShadedMaterialData(
...
+        material.hasNoneRoughnessMetallicTexture(),
+        "ResourceManager::buildPbrShadedMaterialData(): if both the metallic "
+        "and roughness texture exist, they must be packed in the same texture "
+        "based on glTF 2.0 Spec.",
+        finalMaterial);
   }
 
+  // TODO:
+  // Support NormalRoughnessMetallicTexture packing
+  CORRADE_ASSERT(!material.hasNormalRoughnessMetallicTexture(),
```
bigbike, line 1611: @mosra : FYI: I just added a corrade assertion here after you approved this diff.

mosra: I just saw :) `hasNormalRoughnessMetallicTexture()` and `hasNoneRoughnessMetallicTexture()` are mutually exclusive, never both `true`, so *technically* this assert is unnecessary, however one extra check won't hurt I think.

bigbike: Oh, OK. Makes sense.

I would recommend you to put this sentence:
>hasNormalRoughnessMetallicTexture() and hasNoneRoughnessMetallicTexture() are mutually exclusive, never both true

to your wiki even it becomes so obvious after thinking about it for another second.

PR #906, opened 2020-10-28
--[WIP]Remove SceneConfiguration.cpp

jturner65:
## Motivation and Context
SceneConfiguration.cpp is now redundant, obviated by the MetadataMediator and SceneDataset configuration handling.   Instead of a SimulatorConfiguration object holding a SceneConfiguration object, it now has a string, activeSceneID, to represent the handle of the active scene or stage.  This has been changed in the python and js bindings to be "scene_id" from "scene.id".

**This is a breaking change**  If you reference a SceneConfiguration object, this is no longer supported.  If you wish to specify the name of a scene in your SimulatorConfiguration you must modify your python code like so : 

From : 
`cfg.scene.id = "data/scene_datasets/habitat-test-scenes/van-gogh-room.glb"`

To:
`cfg.scene_id = "data/scene_datasets/habitat-test-scenes/van-gogh-room.glb"`

## How Has This Been Tested
C++ and python tests passed

## Types of changes

## Checklist


jturner65:
Note : This change also impacts Hab-Lab, and will only pass the hab-lab tests when[ the requisite changes are merged to hab-lab master.](https://github.com/facebookresearch/habitat-lab/pull/516)

src/esp/bindings_js/bindings_js.cpp
```diff
@@ -153,20 +153,10 @@ EMSCRIPTEN_BINDINGS(habitat_sim_bindings_js) {
...
-      .property("id", &SceneConfiguration::id)
-      .property("filepaths", &SceneConfiguration::filepaths)
-      .property("sceneUpDir", &SceneConfiguration::sceneUpDir)
-      .property("sceneFrontDir", &SceneConfiguration::sceneFrontDir)
-      .property("sceneScaleUnit", &SceneConfiguration::sceneScaleUnit);
-
   em::class_<SimulatorConfiguration>("SimulatorConfiguration")
       .smart_ptr_constructor("SimulatorConfiguration",
                              &SimulatorConfiguration::create<>)
-      .property("scene", &SimulatorConfiguration::scene)
```
Skylion007, line 169: Can't we not add a dummy property here as a deprecated shim?

jturner65: I could, but I wanted things to break so that folks could find and fix any problems. Besides, it would have to be an object, because code refs to this follow the form scene.id so just having some kind of dummy here wouldn't work easily.

tests/test_simulator.py
```diff
@@ -15,7 +15,7 @@ def test_no_navmesh_smoke():
     # with no navmesh and the navmesh isn't used for any exisitng sensors
     agent_config.sensor_specifications = []
 
-    sim_cfg.scene_id = "data/scene_datasets/habitat-test-scenes/stage_floor1.glb"
+    sim_cfg.scene_id = "data/test_assets/scenes/stage_floor1.glb"
```
Skylion007, line 18: Can we have assert in simulator.py that the scene_id exists?

jturner65: Something along these lines obviously need to be explored as scene import is constructed.

src/esp/bindings_js/bindings_js.cpp
Skylion007: Just to revisit, this could temporarily unblock a lot of the pain you are having right now with the python bindings.

jturner65: I do not see how.  Originally scene represented the struct (SceneConfiguration) that this PR is removing, which had as a member the string "id", while now it is replaced with "scene_id" which represents a string. Assignment originally was made to scene.id.  Not sure how i could dummy this in any acceptable way.

erikwijmans: Python is wild: https://github.com/facebookresearch/habitat-sim/compare/Remove_SceneConfiguration-backwards-compat?expand=1#diff-08c325c91ce6aa20a6124d796c72d68934251a35f1cbc6e24421e5c23617de69

If we do want to add a backwards compatibility layer here, we should do it after we have verified that our code doesn't need it tho.

jturner65: Thanks Erik.  I like that deprecation protocol.  If I'm still having problems with this change against hab-lab, I'll make use of it.

Skylion007 (approved):

Skylion007:
Approval if you can get tests to pass.

tests/test_simulator.py
Skylion007: This really feels like an important blocker because segfaulting when someone makes a typo is not desirable behavior.

Skylion007:
Also please pull from master so you can make sure it passes the latest clang-tidy checks.

aclegg3 (approved):
👍  Tests pass. Let's ship it after reverting ci config changes.

PR #801, opened 2020-09-15
[bugfix] Activate island when removing sleeping objects

aclegg3:
## Motivation and Context

An issue was identified where removing a sleeping object supporting other objects would not activate the supported objects, causing them to float instead of falling as expected. This PR fixes the issue by activating a sleeping object's collision island upon removal.

## How Has This Been Tested

Added `TestRemoveSleepingSupport ` to `PhysicsTest.cpp`.

Testing in viewer:
Green outline denotes sleeping object.
**Before**:
stacking_before

**After**:
stacking

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #913, opened 2020-11-04
[bugfix] Remove STATIC Support

aclegg3:
## Motivation and Context

Previously PR #801 fixed an issue where removal of sleeping objects did not wake nearby objects causing simulation artifacts. This was done by iteratively waking all objects which were part of the same simulation island as the sleeping object (`activateCollisionIsland`). 

However, `STATIC` btRigidBodies do not share a simulation island tag with non-static objects which they are in contact with, causing this previous fix to miss the case that the supporting object is STATIC, not sleeping.

This PR adds a step to the `activateCollisionIsland` which checks the previously computed overlapping pairs to collect all relevant simulation islands to activate.

## How Has This Been Tested

Adjusted the existing test to catch this case.

Viewer testing: bottom object is STATIC. Green denotes STATIC or sleeping. First remove a sleeping DYNAMIC object, then remove the bottom STATIC support.

**Before:**
support_bugfix_before

**After:**
support_bugfix_after

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletRigidObject.cpp
+++ b/src/esp/physics/bullet/BulletRigidObject.cpp
@@ -381,12 +381,34 @@ void BulletRigidObject::constructAndAddRigidBody(MotionType mt) {
 }
 
 void BulletRigidObject::activateCollisionIsland() {
+  btCollisionObject* thisColObj = bObjectRigidBody_.get();
+
+  // first query overlapping pairs of the current object from the most recent
+  // broadphase to collect relevant simulation islands.
+  std::set<int> overlappingSimulationIslands = {thisColObj->getIslandTag()};
+  auto& pairCache =
+      bWorld_->getCollisionWorld()->getPairCache()->getOverlappingPairArray();
+
+  for (int i = 0; i < pairCache.size(); ++i) {
+    if (pairCache.at(i).m_pProxy0->m_clientObject == thisColObj) {
+      overlappingSimulationIslands.insert(
+          static_cast<btCollisionObject*>(
+              pairCache.at(i).m_pProxy1->m_clientObject)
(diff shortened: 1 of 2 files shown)


Skylion007 (approved):

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -381,12 +381,34 @@ void BulletRigidObject::constructAndAddRigidBody(MotionType mt) {
 }
 
 void BulletRigidObject::activateCollisionIsland() {
+  btCollisionObject* thisColObj = bObjectRigidBody_.get();
+
+  // first query overlapping pairs of the current object from the most recent
+  // broadphase to collect relevant simulation islands.
+  std::set<int> overlappingSimulationIslands = {thisColObj->getIslandTag()};
```
eundersander, line 388: Nit: Really just food for thought: in Cruise's C++ codebase as well as in the game industry, I'm used to seeing an effort to avoid frequent, per-frame dynamic allocations, such as this local set container which must grow dynamically.

Unfortunately I've never seen a really great alternative. Sometimes a codebase might declare this container as static and just clear the size to zero every frame, reusing the memory, but this is error-prone and non multithreading-safe. Sometimes a codebase offers a special heap allocator which is very fast but mostly requires FILO (stack-ordered) allocs and frees.

eundersander (approved):
LGTM. I left a comment about memory allocation.

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: Hopefully this won't be per-frame, but I get what you are saying. 👍

PR #916, opened 2020-11-04
Enables more clang-tidy fixes and fixes more warning

Skylion007:
## Motivation and Context

* It improves the state of code and helps make sure all unused variables are purposely unused.
* Adds a few warnings that should could help stop future bugs.
* Adds some uncontroversial modernize linters that didn't change our codebase.

## How Has This Been Tested

Locally

## Types of changes

## Checklist


src/esp/scene/Mp3dSemanticScene.cpp
```diff
@@ -73,12 +75,13 @@ std::string Mp3dObjectCategory::name(const std::string& mapping) const {
   }
 }
 
-int Mp3dRegionCategory::index(const std::string& mapping) const {
+int Mp3dRegionCategory::index(CORRADE_UNUSED const std::string& mapping) const {
```
jturner65, line 78: Not sure if this matters or not, but we usually put this macro in the header.

Skylion007: @mosra Does it matter?

jturner65 (approved):
With possible exception of macro location (header vs. cpp? We usually use header, not sure if it matters), LGTM.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -98,9 +98,11 @@ void initGfxBindings(py::module& m) {
...
              return &self;
            })
-      .def("__exit__", [](RenderTarget& self, const py::object& exc_type,
-                          const py::object& exc_value,
-                          const py::object& traceback) { self.renderExit(); })
+      .def(
+          "__exit__",
+          [](RenderTarget& self, CORRADE_UNUSED const py::object& exc_type,
+             CORRADE_UNUSED const py::object& exc_value,
+             CORRADE_UNUSED const py::object& traceback) { self.renderExit(); })
```
mosra, line 105: I'd just remove the names of the variables instead

```cpp
          [](RenderTarget& self, const py::object&,
             const py::object&,
             const py::object&) { self.renderExit(); })
```

that should make clang-tidy happy as well

src/esp/metadata/managers/LightLayoutAttributesManager.cpp
```diff
@@ -140,7 +140,7 @@ void LightLayoutAttributesManager::setLightInstanceValsFromJSONDoc(
 
 LightLayoutAttributes::ptr LightLayoutAttributesManager::initNewObjectInternal(
     const std::string& handleName,
-    bool builtFromConfig) {
+    CORRADE_UNUSED bool builtFromConfig) {
```
mosra, line 143: Same here and in other `cpp` files and as @jturner65 says, just drop the name altogether

src/esp/scene/Mp3dSemanticScene.cpp
mosra: Should be just 

```cpp
int Mp3dRegionCategory::index(const std::string&) const {
```

in all cpp files, and `CORRADE_UNUSED` in headers (so the parameter name is kept for documentation purposes)

mosra (changes requested):

Skylion007:
All reviewer comments have been merged in, I will merge it in if it builds.

src/esp/geo/CoordinateFrame.cpp
```diff
@@ -4,16 +4,18 @@
...
 namespace geo {
 
-CoordinateFrame::CoordinateFrame(const vec3f& up /* = ESP_UP */,
-                                 const vec3f& front /* = ESP_FRONT */,
-                                 const vec3f& origin /* = vec3f(0, 0, 0) */)
-    : up_(up), front_(front), origin_(origin) {
+CoordinateFrame::CoordinateFrame(vec3f up /* = ESP_UP */,
+                                 vec3f front /* = ESP_FRONT */,
+                                 vec3f origin /* = vec3f(0, 0, 0) */)
+    : up_(std::move(up)), front_(std::move(front)), origin_(std::move(origin)) {
```
mosra, line 18: Not sure about this change. The `vec3f` is (or should be, but it's Eigen so I can't tell) a trivially copyable type so `std::move()` is useless. And we're commonly passing vectors by `const&` everywhere so this causes inconsistency.

What's the clang-tidy warning that triggered this?

Skylion007: https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html

PR #918, opened 2020-11-04
[WIP] MotionType::VISUALIZATION_ONLY

aclegg3:
## Motivation and Context

This PR adds the `VISUALIZATION_ONLY` `MotionType` allowing the Habitat Simulator object API to be leverage to create and manipulate objects which can be seen in visual sensors but do not impact physics simulation or collision checking/response.

Added support for "is_visible" and "is_collidable" tags in object configuration files.

Follow-up work:
- Add an API for programmatically adding/removing the render asset (ie., modifying "is_visible").

## How Has This Been Tested

Added tests for `VISUALIZATION_ONLY` object behavior when checking contacts and raycasting.

Possible additional testing I didn't think was necessary:
- duplicate remove_object() tests for sleeping islands. Same method used as when removing an object, so nothing really new to test here.
- test switching to/from various other MotionTypes and expected simulation behaviors.We re-create the `btRigidBody` in these case by the same method as switching between other types, so nothing new to test.

## Types of changes

## Checklist


docs/pages/attributesJSON.rst
```diff
@@ -177,6 +174,12 @@ Below are object-specific physical quantities.  These values will override simil
 "requires_lighting"
 	- boolean
 	- Whether the object should be rendered with lighting or flat shading.
+"is_visible"
+	- boolean
+	- Whether the object's render asset should be created upon instancing.
+"is_collidable"
```
eundersander, line 180: If it's not collidable, does that also imply no dynamics? Not affected by gravity?

docs/pages/attributesJSON.rst
```diff
@@ -177,6 +174,12 @@ Below are object-specific physical quantities.  These values will override simil
...
+	- boolean
+	- Whether the object should be added to the collision and physics simulation world upon instancing. If false, object will be initialized with MotionType::VISUALIZATION_ONLY.
 "mass"
```
eundersander, line 183: It would be nice if there was some clean indication of what I need to specify or not for a VISUALIZATION_ONLY object. Do I need to choose a mass? A COM? margin? etc. It would be nice if these were all lumped into a "dynamics" section that could simply be omitted entirely for VISUALIZATION_ONLY objects.

src/esp/bindings/PhysicsBindings.cpp
```diff
@@ -20,7 +20,8 @@ void initPhysicsBindings(py::module& m) {
       .value("UNDEFINED", MotionType::UNDEFINED)
       .value("STATIC", MotionType::STATIC)
       .value("KINEMATIC", MotionType::KINEMATIC)
-      .value("DYNAMIC", MotionType::DYNAMIC);
+      .value("DYNAMIC", MotionType::DYNAMIC)
+      .value("VISUALIZATION_ONLY", MotionType::VISUALIZATION_ONLY);
```
eundersander, line 24: Can we rename MotionType? It feels like we're abusing this now. VISUALIZATION_ONLY doesn't really refer to how this thing moves. It moves exactly the same as KINEMATIC, true?

src/esp/physics/RigidBase.h
```diff
@@ -63,7 +63,13 @@ enum class MotionType {
...
    * @ref BulletRigidObject.
    */
-  DYNAMIC
+  DYNAMIC,
+
+  /**
+   * The object is not simulated or active in the collision world, but can be
+   * updated kinematically.
+   */
+  VISUALIZATION_ONLY
```
eundersander, line 72: Not clear from description how this is different from KINEMATIC.

eundersander (commented):
This looks reasonable, except for naming. PhysicsManager, MotionType, BulletRigidObject... these names are really conveying the wrong thing at this point. We're going to confuse Habitat users. We're even going to confuse ourselves talking about these objects in meetings. ("A BulletRigidObject doesn't necessarily have an attached Bullet rigid body???").

If you need to rush this PR in, I'm not opposed to merging it as-is, but let's create a Task to track the tech debt here: cleaning up the naming.

src/esp/physics/RigidBase.h
eundersander, line 72: At some point, we're going to add the capability for users to configure what collides with what. I'm talking about collision filtering. For example, we already configure different parts of the fetch to either collide with the static environment or not: https://github.com/ASzot/p-viz-plan/pull/2 .

Let's fast-forward to the day in the future where we have this feature. How is a VISUALIZATION_ONLY object different from a KINEMATIC object that has been configured not to collide with anything? Is this separate type even needed?

docs/pages/attributesJSON.rst
aclegg3: Thanks for the comment here. I think re-naming this field to something more semantically accurate makes sense. Maybe "is_physical", or even "is_visualization_only".

docs/pages/attributesJSON.rst
aclegg3: Good point. Yeah, you can specify any or none of these. Since a VISUALIZATION_ONLY is a MotionType, these objects can be enabled for physics simulation during runtime, so these properties could still be entirely relevant depending on the user's intentions.

src/esp/bindings/PhysicsBindings.cpp
aclegg3: True, we could re-name the type to KINEMATIC_VISUALIZATION_ONLY or we could re-name MotionType to something else. Thoughts?

aclegg3:
> "A BulletRigidObject doesn't necessarily have an attached Bullet rigid body???".
Well in Bullet simulation mode, all objects have a btRigidBody, it just may not be active in the world. 

Regarding PhysicsManager, @jturner65 has been making similar comments for awhile. Some kind of intermediate "ObjectManager" may be called for with a subset on kinematic functionalities. Open to discussing this.

jturner65 (commented):
I think this kind of behavior should be governed by a flag toggling collisions instead of a motion type (i.e. the existing is_collidable flag).  From a user's perspective, I can imagine visualization objects able to exhibit both kinematic and dynamic behavior being desirable.

docs/pages/attributesJSON.rst
```diff
@@ -177,6 +174,12 @@ Below are object-specific physical quantities.  These values will override simil
 "requires_lighting"
 	- boolean
 	- Whether the object should be rendered with lighting or flat shading.
+"is_visible"
+	- boolean
+	- Whether the object's render asset should be created upon instancing.
```
jturner65, line 179: Functionality question : Can this be toggled after instancing, from the object's perspective? (i.e. can render assets be programmatically "turned on" and "turned off"?)

src/esp/metadata/managers/ObjectAttributesManager.cpp
```diff
@@ -114,6 +114,19 @@ void ObjectAttributesManager::setValsFromJSONDoc(
...
+  io::jsonIntoSetter<bool>(
+      jsonConfig, "is_visible",
+      std::bind(&ObjectAttributes::setIsVisible, objAttributes, _1));
+
+  // Whether or not the object is initiialy configured to be
+  // MotionType::VISUALIZATION_ONLY with instanced.
+  io::jsonIntoSetter<bool>(
+      jsonConfig, "is_collidable",
+      std::bind(&ObjectAttributes::setIsCollidable, objAttributes, _1));
+
```
jturner65, line 129: can the tags used to store the values in the ObjectAttributes for these two fields' setter/getters be changed to match the json tags? (i.e. from "isVisible" and "isCollidable" to "is_visible" and "is_collidable") This is for future exporting.

jturner65 (commented):
A few questions/requests :)

docs/pages/attributesJSON.rst
aclegg3: In the TODO list of the PR description. Not currently implemented.

docs/pages/attributesJSON.rst
Skylion007: @eundersander @aclegg3  There has to be a term from this we can borrow from game_design, right?

aclegg3:
> I think this kind of behavior should be governed by a flag toggling collisions instead of a motion type (i.e. the existing is_collidable flag). From a user's perspective, I can imagine visualization objects able to exhibit both kinematic and dynamic behavior being desirable.

I'm not sure we want to support a dynamic rigid object with no collision shape. It would just fall forever unless manually controlled to do otherwise. 🤔  Only reason I can think of would be visual effects like particle systems.

docs/pages/attributesJSON.rst
Skylion007: * incorporeal - having no physical body or form 
* ethereal - lack material substance
* is_ghost - ghosts can pass through things. ;)

jturner65:
> I'm not sure we want to support a dynamic rigid object with no collision shape. It would just fall forever unless manually controlled to do otherwise. thinking Only reason I can think of would be visual effects like particle systems.

Animation as well, as was stated elsewhere. I don't disagree that some management might be necessary for such objects (we probably ought to have some kind of distant object clean up anyway), but my point is just that it seems a bit heavy handed, and certainly semantically inappropriate, to manage this behavior through motion type.

Skylion007:
> 
> 
> > I think this kind of behavior should be governed by a flag toggling collisions instead of a motion type (i.e. the existing is_collidable flag). From a user's perspective, I can imagine visualization objects able to exhibit both kinematic and dynamic behavior being desirable.
> 
> I'm not sure we want to support a dynamic rigid object with no collision shape. It would just fall forever unless manually controlled to do otherwise. 🤔 Only reason I can think of would be visual effects like particle systems.

What it seems we really want are ethereal/ghost objects which aren't affected by gravity.

src/esp/bindings/PhysicsBindings.cpp
Skylion007: @aclegg3  Actually, wouldn't visualization only be the same an `MotionType::DISABLED` After all, nothing can be move it ? I would prefer `MotionType::ETHEREAL`.

Skylion007:
Well, this is quite an odd CI failure...

Skylion007:
@aclegg3 You may want to modify this function in this PR: https://github.com/facebookresearch/habitat-sim/blob/706b23f823cd83b368438fe7f475a89b6ee179ef/examples/tutorials/nb_python/ECCV_2020_Interactivity.py#L1028

Also we should really move that function to viz_utils. 👍

Skylion007:
CI error was just it being out of sync with master.

aclegg3:
Feature replaced with #932.

PR #926, opened 2020-11-10
Enable mypy on tests

Skylion007:
## Motivation and Context

This change cleans up some of the typing in Habitat_Sim further and enables mypy static checking on tests.

Also fixes a broken value in the mypy config and enables some brand new features like allowing for redefinition of types (ie. variable name reuse in local scope).

## How Has This Been Tested
Locally with Pytest

## Types of changes

## Checklist


habitat_sim/agent/controls/pyrobot_noisy_controls.py
```diff
@@ -26,14 +26,13 @@
 from habitat_sim.registry import registry
 
 
-@attr.s(auto_attribs=True)
```
erikwijmans, line 29: Why no more `attr.s`?  This was nice for debugging since I could just print the object.  If you want to get rid of `__init__`, you can add `init=False` to the decorator.

Skylion007: Yeah, in the future we may want to consider using dataclasses since they are part of the standard python library in 3.7>=

erikwijmans (approved):

PR #928, opened 2020-11-10
--Enable override of default Attributes Registration checks

jturner65:
## Motivation and Context
Currently to register an object or stage attributes the specified render asset handle must exist, either as a valid object on disk, a valid existing primitive, or (in the case of a stage) be "none".  If none of these conditions are true, the registration will fail and the attributes will not be added to the attributesManager library.  This is useful in case errors exist in the attributes, so that object creation does not fail due to misnamed target assets.  

The ability to override registration checks for object and stage attributes checks is useful for when render assets are computationally/"hand" crafted meshes.  This PR adds a boolean argument to the registration process that allows for overriding the non-existing asset failure with a warning. 

The default behavior will be to fail if specified render handle does not correspond to a known/knowable asset.

## How Has This Been Tested
C++ and python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/managers/SceneDatasetAttributesManager.cpp
+++ b/src/esp/metadata/managers/SceneDatasetAttributesManager.cpp
@@ -298,7 +298,8 @@ void SceneDatasetAttributesManager::readDatasetConfigsJSONCell(
 
 int SceneDatasetAttributesManager::registerObjectFinalize(
     attributes::SceneDatasetAttributes::ptr SceneDatasetAttributes,
-    const std::string& SceneDatasetAttributesHandle) {
+    const std::string& SceneDatasetAttributesHandle,
+    bool forceRegistration) {
   // adds template to library, and returns either the ID of the existing
   // template referenced by SceneDatasetAttributesHandle, or the next available
   // ID if not found.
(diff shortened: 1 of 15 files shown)


src/esp/metadata/managers/SceneDatasetAttributesManager.cpp
```diff
@@ -298,7 +298,8 @@ void SceneDatasetAttributesManager::readDatasetConfigsJSONCell(
 
 int SceneDatasetAttributesManager::registerObjectFinalize(
     attributes::SceneDatasetAttributes::ptr SceneDatasetAttributes,
-    const std::string& SceneDatasetAttributesHandle) {
+    const std::string& SceneDatasetAttributesHandle,
+    bool forceRegistration) {
```
Skylion007, line 302:
```suggestion
    bool) {
```

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -46,7 +46,8 @@ void StageAttributesManager::buildCtorFuncPtrMaps() {
 
 int StageAttributesManager::registerObjectFinalize(
     StageAttributes::ptr stageAttributes,
-    const std::string& stageAttributesHandle) {
+    const std::string& stageAttributesHandle,
+    bool forceRegistration) {
```
Skylion007, line 50:
```suggestion    
    bool) {
```

Skylion007 (changes requested):

Skylion007 (approved):

aclegg3 (approved):

PR #933, opened 2020-11-12
Orthographic camera support

jturner65:
## Motivation and Context
This PR introduces Orthographic cameras.  It also modifies Viewer so that the camera can be swapped between orthographic and perspective on the fly, without requiring the agent or the sensor to be remade.  Zoom support is also added, and can be seen in Viewer by rolling the mouse wheel.
    

## Types of changes
C++ and python tests

## Checklist


jturner65:
Example image

orthoPass1

src/esp/bindings/SensorBindings.cpp
```diff
@@ -48,6 +48,10 @@ void initSensorBindings(py::module& m) {
       .value("DEPTH", SensorType::DEPTH)
       .value("SEMANTIC", SensorType::SEMANTIC);
 
+  py::enum_<SensorSubtype>(m, "SensorSubType")
+      .value("PINHOLE", SensorSubtype::PINHOLE)
+      .value("ORHTOGRAPHIC", SensorSubtype::ORHTOGRAPHIC);
```
Skylion007, line 53: Typo

src/esp/sensor/Sensor.h
```diff
@@ -37,11 +37,16 @@ enum class ObservationSpaceType {
...
+  PINHOLE = 0,
+  ORHTOGRAPHIC = 1,
+};
+
 // Specifies the configuration parameters of a sensor
 struct SensorSpec {
   std::string uuid = "rgba_camera";
   SensorType sensorType = SensorType::COLOR;
-  std::string sensorSubtype = "pinhole";
+  SensorSubtype sensorSubtype = SensorSubtype::PINHOLE;
```
jturner65, line 49: Please note the addition of this enum to describe camera type.  Originally a string holding "pinhole" it did not seem to be used anywhere.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -61,6 +61,11 @@ void initGfxBindings(py::module& m) {
         Set this `Camera`'s projection matrix.
       )",
            "width"_a, "height"_a, "znear"_a, "zfar"_a, "hfov"_a)
+      .def("set_ortho_projection_matrix",
```
Skylion007, line 64: Add a Pytest to demonstrate usage.

src/utils/viewer/viewer.cpp
```diff
@@ -394,6 +394,9 @@ Viewer::Viewer(const Arguments& arguments)
   };
   agentConfig.sensorSpecifications[0]->resolution =
       esp::vec2i(viewportSize[1], viewportSize[0]);
+  // try orthographic camera
+  agentConfig.sensorSpecifications[0]->sensorSubtype =
+      esp::sensor::SensorSubtype::ORHTOGRAPHIC;
```
Skylion007, line 399: TYPO again

src/esp/sensor/Sensor.h
```diff
@@ -37,11 +37,16 @@ enum class ObservationSpaceType {
   TEXT = 2,
 };
 
+enum class SensorSubtype {
+  PINHOLE = 0,
+  ORHTOGRAPHIC = 1,
```
Skylion007, line 42: TYPO

src/esp/gfx/RenderCamera.cpp
```diff
@@ -76,6 +76,18 @@ RenderCamera& RenderCamera::setProjectionMatrix(int width,
...
 
+RenderCamera& RenderCamera::setOrthoProjectionMatrix(int width,
+                                                     int height,
+                                                     float znear,
+                                                     float zfar) {
+  auto orthoMat = Mn::Matrix4::orthographicProjection(Mn::Vector2{16.0f, 9.0f},
+                                                      znear, zfar);
+
+  MagnumCamera::setProjectionMatrix(orthoMat).setViewport(
+      Magnum::Vector2i(width, height));
```
jturner65, line 87: The {16.0f, 9.0f} is arbitrary, hacky and inappropriate, but I am still figuring out the appropriate configuration for this.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -76,6 +76,18 @@ RenderCamera& RenderCamera::setProjectionMatrix(int width,
   return *this;
 }
 
+RenderCamera& RenderCamera::setOrthoProjectionMatrix(int width,
+                                                     int height,
+                                                     float znear,
+                                                     float zfar) {
+  auto orthoMat = Mn::Matrix4::orthographicProjection(Mn::Vector2{16.0f, 9.0f},
```
Skylion007, line 83: magic numbers. Please describe with appropriately named constant.

src/esp/agent/Agent.cpp
```diff
@@ -31,10 +32,21 @@ Agent::Agent(scene::SceneNode& agentNode, const AgentConfiguration& cfg)
...
+      case sensor::SensorSubtype::PINHOLE: {
+        sensors_.add(sensor::PinholeCamera::create(sensorNode, spec));
+        break;
+      }
+      case sensor::SensorSubtype::ORHTOGRAPHIC: {
+        sensors_.add(sensor::OrthoCamera::create(sensorNode, spec));
+        break;
+      }
+      default: {
+        sensors_.add(sensor::PinholeCamera::create(sensorNode, spec));
```
Skylion007, line 45: Shouldn't we error?

Skylion007 (changes requested):
Serious typos

Skylion007:
@bigbike Would appreciate your input on this

src/esp/sensor/Sensor.h
Skylion007: This seems more like CAMERA_TYPE / PROJECTION_TYPE than sensor subtype

src/esp/agent/Agent.cpp
jturner65: I'm defaulting to pinhole because that field originally (sensorSubType) was ignored. If you think an error would be better choice for this, I can do that too.

Skylion007:
@mosra Seem to be getting weird magnum warnings from the orthomatrix. Not sure how to fix.

src/esp/gfx/RenderCamera.cpp
Skylion007: @jturner65 Does it work for other scenes?

src/esp/agent/Agent.cpp
Skylion007: I am sayiing we add one more check if it doesn't equal "pinhole," then we complain. I'd rather not have it fail silently.

src/esp/sensor/Sensor.h
jturner65: Just using the previous terminology - I wanted to keep the semantic changes to a minimum.

src/esp/gfx/RenderCamera.cpp
jturner65: In this very early WIP i have it hardcoded to run in viewer, regardless of scene, so it should.

jturner65:
> @mosra Seem to be getting weird magnum warnings from the orthomatrix. Not sure how to fix.

Not warnings, just weird behavior.

src/esp/gfx/RenderCamera.cpp
jturner65: just tested running viewer in other scenes.  Seems to work :)

src/esp/sensor/OrthoCamera.cpp
```diff
@@ -0,0 +1,180 @@
...
+        obs.buffer->data});
+  } else if (spec_->sensorType == SensorType::DEPTH) {
+    renderTarget().readFrameDepth(Magnum::MutableImageView2D{
+        Magnum::PixelFormat::R32F, renderTarget().framebufferSize(),
+        obs.buffer->data});
+  } else {
+    renderTarget().readFrameRgba(Magnum::MutableImageView2D{
+        Magnum::PixelFormat::RGBA8Unorm, renderTarget().framebufferSize(),
+        obs.buffer->data});
+  }
```
Skylion007, line 157: Future Proof by doing an elif notImplemented complaint.

Skylion007: Or maybe a switch statement would be more readable here?

jturner65: This code was lifted verbatim from pinhole. Probably ought to be moved to VisualSensor.

Skylion007: Ah leave it as then. @bigbike Something to think about during sensor redesign.

Skylion007 (approved):
Looks good to me, but I would like to here @bigbike and @mosra 's opinion

src/utils/viewer/viewer.cpp
```diff
@@ -273,6 +273,12 @@ Viewer::Viewer(const Arguments& arguments)
       .setHelp("object-dir",
                "Provide a directory to search for object config files "
                "(relative to habitat-sim directory).")
+      .addBooleanOption("use-ortho-camera")
+      .setHelp("use-ortho-camera", "Use orthographic camera to view scene.")
+      .addOption("ortho-scale", ".1")
```
Skylion007, line 278: You could combine these two args, but making use-ortho-camera True if orthoscale != 0/ -1/ NaN. etc... or even if it's just specified at all.

Skylion007 (changes requested):
Really would like to here from @mosra @bigbike on this PR.

src/esp/agent/Agent.cpp
```diff
@@ -31,10 +32,21 @@ Agent::Agent(scene::SceneNode& agentNode, const AgentConfiguration& cfg)
     // sensor
 
     auto& sensorNode = agentNode.createChild();
-    sensors_.add(
-        sensor::PinholeCamera::create(sensorNode, spec));  // transformed within
+    switch (spec->sensorSubtype) {
+      case sensor::SensorSubtype::PINHOLE: {
```
bigbike, line 36: SensorSubType::Pinhole
Please use "Pascal" style for the enum.

Skylion007: It occured me to we also need a str to enum mapping potentially or allow the ENUMs to be exposed on the habitat_lab side in the YAML configs.

src/esp/sensor/OrthoCamera.cpp
```diff
@@ -0,0 +1,181 @@
...
+  scale_ = std::atof(spec_->parameters.at("ortho_scale").c_str());
+}
+
+OrthoCamera& OrthoCamera::setProjectionMatrix(gfx::RenderCamera& targetCamera) {
+  // use ortho version of function
+  targetCamera.setOrthoProjectionMatrix(width_, height_, near_, far_, scale_);
+  return *this;
+}
+
+OrthoCamera& OrthoCamera::setTransformationMatrix(
```
bigbike, line 38: My biggest concern is that there is a lot of code duplication (from PinholeCamera) in this file.
We may need to figure out a way to solve this problem.

If you lift them into the visualSensor class, be careful. For each function, we need to think twice. Would all visual sensors need this function?

@mosra : Thoughts?

src/esp/sensor/Sensor.h
```diff
@@ -37,14 +37,20 @@ enum class ObservationSpaceType {
   TEXT = 2,
 };
 
+enum class SensorSubtype {
```
bigbike, line 40: SensorSubType

src/esp/sensor/Sensor.h
```diff
@@ -37,14 +37,20 @@ enum class ObservationSpaceType {
...
+  PINHOLE = 0,
```
bigbike, line 41: Pinhole = 0,
Orthographic = 1,

src/esp/sensor/Sensor.h
```diff
@@ -37,14 +37,20 @@ enum class ObservationSpaceType {
...
+  ORTHOGRAPHIC = 1,
+};
+
 // Specifies the configuration parameters of a sensor
 struct SensorSpec {
   std::string uuid = "rgba_camera";
   SensorType sensorType = SensorType::COLOR;
-  std::string sensorSubtype = "pinhole";
```
bigbike, line 44: I think before we use string here is because it would be very convenient to set it from the python interface via python bindings.

Skylion007: That's a great point!

src/esp/sensor/OrthoCamera.cpp
jturner65: Yes, i was concerned about the duplication as well, but I did not want to promote anything to visualSensor if they were not appropriate for all visual sensors.  I also knew the redesign is pending and didn't want to mess with that.

bigbike: Not a problem at all. In my opinion, this PR is part of the "camera redesign" project, not a blocker. :)
I think the only difference between the `PinholeCamera` and the `OrthoCamera` is the projection, whether it is perspective or orthogonal. Correct? 
IF it is correct, my idea is that we rename `PinholeCamera` to `CameraSensor` for example, which has the `enum ProjectionType`  properly set to `ProjectionType::Perspective` or `ProjectionType::Orthogonal`. 

Did I miss anything that cannot be merged?

eundersander: I like @bigbike 's suggestion, `CameraSensor` with different projection types.

It's a little messy how to handle `hfov_` and `scale_` members which only apply to one or the other projection type, but that hardly justifies making two separate classes.

Skylion007: Should these really be abstracted out in the class sense? Or should the code be placed in a different file and then use std::bind or lambdas instead.

src/esp/sensor/Sensor.h
mosra: Enum is just as convenient.

src/esp/sensor/OrthoCamera.cpp
mosra: I don't like the duplication either, this makes it a hell to maintain.

> It's a little messy how to handle `hfov_` and `scale_` members which only apply to one or the other projection type, but that hardly justifies making two separate classes.

You can express the perspective camera [in terms of near projection plane size](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Matrix4.html#aeed8f8ff2daedc28b7e3638e8e4ded47), which I'd say is pretty much equivalent to the orthographic projection size.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -61,6 +61,11 @@ void initGfxBindings(py::module& m) {
...
+           &RenderCamera::setOrthoProjectionMatrix, R"(
+        Set this `Orthographic Camera`'s projection matrix.
```
mosra, line 66: Naming/consistency: why not have `set_orthographic_projection_matrix()` instead if there's already a`SensorSubType::ORTHOGRAPHIC`? Same for the viewer option, `--use-orthographic-camera`.

Or alternatively use the contraction everywhere. Having both short and long form seems like an unnecessary variation that hurts API usability.

mosra:
>> @mosra Seem to be getting weird magnum warnings from the orthomatrix.
>
> Not warnings, just weird behavior.

@Skylion007 @jturner65 What kind of warnings? What weird behavior?

- - -

Besides the code duplication concern, I'm not sure about the `scale` parameter, and especially not about having it fixed upfront:

- Because the orthographic projection has no perspective shortening, a zoom done via moving the camera closer / further away doesn't work there. A different approach that should work for both the perspective and orthographic case without having to branch on camera type would be to express both in terms of [near plane size](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Matrix4.html#aeed8f8ff2daedc28b7e3638e8e4ded47) and implement zoom via scaling this size -- basically what you do now for the orthographic camera would be done for the perspective as well, for which this will result in a change of the FoV.
- Then I would change the API to allow this "zoom" to be modified at runtime instead of being specified at the beginning -- I don't know exactly how the viewer behaves, but I suppose the user may want to zoom in/out for inspecting the scene. Forcing them to specify this scale via a command-line option isn't a good UX ;)

jturner65:
> > > @mosra Seem to be getting weird magnum warnings from the orthomatrix.
> > 
> > 
> > Not warnings, just weird behavior.
> 
> @Skylion007 @jturner65 What kind of warnings? What weird behavior?
> 

Just took me a little bit to realize I had to scale the Ortho view. :)

> Besides the code duplication concern, I'm not sure about the `scale` parameter, and especially not about having it fixed upfront:
> 
>     * Because the orthographic projection has no perspective shortening, a zoom done via moving the camera closer / further away doesn't work there. A different approach that should work for both the perspective and orthographic case without having to branch on camera type would be to express both in terms of [near plane size](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Matrix4.html#aeed8f8ff2daedc28b7e3638e8e4ded47) and implement zoom via scaling this size -- basically what you do now for the orthographic camera would be done for the perspective as well, for which this will result in a change of the FoV.
> 
>     * Then I would change the API to allow this "zoom" to be modified at runtime instead of being specified at the beginning (@bigbike something to think about in the sensor redesign as well?) -- I don't know exactly how the viewer behaves, but I suppose the user may want to zoom in/out for inspecting the scene. Forcing them to specify this scale via a command-line option isn't a good UX ;)

Thanks @mosra, this was the kind of thing I was shooting for eventually (allowing the mouseWheel to act as a zoom, for example, was working on that in Viewer on Friday).

src/esp/sensor/OrthographicCamera.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "OrthographicCamera.h"
+
+namespace esp {
+namespace sensor {
+
+void OrthographicCamera::setProjectionParameters_TypeSpecific(
+    const SensorSpec::ptr& spec) {
+  scale_ = std::atof(spec->parameters.at("ortho_scale").c_str());
+}
+Mn::Matrix4 OrthographicCamera::recalcBaseProjectionMatrix() {
```
bigbike, line 14: recalc --> recompute
Long function name is never be a problem.

src/esp/sensor/CameraSensor.h
```diff
@@ -0,0 +1,131 @@
...
+  virtual void setProjectionParameters_TypeSpecific(
+      const SensorSpec::ptr& spec) = 0;
+
+  virtual Magnum::Matrix4 recalcBaseProjectionMatrix() = 0;
+
+  /**
+   * @brief Recalculate the projection Matrix used by this Camera Sensor, which
+   * should be recalculated every time near plane size changes.
+   */
+  void recalcProjectionMatrix() {
```
bigbike, line 83: recomputeProjectionMatrix()

src/utils/viewer/viewer.cpp
```diff
@@ -732,11 +752,17 @@ void Viewer::mouseScrollEvent(MouseScrollEvent& event) {
...
+  float mod = event.offset().y() > 0 ? modVal : 1.0 / modVal;
+  auto cameraSensor =
+      defaultAgent_->getSensorSuite().getSensors()["rgba_camera"];
+  auto cam = std::static_pointer_cast<esp::sensor::CameraSensor>(cameraSensor);
+  cam->setZoom(mod);
   redraw();
 
   event.setAccepted();
-}
+}  // namespace
```
bigbike, line 765: the comment is wrong.

src/utils/viewer/viewer.cpp
bigbike: I may not agree with you.
Compared to the  "use-ortho-camera", "ortho-scale" is a more advanced setting. You cannot expect any user is willing to set it. Most of them simply needs a toggle while keeping the other default parameters not changed.

src/esp/sensor/CameraSensor.h
```diff
@@ -0,0 +1,131 @@
...
+    projectionMatrix_ = zoomMatrix_ * recalcBaseProjectionMatrix();
+  }
+
+  /**
+   * @brief Read the observation that was rendered by the simulator
+   * @param[in,out] obs Instance of Observation class in which the observation
+   * will be stored
+   */
+  void readObservation(Observation& obs);
```
bigbike, line 92: Maybe we also need to make it virtual (for the future)?

jturner65:
Ok, most recent commit has the following substantive changes, among other things.  This is the proposed Final Commit, reviewer-willing :) 

- Removed orthographic and pinhole classes.  CameraSensor is sensor for camera, type of camera governs how class variable baseProjMatrix_  is synthesized, and changing type on existing camera will change projection. 
- Added ability to zoom/change FOV.  This modifies a class variable,  zoomMatrix_
- Added class variable projectionMatrix_ = zoomMatrix_ * baseProjMatrix_;  this is updated whenever either change
- Viewer now supports single tag "--orthographic" which has viewer start in orthographic camera mode.
- Viewer now supports toggling between ortho and perspecitve camera, by pressing 'c'
- Viewer now supports zooming, by spinning mouse wheel (using shift + wheel makes for finer grained zoom)

src/esp/sensor/Sensor.cpp
```diff
@@ -12,7 +12,8 @@ namespace esp {
 namespace sensor {
 
 Sensor::Sensor(scene::SceneNode& node, SensorSpec::ptr spec)
-    : Magnum::SceneGraph::AbstractFeature3D{node}, spec_(std::move(spec)) {
+    : Magnum::SceneGraph::AbstractFeature3D{node},
+      spec_(std::move(spec)) {  // why are we moving?
```
Skylion007, line 16: Performance. See clang-tidy performance checks.

src/esp/agent/Agent.cpp
Skylion007: getattr should work for this

src/esp/sensor/Sensor.cpp
jturner65: Moving a pointer makes that much of a difference?

bigbike (approved):

src/esp/bindings/GfxBindings.cpp
```diff
@@ -57,10 +57,16 @@ void initGfxBindings(py::module& m) {
...
-      .def("set_projection_matrix", &RenderCamera::setProjectionMatrix, R"(
-        Set this `Camera`'s projection matrix.
-      )",
-           "width"_a, "height"_a, "znear"_a, "zfar"_a, "hfov"_a)
+      .def("set_projection_matrix",
+           static_cast<RenderCamera& (RenderCamera::*)(int, int, float, float,
+                                                       float)>(
+               &RenderCamera::setProjectionMatrix),
+           R"(Set this `Camera`'s projection matrix.)", "width"_a, "height"_a,
+           "znear"_a, "zfar"_a, "hfov"_a)
```
mosra, line 65: I'm probably annoying with this already, but can the FoV be `Mn::Deg`? :grin: Now that you're redoing most of the API anyway.

Both in Python and C++.

src/esp/sensor/Sensor.h
```diff
@@ -37,14 +37,20 @@ enum class ObservationSpaceType {
   TEXT = 2,
 };
 
+enum class SensorSubType {
+  Pinhole = 0,
```
mosra, line 41: Dumb question, suggesting a possible rename -- what's the more common / clearer term? Perspective or pinhole camera?

src/utils/viewer/viewer.cpp
```diff
@@ -156,8 +163,9 @@ Key Commands:
   'q': Query the agent's state and print to terminal.
 
   Utilities:
+  'c' switch ortho/perspective camera.
```
mosra, line 166: FYI, in [magnum-player](https://magnum.graphics/showcase/player/) I'm using `Num5` to switch between the two, which matches what Blender uses.

src/utils/viewer/viewer.cpp
```diff
@@ -394,6 +405,15 @@ Viewer::Viewer(const Arguments& arguments)
...
       esp::vec2i(viewportSize[1], viewportSize[0]);
+  if (args.isSet("orthographic")) {
+    // try orthographic camera
+    agentConfig.sensorSpecifications[0]->sensorSubType =
+        esp::sensor::SensorSubType::Orthographic;
+
+  } else {
+    agentConfig.sensorSpecifications[0]->sensorSubType =
+        esp::sensor::SensorSubType::Pinhole;
+  }
```
mosra, line 416:
```suggestion
  agentConfig.sensorSpecifications[0]->sensorSubType =
      args.isSet("orthographic") ?
      esp::sensor::SensorSubType::Orthographic : 
      esp::sensor::SensorSubType::Pinhole;
```

Sorry, code golf, won't do this again :)

src/utils/viewer/viewer.cpp
```diff
@@ -412,6 +432,24 @@ Viewer::Viewer(const Arguments& arguments)
...
+  switch (oldCameraType) {
+    case esp::sensor::SensorSubType::Pinhole: {
+      cam->setCameraType(esp::sensor::SensorSubType::Orthographic);
+      break;
+    }
+    case esp::sensor::SensorSubType::Orthographic: {
+      cam->setCameraType(esp::sensor::SensorSubType::Pinhole);
+      break;
+    }
+  }
```
mosra, line 450:
```suggestion
  switch (oldCameraType) {
    case esp::sensor::SensorSubType::Pinhole: {
      cam->setCameraType(esp::sensor::SensorSubType::Orthographic);
      return;
    }
    case esp::sensor::SensorSubType::Orthographic: {
      cam->setCameraType(esp::sensor::SensorSubType::Pinhole);
      return;
    }
  }

  CORRADE_INTERNAL_ASSERT_UNREACHABLE();
```

This way, when we get a fisheye camera next, this place won't silently stop working but yet (compared to putting that assert in a `default:` clause) the compiler will still notify you that you have an unhandled enum value when the `SensorSubType` gets extended.

src/utils/viewer/viewer.cpp
```diff
@@ -732,11 +770,17 @@ void Viewer::mouseScrollEvent(MouseScrollEvent& event) {
   if (!event.offset().y()) {
     return;
   }
-
+  // Use shift for fine-grained zooming
+  float modVal = (event.modifiers() & MouseEvent::Modifier::Shift) ? 1.01 : 1.1;
+  float mod = event.offset().y() > 0 ? modVal : 1.0 / modVal;
+  auto cameraSensor =
+      defaultAgent_->getSensorSuite().getSensors()["rgba_camera"];
+  auto cam = std::static_pointer_cast<esp::sensor::CameraSensor>(cameraSensor);
```
mosra, line 778:
```suggestion
  auto& cameraSensor = // what is this type, actually?
      *defaultAgent_->getSensorSuite().getSensors()["rgba_camera"];
  auto& cam = static_cast<esp::sensor::CameraSensor&>(cameraSensor);
```

Less unneeded shared pointer copies ;) Also, could the first `auto` be an explicitly typed-out type? Because I personally can't tell what it is without browsing the docs first.

mosra (commented):
Very random comments, hope that's not too annoying.

src/esp/sensor/Sensor.h
jturner65: Probably perspective, but Pinhole also implies infinite depth of field, which is appropriate here, right?

src/utils/viewer/viewer.cpp
jturner65: I'll do that too then.

src/esp/bindings/GfxBindings.cpp
jturner65: Maybe this could wait for the sensor redesign?  I'm not really touching those components (the RenderCamera::setProjectionMatrix I used just sets the matrix).

src/esp/sensor/Sensor.h
mosra: Does it? I thought the code gives the user an option to set the far plane to a non-infinite value.

jturner65: Well, that's far-clipping, right?  Not blur.  Everything between near and far plane is in focus, so it's like a pinhole camera.

jturner65: I mean, if this were just a graphics app, then yeah, it would go without saying that it should be perspective.  But i can understand that vision folks would look at this and value the distinction that this is a pinhole camera. Not really sure, to be honest.

src/esp/bindings/GfxBindings.cpp
bigbike: This PR is part of sensor redesign project. :)
Would you do me a favor and solve it here? Otherwise I think it will be forgotten.
Thanks!

src/esp/sensor/Sensor.h
mosra: Ah, I see, I see. Yeah you're right, good point with the DOF.

bigbike: This is to distinguish the projection method, which is perspective VS orthogonal.
I am leaning towards using the "Perspective" here. :)

bigbike: But considering this is actually used under the "CameraSensor" and it is to distinguish different camera sensors, maybe "Pinhole" is reasonable in this case. (Later we may have "Fisheye").

So I do not know. It is up to you.

By the way is there any reason why we do not put it under the "CameraSensor" header file?

src/esp/bindings/GfxBindings.cpp
jturner65: Ok :)

bigbike: Thanks a lot! @jturner65

src/esp/sensor/Sensor.cpp
Skylion007: @jturner65 Feel free to remove it if you can make clang-tidy happy.

src/esp/sensor/Sensor.h
jturner65: Sensor spec is used for all sensors, not just CameraSensors (especially when others exist :) ), and i envisioned the SensorSubType field as a "catch-all" descriptor for some defining secondary feature of a sensor.  This is why I felt it would be more appropriate in Sensor.h than in CameraSensor.h.

Skylion007:
CameraSensor has very few methods exposed on the Python Bindings? Can we add some additional bindings so we can change the FOV and such without restarting.

Skylion007 (changes requested):
CameraSensor has very few methods exposed on the Python Bindings? Can we add some additional bindings so we can change the FOV and Zoom such without restarting.

Also adding a quick Python test of that behavior would be nice.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -112,7 +112,26 @@ void initSensorBindings(py::module& m) {
...
       m, "CameraSensor")
       .def(py::init_alias<std::reference_wrapper<scene::SceneNode>,
-                          const SensorSpec::ptr&>());
+                          const SensorSpec::ptr&>())
+      .def("set_projection_params", &CameraSensor::setProjectionParameters,
+           R"(Specify the projection parameters this CameraSensor should use.
+           Should be consumed by first querying this CameraSensor's SensorSpec
+           and then modifying as necessary.)",
+           "sensor_spec"_a)
+      .def(
```
Skylion007, line 121: Why is this set FOV? It should just be a property FOV that can be modified and has a corresponding setter/getter.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -112,7 +112,26 @@ void initSensorBindings(py::module& m) {
...
+          "set_fov",
+          static_cast<void (CameraSensor::*)(Mn::Deg)>(&CameraSensor::setFOV),
+          R"(Set the field of view to use for this CameraSensor.  Only applicable to 
+          Pinhole Camera Types)",
+          "fov"_a)
+      .def(
```
Skylion007, line 127: Why isn't this a property too?

src/esp/bindings/SensorBindings.cpp
```diff
@@ -112,7 +112,26 @@ void initSensorBindings(py::module& m) {
              VisualSensor, Magnum::SceneGraph::PyFeatureHolder<CameraSensor>>(
       m, "CameraSensor")
       .def(py::init_alias<std::reference_wrapper<scene::SceneNode>,
-                          const SensorSpec::ptr&>());
+                          const SensorSpec::ptr&>())
+      .def("set_projection_params", &CameraSensor::setProjectionParameters,
```
Skylion007, line 116: Is this really just setting the projection_params? Maybe a better API would be reconfigure_with_params or something? A more informative API here would be helpful.

Skylion007 (changes requested):

src/esp/bindings/SensorBindings.cpp
jturner65: because it is a modifier to a zoom scaling matrix, not a property.

Skylion007: camera_model.zoom_scale_matrix ?

jturner65: Why would i expose this though? It is arcane and could easily mess up the projection if not maintained properly.

Skylion007: Then don't other exposing it.

jturner65: Would python users really want to access this?

src/utils/viewer/viewer.cpp
```diff
@@ -156,6 +169,7 @@ Key Commands:
   'q': Query the agent's state and print to terminal.
 
   Utilities:
+  'num5' switch ortho/perspective camera.
```
Skylion007, line 172: Is this numpad 5? What about people on laptops?

jturner65: Just using what mosra suggested above.  No skin off my nose either way.

jturner65: I think the reasoning was that it is the same as what blender uses.

src/esp/bindings/SensorBindings.cpp
Skylion007: My point being, if you don't think they would, then don't bother exposing it.

I feel like they definitely would want to access the FOV though.

src/utils/viewer/viewer.cpp
Skylion007: I would pick another one.

src/esp/bindings/SensorBindings.cpp
jturner65: My rationale for 'restricting' access to that was that it only had meaning when used with a pinhole camera, so i didn't want it to seem like a "first class citizen", so to speak.   I am fine with making it a property if you think that would be appropriate. :).

Skylion007: You could always make it throw an error if modified / accessed for an ortho, no?

src/esp/bindings/SensorBindings.cpp
jturner65: Yes, what it does is take the passed sensor spec and copy the projection-related params into the sensor's spec.  Then, everything relating to those params is rebuilt.

src/utils/viewer/viewer.cpp
jturner65: I originally had 'c' :)

bigbike: In blender there is a function called "Emulate numpad". We do not have it here.
How about just using number '5'? KeyEvent::Key::Five

src/esp/sensor/Sensor.cpp
Skylion007: @jturner65 I removed it.

Skylion007: /home/circleci/project/habitat-sim/src/esp/sensor/Sensor.cpp:15:58: error: parameter 'spec' is passed by value and only copied once; consider moving it to avoid unnecessary copies [performance-unnecessary-value-param,-warnings-as-errors]

https://stackoverflow.com/questions/41871115/why-would-i-stdmove-an-stdshared-ptr
https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html

jturner65:
I'm just going to roll these back, @Skylion007 - moving it limits unnecessary overhead with atomic incrementers/decrementers.

src/utils/viewer/viewer.cpp
mosra: > What about people on laptops?

I have a numpad on my laptop, what's the problem? :trollface: 

Numpad five was chosen because other numpad keys switch between front/top/side views. Five alone doesn't have any meaning, and it would have *actual* issues on my layout because I'd need to do <kbd>Shift</kbd> + <kbd>ř</kbd> to get 5.

I'd say go back to what you had before, then.

src/esp/bindings/SensorBindings.cpp
jturner65: How about just an informational message?  An error seems drastic, but I will defer to your greater experience as a user of the system :).

jturner65:
This most recent commit adds python property bindings for the various frustum metrics (width, height, near plane dist, far plane dist) as properties, and recalculates appropriate projection matrix when any of these values, along with CameraType and fov, are changed.  Note : if someone is using an orthographic camera, and they change FOV, it will have no effect on the current camera, but will store the change, in case they go back to a perspective camera at some point in the future.

src/utils/viewer/viewer.cpp
```diff
@@ -142,6 +154,8 @@ Mouse Functions:
     (With 'enable-physics') Click a surface to instance a random primitive object at that location.
   SHIFT-RIGHT:
     Click a mesh to highlight it.
+  WHEEL:
+    Modify orthographic camera zoom/perspective camera FOV (+SHIFT for fine grained control)
```
Skylion007, line 158: Does wheel support pinch scrolling laptops? I don't mind plugging in a mouse for that feature, but it would be nice to add one last key binding for that.

Skylion007 (approved):
Looks good, minus that one nit about zoom. 

I will need to update https://github.com/facebookresearch/habitat-lab/blob/f89d53ea47e784fdf6e78fadccded645bb5913a7/habitat/sims/habitat_simulator/habitat_simulator.py#L235 to add allow users to set the camera_type from the habitat_lab side. I can take are of that once this is merged.

src/esp/bindings/SensorBindings.cpp
Skylion007: Works for me.

aclegg3 (approved):

src/utils/viewer/viewer.cpp
mosra: It does, it always did.

PR #935, opened 2020-11-13
refactor ResourceManager render assets

eundersander:
## Motivation and Context

This PR is (mostly) a no-op refactor of ResourceManager (no behavior change). The motivation is to unblock the upcoming render replay feature.

I introduce the terms "render asset" and "render asset instance". A render asset is basically a mesh. A render asset can be of type InstanceMesh ("IMesh"), PTex, Primitive, or General (see `isRenderAssetGeneral`).

A somewhat unrelated change is mixed in here: I'm converting some usage of ResourceKey (a hash) to std::string so that I can better serialize the key.

This is (mostly) a no-op refactor that conceptually replaces `loadPTexMeshData / loadInstanceMeshData / loadGeneralMeshData` with `loadRenderAssetPTex,createRenderAssetInstancePTex / loadRenderAssetIMesh,createRenderAssetInstanceIMesh / loadRenderAssetGeneral,createRenderAssetInstanceGeneralPrimitive`.

Some things you can do with render assets and instances:
- Load a render asset (without adding anything to the scene graph; see `loadRenderAsset`).
- Create instances of a render asset (add instances to the scene graph so they show up in observations; see `createRenderAssetInstance`).
- Move a render asset instance in the scene (and coming soon, you can hide or unhide it).
- Delete a render asset instance (remove the instance from the scene graph).
- Unload a render asset (only after all its instances have been deleted; at present, I think the only way to unload is to reconfigure/reset the simulator).

More details about render assets:
- Currently, a render asset instance isn't a formal class/object in our code. It's a subtree of scene nodes and drawables. It *is* a formal object in the serialization code I'm writing for render replay.
- A PhysicsObject's attributes include its render asset. A PhysicsObject has an associated render asset instance at runtime.
- A stage has an associated RGBD/"render" render asset instance and optional "semantic" render asset instance.
- An articulated object will have an associated render asset instance per link at runtime (articulated objects aren't merged to master yet).

Some goals/benefits of this refactor:
- Cleaner separation between render and simulation code: a replay renderer will be able to load and display stage visuals without calling loadStage, processing stageAttributes, or loading stage collision geometry.
- unified loading of render assets: loadRenderAsset
- unified creation of instances: createRenderAssetInstance
- Users of render assets are (mostly) agnostic to the underlying type: stage, objects, and articulated object links have associated render assets, but they (mostly) don't care what type.
- Any render asset type can be used for RGBD, Semantic, or both.
- Where render asset types differ, we're explicit. E.g. explicit asserts that convey that PTex doesn't support scale or lighting.

Some possible future use cases this unlocks (these are side benefits; the main motivation for this PR is unblocking render replay):
- preload render assets (load prior to use)
- unload and reload render assets on demand
- add separate semantic render assets to objects or links
- use PTex or InstanceMesh render assets with objects and links

## How Has This Been Tested

Loading stages with render assets of type PTex, InstanceMesh, and General (GLB), in the viewer
Adding physics objects in the viewer
Testing semantic rendering using a modified semantic_id_tutorial.py
I ran SimTest, which tests adding a primitive

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/RenderAssetInstanceCreation.h
+++ b/src/esp/assets/RenderAssetInstanceCreation.h
@@ -1,0 +1,30 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include "Asset.h"
+
+#include "esp/scene/SceneNode.h"
+
+#include "Magnum/Resource.h"
+
+#include <Corrade/Containers/Optional.h>
+
+namespace esp {
+namespace assets {
+
+// parameters to control how a render asset instance is created
+struct RenderAssetInstanceCreation {
+  std::string filepath;  // see also AssetInfo::filepath
+  Corrade::Containers::Optional<Magnum::Vector3> scale;
+  bool isStatic =
+      false;            // inform the renderer that this instance won't be moved
+  bool isRGBD = false;  // use in RGB and depth observations
+  bool isSemantic = false;  // use in semantic observations
+  std::string lightSetupKey;
+};
+
+}  // namespace assets
+}  // namespace esp
--- a/src/esp/assets/RenderAssetInstanceCreationInfo.h
+++ b/src/esp/assets/RenderAssetInstanceCreationInfo.h
@@ -28,9 +28,7 @@ struct RenderAssetInstanceCreationInfo {
   RenderAssetInstanceCreationInfo(
       const std::string& _filepath,
       const Corrade::Containers::Optional<Magnum::Vector3>& _scale,
-      bool isStatic,
-      bool isRGBD,
-      bool isSemantic,
+      const Flags& _flags,
       const std::string& _lightSetupKey);
 
   bool isStatic() const { return bool(flags & Flag::IsStatic); }
--- a/src/esp/assets/RenderAssetInstanceCreationInfo.h
+++ b/src/esp/assets/RenderAssetInstanceCreationInfo.h
@@ -10,6 +10,7 @@
 
 #include "Magnum/Resource.h"
 
+#include <Corrade/Containers/EnumSet.h>
 #include <Corrade/Containers/Optional.h>
 
 namespace esp {
@@ -17,12 +18,28 @@ namespace assets {
 
 // parameters to control how a render asset instance is created
 struct RenderAssetInstanceCreationInfo {
+  enum class Flag : unsigned int {
+    IsStatic = 1 << 0,  // inform the renderer that this instance won't be moved
+    IsRGBD = 1 << 1,    // use in RGB and depth observations
+    IsSemantic = 1 << 2  // use in semantic observations
+  };
+  typedef Corrade::Containers::EnumSet<Flag> Flags;
+
+  RenderAssetInstanceCreationInfo(
+      const std::string& _filepath,
+      const Corrade::Containers::Optional<Magnum::Vector3>& _scale,
+      bool isStatic,
+      bool isRGBD,
+      bool isSemantic,
+      const std::string& _lightSetupKey);
+
+  bool isStatic() const { return bool(flags & Flag::IsStatic); }
+  bool isRGBD() const { return bool(flags & Flag::IsRGBD); }
+  bool isSemantic() const { return bool(flags & Flag::IsSemantic); }
+
   std::string filepath;  // see also AssetInfo::filepath
   Corrade::Containers::Optional<Magnum::Vector3> scale;
-  bool isStatic =
-      false;            // inform the renderer that this instance won't be moved
-  bool isRGBD = false;  // use in RGB and depth observations
-  bool isSemantic = false;  // use in semantic observations
+  Flags flags;
   std::string lightSetupKey;
 };
 
--- a/src/esp/assets/RenderAssetInstanceCreationInfo.h
+++ b/src/esp/assets/RenderAssetInstanceCreationInfo.h
@@ -16,7 +16,7 @@ namespace esp {
 namespace assets {
 
 // parameters to control how a render asset instance is created
-struct RenderAssetInstanceCreation {
+struct RenderAssetInstanceCreationInfo {
   std::string filepath;  // see also AssetInfo::filepath
   Corrade::Containers::Optional<Magnum::Vector3> scale;
   bool isStatic =
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -877,6 +877,7 @@ scene::SceneNode* ResourceManager::createRenderAssetInstancePTex(
     const RenderAssetInstanceCreation& creation,
     scene::SceneNode* parent,
     DrawableGroup* drawables) {
+#ifdef ESP_BUILD_PTEX_SUPPORT
   ASSERT(!creation.scale);  // PTex doesn't support scale
   ASSERT(creation.lightSetupKey ==
          NO_LIGHT_KEY);  // PTex doesn't support lighting
@@ -915,6 +916,11 @@ scene::SceneNode* ResourceManager::createRenderAssetInstancePTex(
     computePTexMeshAbsoluteAABBs(*meshes_.at(metaData.meshIndex.first),
                                  staticDrawableInfo);
   return instanceRoot;
+#else
+  LOG(ERROR) << "PTex support not enabled. Enable the BUILD_PTEX_SUPPORT CMake "
+                "option when building.";
+  return nullptr;
+#endif
 }
 
 bool ResourceManager::loadRenderAssetIMesh(const AssetInfo& info) {
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -210,7 +210,7 @@ bool ResourceManager::loadStage(
       auto& semanticRootNode = semanticSceneGraph.getRootNode();
       auto& semanticDrawables = semanticSceneGraph.getDrawables();
 
-      RenderAssetInstanceCreation creation{
+      RenderAssetInstanceCreationInfo creation{
           .filepath = semanticStageFilename,
           .scale = Cr::Containers::NullOpt,
           .isStatic =
@@ -255,7 +255,7 @@ bool ResourceManager::loadStage(
 
   AssetInfo renderInfo = assetInfoMap.at("render");
 
-  RenderAssetInstanceCreation renderCreation{
+  RenderAssetInstanceCreationInfo renderCreation{
       .filepath = renderInfo.filepath,
       .scale = Cr::Containers::NullOpt,
       .isStatic = true,
@@ -265,7 +265,7 @@ bool ResourceManager::loadStage(
 
   bool renderMeshSuccess = loadStageInternal(renderInfo,  // AssetInfo
                                              &renderCreation,
-                        &rootNode,          // parent scene node
+                                             &rootNode,    // parent scene node
                                              &drawables);  //  drawable group
   if (!renderMeshSuccess) {
     LOG(ERROR)
@@ -281,7 +281,7 @@ bool ResourceManager::loadStage(
     // should this be checked to make sure we do not reload?
     bool collisionMeshSuccess = loadStageInternal(colInfo,  // AssetInfo
                                                   nullptr,  // creation
-                          nullptr,            // parent scene node
+                                                  nullptr,  // parent scene node
                                                   nullptr);  // drawable group
 
     if (!collisionMeshSuccess) {
@@ -454,7 +454,7 @@ bool ResourceManager::loadRenderAsset(const AssetInfo& info) {
 }
 
 scene::SceneNode* ResourceManager::createRenderAssetInstance(
-    const RenderAssetInstanceCreation& creation,
+    const RenderAssetInstanceCreationInfo& creation,
     scene::SceneNode* parent,
     DrawableGroup* drawables,
     std::vector<scene::SceneNode*>* visNodeCache) {
@@ -500,7 +500,7 @@ scene::SceneNode* ResourceManager::createRenderAssetInstance(
 
 bool ResourceManager::loadStageInternal(
     const AssetInfo& info,
-    const RenderAssetInstanceCreation* creation,
+    const RenderAssetInstanceCreationInfo* creation,
     scene::SceneNode* parent,
     DrawableGroup* drawables) {
   // scene mesh loading
@@ -521,7 +521,7 @@ bool ResourceManager::loadStageInternal(
           if (!loadRenderAsset(info)) {
             return false;
           }
-      } else {
+        } else {
           if (resourceDict_[filename].assetInfo != info) {
             // Right now, we only allow for an asset to be loaded with one
             // configuration, since generated mesh data may be invalid for a new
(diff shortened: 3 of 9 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -180,31 +188,85 @@ bool ResourceManager::loadStage(
       ((_physicsManager != nullptr) &&
        (_physicsManager->getInitializationAttributes()->getSimulator().compare(
             "none") != 0));
-  const Magnum::ResourceKey& renderLightSetup(stageAttributes->getLightSetup());
+  const std::string renderLightSetupKey(stageAttributes->getLightSetup());
```
eundersander, line 191: See https://cvmlp.slack.com/archives/G0131KVLBLL/p1604955346004100

src/esp/assets/ResourceManager.cpp
```diff
@@ -180,31 +188,85 @@ bool ResourceManager::loadStage(
...
   std::map<std::string, AssetInfo> assetInfoMap =
       createStageAssetInfosFromAttributes(stageAttributes, buildCollisionMesh,
                                           loadSemanticMesh);
 
+  // set equal to current Simulator::activeSemanticSceneID_ value
```
eundersander, line 196: This is not new code. I've simply moved the semantic loading to happen earlier in this function. I did this because I need to know if a render asset instance is going to be used for RGBD/Semantic/both at the time of creation. I want to know as early as possible whether we're going to be in "single scene graph" vs "dual scene graph" mode.

src/esp/assets/ResourceManager.cpp
```diff
@@ -180,31 +188,85 @@ bool ResourceManager::loadStage(
...
+
+      auto& semanticSceneGraph =
+          sceneManagerPtr->getSceneGraph(activeSemanticSceneID);
+      auto& semanticRootNode = semanticSceneGraph.getRootNode();
+      auto& semanticDrawables = semanticSceneGraph.getDrawables();
+
+      RenderAssetInstanceCreation creation{
+          .filepath = semanticStageFilename,
+          .scale = Cr::Containers::NullOpt,
+          .isStatic =
```
eundersander, line 216: isStatic roughly corresponds to the previous logic for computeSemanticAABBs.

Skylion007:
Broke the JS bindings...

src/esp/assets/ResourceManager.cpp
```diff
@@ -440,20 +513,31 @@ bool ResourceManager::loadStageInternal(
           << filename;
       meshSuccess = false;
     } else {
-      if (info.type == AssetType::INSTANCE_MESH) {  // semantic
-        meshSuccess = loadInstanceMeshData(
-            info, parent, drawables, computeAbsoluteAABBs, splitSemanticMesh);
-      } else if (info.type == AssetType::FRL_PTEX_MESH) {
-        meshSuccess = loadPTexMeshData(info, parent, drawables);
-      } else if (info.type == AssetType::SUNCG_SCENE) {
+      if (info.type == AssetType::SUNCG_SCENE) {
```
eundersander, line 516: SunCG scenes are not render assets.

src/esp/assets/ResourceManager.cpp
```diff
@@ -440,20 +513,31 @@ bool ResourceManager::loadStageInternal(
...
+        if (resourceDict_.count(info.filepath) == 0) {
+          if (!loadRenderAsset(info)) {
+            return false;
+          }
+        } else {
+          if (resourceDict_[filename].assetInfo != info) {
+            // Right now, we only allow for an asset to be loaded with one
+            // configuration, since generated mesh data may be invalid for a new
+            // configuration
+            LOG(ERROR)
```
eundersander, line 529: This logic was moved from `loadGeneralMeshData` to here.

src/esp/assets/ResourceManager.cpp
```diff
@@ -822,109 +872,149 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
...
+
+  scene::SceneNode* instanceRoot = &parent->createChild();
+
+  for (int iMesh = start; iMesh <= end; ++iMesh) {
+    auto* pTexMeshData = dynamic_cast<PTexMeshData*>(meshes_[iMesh].get());
+
+    pTexMeshData->uploadBuffersToGPU(false);
+
+    for (int jSubmesh = 0; jSubmesh < pTexMeshData->getSize(); ++jSubmesh) {
+      scene::SceneNode& node = instanceRoot->createChild();
```
eundersander, line 900: This is one of the few behavior changes in the PR. For the instance mesh, instead of adding all these mesh child sceneNodes directly to the parent sceneNode, I first create an `instanceRoot` sceneNode. I want every render asset instance to have a single root sceneNode. This is needed for render replay.

I acknowledge there's probably a slight perf cost. Long-term, our new Vulkan batched renderer probably won't use a SceneNode hierarchy, so this shouldn't matter.

src/esp/assets/ResourceManager.cpp
```diff
@@ -822,109 +872,149 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
...
+
+  std::vector<StaticDrawableInfo> staticDrawableInfo;
+  auto indexPair = getMeshMetaData(creation.filepath).meshIndex;
+  int start = indexPair.first;
+  int end = indexPair.second;
+
+  scene::SceneNode* instanceRoot = &parent->createChild();
+
+  for (uint32_t iMesh = start; iMesh <= end; ++iMesh) {
+    scene::SceneNode& node = instanceRoot->createChild();
```
eundersander, line 983: See https://github.com/facebookresearch/habitat-sim/pull/935/files?diff=split&w=1#r523235399 .

src/esp/assets/ResourceManager.cpp
```diff
@@ -822,109 +872,149 @@ bool ResourceManager::loadPTexMeshData(const AssetInfo& info,
...
     scene::SceneNode* parent,
-    DrawableGroup* drawables,
-    bool computeAbsoluteAABBs,
-    bool splitSemanticMesh /* = true */) {
-  if (info.type != AssetType::INSTANCE_MESH) {
-    LOG(ERROR) << "loadInstanceMeshData only works with INSTANCE_MESH type!";
-    return false;
+    DrawableGroup* drawables) {
+  ASSERT(!creation.scale);  // PTex doesn't support scale
+  ASSERT(creation.lightSetupKey ==
```
eundersander, line 880: Explicit asserts to show what a render asset type supports.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1005,108 +1095,106 @@ bool ResourceManager::loadGeneralMeshData(
...
-    int end = meshMetaData.meshIndex.second;
-    if (0 <= start && start <= end) {
-      for (int iMesh = start; iMesh <= end; ++iMesh) {
-        meshes_[iMesh]->uploadBuffersToGPU(forceReload);
-      }
-    }
-  }  // forceReload
-     // TODO: cache visual nodes added by this process
-  std::vector<scene::SceneNode*> visNodeCache;
+  if (creation.scale) {
```
eundersander, line 1159: Currently, instances for objects use scale and instances for stages do not.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1833,7 +1899,15 @@ bool ResourceManager::loadSUNCGHouseFile(const AssetInfo& houseInfo,
         nodeIds.push_back(id);
         objectNode.setId(nodeIndex);
         if (info.type == AssetType::SUNCG_OBJECT) {
-          loadGeneralMeshData(info, &objectNode, drawables);
+          CHECK(loadRenderAsset(info));
```
eundersander, line 1902: This codepath is untested so who knows if it actually works. https://cvmlp.slack.com/archives/G0131KVLBLL/p1604940887000900

src/esp/assets/ResourceManager.h
```diff
@@ -658,24 +659,21 @@ class ResourceManager {
    * If both parent and drawables are provided, add the mesh to the
    * scene graph for rendering.
    * @param info The @ref AssetInfo for the mesh, already parsed from a file.
+   * @param creation How to instance the render asset, or nullptr if not
+   * instancing.
    * @param parent The @ref scene::SceneNode to which the mesh will be added
-   * as a child.
+   * as a child. See also creation->isRGBD and creation->isSemantic. nullptr if
```
eundersander, line 665: There's an important subtlety here: currently, the way the user chooses RGBD vs Semantic is by passing the appropriate parent and drawables (from either the RGBD or Semantic scene graph). But I'm also requiring the user to specify isRGBD and isSemantic in the creation struct (so that I can record the choice in the replay data).

Eventually, we could refactor this interface so that the user just specifies the creation struct, and then we choose the correct parent and drawables on the backend.

aclegg3 (commented):
Overall looks fine to me. Minor comment that we are using the renderAsset loading pipeline for importing collision proxy assets also. Technically these are render-ready and could be instanced, so I don't think the re-naming is a problem. Just wanted to point this out.

bigbike (commented):
@mosra : Can you take another round of review? Thanks!

src/esp/assets/ResourceManager.cpp
```diff
@@ -422,14 +433,76 @@ esp::geo::CoordinateFrame ResourceManager::buildFrameFromAttributes(
...
+  return meshSuccess;
+}
+
+scene::SceneNode* ResourceManager::createRenderAssetInstance(
+    const RenderAssetInstanceCreation& creation,
+    scene::SceneNode* parent,
+    DrawableGroup* drawables,
+    std::vector<scene::SceneNode*>* visNodeCache) {
+  // assert that asset is already loaded
+  CHECK(resourceDict_.count(creation.filepath));
```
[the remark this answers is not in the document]
mosra: How is a CHECK different from an ASSERT? I'd just use CORRADE_ASSERT here.

src/esp/assets/ResourceManager.cpp
```diff
@@ -180,31 +188,85 @@ bool ResourceManager::loadStage(
...
-                        false,              // split semantic mesh or not
-                        renderLightSetup);  // light setup
 
+  RenderAssetInstanceCreation renderCreation{
+      .filepath = renderInfo.filepath,
+      .scale = Cr::Containers::NullOpt,
+      .isStatic = true,
+      .isRGBD = true,
+      .isSemantic = !isSeparateSemanticScene,
+      .lightSetupKey = renderLightSetupKey};
```
mosra, line 264: Apologies if I'm not up-to-date on C++ standard requirements in Habitat, but designated initializers are standard only since C++20 or require a particular GNU extension, which may hurt portability (especially if we'd want to reconsider Windows support again). I'd personally stay with standard C++ with no compiler-specific extensions.

Same in other places (I spotted one or two more occurence at least).

src/esp/assets/ResourceManager.cpp
```diff
@@ -422,14 +433,76 @@ esp::geo::CoordinateFrame ResourceManager::buildFrameFromAttributes(
...
 
+bool ResourceManager::loadRenderAsset(const AssetInfo& info) {
+  bool meshSuccess = false;
+  if (info.type == AssetType::FRL_PTEX_MESH) {
+    meshSuccess = loadRenderAssetPTex(info);
+  } else if (info.type == AssetType::INSTANCE_MESH) {
+    meshSuccess = loadRenderAssetIMesh(info);
+  } else if (isRenderAssetGeneral(info.type)) {
+    meshSuccess = loadRenderAssetGeneral(info);
+  } else {
```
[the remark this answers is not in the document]
mosra: Yes, compared to what's there now, `CORRADE_INTERNAL_ASSERT_UNREACHABLE()` will give the compiler a correct hint, allowing for better optimizations even in release builds.

src/esp/assets/ResourceManager.cpp
```diff
@@ -422,14 +433,76 @@ esp::geo::CoordinateFrame ResourceManager::buildFrameFromAttributes(
...
+        !visNodeCache);  // createRenderAssetInstanceIMesh doesn't support this
+    newNode = createRenderAssetInstanceIMesh(creation, parent, drawables);
+  } else if (isRenderAssetGeneral(info.type) ||
+             info.type == AssetType::PRIMITIVE) {
+    newNode = createRenderAssetInstanceGeneralPrimitive(
+        creation, parent, drawables, visNodeCache);
+  } else {
+    ASSERT(false);  // createRenderAssetInstance doesn't yet support the
+                    // requested asset type
+    newNode = nullptr;
```
mosra, line 490: `CORRADE_INTERNAL_ASSERT_UNREACHABLE()` here again, and because it marks the code as unreachable, the `newNode = nullptr` isn't needed to suppress "potentially uninitialized variable" warnings.

src/esp/assets/ResourceManager.cpp
```diff
@@ -440,20 +513,31 @@ bool ResourceManager::loadStageInternal(
...
-        // Unknown type, just load general mesh data
-        meshSuccess = loadGeneralMeshData(info, parent, drawables,
-                                          computeAbsoluteAABBs, lightSetupKey);
+        // load render asset if necessary
+        if (resourceDict_.count(info.filepath) == 0) {
+          if (!loadRenderAsset(info)) {
+            return false;
+          }
+        } else {
+          if (resourceDict_[filename].assetInfo != info) {
```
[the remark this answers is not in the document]
mosra: `Mn::Warning{} << bla` :) Behaviorally same as `Error{}` but can be redirected / muted independently of it.

mosra (commented):
I feel a bit like being thrown in a cold water here -- I'm not fully aware of the progress / design / ideas around replay, so I can't really do a thorough review without first accompanying myself with everything related to this work (which I'm not sure I have capacity for, right now).

So all I can offer is random hints and suggestions re Magnum usage :) Hope that's fine. 

Oh, and in addition to the comments below basically everything @bigbike says.

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
bigbike: CHECK simply outputs Warning while ASSERT interrupts system running, I believe.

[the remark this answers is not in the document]
mosra: Oh. But the comment says `assert` and I feel like it should be an assert :)

src/esp/assets/RenderAssetInstanceCreation.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "Magnum/Resource.h"
+
+#include <Corrade/Containers/Optional.h>
+
+namespace esp {
+namespace assets {
+
+// parameters to control how a render asset instance is created
+struct RenderAssetInstanceCreation {
```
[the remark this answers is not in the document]
jturner65: Do we want to rename the file to match the struct?

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
aclegg3: I believe `CHECK` will also abort if the condition is not met. Similar to CORRADE_ASSERT.

jturner65 (approved):
One minor question (feel free to ignore) otherwise LGTM.

src/esp/assets/ResourceManager.cpp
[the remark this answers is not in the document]
bigbike: OK. I checked the google docs. Yes, you guys are right. I was wrong.
`// CHECK dies with a fatal error if condition is not true. `

Skylion007:
Pull from master to fix pre-commit

src/esp/assets/RenderAssetInstanceCreationInfo.h
```diff
@@ -0,0 +1,47 @@
...
+    IsStatic = 1 << 0,  // inform the renderer that this instance won't be moved
+    IsRGBD = 1 << 1,    // use in RGB and depth observations
+    IsSemantic = 1 << 2  // use in semantic observations
+  };
+  typedef Corrade::Containers::EnumSet<Flag> Flags;
+
+  RenderAssetInstanceCreationInfo(
+      const std::string& _filepath,
+      const Corrade::Containers::Optional<Magnum::Vector3>& _scale,
+      bool isStatic,
```
bigbike, line 31: Because you have defined Flags, these 3 booleans can be replaced by a 'Flags flags`, aren't they?
That is the purpose to have the Flags introduced. Did I miss anything?

aclegg3 (approved):
LGTM

src/esp/assets/ResourceManager.cpp
```diff
@@ -82,6 +82,14 @@ constexpr char ResourceManager::DEFAULT_LIGHTING_KEY[];
 constexpr char ResourceManager::DEFAULT_MATERIAL_KEY[];
 constexpr char ResourceManager::WHITE_MATERIAL_KEY[];
 constexpr char ResourceManager::PER_VERTEX_OBJECT_ID_MATERIAL_KEY[];
+
+namespace {
+bool isRenderAssetGeneral(AssetType type) {
+  return type == AssetType::MP3D_MESH || type == AssetType::UNKNOWN ||
+         type == AssetType::SUNCG_OBJECT;
```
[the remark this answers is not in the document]
eundersander: I chatted with Kathryn Duvall from FB legal and she says this will be fine. What I'm trying to do here is not break the SUNCG loading codepath, with the big caveat that we can't test this loading path on SUNCG data or otherwise use SUNCG data. We should consider support for SUNCG data to be deprecated. I will add a runtime warning about this codepath being deprecated.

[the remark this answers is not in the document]
dhruvbatra: Awesome! Thanks for following up on this!

bigbike (approved):

PR #937, opened 2020-11-16
add BulletPhysicsManager::getContactPoints and python bindings

eundersander:
## Motivation and Context

`BulletPhysicsManager::getContactPoints` returns `ContactPointData` objects describing the contacts from the most recent physics substep. This implementation is roughly identical to PyBullet's `getContactPoints`.

Users should take care to call stepWorld with a dt equal to the Bullet substep dt prior to calling getContactPoints, because getContactPoints only returns data about the most recent substep. There's a warning to encourage correct usage.

The most obvious usage is to get info about the recent substep during physics simulation. However, another planned use case is similar to an async "what if" collision query, as follows:
1. Save the state of the physics world.
2. Move desired query object(s) to a desired query pose.
3. stepWorld(substepDt)
4. getContactPoints()
5. Restore the state of the physics world.

There are more direct, efficient ways to do async queries in C++ Bullet, but the steps above are common in existing PyBullet user code, e.g. https://github.com/caelan/pybullet-planning/blob/8ec64398f41a2151de5ce1ab6213cfb220cde924/examples/test_kuka_pick.py#L22 . So let's offer this function in Habitat Bullet in order to be compatible with PyBullet.

## How Has This Been Tested

Ad hoc testing in viewer and rigid_object_tutorial.py. We'll need to add unit tests before merging this into `master`.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/PhysicsBindings.cpp
+++ b/src/esp/bindings/PhysicsBindings.cpp
@@ -61,6 +61,28 @@ void initPhysicsBindings(py::module& m) {
       .def_readonly("hits", &RaycastResults::hits)
       .def_readonly("ray", &RaycastResults::ray)
       .def("has_hits", &RaycastResults::hasHits);
+
+  py::class_<ContactPointData, ContactPointData::ptr>(m, "ContactPointData")
+      .def(py::init(&ContactPointData::create<>))
+      .def_readwrite("object_id_a", &ContactPointData::objectIdA)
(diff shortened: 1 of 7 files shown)


Skylion007 (approved):

Skylion007:
Yet another reason, we really should just wrap the PyBullet API.

eundersander:
I glanced at the first CI compile error and it looks unrelated to my change. I think CI is broken for this branch, so I'll ignore the CI error and merge anyway.

PR #942, opened 2020-11-18
Use Bullet 2.89 as a CMake subproject

mosra:
## Motivation and Context

The main motivation here is to get all Habitat users on the same version of Bullet. Compare to the current setup of having the Hab user install Bullet on her system: conda and package managers currently give different versions of Bullet across platforms.

This PR chooses Bullet 2.89 because that's what most current Hab users are "on" (again, depends on platform). However, another big motivation for this PR is to unblock a version upgrade. We expect to upgrade to 3.06 very soon. 

There are various downsides to no longer using a system install of Bullet: increased build time (building Bullet from source), increased conda distribution size, increased space on disk for the Habitat-sim worktree (all the Bullet source code), etc.

@mosra has included a `USE_SYSTEM_BULLET` build option which lets a user get the old behavior and avoid some downsides. I strongly suggest we don't "officially support" this build option. I.e., if a user posts a github issue about a problem with Habitat and system-installed Bullet, our first and probably only reply should be "switch to the submodule Bullet".

Notes from @mosra 

90% of the work is done in Magnum's [FindBullet.cmake](https://github.com/mosra/magnum-integration/blob/master/modules/FindBullet.cmake) and this only uses it. I added a `USE_SYSTEM_BULLET` option for people like me who don't want to build an extra copy of it every time.

Some caveats:

 * Bullet's buildsystem doesn't correctly express dependencies between static libraries and due to how CMake handles linking to targets, it's *impossible* to link them in correct order. This can be only mitigated with CMake 3.13, but I assume Habitat's minimum CMake version requirement is not that high yet. Another alternative is opening a PR on Bullet's repo and fixing it for them. Because of this, bundled Bullet is built as dynamic libraries.
 * However, Bullet doesn't explicitly export symbols in their shared libraries and Habitat compiles with -fvisibility=hidden, which means there are linker errors unless the hidden visibility is disabled for the Bullet subproject.
 * There's a gigaton of CMake warnings coming out of the Bullet submodule. Unfortunately there isn't anything that could be done to fix those, except for (again) opening a PR on their repo.

## How Has This Been Tested

The builds I tested (@eundersander ):
- bundled bullet and system bullet.
- cmake 3.12 (builds bullet shared libs) and cmake 3.14 (builds bullet static libs)
- Fedora, Ubuntu, and Mac
- with and without --bullet

Cc: @Ram81

### DIFF
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -148,7 +148,7 @@ if(BUILD_WITH_BULLET AND NOT USE_SYSTEM_BULLET)
   # libs, causing linker errors on Magnum side. If you have CMake 3.13, the
   # Find module is able to correct that on its own, otherwise you need to
   # enable BUILD_SHARED_LIBS to build as shared.
-  if(NOT CORRADE_TARGET_EMSCRIPTEN)
+  if((NOT CORRADE_TARGET_EMSCRIPTEN) AND CMAKE_VERSION VERSION_LESS 3.13)
     set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
   else()
     # On Emscripten we require 3.13, so there it's fine (and there we can't use
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -144,7 +144,7 @@ if(BUILD_WITH_BULLET AND NOT USE_SYSTEM_BULLET)
   # This is needed in case BUILD_EXTRAS is enabled, as you'd get a CMake syntax
   # error otherwise
   set(PKGCONFIG_INSTALL_PREFIX "lib${LIB_SUFFIX}/pkgconfig/")
- 
+
   # caches CXX_FLAGS so we can reset them at the end
   set(_PREV_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
 
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -148,7 +148,13 @@ if(BUILD_WITH_BULLET AND NOT USE_SYSTEM_BULLET)
   # libs, causing linker errors on Magnum side. If you have CMake 3.13, the
   # Find module is able to correct that on its own, otherwise you need to
   # enable BUILD_SHARED_LIBS to build as shared.
-  set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
+  if(NOT CORRADE_TARGET_EMSCRIPTEN)
+    set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
+  else()
+    # On Emscripten we require 3.13, so there it's fine (and there we can't use
+    # shared libs)
+    set(BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
+  endif()
   # ... and because we have to build shared libs, we need exported symbols,
   # however the whole Habitat is built with -fvisibility=hidden and Bullet
   # doesn't export any of its symbols and relies on symbols being visible by
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -144,6 +144,10 @@ if(BUILD_WITH_BULLET AND NOT USE_SYSTEM_BULLET)
   # This is needed in case BUILD_EXTRAS is enabled, as you'd get a CMake syntax
   # error otherwise
   set(PKGCONFIG_INSTALL_PREFIX "lib${LIB_SUFFIX}/pkgconfig/")
+ 
+  # caches CXX_FLAGS so we can reset them at the end
+  set(_PREV_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+
   # Bullet's buildsystem doesn't correctly express dependencies between static
   # libs, causing linker errors on Magnum side. If you have CMake 3.13, the
   # Find module is able to correct that on its own, otherwise you need to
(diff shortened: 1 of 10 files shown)


mosra:
@Skylion007 oh, dang, forgot about emscripten. There you have to keep `BUILD_SHARED_LIBS` on `OFF` in `dependencies.cmake` and be sure to use CMake 3.13+.

Skylion007:
@mosra Not sure how to handle that inside of CMake. 

Also habitat's minimum CMake requirement is that for the emscriptem build currently so I don't it's that bad to raise the CMake requirement that much.

mosra:
@Skylion007 Pushed a change that hopefully does it. I don't have the Emscripten build set up locally, so unfortunately I can't test myself. Hope you can :)

src/cmake/dependencies.cmake
```diff
@@ -127,6 +127,44 @@ if(BUILD_PYTHON_BINDINGS)
...
+  set(BUILD_EXTRAS OFF CACHE BOOL "" FORCE)
+  set(BUILD_BULLET3 OFF CACHE BOOL "" FORCE)
+  # This is needed in case BUILD_EXTRAS is enabled, as you'd get a CMake syntax
+  # error otherwise
+  set(PKGCONFIG_INSTALL_PREFIX "lib${LIB_SUFFIX}/pkgconfig/")
+  # Bullet's buildsystem doesn't correctly express dependencies between static
+  # libs, causing linker errors on Magnum side. If you have CMake 3.13, the
+  # Find module is able to correct that on its own, otherwise you need to
+  # enable BUILD_SHARED_LIBS to build as shared.
+  if(NOT CORRADE_TARGET_EMSCRIPTEN)
```
erikwijmans, line 151: Can you add a version check here and build static if >= 3.13?  I'd like to statically link this for conda as that'll reduce the number of things we need to worry about rpath. The conda-build is just grabbing latest cmake, so no need to worry about versions there.

erikwijmans (commented):
On the subject of the conda build, can you remove these lines:

https://github.com/facebookresearch/habitat-sim/blob/master/conda-build/macos_matrix_builder.py#L58
https://github.com/facebookresearch/habitat-sim/blob/master/conda-build/linux_matrix_builder.py#L82

https://github.com/facebookresearch/habitat-sim/blob/master/conda-build/habitat-sim/meta.yaml#L17
https://github.com/facebookresearch/habitat-sim/blob/master/conda-build/habitat-sim/meta.yaml#L66

mosra:
If someone is able to take over and finish this, I'll be happy (@eundersander?). I opened this just to unblock everyone with the build issues, but I should really be focusing on something else ;)

erikwijmans:
> If someone is able to take over and finish this, I'll be happy (@eundersander?). I opened this just to unblock everyone with the build issues, but I should really be focusing on something else ;)

Made my requested changes then :)

src/cmake/dependencies.cmake
```diff
@@ -148,7 +148,7 @@ if(BUILD_WITH_BULLET AND NOT USE_SYSTEM_BULLET)
   # libs, causing linker errors on Magnum side. If you have CMake 3.13, the
   # Find module is able to correct that on its own, otherwise you need to
   # enable BUILD_SHARED_LIBS to build as shared.
-  if(NOT CORRADE_TARGET_EMSCRIPTEN)
+  if((NOT CORRADE_TARGET_EMSCRIPTEN) AND CMAKE_VERSION VERSION_LESS 3.13)
     set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
   else()
     # On Emscripten we require 3.13, so there it's fine (and there we can't use
```
Skylion007, line 154: Actually, we don't error anymore on CMake < 3.13. Don't we have to add a specific additional require or else people could fail on emscripten silently if build_with_bullet and < 3.13. I'd add an error if ```CORRADE_TARGET_EMSCRIPTEN AND CMAKE_VERSION VERSION_LESS 3.13```

mosra: Don't worry, I already thought about that -- it's done by FindBullet itself ;) https://github.com/mosra/magnum-integration/blob/a897805489102030da09644f264983944ddbf824/modules/FindBullet.cmake#L73-L75

Skylion007: Ah, just tested this. Good thinking.

Skylion007 (approved):

Skylion007:
Just one thing to note: since we are exporting all the debug symbols from Bullet, this does increase our binary size to 267mb.

mosra:
> 267mb

Hahah. Can you put the `-fvisibility=hidden` workaround into an `if()` so it's done only when building as shared?

erikwijmans:
Good catch there @Skylion007

src/cmake/dependencies.cmake
```diff
@@ -150,19 +150,19 @@ if(BUILD_WITH_BULLET AND NOT USE_SYSTEM_BULLET)
   # enable BUILD_SHARED_LIBS to build as shared.
   if((NOT CORRADE_TARGET_EMSCRIPTEN) AND CMAKE_VERSION VERSION_LESS 3.13)
     set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
+    # however the whole Habitat is built with -fvisibility=hidden and Bullet
+    # doesn't export any of its symbols and relies on symbols being visible by
+    # default. Which means we have to compile it without hidden visibility.
+    set(_PREV_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    string(REPLACE "-fvisibility=hidden" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    add_subdirectory(${DEPS_DIR}/bullet3 EXCLUDE_FROM_ALL)
```
mosra, line 158: wait no, this isn't right :D the `add_subdirectory()` needs to be outside this `if`, plus you forgot half of the comment below

... and then at the end, after `add_subdirectory()` this `if()` again, which resets the CMAKE_CXX_FLAGS back

eundersander:
>If someone is able to take over and finish this, I'll be happy (@eundersander?). 
Thanks for helping with this. I can babysit this PR from here out. I'll do a little testing at least before merging this.

src/cmake/dependencies.cmake
```diff
@@ -127,6 +127,47 @@ if(BUILD_PYTHON_BINDINGS)
...
+  # Bullet's buildsystem doesn't correctly express dependencies between static
+  # libs, causing linker errors on Magnum side. If you have CMake 3.13, the
+  # Find module is able to correct that on its own, otherwise you need to
+  # enable BUILD_SHARED_LIBS to build as shared.
+  if((NOT CORRADE_TARGET_EMSCRIPTEN) AND CMAKE_VERSION VERSION_LESS 3.13)
+    set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
+    # however the whole Habitat is built with -fvisibility=hidden and Bullet
+    # doesn't export any of its symbols and relies on symbols being visible by
+    # default. Which means we have to compile it without hidden visibility.
+    # ... and because we have to build shared libs, we need exported symbols,
```
mosra, line 160: Come on. The comment lines aren't even in correct order. This goes first.

Skylion007:
Why are we using Bullet 2.89? I know it's confusing but it looks like Erwin is using the v3 tag now...

mosra:
Can you please stop doing unapproved changes? @eundersander wanted to stay on 2.89 for now so that's why this PR pins to the 2.89 tag.

Skylion007:
I'll revert. @eundersander Any specific reason we want to be holden to 2.89 vs the latest version?

aclegg3:
> I'll revert. @eundersander Any specific reason we want to be holden to 2.89 vs the latest version?

2.89 was released in Dec. 2019. Current release of Bullet is [v3.06](https://github.com/bulletphysics/bullet3/releases/tag/3.06). Looks like a couple of nice changes here. E.g. added explicit DynamicType (static, kinematic, dynamic). I think we should consider this release version.

eundersander:
> I'll revert. @eundersander Any specific reason we want to be holden to 2.89 vs the latest version?

I'd like to keep this PR as small as possible. Habitat is currently mostly "on" 2.89, in the sense that 1. Mac package manager gives you 2.89, and 2. Andrew Szot told me he's on 2.89.

Let's wait and see what we break here by making Bullet a submodule. After the dust settles, we can upgrade to 3.06.

eundersander (approved):
LGTM. I added notes about my testing in the PR description.

PR #946, opened 2020-11-22
Refactor Multi-Agent Sensors API and allow dynamic adding of Sensors

### LINKED ISSUE #856: Add clean Python API to add sensors without restarting the simulator
## 🚀 Feature

According to @erikwijmans , we can actually add sensors after the simulator has been instationated by we currently don't have clean bindings to do so. 

` You need to add the sensor spec to the agent's _sensors SensorSuite via .add(sensor_object) and then you add the actual rendering object to the simulators _sensors dictionary (the habitat_sim.simulator.Sensor object)`

We should add a nice Python function to do so.
## Motivation
Removes another reason to have to restart the simulator and is a useful feature. It would be useful to doing things like creating 360 sensors on the fly in Habitat_Lab. 

## Pitch
You should be able to add and remove sensors without restarting the environment. (A fun use case would be a robot turning on an arm mounted camera for example).

Skylion007:
## Motivation and Context

Closes #856 

This also allows other static sensors to be added to the scene (like security cameras / static ortho cameras / bird eye views etc...)

## How Has This Been Tested
Locally. Will add a test case shortly, just want to finalize the API. 

Also expanded the _sensor_dict to be multi-agent compatible, let me know if I should be doing it in a different way, like by making _sensors a ```dict[int, dict]``` instead. 
With this, the only thing unblocking multi-agent support on the habitat-sim part would be testing and the step function.

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -142,9 +142,17 @@ class Agent(object):
         if reconfigure_sensors:
             self._sensors.clear()
             for spec in self.agent_config.sensor_specifications:
-                self.add_sensor(spec)
+                self.add_sensor(spec, modify_agent_config=False)
 
-    def add_sensor(self, spec: hsim.SensorSpec) -> None:
+    def add_sensor(
+        self, spec: hsim.SensorSpec, modify_agent_config: bool = True
+    ) -> None:
+        assert (
+            spec.uuid not in self._sensors
+        ), f"Error, {spec.uuid} already exists in the sensor suite"
+        if modify_agent_config:
+            assert spec not in self.agent_config.sensor_specifications
+            self.agent_config.sensor_specifications.append(spec)
         self._sensors.add(hsim.CameraSensor(self.scene_node.create_child(), spec))
 
     def act(self, action_id: Any) -> bool:
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -142,9 +142,10 @@ class Agent(object):
         if reconfigure_sensors:
             self._sensors.clear()
             for spec in self.agent_config.sensor_specifications:
-                self._sensors.add(
-                    hsim.CameraSensor(self.scene_node.create_child(), spec)
-                )
+                self.add_sensor(spec)
+
+    def add_sensor(self, spec: hsim.SensorSpec) -> None:
+        self._sensors.add(hsim.CameraSensor(self.scene_node.create_child(), spec))
 
     def act(self, action_id: Any) -> bool:
         r"""Take the action specified by action_id
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -142,9 +142,9 @@ class Agent(object):
         if reconfigure_sensors:
             self._sensors.clear()
             for spec in self.agent_config.sensor_specifications:
-                self.add_sensor(spec, modify_agent_config=False)
+                self._add_sensor(spec, modify_agent_config=False)
 
-    def add_sensor(
+    def _add_sensor(
         self, spec: hsim.SensorSpec, modify_agent_config: bool = True
     ) -> None:
         assert (
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -66,13 +66,13 @@ class Simulator(SimulatorBackend):
     config: Configuration
     agents: List[Agent] = attr.ib(factory=list, init=False)
     _num_total_frames: int = attr.ib(default=0, init=False)
-    _default_agent_id: Optional[int] = attr.ib(default=0)
+    _default_agent_id: Optional[int] = attr.ib(default=None, init=False)
     __sensors: List[Dict[str, "Sensor"]] = attr.ib(factory=list, init=False)
     _initialized: bool = attr.ib(default=False, init=False)
     _previous_step_time: float = attr.ib(
         default=0.0, init=False
     )  # track the compute time of each step
-    _last_state: List[Optional[AgentState]] = attr.ib(factory=list, init=False)
+    __last_state: List[Optional[AgentState]] = attr.ib(factory=list, init=False)
 
     @staticmethod
     def _sanitize_config(config: Configuration) -> None:
@@ -130,7 +130,7 @@ class Simulator(SimulatorBackend):
 
         self.config = None
 
-        self._last_state = []
+        self.__last_state = []
 
         super().close()
 
@@ -231,7 +231,7 @@ class Simulator(SimulatorBackend):
         self.__sensors: List[Dict[str, Sensor]] = [
             dict() for i in range(len(config.agents))
         ]
-        self._last_state = [None] * len(self.__sensors)
+        self.__last_state = [None] * len(self.__sensors)
         for agent_id, agent_cfg in enumerate(config.agents):
             for spec in agent_cfg.sensor_specifications:
                 self._update_simulator_sensors(spec.uuid, agent_id=agent_id)
@@ -263,7 +263,7 @@ class Simulator(SimulatorBackend):
                 )
 
         agent.set_state(initial_state, is_initial=True)
-        self._last_state[agent_id] = agent.state
+        self.__last_state[agent_id] = agent.state
         return agent
 
     @overload
@@ -307,12 +307,20 @@ class Simulator(SimulatorBackend):
         return self.get_agent(self._default_agent_id)
 
     @property
+    def _last_state(self) -> AgentState:
+        return self.__last_state[self._default_agent_id]
+
+    @_last_state.setter
+    def _last_state(self, state: AgentState) -> None:
+        self.__last_state[self._default_agent_id] = state
+
+    @property
     def _sensors(self) -> Dict[str, "Sensor"]:
         # TODO Deprecate and remove
         return self.__sensors[self._default_agent_id]
 
     def last_state(self, agent_id: int = 0) -> Optional[AgentState]:
-        return self._last_state[agent_id]
+        return self.__last_state[agent_id]
 
     @overload
     def step(
@@ -344,7 +352,7 @@ class Simulator(SimulatorBackend):
         for agent_id, agent_act in action.items():
             agent = self.get_agent(agent_id)
             collided_dict[agent_id] = agent.act(agent_act)
-            self._last_state[agent_id] = agent.state
+            self.__last_state[agent_id] = agent.state
 
         # step physics by dt
         step_start_Time = time.time()
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -144,7 +144,7 @@ class Simulator(SimulatorBackend):
         super().seed(new_seed)
         self.pathfinder.seed(new_seed)
 
-    def reset(self) -> Dict[str, ndarray]:
+    def reset(self) -> Dict[str, Union[ndarray, "Tensor"]]:
         super().reset()
         for i in range(len(self.agents)):
             self.reset_agent(i)
@@ -242,17 +242,25 @@ class Simulator(SimulatorBackend):
         )
 
     def add_sensor(self, sensor_spec: SensorSpec, agent_id: int = 0) -> None:
-        if sensor_spec.sensor_type == SensorType.SEMANTIC:
-            raise NotImplementedError(
-                "Adding a semantic sensor currently requires restarting the simulator"
+        if (
+            (
+                not self.config.sim_cfg.load_semantic_mesh
+                and sensor_spec.sensor_type == SensorType.SEMANTIC
             )
-        elif (
-            not self.config.sim_cfg.requires_texture
-            and sensor_spec.sensor_type == SensorType.COLOR
+            or (
+                not self.config.sim_cfg.requires_textures
+                and sensor_spec.sensor_type == SensorType.COLOR
+            )
+            or (
+                not self.config.sim_cfg.create_renderer
+                and sensor_spec.sensor_type == SensorType.DEPTH
+            )
         ):
+            sensor_type = sensor_spec.sensor_type
             raise ValueError(
-                """Error: Textures not loaded.
-                Cannot dynamically add COLOR sensors unless one already exists."""
+                f"""Data for {sensor_type} sensor was not loaded during Simulator init.
+                    Cannot dynamically add a {sensor_type} sensor unless one already exists.
+                    """
             )
         agent = self.get_agent(agent_id=agent_id)
         agent._add_sensor(sensor_spec)
@@ -368,7 +376,7 @@ class Simulator(SimulatorBackend):
         for agent_id, agent_act in action.items():
             agent = self.get_agent(agent_id)
             collided_dict[agent_id] = agent.act(agent_act)
-            self.__last_state[agent_id] = agent.state
+            self.__last_state[agent_id] = agent.get_state()
 
         # step physics by dt
         step_start_Time = time.time()
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -6,7 +6,7 @@
 
 import time
 from os import path as osp
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union, cast, overload
 
 import attr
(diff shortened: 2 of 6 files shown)


habitat_sim/agent/agent.py
```diff
@@ -142,9 +142,10 @@ def reconfigure(
         if reconfigure_sensors:
             self._sensors.clear()
             for spec in self.agent_config.sensor_specifications:
-                self._sensors.add(
-                    hsim.CameraSensor(self.scene_node.create_child(), spec)
-                )
+                self.add_sensor(spec)
+
+    def add_sensor(self, spec: hsim.SensorSpec) -> None:
```
Skylion007, line 147: Should the agent config be updated here for dynamically adding the Sensor?

erikwijmans: Yeah. Also makes sense to add an assert that the sensor uuid isn't already in the config.

habitat_sim/simulator.py
```diff
@@ -25,12 +25,17 @@
...
 from habitat_sim.logging import logger
 from habitat_sim.nav import GreedyGeodesicFollower, NavMeshSettings, PathFinder
-from habitat_sim.sensor import SensorType
+from habitat_sim.sensor import SensorSpec, SensorType
 from habitat_sim.sensors.noise_models import make_sensor_noise_model
 from habitat_sim.sim import SimulatorBackend, SimulatorConfiguration
 from habitat_sim.utils.common import quat_from_angle_axis
 
 
+def _overwrite_sensor_spec_uuid(spec: SensorSpec, agent_id: int = 0) -> SensorSpec:
```
Skylion007, line 34: @erikwijmans Is overwriting the UUID like this for multi-agents the way to go?

Skylion007: Maybe I should make this method a class function so that it can be overrided?

erikwijmans: IMO returning a list of observation-dicts makes more sense for multi-agent

habitat_sim/agent/agent.py
```diff
@@ -142,9 +142,18 @@ def reconfigure(
         if reconfigure_sensors:
             self._sensors.clear()
             for spec in self.agent_config.sensor_specifications:
-                self._sensors.add(
-                    hsim.CameraSensor(self.scene_node.create_child(), spec)
-                )
+                self.add_sensor(spec, modify_agent_config=False)
+
+    def add_sensor(
```
erikwijmans, line 147: This should be a private method since it doesn't make sense for anyone to call it besides the simulator.

habitat_sim/simulator.py
Skylion007: @erikwijmans Yeah, I just wanted to minimize API breakage between Habitat-API and Habitat-Lab for now. This is for the use case for having multiple static sensors around the scene (IE. Security Cameras) or alternatively agents that are moved outside of step.

habitat_sim/agent/agent.py
Skylion007: Sounds good.

habitat_sim/simulator.py
erikwijmans: But the question was specifically "is overwriting the UUID like this for multi-agents the way to go?".

We can always still return just a dict when there is a single agent.

Skylion007: While UUID implies UNIQUE to me, meaning that any sensor regardless of agent should have a UNIQUE UID.

Skylion007: Since the UUID is part of the config though, it prevents a config from being reused for multiple agents so some UUID mutation has / should occur in my opinion.

Skylion007: Maybe the UUID overwriting could be done in Habitat-Lab?

erikwijmans: I disagree that it has to be globally unique.  It has to be unique for that agent, but I don't see what it needs to be globally unique.

jturner65 (approved):
LGTM.

Skylion007:
@erikwijmans Completely redesigned the API as you asked. I still need to fix that one faulty test case in tests, but overloading our APIs like that really makes typing and API design a bit difficult.

habitat_sim/simulator.py
Skylion007: @erikwijmans I just started the redesign, but our call guarantees get quite messy but doing this. Really don't like returning different types like that...

Skylion007:
@erikwijmans The question is should we have the user specify a dict of actions or a list? I am partial to a dict since it allow some agents to have NOOP actions, but I am open to suggestions.

habitat_sim/simulator.py
Skylion007: Okay, I think I cleaned it up quite a bit.

Skylion007:
@erikwijmans I think we may want step to return a `Dict[str, Dict]`, instead of a List. Thoughts?

habitat_sim/simulator.py
```diff
@@ -106,20 +106,20 @@ def __attrs_post_init__(self) -> None:
...
-        for sensor in self._sensors.values():
-            sensor.close()
-            del sensor
+        for agent_sensorsuite in self.__sensors:
+            for sensor in agent_sensorsuite.values():
+                sensor.close()
+                del sensor
 
-        self._sensors = {}
+        self.__sensors = []
```
erikwijmans, line 114: Why was this changed to `__`?

erikwijmans (commented):
In terms of returning a list vs. returning a dict, since the actions are input as a dict, returning a dict makes sense.


Then, for consistency, having `get_sensor_observations` also return a dictionary makes sense

habitat_sim/simulator.py
Skylion007: For backwards compatibility. It's not a public API, but we have a lot of calls into. I emulate the old behavior with an '@property' getter.

habitat_sim/simulator.py
```diff
@@ -247,36 +256,96 @@ def initialize_agent(
...
+        Dict[str, Union[bool, ndarray, "Tensor"]],
+    ]:
         self._num_total_frames += 1
-        collided = self._default_agent.act(action)
-        self._last_state = self._default_agent.get_state()
+        return_single = False
+        if not isinstance(action, dict):
+            action = {self._default_agent_id: action}
+            return_single = True
+        agent_ids = sorted(action.keys())
```
Skylion007, line 330: I really don't like this behavior, it's kind of wonky to return a List of Dicts in sorted order. Should really be a Dict of Dict's, right @erikwijmans ?

erikwijmans: Yeah, Dict of Dict's makes sense.  `get_sensor_observations` should also then be a dict of dicts for consistency.

habitat_sim/simulator.py
erikwijmans: Hmm, well, we should make a public API for getting the sensor objects.

Skylion007: Yeah, that can be a separate PR though.

habitat_sim/simulator.py
Skylion007: Less sure about this since the IDs are passed in directly as a list.

Skylion007: Both solutions are annoying for different reasons. Iterating through a dict of a dicts is a bit painful and requires some subtle instancing to detect properly compared to a list at least but figuring out which ordering is return is a bit annoying.

We could always sort the list of agent dicts by agent_ids, but that can get a bit messy if not all agents generate observations...

erikwijmans: Yeah, I agree that it isn't ideal.  My thought is that even if dict of dicts is slightly more awkward, consistency is better because if one is list of dicts while the other is dict of dicts, that will require two different code paths.

What about returning an OrderedDict so `.values()` get's you the current list?

Skylion007:
@erikwijmans I don't this will actually work as is in habitat_lab, because habitat_lab currently maps the action as a dict of the action value (str or int) to a the action arguments (like the location for teleport). To pass a multi-agent action on Habitat-Lab won't work since all the action values [str, int] are wrapped in a dict and dicts are an unhashable type and therefore cannot be keys.

Skylion007:
I don't think sets or lists are hashable types, either... Maybe we need to create a MultiAgent Action class?

erikwijmans:
I think above got resolved in DMs?

Skylion007:
@erikwijmans Yeah, it did. Happy with how it is now.

erikwijmans (approved):

habitat_sim/simulator.py
```diff
@@ -216,17 +225,46 @@ def __set_from_config(self, config: Configuration):
...
+
+    def _update_simulator_sensors(self, uuid: str, agent_id: int = 0) -> None:
+        self.__sensors[agent_id][uuid] = Sensor(
+            sim=self, agent=self.get_agent(agent_id), sensor_id=uuid
+        )
 
-        for i in range(len(self.agents)):
-            self.initialize_agent(i)
+    def add_sensor(self, sensor_spec: SensorSpec, agent_id: int = 0) -> None:
+        if (
```
Skylion007, line 245: @bigbike This is going to get messy as we add more SensorTypes. We should consider a way of making this cleaner.

Skylion007:
So I read about into this: https://medium.com/@vermashresth/craft-and-solve-multi-agent-problems-using-rllib-and-tensorforce-a3bd1bb6f556 . and there are some current issues with the API:

* "reset should return an observation dict for all agents" . Should reset take a 'returned_agent_ids values?
* There should be a way to get reward measures for agents that don't act, (I think this is actually find depending on how we implement it on the Habitat-Lab level.)

habitat_sim/simulator.py
mathfac: Maybe, those checks should live close to Sensor code of particular type and should be checked before particular sensor type creation in `update_simulator_sensors` method.

habitat_sim/simulator.py
```diff
@@ -244,43 +288,124 @@ def initialize_agent(
                 )
 
         agent.set_state(initial_state, is_initial=True)
-        self._last_state = agent.state
+        if agent_id == len(self.__last_state):
+            self.__last_state.append(agent.state)
```
mathfac, line 292: Warning: mix of dictionaries and lists. What if agent_ids aren't sequential.

habitat_sim/simulator.py
```diff
@@ -244,43 +288,124 @@ def initialize_agent(
...
+        dt: float = 1.0 / 60.0,
+    ) -> Union[
+        Dict[str, Union[bool, ndarray, "Tensor"]],
+        Dict[int, Dict[str, Union[bool, ndarray, "Tensor"]]],
+    ]:
         self._num_total_frames += 1
-        collided = self._default_agent.act(action)
-        self._last_state = self._default_agent.get_state()
+        if isinstance(action, MutableMapping):
+            return_single = False
```
mathfac, line 384: nit: to be consistent with `get_observations` worth handle `return_single = True` in if condition.

habitat_sim/simulator.py
```diff
@@ -244,43 +288,124 @@ def initialize_agent(
...
+        else:
+            action = cast(Dict[int, Union[str, int]], {self._default_agent_id: action})
+            return_single = True
+        collided_dict: Dict[int, bool] = {}
```
mathfac, line 388:
```suggestion

        collided_dict: Dict[int, bool] = {}
```

habitat_sim/simulator.py
```diff
@@ -244,43 +288,124 @@ def initialize_agent(
...
+        for agent_id, agent_act in action.items():
+            agent = self.get_agent(agent_id)
+            collided_dict[agent_id] = agent.act(agent_act)
+            self.__last_state[agent_id] = agent.get_state()
```
mathfac, line 392: Worth to make `__last_state` a dict as well to avoid handling list vs dict issues, like out of index.

mathfac (changes requested):
@Skylion007, thank you a lot for contributing to Multi-Agent direction. I think we should lock the scope for this PR as we will have more insights once we add first model baseline for the multi-agent functionality. Then we can make better design choices.

I think "reset should return an observation dict for all agents" make sense. For this we need to have dict or param of agent number at least.
The reward calculation is Hab-Lab area, there we can calculate measure for all agents.

Requested some changes.

habitat_sim/simulator.py
Skylion007: @mathfac They have to be at init time currently. There is now way to create a non-sequential agent_id

habitat_sim/simulator.py
Skylion007: It makes it a lot messier. I originally had it this way. See the earlier commits.

habitat_sim/simulator.py
Skylion007: get_agent will also currently cause out of index, so I don't see the problem?

Skylion007: If we really wanted it to be clean, we would move last state handling into the agent itself. There is no reason for the Sim to keep track of it. The agent can perfectly keep track of it.

habitat_sim/simulator.py
Skylion007: @mathfac They should be handled in the C++ API or have a boolean on the Sensors that correspond which flags are valid / needed for an arbitrary sensor type.

habitat_sim/simulator.py
```diff
@@ -135,7 +146,7 @@ def seed(self, new_seed: int) -> None:
         super().seed(new_seed)
         self.pathfinder.seed(new_seed)
 
-    def reset(self) -> Dict[str, ndarray]:
+    def reset(self) -> Dict[str, Union[ndarray, "Tensor"]]:
```
mathfac, line 149: For now `reset()` will return observations only for agent_id == 0.

mathfac:
Thank you for the answers regarding dict vs list. We should be good to go with the PR as soon as reset for all agents will be supported.

unnat:
@Skylion007 First, thank you for considering me for external feedback. Also, great to see improved multi-agent support in Habitat!

With reference to:
```
        for agent_id, agent_act in action.items():
            agent = self.get_agent(agent_id)
            collided_dict[agent_id] = agent.act(agent_act)
            self.__last_state[agent_id] = agent.get_state()
```
Here, actions are being sequentially executed, which is sufficient for navigational tasks. As we tackle multi-agent tasks, we would face situations where the success of the action of one agent is dependent on the actions+states of the other agents. For example, [in this work](https://arxiv.org/pdf/1910.05512.pdf#page=8) the door opens only if another agent is close to a switch (dependence on the state of other agents). Similarly,  agents can [move heavy objects](https://arxiv.org/pdf/2007.04979.pdf#page=6) only if they are execute coordinated actions (dependence on the action of other agents).

Hence, allowing the flexibility where a user/developer can code up the environment's transition dynamics as a function of the list of actions and list of current states of the agents, might be a robust way forward for more interesting coordinated tasks.

Skylion007:
> @Skylion007 First, thank you for considering me for external feedback. Also, great to see improved multi-agent support in Habitat!
> 
> With reference to:
> 
> ```
>         for agent_id, agent_act in action.items():
>             agent = self.get_agent(agent_id)
>             collided_dict[agent_id] = agent.act(agent_act)
>             self.__last_state[agent_id] = agent.get_state()
> ```
> 
> Here, actions are being sequentially executed, which is sufficient for navigational tasks. As we tackle multi-agent tasks, we would face situations where the success of the action of one agent is dependent on the actions+states of the other agents. For example, [in this work](https://arxiv.org/pdf/1910.05512.pdf#page=8) the door opens only if another agent is close to a switch (dependence on the state of other agents). Similarly, agents can [move heavy objects](https://arxiv.org/pdf/2007.04979.pdf#page=6) only if they are execute coordinated actions (dependence on the action of other agents).
> 
> Hence, allowing the flexibility where a user/developer can code up the environment's transition dynamics as a function of the list of actions and list of current states of the agents, might be a general way forward for interesting coordinated tasks. Perhaps, food for thought for later edits, if not now.

It seems like these conditions are best handled as Task actions in the Habitat-API though. Maybe the thing we should really allow is the ability to call step without advancing physics?.

PR #954, opened 2020-12-03
--Load Scene Instance 1.0

jturner65:
## Motivation and Context
This PR was [replaced by this newer version](https://github.com/facebookresearch/habitat-sim/pull/1034)

## How Has This Been Tested

C++ and python tests 
## Types of changes

## Checklist


src/esp/bindings/MetadataMediatorBindings.cpp
```diff
@@ -0,0 +1,56 @@
...
+void initMetadataMediatorBindings(py::module& m) {
+  py::class_<MetadataMediator, MetadataMediator::ptr>(m, "MetadataMediator")
+      .def_property(
+          "active_dataset", &MetadataMediator::getActiveSceneDatasetName,
+          &MetadataMediator::setActiveSceneDatasetName,
+          R"(The currently active dataset being used.  Will attempt to load
+            configuration files specified if does not already exist.)")
+
+      /* --- Template Manager accessors --- */
+      .def("get_asset_template_manager",
```
Skylion007, line 27: Why aren't these values read_only properties?

src/esp/metadata/MetadataMediator.cpp
```diff
@@ -6,24 +6,58 @@
...
-                                     bool overwrite) {
+bool MetadataMediator::createPhysicsManagerAttributes(
+    const std::string& _physicsManagerAttributesPath) {
+  bool exists = physicsAttributesManager_->getObjectLibHasHandle(
+      _physicsManagerAttributesPath);
+  if (!exists) {
+    auto physAttrs = physicsAttributesManager_->createObject(
+        _physicsManagerAttributesPath, true);
+
+    if (nullptr == physAttrs) {
```
Skylion007, line 39: YODA condition

src/esp/metadata/MetadataMediator.cpp
```diff
@@ -37,20 +71,47 @@ bool MetadataMediator::createDataset(const std::string& sceneDatasetName,
       sceneDatasetAttributesManager_->createObject(sceneDatasetName, true);
   if (nullptr == datasetAttribs) {
```
Skylion007, line 72: Another yoda

src/esp/bindings/GfxBindings.cpp
```diff
@@ -151,6 +151,11 @@ void initGfxBindings(py::module& m) {
       .value("GLOBAL", LightPositionModel::GLOBAL)
       .value("OBJECT", LightPositionModel::OBJECT);
 
+  py::enum_<LightType>(
```
Skylion007, line 154: We really should have a way to extract these lights types for the GLBs. For both importing lit scenes as well as importing objects like lamps.

jturner65: I agree.  Currently this is only to facilitate the creation of LightInfos appropriately from LightInstanceAttributes, but I think eventually the LightInfos would benefit from having the desired type more explicitly specified.

src/esp/metadata/MetadataMediator.h
```diff
@@ -30,9 +30,8 @@ namespace metadata {
 class MetadataMediator {
  public:
   MetadataMediator(const sim::SimulatorConfiguration& cfg);
-  MetadataMediator(const std::string& _activeSceneDataset = "default",
-                   const std::string& _physicsManagerAttributesPath =
-                       ESP_DEFAULT_PHYSICS_CONFIG_REL_PATH);
+
+  MetadataMediator() : MetadataMediator(sim::SimulatorConfiguration{}) {}
```
jturner65, line 34: This will have the created MM be based on a default SimulatorConfiguration. This is to unify handling of MM both from Simulator and if instantiated without an existing sim.

TODO : one caveat to this could be that if someone wanted to cycle through sim close/reconfigure whilst retaining an MM, they would need to remember to call reconfigure with MM's simulator configuration, or else MM's will be replaced with whatever config they use.

PR #958, opened 2020-12-04
forceSeparateSemanticSceneGraph and loadAndCreateRenderAssetInstance

eundersander:
## Motivation and Context

This PR supports the upcoming render-replay feature.

### forceSeparateSemanticSceneGraph
A review of two "modes" for our renderer: depending on the stage, we either create a single scene graph (for RGB, depth, and semantic observations), or we create two scene graphs (loosely speaking, one is for RGB+depth and one is for semantic).

Previously, the "mode" decision was coupled to what stage you're loading (some stages have separate RGBD and Semantic models, others don't). For playback of a render replay, we won't load a stage at all (we'll use EMPTY_STAGE). However, I still need a way to select the mode (before loading any models). So, I've added SimulatorConfiguration.forceSeparateSemanticSceneGraph. 

### loadAndCreateRenderAssetInstance
This is pretty straightforward usage of our existing loadRenderAsset and createRenderAssetInstance. The only complication here is the selection of scene graph based on the creation flags. RenderAssetInstanceCreationInfo has isStatic, isRGBD, and isSemantic flags. Currently, we only support a few combinations of these flags:
- Non-static instances must have both isRGBD and isSemantic.
- For a static instance, it depends on the renderer mode (see above). In practice, this means a user has to be careful to use forceSeparateSemanticSceneGraph correctly before playing back a render replay. Here's an example: [replay_tutorial.py](https://github.com/eundersander/habitat-sim/blob/eundersander/serialize_for_render4/examples/tutorials/nb_python/replay_tutorial.py)
## How Has This Been Tested

I've added a unit test. I've also used this a lot in my in-progress render-replay branch, e.g. [replay_tutorial.py](https://github.com/eundersander/habitat-sim/blob/eundersander/serialize_for_render4/examples/tutorials/nb_python/replay_tutorial.py)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -443,6 +443,66 @@ esp::geo::CoordinateFrame ResourceManager::buildFrameFromAttributes(
   }
 }  // ResourceManager::buildCoordFrameFromAttribVals
 
+scene::SceneNode* ResourceManager::loadAndCreateRenderAssetInstance(
+    const AssetInfo& assetInfo,
+    const RenderAssetInstanceCreationInfo& creation,
+    esp::scene::SceneManager* sceneManagerPtr,
+    const std::vector<int>& activeSceneIDs) {
+  // We map isStatic, isSemantic, and isRGBD to a scene graph.
+  int sceneID = -1;
+  if (!creation.isStatic()) {
+    // Non-static instances must always get added to the RGBD scene graph, with
+    // nodeType==OBJECT, and they will be drawn for both RGBD and Semantic
+    // sensors.
+    if (!(creation.isSemantic() && creation.isRGBD())) {
+      LOG(WARNING) << "unsupported instance creation flags for asset ["
+                   << assetInfo.filepath << "]";
+      return nullptr;
+    }
+    sceneID = activeSceneIDs[0];
+  } else {
+    if (creation.isSemantic() && creation.isRGBD()) {
+      if (activeSceneIDs[1] != activeSceneIDs[0]) {
+        // Because we have a separate semantic scene graph, we can't support a
+        // static instance with both isSemantic and isRGBD.
+        LOG(WARNING)
+            << "unsupported instance creation flags for asset ["
+            << assetInfo.filepath
+            << "] with "
+               "SimulatorConfiguration::forceSeparateSemanticSceneGraph=true.";
+        return nullptr;
+      }
+      sceneID = activeSceneIDs[0];
+    } else {
+      if (activeSceneIDs[1] == activeSceneIDs[0]) {
+        // A separate semantic scene graph wasn't constructed, so we can't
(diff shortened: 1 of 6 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -182,7 +182,8 @@ bool ResourceManager::loadStage(
     const std::shared_ptr<physics::PhysicsManager>& _physicsManager,
     esp::scene::SceneManager* sceneManagerPtr,
     std::vector<int>& activeSceneIDs,
-    bool loadSemanticMesh) {
```
eundersander, line 185: I renamed loadSemanticMesh to createSemanticMesh just to match the var name in the header.

src/esp/bindings/SimBindings.cpp
```diff
@@ -45,6 +45,8 @@ void initSimBindings(py::module& m) {
                      &SimulatorConfiguration::sceneLightSetup)
       .def_readwrite("load_semantic_mesh",
                      &SimulatorConfiguration::loadSemanticMesh)
+      .def_readwrite("force_separate_semantic_scene_graph",
+                     &SimulatorConfiguration::forceSeparateSemanticSceneGraph)
```
jturner65, line 49: Maybe add a docstring explaining this?  Really need them for all of the config values, but perhaps this is a good place to start :)

src/esp/sim/SimulatorConfiguration.h
```diff
@@ -35,6 +35,12 @@ struct SimulatorConfiguration {
    * @brief Whether or not to load the semantic mesh
    */
   bool loadSemanticMesh = true;
+  /**
+   * Force creation of a separate semantic scene graph, even when no semantic
+   * mesh is loaded for the stage. This is required to support playback of any
+   * replay that includes a semantic-only render asset instance.
+   */
+  bool forceSeparateSemanticSceneGraph = false;
```
jturner65, line 43: What is the impact of this separation when no replay is desired?

jturner65 (approved):
A question and a comment, but otherwise LGTM.

src/esp/sim/SimulatorConfiguration.h
eundersander: If your stage has a semantic mesh (and it loads successfully), there's no impact.

Otherwise, the impact is that your stage's RGBD mesh only gets used for RGBD observations, instead of for all observations. This is probably never desirable, so my comment here and in the docstring just says to set it to false (unless configuring for gfx replay).

I realize this is pretty confusing. It's hopefully temporary. In the long term, I'd like to see us refactor RGBD and Semantic scene graphs. The existing code makes some rigid assumptions about stages and objects, and it's a little confusing (see Simulator.py Sensor.draw_observation). I'm trying to avoid the concepts of stages and objects in my gfx replay code (all things are just render asset instances).

PR #959, opened 2020-12-05
rapidjson serialization for some Habitat types

eundersander:
## Motivation and Context

I've added rapidjson serialization for some Habitat types. This supports the upcoming render-replay feature. It may also be useful for other serialization needs like Sim Replay.

Render-replay may eventually support other serialization formats, but json is good enough for the MVP. Some benefits of json: 1. human-readable and easy to debug, and 2. I can easily load it into Unity using C#.

Because I have a lot of user types to serialize, I've also added some rapidjson helper functions. See JsonAllTypes.h. These helpers are designed so that every builtin type and user type can be serialized/deserialized with `esp::io::AddMember`/`ReadMember`. This leads to uniform and readable serialization code. To achieve this, every complex user type defines `ToRJsonValue`/`FromRJsonValue`, and then template versions of `AddMember`/`ReadMember` will automatically use these.

There's a bit of duplication with existing serialization helpers in json.h (e.g. Magnum::Vector3). I was unable to use those helpers because they don't quite fit the pattern I'm using involving rapidjson::Values.

## How Has This Been Tested

Unit tests in IOTest. I've also used this a lot in my in-progress render-replay branch, e.g. [Recorder.cpp](https://github.com/eundersander/habitat-sim/blob/eundersander/serialize_for_render4/src/esp/gfx/replay/Recorder.cpp)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/replay/Keyframe.h
+++ b/src/esp/gfx/replay/Keyframe.h
@@ -1,0 +1,40 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include "esp/assets/Asset.h"
+#include "esp/assets/RenderAssetInstanceCreationInfo.h"
+
+namespace esp {
+namespace scene {
+class SceneNode;
+}
+namespace gfx {
+namespace replay {
+
+using RenderAssetInstanceKey = uint32_t;
+
+// serialization/replay-friendly Transform class
+struct Transform {
+  Magnum::Vector3 translation;
+  Magnum::Quaternion rotation;
+  bool operator==(const Transform& rhs) const {
+    return translation == rhs.translation && rotation == rhs.rotation;
+  }
+};
+
+struct RenderAssetInstanceState {
+  Transform absTransform;  // localToWorld
+  // note we currently only support semanticId per instance, not per drawable
+  int semanticId = -1;
+  // note we don't currently support runtime changes to lightSetupKey
+  bool operator==(const RenderAssetInstanceState& rhs) const {
+    return absTransform == rhs.absTransform && semanticId == rhs.semanticId;
+  }
+};
+
+}  // namespace replay
+}  // namespace gfx
+}  // namespace esp
--- a/src/esp/io/JsonEspTypes.h
+++ b/src/esp/io/JsonEspTypes.h
@@ -1,0 +1,144 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+/** @file
+ * @brief See JsonAllTypes.h. Don't include this header directly in user code.
+ */
+
+#include "JsonBuiltinTypes.h"
+#include "JsonMagnumTypes.h"
+
+#include "esp/assets/RenderAssetInstanceCreationInfo.h"
+#include "esp/core/esp.h"
+#include "esp/gfx/replay/Keyframe.h"
+
+namespace esp {
+namespace io {
+
+inline RJsonValue ToRJsonValue(const esp::vec3f& vec,
+                               RJsonAllocator& allocator) {
+  RJsonValue floatsArray(rapidjson::kArrayType);
+  for (int i = 0; i < vec.size(); i++) {
+    floatsArray.PushBack(vec.data()[i], allocator);
+  }
+  return floatsArray;
+}
+
+inline void FromRJsonValue(const RJsonValue& floatsArray, esp::vec3f& vec) {
+  // TODO: helper for objects that are simply arrays, with error-handling
+  ASSERT(floatsArray.Size() == vec.size());
+  for (int i = 0; i < vec.size(); i++) {
+    vec.data()[i] = floatsArray[i].GetFloat();
+  }
+}
+
+inline RJsonValue ToRJsonValue(const esp::geo::CoordinateFrame& frame,
+                               RJsonAllocator& allocator) {
+  RJsonValue obj(rapidjson::kObjectType);
+  AddMember(obj, "up", frame.up(), allocator);
+  AddMember(obj, "front", frame.front(), allocator);
+  AddMember(obj, "origin", frame.origin(), allocator);
+  return obj;
+}
+
+inline void FromRJsonValue(const RJsonValue& obj,
+                           esp::geo::CoordinateFrame& frame) {
+  esp::vec3f up;
+  esp::vec3f front;
+  esp::vec3f origin;
+  ReadMember(obj, "up", up);
+  ReadMember(obj, "front", front);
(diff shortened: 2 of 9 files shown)


src/esp/gfx/replay/Keyframe.h
```diff
@@ -0,0 +1,40 @@
...
+namespace scene {
+class SceneNode;
+}
+namespace gfx {
+namespace replay {
+
+using RenderAssetInstanceKey = uint32_t;
+
+// serialization/replay-friendly Transform class
+struct Transform {
```
eundersander, line 20: I don't use Transform and RenderAssetInstanceState in this PR, but they do showcase the serialization code. If you have questions/concerns about what exactly I'm serializing for the render-replay feature, I suggest waiting for that PR (coming soon).

src/esp/gfx/replay/Keyframe.h
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
eundersander, line 1: I'm introducing this replay subfolder because I have a lot of files incoming for render-replay.

src/esp/io/JsonEspTypes.h
```diff
@@ -0,0 +1,144 @@
...
+#include "JsonMagnumTypes.h"
+
+#include "esp/assets/RenderAssetInstanceCreationInfo.h"
+#include "esp/core/esp.h"
+#include "esp/gfx/replay/Keyframe.h"
+
+namespace esp {
+namespace io {
+
+inline RJsonValue ToRJsonValue(const esp::vec3f& vec,
```
eundersander, line 21: All the ToRJsonValue/FromJsonValue here are inline. This has nothing to do with runtime perf. My thinking is: (1) it's tedious to do declarations *and* definitions for all these small functions (and all the other types we'll add here), so let's just do inline definitions, and (2) these functions are all really small so it's not expensive to inline them.

I have at least one more complex object coming soon (later PR) that I won't inline because it feels like too much code to inline.

src/esp/io/JsonEspTypes.h
```diff
@@ -0,0 +1,144 @@
...
+  return floatsArray;
+}
+
+inline void FromRJsonValue(const RJsonValue& floatsArray, esp::vec3f& vec) {
+  // TODO: helper for objects that are simply arrays, with error-handling
+  ASSERT(floatsArray.Size() == vec.size());
+  for (int i = 0; i < vec.size(); i++) {
+    vec.data()[i] = floatsArray[i].GetFloat();
+  }
+}
```
jturner65, line 36: Many of the "from" functions already exist [here](https://github.com/facebookresearch/habitat-sim/blob/9c37189d0ac31d6cd5c918766d84b4a51128266e/src/esp/io/json.h#L54)

src/esp/io/JsonStlTypes.h
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
```
jturner65, line 5: Full header guards please, for consistency.

jturner65 (commented):
Please change pragmas to include-based header guards, for consistency.   I've marked the ones.

Also, there's partial duplication with [existing code](https://github.com/facebookresearch/habitat-sim/blob/9c37189d0ac31d6cd5c918766d84b4a51128266e/src/esp/io/json.h#L54) that already covers the functionality that many of the FromRJsonValue functions provide.  Perhaps the existing functionality can be modified to fit your needs?

src/esp/gfx/replay/Keyframe.h
eundersander: The render-replay feature is logically a subset of `gfx`. We will probably have a Sim Replay feature at some point; it would *not* use this Keyframe class. My placement of render-replay files under `gfx` will help distinguish those files from Sim Replay files.

eundersander:
>Also, there's partial duplication with existing code that already covers the functionality that many of the FromRJsonValue functions provide. Perhaps the existing functionality can be modified to fit your needs?

I did look at json.h. Comparing the existing helpers in json.h versus my needs and JsonAllTypes.h:
- jsonIntoVal doesn't support reading arbitrary types from json arrays.
- json.h helpers are read-only whereas I need read and write.
- Separating member lookup and object parsing as I've done with FromRJsonValue/ReadMember lets us handle the missing-member error case in just one place in code: JsonBuiltinTypes.hpp ReadMember (admittedly, this is still a todo for me).
- I'm not quite sure about the best way forward for error-handling, but I'm thinking that checking a return value on every ReadMember is too high a burden and I hope to avoid that.
- Just to be clear, json.h has much better error-handling than JsonAllTypes.h at the moment; I've put TODOs in my code where I still need to do this.

I don't mean to nitpick json.h; these are very subtle design differences. In the short term, I argue let's allow this duplication. In the long term, John and I or the whole team can debate these differences and settle on a consensus design.

jturner65:
> 
> 
> > Also, there's partial duplication with existing code that already covers the functionality that many of the FromRJsonValue functions provide. Perhaps the existing functionality can be modified to fit your needs?
> 
> I did look at json.h. Comparing the existing helpers in json.h versus my needs and JsonAllTypes.h:
> 
>     * jsonIntoVal doesn't support reading arbitrary types from json arrays.
> 
>     * json.h helpers are read-only whereas I need read and write.

All of which can be added.  What is currently in json.h is what was needed at the time.  Expansion to cover other types is trivial, accomplished via template argument or type inference.

> 
>     * Separating member lookup and object parsing as I've done with FromRJsonValue/ReadMember lets us handle the missing-member error case in just one place in code: JsonBuiltinTypes.hpp ReadMember (admittedly, this is still a todo for me).

The current json.h functions allow for the caller to manage the error, if necessary, which I think is much cleaner and more efficient.  Not every consumer is going to require every field be present in the source JSON document.

> 
>     * I'm not quite sure about the best way forward for error-handling, but I'm thinking that checking a return value on every ReadMember is too high a burden and I hope to avoid that.
> 
>     * Just to be clear, json.h has much better error-handling than JsonAllTypes.h at the moment; I've put TODOs in my code where I still need to do this.
> 
> 
> I don't mean to nitpick json.h; these are very subtle design differences. In the short term, I argue let's allow this duplication. In the long term, John and I or the whole team can debate these differences and settle on a consensus design.

I would always argue **against** introducing code duplication in a main branch in the interest of time.  My vote in cases like this would always be either consume (and of course expand as needed) the existing functionality or, if a real need for improvement/reworking exists that the existing code cannot be made to handle, then refactor the existing consumers to instead consume the new code.

I understand that you have a lot of code held up by this.  This also impacts Scene import.  Let's clear this up so you can make progress.

eundersander:
>I understand that you have a lot of code held up by this. This also impacts Scene import. Let's clear this up so you can make progress.
Yeah, let's continue this discussion offline.

eundersander:
Revised as https://github.com/facebookresearch/habitat-sim/pull/980

PR #980, opened 2020-12-17
rapidjson serialization and refactor

eundersander:
## Motivation and Context

This is a second stab at json helpers for writing, which has grown into a larger refactor. The earlier work is here: https://github.com/facebookresearch/habitat-sim/pull/959 . Per @jturner65 's [suggestion](https://github.com/facebookresearch/habitat-sim/pull/959#issuecomment-740161636), I've refactored to consolidate my new write helpers with the existing read helpers.

From JsonAllTypes.h:
These helpers are designed so that every builtin type and user type can be serialized/deserialized with esp::io::addMember/readMember. The leads to uniform and readable serialization code. To achieve this, every user type defines toJsonValue/fromJsonValue, and then template versions of addMember/readMember will automatically use these.

## How Has This Been Tested

See IOTest. I've also tested with the upcoming render-replay feature.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/io/JsonBuiltinTypes.h
+++ b/src/esp/io/JsonBuiltinTypes.h
@@ -16,6 +16,7 @@
 #include <rapidjson/document.h>
 
 #include <string>
+#include <typeinfo>
 
 namespace esp {
 namespace io {
--- a/src/esp/io/JsonBuiltinTypes.h
+++ b/src/esp/io/JsonBuiltinTypes.h
@@ -1,0 +1,285 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#ifndef ESP_IO_JSONBUILTINTYPES_H_
+#define ESP_IO_JSONBUILTINTYPES_H_
+
+/** @file
+ * @brief See JsonAllTypes.h. Don't include this header directly in user code.
+ */
+
+#include "esp/core/logging.h"
+
+#include <Corrade/Utility/Macros.h>
+
+#include <rapidjson/document.h>
+
+#include <string>
+
+namespace esp {
+namespace io {
+
+// Make these types easier to read/type. We'll use them everywhere.
+typedef rapidjson::GenericValue<rapidjson::UTF8<> > JsonGenericValue;
+typedef rapidjson::MemoryPoolAllocator<> JsonAllocator;
+
+// template addMember/readMember to match any type. These are declared here,
+// before all toJsonValue/fromJsonValue definitions, and they are defined
+// later in JsonBuiltinTypes.hpp. The quirky ordering is to avoid some compile
+// errors.
+
+template <typename T>
+void addMember(JsonGenericValue& value,
+               rapidjson::GenericStringRef<char> name,
+               const T& obj,
+               JsonAllocator& allocator);
+
+template <typename T>
+bool readMember(const JsonGenericValue& value, const char* name, T& x);
+
+/**
+ * @brief Fallback implementation for fromJsonValue to produce a runtime error
+ * for types that haven't implemented fromJsonValue.
+ */
+template <typename T>
+bool fromJsonValue(CORRADE_UNUSED const JsonGenericValue& obj,
+                   CORRADE_UNUSED T& val) {
+  // If you've already implemented fromJsonValue for your type and you're still
+  // hitting this, the underlying issue might be the ordering among json helper
+  // definitions.
+  LOG(ERROR) << "Unsupported type. Aborting. You need to implement "
+                "fromJsonValue for typeid(T).name() = "
+             << typeid(T).name() << ".";
+  return false;
+}
+
+/**
+ * @brief Fallback implementation for toJsonValue to produce a runtime error for
+ * types that haven't implemented toJsonValue.
+ */
+template <typename T>
+JsonGenericValue toJsonValue(const T&, JsonAllocator&) {
+  // If you've already implemented toJsonValue for your type and you're still
+  // hitting this, the underlying issue might be the ordering among json helper
+  // definitions.
+  LOG(ERROR) << "Unsupported type. Aborting. You need to implement toJsonValue "
+                "for typeid(T).name() = "
+             << typeid(T).name() << ".";
+  return JsonGenericValue(rapidjson::kObjectType);
+}
+
+// toJsonValue wrappers for the 7 rapidjson builtin types. A JsonGenericValue
+// can be directly constructed from the builtin types.
+
+inline JsonGenericValue toJsonValue(bool x, JsonAllocator&) {
+  return JsonGenericValue(x);
+}
+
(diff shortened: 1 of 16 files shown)


src/esp/io/JsonBuiltinTypes.h
```diff
@@ -0,0 +1,285 @@
...
+
+/**
+ * @brief Populate passed @p val with value. Returns whether successfully
+ * populated, or not. Logs an error if inappropriate type.
+ *
+ * @param obj string tag to look for in json doc
+ * @param val destination value to be populated
+ * @return whether successful or not
+ */
+inline bool fromJsonValue(const JsonGenericValue& obj, bool& val) {
```
eundersander, line 113: The implementation and documentation for these was ported from jsonToVal functions in json.h.

src/esp/io/json.h
```diff
@@ -39,285 +40,6 @@ std::string jsonToString(const JsonDocument& d);
...
- * getters.
- *
- * @tparam T type of value to be populated.
- * @param d json document to parse
- * @param tag string tag to look for in json doc
- * @param val destination value to be populated
- * @return whether successful or not
- */
-template <typename T>
-bool jsonIntoVal(CORRADE_UNUSED const JsonGenericValue& d,
```
eundersander, line 54: See new version at https://github.com/facebookresearch/habitat-sim/pull/980/files#diff-f61fe6b9b46818540dbea3ad5fa73bf56369af872af0b12c14999ec1ff8a293cR46

src/esp/io/json.h
```diff
@@ -39,285 +40,6 @@ std::string jsonToString(const JsonDocument& d);
...
- * is found and successfully populated, or not. Logs an error if tag is found
- * but is inappropriate type.
- *
- * @param d json document to parse
- * @param tag string tag to look for in json doc
- * @param val destination value to be populated
- * @return whether successful or not
- */
-template <>
-inline bool jsonIntoVal(const JsonGenericValue& d,
```
eundersander, line 74: See ported version at https://github.com/facebookresearch/habitat-sim/pull/980/files#diff-f61fe6b9b46818540dbea3ad5fa73bf56369af872af0b12c14999ec1ff8a293cR215

src/esp/io/json.h
```diff
@@ -39,285 +40,6 @@ std::string jsonToString(const JsonDocument& d);
...
- * populated, or not. Logs an error if tag is found but is inappropriate type.
- *
- * @param d json document to parse
- * @param tag string tag to look for in json doc
- * @param val destination value to be populated
- * @return whether successful or not
- */
-
-template <>
-inline bool jsonIntoVal(const JsonGenericValue& d,
```
eundersander, line 226: See ported version at https://github.com/facebookresearch/habitat-sim/pull/980/files#diff-b2a6b781121dc31a434d1c5a0dd7ffbc1caa2d1e386168987d8bccabcffe3f80R38

src/esp/io/json.h
```diff
@@ -39,285 +40,6 @@ std::string jsonToString(const JsonDocument& d);
...
- * configured
- *
- * @param d json document to parse
- * @param tag string tag to look for in json doc
- * @param val destination std::map to be populated
- * @return whether successful or not
- */
-
-template <>
-inline bool jsonIntoVal(const JsonGenericValue& d,
```
eundersander, line 296: moved to JsonStlTypes.h https://github.com/facebookresearch/habitat-sim/pull/980/files#diff-eaa098e6aba025291162ab384a70517874de6a4e24f2eafb550c7f4bdee477f0R56

src/esp/io/JsonEspTypes.h
```diff
@@ -0,0 +1,161 @@
...
+  for (int i = 0; i < vec.size(); i++) {
+    floatsArray.PushBack(vec.data()[i], allocator);
+  }
+  return floatsArray;
+}
+
+inline bool fromJsonValue(const JsonGenericValue& floatsArray,
+                          esp::vec3f& vec) {
+  // TODO: helper for objects that are simply arrays, with error-handling
+  ASSERT(floatsArray.Size() == vec.size());
```
jturner65, line 34: Do we want to fail an assertion if array size is not correct? Or could we just return false with a message?  (That's what I would do before in the old jsonIntoVal functions.)

src/esp/io/JsonEspTypes.h
```diff
@@ -0,0 +1,161 @@
...
+inline bool fromJsonValue(const JsonGenericValue& obj,
+                          esp::geo::CoordinateFrame& frame) {
+  esp::vec3f up;
+  esp::vec3f front;
+  esp::vec3f origin;
+  readMember(obj, "up", up);
+  readMember(obj, "front", front);
+  readMember(obj, "origin", origin);
+  frame = esp::geo::CoordinateFrame(up, front, origin);
+  return true;
```
jturner65, line 59: Could this one return an accurate bool code based on the results of each readmember? Would be helpful for some places that I would use this in managed containers. perhaps something like

```
<...>
bool success = true;
success &= readMember(obj, "up", up);
success &= readMember(obj, "front", front);
success &= readMember(obj, "origin", origin);
<...>
return success;
```

src/esp/io/JsonEspTypes.h
```diff
@@ -0,0 +1,161 @@
...
+  addMember(obj, "rotation", x.rotation, allocator);
+  return obj;
+}
+
+inline bool fromJsonValue(const JsonGenericValue& obj,
+                          esp::gfx::replay::Transform& x) {
+  readMember(obj, "translation", x.translation);
+  readMember(obj, "rotation", x.rotation);
+  return true;
+}
```
jturner65, line 134: Here as well, I would like to have an accurate return value. (I could see a useful application of these transformations in the scene instance attributes, but I would want to have an accurate state of the read value).  So something like this :

```
bool success = true;
success &= readMember(obj, "translation", x.translation);
success &= readMember(obj, "rotation", x.rotation);
return success;
```

jturner65 (approved):
A comment and 2 requests.  Otherwise, looks great to me :).

src/esp/io/JsonEspTypes.h
eundersander: Sure.

>I could see a useful application of these transformations in the scene instance attributes,
Beware, this class in in the replay namespace for a reason and might change at the whims of what replay needs, e.g. adding scale, or switching from quaternion to Matrix4. I don't recommend for general usage outside replay.

src/esp/io/JsonEspTypes.h
eundersander: I'll change to make this work similar to Magnum::Vector3.

src/esp/io/JsonEspTypes.h
jturner65: I see, that makes sense.  Thanks for the clarification.

jturner65: Might be nice to have a generic transformation read/write path, but that doesn't have to be in this PR. :).

PR #983, opened 2020-12-17
render replay integration and python bindings

eundersander:
## Motivation and Context

This PR integrates the render-replay Recorder and Player into Simulator and ResourceManager and sets up python bindings to use them. Recorder and Player are already merged and unit-tested so this PR is fairly small and just does the hookup/integration.

~Note file reading/writing is still disabled (waiting on https://github.com/facebookresearch/habitat-sim/pull/980 ) so this is more like a stub implementation. I may wait for https://github.com/facebookresearch/habitat-sim/pull/980 to merge before merging this, but I'd like to get this under review now.~ Update: https://github.com/facebookresearch/habitat-sim/pull/980 has merged.

I have a working replay tutorial which I was planning to merge as a separate PR (or I can add it here if requested). Anyway it's [here](https://github.com/eundersander/habitat-sim/blob/eundersander/replay_tutorial/examples/tutorials/nb_python/replay_tutorial.py) if you want to check out usage of the python bindings.

## How Has This Been Tested

This integration was tested in upcoming [replay tutorial](https://github.com/eundersander/habitat-sim/blob/eundersander/replay_tutorial/examples/tutorials/nb_python/replay_tutorial.py), including testing the exception-throwing in the bindings.

Note that Player and Recorder are already individually unit-tested in GfxReplayTest.cpp (already in master). The error-handling changes to Player here include new unit tests.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/GfxReplayBindings.cpp
+++ b/src/esp/bindings/GfxReplayBindings.cpp
@@ -1,0 +1,89 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "esp/bindings/bindings.h"
+
+#include <Magnum/PythonBindings.h>
+#include <Magnum/SceneGraph/PythonBindings.h>
+
+#include "esp/gfx/replay/Player.h"
+#include "esp/gfx/replay/ReplayManager.h"
+
+namespace py = pybind11;
+using py::literals::operator""_a;
+
+namespace esp {
+namespace gfx {
+namespace replay {
+
+void initGfxReplayBindings(py::module& m) {
+  py::class_<Player, Player::ptr>(m, "Player")
+      .def("get_num_keyframes", &Player::getNumKeyframes,
+           R"(Get the currently-set keyframe, or -1 if no keyframe is set.)")
+
+      .def(
+          "set_keyframe_index", &Player::setKeyframeIndex,
+          R"(Set a keyframe by index, or pass -1 to clear the currently-set keyframe.)")
+
+      .def("get_keyframe_index", &Player::getKeyframeIndex,
+           R"(Get the number of keyframes read from file.)")
+
+      .def(
+          "get_user_transform",
+          [](Player& self, const std::string& name) {
+            Magnum::Vector3 translation;
+            Magnum::Quaternion rotation;
+            bool found = self.getUserTransform(name, &translation, &rotation);
+            return found ? py::make_tuple(translation, rotation)
+                         : py::make_tuple(nullptr, nullptr);
+          },
+          R"(todo)");
+
+  py::class_<ReplayManager, ReplayManager::ptr>(m, "ReplayManager")
+      .def(
+          "save_keyframe",
+          [](ReplayManager& self) {
+            if (!self.getRecorder()) {
+              LOG(ERROR) << "save_keyframe: not enabled. See "
+                            "SimulatorConfiguration::enableGfxReplaySave.";
+              return;
+            }
+            self.getRecorder()->saveKeyframe();
+          },
+          R"(Save a render keyframe; a render keyframe can be loaded later and used to draw observations.)")
+
+      .def(
+          "add_user_transform_to_keyframe",
+          [](ReplayManager& self, const std::string& name,
+             const Magnum::Vector3& translation,
+             const Magnum::Quaternion& rotation) {
+            if (!self.getRecorder()) {
+              LOG(ERROR) << "add_user_transform_to_keyframe: not enabled. See "
+                            "SimulatorConfiguration::enableGfxReplaySave.";
+              return;
+            }
+            self.getRecorder()->addUserTransformToKeyframe(name, translation,
+                                                           rotation);
+          },
+          R"(Add a user transform to the current render keyframe; it will get stored with the keyframe and will be available later upon loading the keyframe)")
+
+      .def(
+          "write_saved_keyframes_to_file",
+          [](ReplayManager& self, const std::string& filepath) {
+            if (!self.getRecorder()) {
+              LOG(ERROR) << "write_saved_keyframes_to_file: not enabled. See "
+                            "SimulatorConfiguration::enableGfxReplaySave.";
+              return;
+            }
+            self.getRecorder()->writeSavedKeyframesToFile(filepath);
+          },
+          R"(Write all saved keyframes to a file, then discard the keyframes.)")
+
+      .def("read_keyframes_from_file", &ReplayManager::readKeyframesFromFile,
+           R"(Create a Player object from a replay file.)");
+}
+
+}  // namespace replay
+}  // namespace gfx
+}  // namespace esp
--- a/src/esp/bindings/GfxReplayBindings.cpp
+++ b/src/esp/bindings/GfxReplayBindings.cpp
@@ -35,19 +35,20 @@ void initGfxReplayBindings(py::module& m) {
             Magnum::Vector3 translation;
             Magnum::Quaternion rotation;
             bool found = self.getUserTransform(name, &translation, &rotation);
-            return found ? py::make_tuple(translation, rotation)
-                         : py::make_tuple(nullptr, nullptr);
+            return found ? py::cast<py::object>(
+                               py::make_tuple(translation, rotation))
+                         : py::cast<py::object>(Py_None);
           },
-          R"(todo)");
+          R"(Get a previously-added user transform. See also ReplayManager.add_user_transform_to_keyframe.)");
 
   py::class_<ReplayManager, ReplayManager::ptr>(m, "ReplayManager")
       .def(
           "save_keyframe",
           [](ReplayManager& self) {
             if (!self.getRecorder()) {
-              LOG(ERROR) << "save_keyframe: not enabled. See "
-                            "SimulatorConfiguration::enableGfxReplaySave.";
-              return;
+              throw std::runtime_error(
+                  "replay save not enabled. See "
+                  "SimulatorConfiguration.enable_gfx_replay_save.");
             }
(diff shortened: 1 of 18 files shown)


src/esp/bindings/SimBindings.cpp
```diff
@@ -72,6 +77,9 @@ void initSimBindings(py::module& m) {
             Not available for all datasets
             )")
       .def_property_readonly("renderer", &Simulator::getRenderer)
+      .def_property_readonly(
+          "gfx_replay", &Simulator::getGfxReplayManager,
```
mosra, line 81:
```suggestion
          "gfx_replay_manager", &Simulator::getGfxReplayManager,
```

to be consistent with the C++ method?

src/esp/bindings/GfxReplayBindings.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  py::class_<ReplayManager, ReplayManager::ptr>(m, "ReplayManager")
+      .def(
+          "save_keyframe",
+          [](ReplayManager& self) {
+            if (!self.getRecorder()) {
+              LOG(ERROR) << "save_keyframe: not enabled. See "
+                            "SimulatorConfiguration::enableGfxReplaySave.";
+              return;
+            }
```
mosra, line 51: shouldn't all these `LOG(ERROR)` be rather exceptions instead of printing an error and silently doing nothing? i feel that's more pythonic

src/esp/bindings/GfxReplayBindings.cpp
```diff
@@ -0,0 +1,89 @@
...
+           R"(Get the number of keyframes read from file.)")
+
+      .def(
+          "get_user_transform",
+          [](Player& self, const std::string& name) {
+            Magnum::Vector3 translation;
+            Magnum::Quaternion rotation;
+            bool found = self.getUserTransform(name, &translation, &rotation);
+            return found ? py::make_tuple(translation, rotation)
+                         : py::make_tuple(nullptr, nullptr);
```
mosra, line 39: Hmm, wouldn't it be better for `getUserTransform()` to return `Containers::Optional<std::pair<Vector3, Quaternion>>`? It would be both less code (and less checks) in its implementation as well as in any call site:

```cpp
auto found = self.getUserTransform(name);
if(found) return *found; // std::pair converts to a 2-tuple automatically
return py::none;
```

And returning `None` in this case feels more pythonic too, since that works properly with `if` -- however you may need to change the return type to `py::object` and then do some additional gymnastics with `py::cast` to make the above work. Or raise an exception in that case as well? Not sure what's the intended semantics for when this "fails".

Alternatively I could look into exposing Optional in a way that makes pybind implicitly understand it as a Python optional. Or (assuming Habitat is C++17) you could convert it to a `std::optional` and [let pybind do the rest](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#c-17-library-containers) (you need [`#include <Corrade/Containers/OptionalStl.h>`](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1Optional.html#Containers-Optional-stl) and an explicit conversion for this).

mosra (commented):
More of a "pybind usage" review than actually comments relevant to the render replay functionality, as I still need to catch up on all that. 

I hope some of this is useful :)

src/esp/bindings/GfxReplayBindings.cpp
eundersander: Sounds good, I'll investigate.

src/esp/bindings/GfxReplayBindings.cpp
eundersander: >Or raise an exception in that case as well? Not sure what's the intended semantics for when this "fails".
This should behave more like a find. It is non exceptional to return None. I will make a change here along the lines of your suggestion.

src/esp/bindings/GfxReplayBindings.cpp
mosra: Pybind should have most of the usual Python exceptions exposed, so it should be as easy as `throw`ing one of them: https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

src/esp/bindings/GfxReplayBindings.cpp
eundersander: Oops, I misread some of this.

I'm going to change the python interface as you suggest. I'll return None instead of a pair of None.

I see you were also suggesting a change to getUserTransform to return an optional pair. I'm skeptical. For the common case of a found transform, I think the compiler will have a harder time of avoiding copying at the callsite if it has to construct an optional. I'm thinking about future usage in perf-critical C++ code, not this usage here in the python binding.

mosra: > I think the compiler will have a harder time of avoiding copying at the callsite if it has to construct an optional.

It will either copy 32 bytes (the contents of the Optional are trivially copyable, and if ~~you~~ a profiler really insists I could make the whole Optional trivially copyable for trivial types) or zero-init one byte, whereas with the out-pointer it has to do two separate copies since it can't know the pointer locations beforehand... And all that is practically inlined, which has minimal overhead compared to the function call inside `getUserTransform()`. Anyway, I think that's fine, there's *way* worse crimes commited elsewhere in the code, especially all the shared pointers ;)

eundersander: >Anyway, I think that's fine, there's way worse crimes commited elsewhere in the code, especially all the shared pointers ;)

I'll leave the code as-is for now. I'm curious to discuss this when time allows, maybe a Thursday meeting in Jan... e.g., what is the best practice here? What if the optional thing here were a potentially large std::vector?

src/esp/bindings/GfxReplayBindings.cpp
eundersander: I throw std::runtime_error, which pybind turns into a python RuntimeError.

src/esp/bindings/GfxReplayBindings.cpp
mosra: I don't know the exact C++17 rules because I'm still ~~living in the cave~~ mainly focusing on C++11, but here's the thing: https://en.cppreference.com/w/cpp/language/copy_elision

AFAIK, in C++17 the compiler is *required* to construct the returned value in-place at the call site instead of first constructing it inside the function and then copy-constructing it to the final place it's returned to. In C++11 it's not required to do that, but is *allowed* to do such optimization. The spec puts some extra restrictions depending on the context, but usually it should work. 

Then, with this rule in place, construction of an empty `Optional` is just clearing a single byte to zero, no matter how huge the payload is, so that's not too far from returning a `bool`, assuming it gets all inlined in a release build. Copying of such (empty) container is the same -- it only clears a single byte and leaves the rest untouched. Compared to `Pointer` / `std::unique_ptr` that has a similar high-level semantic this doesn't involve any heap allocation, just an in-place construction.

eundersander: I'm confident in the compiler avoiding copying for return values. I love how I can return big complex objects in by value nowadays.

What I'm worried about here is, imagine if the caller code wants to store the translation vector and/or the rotation somewhere in particular like an array of vectors. With my existing signature, this can be done without copying. If returning an optional, I don't see how the copy from the optional to the array could be avoided.

mosra: Ah, yep, that can't be avoided if copying them to separate places. 

OTOH I think the function call overhead is *still* much bigger than the cost of extra copy, except if LTO inlines the whole thing (which you can't guarantee). If you'd want to really optimize for that case and mitigate the call overhead, you'd need some batch variant that gives back N values at the same time.

eundersander: We could go back and forth on this. I appreciate your thoughts. I'm not fully convinced that returning an optional by value is the new best practice for C++. But I'll continue to think about it.

Are you comfortable approving this PR as-is?

mosra: I don't feel comfortable being the only reviewer, especially because I didn't see the other 90% of the code :)

Can this wait for Monday for others to accept? If not, consider it approved from my side.

bigbike (approved):

src/esp/bindings/GfxReplayBindings.cpp
```diff
@@ -0,0 +1,90 @@
...
+namespace py = pybind11;
+using py::literals::operator""_a;
+
+namespace esp {
+namespace gfx {
+namespace replay {
+
+void initGfxReplayBindings(py::module& m) {
+  py::class_<Player, Player::ptr>(m, "Player")
+      .def("get_num_keyframes", &Player::getNumKeyframes,
```
erikwijmans, line 22: Pure getters and setters like "get_num_keyframes" and "get/set_keyframe_index" should be exposed with `def_property` and `def_property_readonly` as this better fits python

Skylion007: Agreed!

eundersander: Just want to double-check here: in this case, set_keyframe_index is a very heavyweight operation--it might include creating render asset instances as necessary to restore the scene graph to the desired state.

Given that, do you still think a property is appropriate?

I want to discourage code like this:

```
player.keyframe_index = x
if some_condition:
  player.keyframe_index += 1
```  
I would prefer:
```
keyframe_index = x
if some_condition:
  keyframe_index += 1
player.set_keyframe_index(keyframe_index)
```

erikwijmans: Ah, good point.  Heavy weight properties can certainly be counter intuitive.

eundersander: I went ahead and merged this as-is but I realize I should change get_num_keyframes to be a read-only property. I'll make a separate PR soon.

PR #984, opened 2020-12-18
Replace WebVR integration with WebXR

BlaiseRitchie:
## Motivation and Context

The existing WebVR integration doesn't work with newer browsers

## How Has This Been Tested

Tested locally with a Rift S on Chrome 87.0.4280.88

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist

- [ x ] I have read the **CONTRIBUTING** document.
- [ x ] I have completed my CLA (see **CONTRIBUTING**)

### DIFF
--- a/src/esp/bindings_js/index.js
+++ b/src/esp/bindings_js/index.js
@@ -44,12 +44,13 @@ Module.preRun.push(() => {
   }
 });
 
-Module.onRuntimeInitialized = () => {
+Module.onRuntimeInitialized = async function() {
   console.log("hsim_bindings initialized");
   let demo;
   if (window.vrEnabled) {
-    if (navigator && navigator.getVRDisplays) {
-      console.log("Web VR is supported");
+    const supported = await navigator.xr.isSessionSupported("immersive-vr");
+    if (supported) {
+      console.log("WebXR is supported");
       demo = new VRDemo();
     }
   } else if (window.viewerEnabled) {
--- a/src/esp/bindings_js/modules/vr_demo.js
+++ b/src/esp/bindings_js/modules/vr_demo.js
@@ -2,19 +2,23 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-/* global VRFrameData */
+/* global XRWebGLLayer */
 import WebDemo from "./web_demo";
-import { defaultAgentConfig, defaultEpisode } from "./defaults";
+import { defaultAgentConfig } from "./defaults";
+
+const BUTTON_ID = "vr_button";
+const pointToArray = p => [p.x, p.y, p.z, p.w];
 
 class VRDemo extends WebDemo {
-  normalSceneFrame;
-  vrSceneFrame;
   fpsElement;
   lastPaintTime;
-  frameData = new VRFrameData();
-  previousRotation = false;
-  previousPosition = false;
-  currentVrDisplay = false;
+
+  webXRSession = null;
+  xrReferenceSpace = null;
+  gl = null;
+  tex = null;
+  inds = null;
+
   fps = 0;
   skipFrames = 60;
   currentFramesSkipped = 0;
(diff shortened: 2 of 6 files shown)


src/esp/bindings_js/index.js
```diff
@@ -44,12 +44,13 @@ Module.preRun.push(() => {
...
 });
 
-Module.onRuntimeInitialized = () => {
+Module.onRuntimeInitialized = async function() {
   console.log("hsim_bindings initialized");
   let demo;
   if (window.vrEnabled) {
-    if (navigator && navigator.getVRDisplays) {
-      console.log("Web VR is supported");
+    let supported = await navigator.xr.isSessionSupported("immersive-vr");
```
Skylion007, line 51: Nit: shouldn't this be `const`.

src/esp/bindings_js/modules/web_demo.js
```diff
@@ -50,19 +50,30 @@ class WebDemo {
...
     const sensorConfigs = [
       {
         uuid: "rgb",
+        sensorType: Module.SensorType.COLOR
+      },
+      {
+        uuid: "left_eye",
         sensorType: Module.SensorType.COLOR,
-        resolution: [480, 640]
+        resolution: [1024, 1024]
```
Skylion007, line 67: Hold on. The position of these two sensors should be offset by the VR headsets (interpupilary distance (IPD) / 2) (or the average IPD for the population if not provided. And doesn't the rgb sensor become redundant with this setup. No reason to waste render cycles on it.

src/esp/bindings_js/modules/vr_demo.js
```diff
@@ -25,96 +28,259 @@ class VRDemo extends WebDemo {
...
+
     const agent = this.simenv.sim.getAgent(this.simenv.selectedAgentId);
-    let prevAgentState = this.simenv.createAgentState({});
-    agent.getState(prevAgentState);
 
-    const simRotation = prevAgentState.rotation.slice();
-    const simPosition = prevAgentState.position.slice();
+    const VIEW_SENSORS = ["left_eye", "right_eye"];
+    for (var iView = 0; iView < pose.views.length; ++iView) {
+      let view = pose.views[iView];
```
Skylion007, line 266: NIT: Some of these should const as well.

src/esp/bindings_js/modules/web_demo.js
BlaiseRitchie: I originally had a position with IPD set here, but it just gets overridden by the transforms coming from WebXR every frame anyway.

Skylion007: See the stereo example agent: https://github.com/facebookresearch/habitat-sim/blob/b6faca590ccb52a672074c0b14b8d7dd47b4d3de/examples/tutorials/stereo_agent.py

src/esp/bindings_js/modules/vr_demo.js
```diff
@@ -25,96 +28,259 @@ class VRDemo extends WebDemo {
...
-      poseRotation.forEach((item, index) => {
-        simRotation[index] += item - this.previousRotation[index];
-      });
-      posePosition.forEach((item, index) => {
-        simPosition[index] += item - this.previousPosition[index];
-      });
-      this.previousRotation = poseRotation.slice();
-      this.previousPosition = posePosition.slice();
+      const pointToArray = p => [p.x, p.y, p.z, p.w];
+      sensor.setLocalTransform(
```
Skylion007, line 273: This only needs to be set once, right? Not every time the sensors are drawn? I guess it depends on what the agent "body" is and whether you are using collisions? Because it seems like it would be easy to just update the agent's overall transform if it's just the head.

Skylion007 (changes requested):
Most of this looks good except for the stereo right eye / left eye sensor should be initialized to a good default value. 

Also would it be possible to add a quick test like we have for the default viewer?
https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/bindings_js/tests/viewer_demo.visual.test.js Don't have to do a pixelwise sensor compare, just want to make sure it doesn't throw any errors.

src/esp/bindings_js/modules/web_demo.js
Skylion007: Ah I see, I would still have it set as a default.

src/esp/bindings_js/modules/vr_demo.js
```diff
@@ -25,96 +28,259 @@ class VRDemo extends WebDemo {
...
 
-      poseRotation.forEach((item, index) => {
-        simRotation[index] += item - this.previousRotation[index];
-      });
-      posePosition.forEach((item, index) => {
-        simPosition[index] += item - this.previousPosition[index];
-      });
-      this.previousRotation = poseRotation.slice();
-      this.previousPosition = posePosition.slice();
+      const pointToArray = p => [p.x, p.y, p.z, p.w];
```
Skylion007, line 272: Does this lambda need to be regenerated twi every frame. Might want to move it out of this hot code.

src/esp/bindings_js/modules/vr_demo.js
BlaiseRitchie: I didn't change the agent transform because it needs to be on the ground to interact correctly with the environment navmesh. And these transforms change every frame as the viewer's head pose moves.

eundersander:
>Also would it be possible to add a quick test like we have for the default viewer?
https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/bindings_js/tests/viewer_demo.visual.test.js Don't have to do a pixelwise sensor compare, just want to make sure it doesn't throw any errors.

This seems like a reasonable suggestion and @BlaiseRitchie you're welcome to take a stab. But @Skylion007 I suggest we allow this to merge without the test (i.e., don't block the PR on this). I don't think the older VR code had a test. I don't want to burden contributors and discourage them by overloading them with testing requests. Someone on Hab Sim Infra can add the test later.

src/esp/bindings_js/webvr.html
```diff
@@ -21,7 +21,13 @@ <h1>Habitat</h1>
     <div id="status">Initialization...</div>
     <div id="status-description"></div>
   </div>
+
+
+  <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/webxr-polyfill@latest/build/webxr-polyfill.js"></script>
```
Skylion007, line 26: We should double check this doesn't cause a CORS error on our website. Thoughts @eundersander . Worst case we could always bundle it.

eundersander: Seems fine to me.

I put a question to our Slack dev channel about how our website gets updated, e.g. https://aihabitat.org/demo/ .

But I don't think @BlaiseRitchie 's changes here will get automatically pushed to our website. If I'm correct, then your concern about CORS error is more about if/when anyone tries to push this app to a website (right?), so I guess we could defer the question and testing until then.

Linking to this jsdelivr URL seems to be good practice according to https://github.com/immersive-web/webxr-polyfill .

Skylion007 (approved):
Looks good to me.

PR #987, opened 2020-12-19
Fix bugs in cuda noise model

erikwijmans:
## Motivation and Context

The CUDA and CPU noise model implementations were producing pretty different results, this PR fixes that.

## How Has This Been Tested

With the new test (it use to fail before I fixed the bugs).

## Types of changes

### DIFF
--- a/src/esp/sensor/RedwoodNoiseModel.cu
+++ b/src/esp/sensor/RedwoodNoiseModel.cu
@@ -5,7 +5,6 @@
 #include "RedwoodNoiseModel.cuh"
 
 #include <algorithm>
-#include <iostream>
 
 #include <cuda_runtime.h>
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

src/esp/sensor/RedwoodNoiseModel.cu
```diff
@@ -5,13 +5,14 @@
 #include "RedwoodNoiseModel.cuh"
 
 #include <algorithm>
+#include <iostream>
```
Skylion007, line 8: NIT: Is this left from debugging?

erikwijmans: Yep, thanks!

PR #990, opened 2020-12-21
--Create MetadataMediator python bindings.

jturner65:
## Motivation and Context
This PR introduces python bindings to facilitate access to Metadata Mediator and the functionality it provides (i.e. access to attributes managers) independent of chaining through Simulator.  Simulator-specific access to attributes managers will probably remain, but the MM is intended to be able to persist through iterations of Simulator, so providing direct python access is essential, and will be expanded as needed.   Pytests to follow in future PR (as part of scene import). 

## How Has This Been Tested
existing c++ and python tests

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/MetadataMediatorBindings.cpp
+++ b/src/esp/bindings/MetadataMediatorBindings.cpp
@@ -1,0 +1,60 @@
+
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "esp/bindings/bindings.h"
+
+#include <Magnum/Magnum.h>
+#include <Magnum/PythonBindings.h>
+
+#include "esp/metadata/MetadataMediator.h"
+namespace py = pybind11;
+using py::literals::operator""_a;
+
+namespace esp {
+namespace metadata {
+
+void initMetadataMediatorBindings(py::module& m) {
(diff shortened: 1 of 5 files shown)


Skylion007 (approved):

src/esp/bindings/MetadataMediatorBindings.cpp
```diff
@@ -0,0 +1,60 @@
...
+      .def_property(
+          "active_dataset", &MetadataMediator::getActiveSceneDatasetName,
+          &MetadataMediator::setActiveSceneDatasetName,
+          R"(The currently active dataset being used.  Will attempt to load
+            configuration files specified if does not already exist.)")
+
+      /* --- Template Manager accessors --- */
+      .def_property_readonly(
+          "asset_template_manager",
+          &MetadataMediator::getAssetAttributesManager,
```
eundersander, line 29: What's the reasoning on these names not quite matching? I would naively expect a property called "asset_attributes_manager" since that's the name of the C++ function.

jturner65: It was thought that the term "Template" was a  more favorable term than "Attributes" for the python code, since it infers that these constructs are used to provide a template for construction.  We did not use the term "Template" in the c++ implementations because of the language overloads inherent in the term.  

I agree, from an Infra dev perspective, it is somewhat confusing, but our rationale during the original design was that we felt that most python users won't consume the c++ directly and so we don't have to constrain the naming as tightly.

eundersander (approved):
LGTM

PR #994, opened 2020-12-23
[Clang-Tidy] Tighten clang-tidy warnings into errors for bugprone.

Skylion007:
## Motivation and Context

* There were a few annoying RAII warnings and bugprone-branch clones that were being emitted as warnings instead of errors. It seems like the RAII warnings are false positives though since Caffe2 has similar nolints. False positive occurs when a module is constructed with nodefs, which usually is an error, but in these few cases is not. When it is not an error, nolint comments should be explicit.
* Now the main warning left in the code base is implicit narrowing conversions.

## How Has This Been Tested
* Locally and on CircleCI

## Types of changes

-
## Checklist

### DIFF
--- a/src/esp/bindings/AttributesManagersBindings.cpp
+++ b/src/esp/bindings/AttributesManagersBindings.cpp
@@ -274,12 +274,14 @@ void initAttributesManagersBindings(py::module& m) {
 
   // ==== Light Layout Attributes Template manager ====
   declareBaseAttributesManager<LightLayoutAttributes>(m, "BaseLightLayout");
+  // NOLINTNEXTLINE(bugprone-unused-raii)
   py::class_<LightLayoutAttributesManager,
              AttributesManager<LightLayoutAttributes>,
              LightLayoutAttributesManager::ptr>(m,
                                                 "LightLayoutAttributesManager");
   // ==== Object Attributes Template manager ====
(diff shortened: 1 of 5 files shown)


src/esp/bindings/AttributesManagersBindings.cpp
```diff
@@ -274,12 +274,14 @@ void initAttributesManagersBindings(py::module& m) {
 
   // ==== Light Layout Attributes Template manager ====
   declareBaseAttributesManager<LightLayoutAttributes>(m, "BaseLightLayout");
+  // NOLINTNEXTLINE(bugprone-unused-raii)
```
Skylion007, line 277: Would like some confirmations from the reviewers that these are false positives.

erikwijmans: Yeah, those are all false positives.  We should probably turn this off globally for src/esp/bindings.  The most common use of RAII is locking mutexs and there likely won't ever be any mutexs in bindings.

Skylion007: Unfortunately, disabling a single check from clang-tidy in a subfolder is rather difficult unless we upgrade to clang-tidy 11 which supports inheriting from the parent config. As such, I'll just leave them as warnings.

jturner65 (approved):
LGTM

PR #1000, opened 2020-12-30
[sensor] A camera to render the scene into the cubemap

bigbike:
## Motivation and Context
This is a camera that can be put into the scene (attaching to a node in the scene graph), and render the surroundings into the cubemap.
We need the cubemap for:
- Fisheye sensor;
- Pbr Image based rendering;

More about this camera:
- It is a render camera that can render the scene. So it is inherited from the RenderCamera;
- The FOV is fixed to 90 degrees;
- The rendering plane is a square (width must be the same as height.);
- It has a convenient function called SwitchToFace() to control which cubemap's face it is rendering to;

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/CubeMapCamera.cpp
+++ b/src/esp/gfx/CubeMapCamera.cpp
@@ -40,32 +40,12 @@ CubeMapCamera& CubeMapCamera::restoreTransformation() {
 }
 
 CubeMapCamera& CubeMapCamera::switchToFace(unsigned int cubeSideIndex) {
-  switch (cubeSideIndex) {
-    case 0:
-      switchToFace(Mn::GL::CubeMapCoordinate::PositiveX);
-      break;
-    case 1:
-      switchToFace(Mn::GL::CubeMapCoordinate::NegativeX);
-      break;
-    case 2:
-      switchToFace(Mn::GL::CubeMapCoordinate::PositiveY);
-      break;
-    case 3:
-      switchToFace(Mn::GL::CubeMapCoordinate::NegativeY);
-      break;
-    case 4:
-      switchToFace(Mn::GL::CubeMapCoordinate::PositiveZ);
-      break;
-    case 5:
-      switchToFace(Mn::GL::CubeMapCoordinate::NegativeZ);
-      break;
-    default:
-      LOG(ERROR)
-          << "Warning: CubeMapCamera::switchToFace(): the index of cube side "
-          << cubeSideIndex
-          << " is illegal. Camera orientation stays unchanged.";
-      break;
-  }
+  CORRADE_ASSERT(cubeSideIndex < 6,
+                 "CubeMapCamera::switchToFace(): the index of the cube side,"
+                     << cubeSideIndex << "is illegal.",
+                 *this);
+  switchToFace(Mn::GL::CubeMapCoordinate(
+      int(Mn::GL::CubeMapCoordinate::PositiveX) + cubeSideIndex));
   return *this;
 }
 
@@ -95,7 +75,7 @@ CubeMapCamera& CubeMapCamera::switchToFace(Mn::GL::CubeMapCoordinate cubeSide) {
         return Mn::Matrix4::lookAt(eye, Mn::Vector3{0.0, 0.0, 1.0}, -yUp);
         break;
       default:
-        return Mn::Matrix4{Magnum::Math::IdentityInit};
+        CORRADE_INTERNAL_ASSERT_UNREACHABLE();
         break;
     }
   };
@@ -107,13 +87,15 @@ CubeMapCamera& CubeMapCamera::switchToFace(Mn::GL::CubeMapCoordinate cubeSide) {
 CubeMapCamera& CubeMapCamera::setProjectionMatrix(int width,
                                                   float znear,
                                                   float zfar) {
+  // NOLINTNEXTLINE(google-build-using-namespace)
+  using namespace Mn::Math::Literals;
   MagnumCamera::setProjectionMatrix(
       Mn::Matrix4::perspectiveProjection(
-          Mn::Deg{90.0},  // horizontal field of view angle
-          1.0,            // aspect ratio (width/height)
-          znear,          // z-near plane
-          zfar))          // z-far plane
-      .setViewport(Magnum::Vector2i(width, width));
+          Mn::Deg{90.0_degf},  // horizontal field of view angle
+          1.0,                 // aspect ratio (width/height)
+          znear,               // z-near plane
+          zfar))               // z-far plane
+      .setViewport({width, width});
   return *this;
 }
 
--- a/src/esp/gfx/CubeMapCamera.cpp
+++ b/src/esp/gfx/CubeMapCamera.cpp
@@ -1,0 +1,121 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree
+
+#include "CubeMapCamera.h"
+#include <Magnum/EigenIntegration/Integration.h>
+
+namespace Mn = Magnum;
+namespace Cr = Corrade;
+
+namespace esp {
+namespace gfx {
+
+CubeMapCamera::CubeMapCamera(scene::SceneNode& node) : RenderCamera(node) {
+  updateOriginalViewingMatrix();
+}
+CubeMapCamera::CubeMapCamera(scene::SceneNode& node,
+                             const vec3f& eye,
+                             const vec3f& target,
+                             const vec3f& up)
+    : CubeMapCamera(node,
+                    Mn::Vector3{eye},
+                    Mn::Vector3{target},
+                    Mn::Vector3{up}) {}
+CubeMapCamera::CubeMapCamera(scene::SceneNode& node,
+                             const Mn::Vector3& eye,
+                             const Mn::Vector3& target,
+                             const Mn::Vector3& up)
+    : RenderCamera(node, eye, target, up) {
+  updateOriginalViewingMatrix();
+}
+CubeMapCamera& CubeMapCamera::updateOriginalViewingMatrix() {
+  originalViewingMatrix_ = this->node().transformation();
+  return *this;
+}
+
+CubeMapCamera& CubeMapCamera::restoreTransformation() {
+  this->node().setTransformation(originalViewingMatrix_);
+  return *this;
+}
+
+CubeMapCamera& CubeMapCamera::switchToFace(unsigned int cubeSideIndex) {
+  switch (cubeSideIndex) {
+    case 0:
+      switchToFace(Mn::GL::CubeMapCoordinate::PositiveX);
+      break;
+    case 1:
+      switchToFace(Mn::GL::CubeMapCoordinate::NegativeX);
+      break;
+    case 2:
+      switchToFace(Mn::GL::CubeMapCoordinate::PositiveY);
+      break;
+    case 3:
+      switchToFace(Mn::GL::CubeMapCoordinate::NegativeY);
+      break;
+    case 4:
+      switchToFace(Mn::GL::CubeMapCoordinate::PositiveZ);
+      break;
+    case 5:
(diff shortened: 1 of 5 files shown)


bigbike:
Just noticed that this is PR 1K.

GREAT.

src/esp/gfx/CubeMapCamera.cpp
```diff
@@ -0,0 +1,121 @@
...
+        return Mn::Matrix4::lookAt(eye, Mn::Vector3{0.0, -1.0, 0.0}, zUp);
+        break;
+      case Mn::GL::CubeMapCoordinate::PositiveZ:
+        return Mn::Matrix4::lookAt(eye, Mn::Vector3{0.0, 0.0, -1.0}, -yUp);
+        break;
+      case Mn::GL::CubeMapCoordinate::NegativeZ:
+        return Mn::Matrix4::lookAt(eye, Mn::Vector3{0.0, 0.0, 1.0}, -yUp);
+        break;
+      default:
+        return Mn::Matrix4{Magnum::Math::IdentityInit};
```
mosra, line 98: assert here?

src/esp/gfx/CubeMapCamera.cpp
```diff
@@ -0,0 +1,121 @@
...
+                                 cameraLocalTransform());
+  return *this;
+}
+
+CubeMapCamera& CubeMapCamera::setProjectionMatrix(int width,
+                                                  float znear,
+                                                  float zfar) {
+  MagnumCamera::setProjectionMatrix(
+      Mn::Matrix4::perspectiveProjection(
+          Mn::Deg{90.0},  // horizontal field of view angle
```
mosra, line 112:
```suggestion
          90.0_degf,      // horizontal field of view angle
```

Why not use the literals? :)

(`using namespace Mn::Math::Literals;` somewhere of course)

src/esp/gfx/CubeMapCamera.cpp
```diff
@@ -0,0 +1,121 @@
...
+          1.0,            // aspect ratio (width/height)
+          znear,          // z-near plane
+          zfar))          // z-far plane
+      .setViewport(Magnum::Vector2i(width, width));
```
mosra, line 116:
```suggestion
      .setViewport(Magnum::Vector2i(width));
```

or

```suggestion
      .setViewport({width, width});
```

are both shorter

src/esp/gfx/CubeMapCamera.h
```diff
@@ -0,0 +1,125 @@
...
+   * 4: +Z
+   * 5: -Z
+   *           +----+
+   *           | -Y |
+   * +----+----+----+----+
+   * | -Z | -X | +Z | +X |
+   * +----+----+----+----+
+   *           | +Y |
+   *           +----+
+   * @return Reference to self (for method chaining)
```
mosra, line 78: The ASCII art will render pretty bad in the docs. Put it outside the `@param` documentation and wrap in a ``` block.

src/esp/gfx/CubeMapCamera.h
```diff
@@ -0,0 +1,125 @@
...
+  CubeMapCamera& switchToFace(unsigned int cubeSideIndex);
+
+  /**
+   * Calling the the setProjectionMatrix from base class is not allowed
+   */
+  RenderCamera& setProjectionMatrix(int width,
+                                    int height,
+                                    float znear,
+                                    float zfar,
+                                    float hfov) = delete;
```
mosra, line 89: > Calling the the setProjectionMatrix from base class is not allowed

Wait but the base implementation isn't `virtual` -- can it happen that someone calls this function from the base class pointer? And what about `setOrthographicProjectionMatrix()`, shouldn't it get deleted as well?

Alternatively and possibly more user friendly in case it's really intended to have those `virtual` and called from the base class pointer, would be to not delete this but instead assert that width and height is the same and FoV is 90°.

src/esp/gfx/CubeMapCamera.cpp
```diff
@@ -0,0 +1,121 @@
...
+      break;
+    case 3:
+      switchToFace(Mn::GL::CubeMapCoordinate::NegativeY);
+      break;
+    case 4:
+      switchToFace(Mn::GL::CubeMapCoordinate::PositiveZ);
+      break;
+    case 5:
+      switchToFace(Mn::GL::CubeMapCoordinate::NegativeZ);
+      break;
```
mosra, line 61: FYI the enums are defined in GL to have consecutive indices, so you can do something like 

```cpp
switchToFace(Mn::GL::CubeMapCoordinate(int(Mn::GL::CubeMapCoordinate::PositiveX) + cubeSideIndex));
```

instead of the giant switch.

src/esp/gfx/CubeMapCamera.cpp
```diff
@@ -0,0 +1,121 @@
...
+        break;
+      case Mn::GL::CubeMapCoordinate::NegativeY:
+        return Mn::Matrix4::lookAt(eye, Mn::Vector3{0.0, -1.0, 0.0}, zUp);
+        break;
+      case Mn::GL::CubeMapCoordinate::PositiveZ:
+        return Mn::Matrix4::lookAt(eye, Mn::Vector3{0.0, 0.0, -1.0}, -yUp);
+        break;
+      case Mn::GL::CubeMapCoordinate::NegativeZ:
+        return Mn::Matrix4::lookAt(eye, Mn::Vector3{0.0, 0.0, 1.0}, -yUp);
+        break;
```
mosra, line 96: The `break`s are not needed after a return -- some compilers will even warn about those being unreachable.

mosra (commented):
Congrats for #1000 :tada: 

A bunch of comments, mostly suggesting simpler / easier Magnum usage.

src/esp/gfx/CubeMapCamera.cpp
Skylion007: Convenient!

src/esp/gfx/CubeMapCamera.cpp
bigbike: Facebook coding convention requires "Always end a non-empty case with a break." (see here: [the public link](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es78-dont-rely-on-implicit-fallthrough-in-switch-statements)). Of course I understand here it is a special case because it is a "return" before the "break". But I prefer to keep the breaks here. It is a good habit and less error-prone in case any refactoring happen in the future.

So far I did not see any compiler give me such warnings, neither on MacOS nor on Linux. Thanks for the heads up!

src/esp/gfx/CubeMapCamera.cpp
bigbike: OK, that makes sense too. Let the developer fail faster to save debugging time.

src/esp/gfx/CubeMapCamera.cpp
bigbike: Good to know. Will do. Thanks.

src/esp/gfx/CubeMapCamera.h
bigbike: I see. Good to know. Will do.

src/esp/gfx/CubeMapCamera.h
bigbike: My purpose to delete it is that any developer who accidentally calls this function from the base class would get caught by the compiler, so she can fix it immediately. ("fail-fast")

And the `RenderCamera` class is NOT the interface class of the `CubeMapCamera`. I did not add `virtual` because no polymorphism is needed here. Anyone who tries to use it in such a way is misinterpreting the purpose. Ideally, a `RenderCamera` pointer should never be pointing to a `CubeMapCamera` instance.

However, I also like the following suggestions though. But again, I do not think polymorphism is needed here.
>Alternatively and possibly more user friendly in case it's really intended to have those virtual and called from the base class pointer, would be to not delete this but instead assert that width and height is the same and FoV is 90°.

bigbike:
@mosra : thoughts?

src/esp/gfx/CubeMapCamera.cpp
mosra: > So far I did not see any compiler give me such warnings

Not sure if `-Wpedantic -Wextra` is enabled on Habitat code, but I remember either those or the `/W4` warning level on MSVC complaining about "unreachable code after return". Or I might be misremembering, or maybe this particular case isn't checked. Conversely, on recent compilers these warning levels yell at you when you forget an actually important `break`, and I don't think they would warn here :)

> Always end a non-empty case with a break

Every ~~exception has a rule~~ rule has an exception, from my PoV the `break` is never reached and thus extraneous.

src/esp/gfx/CubeMapCamera.h
mosra: > CubeMapCamera is subclassed from RenderCamera simply to avoid duplicating functions. Ideally, a RenderCamera pointer should never be pointing to a CubeMapCamera instance.

I got some crazy ideas about private inheritance but ... eh. Keep it like this, then :)

src/esp/gfx/CubeMapCamera.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+CubeMapCamera& CubeMapCamera::setProjectionMatrix(int width,
+                                                  float znear,
+                                                  float zfar) {
+  // NOLINTNEXTLINE(google-build-using-namespace)
+  using namespace Mn::Math::Literals;
+  MagnumCamera::setProjectionMatrix(
+      Mn::Matrix4::perspectiveProjection(
+          Mn::Deg{90.0_degf},  // horizontal field of view angle
```
mosra, line 94: Err, I mean, you can use *just* the literal, which is why it's there, to make the code shorter to type and easier to read ;)

```suggestion
          90.0_degf,           // horizontal field of view angle
```

bigbike: Oops. Yeah, you are right. Forgot to get rid of the Mn::Deg. Will do.

src/esp/gfx/CubeMapCamera.cpp
bigbike: >Conversely, on recent compilers these warning levels yell at you when you forget an actually important break

It is quite dangerous. What if the developer ignores such warnings.

Yes, I totally understand your points here. It can reduce the noise from certain compilers, and make the code a bit simpler.

However, like I said, I believe it is a good habit to have a `break` in each case, and it follows Facebook coding convention we agreed on. It is just like we require `{}` in each if clause no matter if it is just one line of code.

More importantly, I did not see any significant benefits from making an exception to the rule. It just make one or two specific compiler less noisy (even the noise is not harmful). If we remove the `break`, however, it makes the code potentially more error-prone in the future refactoring. 

Shall we agree on the disagree here? Hope it is not deal breaker in order to get your approval on the code. :)

Skylion007:
Really excited to see the Python bindings for this, I guess that is out of the scope of this PR though?

bigbike:
@Skylion007 : You are right. Will do it in the other PRs. The priority here is to make it (the fisheye sensor) work in the simulator side as quickly as possible, so that the my intern will be unblocked. After this one, there are a couple of big diffs coming.
Even better, if you can join and help us to write the python bindings in parallel, it would greatly expedite everything. Hope you can consdier. :)

src/esp/gfx/CubeMapCamera.cpp
mosra: Yeah it's not a deal breaker for me, just sharing my view :)

> What if the developer ignores such warnings.

`-Werror=implicit-fallthrough` turns that into a *loud yell* :P 

And (just FYI) for a portable `[[fallthrough]]` there's a [CORRADE_FALLTHROUGH](https://doc.magnum.graphics/corrade/Macros_8h.html#af3cd136001e4eb1fc1a78aa17fe8232e).

mosra (approved):

bigbike:
@mosra : The quality of the code is significantly improved. Thank you very much for your thorough review every time!!

PR #1003, opened 2021-01-02
add JS replay viewer

eundersander:
## Motivation and Context

Gfx-replay viewer in the browser. This is a demo and reference for building other JS apps.

The live demo is hosted here: [aihabitat.org/replay_demo](https://www.aihabitat.org/replay_demo/). Some data files for this demo are temporarily hosted [here](https://www.ericundersander.com/habitat/data/).

replay_viewer_dominos2

- ~50 objects added to the scene
- 60 FPS
- smooth keyboard controls (WASD + arrow keys)

Previously, we've found it impractical to simulate so many objects in our javascript client, due to poor runtime perf. Here, I'm not actually simulating dynamics on the client. I'm just playing back a "replay" (recorded keyframes from an earlier simulation). It's a nice demo of the new gfx-replay feature, but it's not interactive; it's more of a mockup. However, imagine if our javascript client sent user input to the server and then received these keyframes (continuously, at some reasonable rate). This is the client-server architecture I want us to move to in the future. It will enable more complex physics, GPU-accelerated pytorch for human-and-model-in-the-loop, multiple users, and VR.

The data files for this viewer app (json and glb) are currently hosted at ericundersander.com. This is a temporary host until I find something better. Download them locally to your data folder if you want to iterate on this demo locally.

## How Has This Been Tested

Ad hoc testing in Fedora Chrome.

I will do more testing on other platforms and update here before merging.

## Types of changes

## Checklist


src/esp/bindings_js/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ set(
   ${SCENE_DATASETS}/habitat-test-scenes/van-gogh-room.navmesh
   ${MAGNUM_WINDOWLESSEMSCRIPTENAPPLICATION_JS}
   ${MAGNUM_WEBAPPLICATION_CSS}
+  ${DATA_DIR}
```
eundersander, line 34: This creates a symlink to `data` so that all your local data files become available when choosing files to preload in index.js.

src/esp/bindings_js/index.js
```diff
@@ -34,20 +35,24 @@ function preload(url) {
...
 Module.preRun.push(() => {
-  let config = {};
-  config.scene = defaultScene;
-  buildConfigFromURLParameters(config);
-  window.config = config;
-  const scene = config.scene;
-  Module.scene = preload(scene);
-  const fileNoExtension = scene.substr(0, scene.lastIndexOf("."));
+  if (window.replayViewerEnabled) {
+    ReplayViewer.preloadFiles(preload);
```
eundersander, line 39: Going forward, index.js is the entry point for all our "JS apps" including bindings.html and replay_viewer.html. I do think the choice of files to preload should be specific to the app, so here I delegate preloading to the ReplayViewer class.

I'm open to discussion about how to support multiple JS apps.

src/esp/bindings_js/modules/replay_viewer.js
```diff
@@ -0,0 +1,373 @@
...
+            currIndex == this.player.numKeyframes - 1 ? 0 : currIndex + 1;
+          this.player.setKeyframeIndex(newIndex);
+        }
+      }
+
+      for (let action of this.activeActions) {
+        this.handleAction(action);
+      }
+
+      this.render();
```
eundersander, line 61: Notice the Replay Viewer doesn't call sim.step. It doesn't use the agent action API.

src/esp/bindings_js/modules/replay_viewer.js
```diff
@@ -0,0 +1,373 @@
...
+      action == "strafeRight"
+    ) {
+      let stepDist = 0.06;
+      let stepSignedDist;
+      let adjustedRotation;
+      if (action == "moveForward" || action == "moveBackward") {
+        stepSignedDist = action == "moveForward" ? stepDist : -stepDist;
+        adjustedRotation = -this.agentRotation - Math.PI * 0.5;
+      } else {
+        stepSignedDist = action == "strafeLeft" ? stepDist : -stepDist;
```
eundersander, line 191: Notice this design decision: because this app is not tied to the agent action API, I'm able to easily add new interactivity such as strafing without touching the C++.

src/esp/bindings_js/modules/replay_viewer.js
```diff
@@ -0,0 +1,373 @@
...
+}
+
+class ReplayViewer {
+  static preloadFiles(preloadFunc) {
+    let dataUrlBase = "";
+    let useLocalSourceFiles = window.location.href.indexOf("0.0.0.0") != -1;
+    if (useLocalSourceFiles) {
+      dataUrlBase = "data";
+    } else {
+      dataUrlBase = "https://www.ericundersander.com/habitat/data";
```
eundersander, line 282: Temporary hosting for these files

dhruvbatra:
Looks awesome!

CC: @rutadesai @BlaiseRitchie -- this functionality is coming.

src/esp/bindings_js/modules/replay_viewer.js
Skylion007: I am really not a fan of hard coding this. Can't we force the user to specify it?

eundersander: >Can't we force the user to specify it?
I don't have a good answer here. I'm also not a fan of hard-coding this.

In practice, the "user" here is a Hab user looking to use replay_viewer.js as a reference to deploy their own replay. And they need to override not just the base URL but the full set of files (including all GLBs). This user can fork hab-sim and modify replay_viewer.js in place--they'll hard-code their specific set of files including specific URLs.

For now, I'll just remove the reference to ericundersander.com. Any user looking to deploy this will need to figure out their own URLs.

src/esp/bindings_js/bindings_js.cpp
```diff
@@ -79,9 +82,16 @@ vec4f eulerToQuaternion(const vec3f& q) {
       .coeffs();
 }
 
+vec4f quaternionMultiply(const vec4f& q1, const vec4f& q2) {
+  auto product = Magnum::Quaternion(quatf(q1)) * Magnum::Quaternion(quatf(q2));
+  return vec4f(product.data()[0], product.data()[1], product.data()[2],
+               product.data()[3]);
```
mosra, line 88: `vec4f` is Eigen, right?

```suggestion
  return Mn::EigenIntegration::cast<vec4f>(product);
``` 

should work. [Unfortunately explicit cast doesn't work in this direction because Eigen is special.](https://doc.magnum.graphics/magnum/integration_2src_2Magnum_2EigenIntegration_2Integration_8h.html#EigenIntegration-math)

mosra (commented):
I know next to nothing about modern JS, so just a tiny comment about the C++ code.

Issue #988, opened 2020-12-20
Rendering equirectangular images from replica in HDR (.hdr or .exr)


elientumba2019:
## ❓ Questions and Help
 Hello. Thanks for this amazing platform.

I have the following questions :　
1) How do I render hdr (.hdr, .exr not rgba) images from the replica dataset using habitat-sim? 
2) How do I render equirectangular images? 

Thank you

Skylion007:
1. See the example from Habitat-Lab about how to render equirect images from a cubemap. 
2. Pretty sure Magnum supports reading/writing HDR files, but I don't know if we actually support HDR rendering in Habtiat-Sim. @bigbike @mosra

elientumba2019:
I have checked all the examples available in the repo and could not find the one about how to render equirectangular images from a cube map. could you please point me to that resource? Thank you 
@Skylion007

elientumba2019:
@bigbike , @mosra . Hello, I hope you are well. Does habitat support writing HDR files? exr, or hdr. Specifically for the replica dataset as it comes with high dynamic range textures. Thank you

waldenlakes:
@mosra @bigbike I also want to know how to render HDR images. Looking forward to your help. Thank you!

mosra:
Magnum does support [writing HDR and EXR files](https://doc.magnum.graphics/magnum/file-formats.html) but Habitat isn't ready for that right now -- it always renders to a 8-bit framebuffer. It would need to be updated ([here](https://github.com/facebookresearch/habitat-sim/blob/a3820c4ea7dad72026d8b2a6562349fd31b3ac8b/src/esp/gfx/RenderTarget.cpp#L64), [here](https://github.com/facebookresearch/habitat-sim/blob/d09b0bdbc6404b0f19c83d6ef61f178b27e48fe2/src/esp/sensor/CameraSensor.cpp#L193) and probably some other places as well) to add an option for 16-/32-bit float output next to 8-bit.

waldenlakes:
[...]

PR #1011, opened 2021-01-07
[sensor] CubeMap

bigbike:
## Motivation and Context

This CubeMap stores cubemap textures (RGB, depth etc.) within, provides functions to setup, render to, save and load such textures.

Cubemap texture is needed in fisheye sensor, and PBR image based lighting.

## Types of changes

## Checklist


Skylion007:
Any reason the semantic sensor format is excluded for now? 
I assume the actual "VisualSensor" implementation / distinction comes later?

bigbike:
@Skylion007 : Simply because I do not have enough time to develop and debug it at the moment. :)
The plan is to have the infra pipeline of the fisheye sensor work first, and make it happen as quickly as possible. This is a kind of 0 to 1 change. Once we have that, we can do incremental changes on top of it, e.g., adding the semantic sensor to it.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+    // But the question is it has NOTHING to do with the framebuffer
+    // anymore at this moment.
+    Mn::Image2D image = textures_[type]->image(
+        convertFaceIndexToCubeMapCoordinate(iFace), 0, {getPixelFormat(type)});
+
+    std::string filename;
+    switch (type) {
+      case TextureType::Color: {
+        filename = imageFilePrefix + getTextureTypeFilenameString(type) +
+                   coordStrings[iFace] + std::string{".png"};
```
mosra, line 230: FYI there's [Utility::format()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#a5b85466788fc3d71b143909dda1ef44a), works the same as the Python flavor and will avoid a bunch of unnecessary allocations during the gradual string concatenation:

```suggestion
        filename = Cr::Utility::formatString("{}{}{}.png", imageFilePrefix, 
            getTextureTypeFilenameString(type), coordStrings[iFace]);
```

To save further needless allocations you could make `getTextureTypeFilenameString()` and `coordStrings` just plain `const char*`, or if you want to get fancy, a [Containers::StringView](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BasicStringView.html) (which is brand new, done for more efficiency in the Vulkan backend).

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+  auto loadImporter = [&](const std::string& plugin) {
+    Cr::PluginManager::Manager<Mn::Trade::AbstractImporter> manager;
+    Cr::Containers::Pointer<Mn::Trade::AbstractImporter> importer =
+        manager.loadAndInstantiate(plugin);
+    CORRADE_ASSERT(importer,
+                   "CubeMap::loadTexture(): cannot initialize" << plugin, );
+
+    imageImporterManger_.set<Mn::Trade::AbstractImporter>(
+        plugin, importer.release(), Mn::ResourceDataState::Final,
+        Mn::ResourcePolicy::Manual);
```
mosra, line 322: Wait this *can't* work, the `importer` is not allowed to outlive the `manager` (because in case of dynamic plugins the manager is responsible for `dlopen()`/`dlclose()`ing the plugin binary and other reasons). 

If it works and doesn't assert then it's maybe because for static plugins it's *mostly* fine, but is generally not a correct workflow.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+    texture->setSubImage(convertFaceIndexToCubeMapCoordinate(iFace), 0, {},
+                         *image);
+  }
+  // Color texture ONLY, NOT for depth
+  if (flags_ & (Flag::BuildMipMap | Flag::ColorTexture)) {
+    texture->generateMipmap();
+  }
+  // We don't need the importer anymore
+  imageImporterManger_.free<Mn::Trade::AbstractImporter>();
```
mosra, line 395: Why do you need a ResourceManager at all, then, if the plugin doesn't even outlive the function it was instantiated in? It could be just a function-local variable.

src/esp/gfx/CubeMap.cpp
mosra, line 230: LDR images are not really usable for IBL anyway, so rendering to a 8-bit framebuffer and saving as a PNG is not going to cut it in the long run. 

Which makes me wondering, wouldn't it just be easiest to have Magnum *properly* support OpenEXR? I was checking this just before Xmas due to #988 and the format is:

- 16-bit floats for color,
- 32-bit floats for Z, 
- understands mipmap,
- *and* has builtin support for cubemaps

so basically everything you need, and you would be able to store everything in a single file.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+    default:
+      CORRADE_INTERNAL_ASSERT_UNREACHABLE();
+      break;
+  }
+  for (int iFace = 0; iFace < 6; ++iFace) {
+    // open image file
+    std::string filename =
+        imageFilePrefix + getTextureTypeFilenameString(type) +
+        coordStrings[iFace] + "." + std::string{imageFileExtension};
```
mosra, line 367: `Utility::formatString()` could be used here as well

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+      imageSize = size.x();
+      reset(imageSize);
+    } else {
+      CORRADE_ASSERT(size.x() == imageSize,
+                     " CubeMap::loadTexture(): texture images must have the "
+                     "same size.", );
+    }
+
+    texture->setSubImage(convertFaceIndexToCubeMapCoordinate(iFace), 0, {},
+                         *image);
```
mosra, line 388: Did you test with HDR images? Because I expect this to not work, as the HDR images give you a RGB but the texture is just a single channel.

Maybe GL does some conversion under the hood, but I wouldn't bet on that.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+  }
+  // Color texture ONLY, NOT for depth
+  if (flags_ & (Flag::BuildMipMap | Flag::ColorTexture)) {
```
mosra, line 391: Same note about the wrong bitwise operations as above.

src/esp/gfx/CubeMap.h
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * @brief Render to cubemap texture using the camera
+   * @param camera, a cubemap camera
+   * NOTE: It will NOT automatically generate the mipmap for the user
+   */
+  void renderToTexture(CubeMapCamera& camera,
+                       scene::SceneGraph& sceneGraph,
+                       RenderCamera::Flags flags);
+
+ protected:
```
mosra, line 141: private?

src/esp/gfx/CubeMap.h
```diff
@@ -0,0 +1,202 @@
...
+#include "esp/gfx/CubeMapCamera.h"
+#include "esp/gfx/RenderCamera.h"
+#include "esp/scene/SceneGraph.h"
+#include "esp/scene/SceneNode.h"
+
+namespace esp {
+namespace gfx {
+class CubeMap {
+ public:
+  static void enableSeamlessCubeMapTexture();
```
mosra, line 27: Why not just call this implicitly on construction or something? It's always good to have this enabled, and this makes it too easy to forget (and the function doesn't seem to be used anywhere in this PR).

Skylion007:
> @Skylion007 : Simply because I do not have enough time to develop and debug it at the moment. :)
> The plan is to have the infra pipeline of the fisheye sensor work first, and make it happen as quickly as possible. This is a kind of 0 to 1 change. Once we have that, we can do incremental changes on top of it, e.g., adding the semantic sensor to it.

Just a little confused how to hook this up to the current VisualSensor API to generate an array of textures at the moment. I assume more boilerplate / Sensor abstraction will be in a subsequent PR?

bigbike:
@Skylion007 : if you are really interested in it, and would like to have a sneak peak, check the "cubemap" branch. Fisheye sensor is implemented there.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+  Cr::Containers::Pointer<Mn::Trade::AbstractImageConverter> converter;
+  if (!(converter = manager.loadAndInstantiate("AnyImageConverter"))) {
+    return false;
+  }
+
+  std::string coordStrings[6] = {".+X", ".-X", ".+Y", ".-Y", ".+Z", ".-Z"};
+  for (int iFace = 0; iFace < 6; ++iFace) {
+    // TODO: use framebuffer.read() instead somehow?
+    // But the question is it has NOTHING to do with the framebuffer
+    // anymore at this moment.
```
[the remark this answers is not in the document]
bigbike: I considered your proposal when I wrote it. 
In OpenGL, changing the rendering target (FBO) is quite expensive. 
This is based on my experience and it matches what is described in "Real-time Rendering (4th)". Here is a screenshot.
Screenshot from 2021-01-08 15-58-26
The data is as of 2014 though. I did not verify if it is still the case on the GPU in recent 2 years. I can try it later. In this version, I would like to keep it as this (One FBO).

VK is completely different topic.

src/esp/gfx/CubeMap.cpp
bigbike: >DR images are not really usable for IBL anyway

It still works, and the final result is just not that great.

So what you suggest here? 

Practical way is that I keep the way it is now, while you can develop the Magnum to support OpenEXR. After you have it, we can refactor the code.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+    prepareToDraw(iFace);
+
+    // TODO:
+    // camera should have flags so that it can do "low quality" rendering,
+    // e.g., no normal maps, no specular lighting, low-poly meshes,
+    // low-quality textures.
+
+    for (auto& it : sceneGraph.getDrawableGroups()) {
+      // TODO: remove || true
+      if (it.second.prepareForDraw(camera) || true) {
```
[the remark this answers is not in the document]
bigbike: Are you asking `it.second.prepareForDraw(camera)`?
This is not something newly invented. I just copied it from any drawEvent().

bigbike (commented):
I have not read all the reviewer's comments.
This is just my response to the ones I already went through. Will keep you updated.

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
mosra: > In OpenGL, changing the rendering target (FBO) is quite expensive.

Yes, that's correct. But what is commonly *even more* expensive (and what the book doesn't mention) is changing attachments to an existing framebuffer, because the driver has to redo all validation instead of switching to a different but already validated target. :) Though of course it's all just words until a benchmark can prove either way.

> VK is completely different topic.

I don't think it is, GL drivers on Mac are built on top of Metal, which has similar constraints about render targets and pipeline state as Vulkan, and Vulkan was designed based on the experience coming from what had to be done in GL and D3D drivers internally. GL can't just magically be faster at the same thing Vulkan is slow at.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+    case TextureType::Depth:
+      CORRADE_ASSERT(flags_ & Flag::DepthTexture,
+                     functionNameStr.c_str()
+                         << "instance was not created with depth "
+                            "texture output enabled.", );
+      break;
+    default:
+      CORRADE_INTERNAL_ASSERT_UNREACHABLE();
+      break;
+  }
```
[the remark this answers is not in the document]
mosra: > But what if a developer ignores that warning?

Sure. The point is that the developers *can* get the warning at build time already if they choose to pay attention to new warnings, instead of later at runtime (where the code path might not even get hit until several months later).

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
mosra: Yeah I was wondering what is the reason for `prepareForDraw()` failing, what can cause that? Is it even valid for it to fail here? Shouldn't it be asserted? etc.

src/esp/gfx/CubeMap.cpp
mosra: Yeah I'm just exploring the options. If OpenEXR isn't deemed useful, then great, less work for me; if it makes sense to have it / if it saves you or someone else extra work, then let's consider doing it.

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
bigbike: You analysis sounds quite reasonable. When I said "VK is completely different topic". I did not mean GL can be faster 
 on something where VK is slow. My point is that experience gained on VK is not necessarily true on GL. For example, here, when being implemented with VK, the method with "6 FBOs" would be faster than the one with "single FBO". It is likely the same case on GL, but I will have to benchmark it to verify.

src/esp/gfx/CubeMap.h
bigbike: I simply hope it can be called once, not every time in the constructor.

I can change.

src/esp/gfx/CubeMap.cpp
bigbike: NO, I have not tested it with HDR yet! I was about to ask how to implemented in the correct way.
You mentioned to me on the slack channel that somehow I need to convert it (from RGB separate channels to a singl channel). Any code snippet to demo how to do it would be appreciated here.

src/esp/gfx/CubeMap.h
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * TODO: ObjectId
+     */
+    /**
+     * Build mipmap for cubemap color texture
+     * By default, NO mipmap will be built, only 1 level
+     * By turning on this option, it will build the mipmap for the color texture
+     * if any.
+     */
+    BuildMipMap = 1 << 3,
```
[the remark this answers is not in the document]
bigbike: Will do!

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
mosra: Actually, scratch that. I got a new idea that I think could work much better -- attach all six cubemap faces as six separate color attachments at once and then only call `mapForDraw()` where you draw to a different attachment every time.

src/esp/gfx/CubeMap.cpp
bigbike: Actually, look at the pixel format I used for the color texture. RGBA8. Let us temporarily forget about the external HDR images, even internally, when I render the scene into such a texture, it is already has limitation for IBL because it is LBR. Correct? What is the correct pixel format then?

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
bigbike: >attach all six cubemap faces as six separate color attachments at once

Haha! Nice! Thank you. Why did not I think of that? :)

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,399 @@
...
+  // CAREFUL!!!
+  // switchToFace() will change the local transformation of this camera node!
+  // If you do not do anything, in next rendering cycle, since
+  // camera.updateOriginalViewingMatrix() is called, the original viewing matrix
+  // will be updated by mistake!!! To prevent such mistakes, local
+  // transformation of this camera node must be reset.
+  camera.restoreTransformation();
+
+  // Color texture ONLY, NOT for depth
+  if (flags_ & (Flag::BuildMipMap | Flag::ColorTexture)) {
```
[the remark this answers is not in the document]
bigbike: OK. It is a bug. Good Catch!!

src/esp/gfx/CubeMap.cpp
bigbike: This might be the reason of the super rare segfault I mentioned to you a while ago.

src/esp/gfx/CubeMap.cpp
mosra: > Actually, look at the pixel format I used for the color texture.

Yep, that's what I mentioned in #988 as well. For OpenEXR it'd be 16-bit floats, so RGBA16F.

src/esp/gfx/CubeMap.cpp
mosra: ... or we can go with OpenEXR and don't bother with limitations of the HDR format :)

It's not hard, basically you convert a tightly-packed stream of RGBRGBRGB... data to RRRR..., throwing away the two other channels.

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
bigbike: Just look into this function:
```
  /**
   * @return Whether the @ref DrawableGroup is in a valid state to be drawn
   */
  virtual bool prepareForDraw(const RenderCamera&) { return true; }
```

I guess in the future, we may have different drawable groups, and not all of them would be drawn.

src/esp/gfx/CubeMap.h
```diff
@@ -0,0 +1,202 @@
...
+   * @brief check if the class instance is created with corresponding texture
+   * enabled
+   */
+  void textureTypeSanityCheck(TextureType type,
+                              const std::string& functionNameStr);
+
+  /**
+   * @brief convert cube face index to Magnum::GL::CubeMapCoordinate
+   */
+  Magnum::GL::CubeMapCoordinate convertFaceIndexToCubeMapCoordinate(
```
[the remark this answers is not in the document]
bigbike: Very good point. Makes a lot of sense. Will change.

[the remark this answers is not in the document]
bigbike: > to avoid a false impression that they're needed outside of the class as well. 

Actually, they are private functions, so they cannot be called outside of the class.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,396 @@
...
+    switch (type) {
+      case TextureType::Color:
+        texture->setSubImage(convertFaceIndexToCubeMapCoordinate(iFace), 0, {},
+                             *image);
+        break;
+
+      case TextureType::Depth: {
+        // R32F means 4 bytes per pixel
+        const int dim = static_cast<int>(std::sqrt(image->pixelSize() / 4));
+        // reinterpret the data as a R32F image
```
bigbike, line 325: @mosra: can I do it in this way?
What I am thinking is no matter what is the input's pixel format, whether it is RGB32F, or anything else, reinterpret it as R32F. 
Is this code correct?

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,396 @@
...
+      case TextureType::Depth: {
+        filename = Cr::Utility::formatString("{}.{}.{}.hdr", imageFilePrefix,
+                                             getTextureTypeFilenameString(type),
+                                             coordStrings[iFace]);
+      } break;
+    }
+    CORRADE_ASSERT(!filename.empty(),
+                   "CubeMap::saveTexture(): Unknown texture type.", false);
+
+    if (!converter->exportToFile(image, filename)) {
```
bigbike, line 253: For the depth texture, whose pixel format is R32F, would this (`converter->exportToFile(image, filename)`) be good enough to save the data as .hdr?

bigbike:
Two TODOs before it is finished:

- Double confirm the "depth texture save/load" function works correctly;
- One framebuffer with 6 attached Color attachment, and 6 depth components;

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
mosra: Hmm. Doesn't really *feel* like a good API design to me (because I had to ask what was the actual meaning). 

Nevertheless, this means the `|| true` should then be removed, as your TODO says.

src/esp/gfx/CubeMap.cpp
mosra: Yes, the plugin will [internally expand to RGB, [as the docs say](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StbImageConverter.html).

src/esp/gfx/CubeMap.cpp
mosra: No, because the green channel of the first pixel will be now treated as the second pixel, the blue channel of the first pixel will be now treated as the third pixel and so on.

And the `dim` is also wrong. Completely. The `pixelSize()` is a *size of the pixel*, which was originally 4, now it will be 1. 

Did you even try to run the code? :)

src/esp/gfx/CubeMap.cpp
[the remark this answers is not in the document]
bigbike: I guess Matthew and Alex put it here last year. Will need to sync up with them first.

src/esp/gfx/CubeMap.cpp
bigbike: No, I did not have hdr images to try it. I should try it. My bad.
I simply do not have time to do it during the weekend, and I cannot do it in this PR, because it only introduces the CubeMap class. No one is calling it now.

bigbike: So can I just do:
```
        Mn::ImageView2D imageView(Mn::PixelFormat::R32F, image->size(),
                                  image->data());
                                  ```

Comment (mosra) [reply src/esp/gfx/CubeMap.cpp:328]: (Replied to the second comment on Slack.)

> I cannot do it in this PR, because it only introduces the CubeMap class.

Yes but what's the point of the PR having code that's just completely broken? Not put it at all there, then. In my opinion, a PR should ideally be *atomic*, that is a minimal but complete working piece of code, working correctly, documented and including tests. Anything less leads to technical debt and headaches.

Comment (bigbike) [reply src/esp/gfx/CubeMap.cpp:328]: > Yes but what's the point of the PR having code that's just completely broken? 

I never say I will merge it like this. I am working on this code to get it correct. That is the purpose to have the code review, get advices from the reviewer, fix anything that needs to be fixed.

And when I say:
>I cannot do it in this PR, because it only introduces the CubeMap class.

Please do not get me wrong. I cannot test it here in this PR, it does not mean I would not test it. Actually I am testing it in my other branch called "cubemap".

>In my opinion, a PR should ideally be atomic, that is a minimal but complete working piece of code, working correctly, documented and including tests. 

I agree. We are on the same page. Again, I never, ever meant to push it in current shape. I am still working on it.

File: src/esp/gfx/CubeMap.cpp [hunk 17]
```diff
@@ -0,0 +1,438 @@
...
+
+  CORRADE_INTERNAL_ASSERT(
+      frameBuffer_.checkStatus(Mn::GL::FramebufferTarget::Draw) ==
+      Mn::GL::Framebuffer::Status::Complete);
+}
+
+void CubeMap::mapForDraw(unsigned int colorAttachmentIndex) {
+  frameBuffer_.mapForDraw({
+      {Mn::Shaders::Generic3D::ColorOutput,
+       colorAttachment[colorAttachmentIndex]},
```
mosra, line 233: Why have an array for this and not directly use

```suggestion
       Mn::GL::Framebuffer::ColorAttachment{colorAttachmentIndex}},
```

? The array indirection only complicates that in my opinion.

Hm and given that you need to re-attach the depth buffer every time, I'm not sure about the efficiency of this suggested approach either...

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,438 @@
...
+        // The pixel format for depth texture is R32F. When it is saved as hdr,
+        // the single channel is expanded to three channels by repeating the R
+        // channel 3 times (and it becomes RGB32F).
+        // That means when it is loaded, we need to dedup by taking just the
+        // first component (R component) out of each pixel.
+        Cr::Containers::StridedArrayView2D<const Mn::Color3>
+            imageStridedArrayView = imageData->pixels<Mn::Color3>();
+
+        Cr::Containers::Array<float> depthImage{
+            static_cast<size_t>(size.x() * size.y())};
```
mosra, line 360:
```suggestion
            static_cast<size_t>(size.product())};
```

;)

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,438 @@
...
+        unsigned int idx = 0;
+        for (auto row : imageStridedArrayView) {
+          for (const Mn::Color3& pixel : row) {
+            depthImage[idx++] = pixel.r();
+          }
+        }
```
mosra, line 366: If you want to get fancy, it's possible to do the copy in a potentially more efficient way using [Utility::copy()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#a6cd0c04124a0d7c4879d4cf6da3556a2). It takes two views of the same type, dimensions and size as input, so you need to turn the `imageStridedArrayView` into a float view as well and make a view on the `depthImage`:

```cpp
/* This casts the Color3 view to a float, which effectively uses only four bytes of each element, 
   which is the red channel */
Cr::Containers::StridedArrayView2D<const float> red = 
    Cr::Containers::arrayCast<const float>(imageStridedArrayView);
/* Turns the `depthImage` into a tightly packed 2D view of the same rows and pixels as 
   `imageStridedArrayView` */
Cr::Containers::StridedArrayView2D<float> output{depthImage, {std::size_t(size.y()), std::size_t(size.x())}};
/* And this does a copy of the data */
Cr::Utility::copy(red, output);
```

Be sure to test this, I might have made some minor mistake and caused copy() to assert ;)

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,438 @@
...
+      case TextureType::Depth: {
+        // The pixel format for depth texture is R32F. When it is saved as hdr,
+        // the single channel is expanded to three channels by repeating the R
+        // channel 3 times (and it becomes RGB32F).
+        // That means when it is loaded, we need to dedup by taking just the
+        // first component (R component) out of each pixel.
+        Cr::Containers::StridedArrayView2D<const Mn::Color3>
+            imageStridedArrayView = imageData->pixels<Mn::Color3>();
+
+        Cr::Containers::Array<float> depthImage{
```
mosra, line 359:
```suggestion
        Cr::Containers::Array<float> depthImage{Cr::Containers::NoInit,
```

to avoid zero-initializing the whole array, since it's overwritten anyway -- for bigger sizes the time saving is quite significant

src/esp/gfx/CubeMap.cpp
bigbike: It still helps in the color sensor case. The performance got visibly improved. See the following test results:
Before:
before-change-fps

After:
after-change-fps

bigbike: > The array indirection only complicates that in my opinion.

Yes, I see your point. I defined them in the very beginning of this file and by doing so, later if we would like to define the another color attachment for the objectId,  it is less error prone. (I would remember that the first 6 color attachments are already taken.)

mosra: Oh nice, the frame profiler is already useful :heart: 

If you have time, could you try how the CPU overhead would be with six different framebuffer objects (my original suggestion)?

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+  Cr::PluginManager::Manager<Mn::Trade::AbstractImageConverter> manager;
+  Cr::Containers::Pointer<Mn::Trade::AbstractImageConverter> converter;
+  if (!(converter = manager.loadAndInstantiate("AnyImageConverter"))) {
+    return false;
+  }
+
+  const char* coordStrings[6] = {"+X", "-X", "+Y", "-Y", "+Z", "-Z"};
+  for (int iFace = 0; iFace < 6; ++iFace) {
+    Mn::Image2D image = textures_[type]->image(
```
bigbike, line 253: @mosra :
For depth texture, can I get the image of each side using this `image(...)`?
I I tried to print out the 1st 30 floats (30 depth values) using:
```
        Mn::Debug{} << Cr::Containers::arrayCast<float>(image.data().prefix(120));
```
And they were all 0.

(I also verified the color texture, it was OK. Non-zero values. `saveTexture` and `loadTexture` worked as expected for the rgb cubemap, but not for depth cubemap, as all the pixels were "0".)

mosra: If you get all zeros, then you probably get a GL error as well (you can check with the `--magnum-gpu-validation` option). Which is probably due to the R32F format being used. If you do the following:

```cpp
        convertFaceIndexToCubeMapCoordinate(iFace), 0, {GL::PixelFormat::DepthComponent, GL::PixelType::Float});
```

it should give you the proper values, I think. But then you have to "reinterpret" the image as a R32F again when saving the image. Something like

```cpp
ImageView2D depthAsRedChannel{image.storage(), PixelFormat::R32F, image.size(), image.data()};
converter->exportToFile(depthAsRedChannel, filename) ...
```

bigbike: quoted from @mosra from other conversations :
"because you're trying to read R32F while it's a depth format (for that you need to use a GL::PixelFormat and GL::PixelType pair, as there are no generic depth PixelFormat values)"
"to explain further: because no image formats so far support storing depth data, there's no generic depth PixelFormat ... but that will probably change with the EXR support implemented, and all this will become simpler"

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,507 @@
...
+// LICENSE file in the root directory of this source tree
+#include "CubeMap.h"
+#include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Containers/Optional.h>
+#include <Corrade/Containers/StridedArrayView.h>
+#include <Corrade/PluginManager/Manager.h>
+#include <Corrade/Utility/Algorithms.h>
+#include <Corrade/Utility/Assert.h>
+// XXX
+#include <Corrade/Utility/Debug.h>
```
bigbike, line 12: Any thing started with XXX is for debugging, and will be removed later.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,507 @@
...
+            image.storage(), Mn::PixelFormat::R32F, image.size(), image.data()};
+        if (!converter->exportToFile(depthAsRedChannel, filename)) {
+          return false;
+        }
+        // XXX
+        // display 8 floats
+        /*
+        Mn::Debug{} << "========"
+                    << Cr::Containers::arrayCast<float>(
+                           image.data().prefix(32));
```
bigbike, line 306: I printed out 8 floats (depth values for 8 pixels), I expect when the image is loaded in the `loadTexture(...)` later, the values should be the same.
BUT THEY ARE NOT. I suspect something is wrong in the `loadTexture(...)`

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,507 @@
...
+        Mn::Debug{} << "image size in bytes: " << image.data().size();
```
bigbike, line 307: I printed out the size, which was 5760000. This is correct since the image resolution is 1200 x 1200.
 5760000 bytes = 1200 x 1200 x 4bytes/float

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,507 @@
...
+        "{}.{}.{}.{}", imageFilePrefix, getTextureTypeFilenameString(type),
+        coordStrings[iFace], imageFileExtension);
+
+    importer->openFile(filename);
+    Cr::Containers::Optional<Mn::Trade::ImageData2D> imageData =
+        importer->image2D(0);
+
+    // XXX
+    /*
+    Mn::Debug{} << "imagedata raw data size = " << (*imageData).data().size();
```
bigbike, line 362: I printed out the size, which was 17280000, and it is correct. because 17280000 = 5760000 x 3.
3 means when the image (R32F) is saved to .hdr format, it is expended to RGB32F. And 5760000 matched with the number printed out in the `saveTexture(...)`. See my previous comments in that function.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,507 @@
...
+    {
+      auto view = Cr::Containers::arrayCast<const float>((*imageData).data());
+      Mn::Debug{} << "first 4 rgbs from imagedata"
+                  << view.prefix(12);  // first 12 floats --> 4 Rgbs
```
bigbike, line 366: Here is the problem.
I printed out the first 12 floats (4 RGBs)
Every float from the image becomes: 0.996094, which is different from the values in the `saveTexture(...)`

src/esp/gfx/CubeMap.cpp
bigbike: For example, here the output can be:
{0.99788, 0.99788, 0.99788, 0.997879, 0.997879, 0.997879, 0.997879, 0.997878}

CAREFUL: some values are slightly different.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,507 @@
...
+        unsigned int idx = 0;
+        for (auto row : imageStridedArrayView) {
+          for (const Mn::Color3& pixel : row) {
+            depthImage[idx++] = pixel.r();
+          }
+        }
+        */
+        // Mn::Debug{} << "depthImage (top 9) = " << depthImage.prefix(9);
+        // Mn::Debug{} << "imageData size: " << imageData->size();
+        Mn::ImageView2D imageView(Mn::GL::PixelFormat::DepthComponent,
```
bigbike, line 426: Note: compared to the previous commit, I replace the R32F by `Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float`

Hope it is the correct thing to do.

mosra:
For the record, the difference in values is because [Radiance HDR](https://en.wikipedia.org/wiki/RGBE_image_format) stores the values as 8-bit RGB + 8-bit exponent, not a full 32-bit float per channel. Two options:

- either you'll live with it for now
- or wait until EXR support in magnum is done, which should give you an ability to store the full 32-bit float depth channel

Additionally, https://github.com/mosra/magnum-plugins/commit/6c34b357c90703771f86a3598ecd162ff40329e3 should simplify the HDR file loading as it allows you to override the imported channel count. Feel free to upgrade and adapt the code to use this feature.

src/esp/gfx/CubeMap.h
```diff
@@ -0,0 +1,203 @@
...
+  int imageSize_ = 0;
+  std::map<TextureType, std::unique_ptr<Magnum::GL::CubeMapTexture>> textures_;
+
+  /**
+   * @brief Recreate textures
+   */
+  void recreateTexture();
+
+  // framebuffers (one for every cube side)
+  Magnum::GL::Framebuffer frameBuffer_[6]{
```
bigbike, line 162: Is there a better way to initialize this?

src/esp/gfx/CubeMap.cpp
bigbike: Yes, I did the tests:
1 FBO + 6 colorAttachements VS 6FBOs

In "1FBO + 6": It needs to attach depth buffer every frame. 
(Note: I also have fbo.bind(...) in every frame but since the fbo never changes, I assume Magnum will NOT actually bind it again and again.)

In 6FBOs: It does NOT need to attach any color or depth buffer, but need to rebind FBOs each time.

The following pic is for "1FBO+6":


The following pic is for "6FBOs":
6fbos

In terms of CPU time: 1FBO+6 < 6FBOs
In terms of GPU time: 1FBO+6 > 6FBOs
Overall time: 1FBO+6 > 6FBOs (winner)
This AB test has been repeated multiple times, and I used my cellphone to capture the results (screenshot app will significantly affect the measurement.). The results are trustable.

@mosra: Can you explain why 6FBOs beats 1FBO+6 on GPU?

bigbike: The PR is updated to use the 6FBOs based on the result.

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,497 @@
...
+          Mn::GL::Framebuffer::BufferAttachment::Depth,
+          optionalDepthBuffer_[index]);
+    }
+  }
+}
+
+void CubeMap::prepareToDraw(unsigned int cubeSideIndex) {
+  // Note: we ONLY need to map shader output to color attachment when necessary,
+  // which means in depth texture mode, we do NOT need to do this
+  if (flags_ & CubeMap::Flag::ColorTexture) {
```
bigbike, line 222: @mosra: 
I also did tests in "DEPTH texture mode" where color output is NOT needed.
My settings:
**Control group:** has normal color buffer output.
**Test group A:** has normal color buffer output, but DISABLE the color output (magnum wrapper glColorMask(false ...) when entering rendering, and reenable it after rendering.
**Test group B:** NO color attachment, thus NO mapForDraw

Result:
Control group:
control
Test A:
testa
Test B:
testB

In terms of rendering time: 
Control > Test A > Test B (winner).

**No color attachment can beat the simply mask writing**, if my experiments are correct.
(PR is updated to use this method.)

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,458 @@
...
+  std::string importerName{"AnyImageImporter"};
+  if (type == TextureType::Depth) {
+    importerName = "StbImageImporter";
+  }
+  Cr::Containers::Pointer<Mn::Trade::AbstractImporter> importer =
+      manager.loadAndInstantiate(importerName);
+  CORRADE_INTERNAL_ASSERT(importer);
+  if (type == TextureType::Depth) {
+    // Override image channel count. Because depth info should be R32F
+    importer->configuration().setValue("forceChannelCount", 1);
```
bigbike, line 339: @mosra:

I just updated the PR using `forceChannelCount` based on my understanding. Still no luck here. 
The data saved in the `saveTexture` are different from the data loaded from the `loadTexture` (I printed out the first 12 floats).

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,458 @@
...
+            convertFaceIndexToCubeMapCoordinate(iFace), 0,
+            {Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float});
+        Mn::ImageView2D depthAsRedChannel{
+            image.storage(), Mn::PixelFormat::R32F, image.size(), image.data()};
+        if (!converter->exportToFile(depthAsRedChannel, filename)) {
+          return false;
+        }
+        // XXX
+        // display 1st 4 rgbs
+        Mn::Debug{} << "1st 4 rgb ==== SAVE"
```
bigbike, line 294: I output the first 12 floats here.
example output:
`{0.997912, 0.997912, 0.997912, 0.997911, 0.997911, 0.997911, 0.997911, 0.997911, 0.997911, 0.997911, 0.997911, 0.997912}`

src/esp/gfx/CubeMap.cpp
```diff
@@ -0,0 +1,458 @@
...
+      case TextureType::Depth: {
+        CORRADE_INTERNAL_ASSERT(imageData->format() == Mn::PixelFormat::R32F);
+        Mn::ImageView2D imageView(Mn::GL::PixelFormat::DepthComponent,
+                                  Mn::GL::PixelType::Float, imageData->size(),
+                                  imageData->data());
+        texture->setSubImage(convertFaceIndexToCubeMapCoordinate(iFace), 0, {},
+                             imageView);
+        // XXX
+        Mn::Debug{} << Cr::Containers::arrayCast<const float>(
+            imageView.data().prefix(32));
```
bigbike, line 386: Here, I output the first 12 floats from the loaded image. 
example outputs:
`{0.996094, 0.996094, 0.996094, 0.996094, 0.996094, 0.996094, 0.996094, 0.996094, 0.996094, 0.996094, 0.996094, 0.996094}`
Everything is 0.996094 again.

src/esp/gfx/CubeMap.cpp
mosra: > The data saved in the saveTexture are different from the data loaded from the loadTexture

That's inherent to the HDR format, as I on Slack few weeks ago -- see https://en.wikipedia.org/wiki/RGBE_image_format for reference.

> Note that I did not touch the saveTexturefunction which was still using anyImageConverterto dump out depth info. Should I use stbImageConverter`?

Ah, yeah... if you still get three channels on input and not just one, you need to use `StbImageImporter` directly, not the `AnyImageImporter` proxy. (Sorry about that. I have a TODO there, the configuration values aren't propagated from `AnyImageImporter` currently because of some missing APIs in Corrade.)

src/esp/gfx/CubeMap.h
mosra: Probably not... except if you'd want to use [StaticArray](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1StaticArray.html#Containers-StaticArray-initialization), which has the magic (I hope it works with move-only types):

```cpp
Corrade::Containers::StaticArray<6, Magnum::GL:Framebuffer> frameBuffer_{Corrade::Containers::DirectInit, Magnum::NoCreate};
```

src/esp/gfx/CubeMap.cpp
mosra: I don't want to be mean, but ... were the screenshots really necessary? Just writing out the few numbers would be enough, with this the signal:noise ratio is *really high*.

bigbike:
Put here for the record:
The precision of radiance HDR format cannot satisfy our need here (e.g., we will lose precision when depth texture is saved). We are going to check it in like this for now. Will revisit this problem when other formats such as EXR are supported in Magnum.

@mosra : Code is cleaned. Would you take a final look next week? Thanks.

src/esp/gfx/CubeMap.h
```diff
@@ -0,0 +1,207 @@
...
+   * NOTE: It will NOT automatically generate the mipmap for the user
+   */
+  void renderToTexture(CubeMapCamera& camera,
+                       scene::SceneGraph& sceneGraph,
+                       RenderCamera::Flags flags);
+
+ private:
+  Flags flags_;
+  int imageSize_ = 0;
+  std::map<TextureType, std::unique_ptr<Magnum::GL::CubeMapTexture>> textures_;
```
mosra, line 163: Not a fan of `std::map`, especially if you index it by a monotonic enum; the unique pointer is totally unnecessary also. Could be just a plain C array indexed by `TextureType`, if you want extra robustness you could wrap it in some getter, like

```cpp
enum class TextureType: uint8_t { 
    Color,
    Depth,
    ...
    Count
 };

Magnum::GL::CubeMapTexture textures_[uint8_t(TextureType::Count)];

Magnum::GL::CubeMapTexture& texture(TextureType type) {
    CORRADE_INTERNAL_ASSERT(uint8_t(type) < Containers::arraySize(textures_));
    return textures_[uint8_t(type)];
}
```

src/esp/gfx/CubeMap.h
```diff
@@ -0,0 +1,207 @@
...
+   * @brief Recreate textures
+   */
+  void recreateTexture();
+
+  // framebuffers (one for every cube side)
+  Corrade::Containers::StaticArray<6, Magnum::GL::Framebuffer> frameBuffer_{
+      Corrade::Containers::DirectInit, Magnum::NoCreate};
+
+  // in case there is no need to output depth texture, we need a depth buffer
+  Magnum::GL::Renderbuffer optionalDepthBuffer_[6];
```
mosra, line 175: `NoCreate` those as well? Otherwise it unnecessarily creates 6 GL renderbuffers even if not needed.

bigbike:
@mosra : Thanks for the comments. Fixed. Take a look.

src/esp/gfx/CubeMap.h
```diff
@@ -36,7 +36,12 @@ class CubeMap {
     Depth = 1,
     // TODO: ObjectId
     // TODO: HDR color
+
+    // you will have to update it when new type is added
+    Count = 2,
   };
+  static const uint8_t numTextureTypes =
+      static_cast<uint8_t>(TextureType::Count);
```
mosra, line 44: You don't need this, it's redundant. Use `uint8_t(TextureType::Count)` in the two places where it's needed instead.

src/esp/gfx/CubeMap.h
```diff
@@ -36,7 +36,12 @@ class CubeMap {
     Depth = 1,
     // TODO: ObjectId
     // TODO: HDR color
+
+    // you will have to update it when new type is added
+    Count = 2,
```
mosra, line 41:
```suggestion
    Count
```

Keep it without a number, that's the trick. The compiler will automatically assign a value larger than the previous enum value and thus you don't have to explicitly take care of this.

src/esp/gfx/CubeMap.cpp
```diff
@@ -109,6 +109,12 @@ Mn::PixelFormat getPixelFormat(CubeMap::TextureType type) {
   CORRADE_INTERNAL_ASSERT_UNREACHABLE();
 }
 
+uint8_t textureIndex(CubeMap::TextureType type) {
+  uint8_t idx = static_cast<uint8_t>(type);
+  CORRADE_INTERNAL_ASSERT(idx < CubeMap::numTextureTypes);
+  return idx;
+}
```
mosra, line 123: I thought this would wrap the `textures_` access, to avoid accidentally accessing the member directly without a range check. That will also be shorter -- instead of

```cpp
textures_[textureIndex(type)]
```

everywhere, you'd do just

```cpp
getTexture(type)
```

for example. For even more "security" you could rename `textures_` to `textures_DO_NOT_TOUCH_` :D

src/esp/gfx/CubeMap.h
bigbike: @mosra : Honestly, I knew this trick. And what I did before was 1) I would not use the number, 2) I would just use the `enum` instead of the `enum class` so that you do `texture_[TextureType::Color]` directly without cast.

Later, Facebook coding convention does not allow me to do it. All the items within the enum has to have a number.  And later, in the habitat project, `enum` is not allowed, and I can only use `enum class`.

src/esp/gfx/CubeMap.h
bigbike: Yes, in terms of code, it is just 2 places.
But in terms of usage, I hope to avoid casting it every time even it is light operation.

src/esp/gfx/CubeMap.h
mosra: > Facebook coding convention requires me to put a number to each item in the enum

It's actively harmful in this case and results in extra maintenance burden, so I'd suggest to not follow that rule here, no matter how they insist on that. `enum class` is fine tho.

src/esp/gfx/CubeMap.h
mosra: I made the comment because this extra variable made the code harder to understand for me -- I had to check what exactly it was, if it really matched what's in the enum, and what was the possibility for it to get out of sync, which then led to the other comment below.

So in my personal opinion just a congitive overhead -- the cast is *a bit* of extra typing but immediately clear to whoever reads the code, and the code gets read much more often than it gets written ;)

bigbike:
@mosra : Updated. Take a look.

mosra (approved):
Aaand we passed the 100 comment milestone! :tada: :see_no_evil: :sweat_smile:

bigbike:
>Aaand we passed the 100 comment milestone! tada see_no_evil sweat_smile

Yeah! Cannot achieve it without your patience and invaluable ideas, comments! Thanks a lot!

PR #1013, opened 2021-01-08
add smooth moving/looking in viewer

charliecb:
## Motivation and Context

Previously, the viewer controls to move and look-around moved the agent in discrete steps, and multiple controls could not be used at the same time. To rationalize this, holding "W" in the viewer had the same effect as doing so in a text editor: the agent moves forward once, pauses, and then stutters forward.

To obtain smoother movement and looking-around, this feature does 3 things:
1. Holding movement keys now results in gliding forward smoothly
2. Roughly, the rate of movement is independent of frame rate (i.e. it is nearly constant in the world time)
3. Multiple movement/look-around keys can be handled simultaneously

## How Has This Been Tested

Only ad-hoc testing so far. I am concerned that physics may behave slightly differently around pauses (spacebar) and frame-by-frame stepping (.) now, since a minor change was made here. I believe that this was a slight fix, but I don't fully understand the original code, so I could have misinterpreted the situation.

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -2,8 +2,8 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 
-#include <stdlib.h>
 #include <math.h>
+#include <stdlib.h>
 #include <ctime>
 
 #include <Magnum/configure.h>
@@ -52,9 +52,10 @@
 #include "ObjectPickingHelper.h"
 #include "esp/physics/configure.h"
 
-constexpr float moveSensitivity = 0.08f;
-constexpr float lookSensitivity = 0.8f;
+constexpr float moveSensitivity = 0.07f;
+constexpr float lookSensitivity = 0.9f;
 constexpr float rgbSensorHeight = 1.5f;
+constexpr float agentActionsPerSecond = 60.0f;
 
 // for ease of access
 namespace Cr = Corrade;
@@ -81,20 +82,14 @@ class Viewer : public Mn::Platform::Application {
   explicit Viewer(const Arguments& arguments);
 
  private:
-
-  // Keys for moving/looking are recorded according to whether they are currently being pressed
+  // Keys for moving/looking are recorded according to whether they are
+  // currently being pressed
   std::map<KeyEvent::Key, bool> keysPressed = {
-    {KeyEvent::Key::Left, false},
-    {KeyEvent::Key::Right, false},
-    {KeyEvent::Key::Up, false},
-    {KeyEvent::Key::Down, false},
-    {KeyEvent::Key::A, false},
-    {KeyEvent::Key::D, false},
-    {KeyEvent::Key::S, false},
-    {KeyEvent::Key::W, false},
-    {KeyEvent::Key::X, false},
-    {KeyEvent::Key::Z, false}
-  };
+      {KeyEvent::Key::Left, false}, {KeyEvent::Key::Right, false},
+      {KeyEvent::Key::Up, false},   {KeyEvent::Key::Down, false},
+      {KeyEvent::Key::A, false},    {KeyEvent::Key::D, false},
+      {KeyEvent::Key::S, false},    {KeyEvent::Key::W, false},
+      {KeyEvent::Key::X, false},    {KeyEvent::Key::Z, false}};
 
   void drawEvent() override;
   void viewportEvent(ViewportEvent& event) override;
@@ -682,12 +677,10 @@ void Viewer::drawEvent() {
   Mn::GL::defaultFramebuffer.clear(Mn::GL::FramebufferClear::Color |
                                    Mn::GL::FramebufferClear::Depth);
 
-  // Move and look actions should occur at a rate of 60 per second,
-  //  which is the optimal frame rate (so 1 event per frame in the optimal case)
-  //  In the non-optimal case, the number of actions per frame is given by repetitionsToPerform:
+  // Agent actions should occur at a fixed rate per second
   timeSinceLastSimulation += timeline_.previousFrameDuration();
-  int repetitionsToPerform = timeSinceLastSimulation * 60.0;
-  moveAndLook(repetitionsToPerform);
+  int numAgentActions = timeSinceLastSimulation * agentActionsPerSecond;
+  moveAndLook(numAgentActions);
 
   // occasionally a frame will pass quicker than 1/60 seconds
   if (timeSinceLastSimulation >= 1.0 / 60.0) {
@@ -799,8 +792,6 @@ void Viewer::drawEvent() {
 }
 
 void Viewer::moveAndLook(int repetitions) {
-  if (repetitions <= 2) {repetitions = 1;}    // for visual smoothness, for small deviations around 60fps, we default to repetitions=1.
-
   for (int i = 0; i < repetitions; i++) {
     if (keysPressed[KeyEvent::Key::Left]) {
       defaultAgent_->act("turnLeft");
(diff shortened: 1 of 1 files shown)


erikwijmans (commented):
The downside of this approach is that unless the agent is super fast (like it is now) it has the effect of doing the same thing as disabling sliding, which makes human control much rougher (the agent will get stuck on any little thing).  One way around this is to instead of taking lots of little steps, still take the big steps but interpolate the agent between the current state and the next state. Magnum does already have quat slerp for us so it should be fairly straight forward.

src/utils/viewer/viewer.cpp
```diff
@@ -769,6 +798,44 @@ void Viewer::drawEvent() {
...
+    }
+    if (keysPressed[KeyEvent::Key::W]) {
+      defaultAgent_->act("moveForward");
+    }
+    if (keysPressed[KeyEvent::Key::X]) {
+      defaultAgent_->act("moveDown");
+    }
+    if (keysPressed[KeyEvent::Key::Z]) {
+      defaultAgent_->act("moveUp");
+    }
```
jturner65, line 835: recAgentLocation() should be called in here, as in the original movement cases.  At the end of this section, if any of the movement actions have executed (if the agent's location has changed in any way), recAgentLocation() should be called.

src/utils/viewer/viewer.cpp
```diff
@@ -1048,6 +1078,25 @@ void Viewer::keyPressEvent(KeyEvent& event) {
     default:
       break;
   }
+
+  // Update map of moving/looking keys which are currently pressed
+  for (const auto& [k, b] : keysPressed) {
+    if (key == k) {
+      keysPressed[key] = true;
+    }
+  }
```
jturner65, line 1087: There will only ever be one key pressed or released per call to keyPressEvent or keyReleaseEvent, so the loops should not be needed.

jturner65 (changes requested):
Added a few change requests.  Nice first PR :)

src/utils/viewer/viewer.cpp
jturner65: use 
`if (keyPressed.count(key) > 0)`
instead of the loop and if check.

src/utils/viewer/viewer.cpp
```diff
@@ -1048,6 +1078,25 @@ void Viewer::keyPressEvent(KeyEvent& event) {
...
+}
+
+void Viewer::keyReleaseEvent(KeyEvent& event) {
+  // Update map of moving/looking keys which are currently pressed
+  const auto key = event.key();
+  for (const auto& [k, b] : keysPressed) {
+    if (key == k) {
+      keysPressed[key] = false;
+    }
+  }
```
jturner65, line 1099: you should use
`if (keyPressed.count(key) > 0)`
instead of the loop and if check here too.

jturner65:
Won't this PR also modify the agent's speed? If the user is pressing multiple movement keys, the agent will be going unit velocity along each axis.  Is this desirable? @aclegg3 Thoughts?

jturner65:
Won't this PR also modify the agent's speed? If the user is pressing multiple movement keys, won't the agent move in unit speed along each axis?  This means the agent's linear velocity is going to vary depending on heading.  Is this desirable? @aclegg3 Thoughts?

jturner65:
Won't the functionality in this PR cause the agent to move with variable speed? If the user is pressing multiple movement keys, the agent will move in unit speed along each axis of key pressed.  This means the magnitude of the agent's linear velocity is going to vary depending on heading.  I'm not sure if this is a big deal or not, but is this desirable? @aclegg3 Thoughts?

aclegg3 (commented):
Overall, these changes feel great. They seem to work well with low FPS also. Happy to approve once other concerns are addressed. 👍

aclegg3:
> The downside of this approach is that unless the agent is super fast (like it is now) it has the effect of doing the same thing as disabling sliding, which makes human control much rougher (the agent will get stuck on any little thing). 

I'm actually not finding this to be the case in practice. If you reduce the `moveSensitivity` configured here by 10x to 0.008 the sliding still feels good. I think the current approach should be fine for interactive applications.

eundersander:
>The downside of this approach is that unless the agent is super fast (like it is now) it has the effect of doing the same thing as disabling sliding, which makes human control much rougher (the agent will get stuck on any little thing).

>I'm actually not finding this to be the case in practice. If you reduce the moveSensitivity configured here by 10x to 0.008 the sliding still feels good. I think the current approach should be fine for interactive applications.

I tried this branch just now and I think the sliding feels good. For example, if you walk almost straight at a wall, you slide very slowly, but if you walk almost parallel to the wall (slightly into it), you walk almost full speed.

erikwijmans:
Yeah, seems like a 10x reduction in speed is enough to not make it feel speedy but not enough to effectively disable sliding.  Disregard the comment above :)

src/utils/viewer/viewer.cpp
```diff
@@ -769,6 +798,44 @@ void Viewer::drawEvent() {
   redraw();
 }
 
+void Viewer::moveAndLook(int repetitions) {
+  if (repetitions <= 2) {repetitions = 1;}    // for visual smoothness, for small deviations around 60fps, we default to repetitions=1.
```
eundersander, line 802: We discussed offline and agreed this isn't necessary, so I expect you'll remove this line.

src/utils/viewer/viewer.cpp
```diff
@@ -664,13 +682,24 @@ void Viewer::drawEvent() {
...
+  // Move and look actions should occur at a rate of 60 per second,
+  //  which is the optimal frame rate (so 1 event per frame in the optimal case)
+  //  In the non-optimal case, the number of actions per frame is given by repetitionsToPerform:
   timeSinceLastSimulation += timeline_.previousFrameDuration();
-  if (timeSinceLastSimulation >= 1.0 / 60.0 &&
-      (simulating_ || simulateSingleStep_)) {
-    simulator_->stepWorld(1.0 / 60.0);
-    timeSinceLastSimulation = 0.0;
-    simulateSingleStep_ = false;
+  int repetitionsToPerform = timeSinceLastSimulation * 60.0;
```
eundersander, line 689: Let's rename this "60" magic number: `constexpr float agentActionsPerSecond = 60.0f`. And lets rename `repetitionsToPerform` to `numAgentActions`.

src/utils/viewer/viewer.cpp
```diff
@@ -769,6 +791,48 @@ void Viewer::drawEvent() {
...
+      recAgentLocation();
+    }
+    if (keysPressed[KeyEvent::Key::X]) {
+      defaultAgent_->act("moveDown");
+      recAgentLocation();
+    }
+    if (keysPressed[KeyEvent::Key::Z]) {
+      defaultAgent_->act("moveUp");
+      recAgentLocation();
+    }
```
jturner65, line 832: This is very close, but you don't really want to record the movements quite like this - we don't want to record multiple locations in a single pass, considering these changes together are all reflecting a desired single movement.  We'd prefer this to be a kind of latch, where any movement would set a flag that would then prompt the recAgentLocation method.

In other words, set a boolean to true if any of the changes happen, and if the bool is true then call recAgentLocation() 1 time.

So, at line 808, something like 
`bool moved = false;`

and then in every if check : 
`moved = true;`

and finally, after the Z check is completed (line 833) : 

```
if moved {
     recAgentLocation();
}
```

jturner65 (changes requested):
Very close.  Just one more change, please.

src/utils/viewer/viewer.cpp
charliecb: Ah, got it--sorry about that. Just checking: if the agent is moved several times in a single frame due to lag (i.e. repetitions is greater than 1), should we record the agent's location at each step between the frames, or only once per frame? I assumed the former, which I believe corresponds to calling recAgentLocation inside the for loop, and just pushed a commit with this.

jturner65 (approved):
Looks good!

src/utils/viewer/viewer.cpp
jturner65: Yes, that is the correct mechanism.  In viewer, these locations are being recorded so that a demonstration of the trajectory recording/traj-pipe synthesis process can be provided to the user.   We'd want to preserve every location the agent comes to rest in.

aclegg3 (approved):
LGTM.

PR #1018, opened 2021-01-11
Pytest bugfix and PyTorch dependency softening

Skylion007:
## Motivation and Context

Fixes two bugs. One in our hypothesis code that only generated a single navigable point for our hypothesis testing, and softened some hard dependency on PyTorch that were not necessary.

This should resolve the CircleCI issues.

```
E                               hypothesis.errors.FailedHealthCheck: tests/test_snap_point.py::test_snap_point uses the 'test_data' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.
E                               See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.function_scoped_fixture to the suppress_health_check settings for this test.
```
Error caused by new version of Hypothesis posted above.

## How Has This Been Tested
Locally with PyTest in an environment that doesn't have PyTorch installed

## Types of changes

## Checklist

@erikwijmans Not sure if this is actually supposed to be evaluated on the same point or not. I can add a separate hypothesis test that operates on a fixed random point if that is what is desired.

### DIFF
--- a/habitat_sim/sensors/noise_models/no_noise_model.py
+++ b/habitat_sim/sensors/noise_models/no_noise_model.py
@@ -9,7 +9,6 @@ from typing import Union
 import attr
 import numpy as np
 from numpy import ndarray
-from torch import Tensor
 
 from habitat_sim.registry import registry
 from habitat_sim.sensor import SensorType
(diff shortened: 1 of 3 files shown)


jturner65 (approved):
LGTM!

erikwijmans (approved):
Sure, no idea why hypothesis just started complaining about this

eundersander (approved):
LGTM

PR #1025, opened 2021-01-14
[viewer.cpp] Saving/loading camera position/orientation (support for more general scene nodes)

charliecb:
## Motivation and Context

This feature has been requested by Yili, and he wrote the original code in the branch demo-pbr-before-after.

This adds functionality to save and load the positions of scene nodes, particularly the camera. Currently, the camera position/orientation can be saved to the filesystem by its timestamp for future use. Saved camera position/orientation can be loaded from the filesystem with a timestamp specified in the command line (or else if the timestamp is not specified, from the last save in the current instance of viewer).

In practicality, the camera is usually locked on the agent. So in order for loading camera position/orientation to take effect, a feature called flying camera mode is introduced, in which the camera is unlocked from the agent.

## How Has This Been Tested

Handling and logging for the following errors:
1. Filesystem is not configured properly (directory to store saves doesn't exist)
2. Saved data file was corrupted and no longer represents a valid transformation matrix
3. Other file system errors (lack of permission, etc)
4. User does not provide filename to load data from, or attempts to load a file before having saved one

Ad-hoc testing for loading camera transformations (e.g. robust against pauses in simulation, etc).

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -405,7 +405,8 @@ Viewer::Viewer(const Arguments& arguments)
       .setHelp("recompute-navmesh",
                "Programmatically re-generate the scene navmesh.")
       .addOption("camera-transform-filepath")
-      .setHelp("camera", "Specify path to load camera transform from.")
+      .setHelp("camera-transform-filepath",
+               "Specify path to load camera transform from.")
       .parse(arguments.argc, arguments.argv);
 
   const auto viewportSize = Mn::GL::defaultFramebuffer.viewport().size();
@@ -583,9 +584,9 @@ void Viewer::switchCameraType() {
 }
 
 void Viewer::saveCameraTransformToFile() {
-  const char* saved_transformations_directory_ = "saved_transformations/";
-  if (!Cr::Utility::Directory::exists(saved_transformations_directory_)) {
-    Cr::Utility::Directory::mkpath(saved_transformations_directory_);
+  const char* saved_transformations_directory = "saved_transformations/";
+  if (!Cr::Utility::Directory::exists(saved_transformations_directory)) {
+    Cr::Utility::Directory::mkpath(saved_transformations_directory);
   }
 
   // update temporary save
@@ -595,7 +596,7 @@ void Viewer::saveCameraTransformToFile() {
   saveNodeTransformToFile(
       renderCamera_->node(),
       Cr::Utility::formatString("{}camera.{}.txt",
-                                saved_transformations_directory_,
+                                saved_transformations_directory,
                                 getCurrentTimeString()));
 }
 
@@ -605,12 +606,12 @@ void Viewer::loadCameraTransformFromFile() {
     loadNodeTransformFromFile(renderCamera_->node(), cameraLoadPath_);
   } else {
     // attempting to load from last temporary save
-    LOG(INFO)
-        << "Note: Camera transform file not specified, attempting to load from "
+    LOG(WARNING)
+        << "Camera transform file not specified, attempting to load from "
            "current instance. Use --camera-transform-filepath to specify file "
            "to load from.";
     if (!lastCameraSave_) {
-      LOG(WARNING) << "No transformation saved in current instance.";
+      LOG(ERROR) << "No transformation saved in current instance.";
       return;
     }
 
@@ -620,7 +621,7 @@ void Viewer::loadCameraTransformFromFile() {
   }
 
   if (!flyingCameraMode_) {
-    LOG(INFO) << "Note: the flying camera mode is currently OFF. You may not "
+    LOG(INFO) << "Note: The flying camera mode is currently OFF. You may not "
                  "see any view change.";
   }
 }
@@ -629,7 +630,7 @@ void Viewer::saveNodeTransformToFile(esp::scene::SceneNode& node,
                                      const std::string& filename) {
   std::ofstream file(filename);
   if (!file.good()) {
-    LOG(WARNING) << "Cannot open " << filename << " to output data.";
+    LOG(ERROR) << "Cannot open " << filename << " to output data.";
     return;
   }
 
@@ -649,7 +650,7 @@ void Viewer::loadNodeTransformFromFile(esp::scene::SceneNode& node,
                                        const std::string& filename) {
   std::ifstream file(filename);
   if (!file.good()) {
-    LOG(WARNING) << "Cannot open " << filename << " to load data.";
+    LOG(ERROR) << "Cannot open " << filename << " to load data.";
     return;
   }
 
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -5,6 +5,7 @@
 #include <math.h>
 #include <stdlib.h>
 #include <ctime>
+#include <fstream>
 
 #include <Magnum/configure.h>
 #include <Magnum/ImGuiIntegration/Context.hpp>
@@ -147,6 +148,13 @@ class Viewer : public Mn::Platform::Application {
   void switchCameraType();
   Mn::Vector3 randomDirection();
 
+  void saveCameraTransformToFile();
+  void saveNodeTransformToFile(esp::scene::SceneNode& node,
+                               std::string filename);
+  void loadCameraTransformFromFile();
+  void loadNodeTransformFromFile(esp::scene::SceneNode& node,
+                                 std::string filename);
+
   //! string rep of time when viewer application was started
   std::string viewerStartTimeString = getCurrentTimeString();
   void screenshot();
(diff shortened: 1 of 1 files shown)


src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+    Cr::Utility::Directory::mkpath(saved_transformations_directory);
+  }
+
+  // update temporary save
+  lastCameraSave_ = renderCamera_->node().absoluteTransformation();
+  cameraSaved = true;
+
+  // update save in file system
+  saveNodeTransformToFile(renderCamera_->node(),
+                          saved_transformations_directory + "camera." +
```
bigbike, line 563: you can use:
 [Utility::format()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#a5b85466788fc3d71b143909dda1ef44a),

src/utils/viewer/viewer.cpp
```diff
@@ -374,6 +391,11 @@ Viewer::Viewer(const Arguments& arguments)
       .addBooleanOption("recompute-navmesh")
       .setHelp("recompute-navmesh",
                "Programmatically re-generate the scene navmesh.")
+      .addOption("camera")
+      .setHelp("camera",
+               "Specify time to load camera transformation from. Format: "
```
bigbike, line 396: This is "too complicated" for the user.
You can just provide a string as the "full path" to the file to be loaded.
e.g., user can just provide: "~/models/hahaha/cameraTransform.txt"

src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+    // loading from file system
+    loadNodeTransformFromFile(renderCamera_->node(),
+                              saved_transformations_directory + "camera." +
+                                  cameraLoadTimeString_ + ".txt");
+  } else {
+    // attempting to load from last temporary save
+    LOG(INFO) << "Camera transform file not specified, attempting to load from "
+                 "current instance. Use --camera to specify file to load from.";
+    if (!cameraSaved) {
+      LOG(INFO) << "No transformation saved in current instance. Save a "
```
bigbike, line 584: Too verbose.
I would recommend to remove "Save a transformation before trying to load one.".

src/utils/viewer/viewer.cpp
```diff
@@ -321,10 +332,16 @@ Key Commands:
   esp::scene::SceneGraph* activeSceneGraph_ = nullptr;
   bool drawObjectBBs = false;
 
+  const std::string saved_transformations_directory = "saved_transformations/";
+  std::string cameraLoadTimeString_;
+  bool cameraSaved = false;
```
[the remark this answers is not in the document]
bigbike: Second, you may not need this boolean because
you can specify `lastCameraSave_` as a [Corrade::Containers::Optional ](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1Optional.html)

src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+  }
+
+  if (!cameraLoadTimeString_.empty()) {
+    // loading from file system
+    loadNodeTransformFromFile(renderCamera_->node(),
+                              saved_transformations_directory + "camera." +
+                                  cameraLoadTimeString_ + ".txt");
+  } else {
+    // attempting to load from last temporary save
+    LOG(INFO) << "Camera transform file not specified, attempting to load from "
```
bigbike, line 581: Start with: `Note:` or `Warning:`

src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+    for (int row = 0; row < 4; ++row) {
+      file >> cols[col][row];
+    }
+  }
+  Mn::Matrix4 transform{cols[0], cols[1], cols[2], cols[3]};
+  file.close();
+
+  // checking for corruption
+  if (!transform.isRigidTransformation()) {
+    LOG(INFO) << "File error: Data loaded from " << filename
```
bigbike, line 641: `File error:` --> `Warning: `

src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+              << " is not a valid transformation.";
+    return;
+  }
+
+  LOG(INFO) << "Transformation matrix loaded from " << filename << " : "
```
bigbike, line 646: I would place this LOG after the actual transformation is set.

src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+  cameraSaved = true;
+
+  // update save in file system
+  saveNodeTransformToFile(renderCamera_->node(),
+                          saved_transformations_directory + "camera." +
+                              getCurrentTimeString() + ".txt");
+}
+
+void Viewer::loadCameraTransformFromFile() {
+  if (!Cr::Utility::Directory::exists(saved_transformations_directory)) {
```
bigbike, line 568: Like I said, the user should have the freedom to load any file store anywhere in her file system.

src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+    if (!cameraSaved) {
+      LOG(INFO) << "No transformation saved in current instance. Save a "
+                   "transformation before trying to load one.";
+      return;
+    }
+
+    LOG(INFO) << "Transformation matrix loaded from current instance : "
+              << Eigen::Map<esp::mat4f>(lastCameraSave_.data());
+
+    renderCamera_->node().setTransformation(lastCameraSave_);
```
bigbike, line 592: Meanwhile would you need to upgrade the `lastCameraSave_`?

src/utils/viewer/viewer.cpp
```diff
@@ -525,6 +549,106 @@ void Viewer::switchCameraType() {
...
+  }
+
+  if (!flyingCameraMode_) {
+    LOG(INFO) << "Note: flying camera mode is OFF, so loading camera transform "
```
bigbike, line 596: "Note: the flying camera mode is currently OFF. You may not see any view change."

bigbike (changes requested):
Please also merge the latest master branch to this one, and resolve the conflicts if any.

src/utils/viewer/viewer.cpp
```diff
@@ -321,10 +332,16 @@ Key Commands:
   esp::scene::SceneGraph* activeSceneGraph_ = nullptr;
   bool drawObjectBBs = false;
 
+  const std::string saved_transformations_directory = "saved_transformations/";
+  std::string cameraLoadTimeString_;
```
bigbike, line 336: Can it be a local variable instead?

src/utils/viewer/viewer.cpp
```diff
@@ -144,6 +145,13 @@ class Viewer : public Mn::Platform::Application {
   void switchCameraType();
   Mn::Vector3 randomDirection();
 
+  void saveCameraTransformToFile();
+  void saveNodeTransformToFile(esp::scene::SceneNode& node,
+                               std::string filename);
+  void loadCameraTransformFromFile();
+  void loadNodeTransformFromFile(esp::scene::SceneNode& node,
+                                 std::string filename);
```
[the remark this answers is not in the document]
eundersander: er, `const std::string&`

src/utils/viewer/viewer.cpp
eundersander: Agree, and I would rename the command-line option from "camera" to something like "camera-transform-filepath".

src/utils/viewer/viewer.cpp
eundersander: I also suggest using `LOG(INFO)`, `LOG(WARNING)`, or `LOG(ERROR)` based on your intention. Users can choose to filter their debug output based on these categories.

src/utils/viewer/viewer.cpp
charliecb: Okay, my fix was to simply remove this conditional, because if the file doesn't exist it'll be caught by file.good() later. Let me know if something else would be better.

src/utils/viewer/viewer.cpp
charliecb: I'm not sure what you mean, do you mind explaining more?

src/utils/viewer/viewer.cpp
bigbike: That is a bit too complicated, in my opinion.
Just "camera-transform-filepath" is good enough.
The save directory can just use what was implemented here.

src/utils/viewer/viewer.cpp
bigbike: Oh, I was wrong! My bad.
This branch is to handle the setting when the external file is missing.

src/utils/viewer/viewer.cpp
bigbike: You can consider what Eric suggested, "LOG(WARNING)".

bigbike (approved):
Looks good to me.

charliecb:
Confirming that transformation loading method has been tested in all cases.

bigbike:
It seems it failed the CI test. Please fix it before committing it in.

PR #1026, opened 2021-01-14
[PYTHON_BINDINGS] Expose navmesh triangles to Python

Skylion007:
## Motivation and Context

It allows Python code to directly get the triangles from the NavMesh, before it was impossible to get NavMesh data into Python directly. This will add a property to pathfinders that is an array of 3D arrays indicating the triangles.

## How Has This Been Tested
Locally, it builds properly 

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/ShortestPathBindings.cpp
+++ b/src/esp/bindings/ShortestPathBindings.cpp
@@ -100,8 +100,8 @@ void initShortestPathBindings(py::module& m) {
           "navmesh_vertices",
           [](PathFinder& self) { return self.getNavMeshData()->vbo; })
       .def_property_readonly(
-          "navmesh_normals",
-          [](PathFinder& self) { return self.getNavMeshData()->nbo; })
+          "navmesh_vertex_indices",
+          [](PathFinder& self) { return self.getNavMeshData()->ibo; })
       .def("load_nav_mesh", &PathFinder::loadNavMesh)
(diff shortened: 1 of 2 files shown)


src/esp/bindings/ShortestPathBindings.cpp
```diff
@@ -95,6 +96,12 @@ void initShortestPathBindings(py::module& m) {
       .def("island_radius", &PathFinder::islandRadius, "pt"_a)
       .def_property_readonly("is_loaded", &PathFinder::isLoaded)
       .def_property_readonly("navigable_area", &PathFinder::getNavigableArea)
+      .def_property_readonly(
```
erikwijmans, line 99: These should probably be methods to discourage them from being called in a tight loop

Skylion007: Haven't heard that argument before... we could cache it potentially?

erikwijmans: The general expectation of properties IMO is that they are lightweight,

```python
for vid in vids:
    pathfinder.navmesh_vertices[vid]
```

seems fine while
```python
for vid in vids:
    pathfinder.build_navmesh_vertices()[vid]
```

definitely doesn't.


Yeah, caching would be good.

Skylion007: @erikwijmans Done

erikwijmans (approved):

PR #1034, opened 2021-01-18
Scene Import 1.0

jturner65:
## Motivation and Context
This PR replaces [the old version found here](https://github.com/facebookresearch/habitat-sim/pull/954) since much of the old PR was split out into individual PRs that have been merged already.

This PR adds the support for importing scene instances.  A scene instance describes a scene layout, using handles to attributes for the various assets used that are specified/defined in a scene dataset configuration file.  The scene instance can describe the assets and transformations needed to build instances of a stage, 1 or more objects, a lighting configuration, navmesh and semantic scene descriptor files. 

The stage and object instances described in a scene instance include their necessary transformations to align them properly in the scene.  For objects, it also includes the motion type the object possesses, and whether or not the position of the placed object should be measured from the COM or from some asset-related origin.  

All of Scene Instance Import functionality is driven by data specified in SimulatorConfiguration.

While this functionality is designed as part of the scene dataset configuration functionality that has been added, it is fully backwards compatible with the current scene/stage loading mechanism, and intended to remain so.  If the user does not specify an existing scene dataset configuration, the system uses a default that can be modified and expanded (and in the next version of this functionality, saved as well).  If the user attempts to load a stage via reconfigure, this functionality will internally construct a scene instance containing that stage and add it to the default dataset.

Future updates to this process are going to include the ability to save stage and object arrangements from within habitat-sim as scene instance configurations, as well as saving scene dataset configurations constructed/modified within habitat.

## Types of changes

## Checklist

### DIFF
--- a/data/test_assets/dataset_tests/scenes/dataset_test_scene.scene_instance.json
+++ b/data/test_assets/dataset_tests/scenes/dataset_test_scene.scene_instance.json
@@ -1,15 +1,18 @@
 {
-	"translation_origin":"asset_local",
+	"translation_origin":"com",
 	"stage_instance":{
+		"translation_origin":"asset_local",
 		"template_name":"modified_test_stage",
 		"translation":[1.1,2.2,3.3]
 	},
 	"object_instances":[
 	{
+		"translation_origin":"asset_local",
 		"template_name":"dataset_test_object1",
 		"translation":[0.1,0.2,0.3]
 	},
 	{
+		"translation_origin":"asset_local",
 		"template_name":"dataset_test_object2",
 		"translation":[0.3,0.4,0.5]
 	}
--- a/data/test_assets/dataset_tests/scenes/dataset_test_scene.scene_instance.json
+++ b/data/test_assets/dataset_tests/scenes/dataset_test_scene.scene_instance.json
@@ -1,4 +1,5 @@
 {
+	"translation_origin":"asset_local",
 	"stage_instance":{
 		"template_name":"modified_test_stage",
 		"translation":[1.1,2.2,3.3]
@@ -14,6 +15,7 @@
 	}
 	],
 	"default_lighting":"modified_test_lights",
+	"frustum_culling":true,
 	"navmesh_instance":"navmesh_path1",
 	"semantic_scene_instance":"semantic_descriptor_path1"
 }
--- a/data/test_assets/dataset_tests/scenes/dataset_test_scene.scene_instance.json
+++ b/data/test_assets/dataset_tests/scenes/dataset_test_scene.scene_instance.json
@@ -18,7 +18,6 @@
 	}
 	],
 	"default_lighting":"modified_test_lights",
-	"frustum_culling":true,
 	"navmesh_instance":"navmesh_path1",
 	"semantic_scene_instance":"semantic_descriptor_path1"
 }
--- a/src/esp/metadata/MetadataMediator.cpp
+++ b/src/esp/metadata/MetadataMediator.cpp
@@ -15,17 +15,6 @@ MetadataMediator::MetadataMediator(const sim::SimulatorConfiguration& cfg)
   setSimulatorConfiguration(simConfig_);
 }  // MetadataMediator ctor (SimulatorConfiguration)
 
-MetadataMediator::MetadataMediator(
-    const std::string& _activeSceneDataset,
-    const std::string& _physicsManagerAttributesPath)
-    : activeSceneDataset_(_activeSceneDataset),
-      currPhysicsManagerAttributes_(_physicsManagerAttributesPath),
-      simConfig_(sim::SimulatorConfiguration{}) {
-  simConfig_.sceneDatasetConfigFile = _activeSceneDataset;
-  simConfig_.physicsConfigFile = _physicsManagerAttributesPath;
-  buildAttributesManagers();
-  setSimulatorConfiguration(simConfig_);
-}  // MetadataMediator ctor
 void MetadataMediator::buildAttributesManagers() {
   physicsAttributesManager_ = managers::PhysicsAttributesManager::create();
 
--- a/src/esp/metadata/MetadataMediator.cpp
+++ b/src/esp/metadata/MetadataMediator.cpp
@@ -327,10 +327,10 @@ attributes::SceneAttributes::ptr MetadataMediator::makeSceneAndReferenceStage(
   // create scene attributes with passed name
   attributes::SceneAttributes::ptr sceneAttributes =
       dsSceneAttrMgr->createDefaultObject(sceneName, false);
-  // create stage instance attributes and set its name (scene and stage here
-  // share same name)
+  // create stage instance attributes and set its name (from stage attributes)
   sceneAttributes->setStageInstance(
-      dsSceneAttrMgr->createEmptyInstanceAttributes(sceneName));
+      dsSceneAttrMgr->createEmptyInstanceAttributes(
+          stageAttributes->getHandle()));
 
   // The following is to manage stage files that have navmesh and semantic scene
   // descriptor ("house file") handles in them. This mechanism has been
--- a/src/esp/metadata/MetadataMediator.cpp
+++ b/src/esp/metadata/MetadataMediator.cpp
@@ -37,6 +37,7 @@ void MetadataMediator::buildAttributesManagers() {
   createSceneDataset("default");
   // should always have default physicsManagerAttributesPath
   createPhysicsManagerAttributes(ESP_DEFAULT_PHYSICS_CONFIG_REL_PATH);
+  // after this setSimulatorConfiguration will be called
 }  // MetadataMediator::buildAttributesManagers
 
 bool MetadataMediator::setSimulatorConfiguration(
@@ -58,7 +59,8 @@ bool MetadataMediator::setSimulatorConfiguration(
   success = setCurrPhysicsAttributesHandle(simConfig_.physicsConfigFile);
   if (!success) {
     // something failed about setting up physics attributes
-    LOG(ERROR) << "MetadataMediator::setSimulatorConfiguration : ";
+    LOG(ERROR) << "MetadataMediator::setSimulatorConfiguration : Some error "
+                  "prevented current physics ";
 
     return false;
   }
@@ -287,7 +289,9 @@ attributes::SceneAttributes::ptr MetadataMediator::getSceneAttributesByName(
                    "SceneAttributes with same name and adding to Dataset.";
       // create a new SceneAttributes, and give it a
       // SceneObjectInstanceAttributes for the stage with the same name.
-      sceneAttributes = makeSceneAndReferenceStage(dsSceneAttrMgr, sceneName);
+      sceneAttributes = makeSceneAndReferenceStage(
+          datasetAttr, dsStageAttrMgr->getObjectByHandle(sceneName),
+          dsSceneAttrMgr, sceneName);
 
     } else {
       // 4.  Existing stage config/asset on disk, but not in current dataset.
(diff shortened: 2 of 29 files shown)


src/esp/sim/Simulator.h
```diff
@@ -848,7 +848,63 @@ class Simulator {
...
 
-  //! sample a random valid AgentState in passed agentState
+  /**
+   * @brief Creates a pathfinder and loads the navmesh from the specified
+   * location if it exists, otherwise returns a warning.
+   * @param navmeshFilename The filename of the navmesh to load when pathfinder
+   * is created
+   * @return Whether successful or not
+   */
+  bool createPathfinder(const std::string& navmeshFilename);
```
Skylion007, line 858: Wow, why is the necessary over the existing pathfinder API? shouldn't this be a method on pathfinder? Why should it be its own method. (There are already a lot of duplicate / unused methods on Simulator.h, so we should reuse / refactor when possible.)

src/esp/sim/Simulator.cpp
```diff
@@ -117,171 +126,405 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
...
+  metadata::attributes::SceneAttributes::cptr curSceneInstanceAttributes =
+      setSceneInstanceAttributes(activeSceneName);
+
+  // TODO : reset may eventually have all the scene instance instantiation code
+  // so that scenes can be reset
+  reset();
+  return true;
+}  // Simulator::createSceneInstanceNoRenderer
+
+bool Simulator::createPathfinder(const std::string& navmeshFilename) {
```
Skylion007, line 458: Just confused why this method was abstracted out. Shouldn't default navmesh recomputation be factored out of the constructor too?. This just seems like it does create + load. Maybe these logging should be moved to the pathfinder's loadNavMesh logic somewhere?

Skylion007: Feel like this should live as a static method on the pathfinder class, don't you think? No simulator or pathfinder specific variables are used, feel like this should just be a constructor for Pathfinder that takes in a navmesh string. You can then use pathfinder.isloaded to figure out whether or not it loaded a navmesh.

src/esp/sim/Simulator.cpp
```diff
@@ -117,171 +126,405 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
...
-        } else if (endsWith(houseFilename, ".scn")) {
-          scene::SemanticScene::loadGibsonHouse(houseFilename, *semanticScene_);
+        if (endsWith(semanticSceneDescFilename, ".house")) {
+          present = scene::SemanticScene::loadMp3dHouse(
+              semanticSceneDescFilename, *semanticScene_);
+        } else if (endsWith(semanticSceneDescFilename, ".scn")) {
+          present = scene::SemanticScene::loadGibsonHouse(
+              semanticSceneDescFilename, *semanticScene_);
+        } else {
+          present = false;
```
Skylion007, line 507: Since present defaults to false, are these else ```present = false``` even needed? All these switches are mutually exclusive. Feels like it could mask an issue.

Skylion007 (changes requested):

Skylion007:
Do we have any validate JSON functions, currently? I feel like those would be useful.

src/esp/sim/Simulator.h
jturner65: This is to support the eventual addition of multiple navmeshes, each of which will require their own pathfinder.

src/esp/sim/Simulator.cpp
jturner65: I like that.

jturner65: In an effort to minimize the impact this PR has, I am going to remove the mentioned methods and move the code contained to the only place the method was called.  The future design considerations this was laying the groundwork for are not yet on the horizon anyway.

src/esp/sim/Simulator.cpp
jturner65: Here again I parsed the code into an individual method in the interest of future development, when it does not really need to be split from where it is called right now, so I'll return this code to where it was originally executed.

src/esp/sim/Simulator.cpp
```diff
@@ -117,171 +126,376 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
...
+    physicsManager_->setTranslation(objID, translate);
+    physicsManager_->setRotation(objID, objInst->getRotation());
+    // set object's motion type if different than set value
+    const physics::MotionType attrObjMotionType =
+        static_cast<physics::MotionType>(objInst->getMotionType());
+    const physics::MotionType objMotionType =
+        physicsManager_->getObjectMotionType(objID);
+
+    if ((attrObjMotionType != objMotionType) &&
+        (attrObjMotionType != physics::MotionType::UNDEFINED)) {
```
aclegg3, line 472: The setter will check for equivalency before doing anything, so you could simplify this.

aclegg3 (commented):
Looking good. Some questions and suggestions. Would also like to see some tested example scene instances, maybe a test_dataset created from the test_assets directory?

src/esp/metadata/MetadataMediator.cpp
```diff
@@ -36,39 +117,271 @@ bool MetadataMediator::createDataset(const std::string& sceneDatasetName,
...
+        // SceneObjectInstanceAttributes for the stage with the same name.
+        sceneAttributes = makeSceneAndReferenceStage(
+            datasetAttr, stageAttributes, dsSceneAttrMgr, sceneName);
+      }
+    }
+  }
+  // make sure that all stage, object and lighting attributes referenced in
+  // scene attributes are loaded in dataset, as well as the scene attributes
+  // itself.
+  datasetAttr->addNewSceneInstanceToDataset(sceneAttributes);
```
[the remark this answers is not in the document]
jturner65: That's what the function does.  If it does not find named attributes within the dataset corresponding to those specified in the scene instance, it attempts to create them.   This process will not "fail" per se, but empty results/unfound attributes may not be what the user wishes for.  Or, it may be, depending on the source of the scene instance attributes that has been created.

src/esp/sim/Simulator.cpp
```diff
@@ -79,15 +79,24 @@ void Simulator::close() {
...
-    metadataMediator_ =
-        metadata::MetadataMediator::create(cfg.sceneDatasetConfigFile);
+    metadataMediator_ = metadata::MetadataMediator::create(cfg);
+    LOG(WARNING) << "Simulator::reconfigure : create MM w/ cfg w/setting "
+                 << cfg.overrideSceneLightDefaults;
   } else {
-    metadataMediator_->setActiveSceneDatasetName(cfg.sceneDatasetConfigFile);
+    metadataMediator_->setSimulatorConfiguration(cfg);
+    LOG(WARNING) << "Simulator::reconfigure : set sim MM w/ cfg w/setting "
+                 << cfg.overrideSceneLightDefaults;
```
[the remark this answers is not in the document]
jturner65: Was using a warning to track results in tests.  This can be downgraded to info i guess.

src/esp/metadata/attributes/SceneAttributes.cpp
```diff
@@ -18,10 +18,28 @@ const std::map<std::string, esp::physics::MotionType>
 SceneObjectInstanceAttributes::SceneObjectInstanceAttributes(
     const std::string& handle)
     : AbstractAttributes("SceneObjectInstanceAttributes", handle) {
-  setMotionType(-1);  // defaults to unknown
+  // defaults to unknown/undefined
+  setMotionType(static_cast<int>(esp::physics::MotionType::UNDEFINED));
```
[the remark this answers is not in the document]
jturner65: This is only used to override the default setting already in place.  So this comes into play in the following code from Simulator.cpp (466) :

```
    const physics::MotionType attrObjMotionType =
        static_cast<physics::MotionType>(objInst->getMotionType());
    const physics::MotionType objMotionType =
        physicsManager_->getObjectMotionType(objID);

    if ((attrObjMotionType != objMotionType) &&
        (attrObjMotionType != physics::MotionType::UNDEFINED)) {
      physicsManager_->setObjectMotionType(objID, attrObjMotionType);
    }
```

src/esp/sim/Simulator.cpp
jturner65: I was thinking of doing that, but I wanted it visually obvious to implementers/supporters the exact conditions where this would be executed.

src/esp/sim/Simulator.cpp
```diff
@@ -117,171 +126,376 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
...
+    // some candidate for SSD exists in scene instance attributes.
+    LOG(INFO) << "Simulator::setSceneInstanceAttributes : SceneInstance : "
+              << activeSceneName
+              << " proposed Semantic Scene Descriptor filename : "
+              << semanticSceneDescFilename;
+    // get stage attributes for current scene instance
+    auto stageAttributes = metadataMediator_->getNamedStageAttributesCopy(
+        curSceneInstanceAttributes->getStageInstance()->getHandle());
+    if (stageAttributes != nullptr) {
+      // assetType of stage is used to specify semanctic scene descriptor format
```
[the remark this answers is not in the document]
jturner65: We need to come up with a better/expanded design for handling this, I agree.  I would also prefer not to be tethered to the stage attributes for this - SSD loading should be completely independent of stage and stage type. 

We could modify the check to use the file name given for the ssd, and if that fails to yield a usable type, then fall back on the scene instance. -shrug-.

src/esp/metadata/attributes/SceneAttributes.cpp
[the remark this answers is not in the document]
jturner65: ..which has been changed heh.  Now it will only attempt to change the value if the object instance motion type value is not undefined.
```
    // set object's motion type if different than set value
    const physics::MotionType attrObjMotionType =
        static_cast<physics::MotionType>(objInst->getMotionType());
    if (attrObjMotionType != physics::MotionType::UNDEFINED) {
      physicsManager_->setObjectMotionType(objID, attrObjMotionType);
    }
```

src/esp/sim/Simulator.h
Skylion007: Still should remove the amount stat calls to the file system. Doesn't pathfinder already do a file exists check?

jturner65: I agree that fewer OS hits would be a great thing.  What you see was the existing mechanism for loading navmesh in reconfigure, which is why I have not modified it.  

The loading function does check for file existence, along with multiple other failure conditions, and returns only a boolean success value.  By having this test, we can know for certain whether failure is caused by non-existing file, or some other condition, without having to refactor PathFinder::loadNavMesh, which I felt is a bit beyond the scope of this PR.

src/esp/sim/Simulator.cpp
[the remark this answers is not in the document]
jturner65: The mechanism has been modified somewhat.  If an ssd filename is specified in the scene instance attributes, this is first checked for existence, and if it does exist, it is attempted to be loaded (currently loading is determined by its name).  If what is specified does not load properly or is not found, then the previous implementation (pivoting on stageAttributes type) is executed.

data/test_assets/dataset_tests/scenes/dataset_test_scene.scene_instance.json
```diff
@@ -1,19 +1,24 @@
...
 		"translation":[0.1,0.2,0.3]
 	},
 	{
+		"translation_origin":"asset_local",
 		"template_name":"dataset_test_object2",
 		"translation":[0.3,0.4,0.5]
 	}
 	],
 	"default_lighting":"modified_test_lights",
+	"frustum_culling":true,
```
aclegg3, line 21: I think we decided to remove this from config system and instead just keep SimulatorConfiguration as the entry point.

aclegg3 (approved):
Looks like we're ready. 👍

PR #1036, opened 2021-01-19
[sensor] Embed RenderCamera in CameraSensor

vauduong:
## Motivation and Context
Currently, when a VisualSensor wants to render something, we get the defaultRenderCamera from the SceneGraph and set the transformation, projection matrix, and viewport of the VisualSensor to the defaultRenderCamera. 
SensorCameraWeeklyUpdates 2 pptx
SensorCameraWeeklyUpdates 2 pptx (1)

With this change, renderCamera_ is now a member variable of CameraSensor. Upon initialization of CameraSensor, renderCamera_ is initialized at the same SceneNode as CameraSensor and the transformation, projection matrix, and viewport of the RenderCamera are set to that of the CameraSensor at initialization and upon changes to the specs. 

This is a more user-friendly approach that involves less passing around of variables between the SceneGraph, VisualSensor, and RenderCamera. Viewer gets renderCamera_ from the VisualSensor in the SceneGraph and uses it to render the scene. We also deprecate defaultRenderCamera in the SceneGraph, as now each CameraSensor has its own RenderCamera.

In this PR, I also:
- Remove VisualSensor methods to set attributes of the RenderCamera and set RenderCamera attributes directly
- Fix viewer crashing on viewer::ViewportEvent
- Rename CameraSensor::modZoom to CameraSensor::modifyZoom
- Add comments to clarify methods in CameraSensor
- Update tests to reflect changes

Currently, this PR fails test_examples.py in habitat-lab tests due to a habitat-lab test calling the deprecated function SceneGraph::getDefaultRenderCamera: https://app.circleci.com/pipelines/github/facebookresearch/habitat-sim/4737/workflows/ba78662a-b7bf-4a16-9eda-0d9544ff54a5/jobs/20279

This will be fixed with the following PR in habitat-lab: facebookresearch/habitat-lab#575

## How Has This Been Tested

Built and tested functionality

## Types of changes

## Checklist


bigbike (changes requested):
VisualSensor is an interface class, and its subclass CameraSensor actually does the job. 
The task is to embed the renderCamera in the CameraSensor not in the VisualSensor.

src/esp/sensor/VisualSensor.h
```diff
@@ -108,9 +110,18 @@ class VisualSensor : public Sensor {
...
+   * @brief Returns RenderCamera of CameraSensor
+   */
+  gfx::RenderCamera& getRenderCamera() { return renderCamera_; }
+
  protected:
   std::unique_ptr<gfx::RenderTarget> tgt_;
 
+  /** @brief RenderCamera for sensor
+   */
+  gfx::RenderCamera renderCamera_;
```
erikwijmans, line 123: This will need to be a pointer that doesn't have delete called on it by the destructor.  Magnum will cleanup things like gfx::RenderCamera when the scene graph is destroyed so this is currently being double freed (that is what is causing the assert failure on the CI)

bigbike: @vauduong : To address what @erikwijmans pointed out, I would recommend to use a raw pointer (a member variable of the CameraSensor) to point to a renderCamera.
A good reference is here:
https://github.com/facebookresearch/habitat-sim/compare/cubemap#diff-2e8bced9d5e63f6f91584f0cab332f30fe2c372c654ebfbd6ae6485dbe2d6475R116

src/esp/gfx/Renderer.cpp
```diff
@@ -44,15 +45,14 @@ struct Renderer::Impl {
     }
   }
 
-  void draw(sensor::VisualSensor& visualSensor,
+  void draw(esp::sensor::CameraSensor& cameraSensor,
```
bigbike, line 48: What is the reason to change the VisualSensor to a CameraSensor?

src/esp/sensor/VisualSensor.h
```diff
@@ -108,6 +110,14 @@ class VisualSensor : public Sensor {
     return false;
   }
 
+  /**
+   * @brief Returns RenderCamera
+   */
+  virtual gfx::RenderCamera* getRenderCamera() {
```
erikwijmans, line 116: This should either be pure virtual or it should have a hard assert in it.  Currently this will just segfault later when you dereference the null pointer, which makes debugging harder.

vauduong: Thanks for the feedback!

src/esp/gfx/Renderer.cpp
vauduong: Thanks for the feedback, I was definitely not aware of all the uses of virtual functions in cpp, and have learned a lot! This is fixed now.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -81,9 +81,9 @@ void initGfxBindings(py::module& m) {
       .def(py::init(&Renderer::create<>))
       .def(
           "draw",
-          [](Renderer& self, sensor::VisualSensor& visualSensor,
+          [](Renderer& self, esp::sensor::CameraSensor& cameraSensor,
```
bigbike, line 84: I think you need to undo this.

src/esp/gfx/Renderer.h
```diff
@@ -8,7 +8,7 @@
 #include "esp/core/esp.h"
 #include "esp/gfx/RenderCamera.h"
 #include "esp/scene/SceneGraph.h"
-#include "esp/sensor/VisualSensor.h"
+#include "esp/sensor/CameraSensor.h"
```
bigbike, line 11: This is wrong.

src/tests/GfxReplayTest.cpp
```diff
@@ -121,9 +121,13 @@ TEST(GfxReplayTest, player) {
   // retrieve last child of scene root node
   auto& rootNode = sceneGraph.getRootNode();
   const auto* lastRootChild = rootNode.children().first();
-  ASSERT(lastRootChild);
-  while (lastRootChild->nextSibling()) {
-    lastRootChild = lastRootChild->nextSibling();
+  if (lastRootChild == NULL) {
+    lastRootChild = rootNode;
```
bigbike, line 125: @eundersander :
Maybe it is a naive question. I did not quite understand the `lastRootChild`. What does it do? 
(I am not so sure if Vivian is doing the right thing here.)

src/esp/sensor/CameraSensor.cpp
```diff
@@ -66,18 +67,9 @@ void CameraSensor::recomputeBaseProjectionMatrix() {
...
-
-CameraSensor& CameraSensor::setTransformationMatrix(
-    gfx::RenderCamera& targetCamera) {
-  CORRADE_ASSERT(!scene::SceneGraph::isRootNode(targetCamera.node()),
-                 "CameraSensor::setTransformationMatrix: target camera cannot "
-                 "be on the root node of the scene graph",
-                 *this);
+auto CameraSensor::computeTransformationMatrix() {
+  // target camera cannot be on the root node of the scene graph"
+  ASSERT(!scene::SceneGraph::isRootNode(renderCamera_->node()));
```
bigbike, line 72: You can do it either in this PR or in another PR:
try to convert all the ASSERT to CORRADE_ASSERT in the CameraSensor.cpp

src/esp/sensor/CameraSensor.h
```diff
@@ -23,14 +23,9 @@ class CameraSensor : public VisualSensor {
...
-  virtual CameraSensor& setProjectionMatrix(
-      gfx::RenderCamera& targetCamera) override;
-  // set the transformation matrix to the given render camera
-  virtual CameraSensor& setTransformationMatrix(
-      gfx::RenderCamera& targetCamera) override;
-  // set the view port to the given render camera
-  virtual CameraSensor& setViewport(gfx::RenderCamera& targetCamera) override;
+  virtual gfx::RenderCamera* getRenderCamera() override;
+
+  auto computeTransformationMatrix();
```
bigbike, line 28: figure out what is exactly the return type.
Please do not use auto.

src/tests/GfxReplayTest.cpp
vauduong: Locally, I'm trying to do the following:
```
if (!lastRootChild) {
    lastRootChild = &rootNode;
  } else {
    ASSERT(lastRootChild);
    while (lastRootChild->nextSibling()) {
      lastRootChild = lastRootChild->nextSibling();
    }
  }
```
This works, but later I fail the test here:
```
else if (keyframeIndex == 0) {
      // assert that a new node was created under root
      ASSERT(lastRootChild->nextSibling());
```

Do you have any feedback on understanding this test?

eundersander: >I did not quite understand the lastRootChild. What does it do?
The scene graph has a root node, and the root node in general has a list of N children. lastRootChild is meant to be a pointer to the last child of the root node. The ordering is due to the order in which children are added to the scene graph.

This test doesn't make any assumption about what exactly that last child is, but it does assume that there is at least N=1 children of the root node upon reaching this point in the code. It seems that Vivian's change is causing that assumption to no longer hold (N=0 for some case), hence her change here. That is fine.

Later, I use lastRootChild to test whether an *additional* child of the root node (i.e., does number of children now equal N + 1?). In the case of N=0, the test needs to be done differently.

One way to clean all this up would be to write a helper function, getNumberOfChildrenOfRoot(). You can modify my tests to just verify that the count is increased by 1 where expected.

vauduong: That makes sense, thanks for the clarification!

src/esp/sensor/CameraSensor.cpp
```diff
@@ -20,6 +20,7 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
       baseProjMatrix_(Magnum::Math::IdentityInit),
       zoomMatrix_(Magnum::Math::IdentityInit) {
   setProjectionParameters(spec);
+  renderCamera_ = new gfx::RenderCamera(cameraNode);
```
bigbike, line 23: After you created this renderCamera, you have to set it.
You will have to set the projection matrix, view port once since it will not change during the lifetime of the sensor.
By doing so, you do not have to set the projection matrix again and again in the `getObservation()` function.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -150,6 +140,11 @@ bool CameraSensor::drawObservation(sim::Simulator& sim) {
   if (sim.isFrustumCullingEnabled())
     flags |= gfx::RenderCamera::Flag::FrustumCulling;
 
+  // Set projection matrix, transformation matrix, and viewport
+  renderCamera_->setProjectionMatrix(width_, height_, projectionMatrix_);
```
bigbike: You do not need to set it every time.
You can consider **setting it once if there is any changes** to the projectionMatrix_.
- So you will set it first in the constructor;
- Then you will set it when `projectionMatrix_` needs to be updated. Note that `projectionMatrix_` is recomputed in the function `recomputeProjectionMatrix()`, which means you can call `renderCamera_->setProjectionMatrix(...)` there. 

Careful: after you make the above changes, re-visit the `viewportEvent()` in the viewer.cpp. That function handles the case where the window size is changed (e.g., you use your mouse to enlarge the window). You have to make sure changes on width and height can eventually trigger the re-computation of the projection matrix!!

src/esp/sensor/CameraSensor.cpp
```diff
@@ -150,6 +140,11 @@ bool CameraSensor::drawObservation(sim::Simulator& sim) {
...
+  renderCamera_->node().setTransformation(this->computeTransformationMatrix());
```
bigbike: You do not need to set it every time since the renderCamera and the sensor are attaching to the same node so it always has the same transformation as the sensor!

src/utils/viewer/viewer.cpp
```diff
@@ -536,11 +536,11 @@ Viewer::Viewer(const Arguments& arguments)
 
   // Set up camera
   activeSceneGraph_ = &simulator_->getActiveSceneGraph();
-  renderCamera_ = &activeSceneGraph_->getDefaultRenderCamera();
-  renderCamera_->setAspectRatioPolicy(
-      Mn::SceneGraph::AspectRatioPolicy::Extend);
   defaultAgent_ = simulator_->getAgent(defaultAgentId_);
   agentBodyNode_ = &defaultAgent_->node();
+  renderCamera_ = getAgentCamera().getRenderCamera();
+  renderCamera_->setAspectRatioPolicy(
```
bigbike, line 542: now since the renderCamera is embedded in the sensor. 
 please
 `setAspectRatioPolicy(
      Mn::SceneGraph::AspectRatioPolicy::Extend)`
      in the CameraSensor constructor, after you create the renderCamera.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -150,6 +140,11 @@ bool CameraSensor::drawObservation(sim::Simulator& sim) {
...
+  renderCamera_->setViewport(this->framebufferSize());
```
bigbike: You can leave it as this for now.
But the logic is the same as the projection matrix, namely, you **only need to set it once** if there is any changes

src/esp/sensor/CameraSensor.h
bigbike: I am wondering why you still need this function??

To answer this question, let us first look at another one:
Why do we need a function to set the transformation matrix before? 
Because the camera (the `defaultRenderCamera`) was responsible for the rendering requests from all the sensors. For example, sensor A asked it to render the scene (so you have to set the render camera based on the sensor A's spec), then, sensor B asked it to render the scene, so you have to refresh the render camera settings with the spec from sensor B.
old-camera-design
[Figure illustrates the OLD design]

Now, after this refactoring, this is not the case. Each sensor has its own embedded camera! It moves as the sensor moves, which means the transformation matrix of the render camera is always up to date. You do not have to set it every time in the `getObeservation()` (namely, you do not need: `renderCamera_->node().setTransformation(this->computeTransformationMatrix());`)

And I do not think you need this `computeTransformationMatrix` at all! It can be fully deprecated.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -66,18 +67,9 @@ void CameraSensor::recomputeBaseProjectionMatrix() {
...
-  return *this;
-}
-
-CameraSensor& CameraSensor::setTransformationMatrix(
-    gfx::RenderCamera& targetCamera) {
-  CORRADE_ASSERT(!scene::SceneGraph::isRootNode(targetCamera.node()),
-                 "CameraSensor::setTransformationMatrix: target camera cannot "
-                 "be on the root node of the scene graph",
-                 *this);
+Magnum::Matrix4 CameraSensor::computeTransformationMatrix() {
```
erikwijmans, line 70: I think this entire function no longer needs to exist.  This was a thing when we needed to apply the transform of a given sensor to the camera, but now `renderCamera_->node() == this->node()` so this function should be a no-op.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -66,50 +71,9 @@ void CameraSensor::recomputeBaseProjectionMatrix() {
...
-  targetCamera.node().setTransformation(relativeTransform);
-  return *this;
-}
-
-CameraSensor& CameraSensor::setViewport(gfx::RenderCamera& targetCamera) {
-  targetCamera.setViewport(this->framebufferSize());
-  return *this;
+void CameraSensor::recomputeProjectionMatrix() {
+  projectionMatrix_ = zoomMatrix_ * baseProjMatrix_;
+  // renderCamera_->setProjectionMatrix(width_, height_, projectionMatrix_);
```
bigbike, line 76: why is it commented out?

vauduong: Sorry, was testing something out and forgot to uncomment! Will fix

vauduong:
Still need to fix test_gfx, seems like I put the binding for visual_sensor.get_render_camera in the wrong place

Skylion007:
@vauduong  MyPy is complaining because the object you are modifying is a Sensor not a VisualSensor so mypy is complaining. You need to assert that the sensor is of the PyBind11 instancetype for mypy to be happy. This can be done by adding an  ```assert isinstance(sensor, habitat_sim.sensors.VisualSensor)``` to allow the typing system to downcast it and know that the sensor is a visual_sensor. You could also use the python ```typing.cast``` function, but best to avoid it. Alternatively, you could just silence the error with noqa.

This seems unrelated to the C++ segfault that is currently happening. It's possible you may need to change the PyBind11 return type so that Python can take ownership: https://pybind11.readthedocs.io/en/stable/advanced/functions.html?highlight=return#return-value-policies.

run 
```python -X faulthandler  examples/example.py --scene data/scene_datasets/habitat-test-scenes/van-gogh-room.glb --silent --test_fps_regression $FPS_THRESHOLD``` to see exactly where the segfault is ocuring.

**Comment by erikwijmans** [comment]
You can also pop things into gdb/lldb to debug cpp code called from python, i.e.

```
lldb python
>  r examples/example.py --scene data/scene_datasets/habitat-test-scenes/van-gogh-room.glb --silent --test_fps_regression 200
```

**Comment by Skylion007** [comment]
@vauduong Segfault occurs here:
tests/test_gfx.py Fatal Python error: Segmentation fault

Current thread 0x0000000108582dc0 (most recent call first):
  File "./habitat_sim/agent/agent.py", line 156 in _add_sensor

**Comment by Skylion007** [comment]
@vauduong here is the lldb stack trace:
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x0000000153dbfe09 habitat_sim_bindings.cpython-37m-darwin.so`Magnum::SceneGraph::Camera<3u, float>::setProjectionMatrix(this=0x0005000000000004, matrix=0x000000018e5cc8b0) at Camera.hpp:75:26 [opt]
   72  	}
   73
   74  	template<UnsignedInt dimensions, class T> Camera<dimensions, T>& Camera<dimensions, T>::setProjectionMatrix(const MatrixTypeFor<dimensions, T>& matrix) {
-> 75  	    _rawProjectionMatrix = matrix;
   76  	    fixAspectRatio();
   77  	    return *this;
   78  	}

**Review by Skylion007 — CHANGES_REQUESTED** [review]

File: src/esp/bindings/SensorBindings.cpp [hunk 16]
```diff
@@ -107,7 +107,8 @@ void initSensorBindings(py::module& m) {
                                                                 "VisualSensor")
       .def(
           "get_render_camera", &VisualSensor::getRenderCamera,
```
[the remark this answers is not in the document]
vauduong: From my understanding, render_camera is an instance of RenderCamera, and we want to be able to get the object and call methods on it, rather than just get a value. You can see in GfxBindings the bindings for methods of RenderCamera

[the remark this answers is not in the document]
Skylion007: Yeah, you can call methods  on a property. A property doesn't have to be a single value.

src/esp/gfx/Renderer.cpp
```diff
@@ -48,11 +49,7 @@ struct Renderer::Impl {
             scene::SceneGraph& sceneGraph,
             RenderCamera::Flags flags) {
     ASSERT(visualSensor.isVisualSensor());
```
bigbike, line 51: In the argument list, `visualSensor` is a reference of `VisualSensor` instance. 
That means `isVisualSensor` always returns true. 
I would remove this ASSERT.

Skylion007:
Looks mostly good, but render_camera really should be a read_only property instead of method.

vauduong:
> Looks mostly good, but render_camera really should be a read_only property instead of method.

Hmm, it seems like that won't work, getting the following error when I make get_render_camera a read_only property
`>           render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera()
E           TypeError: 'habitat_sim._ext.habitat_sim_bindings.Camera' object is not callable`

Skylion007:
> 
> 
> > Looks mostly good, but render_camera really should be a read_only property instead of method.
> 
> Hmm, it seems like that won't work, getting the following error when I make get_render_camera a read_only property
> `> render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera() E TypeError: 'habitat_sim._ext.habitat_sim_bindings.Camera' object is not callable`

It's  a property not a function. calling 
```
render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera
```
but if you make a property, it's the object an no longer the getter for it. Therefore you should rename it should be:

```
render_camera = sim._sensors["color_sensor"]._sensor_object.render_camera
```

vauduong:
Currently, this PR fails test_examples.py in habitat-lab tests due to a habitat-lab test calling the deprecated function SceneGraph::getDefaultRenderCamera: https://app.circleci.com/pipelines/github/facebookresearch/habitat-sim/4737/workflows/ba78662a-b7bf-4a16-9eda-0d9544ff54a5/jobs/20279


This will be fixed with the following PR in habitat-lab: https://github.com/facebookresearch/habitat-lab/pull/575

vauduong:
> > > Looks mostly good, but render_camera really should be a read_only property instead of method.
> > 
> > 
> > Hmm, it seems like that won't work, getting the following error when I make get_render_camera a read_only property
> > `> render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera() E TypeError: 'habitat_sim._ext.habitat_sim_bindings.Camera' object is not callable`
> 
> It's a property not a function. calling the following would work
> 
> ```python
> render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera
> ```
> 
> but if you make a property, it's the object itself and not the getter for it. Therefore you should rename it and the resulting code should be:
> 
> ```python
> render_camera = sim._sensors["color_sensor"]._sensor_object.render_camera
> ```

Got it, thanks Aaron!

Skylion007:
> 
> 
> > > > Looks mostly good, but render_camera really should be a read_only property instead of method.
> > > 
> > > 
> > > Hmm, it seems like that won't work, getting the following error when I make get_render_camera a read_only property
> > > `> render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera() E TypeError: 'habitat_sim._ext.habitat_sim_bindings.Camera' object is not callable`
> > 
> > 
> > It's a property not a function. calling the following would work
> > ```python
> > render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera
> > ```
> > 
> > 
> > but if you make a property, it's the object itself and not the getter for it. Therefore you should rename it and the resulting code should be:
> > ```python
> > render_camera = sim._sensors["color_sensor"]._sensor_object.render_camera
> > ```
> 
> Got it, thanks Aaron!

The examples script that use the old get_render_camera will still fail. You need to update those as well.

vauduong:
> > > > > Looks mostly good, but render_camera really should be a read_only property instead of method.
> > > > 
> > > > 
> > > > Hmm, it seems like that won't work, getting the following error when I make get_render_camera a read_only property
> > > > `> render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera() E TypeError: 'habitat_sim._ext.habitat_sim_bindings.Camera' object is not callable`
> > > 
> > > 
> > > It's a property not a function. calling the following would work
> > > ```python
> > > render_camera = sim._sensors["color_sensor"]._sensor_object.get_render_camera
> > > ```
> > > 
> > > 
> > > but if you make a property, it's the object itself and not the getter for it. Therefore you should rename it and the resulting code should be:
> > > ```python
> > > render_camera = sim._sensors["color_sensor"]._sensor_object.render_camera
> > > ```
> > 
> > 
> > Got it, thanks Aaron!
> 
> The examples script that use the old get_render_camera will still fail. You need to update those as well.

Oops, my bad. Thanks for the catch Aaron, that's fixed now!

Skylion007:
Please remove the .mp4 file that was created.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -105,6 +105,10 @@ void initSensorBindings(py::module& m) {
   py::class_<VisualSensor, Magnum::SceneGraph::PyFeature<VisualSensor>, Sensor,
              Magnum::SceneGraph::PyFeatureHolder<VisualSensor>>(m,
                                                                 "VisualSensor")
+      .def_property_readonly(
+          "get_render_camera", &VisualSensor::getRenderCamera,
```
erikwijmans, line 109: Having a property is much nicer, thank you for doing that.  For properties, no `get_` tho, just `render_camera`

vauduong:
Thank you both! That's fixed!

bigbike:
I still see a failure on Habitat-Lab CI test.
Is that OK? @Skylion007 @mathfac

bigbike:
@mosra : If you have time, would you like to also take a quick pass? :)

erikwijmans:
> I still see a failure on Habitat-Lab CI test.
> Is that OK?

Yeah, @vauduong has a PR to fix the one test that is failing open on hab-lab that we will land after this.

src/utils/viewer/viewer.cpp
```diff
@@ -1011,14 +1009,16 @@ void Viewer::viewportEvent(ViewportEvent& event) {
...
       simulator_->getRenderer()->bindRenderTarget(*visualSensor);
     }
   }
-  Mn::GL::defaultFramebuffer.setViewport({{}, framebufferSize()});
+  Mn::GL::defaultFramebuffer.setViewport({{}, event.framebufferSize()});
+
   if (flyingCameraMode_) {
-    renderCamera_->setViewport(event.windowSize());
+    renderCamera_->setViewport(event.framebufferSize());
   }
```
mosra, line 1022: Nice, even fixing HiDPI issues :)

mosra (commented):
A lot of code deleted, a lot of documentation added. What's not to like :)

src/esp/sensor/CameraSensor.cpp
```diff
@@ -38,6 +44,15 @@ void CameraSensor::setProjectionParameters(const SensorSpec::ptr& spec) {
 
 }  // setProjectionParameters
 
+void CameraSensor::recomputeProjectionMatrix() {
+  projectionMatrix_ = zoomMatrix_ * baseProjMatrix_;
+  // update renderCamera_'s projectionMatrix every time the sensor's
+  // projectionMatrix changes
+  if (renderCamera_) {
```
eundersander, line 51: Is there any scenario where the CameraSensor instance *doesn't* have a renderCamera?

src/tests/GfxReplayTest.cpp
```diff
@@ -28,6 +28,23 @@ using esp::assets::ResourceManager;
 using esp::metadata::MetadataMediator;
 using esp::scene::SceneManager;
 
+// Helper function to get numberOfChildrenOfRoot
+int getNumberOfChildrenOfRoot(esp::scene::SceneNode& rootNode) {
+  int numberOfChildrenOfRoot = 1;
+  const auto* lastRootChild = rootNode.children().first();
+  if (!lastRootChild) {
+    lastRootChild = &rootNode;
```
eundersander, line 36: This assignment seems unnecessary.

src/tests/GfxReplayTest.cpp
```diff
@@ -201,21 +224,35 @@ TEST(GfxReplayTest, player) {
...
+                // was applied to a new child node
+        const auto* rootChild = rootNode.children().first();
+        ASSERT(rootChild);
+        const esp::scene::SceneNode* instanceNode =
+            static_cast<const esp::scene::SceneNode*>(rootChild);
+        ASSERT(instanceNode->translation() == Mn::Vector3(1.f, 2.f, 3.f));
+        ASSERT(instanceNode->getSemanticId() == semanticId);
+      } else {  // if rootNode had children originally, then stateUpdate was
+                // applied to a sibling of lastRootChild
+        ASSERT(lastRootChild->nextSibling());
```
eundersander, line 249: I expected all these tests involving lastRootChild to be replaced with testing on the number of children.

eundersander: It seems like you're only using `lastRootChild` here in this else statement, for the case that `numberOfChildren` > 0. So maybe you can just make `lastRootChild` a local variable inside this else statement, instead of computing it earlier. I'd like us to avoid the `lastRootChild = &rootNode` because that just doesn't make sense to me (the root node is not a child).

eundersander (commented):
GfxReplayTest needs some tweaking to be more readable, otherwise looking good!

src/tests/GfxReplayTest.cpp
vauduong: Thanks for the feedback @eundersander! I modified the comments of the test to be more clear, and use the number of children test wherever possible. I also got rid of `lastRootChild = &rootNode` and use the number of children to get the lastRootChild in the test with the new child. Hope this is more clear and let me know if there are any more changes you'd like to see!

vauduong:
Thanks @eundersander ! I updated the PR with your feedback, could you give it another review when you get the chance?

bigbike:
@eundersander 
Hi, Eric. When you are available, hope you can take a look. See if Vivian has successfully addressed all your concerns. Thanks!

eundersander:
>I updated the PR with your feedback, could you give it another review when you get the chance?
LGTM. I see you have other outstanding requested changes so I suggest resolving them and then finding someone to give a green check approval before merging.

Skylion007 (approved):
Looks good to me. Please update the habitat-lab PR though since get_render_camera() was turned into a property.

bigbike (approved):
Please fix the hablab later once it is merged in. Thanks!

vauduong:
Thanks @bigbike @Skylion007 ! The hab-lab PR is up-to-date with the changes here.

PR #1015, opened 2021-01-11
Integrate per-frame profiler into viewer.cpp

vauduong:
## Motivation and Context

We wish to integrate the FrameProfiler functionality which is present in magnum-player into the habitat simulator viewer.
## How Has This Been Tested

Build and run, compare functionality to magnum-player profiling tool

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1039, opened 2021-01-19
Add comments to clarify per-frame profiler

vauduong:
## Motivation and Context

We added a per-frame profiler in #1015 to display frame duration, cpu duration, and gpu duration at runtime to be aware of bottlenecks in data processing when running our viewer. This PR adds comments to clarify how to interpret values.

## How Has This Been Tested
Build and run

## Types of changes

- [ x] Docs change / refactoring / dependency upgrade

## Checklist

- [ x] I have read the **CONTRIBUTING** document.
- [x ] I have completed my CLA (see **CONTRIBUTING**)

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -551,45 +551,35 @@ Viewer::Viewer(const Arguments& arguments)
    * idling, and CpuDuration and GpuDuration should be roughly equal for faster
    * rendering times
    *
-   * FrameTime: (Units::Nanoseconds) Time to render per frame, 1/FPS, 2 frame
-   * delay
+   * FrameTime: (Units::Nanoseconds) Time to render per frame, 1/FPS
    *
    * CpuDuration: (Units::Nanoseconds) CPU time spent processing events,
    * physics, traversing SceneGraph, and submitting data to GPU/drivers per
    * frame
-   * Measured using std::chrono::high_resolution_clock, 1 frame delay
    *
-   * GpuDuration: (Units::Nanoseconds) GPU time spent rendering data submitted
-   * by CPU per frame
-   * Uses asynchronous querying to measure the amount of time
-   * to fully complete a set of GL commands without stalling rendering, 3 frame
-   * delay
-   * Asynchronous querying extensions: ARB_timer_query (OpenGL 3.3),
-   * EXT_disjoint_timer_query (OpenGL ES, WebGL), EXT_disjoint_timer_query
-   * (WebGL2)
-   * Requires an active OpenGL context
+   * GpuDuration: (Units::Nanoseconds) Measures how much time it takes for the
+   * GPU to process all work submitted by CPU Uses asynchronous querying to
+   * measure the amount of time to fully complete a set of GL commands without
+   * stalling rendering
    */
   Mn::DebugTools::GLFrameProfiler::Values profilerValues =
       Mn::DebugTools::GLFrameProfiler::Value::FrameTime |
       Mn::DebugTools::GLFrameProfiler::Value::CpuDuration |
       Mn::DebugTools::GLFrameProfiler::Value::GpuDuration;
 
-/**
- * VertexFetchRatio and PrimitiveClipRatio only supported for GL 4.6
- *
- * VertexFetchRatio: (Units::RatioThousandths) Ratio of vertex shader
- * invocations to count of vertices submitted
- *
- * PrimitiveClipRatio:  (Units::PercentageThousandths) Ratio of primitives
- * discarded by the clipping stage to count of primitives submitted
- */
+// VertexFetchRatio and PrimitiveClipRatio only supported for GL 4.6
 #ifndef MAGNUM_TARGET_GLES
   if (Mn::GL::Context::current()
           .isExtensionSupported<
               Mn::GL::Extensions::ARB::pipeline_statistics_query>()) {
     profilerValues |=
-        Mn::DebugTools::GLFrameProfiler::Value::VertexFetchRatio |
-        Mn::DebugTools::GLFrameProfiler::Value::PrimitiveClipRatio;
+        Mn::DebugTools::GLFrameProfiler::Value::
+            VertexFetchRatio |  // Ratio of vertex shader invocations to count
+                                // of vertices submitted
+        Mn::DebugTools::GLFrameProfiler::Value::
+            PrimitiveClipRatio;  // Ratio of primitives discarded by the
+                                 // clipping stage to count of primitives
+                                 // submitted
   }
 #endif
 
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -523,11 +523,11 @@ Viewer::Viewer(const Arguments& arguments)
   objectPickingHelper_ = std::make_unique<ObjectPickingHelper>(viewportSize);
   timeline_.start();
 
-  // Set up per frame profiler
+  // Set up per frame profiler to be aware of bottlenecking in processing data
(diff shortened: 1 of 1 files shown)


src/utils/viewer/viewer.cpp
```diff
@@ -523,23 +523,32 @@ Viewer::Viewer(const Arguments& arguments)
...
-  // Set up per frame profiler
+  // Set up per frame profiler to be aware of bottlenecking in processing data
   Mn::DebugTools::GLFrameProfiler::Values profilerValues =
-      Mn::DebugTools::GLFrameProfiler::Value::FrameTime |
-      Mn::DebugTools::GLFrameProfiler::Value::CpuDuration |
-      Mn::DebugTools::GLFrameProfiler::Value::GpuDuration;
+      Mn::DebugTools::GLFrameProfiler::Value::FrameTime |  // Time to render per
+                                                           // frame frame
+      Mn::DebugTools::GLFrameProfiler::Value::
+          CpuDuration |  // Time to process action (eg. physics, key presses)
```
dhruvbatra, line 531: Is physics/simulation and UI processing the only two broad categories? Is there anything else?

src/utils/viewer/viewer.cpp
```diff
@@ -523,23 +523,32 @@ Viewer::Viewer(const Arguments& arguments)
...
+                         // data per frame
+      Mn::DebugTools::GLFrameProfiler::Value::
+          GpuDuration;  // Time to process graphics data per frame
```
dhruvbatra, line 534: "Process graphics data" = rendering? Or do we mean something else?

bigbike: @dhruvbatra : "GpuDruation" is using asynchronous query such as ARB_timer_query, to record the amount of time that GPU takes to fully complete a set of scoped GL commands.

bigbike:
@mosra : Saw you are in the reviewer list.
Any comments would be highly appreciated. Thanks!

vauduong:
@mosra : Yes, would appreciate any feedback on writing more precise comments to help users understand what the GLProfiler is doing!

src/utils/viewer/viewer.cpp
mosra: The profiler *currently* measures just one scope, that's its limitation, except if more than one profiler instance would be used for more scopes, as i suggested on #1015. That's a possible path for extending this functionality. 

(I might be expanding the Magnum functionality eventually as well, depending on what will be interesting to measure for the Vulkan backend.)

src/utils/viewer/viewer.cpp
mosra: I think this could be an easy enough interpretation of the numbers:

- `FrameTime` is an inverse of FPS and should be 16.6 ms or less for smooth interactivity. *In my opinion* it's better than FPS because you can better measure improvements. "FPS increased by 10" means a totally different thing if it was from 10 to 20 or from 900 to 910, while "frame time decreased by 10 ms" always has a clear meaning.
- `CpuDuration` measures how much CPU time was spent doing all work in a frame -- event processing, physics, but also traversing the scenegraph, submitting work for the GPU or the driver overhead. It should be less than GPU duration, if it's not then the GPU is sitting there bored.
- `GpuDuration` measured how much time did it take for the GPU to process all work submitted by the CPU. If it's less than the CPU time then if you reduce the CPU / driver overhead you can do / draw more in a frame (or draw frames faster), if it's significantly more than the CPU time then you're GPU-bound -- rendering too dense meshes, having too high texture resolution, or maybe just rendering a lot of objects that end up occluded or out of the view or having inefficient layout of the mesh data. Generally, since the aim here is to render as fast as possible (and not power efficiency for example), the CPU and GPU time should be roughly equal. 

@vauduong in case you haven't stumbled upon it yet, there's an article about the geometry pipeline in Magnum, with more info about how to interpret the values and what makes meshes slow or fast to render: https://blog.magnum.graphics/announcements/new-geometry-pipeline/ It's *scary long* but written hopefully in a general enough way that might give you useful info even if you don't end up using Magnum further in your career ;)

mosra:
Replied above :) 

I'm realizing some of this info could go straight into Magnum docs as well, because [the explanation is currently a bit underwhelming](https://doc.magnum.graphics/magnum/classMagnum_1_1DebugTools_1_1GLFrameProfiler.html#a49fa9197c6d7993ab67652d89171b9f1).

src/utils/viewer/viewer.cpp
vauduong: Thanks for the helpful comments @mosras! I did stumble upon that article and it was also super informative :)

src/utils/viewer/viewer.cpp
```diff
@@ -545,13 +545,44 @@ Viewer::Viewer(const Arguments& arguments)
...
+   * physics, traversing SceneGraph, and submitting data to GPU/drivers per
+   * frame
+   * Measured using std::chrono::high_resolution_clock, 1 frame delay
+   *
+   * GpuDuration: (Units::Nanoseconds) GPU time spent rendering data submitted
+   * by CPU per frame
+   * Uses asynchronous querying to measure the amount of time
+   * to fully complete a set of GL commands without stalling rendering, 3 frame
+   * delay
+   * Asynchronous querying extensions: ARB_timer_query (OpenGL 3.3),
```
bigbike, line 567: I will recommend removing such details (L567 - L570)

src/utils/viewer/viewer.cpp
```diff
@@ -545,13 +545,44 @@ Viewer::Viewer(const Arguments& arguments)
...
+   * idling, and CpuDuration and GpuDuration should be roughly equal for faster
+   * rendering times
+   *
+   * FrameTime: (Units::Nanoseconds) Time to render per frame, 1/FPS, 2 frame
+   * delay
+   *
+   * CpuDuration: (Units::Nanoseconds) CPU time spent processing events,
+   * physics, traversing SceneGraph, and submitting data to GPU/drivers per
+   * frame
+   * Measured using std::chrono::high_resolution_clock, 1 frame delay
```
bigbike, line 560: remove details such as L560.

src/utils/viewer/viewer.cpp
```diff
@@ -545,13 +545,44 @@ Viewer::Viewer(const Arguments& arguments)
...
+   *
+   * GpuDuration: (Units::Nanoseconds) GPU time spent rendering data submitted
```
bigbike, line 562: measured how much time it takes for the GPU to process all work submitted by the CPU.

src/utils/viewer/viewer.cpp
```diff
@@ -545,13 +545,44 @@ Viewer::Viewer(const Arguments& arguments)
...
+   * by CPU per frame
+   * Uses asynchronous querying to measure the amount of time
+   * to fully complete a set of GL commands without stalling rendering, 3 frame
```
bigbike, line 565: remove 3 frame delay.

src/utils/viewer/viewer.cpp
```diff
@@ -545,13 +545,44 @@ Viewer::Viewer(const Arguments& arguments)
...
+   * (WebGL2)
+   * Requires an active OpenGL context
+   */
   Mn::DebugTools::GLFrameProfiler::Values profilerValues =
       Mn::DebugTools::GLFrameProfiler::Value::FrameTime |
       Mn::DebugTools::GLFrameProfiler::Value::CpuDuration |
       Mn::DebugTools::GLFrameProfiler::Value::GpuDuration;
 
-// VertexFetchRatio and PrimitiveClipRatio only supported for GL 4.6
+/**
```
bigbike, line 577: You do not need this. Undo the change.

PR #1040, opened 2021-01-20
--[BUGFIX] query both x and y offset in mouseScrollEvent for an active scroll.

jturner65:
## Motivation and Context
This PR fixes a bug with how viewer processes mouseScrollEvent offsets on a mac by looking at both x and y offsets if a mouseScrollEvent is processed.  If shift is pressed, the scroll is defaulted to be horizontal on a mac, whereas previously in viewer we only examined the y offset to determine if a valid scroll has occurred, and in what direction.  

## How Has This Been Tested

C++ and python
## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -974,12 +974,15 @@ void Viewer::mouseReleaseEvent(MouseEvent& event) {
 }
 
 void Viewer::mouseScrollEvent(MouseScrollEvent& event) {
-  if (!event.offset().y()) {
+  // shift+scroll is forced into x direction on mac, seemingly at OS level, so
+  // use both x and y offsets.
+  float scrollModVal = event.offset().y() + event.offset().x();
+  if (!(scrollModVal)) {
     return;
   }
   // Use shift for fine-grained zooming
   float modVal = (event.modifiers() & MouseEvent::Modifier::Shift) ? 1.01 : 1.1;
-  float mod = event.offset().y() > 0 ? modVal : 1.0 / modVal;
+  float mod = scrollModVal > 0 ? modVal : 1.0 / modVal;
   auto& cam = getAgentCamera();
   cam.modZoom(mod);
   redraw();
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -976,7 +976,9 @@ void Viewer::mouseReleaseEvent(MouseEvent& event) {
 void Viewer::mouseScrollEvent(MouseScrollEvent& event) {
   // shift+scroll is forced into x direction on mac, seemingly at OS level, so
   // use both x and y offsets.
-  float scrollModVal = event.offset().y() + event.offset().x();
+  float scrollModVal = abs(event.offset().y()) > abs(event.offset().x())
+                           ? event.offset().y()
+                           : event.offset().x();
   if (!(scrollModVal)) {
     return;
   }


Skylion007 (approved):

src/utils/viewer/viewer.cpp
```diff
@@ -974,12 +974,15 @@ void Viewer::mouseReleaseEvent(MouseEvent& event) {
 }
 
 void Viewer::mouseScrollEvent(MouseScrollEvent& event) {
-  if (!event.offset().y()) {
+  // shift+scroll is forced into x direction on mac, seemingly at OS level, so
+  // use both x and y offsets.
+  float scrollModVal = event.offset().y() + event.offset().x();
+  if (!(scrollModVal)) {
```
mosra, line 980: As we discussed on Slack -- I'd rather want to see this fixed / worked around / configured at some upper level than in the application code...

jturner65: The question remains, though, should we intercept it at a lower level? I would imagine folks might want to see/use this behavior in other UI applications (if any existed -shrug-)

src/utils/viewer/viewer.cpp
```diff
@@ -974,12 +974,15 @@ void Viewer::mouseReleaseEvent(MouseEvent& event) {
 }
 
 void Viewer::mouseScrollEvent(MouseScrollEvent& event) {
-  if (!event.offset().y()) {
+  // shift+scroll is forced into x direction on mac, seemingly at OS level, so
+  // use both x and y offsets.
```
mosra, line 978: Too limit this weird workaround a bit, I'd suggest to wrap it in `#ifdef CORRADE_TARGET_APPLE`, at least.

src/utils/viewer/viewer.cpp
```diff
@@ -974,12 +974,15 @@ void Viewer::mouseReleaseEvent(MouseEvent& event) {
...
+  float scrollModVal = event.offset().y() + event.offset().x();
```
mosra, line 979:
```suggestion
  float scrollModVal = event.offset().max();
```

shorter / less error-prone, maybe? On touchpads you can have both directions and so the two could cancel each other out. I think the `max()` could work reasonably well there too.

src/utils/viewer/viewer.cpp
jturner65: I'm not a chronic mac user, and I don't know if we would ever implement something in sim/viewer that would take advantage of horizontal scrolling explicitly, so I figured this was the route of least commitment :)

src/utils/viewer/viewer.cpp
jturner65: is it an abs max?  Can be + or -

jturner65: Also, the x displacement is roughly 10x as big as the y, which was why i added them, but as you say, max would work too, if it is a magnitude max.

src/utils/viewer/viewer.cpp
mosra: > intercept it at a lower level?

Yeah, depends. If I would patch this in Magnum, I would add some window flag to the Application to toggle this behavior (I'm neither a fan of Apple forcing this to everyone nor a library reverting that behavior unconditionally).

I would also need to check how this behaves in SDL, if it's already accounted for in some way or not.

src/utils/viewer/viewer.cpp
mosra: Ah, no it's not, good point. `Mn::Math::abs(event.offset()).max()` then :)

jturner65: How about this : 

```
  float scrollModVal = abs(event.offset().y()) > abs(event.offset().x())
                           ? event.offset().y()
                           : event.offset().x();
```

jturner65: This way then any mouse scrolling on 1d or 2d mice, platform independent, will accomplish something expected, maybe?

mosra: Yeah let's do that :)

PR #1045, opened 2021-01-25
[Validator] Add agent_state validator

Skylion007:
## Motivation and Context

As a sanity check, prevent users from setting an agent_state which has inf or NaN in any of the arrays. It is not sensible behavior and is indicative of an error. 

## How Has This Been Tested
* This has been tested with PyTest

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -252,6 +252,7 @@ class Agent(object):
         if not infer_sensor_states:
             for k, v in state.sensor_states.items():
                 assert k in self._sensors
+                attr.validate(v)
                 if isinstance(v.rotation, list):
                     v.rotation = quat_from_coeffs(v.rotation)
 
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -22,6 +22,7 @@ from habitat_sim.utils.common import (
     quat_to_magnum,
 )
 from habitat_sim.utils.validators import (
+    NoAttrValidationContext,
     all_is_finite,
     is_unit_length,
     value_is_validated,
@@ -210,6 +211,7 @@ class Agent(object):
 
         return did_collide
 
+    @NoAttrValidationContext()
     def get_state(self) -> AgentState:
         habitat_sim.errors.assert_obj_valid(self.body)
         state = AgentState(
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -59,7 +59,7 @@ class SixDOFPose(object):
     :property rotation: unit quaternion rotation
     """
 
-    position: np.ndarray = np.zeros(3)
+    position: np.ndarray = attr.ib(factory=_triple_zero, validator=all_is_finite)
     rotation: Union[np.quaternion, List] = np.quaternion(1, 0, 0, 0)
 
 
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -252,7 +252,6 @@ class Agent(object):
         if not infer_sensor_states:
             for k, v in state.sensor_states.items():
                 assert k in self._sensors
-                attr.validate(v)
                 if isinstance(v.rotation, list):
                     v.rotation = quat_from_coeffs(v.rotation)
 
--- a/habitat_sim/agent/agent.py
+++ b/habitat_sim/agent/agent.py
@@ -20,6 +20,7 @@ from habitat_sim.utils.common import (
     quat_rotate_vector,
     quat_to_magnum,
 )
+from habitat_sim.utils.validators import all_is_finite
 
 from .controls import ActuationSpec, ObjectControls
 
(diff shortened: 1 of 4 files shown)


habitat_sim/agent/agent.py
```diff
@@ -54,18 +59,20 @@ class SixDOFPose(object):
...
-    angular_velocity: np.ndarray = np.zeros(3)
-    force: np.ndarray = np.zeros(3)
-    torque: np.ndarray = np.zeros(3)
+    velocity: np.ndarray = attr.ib(factory=_triple_zero, validator=all_is_finite)
+    angular_velocity: np.ndarray = attr.ib(
+        factory=_triple_zero, validator=all_is_finite
+    )
+    force: np.ndarray = attr.ib(factory=_triple_zero, validator=all_is_finite)
+    torque: np.ndarray = attr.ib(factory=_triple_zero, validator=all_is_finite)
     sensor_states: Dict[str, SixDOFPose] = attr.Factory(dict)
```
erikwijmans, line 76: Can you add similar checks to `SixDOFPose`?

habitat_sim/agent/agent.py
```diff
@@ -54,18 +59,20 @@ class SixDOFPose(object):
...
-    position: np.ndarray = np.zeros(3)
+    position: np.ndarray = attr.ib(factory=_triple_zero, validator=all_is_finite)
     rotation: Union[np.quaternion, List] = np.quaternion(1, 0, 0, 0)
 
 
 @attr.s(auto_attribs=True, slots=True)
 class AgentState(object):
-    position: np.ndarray = np.zeros(3)
+    position: np.ndarray = attr.ib(factory=_triple_zero, validator=all_is_finite)
     rotation: Union[np.quaternion, List, np.ndarray] = np.quaternion(1, 0, 0, 0)
```
erikwijmans, line 69: Let's also check for the rotation being unit length. `.norm()` for `np. np.quaternion` and `np.linalg.norm` for the others.

Skylion007: Are they any np.quaternion functions to do this @erikwijmans ? A lot of numpy functions like np.finite don't work on that type.

habitat_sim/agent/agent.py
Skylion007: @erikwijmans I do in the SixDOFPose struct itself.

habitat_sim/agent/agent.py
erikwijmans: `.norm()` will do it.  That will not be 1 when there is inf or nan

habitat_sim/agent/agent.py
erikwijmans: Yeah, that is just another attrs class defined right above this one

Skylion007: Not sure what you are asking then?

habitat_sim/utils/validators.py
```diff
@@ -0,0 +1,42 @@
...
+
+def all_is_finite(instance, attribute, value) -> None:
+    if not np.all(np.isfinite(value)):
+        raise ValueError(
+            f"{value} contains non-finite values which are forbidden in the {attribute} of {instance}"
+        )
+
+
+def is_unit_length(instance, attribute, value, tol=1e-5) -> None:
+    if isinstance(value, mn.Quaternion):
```
erikwijmans, line 24: mn.Quaternion has https://doc.magnum.graphics/python/magnum/Quaternion/#is_normalized

erikwijmans (changes requested):
Looking at the regression test on the CI, it doesn't seem like this is causing a large enough one to cause that to error, but this does seem to be slowing things down by 10-20%.  We call `Agent.get_state` on every frame currently, however the state constructed in that should never fail the validators (other things would have errored long before in the renderer).

Looks like attrs validators can be globally disabled via `attr.set_run_validators(False)` (and then re-enabled via `attr.set_run_validators(True)`).  Can you add that to get_state to keep things fast when already know things will be valid?

Skylion007:
Validators are only run on state construction (from __init__) and when
attr.validate is called, so I am not sure that would fix the slowdown. Are
you sure get_state is the culprit? Using pyspy?

Skylion007:
@eundersander Could be useful for some of your profiling tools here.

erikwijmans:
Yeah, I looked at current master vs. this PR.  attrs validators are also run on construction for default values.

Skylion007:
Ah, didn't realized they ran on defaults

erikwijmans:
> Ah, didn't realized they ran on defaults

Neither did I, had to double check :-)

Skylion007:
@erikwijmans This PR did catch a potential bug. 

If initial_state of an agent is None, there is a chance that the sample_random_navigable_point will return an invalid state.

Skylion007:
@erikwijmans Still looks like a 10-20% slowdown...

erikwijmans:
> @erikwijmans Still looks like a 10-20% slowdown...

I am not seeing that locally.  If I compare this and master back to back, all seems about equal.  If I comment out the validator disable, then I see the perf drop.

mathfac:
I find random position sampling for the agent useful and was actually using None starting position/rotation as feature.
For example if user would like to have an episode where agent position is random? Sampling position to fill the episode always require loading the scene first.

Possibly easier way to proceed will be to check sampled start position when it wasn't provided?

Skylion007:
@mathfac @erikwijmans bump? Only thing I don't quite like is doing the deep validation of the dict of sensor poses every time.

erikwijmans (approved):
This seems fine with me

PR #1048, opened 2021-01-26
Physics test suite

ScottCSteinhauser:
## Motivation and Context

Part of my intern project is optimizing habitat's collision detection. In order to test the effects of my changes, I've built a modular profiling system that can execute different simulations and collect relevant metrics. The hope is that this can be used by the rest of the team to test changes made to habitat src.  Here is an example of the metrics it prints:

 ========================= Performance ======================== 
| Number of frames: 676
| Average time per physics step: 0.0024307916855670995
| Average time per frame render: 0.002012223887020314
| Maximum Collisions: 246
| Total time for test: 3.101667881011963
|----- END of bowl_drop_test_2 -----
 ==============================================================

https://user-images.githubusercontent.com/56770849/105782120-7f648200-5f39-11eb-9481-39ae0d7ee7d0.mp4
Figure_1

https://user-images.githubusercontent.com/56770849/105782494-44af1980-5f3a-11eb-818b-b5e3b18955b2.mp4

Currently, there is just one test - bowl_drop_test. This drops a specified number of objects into a container and records how long is takes to simulate and render each frame, as well as the number of active collisions each frame. Once I optimize habitat's collision detection, the goal is that this simulation time will be reduced by a noticeable amount.

There is room for other tests - Alex has suggested a scene contact test and a ray-casting test, and I'm open to any suggestions from the team!

That being said, I'd love to hear ideas from you on how this might be used by the team in the future, changes I should make, and features I should add.

## How Has This Been Tested

This is integrated into PyTest in the test_physics_unit_tests file.


tests/test_physics_unit_tests.py
```diff
@@ -0,0 +1,51 @@
+import itertools
+import multiprocessing
+import runpy
+import sys
+from os import path as osp
+
+import pytest
+
+
+def run_main(*args):
```
Skylion007, line 10: We should probably extract this into a test utility instead of duplicating it.

dhruvbatra:
This is awesome! 

In addition to bowl_drop_test, can I also suggest that we use scenes from ReplicaCAD and iGibson? That way, we'll be directly testing and optimizing the scenes and scenarios that @ASzot is using for his experiments.

Skylion007 (changes requested):

tests/test_physics_unit_tests.py
```diff
@@ -0,0 +1,51 @@
...
+    # run_path has one difference with invoking Python from command-line:
+    # if the target is a file (rather than a directory), it does not add its
+    # parent directory to sys.path. Thus, importing other modules from the
+    # same directory is broken unless sys.path is patched here.
+    if osp.isfile(target):
+        sys.path.insert(0, osp.dirname(target))
+    runpy.run_path(target, run_name="__main__")
+
+
+def powerset(iterable):
```
Skylion007, line 23: function unused. Remove

tests/test_physics_unit_tests.py
```diff
@@ -0,0 +1,51 @@
...
+    s = list(iterable)
+    return itertools.chain.from_iterable(
+        itertools.combinations(s, r) for r in range(len(s) + 1)
+    )
+
+
+def run_main_subproc(args):
```
Skylion007, line 30: This should also be factored out into a pytest utility.

examples/tutorials/physics_unit_tests.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+def remove_all_objects(sim):
+    for id_ in sim.get_existing_object_ids():
+        sim.remove_object(id_)
+
+
+def place_agent(sim, position, orientation):
+    # place our agent in the scene
+    agent_state = habitat_sim.AgentState()
```
Skylion007, line 30: position and state, can and should bet set from the constructor if they are known ahead of time.

tests/test_physics_unit_tests.py
```diff
@@ -0,0 +1,51 @@
...
+    if osp.isfile(target):
+        sys.path.insert(0, osp.dirname(target))
+    runpy.run_path(target, run_name="__main__")
+
+
+def powerset(iterable):
+    s = list(iterable)
+    return itertools.chain.from_iterable(
+        itertools.combinations(s, r) for r in range(len(s) + 1)
+    )
```
jturner65, line 27: This is not required here and should be removed, you're not using it as you formulate your test.

examples/tutorials/physics_unit_tests.py
```diff
@@ -0,0 +1,312 @@
...
+data_path = os.path.join(dir_path, "../../data")
+output_path = os.path.join(dir_path, "physics_unit_tests_output/")
+
+save_index = 0
+
+
+# Define Helper Functions (Taken from rigid_object_tutorial.py)
+
+
+def remove_all_objects(sim):
```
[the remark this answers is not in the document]
jturner65: [Once upon a time](https://github.com/facebookresearch/habitat-sim/pull/639/files#) as part of a PR that added a tutorial for adding primitive objects in python, I moved all repeatedly accessed functions to an [external file](https://github.com/jturner65/habitat-sim/blob/43c5046e4848315a00a9ac4b84390324c3afac50/examples/tutorials/tutorial_functions.py).  That PR was obviated due to the focus and development of the colabs, however.  Perhaps we can revisit the idea?

aclegg3:
> This is awesome!
> 
> In addition to bowl_drop_test, can I also suggest that we use scenes from ReplicaCAD and iGibson? That way, we'll be directly testing and optimizing the scenes and scenarios that @ASzot is using for his experiments.

Agreed. I think it makes sense to devise a scene benchmarking procedure which is general enough to be applied trivially to any provided scene and provides some more detailed information that a single scalar (e.g., where in the scene is collision most expensive). @ScottCSteinhauser and I have discussed implementing some possible approaches as new "scenarios" for this benchmarking framework moving forward.

examples/tutorials/physics_unit_tests.py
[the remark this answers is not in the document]
aclegg3: Maybe this should just be utility we provide through Simulator API if it gets implemented everywhere?

examples/tutorials/physics_unit_tests.py
```diff
@@ -0,0 +1,312 @@
...
+    collisions += cur_collisions
+
+    # [/basics]
+    # return total time to run, time to load, time to simulate physics, time for rendering
+    remove_all_objects(sim)
+    sim.close()
+    return physics_step_times, graphics_render_times, collisions, observations
+
+
+unit_tests = {
```
aclegg3, line 218: I wonder if this structure would benefit from an extra entry with a short description of what a scenario is trying to probe? This could also be recorded in the results log mentioned below. Would make the final output more interpretable.

E.g. "Filling a primitive box vessel with primitive shapes" and "Filling a mesh vessel with convex hull shapes", "Benchmarking dense contact between a variety of convex hull shapes", "Benchmarking raycast performance in a scene", "measuring time-to-sleep in a simple drop scenario", etc...

aclegg3 (commented):
Good first pass on this benchmarking utility.

Let's rename this to "physics_benchmarking.py" and similarly switch from "test" to "benchmark", "experiment", or "scenario" wording throughout as applicable.

Skylion007:
@ScottCSteinhauser The Pytest-Benchmarking plugin fixtures may make your life a lot easier on integrating these benchmarks into Habitat_Sim and giving us more accurate readings/CircleCI tests: https://pytest-benchmark.readthedocs.io/en/latest/

examples/tutorials/physics_unit_tests.py
Skylion007: @aclegg3 Any reason why aren't using the pytest-benchmarking plugin for this?

examples/tutorials/physics_unit_tests.py
```diff
@@ -0,0 +1,312 @@
...
+            output_path + "sim_basics",
+            open_vid=show_video,
+        )
+
+    if display_graph:
+        plt.plot(physics_step_times)
+        plt.title("Time to simulate physics for a frame (%s)" % testId)
+        plt.xlabel("Frame #")
+        plt.ylabel("Time Taken (s)")
+        plt.show()
```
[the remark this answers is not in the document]
Skylion007: Or even better, save the plots instead of displaying them!

examples/tutorials/physics_unit_tests.py
```diff
@@ -0,0 +1,312 @@
...
+    """Drops a specified number of objects into a bowl/box and returns metrics including the time to simulate each frame, render each frame, and the number of collisions each frame. """
+    # start with bowl being bounding boxes
+
+    cfg = make_configuration()
+    try:  # Got to make initialization idiot proof
+        sim.close()
+    except NameError:
+        pass
+
+    sim = habitat_sim.Simulator(cfg)
```
Skylion007, line 116: If you use the ```with  habitat_sim.Simulator(cfg) as sim:``` construct for all these tests you won't have to worry about closing the simulator.

examples/tutorials/physics_unit_tests.py
```diff
@@ -0,0 +1,312 @@
...
+        cur_collisions,
+    ) = simulate(sim, dt=post_throw_settling_time, get_frames=make_video)
+    observations += cur_observations
+    physics_step_times += cur_physics_step_times
+    graphics_render_times += cur_graphics_render_times
+    collisions += cur_collisions
+
+    # [/basics]
+    # return total time to run, time to load, time to simulate physics, time for rendering
+    remove_all_objects(sim)
```
Skylion007, line 213: Is removing all the objects necessary?

Skylion007 (changes requested):

examples/tutorials/physics_unit_tests.py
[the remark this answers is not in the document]
Skylion007: Yeah, there should be a simulator.clear_objects().

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,282 @@
...
+    bowlParts[1].scale = np.array([1.0, 0.6, 0.1])
+    bowlParts[2].scale = np.array([1.0, 0.6, 0.1])
+    bowlParts[3].scale = np.array([0.1, 0.6, 1.0])
+    bowlParts[4].scale = np.array([0.1, 0.6, 1.0])
+
+    for i in range(5):
+        part_name = "bowl_part_" + str(i)
+        obj_templates_mgr.register_template(bowlParts[i], part_name)
+        bowlIDs += [sim.add_object_by_handle(part_name)]
+        sim.set_object_motion_type(habitat_sim.physics.MotionType.KINEMATIC, bowlIDs[i])
```
aclegg3, line 125: This setting may not let your other objects de-activate. Perhaps set the state first and then set these objects as `MotionType::STATIC` instead.

ScottCSteinhauser:
> @ScottCSteinhauser The Pytest-Benchmarking plugin fixtures may make your life a lot easier on integrating these benchmarks into Habitat_Sim and giving us more accurate readings/CircleCI tests: https://pytest-benchmark.readthedocs.io/en/latest/

Thank you for pointing this out - I think I'll find a way to include this in the next PR. It could be valuable to see the test results straight from pytest.

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+
+        box_parts = []
+        boxIDs = []
+        for _i in range(5):
+            box_parts += [obj_templates_mgr.get_template_by_handle(cube_handle)]
+        box_parts[0].scale = np.array([1.0, 0.1, 1.0]) * box_size
+        box_parts[1].scale = np.array([1.0, 0.6, 0.1]) * box_size
+        box_parts[2].scale = np.array([1.0, 0.6, 0.1]) * box_size
+        box_parts[3].scale = np.array([0.1, 0.6, 1.0]) * box_size
+        box_parts[4].scale = np.array([0.1, 0.6, 1.0]) * box_size
```
aclegg3, line 125: You don't need duplicates here.

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+
+        # throw objects into box
+        for i in range(num_objects):
+            cur_id = sim.add_object(object_ids[i % len(object_ids)])
+
+            obj_node = sim.get_object_scene_node(cur_id)
+            obj_bb = obj_node.cumulative_bb
+            diaganal_length = math.sqrt(
+                obj_bb.size_x() ** 2 + obj_bb.size_y() ** 2 + obj_bb.size_z() ** 2
+            )
```
aclegg3, line 169: Typo spelling.
Also, Magnum utils for this:
```suggestion
            diagonal_length = obj_bb.size().length()
```

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+            time_til_next_obj = diaganal_length / (object_speed * box_size) / 2
+            # obj_node.scale(mn.Vector3(3,3,3))
```
aclegg3, line 171: remove comment

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+
+            # set object position and velocity
+            sim.set_translation(np.multiply(np.array([1.50, 2, 1.2]), box_size), cur_id)
+            initial_linear_velocity = mn.Vector3(1, 0, -1)
+            initial_linear_velocity = (
+                initial_linear_velocity.normalized() * object_speed * box_size
+            )
+            sim.set_linear_velocity(initial_linear_velocity, cur_id)
```
aclegg3, line 179: Could be simplified to:
```suggestion
            sim.set_linear_velocity(mn.Vector3(1, 0, -1).normalized()*object_speed*box_size, cur_id)
```

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+            sim.set_object_motion_type(habitat_sim.physics.MotionType.STATIC, boxIDs[i])
+
+        # load some object templates from configuration files
+        object_ids = []
+        scales = {
+            "test_assets/objects/sphere": 1,
+            "test_assets/objects/chair": 0.9,
+            "test_assets/objects/donut": 2,
+            "test_assets/objects/nested_box": 0.5,
+        }
```
aclegg3, line 149: Maybe want to embed this metadata in the object parameter so it can be configured externally? Could be tuples or dictionaries for each entry.

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+    else:
+        # only runs the benchmark/benchmark_sets specified in commandline arguments
+        for test in tests_to_run:
+            if test in benchmark_sets:
+                for sub_test in benchmark_sets[test]:
+                    log_contents += run_test(sub_test)
+            elif test in benchmarks:
+                log_contents += run_test(test)
+            else:
+                print(test, "done not exist.")
```
aclegg3, line 304:
```suggestion
                print(test, " does not exist.")
```

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--no-show-video", dest="show_video", action="store_false")
+    parser.add_argument("--no-make-video", dest="make_video", action="store_false")
+    parser.add_argument("--save-to-log", dest="save_log", action="store_true")
+    parser.add_argument("--graph", default=[], nargs="*", type=str, metavar="N")
+    parser.add_argument("--run", default=[], nargs="*", type=str, metavar="N")
+
+    parser.set_defaults(show_video=True, make_video=True, save_log=False)
+    args, _ = parser.parse_known_args()
```
aclegg3, line 323: Suggestion: error on unknown args so user gets a clear signal if they messed up? 🤷 
```suggestion
    args = parser.parse_args()
```

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+
+    show_video = args.show_video
+    make_video = args.make_video
+    save_log = args.save_log
+    graphs_to_produce = args.graph
+    tests_to_run = args.run
+
+    if make_video and not os.path.exists(output_path):
```
aclegg3, line 331: Will also want to produce the output path for logging and graph saving. Maybe just always create it?
```suggestion
    if not os.path.exists(output_path):
```

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--no-show-video", dest="show_video", action="store_false")
+    parser.add_argument("--no-make-video", dest="make_video", action="store_false")
+    parser.add_argument("--save-to-log", dest="save_log", action="store_true")
+    parser.add_argument("--graph", default=[], nargs="*", type=str, metavar="N")
+    parser.add_argument("--run", default=[], nargs="*", type=str, metavar="N")
```
aclegg3, line 320: Could add help text to less obvious arguments like "graph" and "run". 
E.g.
```
parser.add_argument("--save-to-log", dest="save_log", action="store_true", help="Save benchmark results to a log file in the script output directory.")
```

examples/tutorials/physics_benchmarking.py
```diff
@@ -0,0 +1,334 @@
...
+import habitat_sim
+from habitat_sim.utils import viz_utils as vut
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+data_path = os.path.join(dir_path, "../../data")
+output_path = os.path.join(dir_path, "physics_benchmarking_output/")
+
+save_index = 0
+
+
```
aclegg3, line 19: unused

aclegg3 (commented):
Almost there. Most of these are minor comments or suggestions.

examples/tutorials/physics_benchmarking.py
Skylion007: ^ parse_known_args is a required workaround to get it to work in Python notebooks. This isn't compiled as a colab, but maybe it should be?

examples/tutorials/physics_benchmarking.py
ScottCSteinhauser: If I don't use duplicates, then I can't use the loop below this and would have to manually register each box part. Is it worth it to remove duplicates and change the loop?

tests/helpers/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    # run_path has one difference with invoking Python from command-line:
+    # if the target is a file (rather than a directory), it does not add its
+    # parent directory to sys.path. Thus, importing other modules from the
+    # same directory is broken unless sys.path is patched here.
+    if osp.isfile(target):
+        sys.path.insert(0, osp.dirname(target))
+    runpy.run_path(target, run_name="__main__")
+
+
+def run_main_subproc(args):
```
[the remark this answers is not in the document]
ScottCSteinhauser: Good idea - just removed a re-definition in test_examples.py

examples/tutorials/physics_benchmarking.py
ScottCSteinhauser: > ^ parse_known_args is a required workaround to get it to work in Python notebooks. This isn't compiled as a colab, but maybe it should be?

We could. Just out of curiosity - are colabs used for anything other than tutorials?

aclegg3: @ScottCSteinhauser The real purpose of the Colab is to provide a no-install option to use Habitat-sim in the browser with minimal setup. Tutorials are a good example of why we'd want to do this. 🙂 

The purpose of this script is to benchmark systems, I don't know that we are targeting Colab. I guess an argument could be made that you may want to know physics perf on the Colab node also, but I'd rather keep it clean and simple.

aclegg3 (approved):
Looks like this is modifying the pybind submodule version now after the rebase. Let's fix that.

Otherwise I think we are ready to merge this. 🎉

tests/helpers/utils.py
```diff
@@ -0,0 +1,28 @@
+import multiprocessing
+import runpy
+import sys
+from os import path as osp
+
+
+def run_main(*args):
+    # patch sys.args
+    sys.argv = list(args)
+    print("Args", args)
```
Skylion007, line 10: This should be removed. It could interfere with tests

Skylion007 (changes requested):

tests/helpers/utils.py
ScottCSteinhauser: Oh yeah, thank you for pointing that out.

Skylion007:
I don't think the pybind commit is correct, it should have been updated the v2.6.2, it looks like it's one 2.6.1

PR #1049, opened 2021-01-26
Gfx replay tutorial

eundersander:
## Motivation and Context

Tutorial for Gfx Replay (aka graphics replay, aka render replay).

Gfx Replay is a feature that lets you save the visual state of the sim, restore the visual state later, and either reproduce earlier observations (from the same camera position) or produce new observations (from different camera positions).

Note that restoring the visual state is not the same as restoring the full simulation state. When playing a replay, no physics or other simulation is running. Object ids from the earlier recorded simulation aren't valid and you can't interact with objects in the scene. However, you can move your agent, sensor, or camera to produce new observations from different camera positions.

The recording API (see replay_tutorial.py):
- cfg.enable_gfx_replay_save
- sim.gfx_replay_manager.save_keyframe
- gfx_replay_utils.add_node_user_transform (wraps sim.gfx_replay_manager.add_user_transform_to_keyframe)
- sim.gfx_replay_manager.write_saved_keyframes_to_file

The playback API:
- gfx_replay_utils.make_backend_configuration_for_playback
- sim.gfx_replay_manager.read_keyframes_from_file
- player.set_keyframe_index
- player.get_user_transform

https://user-images.githubusercontent.com/6557808/105935287-902ef980-6006-11eb-9cee-ceb9afdfdded.mp4
sequence

## How Has This Been Tested

Running locally and as Colab.

I was unable to build/test the doc change to index.rst on either Mac or Ubuntu. The change is small so maybe I'll get lucky and it'll be fine.

## Types of changes

## Checklist

### DIFF
--- a/examples/tutorials/colabs/replay_tutorial.ipynb
+++ b/examples/tutorials/colabs/replay_tutorial.ipynb
@@ -8,7 +8,7 @@
     "\n",
     "gfx replay is a feature that lets you save the visual state of the sim, restore the visual state later, and either reproduce earlier observations (from the same camera position) or produce new observations (from different camera positions).\n",
     "\n",
-    "Note that restoring the visual state is not the same as restoring the full simulation state; you can't resume simulation from a replay.\n",
+    "Note that restoring the visual state is not the same as restoring the full simulation state. When playing a replay, no physics or other simulation is running. Object ids from the earlier recorded simulation aren't valid and you can't interact with objects in the scene. However, you can move your agent, sensor, or camera to produce new observations from different camera positions.\n",
     "\n",
     "The recording API:\n",
     "- cfg.enable_gfx_replay_save\n",
@@ -397,7 +397,7 @@
    },
    "source": [
     "## Reconfigure simulator for replay playback.\n",
-    "Note call to gfx_replay_utils.make_backend_configuration_for_playback. Note that we don't specify a scene or stage when reconfiguring for replay playback. need_separate_semantic_scene_graph is generally set to False, but must be set to True if you're replaying a scene that uses a separate semantic mesh."
+    "Note call to gfx_replay_utils.make_backend_configuration_for_playback. Note that we don't specify a scene or stage when reconfiguring for replay playback. need_separate_semantic_scene_graph is generally set to False. If you're using a semantic sensor and replaying a scene that uses a separate semantic mesh (like an MP3D scene), set this to True. If in doubt, be aware there's a Habitat runtime warning that will always catch incorrect usage of this flag."
    ]
   },
   {
@@ -433,7 +433,7 @@
     "lines_to_next_cell": 0
    },
    "source": [
-    "## Place dummy agent with identify transform.\n",
+    "## Place dummy agent with identity transform.\n",
     "For replay playback, we place a dummy agent at the origin and then transform the sensor using the \"sensor\" user transform stored in the replay. In the future, Habitat will offer a cleaner way to play replays without an agent."
    ]
   },
--- a/examples/tutorials/colabs/replay_tutorial.ipynb
+++ b/examples/tutorials/colabs/replay_tutorial.ipynb
@@ -1,0 +1,672 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#Gfx Replay Tutorial\n",
+    "\n",
+    "gfx replay is a feature that lets you save the visual state of the sim, restore the visual state later, and either reproduce earlier observations (from the same camera position) or produce new observations (from different camera positions).\n",
+    "\n",
+    "Note that restoring the visual state is not the same as restoring the full simulation state; you can't resume simulation from a replay.\n",
+    "\n",
+    "The recording API:\n",
+    "- cfg.enable_gfx_replay_save\n",
+    "- sim.gfx_replay_manager.save_keyframe\n",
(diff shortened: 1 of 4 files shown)


examples/tutorials/colabs/replay_tutorial.ipynb
```diff
@@ -0,0 +1,521 @@
+{
```
eundersander, line 1: This is an auto-generated file. Don't review it. Review the py file below instead.

habitat_sim/gfx_replay_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def make_backend_configuration_for_playback(
+    need_separate_semantic_scene_graph=False,
+):
+    backend_cfg = habitat_sim.SimulatorConfiguration()
+    backend_cfg.scene_id = "NONE"  # see Asset.h EMPTY_SCENE
+    backend_cfg.force_separate_semantic_scene_graph = need_separate_semantic_scene_graph
+
+    return backend_cfg
```
jturner65, line 32: Should be in habitat_sim/utils directory, no?

jturner65 (approved):
Comment about file placement (utils files should go in utils dir :) ).  Otherwise, LGTM (once everything passes of course)

habitat_sim/gfx_replay_utils.py
eundersander: Previously, we've discussed that this utils directory should be where we start putting "pure" utilities, I mean, utilities that don't rely on the rest of habitat_sim. We can eventually even modify our CI to enforce this. @aclegg3 is this your understanding of this directory? Perhaps we should rename to "core_utils" ("core" implying few other dependencies)?

In the case if gfx_replay_utils, this is a utility that very much relies on habitat_sim (namely, gfx_replay).

eundersander:
Comment on one of my CI failures:
```
-    "    for i in range(10):\n",
+    "    for _ in range(10):\n",
```
Black and flake8 are literally fighting over this. One changes it from `_` to `i` and the other demands I change it back. I will investigate.

erikwijmans:
I've seen `_i` a couples places in our code, maybe that makes them happy?

habitat_sim/gfx_replay_utils.py
jturner65: Fair enough. Would it be reasonable to find somewhere to put it under habitat-sim/habitat_sim that would help keep things organized within that directory? I'm concerned that habitat-sim/habitat_sim run the risk of becoming a disorganized catch-all.

aclegg3: Currently, the role of the utils directory is muddy in this regard. There are modules which require habitat (e.g. profiling_utils.py, viz_utils.py) and those which don't (e.g. common.py, collect_env.py, and compare_profiles.py). Some of these still require Magnum or other libraries. 

I think we agreed that there should be a separation between utilities that require a Habitat-sim install and those which do not. Perhaps this should manifest as separate "habitat-utils" and "utils" directories, or a single directory with per-file separation as exists now. 

Either way, I support John's suggestion that this should be moved somewhere and it feels like "utils" is the right place at this point (it seems like this should maybe live next to profiling and visualization utilities). Perhaps we can put it there and then follow-up with a refactor for a better division later.

aclegg3 (approved):
This looks good to me. Cool feature. 👍 

Maybe you can add some output videos to the PR description and link this on the docs homepage with the other resources.

examples/tutorials/nb_python/replay_tutorial.py
```diff
@@ -0,0 +1,440 @@
...
+#     display_name: Python 3
+#     name: python3
+# ---
+
+# %% [markdown]
+# #Gfx Replay Tutorial
+#
+# gfx replay is a feature that lets you save the visual state of the sim, restore the visual state later, and either reproduce earlier observations (from the same camera position) or produce new observations (from different camera positions).
+#
+# Note that restoring the visual state is not the same as restoring the full simulation state; you can't resume simulation from a replay.
```
dhruvbatra, line 26: An uninformed user may not understand what all is entailed in "simulation". Can we take the description from the internal post (e.g. "The user can move around in the space (using keyboard control) to view this "object waterfall" from any viewpoint, but she may not interact with these objects.) and add it here as appropriate.

examples/tutorials/nb_python/replay_tutorial.py
```diff
@@ -0,0 +1,440 @@
...
+# gfx replay files are written as JSON.
+# %%
+
+sim.gfx_replay_manager.write_saved_keyframes_to_file(replay_filepath)
+
+remove_all_objects(sim)
+
+# %% [markdown]
+# ## Reconfigure simulator for replay playback.
+# Note call to gfx_replay_utils.make_backend_configuration_for_playback. Note that we don't specify a scene or stage when reconfiguring for replay playback. need_separate_semantic_scene_graph is generally set to False, but must be set to True if you're replaying a scene that uses a separate semantic mesh.
```
dhruvbatra, line 276:
```suggestion
# Note call to gfx_replay_utils.make_backend_configuration_for_playback. Note that we don't specify a scene or stage when reconfiguring for replay playback. need_separate_semantic_scene_graph is generally set to False, but must be set to True if you're replaying a scene that uses a separate semantic mesh (like MP3D dataset).
```

dhruvbatra (approved):
Minor edits, but overall LGTM.

eundersander:
The CI error on build_and_test_habitat_lab_conda is a known, unrelated issue. In the interest of time, I'm going to merge this now.

PR #1052, opened 2021-01-27
Fix the conda binary build

erikwijmans:
## Motivation and Context

`gitpython`'s pip package requires `smmap<4.0,>=3.0.1`, however, it's conda package does not list this requirement properly.  That mismatch means conda-build pulls a version of smmap that will later break the build due to dependency checking for gitpython failing.

## How Has This Been Tested

The CI


jturner65 (approved):
LGTM!

PR #1054, opened 2021-01-27
Make nav point sampling more robust

erikwijmans:
## Motivation and Context

Looking at the recast code, there are two reasons why `navQuery_->findRandomPoint` would fail on a well formed navmesh. 

1. It fails to sample a polygon.  This can happen because the tile is sampled without checking to see if it has any polygons to select
2. There is a machine precision error that causes the sampled point to fail a point-in-poly test.

In both cases, trying again makes sense.  I added a runtime error to throw when the navmesh is definitely bad, no navigable area.

## How Has This Been Tested

Made a little script that samples a couple hundred million points.  That use to error once or twice even tho the navmesh is fine.  It no longer errors.

## Types of changes

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -1344,7 +1344,7 @@ bool PathFinder::build(const NavMeshSettings& bs,
 }
 
 vec3f PathFinder::getRandomNavigablePoint(const int maxTries /*= 10*/) {
-  return pimpl_->getRandomNavigablePoint();
+  return pimpl_->getRandomNavigablePoint(maxTries);
 }
 
 bool PathFinder::findPath(ShortestPath& path) {
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -924,15 +924,29 @@ static float frand() {
 }
 
 vec3f PathFinder::Impl::getRandomNavigablePoint() {
-  dtPolyRef ref;
+  constexpr int MAX_TRIES = 1e3;
   constexpr float inf = std::numeric_limits<float>::infinity();
-  vec3f pt(inf, inf, inf);
-  dtStatus status =
-      navQuery_->findRandomPoint(filter_.get(), frand, &ref, pt.data());
-  if (!dtStatusSucceed(status)) {
-    LOG(ERROR) << "Failed to getRandomNavigablePoint";
+  if (getNavigableArea() <= 0.0)
+    throw std::runtime_error(
+        "NavMesh has no navigable area, this indicates an issue with the "
+        "NavMesh");
+
+  int i = 0;
+  vec3f pt;
+  for (int i = 0; i < MAX_TRIES; ++i) {
+    dtPolyRef ref;
+    dtStatus status =
+        navQuery_->findRandomPoint(filter_.get(), frand, &ref, pt.data());
+    if (dtStatusSucceed(status))
+      break;
   }
-  return pt;
+
+  if (i == MAX_TRIES) {
+    LOG(ERROR) << "Failed to getRandomNavigablePoint";
+    return {inf, inf, inf};
+  } else {
+    return pt;
+  }
 }
 
 namespace {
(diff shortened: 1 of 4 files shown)


src/esp/nav/PathFinder.cpp
```diff
@@ -924,15 +924,30 @@ static float frand() {
...
+    dtPolyRef ref;
+    dtStatus status =
+        navQuery_->findRandomPoint(filter_.get(), frand, &ref, pt.data());
+    if (dtStatusSucceed(status))
+      break;
+  }
+
+  if (i == MAX_TRIES) {
     LOG(ERROR) << "Failed to getRandomNavigablePoint";
+    return {inf, inf, inf};
```
Skylion007, line 947: Maybe we should make this NaN for consistency. I know we had it as inf before though.

src/esp/nav/PathFinder.cpp
```diff
@@ -924,15 +924,30 @@ static float frand() {
 }
 
 vec3f PathFinder::Impl::getRandomNavigablePoint() {
-  dtPolyRef ref;
+  constexpr int MAX_TRIES = 1e3;
```
Skylion007, line 927: Shouldn't this be configurable? as part of the function arg

Skylion007 (changes requested):
I agree on this. Maybe we should return a value of NaN instead of inf for this. I would also make the max_tries configurable as we already use this trick when generating the datasets fo Habitat-Lab so might as well just make the max_tries a function arg here.

aclegg3 (approved):
Looks good, thanks for the investigation.

src/esp/nav/PathFinder.cpp
```diff
@@ -1329,7 +1343,7 @@ bool PathFinder::build(const NavMeshSettings& bs,
   return pimpl_->build(bs, mesh);
 }
 
-vec3f PathFinder::getRandomNavigablePoint() {
+vec3f PathFinder::getRandomNavigablePoint(const int maxTries /*= 10*/) {
   return pimpl_->getRandomNavigablePoint();
```
aclegg3, line 1347: Need to pass the variable 🙂

Skylion007: I'm glad my clang-tidy checks caught this bug appearing on the exact same class. 👍

erikwijmans: Looks like I forgot to check if this would compile locally... this is exactly why I didn't put a default value in the pimpl header...

Skylion007: @erikwijmans Even if you did, we now have clang-tidy checks for unused values. 👍

Skylion007 (approved):

src/esp/nav/PathFinder.cpp
```diff
@@ -225,7 +225,7 @@ struct PathFinder::Impl {
              const float* bmax);
   bool build(const NavMeshSettings& bs, const esp::assets::MeshData& mesh);
 
-  vec3f getRandomNavigablePoint();
+  vec3f getRandomNavigablePoint(const int maxTries);
```
mosra, line 228:
```suggestion
  vec3f getRandomNavigablePoint(int maxTries);
```

it's just noise in declarations, makes sense only if the variable is actually used :)

src/esp/nav/PathFinder.cpp
```diff
@@ -923,16 +923,30 @@ static float frand() {
...
+    dtStatus status =
+        navQuery_->findRandomPoint(filter_.get(), frand, &ref, pt.data());
+    if (dtStatusSucceed(status))
+      break;
+  }
+
+  if (i == maxTries) {
+    LOG(ERROR) << "Failed to getRandomNavigablePoint.  Try increasing max "
+                  "tries if the navmesh is fine but just hard to sample from";
+    return {NAN, NAN, NAN};
```
mosra, line 946: FYI there's [Mn::Constants::nan()](https://doc.magnum.graphics/magnum/structMagnum_1_1Math_1_1Constants.html#a0cc96df25e99d8e386551475d6d171fc) in case you wouldn't want to use the macro (mainly suggesting it because it has other common constants as well, all in one place).

Skylion007:
LGTM, let's merge.

PR #1055, opened 2021-01-27
VHACD (Convex Hull Decomposition) Integration

ScottCSteinhauser:
## Motivation and Context

Integrating VHACD into Habitat-sim will allow us to simplify collision meshes into their approximate convex hulls, speeding up collision detection. 

image

## Types of changes

With this PR, VHACD has been enabled in the project. Users can run convex hull decomposition on an object template of their choice and tune the VHACD parameters in the python code. An example of this can be seen in VHACD_tutorial.py. 

Here's an example of different resolution decompositions on various Replica_CAD objects:

https://user-images.githubusercontent.com/56770849/107542306-ab187680-6b8d-11eb-9ed4-ce3560d02abe.mp4

And here is an example of a performance boost from using a medium resolution CHD as opposed to the original mesh (In this case, the objects/chair.glb was decomposed into 4 convex hulls for the CHD. 100 of these chairs were dropped into a box (Using physics_benchmarking.py's box_drop_test)). 

Screen Shot 2021-02-10 at 10 49 51 AM


Skylion007 (changes requested):
1st: you should run `pre-commit -a ` to run all the necessary linters on these files and make you have pre-commit installed so it warns you in the future. Run pre-commit install after installing it with pip to have these hooks run automatically.
2nd: Is there any reason that VHACD can't be a git submodule? It looks like the git repo was just pasted in there? If it requires some changes, we can fork it and point to that git repo as the submodule.

aclegg3:
> Is there any reason that VHACD can't be a git submodule? It looks like the git repo was just pasted in there? If it requires some changes, we can fork it and point to that git repo as the submodule.

Thanks for bringing this up. This is one of the points @ScottCSteinhauser and I were hoping to get some feedback on. In the meantime, this initial build unblocks further work on the integration while we figure out the details. Yes, we plan to modify the source to extend the interface, so perhaps a forked submodule is the right choice. @mosra Would appreciate your thoughts on this also.

Skylion007:
> 
> 
> > Is there any reason that VHACD can't be a git submodule? It looks like the git repo was just pasted in there? If it requires some changes, we can fork it and point to that git repo as the submodule.
> 
> Thanks for bringing this up. This is one of the points @ScottCSteinhauser and I were hoping to get some feedback on. In the meantime, this initial build unblocks further work on the integration while we figure out the details. Yes, we plan to modify the source to extend the interface, so perhaps a forked submodule is the right choice. @mosra Would appreciate your thoughts on this also.

I would just like to avoid polluting master with various assets that are going to be immediately moved to a submodule. There are lot of binary images in this PR that makes it quite large for instance.

Also if VHACD ever does get bugfixes, this will be trivial to merge, like we recently did on Recast.

src/esp/assets/CMakeLists.txt
```diff
@@ -79,8 +79,11 @@ target_link_libraries(
          MagnumPlugins::StbImageConverter
          MagnumPlugins::TinyGltfImporter
   PRIVATE geo io
+         vhacd
 )
 
+target_include_directories(vhacd PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../deps/v-hacd/src/VHACD_Lib/public)
```
mosra, line 85: Modifying targets from outside is allowed only in CMake 3.13 and newer. I don't think Habitat requires that version yet, and I would suggest not arbitrarily increasing the required CMake version unless absolutely needed -- it adds unnecessary friction for users on LTS distros :) Ubuntu 18.04 LTS has 3.10, so I'd say staying at that version is reasonable. (Just FYI Magnum has stricter requirements, so there the minimal version is still 3.4.)

To make this work on older CMakes, I'd instead do

```suggestion
target_include_directories(assets PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../deps/v-hacd/src/VHACD_Lib/public)
```

... and ideally fork / submit a patch to vhacd so it does a `target_include_directories()` on its own :)

src/esp/assets/ResourceManager.cpp
```diff
@@ -47,6 +47,8 @@
 
 #include "esp/nav/PathFinder.h"
 
+#include "deps/v-hacd/src/VHACD_Lib/public/VHACD.h"
```
mosra, line 50: Hm, since the `public/` include directory is now in the include path, shouldn't this be 

```suggestion
#include "VHACD.h"
```

instead? I'm not a fan of hardcoding the full paths this way :)

src/cmake/dependencies.cmake
```diff
@@ -90,6 +90,11 @@ if(BUILD_ASSIMP_SUPPORT)
   find_package(Assimp REQUIRED)
 endif()
 
+# v-hacd
+set(NO_OPENCL ON CACHE BOOL "NO_OPENCL" FORCE)
+set(NO_OPENMP ON CACHE BOOL "NO_OPENMP" FORCE)
+add_subdirectory("${DEPS_DIR}/v-hacd/src/VHACD_Lib")
```
mosra, line 96: It might be useful to add a comment explaining why `src/VHACD_Lib`  was added as a subdirectory and not `src/`.

(I checked the sources and it makes sense to do so, in order to skip the tests, but explicitly pointing this out could help debugging in case V-HACD's top-level CMakeLists becomes important at some point in the future.)

mosra (commented):
As others pointed out already, I wouldn't embed the whole codebase in here, especially not with its whole history squashed into a single commit. Let's not repeat the same mistake Bullet did, this would make upgrades *really* painful ;)

I'd say an optimistic way to go about this is to fork V-HACD to your personal account, do the required modifications/additions there and submit a PR upstream. Habitat would then temporarily use your fork as a Git submodule and when the PR is merged upstream, we'd switch the submodule to upstream.

Skylion007:
Clang-Tidy is complaining:
```/home/circleci/project/habitat-sim/src/esp/assets/ResourceManager.cpp:97:18: error: Value stored to 'interfaceVHACD' during its initialization is never read [clang-analyzer-deadcode.DeadStores,-warnings-as-errors]
  VHACD::IVHACD* interfaceVHACD = VHACD::CreateVHACD();
                 ^
/home/circleci/project/habitat-sim/src/esp/assets/ResourceManager.cpp:97:18: note: Value stored to 'interfaceVHACD' during its initialization is never read
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2184,147 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+
+  // make LoadedAssetData corresponding to this asset
+  LoadedAssetData loadedAssetData{info, meshMetaData};
+  // TODO : need to free render assets associated with this object if collision
+  // occurs, otherwise leak! (Currently unsupported).
+  // if (resourceDict_.count(trajVisName) != 0) {
+  //   resourceDict_.erase(trajVisName);
+  // }
+
+  Cr::Utility::Debug() << "END";
```
mosra, line 2320: @ScottCSteinhauser printf debugging hint -- you [can use "exclamated" `!Debug{};` to add markers containing the file and line number](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html#Utility-Debug-source-location), without having to type arbitrary text ;)

Also, `Cr::Utility::Debug{}` is aliased to `Mn::Debug{}` for less typing.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2185,209 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+
+    // Create CollisionMeshData and add to collisionMeshGroup vector
+    CollisionMeshData CHCollisionMesh = genCHMeshData->getCollisionMeshData();
+    collisionMeshGroup.push_back(CHCollisionMesh);
+
+    // register GenericMeshData in meshes_ dict
+    meshes_.emplace(meshes_.size(), std::move(genCHMeshData));
+  }
+  // make MeshMetaData
+  int meshEnd = meshes_.size() - 1;
```
Skylion007, line 2334: Doesn't this introduce narrowing conversion errors? Should be unsigned?

src/esp/assets/ResourceManager.h
```diff
@@ -42,6 +42,9 @@
 #include "esp/metadata/MetadataMediator.h"
 #include "esp/metadata/attributes/AttributesBase.h"
 
+//#include <VHACD.h>
```
Skylion007, line 45:
```suggestion
```

src/esp/bindings/SimBindings.cpp
```diff
@@ -285,6 +285,11 @@ void initSimBindings(py::module& m) {
           "recompute_navmesh", &Simulator::recomputeNavMesh, "pathfinder"_a,
           "navmesh_settings"_a, "include_static_objects"_a = false,
           R"(Recompute the NavMesh for a given PathFinder instance using configured NavMeshSettings. Optionally include all MotionType::STATIC objects in the navigability constraints.)")
+      .def(
+          "convex_hull_decomposition", &Simulator::convexHullDecomposition,
```
Skylion007, line 289: Nit: On the Simulator class bindings too, methods should be verbs,  so it should be apply_convex_hull_decomposition or convex_hell_composite? (Should definitely have convex hull decomposition somewhere though for the search term.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2185,209 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+  }  // if no render asset exists
+
+  // fill points and triangles with mesh data
+  std::vector<float> points = std::vector<float>();
+  std::vector<uint32_t> triangles = std::vector<uint32_t>();
+  getPrimitiveMeshData(filename, points, triangles);
+
+  // use VHACD
+  bool res =
+      interfaceVHACD->Compute(&points[0], (unsigned int)(points.size() / 3),
```
Skylion007, line 2276: Instead of casting back and for shouldn't the divison just been done between two unsigned ints instead?

Skylion007 (changes requested):
Just some minor nits. Great job on the bindings

src/esp/bindings/SimBindings.cpp
Skylion007: Actually, Trimesh does this too, which is disappointing so I suppose it's fine.

src/esp/assets/ResourceManager.cpp
Skylion007: Also, we should be trying to avoid C-Style cast's when possible right @aclegg3 ? Also there is a lot of unecessarily throwing away the signedness of ints in this PR. No reason to do narrowing conversions particularly here where you widen the conversion immediately after.

Skylion007:
@ScottCSteinhauser Looks like clang-tidy caught most of my complaints. Just fix them or have clang-tidy apply the fixes for you locally.

src/esp/assets/ResourceManager.cpp
ScottCSteinhauser: I pass meshStart and meshEnd into the MeshMetaData which only accepts ints.

src/esp/bindings/SimBindings.cpp
ScottCSteinhauser: Still a good thought, I went ahead and changed it :)

src/esp/sim/Simulator.cpp
```diff
@@ -1076,6 +1076,27 @@ scene::SceneNode* Simulator::loadAndCreateRenderAssetInstance(
       assetInfo, creation, sceneManager_.get(), tempIDs);
 }
 
+std::string Simulator::convexHullDecomposition(
+    const std::string& filename,
+    const std::string& CHDFilename,
+    const assets::ResourceManager::VHACDParameters& params) {
+  Cr::Utility::Debug() << "VHACD PARAMS RESOLUTION: " << params.m_resolution;
+  resourceManager_->convexHullDecomposition(filename, CHDFilename, params);
+  // 2. create object attributes for the trajectory
```
jturner65, line 1085: Update this comment please :).

src/esp/sim/Simulator.cpp
```diff
@@ -1076,6 +1076,27 @@ scene::SceneNode* Simulator::loadAndCreateRenderAssetInstance(
...
+  auto objAttrMgr = metadataMediator_->getObjectAttributesManager();
+  auto CHDObjAttr = objAttrMgr->createObject(filename, false);
+  // turn off collisions
```
jturner65, line 1088: Update this comment as well.

src/esp/sim/Simulator.h
```diff
@@ -845,6 +845,20 @@ class Simulator {
...
       const assets::RenderAssetInstanceCreationInfo& creation);
 
+  /**
+   * @brief Runs convex hull decomposition on a specified file.
+   *
+   * @param filename The MeshMetaData filename to be converted.
+   * @param CHDFilename The new filename for the CHD collision mesh.
+   * @param params VHACD params that specify resolution, vertices per convex
+   * hull, etc.
+   */
```
jturner65, line 855: Need to have @return here.  Also, need to expand comment to cover all of the functionality (", creates an object attributes referencing the convex hull asset, and returns the handle of this attributes" or something to that effect.)

src/esp/sim/Simulator.cpp
jturner65, line 1088: Update this comment as well.

src/esp/sim/Simulator.cpp
jturner65, line 1085: Update this comment please :).

src/esp/bindings/bindings.cpp
```diff
@@ -18,6 +19,15 @@ void initEspBindings(py::module& m) {
...
       .def_property_readonly("center", &box3f::center);
+  py::class_<assets::ResourceManager::VHACDParameters,
+             assets::ResourceManager::VHACDParameters::ptr>(m,
+                                                            "VHACDParameters")
+      .def(py::init(&assets::ResourceManager::VHACDParameters::create<>))
+      .def_property("resolution",
+                    &assets::ResourceManager::VHACDParameters::getResolution,
+                    &assets::ResourceManager::VHACDParameters::setResolution,
+                    R"(Type of the mesh asset used to render constructions built
+          from this template.)");
```
jturner65, line 30: This comment is incorrect, I think.

src/esp/assets/ResourceManager.h
```diff
@@ -323,6 +335,44 @@ class ResourceManager {
...
+
+  /**
+   * @brief Runs convex hull decomposition on a specified file.
+   *
+   * @param filename The MeshMetaData filename to be converted.
+   * @param CHDFilename The new filename for the CHD collision mesh.
+   * @param params VHACD params that specify resolution, vertices per convex
+   * hull, etc.
+   */
+  void convexHullDecomposition(
```
jturner65, line 372: perhaps `createConvexHullDecomposition` ?

jturner65 (commented):
Some minor comments.  Looking good!

Skylion007:
@ScottCSteinhauser please run clang-tidy locally as there are a lot of style errors and performance issues that need to be fixed on this PR.

src/esp/sim/Simulator.cpp
```diff
@@ -573,10 +573,9 @@ void Simulator::reconfigureReplayManager() {
   resourceManager_->setRecorder(gfxReplayMgr_->getRecorder());
 
   // provide Player callback to replay manager
-  gfxReplayMgr_->setPlayerCallback([this](auto&& PH1, auto&& PH2) {
-    loadAndCreateRenderAssetInstance(std::forward<decltype(PH1)>(PH1),
-                                     std::forward<decltype(PH2)>(PH2));
-  });
+  gfxReplayMgr_->setPlayerCallback(
+      std::bind(&esp::sim::Simulator::loadAndCreateRenderAssetInstance, this,
```
Skylion007: why was this change needed? Isn't std::bind less performant here?

src/esp/sim/Simulator.cpp
```diff
@@ -122,7 +122,7 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
                     "initialized with True.  Call close() to change this.";
   }
 
-  bool success;
+  bool success = 0;
```
Skylion007, line 125: Incorrect initialization, see clang-tidy. Should be false.

src/esp/sim/Simulator.cpp
```diff
@@ -1076,6 +1076,58 @@ scene::SceneNode* Simulator::loadAndCreateRenderAssetInstance(
...
+  CHDObjAttr->setIsCollidable(true);
+  CHDObjAttr->setComputeCOMFromShape(false);
+  CHDObjAttr->setCollisionAssetIsPrimitive(false);
+  CHDObjAttr->setJoinCollisionMeshes(false);
+
+  // if the renderCHD flag is set to true, set the convex hull decomposition to
+  // be the render asset (useful for testing)
+  if (renderCHD) {
+    CHDObjAttr->setRenderAssetHandle(CHDFilename);
+    CHDObjAttr->setRenderAssetIsPrimitive(false);
```
Skylion007, line 1120: This should be false, either way, right? Why is this line duplicated?

src/tests/ResourceManagerTest.cpp
```diff
@@ -86,6 +86,50 @@ TEST(ResourceManagerTest, createJoinedCollisionMesh) {
...
+  // params.setMaxNumVerticesPerCH(10);
+  params.m_resolution = 1000000;
+
+  resourceManager.createConvexHullDecomposition(donutFile, CHdonutFile, params);
+  std::cout << "Nope :/" << std::endl;
+  // transform_box.glb is composed of 6 identical triangulated plane meshes
+  // transformed into a cube via a transform heirarchy. Combined, the resulting
+  // mesh should have 24 vertices and 36 indices with corners at the unit corner
+  // coordinates as defined in the ground truth vectors below.
+  ASSERT(1 > 0);
```
Skylion007, line 130: There has to be an assert macro for this, right @mosra

src/esp/assets/ResourceManager.h
```diff
@@ -84,6 +87,13 @@ class ResourceManager {
   /** @brief Convenience typedef for Importer class */
   using Importer = Mn::Trade::AbstractImporter;
 
+  struct VHACDParameters : VHACD::IVHACD::Parameters {
+    VHACDParameters() { m_oclAcceleration = false; }
```
Skylion007, line 91: Any reason we don't want OCL acceleration? Should add comment as to why.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2185,208 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+    }
+  }
+  Cr::Utility::Directory::appendString(
+      filepath + "/" + new_filename,
+      "# Number of vertices: " + std::to_string(numVertices) + "\n\n");
+
+  // Now do second pass to write indices for each group (node)
+  int globalVertexNum = 1;
+  int numParts = 1;
+  for (MeshTransformNode node : metaData.root.children) {
```
Skylion007, line 2237: Pretty sure this is a unnecessary copy, see clang-tidy fixes

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2185,208 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+  Cr::Utility::Directory::writeString(filepath + "/" + new_filename,
+                                      "# CHD Mesh group \n");
+  // write vertex info to file
+  int numVertices = 0;
+  for (MeshTransformNode node : metaData.root.children) {
+    CollisionMeshData& meshData =
+        meshes_.at(node.meshIDLocal + metaData.meshIndex.first)
+            ->getCollisionMeshData();
+    for (auto& pos : meshData.positions) {
+      Cr::Utility::Directory::appendString(filepath + "/" + new_filename,
```
Skylion007, line 2223: may want to consider a string build here

Skylion007 (changes requested):
A lot of clang-tidy fixes needed, please fix. Most fixes can be automatically done with clang-tidy -fix

src/esp/sim/Simulator.cpp
ScottCSteinhauser: This was giving me an error for some reason. I got std::bind from the current master branch of habitat.

src/esp/assets/ResourceManager.cpp
mosra: Hm. I would actually fully build the string first and then write it at once. `appendString()` opens and closes the file each time and so it's way slower than it should be.

For the string building, [Utility::formatInto()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#aaf6e2f081fa8fd67065dd7b39811f5f6) is what you're looking for, the doc even has a snippet that is not too far from what you do here.

src/esp/assets/ResourceManager.h
mosra: Simple: it's broken :D Both @aclegg3 and me verified that. (Neither of us bothered to check *why*, though.)

src/esp/sim/Simulator.h
mosra: `@returns` [works too](https://www.doxygen.nl/manual/commands.html#cmdreturns), nevertheless consistency is king :)

src/tests/ResourceManagerTest.cpp
mosra: I have no idea what this checks :sweat_smile: 

The test should check at least something, otherwise it can be minimized down to `ASSERT(true)`.

src/esp/sim/Simulator.cpp
```diff
@@ -1076,6 +1076,58 @@ scene::SceneNode* Simulator::loadAndCreateRenderAssetInstance(
...
+std::string Simulator::convexHullDecomposition(
+    const std::string& filename,
+    const assets::ResourceManager::VHACDParameters& params,
+    const bool renderCHD,
+    const bool saveCHDToObj) {
+  Cr::Utility::Debug() << "VHACD PARAMS RESOLUTION: " << params.m_resolution;
+
+  // generate a unique filename
+  std::string CHDFilename =
+      filename.substr(0, filename.find_last_of('.')) + "CHD";
```
mosra, line 1088: [[Utility::Directory::splitExtension(filename)](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1Directory.html#a33b62a435a580891394d5eef5ef40fbb)] ;)

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2185,208 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+  // fill points and triangles with mesh data
+  std::vector<float> points = std::vector<float>();
+  std::vector<uint32_t> triangles = std::vector<uint32_t>();
+  getPrimitiveMeshData(filename, points, triangles);
+
+  // use VHACD
+  bool res =
+      interfaceVHACD->Compute(&points[0], (unsigned int)(points.size() / 3),
+                              (const uint32_t*)&triangles[0],
+                              (unsigned int)(triangles.size() / 3), params);
```
mosra, line 2284: Hm, since you're passing raw pointers here, is the `getPrimitiveMeshData()` function and all the data copying really necessary? The `vbo` is also a list of floats, only grouped to three-component vectors.

Same below, unless VHACD is *really badly designed* (which I don't think it is), magnum's MeshData should be able to reference the data from `ch` directly, without having to copy again. What's the type of `ch.m_points` and `ch.m_triangles`?

src/tests/ResourceManagerTest.cpp
ScottCSteinhauser: I was using this test just to check whether or not the VHACD functions run without segfaults - didn't really require the ASSERT statement. Is there a different place I should put this if I'm not Asserting anything? @mosra @Skylion007

src/tests/ResourceManagerTest.cpp
```diff
@@ -86,6 +86,50 @@ TEST(ResourceManagerTest, createJoinedCollisionMesh) {
...
+
+  esp::assets::MeshData::uptr joinedBox =
+      resourceManager.createJoinedCollisionMesh(donutFile);
+
+  esp::assets::ResourceManager::VHACDParameters params;
+  // params.setMaxNumVerticesPerCH(10);
+  params.m_resolution = 1000000;
+
+  resourceManager.createConvexHullDecomposition(donutFile, CHdonutFile, params);
+  std::cout << "Nope :/" << std::endl;
```
jturner65, line 125: Needs to be removed :).

src/esp/assets/ResourceManager.h
```diff
@@ -323,6 +333,55 @@ class ResourceManager {
...
+   * @brief Get primitive mesh data (points and triangle arrays) from a
+   * MeshMetaData filename.
+   *
+   * @param filename The MeshMetaData filename to receive primitive attributes
+   * from.
+   * @param points The vector of points to be filled with the mesh vertices.
+   * @param triangles The vector of indices to be filled with the mesh index
+   * buffer.
+   */
+  void getPrimitiveMeshData(const std::string& filename,
```
jturner65, line 346: This is a minor thing: since this doesn't "get"/return anything, can we change the name to something like buildPrimitiveMeshData? Also, mark `points` and `triangles` as [in] variables in docs, please :).

ScottCSteinhauser: Good point. As far as the docs go, what is the difference between @param[in] and @param[out]?

jturner65: Oops.  that should have been "[out]", my mistake.

src/esp/assets/ResourceManager.cpp
ScottCSteinhauser: Good catch, passing vbo and ibo straight in worked! I think my assumption was that a vec3f couldn't be iterated on in the same way as 3 floats, but looks like it can :)

setup.py
```diff
@@ -50,6 +50,12 @@ def build_parser():
         action="store_true",
         help="""Build with Bullet simulation engine.""",
     )
+    parser.add_argument(
+        "--vhacd",
```
Skylion007, line 54: Should we keep adding flags needed for the full feature suite? At some point, shouldn't bullet and VHACD become opt out params? @aclegg3 ?

Skylion007: Also by not updating cirlcelci to build with VHACD, none of this code will be checked by clang-tidy?

src/tests/ResourceManagerTest.cpp
aclegg3: I suggest checking something about the output instead. Could check that the exported file was created correctly or you could check that the new render asset is registered somehow.

setup.py
aclegg3: Yeah, that's a good point. We keep Bullet as opt-in because many standard use cases for Habitat don't require dynamics at all. We don't want to impact those users. Maybe we reconsider this once more interactive tasks hit Hab-lab. 

I think it is also ok to keep VHACD as opt-in since there are few use use cases for it at this time.

Agreed that we should add the flag to CI for testing.

Skylion007: Actually, it's built twice in CI, once for clang-tidy/benchmarking and another time for testing.

Skylion007:
@ScottCSteinhauser You are also going to need to add error handling the physics benchmarking script since VHACD isn't built there or skipping if VHACD isn't built. 

Also, we need to add VHACD as a build option to the conda script. I don't see any reason to not include it by default in the generated conda builds.

ScottCSteinhauser:
> @ScottCSteinhauser You are also going to need to add error handling the physics benchmarking script since VHACD isn't built there or skipping if VHACD isn't built.
> 
> Also, we need to add VHACD as a build option to the conda script. I don't see any reason to not include it by default in the generated conda builds.



> @ScottCSteinhauser You are also going to need to add error handling the physics benchmarking script since VHACD isn't built there or skipping if VHACD isn't built.
> 
> Also, we need to add VHACD as a build option to the conda script. I don't see any reason to not include it by default in the generated conda builds.

Yeah for sure - are these changes I need to make to the CircleCI/config.yml file?

Skylion007:
No, you should modify the tests with the appropriate skip connections. The
build params are in conda-build.

.circleci/config.yml
```diff
@@ -507,6 +507,7 @@ jobs:
               # run tests with code coverage
               CORRADE_TEST_COLOR=ON GTEST_COLOR=yes ./build.sh --headless \
                 --bullet \
+                --vhacd \
```
Skylion007, line 510: You also need to add it above so that when clang-tidy is run, it is also built with VHACD so that the interface gets evaluated by clang-tidy.

ScottCSteinhauser:
> You should still guard so that pytest can pass if the it's not built with VHACD. You can probably hack this with a pytest_import skip or a VHACD_enabled constant similar to BULLET.

Hmm I'm not sure how to set/check a constant like that. Is there an example where that is done with bullet you could point me to?

Skylion007:
See the Bullet and CUDA tests in the test suite.

Skylion007:
Almost there, you just need to add the --vhacd arg to this line: https://github.com/facebookresearch/habitat-sim/blob/67e79c09ca0a7d3d82fb4fe228f1e50f5d0163eb/.circleci/config.yml#L446 and also pull from master and ensure it passes the new ClangTidy settings.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2188,188 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+    meshMetaData.root.children.push_back(transformNode);
+  }
+
+  // default material for now
+  auto phongMaterial = gfx::PhongMaterialData::create_unique();
+  phongMaterial->specularColor = {1.0, 1.0, 1.0, 1.0};
+  phongMaterial->ambientColor = Magnum::Color4{1.0};
+  ;
+  phongMaterial->diffuseColor = Magnum::Color4{0.9};
+  ;
```
mosra, line 2344: Weren't there some "reasonable defaults" inside PhongMaterialData already?

src/esp/sim/Simulator.cpp
```diff
@@ -1076,6 +1076,59 @@ scene::SceneNode* Simulator::loadAndCreateRenderAssetInstance(
...
+std::string Simulator::convexHullDecomposition(
+    const std::string& filename,
+    const assets::ResourceManager::VHACDParameters& params,
+    const bool renderCHD,
+    const bool saveCHDToObj) {
+  Cr::Utility::Debug() << "VHACD PARAMS RESOLUTION: " << params.m_resolution;
+
+  // generate a unique filename
+  std::string CHDFilename =
+      Cr::Utility::Directory::splitExtension(filename).first + "CHD";
```
mosra, line 1089:
```suggestion
      Cr::Utility::Directory::splitExtension(filename).first + ".CHD";
```

the period is not contained in `.first`, so you have to add it again

also maybe lowercase? some filesystems are case insensitive and this causes unnecessary friction

src/esp/sim/Simulator.cpp
```diff
@@ -1076,6 +1076,59 @@ scene::SceneNode* Simulator::loadAndCreateRenderAssetInstance(
...
+#ifdef ESP_BUILD_WITH_VHACD
+std::string Simulator::convexHullDecomposition(
+    const std::string& filename,
+    const assets::ResourceManager::VHACDParameters& params,
+    const bool renderCHD,
+    const bool saveCHDToObj) {
+  Cr::Utility::Debug() << "VHACD PARAMS RESOLUTION: " << params.m_resolution;
+
+  // generate a unique filename
+  std::string CHDFilename =
```
mosra, line 1088:
```suggestion
  std::string chdFilename =
```

naming policy: variables should start with lowercase; same elsewhere

src/esp/assets/ResourceManager.cpp
```diff
@@ -2184,5 +2188,188 @@ std::unique_ptr<MeshData> ResourceManager::createJoinedCollisionMesh(
...
+  Cr::Utility::Debug() << "Num Convex Hulls: " << nConvexHulls;
+  Cr::Utility::Debug() << "Resolution: " << params.m_resolution;
+  VHACD::IVHACD::ConvexHull ch{};
+  std::unique_ptr<GenericMeshData> genCHMeshData;
+  for (unsigned int p = 0; p < nConvexHulls; ++p) {
+    // for each convex hull, transfer the data to a newly created  MeshData
+    interfaceVHACD->GetConvexHull(p, ch);
+
+    std::vector<Magnum::UnsignedInt> indices;
+    std::vector<Magnum::Vector3> positions;
```
mosra, line 2281: [From my previous comment](https://github.com/facebookresearch/habitat-sim/pull/1055#discussion_r573129296) -- this doesn't seem to be done yet:

> Same below, unless VHACD is *really badly designed* (which I don't think it is), magnum's MeshData should be able to reference the data from `ch` directly, without having to copy again.

src/esp/sim/Simulator.cpp
```diff
@@ -1076,6 +1076,59 @@ scene::SceneNode* Simulator::loadAndCreateRenderAssetInstance(
...
+  CHDObjAttr->setCollisionAssetIsPrimitive(false);
+  CHDObjAttr->setJoinCollisionMeshes(false);
+
+  // if the renderCHD flag is set to true, set the convex hull decomposition to
+  // be the render asset (useful for testing)
+  if (renderCHD) {
+    CHDObjAttr->setRenderAssetHandle(CHDFilename);
+  } else {
+    CHDObjAttr->setRenderAssetHandle(filename);
+  }
```
mosra, line 1122: code golf:

```suggestion
  CHDObjAttr->setRenderAssetHandle(renderCHD ? CHDFilename : filename);
```

src/esp/sim/Simulator.h
```diff
@@ -845,6 +845,29 @@ class Simulator {
...
+   * @param saveCHDToObj Specifies whether or not to save the newly created
+   * convex hull asset to an obj file.
+   * @return The handle of the newly created object attributes.
+   */
+  std::string convexHullDecomposition(
+      const std::string& filename,
+      const assets::ResourceManager::VHACDParameters& params =
+          assets::ResourceManager::VHACDParameters(),
+      const bool renderCHD = false,
+      const bool saveCHDToObj = false);
```
mosra, line 868:
```suggestion
      bool renderCHD = false,
      bool saveCHDToObj = false);
```

No need to make values (as opposed to `&` references) `const` in a function declaration, has no effect and only adds noise to the generated documentation.

On the other hand, tn the *definition* in the cpp it's fine, and encouraged :)

src/esp/assets/ResourceManager.h
```diff
@@ -84,6 +88,17 @@ class ResourceManager {
   /** @brief Convenience typedef for Importer class */
   using Importer = Mn::Trade::AbstractImporter;
 
+#ifdef ESP_BUILD_WITH_VHACD
+  struct VHACDParameters : VHACD::IVHACD::Parameters {
```
mosra, line 92: Documentation for this struct?

src/esp/assets/ResourceManager.cpp
```diff
@@ -89,7 +91,9 @@ ResourceManager::ResourceManager(
       importerManager_("nonexistent")
 #endif
 {
-
+#ifdef ESP_BUILD_WITH_VHACD
+  interfaceVHACD = VHACD::CreateVHACD();
```
mosra, line 95: Doesn't this need to be freed in some way? I don't see any code to do so in the destructor.

noahm1216:
Good displays for the work!

Don't have anything else to add that hasn't been said

src/esp/assets/ResourceManager.cpp
ScottCSteinhauser: Hmm I'm not quite sure how to do this, or if it's needed. I tried "delete interfaceVHACD" in the destructor but it looks like the destructor for VHACD interface is protected. Would changing interfaceVHACD to be a ptr, and then just setting it to be a nullptr work?

src/esp/assets/ResourceManager.h
```diff
@@ -84,6 +88,17 @@ class ResourceManager {
...
+    VHACDParameters() {
+      m_oclAcceleration = false;  // OCL Acceleration does not work on VHACD
+    }
+    ESP_SMART_POINTERS(VHACDParameters)
+  };
+
+  VHACD::IVHACD* interfaceVHACD;
```
aclegg3, line 99: This could be a unique_ptr instead and then it would destruct itself when ResourceManager is destructed.

examples/tutorials/physics_benchmarking.py
```diff
@@ -141,9 +145,30 @@ def box_drop_test(
                 ]
             ]
             obj_template = obj_templates_mgr.get_template_by_ID(object_ids[-1])
+
             if "scale" in obj.keys():
```
Skylion007, line 149: NIT:
```suggestion
            if "scale" in obj:
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -1232,7 +1242,7 @@ scene::SceneNode* ResourceManager::createRenderAssetInstanceGeneralPrimitive(
   if (creation.scale) {
     // need a new node for scaling because motion state will override scale
     // set at the physical node
-    // perf todo: avoid this if unit scale
+    // perf etgebgutecgfeucdluivtrrtruvrgunc: avoid this if unit scale
```
Skylion007, line 1245: YUBI barf

src/esp/assets/ResourceManager.cpp
```diff
@@ -89,12 +92,19 @@ ResourceManager::ResourceManager(
...
-
+#ifdef ESP_BUILD_WITH_VHACD
+  interfaceVHACD = VHACD::CreateVHACD();
+#endif
   initDefaultLightSetups();
   initDefaultMaterials();
   buildImporters();
 }
 
+ResourceManager::~ResourceManager() {
```
Skylion007, line 103: This needs to be in an ifdef too

tests/test_physics_benchmarking.py
```diff
@@ -1,8 +1,10 @@
 import pytest
 
 import utils
+from habitat_sim.bindings import vhacd_enabled
```
Skylion007, line 4: Calling bindings from the test is not recommended. I am surprised the import linter didn't catch it.

Skylion007 (changes requested):

src/esp/assets/ResourceManager.cpp
mosra: > Would changing interfaceVHACD to be a ptr, and then just setting it to be a nullptr work?

Setting a raw pointer to null doesn't free any memory, no. Using a smart pointer for this won't work either because the destructor is protected.

I did some digging and it seems you have to call `interfaceVHACD->Clean()` and interfaceVHACD->Release()` in order to clean up. In my opinion that's a *really bad* interface (did the author of that thing ever hear about destructors?!), but what can we do, sigh.

src/esp/assets/ResourceManager.h
mosra: Can't, the destructor is protected. See [my comment above](https://github.com/facebookresearch/habitat-sim/pull/1055#discussion_r579428913).

src/esp/assets/ResourceManager.cpp
aclegg3: Haha, yeah, Scott and I chatted about this a couple hours ago. We arrived at the same conclusion. 🤷

tests/test_physics_benchmarking.py
ScottCSteinhauser: Why is it not a good idea? And what would the alternative be?

Skylion007: Nothing should be imported from bindings, it should have a dedicated namespace somewhere. 
```import habitat_sim
habitat_sim.vhacd_enabled
```
or
```
 from habitat_sim.vhacd import vhacd_enabled 
 ```
 
 We don't want the users to use bindings, it's mostly there for backwards compat.

**Review by Skylion007 — CHANGES_REQUESTED** [review]
Biggest remaining issue is this: https://github.com/facebookresearch/habitat-sim/pull/1055#discussion_r577714697

**Comment by ScottCSteinhauser** [comment]
> Biggest remaining issue is this: [#1055 (comment)](https://github.com/facebookresearch/habitat-sim/pull/1055#discussion_r577714697)

I spoke with mosra about this - the way I'm doing this for 

> Biggest remaining issue is this: [#1055 (comment)](https://github.com/facebookresearch/habitat-sim/pull/1055#discussion_r577714697)

I spoke to mosra about this and what I'm doing turns out to be the efficient option. I can't pass ch.m_positions directly in, and I resolved the issue for indices.

**Comment by mosra** [comment]
> I can't pass ch.m_positions directly in

Tangential -- if we realize VHACD is a performance bottleneck, it might be worth to investigate if it really needs to use doubles. Floats might be enough and much faster for all calculations.

**Comment by Skylion007** [comment]
@mosra There is also a performance optimized branch of VHACD on the repo that is going to be the next major version change. We may want to look at that.

**Comment by Skylion007** [comment]
Pull and resolve merge conflicts and ensure it passes clang-tidy afterwards.

File: examples/tutorials/VHACD_tutorial.py [hunk 40]
```diff
@@ -0,0 +1,234 @@
...
+                        -1 * dist,
+                    ]
+                ),
+            )
+            sim.set_object_motion_type(habitat_sim.physics.MotionType.KINEMATIC, cur_id)
+            vel_control = sim.get_object_velocity_control(cur_id)
+            vel_control.controlling_ang_vel = True
+            vel_control.angular_velocity = np.array([0, -1.56, 0])
+
+        # simulate for 3 seconds
```
aclegg3, line 202: typo

examples/tutorials/VHACD_tutorial.py
```diff
@@ -0,0 +1,234 @@
...
+    sensor_specs.append(sensor_spec)
+
+    # agent configuration
+    agent_cfg = habitat_sim.agent.AgentConfiguration()
+    agent_cfg.sensor_specifications = sensor_specs
+
+    return habitat_sim.Configuration(backend_cfg, [agent_cfg])
+
+
+def simulate(sim, dt=1.0, get_frames=True, data=None):
```
aclegg3, line 47: If we're not using the metric collection/logging for the tutorial script let's remove it.

src/esp/assets/ResourceManager.h
```diff
@@ -256,8 +276,8 @@ class ResourceManager {
    * initialization that may be required when @ref metadataMediator_ is changed.
    * @param _MM a reference to the new @ref metadataMediator_.
    */
-  void setMetadataMediator(metadata::MetadataMediator::ptr _MM) {
-    metadataMediator_ = _MM;
+  void setMetadataMediator(metadata::MetadataMediator::ptr MM) {
+    metadataMediator_ = std::move(MM);
```
aclegg3, line 280: Merge error?

Skylion007: Nah, this is right. Clang-Tidy recommends this for performance and style.

Skylion007:
https://github.com/kmammou/v-hacd/tree/performance-improvements VHACD more performant branch btw

ScottCSteinhauser:
Looks like the js_lint CircleCI tests, although I'm unsure why my changes would cause this to break. Any thoughts @Skylion007 @aclegg3 ?

Skylion007:
@ScottCSteinhauser Looks like something broke the entire CI... It's not just you.

Skylion007:
Seems like the error was transient. Working fine now.

examples/tutorials/VHACD_tutorial.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # simulate for 4 seconds
+        simulate(sim, dt=4, get_frames=True, data=data)
+
+        for cur_id in cur_ids:
+            vel_control = sim.get_object_velocity_control(cur_id)
+            vel_control.controlling_ang_vel = True
+            vel_control.angular_velocity = np.array([-1.56, 0, 0])
+
+        # simulate for 3 seconds
```
aclegg3, line 192: typo

aclegg3 (approved):
Looks good. When tests pass, let's merge.

Skylion007 (approved):
Looks good to me assuming tests pass.

PR #1060, opened 2021-01-28
--Verify torch and cuda are installed if required by python tests

jturner65:
## Motivation and Context
This PR introduces code that will skip pytest tests that require soft dependencies that may not be installed, such as torch or torch with cuda

## Types of changes
Python tests pass, appropriately skip non-cuda installs.

## Checklist

### DIFF
--- a/tests/test_profiling_utils.py
+++ b/tests/test_profiling_utils.py
@@ -10,6 +10,9 @@ import os
 from io import StringIO
 from unittest.mock import patch
 
+import pytest
+
+import habitat_sim
 from habitat_sim.utils import profiling_utils
 
 _ENV_VAR_NAME = "HABITAT_PROFILING"
@@ -17,7 +20,8 @@ _ENV_VAR_NAME = "HABITAT_PROFILING"
 # Based on the env var, reloading the profiling_utils module should set
 # profiling_utils._enable_profiling to True or False.
 def test_env_var_enable():
-
+    if not habitat_sim.cuda_enabled:
+        pytest.skip("No CUDA present so skipping test_env_var_enable test")
     # test with env var not set
     os.environ.pop(_ENV_VAR_NAME, None)
     importlib.reload(profiling_utils)
@@ -47,6 +51,8 @@ def test_env_var_enable():
 
 # Create nested ranges and verify the code runs without error.
 def test_nested_range_push_pop():
+    if not habitat_sim.cuda_enabled:
+        pytest.skip("No CUDA present so skipping test_nested_range_push_pop test")
 
     os.environ[_ENV_VAR_NAME] = "1"
     importlib.reload(profiling_utils)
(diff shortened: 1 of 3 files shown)


Skylion007 (changes requested):
For this, you should be using pytest.mark.skipif https://docs.pytest.org/en/reorganize-docs/new-docs/user/skipping.html

jturner65:
> For this, you should be using pytest.mark.skipif https://docs.pytest.org/en/reorganize-docs/new-docs/user/skipping.html

Ok.  I just used the mechanism I found in test_sensors.py

tests/test_profiling_utils.py
```diff
@@ -47,6 +51,8 @@ def test_env_var_enable():
 
 # Create nested ranges and verify the code runs without error.
 def test_nested_range_push_pop():
+    if not habitat_sim.cuda_enabled:
```
erikwijmans, line 54: Why do these tests require that habitat-sim was compiled with CUDA?  That is different than there being CUDA capable devices on the system

tests/test_profiling_utils.py
```diff
@@ -11,17 +11,19 @@
 from unittest.mock import patch
 
 import pytest
+from torch import cuda
```
Skylion007, line 14: Torch is a soft dependency so it shouldn't error if Torch isn't installed, it should just skip tests.

Skylion007 (changes requested):
Make torch a soft dependency

tests/test_profiling_utils.py
```diff
@@ -10,14 +10,20 @@
...
+import pytest
+from torch import cuda
+
 from habitat_sim.utils import profiling_utils
 
 _ENV_VAR_NAME = "HABITAT_PROFILING"
 
 # Based on the env var, reloading the profiling_utils module should set
 # profiling_utils._enable_profiling to True or False.
+@pytest.mark.skipif(
```
Skylion007, line 22: There all identical, you can define the skip if once and then use it everywhere
```
requires_cuda =  pytest.mark.skipif( ...
```
```
@requires_cuda
```

jturner65: Message is different, but otherwise yeah, these are the same check.

tests/test_profiling_utils.py
jturner65: There are other tests that require but do not verify torch installation (test_data_extraction.py, for example).  Not sure what you want to do there.

Skylion007: Yeah, @jturner65 been meaning to fix that.

Skylion007 (changes requested):

tests/test_profiling_utils.py
Skylion007: Just use a try: except ImportError boolean

jturner65:
Added torch import check in test_data_extractor.py as well

tests/test_profiling_utils.py
jturner65: pytest.importorskip seems a bit more elegant, no?

Skylion007 (approved):

PR #1063, opened 2021-01-29
Update version to 0.1.7

aclegg3:


habitat_sim/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 import builtins
 
-__version__ = "0.1.6"
+__version__ = "0.1.7"
```
Skylion007, line 9: The conda version also need to be bumped, no? @mathfac Do we have a document on how to do releases?

mathfac: conda version is taken from this value.

mathfac (approved):

PR #1064, opened 2021-01-29
[Sensor] Decouple sensor initialization from Agent

vauduong:
## Motivation and Context
Currently, agents are constructed using a valid scene node reference and an agent configuration in the c++ habitat-sim function Simulator::addAgent. The agent configuration specifies sensor specifications, and the sensors are initialized in the agent constructor using the provided scene node.   

This PR just focuses on the C++ side, but another PR can be made in the future to bring the functionality to python. 

Sensors can be constructed using any valid scene node, so we wish to decouple sensor initialization from the agent. In this PR, we:

- create a factory class SensorFactory, which contains a static method to return a SensorSuite given a reference to a SceneNode and reference to a vector of SensorSpec::ptrs called SensorSetup. 
- Add method  SensorSuite::merge to merge the sensors from another SensorSuite into its existing map of sensors
- SensorFactory is called in Simulator::AddAgent to provide sensors. 
- Add method Simulator::addSensorToObject to add a sensor at a node, given the node's objectId and the specification of the sensor
- Add test in SimTest to test adding a sensor to an object
- Add test in SensorTest to test SensorFactory methods
- Methods in viewer to demo adding sensors to objects has been moved to https://github.com/vauduong/habitat-sim/tree/add-sensors-to-object-viewer-demo

SensorCameraWeeklyUpdates 3 pptx

To demo this functionality, methods were added to Simulator to keep track of Sensors and get Sensors by object ID; and methods were added to Viewer to add CameraSensors to objects and toggle between agent and object cameras. You can see a video of a banana being pushed here:

https://user-images.githubusercontent.com/15386768/107078519-0f18f480-67b4-11eb-95e3-8cce26c980b0.mov

## How Has This Been Tested
Build and run with demo in viewer
Add test to add object in SimTest
Add test to test SensorFactory in SensorTest

## Types of changes

## Checklist


src/esp/sensor/SensorFactory.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include "esp/scene/SceneNode.h"
+#include "esp/sensor/CameraSensor.h"
+#include "esp/sensor/Sensor.h"
+
+namespace esp {
+sensor::SensorSuite SensorFactory::createSensors(
+    scene::SceneNode& node,
+    const sensor::SensorSetup& sensorSetup) {
+  sensor::SensorSuite sensorSuite = sensor::SensorSuite();
+  for (sensor::SensorSpec::ptr spec : sensorSetup) {
```
Skylion007, line 12: This is already decoupled on the Python side FYI. It's even factored out on the Python side to a sensor can be added without reconfiguring the sim. Seems like this should be refactored to also have a function to add a single sensor like we have on the Python side ```agent.add_sensor()...```

bigbike: Hi @Skylion007 
If possible, I sincerely hope you can help us to take a look at the internal task (I will send you the task number via slack), where we gave concrete implementation steps. Check if something is contradictory to what has been done in the Python side. Thanks!

src/esp/sensor/SensorFactory.h
```diff
@@ -0,0 +1,16 @@
...
+namespace esp {
+class SensorFactory {
+ public:
+  virtual ~SensorFactory();
+  static sensor::SensorSuite createSensors(
+      scene::SceneNode& node,
+      const sensor::SensorSetup& sensorSetup);
+
+ private:
+  ESP_SMART_POINTERS(SensorFactory)
```
jturner65, line 14: Don't think you need this either, since we're not going to make SensorFactory objects.  
Besides, if we do and we want to reference them as smart pointers, this macro needs to be public for it to be usable.

src/esp/sensor/SensorFactory.h
```diff
@@ -0,0 +1,16 @@
+#include "esp/scene/SceneNode.h"
+#include "esp/sensor/CameraSensor.h"
+#include "esp/sensor/Sensor.h"
+
+namespace esp {
+class SensorFactory {
+ public:
+  virtual ~SensorFactory();
```
jturner65, line 8: Why declare a destructor for this class? Probably not going to create an object of this class ever.

src/esp/agent/Agent.h
```diff
@@ -112,6 +112,10 @@ class Agent : public Magnum::SceneGraph::AbstractFeature3D {
 
   void setState(const AgentState& state, const bool resetSensors = true);
 
+  void setSensorSuite(sensor::SensorSuite sensorSuite) {
+    sensors_ = sensorSuite;
```
jturner65, line 116: This argument should be passed by reference, instead of making a copy.

src/esp/sensor/SensorFactory.cpp
jturner65: > This is already decoupled on the Python side FYI. It's even factored out on the Python side to a sensor can be added without reconfiguring the sim. Seems like this should be refactored to also have a function to add a single sensor like we have on the Python side `agent.add_sensor()...`

As an aside, it looks like the sensor creation for the agent in agent.py happens in agent.reconfigure when the agent is created, regardless of whether created sensors are passed into the class or not, which seems like a bug, no? 

Pertinent Agent.py agent code

```
    def __init__(
        self,
        scene_node: hsim.SceneNode,
        agent_config: Optional[AgentConfiguration] = None,
        _sensors: Optional[SensorSuite] = None,
        controls: Optional[ObjectControls] = None,
    ) -> None:
        self.agent_config = agent_config if agent_config else AgentConfiguration()
        
        # Assigning pre-existing sensors is ignored in below reconfigure call
        self._sensors = _sensors if _sensors else SensorSuite()
< ... >        
        # Reconfigure always called with default reconfigure_sensors == True
        self.reconfigure(self.agent_config)
< ... >        

    def reconfigure(
        self, agent_config: AgentConfiguration, reconfigure_sensors: bool = True
    ) -> None:
< ... >        
        # This is always True if called from self.reconfigure called from constructor
        if reconfigure_sensors:
            self._sensors.clear()
            for spec in self.agent_config.sensor_specifications:
                self._add_sensor(spec, modify_agent_config=False)

```

jturner65 (commented):
Some suggestions.

src/esp/agent/Agent.h
jturner65: So : 

`void setSensorSuite(const sensor::SensorSuite& sensorSuite)`

jturner65 (commented):
Please remember to document your added functions.

src/esp/agent/Agent.h
erikwijmans: Shouldn't this be `sensor::SensorSuite&& sensorSuite` as the arg and then `sensors_ = std::move(sensorSuite);`?  `sensors_` is an instance of sensor::SenorSuite, so the assignment will make a copy currently

src/esp/sensor/SensorFactory.cpp
jturner65, line 12: Fix this loop using the following code : 
`for (const sensor::SensorSpec::ptr& spec : sensorSetup)`

jturner65 (commented):
Fix for clang-tidy

src/esp/agent/Agent.h
jturner65: I considered mentioning that, but I wasn't sure if we would want to retain the sensorSuite we pass to this function in the calling code.  If that is not a concern, then the move would probably be a better choice.

jturner65: @bigbike Thoughts?

bigbike: I think @erikwijmans made a very good point here. Nice catch! We should directly move it instead of copying.
Actually I would even consider deleting the copy constructor in the sensorSuite.

src/esp/bindings/SimBindings.cpp
```diff
@@ -93,10 +93,14 @@ void initSimBindings(py::module& m) {
       .def_property_readonly(
           "gfx_replay_manager", &Simulator::getGfxReplayManager,
           R"(Use gfx_replay_manager for replay recording and playback.)")
+      .def_property_read_only("sensor_suite", &Simulator::getSensorSuite)
```
Skylion007, line 96: This is going to conflict with the Simulator python class. Furthermore, the Python sensorsuite is entirely seperate from the C++ one, we need to do more work before merging them.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -156,8 +157,14 @@ void initSensorBindings(py::module& m) {
   py::class_<SensorSuite, SensorSuite::ptr>(m, "SensorSuite")
       .def(py::init(&SensorSuite::create<>))
       .def("add", &SensorSuite::add)
+      .def("merge", &SensorSuite::merge)
       .def("get", &SensorSuite::get, R"(get the sensor by id)");
 
+  // ==== SensorFactory ====
+  py::class_<SensorFactory>(m, "SensorFactory")
```
Skylion007, line 164: This should be a factory method on the Sensorsuite, there is no reason for this particular class to be bound in Python like so. You should add python tests to see the divergent behavior.

src/esp/agent/Agent.h
vauduong: Thank you! That's fixed now.

src/esp/bindings/SimBindings.cpp
vauduong: Thanks for the feedback Aaron, I see now that the python sensor suite is a Dict[str, hsim.Sensor], while the C++ SensorSuite is std::map<std::string, Sensor::ptr>. In this PR, we are making changes that we'd like to see on the C++ side, but I don't know enough about the python side to know if we'd like to see the same changes. It seems like the functionality to just add sensors already exists, right? Do you think I should modify how python sensors work to add them using this SensorFactory method, or just leave it as is?

vauduong: Oh I think I understand now, we should implement a similar SensorFactory method on the python side to create sensors.

bigbike:
@Skylion007 : when you have time, can you take a look at the slack channel? Thanks!

src/esp/sim/Simulator.h
```diff
@@ -251,6 +254,14 @@ class Simulator {
    */
   std::vector<int> getExistingObjectIDs(int sceneID = 0);
 
+  /**
+   * @brief Get the SceneNode of a physics object instanced in a physical scene.
+   * See @ref esp::physics::PhysicsManager::getExistingObjectIDs.
+   * @param objectID The ID of the object identifying it in @ref
+   * @return A reference to the SceneNode
+   */
+  esp::scene::SceneNode& getObjectNode(const int objectId);
```
erikwijmans, line 263: Isn't this the same as getObjectSceneNode?

src/esp/sim/Simulator.h
```diff
@@ -716,6 +727,16 @@ class Simulator {
...
 
+  /**
+   * @brief Initialize sensor and attach to sceneNode of a particular object
+   * with uuid objectName_objectId
+   * @param objectId    Id of the object to which a sensor will be initialized
+   * at its node
+   * @param objectName Name of the object which will be used in sensor's uuid
+   *
+   */
+  void addSensorToObject(const int objectId, const std::string& objectName);
```
erikwijmans, line 738: This seems like a super specific method to have on the simulator class.  It's nice for a demo, but because it takes just an ID and a name, it isn't generally useful because the user has no control over the sensors position or orientation with moving the object.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -156,6 +157,7 @@ void initSensorBindings(py::module& m) {
   py::class_<SensorSuite, SensorSuite::ptr>(m, "SensorSuite")
```
erikwijmans, line 157: Seems like this never got removed, it should have been.  Nothing uses this -- it's just a dictionary so python just uses a `dict`

src/esp/sim/Simulator.cpp
```diff
@@ -1170,6 +1197,10 @@ bool Simulator::drawObservation(const int agentId,
   return false;
 }
 
+bool Simulator::drawObservation(esp::sensor::VisualSensor* visualSensor) {
```
erikwijmans, line 1200: This should be a reference.  Pointer implies it's a nullable or the called method will change what it points to.

src/esp/sim/Simulator.h
vauduong: Thanks Erik! I added this function because I couldn't find sceneID anywhere in viewer, which is a parameter to getObjectSceneNode. But now I see that there are some functions in Simulator which have sceneID = 0 by default, so we could do that for getObjectSceneNode. I'll make that change!

src/esp/sim/Simulator.h
aclegg3: Agreed, this could be a nice utility (possibly in python utils?) given an optional transformation. You can create a new child SceneNode and apply the transformation there. Also think about what happens to the Sensor when the object is deleted.

vauduong: Thanks, that makes a lot of sense - I'll look into adding an optional transformation and making sure to delete the sensor when the object is deleted. 

I see that there is a method Simulator::setTransformation(const Magnum::Matrix4& transform, const int objectID, const int sceneID). Since the sensor is attached to the node at objectID, the sensor will move as well upon transformation, but I imagine that we would want another optional transform so that we could have a 3rd person view of an object as well! Am I understanding this right?

I think a function like this is needed to provide add a sensor to an object, but maybe this is not the best place. Would you recommend moving this to another class?

src/esp/sim/Simulator.cpp
```diff
@@ -664,6 +666,21 @@ void Simulator::removeObject(const int objectID,
       // resourceManager_->removeResourceByName(trajVisAssetName);
     }
   }
+  // if (deleteObjectNode) {
```
vauduong, line 669: Thanks everyone for the helpful feedback! @aclegg3 I did realize Simulator::removeObject was causing a crash, I think from a sensor being attached to a node that no longer exists. 

To fix this, I tried freeing the memory at the pointer to a sensor and/or erasing a sensor from a SensorSuite before PhysicsManager::removeObject is called. I also tried deleting the object scene node after the physical scene node and visual scene node were deleted, so that the destructor would delete things attached to the node. None of these worked, so I'm not sure what to do. 

Does anyone have any guidance for deleting a sensor when its node is deleted?

jturner65: One easy thing - any deleting here should be within `if (sceneHasPhysics(sceneID)) {` check.

src/esp/sim/Simulator.cpp
```diff
@@ -1122,6 +1144,25 @@ agent::Agent::ptr Simulator::getAgent(const int agentId) {
   return agents_[agentId];
 }
 
+void Simulator::addSensorToObject(const int objectId,
+                                  const vec3f& position,
+                                  const vec3f& orientation,
+                                  const vec2i& resolution) {
+  esp::sensor::SensorSpec::ptr objectSensorSpec =
+      esp::sensor::SensorSpec::create_unique();
```
jturner65, line 1152: Shouldn't this be `esp::sensor::SensorSpec::create()`?  Everywhere you reference these SensorSpecs you treat them as shared pointers, not unique ones.

src/esp/sim/Simulator.cpp
jturner65: All the delete code should be within the `if (sceneHasPhysics(sceneID)) {` check.

src/esp/sim/Simulator.cpp
```diff
@@ -664,6 +666,21 @@ void Simulator::removeObject(const int objectID,
...
+  // if (deleteObjectNode) {
+  //   // delete sensor at ObjectNode if there is one
+  //   std::map<std::string, esp::sensor::Sensor::ptr>::iterator sensorIterator
+  //   =
+  //       getSensorSuite().getSensors().find(std::to_string(objectID));
+  //   if (sensorIterator != getSensorSuite().getSensors().end()) {
+  //     //freeing memory at pointer to sensor
+  //     delete sensorIterator->second.get();
+  //     //erasing sensor from SensorSuite
+  //     getSensorSuite().getSensors().erase(sensorIterator);
```
jturner65, line 678: Probably don't want the delete in here.  By erasing sensorIterator from the map, you should be removing the last reference to the sensor's smart pointer, and it should clean up after itself.

src/esp/sim/Simulator.cpp
jturner65: Otherwise, objectSensorSpec should be declared as SensorSpec::uptr

vauduong: Thanks John, that's fixed now and everything works :)

src/esp/sim/Simulator.h
```diff
@@ -716,6 +719,21 @@ class Simulator {
...
+   * @brief Initialize sensor and attach to sceneNode of a particular object
+   * with uuid objectId
+   * @param objectId    Id of the object to which a sensor will be initialized
+   * at its node
+   * @param position  vec3f of position to set spec to
+   * @param orientation vec3f or orientation to set spec to
+   * @param resolution vec2i of resolution to set spec to
+   *
+   */
+  void addSensorToObject(const int objectId,
```
bigbike, line 732: When designing this API, please take a look at the definitions of the other similar "add" functions, such as `addAgent`.

The return can be a pointer to that new sensor. So you can memorize this sensor in your demo. You do not need that global sensorSuite_ in the simulator.
The input should be object Id or a `scene::SceneNode` of that object (You can create an overload function, and use one to call the other.) + a sensorSpec (instead of the pos + orientation + resolution).

src/esp/sim/Simulator.h
```diff
@@ -924,6 +942,9 @@ class Simulator {
   SimulatorConfiguration config_;
 
   std::vector<agent::Agent::ptr> agents_;
+  // SensorSuite of all sensors available to the simulator
+  esp::sensor::SensorSuite sensorSuite_;
```
bigbike, line 946: I do not think you need this `sensorSuite_`. Maybe you can have one defined in the viewer.cpp for your demo, but not in the simulator.
@aclegg3 : How do you think? I remembered that when we discuss this project, you mentioned once that we might need such a global sensorSuite to collect all the sensors? Can you think of any usage case where such a global `sensorSuite` is a must-have?

src/tests/SimTest.cpp
```diff
@@ -627,6 +631,57 @@ void SimTest::buildingPrimAssetObjectTemplates() {
 
 }  // SimTest::buildingPrimAssetObjectTemplates
 
+void SimTest::addingCameraToObject() {
```
bigbike, line 634: Should it be addSensorToObject()?

src/utils/viewer/viewer.cpp
```diff
@@ -330,6 +335,10 @@ Key Commands:
 
   const int defaultAgentId_ = 0;
   esp::agent::Agent::ptr defaultAgent_ = nullptr;
+  std::string lastAddedObjectName_ = "";
```
bigbike, line 338: All the new code in the viewer.cpp is for your demo only. Perhaps you do not have to commit it in with this PR.
So this PR should include probably everything expect the viewer.cpp.
Then you create a new branch on top of this branch, `vauduong:decouple-sensors` (careful, not on top of the `master`) and put the modified `viewer.cpp` in that new branch (no PR is needed.)

src/esp/sim/Simulator.cpp
```diff
@@ -654,6 +656,13 @@ void Simulator::removeObject(const int objectID,
                              bool deleteVisualNode,
                              const int sceneID) {
   if (sceneHasPhysics(sceneID)) {
+    // delete sensor at ObjectNode if there is one
+    std::map<std::string, esp::sensor::Sensor::ptr>::iterator sensorIterator =
+        getSensorSuite().getSensors().find(std::to_string(objectID));
```
bigbike, line 661: why not cache the result of `getSensorSuite().getSensors()` in a local variable?

src/esp/gfx/RenderCamera.h
```diff
@@ -201,7 +201,7 @@ class RenderCamera : public MagnumCamera {
    * @brief Query the cached number of Drawables visible after frustum culling
    * for the most recent render pass.
    */
-  size_t getPreviousNumVisibileDrawables() const {
+  size_t getPreviousNumVisibleDrawables() const {
```
bigbike, line 204: Thanks for fixing this. :)

bigbike (changes requested):

vauduong:
Thanks @bigbike for the feedback! If anyone else could give me another review and feedback on things to fix that would be greatly appreciated as well!

@aclegg3 @eundersander @jturner65 @Skylion007 @mosra @erikwijmans

src/esp/sim/Simulator.h
Skylion007: Python uses its own for backwards compatability, but there is no reason if you can quickly and easily get all the sensors in the scene.

src/esp/sensor/Sensor.h
```diff
@@ -131,6 +132,13 @@ class Sensor : public Magnum::SceneGraph::AbstractFeature3D {
 class SensorSuite {
  public:
   void add(const Sensor::ptr& sensor);
+
+  /**
+   * @brief Concatenate sensorSuite's sensors to existing sensors_
+   * @param[in] sensorSuite Instance of SensorSuite class from which to copy
+   * Sensors
```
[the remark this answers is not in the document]
Skylion007: Hmm, does this follow Python's dict concatenation convention? I think in Python if you update two dicts with one another it will overwrite by default. Something to consider.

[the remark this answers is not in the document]
bigbike: @Skylion007 :
Interesting. No, we follow C++ convention (for example, in C++17, we can directly use [merge()](https://en.cppreference.com/w/cpp/container/map/merge), and it behaves as what we describe here.). If we prefer the other way used in python, we can change accordingly though.

src/esp/sim/Simulator.cpp
```diff
@@ -1084,9 +1101,14 @@ agent::Agent::ptr Simulator::addAgent(
   // attach each agent, each sensor to a scene node, set the local
   // transformation of the sensor w.r.t. the agent (done internally in the
   // constructor of Agent)
-
   auto& agentNode = agentParentNode.createChild();
   agent::Agent::ptr ag = agent::Agent::create(agentNode, agentConfig);
+  esp::sensor::SensorSuite agentSensors =
+      esp::sensor::SensorFactory::createSensors(
+          agentNode, agentConfig.sensorSpecifications);
+  ag->setSensorSuite(std::move(agentSensors));
```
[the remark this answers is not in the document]
vauduong: Yes, that works!

src/esp/sim/Simulator.h
vauduong: Thanks for the comments - I included a global `sensorSuite_` into Simulator to keep track of all the sensors. I thought about just keeping a pointers to the last added sensor, but later reasoned that it would be useful to be able to get all the sensors, not just the agent sensor and the last added sensor. 

I'm happy to change this, but right now it seems like it's not so easy to get all the sensors in the scene. 

Let me know any thoughts or if there is a better way to keep track of all the sensors! :)

bigbike: The question is that why you need to get all the sensors in the scene? 
When you add a sensor to an object, the "add" function should give a pointer to this sensor, and you keep it somewhere so that you can use it later.

I am not against to have such a sensorSuite in the viewer.cpp for example, but not here, in the simulator class.

src/utils/viewer/viewer.cpp
```diff
@@ -718,7 +727,13 @@ int Viewer::addObject(const std::string& objectAttrHandle) {
 int Viewer::addTemplateObject() {
   int numObjTemplates = objectAttrManager_->getNumFileTemplateObjects();
   if (numObjTemplates > 0) {
-    return addObject(objectAttrManager_->getRandomFileTemplateHandle());
+    std::string fileTemplateHandle =
+        objectAttrManager_->getRandomFileTemplateHandle();
+    // Get object name to display in logging text
+    std::string fileName =
+        fileTemplateHandle.substr(fileTemplateHandle.find_last_of("/") + 1);
+    lastAddedObjectName_ = fileName.substr(0, fileName.find_first_of('.'));
```
jturner65, line 735: The functionality of the two lines above can be replaced by using :
```
lastAddedObjectName_ = Cr::Utility::Directory::splitExtension(
          Cr::Utility::Directory::filename(fileTemplateHandle))
          .first;

```

vauduong: Thanks!

src/esp/sim/Simulator.cpp
```diff
@@ -1122,6 +1124,18 @@ agent::Agent::ptr Simulator::getAgent(const int agentId) {
   return agents_[agentId];
 }
 
+esp::sensor::Sensor::ptr Simulator::addSensorToObject(
+    const int objectId,
+    esp::sensor::SensorSpec::ptr& sensorSpec) {
+  esp::sensor::SensorSetup sensorSpecifications = {
+      sensorSpec};  // default SensorSpec
```
bigbike, line 1131: This is not default sensor spec any more, correct? :)

src/tests/SensorTest.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "esp/sensor/SensorFactory.h"
+
+#include "configure.h"
+
+using namespace esp::sensor;
+using namespace esp::scene;
+
+// TODO: Add tests for different Sensors
+
+TEST(SensorTest, SensorFactory) {
```
bigbike, line 19: This is a newly added sensor test, which is really great! Thank you.

Unfortunately, you are following an old test template which is using google test.
We are migrating the test to use the `Cr::TestSuite::Tester`. See the `SimTest.cpp` you have modified.
Would you change it?

src/tests/SimTest.cpp
```diff
@@ -627,6 +631,64 @@ void SimTest::buildingPrimAssetObjectTemplates() {
 
 }  // SimTest::buildingPrimAssetObjectTemplates
 
+void SimTest::addingSensorToObject() {
```
bigbike, line 634: `addSensorToObject(...)`

bigbike (commented):
The PR is improved a lot. We are almost there. :)

vauduong:
@aclegg3 @eundersander @jturner65 @Skylion007 @mosra @erikwijmans @bigbike 
Thanks everyone for the suggestions! viewer code for demo has been moved to https://github.com/vauduong/habitat-sim/tree/add-sensors-to-object-viewer-demo, and I'm looking for any final reviews for this PR before merging.

aclegg3 (approved):
Looks good. 👍

src/esp/sensor/SensorFactory.h
```diff
@@ -0,0 +1,22 @@
+#include "esp/scene/SceneNode.h"
+#include "esp/sensor/CameraSensor.h"
```
bigbike, line 2: @vauduong :

This is something I did not catch.
You do not need this header file here.

src/esp/sensor/SensorFactory.cpp
```diff
@@ -0,0 +1,20 @@
+#include "esp/sensor/SensorFactory.h"
+
+#include "esp/scene/SceneNode.h"
+#include "esp/sensor/CameraSensor.h"
+#include "esp/sensor/Sensor.h"
```
bigbike, line 5: @vauduong : This is duplicated header file.

src/esp/sensor/SensorFactory.h
vauduong: Thank you!

PR #1066, opened 2021-02-01
[sensor] Visualize observation from depth sensor

bigbike:
## Motivation and Context
As titled. In current GUI mode (viewer), we do not have a way to visualize the depth info which is not displayable. This diff is to address this problem.
- It introduces a new class called `SensorInfoVisualizer` as a helper class, which provides and manages the framebuffer, renderbuffer, shaders that are necessary in the visualization.
- It introduces a new shader class called `DepthVisualizerShader` that leverages the existing `depth.vert` and `depth.frag` to visualize the depth info.
- It would work for regular depth sensor, and also the fisheye depth sensor (TODO).
- Scalable: This framework can be easily extended to visualize the other sensor info other than the depth map.
- In viewer, press key '7' to switch between rgb sensor and the depth sensor.

Lastest results:
rgb depth:
rgb-depth-1

Fisheye depth:
fisheye-depth-1

==== **OBSOLETE** ========
Regular RGB sensor:
rgb-viz
Depth sensor in the same camera angle:
depth-viz

## Types of changes

## Checklist


mosra:
Unrelated: it's interesting how @bigbike and @eundersander have almost the same avatar, what's the probability of that happening? :smile:

src/utils/viewer/viewer.cpp
```diff
@@ -931,9 +954,6 @@ void Viewer::drawEvent() {
       Mn::GL::Renderer::disable(Mn::GL::Renderer::Feature::Blending);
     }
     sensorRenderTarget->blitRgbaToDefault();
-    // Immediately bind the main buffer back so that the "imgui" below can work
-    // properly
-    Mn::GL::defaultFramebuffer.bind();
```
mosra, line 936: Why is this removed? What's the reason?

src/utils/viewer/viewer.cpp
```diff
@@ -950,6 +970,9 @@ void Viewer::drawEvent() {
                  ImGuiWindowFlags_NoDecoration | ImGuiWindowFlags_NoBackground |
                      ImGuiWindowFlags_AlwaysAutoResize);
     ImGui::SetWindowFontScale(1.5);
+    if (flyingCameraMode_) {
+      ImGui::Text("In flying Camera MODE. Press key '3' to EXIT.");
+    }
```
mosra, line 975: This seems unrelated to the bulk of changes in this PR.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -22,5 +24,34 @@ void VisualSensor::bindRenderTarget(gfx::RenderTarget::uptr&& tgt) {
...
+          .setDepthUnprojection(*depthUnprojection())
+          .setDepthScaling(depthScaling);
+      // draw to the framebuffer
+      visualizer.draw(shader.key());
+    } break;
+
+    default:
+      // Have to had this default, otherwise the clang-tidy will be pissed off,
+      // and will not let me pass
+      break;
```
mosra, line 52: :laughing: 

what was the error it gave you? this seems silly

src/esp/gfx/SensorInfoVisualizer.cpp
```diff
@@ -0,0 +1,90 @@
...
+  if (frameBuffer_.viewport() != Mn::Range2Di{{}, {framebufferSize}}) {
+    frameBuffer_ = Mn::GL::Framebuffer{{{}, framebufferSize}};
+    colorBuffer_.setStorage(Mn::GL::RenderbufferFormat::RGBA8, framebufferSize);
+    frameBuffer_.attachRenderbuffer(Mn::GL::Framebuffer::ColorAttachment{0},
+                                    colorBuffer_);
+    // depth buffer is 24-bit integer pixel
+    depthBuffer_.setStorage(Mn::GL::RenderbufferFormat::DepthComponent24,
+                            framebufferSize);
+    frameBuffer_.attachRenderbuffer(
+        Mn::GL::Framebuffer::BufferAttachment::Depth, depthBuffer_);
```
mosra, line 37: Do you actually need the depth buffer? Isn't it only rendering a single textured triangle, no matter what kind of data gets visualized?

src/esp/gfx/SensorInfoVisualizer.cpp
```diff
@@ -0,0 +1,90 @@
...
+  CORRADE_ASSERT(
+      frameBuffer_.viewport() == Mn::GL::defaultFramebuffer.viewport(),
+      "SensorInfoVisualizer::blitRgbaToDefault(): a mismatch on viewport"
+      "between current framebuffer and the default framebuffer.", );
+
+  Mn::GL::AbstractFramebuffer::blit(
+      frameBuffer_, Mn::GL::defaultFramebuffer, frameBuffer_.viewport(),
+      Mn::GL::defaultFramebuffer.viewport(), Mn::GL::FramebufferBlit::Color,
+      Mn::GL::FramebufferBlitFilter::Nearest);
+}
```
mosra, line 69: Actually, is the extra framebuffer needed *at all*? Since the only use of this visualizer is in the viewer, where you *do want* to have it in the default framebuffer anyway.

I'd personally just have `draw()` take the target framebuffer it renders to and that's it. I don't see a reason why there would need to be yet another render target for visualization operations.

src/esp/gfx/DepthVisualizerShader.cpp
```diff
@@ -0,0 +1,84 @@
...
+  CORRADE_INTERNAL_ASSERT_OUTPUT(Mn::GL::Shader::compile({vert, frag}));
+
+  attachShaders({vert, frag});
+
+  CORRADE_INTERNAL_ASSERT_OUTPUT(link());
+
+  setUniform(uniformLocation("depthTexture"), DepthTextureUnit);
+
+  depthUnprojectionUniform_ = uniformLocation("depthUnprojection");
+  CORRADE_INTERNAL_ASSERT(depthUnprojectionUniform_ != ID_UNDEFINED);
```
mosra, line 60: I remember we discussed this once already, but I feel like it needs repeating -- `ID_UNDEFINED` is Habitat-specific, it has no relation to Magnum, and if it ever changes for some reason, the assert will no longer work.

If you want to avoid using the `-1` because "coding guidelines forbid magic constants" etc, I'd suggest doing this instead:

```cpp
CORRADE_INTERNAL_ASSERT(depthUnprojectionUniform_ >= 0);
```

Same elsewhere.

src/esp/gfx/DepthVisualizerShader.h
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#ifndef ESP_GFX_DEPTHVISUALIZERSHADER_H_
+#define ESP_GFX_DEPTHVISUALIZERSHADER_H_
+#include <Corrade/Containers/EnumSet.h>
```
mosra, line 7: Include not used here :)

src/esp/gfx/SensorInfoVisualizer.h
```diff
@@ -0,0 +1,111 @@
...
+
+enum class SensorInfoType : Magnum::UnsignedInt {
+  Depth = 0,
+};
+
+/**
+@brief helper class to visualize undisplayble info such as depth, semantic ids
+from the sensor on a framebuffer
+*/
+class SensorInfoVisualizer {
```
mosra, line 30: Why not just `SensorVisualizer`?

src/esp/sensor/VisualSensor.cpp
```diff
@@ -22,5 +24,34 @@ void VisualSensor::bindRenderTarget(gfx::RenderTarget::uptr&& tgt) {
...
+    case SensorType::Depth: {
+      // setup the shader
+      Mn::Resource<Mn::GL::AbstractShaderProgram, gfx::DepthVisualizerShader>
+          shader = visualizer.getShader<gfx::DepthVisualizerShader>(
+              gfx::SensorInfoType::Depth);
+      shader->bindDepthTexture(renderTarget().getDepthTexture())
+          .setDepthUnprojection(*depthUnprojection())
+          .setDepthScaling(depthScaling);
+      // draw to the framebuffer
+      visualizer.draw(shader.key());
```
mosra, line 46: I don't understand the API design here. Why do you first retrieve a shader resource to set its uniforms but then pass just a resource key to `draw()` so `draw()` has to fetch the same shader resource again (and then do additional error handling)? Why not have just a `draw(GL::AbstractShaderProgram&)`? Or maybe even just expose the mesh and call a draw() on it directly from here.

Honestly the more I look at it, the less useful the overly generic `SensorVisualizer` seems to be. I'd make it actually aware of what it's doing instead of splitting the responsibilities partly between this class and the `SensorVisualizer`. That way:

- it can create the shader on-demand and set the uniforms on its own, exposing a simpler API to be used from VisualSensor
- thus you don't need to do the *direct* shader setup here (this is not a good place for such low-level stuff I'd say)
- and so you don't need the complex templated `getShader()` getter and neither the resource manager nor any of that stuff, you can just have the shader(s) one by one directly in the class as the class is fully aware of what it can and cannot do

src/shaders/depth.frag
```diff
@@ -6,20 +6,35 @@ in highp vec2 textureCoordinates;
...
-  originalDepth = depth;
+    #ifdef DEPTH_VISUALIZATER
+    highp float
+    #endif
+    originalDepth = depth;
+  #endif
+  #ifdef DEPTH_VISUALIZATER
+    highp float r = originalDepth / depthScaling;
+    highp float b = 0.5 - 0.5 * r;
+    fragmentColor = vec4(r, r, b, 1.0);
```
mosra, line 38: I'd suggest to use colormaps for better visualization output -- [DebugTools::ColorMap](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1DebugTools_1_1ColorMap.html). Especially the Turbo color map is tuned to make minor differences nicely visible and the output uniformly bright.

Further, to make it more generic, I'd name it a "color map transform" instead of "depth scaling", as that allows you to later use the same API for semantic ID visualization as well. [See how MeshVisualizer does it](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1MeshVisualizer3D.html#Shaders-MeshVisualizer-object-id).

src/shaders/depth.frag
```diff
@@ -6,20 +6,35 @@ in highp vec2 textureCoordinates;
...
+      /* We can afford using == for comparison as 1.0f has an exact
+         representation and the depth is cleared to exactly this value. */
+      depth == 1.0 ? 0.0 :
+      #endif
+      depthUnprojection[1] / (depth + depthUnprojection[0]);
   #else
-  originalDepth = depth;
+    #ifdef DEPTH_VISUALIZATER
+    highp float
+    #endif
```
mosra, line 32: I feel these `#ifdef`s are a bit excessive. Declare the variable once above for both cases so you don't need to do this.

Thinking more about this, I feel like two *completely unrelated* shaders are mashed up together. They're not even exposed through the same C++ class. Since this is for an interactive visualization and not ML, one extra pass won't matter, so what about

1. unprojecting depth as usual using the original shader
2. then have a shader that's dedicated *just* to visualization, which then processes the unprojected depth

That way the `VisualizerShader` doesn't need the depth-related inputs and can be also easily adapted to visualizing any other rendered data, such as semantic IDs.

src/esp/sensor/VisualSensor.cpp
Skylion007: Agreed, this should be more tightly integrated with Sensor. What's the benefit of this class? This class seems like it would be a pain to use easily from Python for instance.

bigbike: To both of you: @mosra @Skylion007 
**Why I need this class?**
I need a container to hold the framebuffer, renderbuffer, shaders, the big triangle (GL mesh) as a canvas etc. These resources which are necessary to do the visualization. I tried to put them in the VisualSensor. Then VisualSensor becomes too much and too ugly. Also I hope this infra can be used in the `NonVisualSensor` in the future. 

**Why I need the framebuffer and renderbuffer?**
If it is only for "visualization", yes, the default framebuffer is good enough. But if I need "interaction", it is not enough. For example, I am thinking it would be awesome if we can "select and highlight" some objects even in the depth or semantic visualization.
(I know I need to further change the shader in this case.)

Now answer the questions separately:

@mosra:
>it can create the shader on-demand 

Constructing a shader is expensive. I prefer to keep it once it is constructed. So I need the ResourceManager either in this class or in the `VisualSensor`. I prefer to do it in this `SensorVisualizer` class so that later after we have developed the `NonVisualSensor` I can use it as well.

>I don't understand the API design here. 
>maybe even just expose the mesh 

Yes, when I coded it up, I wished to call `shader->draw()` directly. But I found I needed that mesh which was NOT exposed (and I do not want to). I was thinking retrieve this shader via key was cheap to do. 

>draw(GL::AbstractShaderProgram&)

OK, I can do it.

>you can just have the shader(s) one by one directly in the class as the class is fully aware of what it can and cannot do

Putting all the shaders (there might be a lot in the future) one by one in which class, the VisualSensor class? It would become very ugly. I guess I do not understand you completely, do I? Hope you can explain a bit further.  

>the complex templated getShader()

Not just for depth, semantic info, I am paving the way for the future, e.g., force, lidar, etc. Any info that is not directly displayable, but has certain ways to be visualized. 

@Skylion007 : 
> What's the benefit of this class?

See my above comments on why I need this class.

>simplify shader setup so that it can be easily reused from VisualSensor.

I believe this is a clean way in order NOT to mess up with the `VisualSensor`.

>doing it like that would make it easier to expose to Python for example

You will have to expose the entire `SensorVisualizer` class simply because it appears as 1 input parameter in the argument list of this function `visualizeObservation` in `VisualSensor`?

src/esp/sensor/VisualSensor.cpp
bigbike: The CI test failed directly in the clang-tidy section.
It says missing code for the other enum terms. 
And look at this [long list](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/sensor/Sensor.h#L20-L32):   
  None = 0,
  Color = 1,
  Depth = 2,
  Normal = 3,
  Semantic = 4,
  Path = 5,
...

src/shaders/depth.frag
bigbike: >  a shader that's dedicated just to visualization

I used to have a dedicated shader. But it is different from what you suggested here -- dedicated to process the **unprojected** depth. My shader will still do both depth unprojection + processing like it does here. And then I thought you might say, well why not leverage the existing depth.frag to avoid code duplication?? Clearly my guess took the wrong direction this time. :)

Well, I guess the focus is that do I need the depth-related input?
Answer: it depends whether I need "visualization ONLY" or "visualization + interaction".
Your comments is based on the former, while I was thinking the latter.
However, I am actually considering adding depth in later (directly read from the `depthRenderTexture` and output to the `gl_FragDepth`) since it gives me the ability for "interaction".

src/shaders/depth.frag
bigbike: Cool stuff. I will study it.

src/utils/viewer/viewer.cpp
bigbike: Good catch. I am supposed to move it out of the bracket. Definitely a bug introduced. Will fix.

src/utils/viewer/viewer.cpp
bigbike: Yeah, I found it was annoying (without such info) when debugging this PR. So try to fix it in this PR.

src/esp/gfx/SensorInfoVisualizer.cpp
bigbike: I need a depth buffer later if I would like to "interact“. For example, if I would like to select and highlight some objects. ( I need to draw these objects on top of this framebuffer.)

bigbike: I need a depth buffer later if I would like to "interact“. For example, if I would like to select and highlight some objects. ( I need to draw these objects on top of this framebuffer.)

src/esp/gfx/SensorInfoVisualizer.cpp
bigbike: See my comments below.

src/esp/sensor/VisualSensor.cpp
bigbike: @Skylion007 : the "default" a bad concept.

src/shaders/depth.frag
bigbike: By the way, I understand you can apply such ColorMap in magnum built-in MeshVisusalizer shader. 
Question: how can I apply it to my customized shader??

src/esp/sensor/VisualSensor.cpp
mosra: Oh, right, I didn't realize there's a lot more values. In that case either enumerating them all or adding a `default`. In this case the default probably makes sense.

mosra:
I feel like I already spent more time on the review here than it would take to implement a `TextureVisualizer` shader directly in magnum, so just in short -- this is how I think it should work:

1. Render everything the usual way into a framebuffer X, with depth unprojected to a texture from a depth buffer. Or render just the depth, because the color will get discarded anyway.
2. Draw a full-screen quad *into the same* framebuffer X, visualizing the depth using a color map, only color, with depth test disabled.

That way you:

- end up with a framebuffer that has the depth visualized *but also* the depth correct, which satisfies your point about "interaction",
- don't need to do anything special with `gl_FragDepth`,
- don't need a whole new Visualizer class with yet another color and depth framebuffer and all the scaffolding and documentation and testing around, as this is reusing what's already there,
- don't need to conflate the depth unprojection shader with visualization (*seriously*, from a design perspective, it just doesn't make sense to conflate the two, it will only cause problems down the road),
- and everything is much more modular and easier to extend -- if, let's say, a semantic ID visualization needs to be done (and it can't be done with [MeshVisualizer]([MeshVisualizer3D](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1MeshVisualizer3D.html)) for some strange reason), it means rendering just another fullscreen triangle with another custom shader to the same color buffer, without the shader again having to take care of depth unprojection or writing to `gl_FragDepth`

bigbike:
Dear reviewers:
The diff is ready to be reviewed again. Code has been significantly refactored based on the reviewers' comments.
In this version the `Magnum::ColorMap` is adopted and the parameter is fine tuned.
See the results (depth sensor) as follows:
rgb-depth-1
fisheye-depth-1

bigbike:
@mosra: This is diff is ready to be reviewed again.

src/esp/gfx/TextureVisualizerShader.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+TextureVisualizerShader& TextureVisualizerShader::setColorMapTransformation(
+    float offset,
+    float scale) {
+  CORRADE_ASSERT(offset >= 0.0,
+                 "TextureVisualizerShader::setColorMapTransformation(): offset"
+                     << offset << "is illegal.",
+                 *this);
+  CORRADE_ASSERT(scale >= 0.0,
+                 "TextureVisualizerShader::setColorMapTransformation(): scale "
```
mosra, line 115:
```suggestion
                 "TextureVisualizerShader::setColorMapTransformation(): scale"
```

src/shaders/textureVisualizer.frag
```diff
@@ -0,0 +1,53 @@
...
+#define colorMapScale colorMapOffsetScale.y
+
+#ifdef DEPTH_TEXTURE
+uniform highp vec2 depthUnprojection;
+#endif
+
+//------------- output ----------------------
+#ifdef EXPLICIT_ATTRIB_LOCATION
+layout(location = OUTPUT_ATTRIBUTE_LOCATION_COLOR)
+#endif
```
mosra, line 24: I think this could do without the `#ifdef EXPLICIT_ATTRIB_LOCATION` around -- in Magnum that's only for compatibility with GL 2.1 / ES2 / WebGL 1, which you don't care about here.

(Also remove the related `addSource()` in the shader cpp, of course.)

mosra (commented):
Found just two very minor things.

I was wondering if it would be useful to allow the user to change the colormap used, but maybe do that only when someone *really needs* that. Turbo is a good default.

mosra (approved):

PR #1071, opened 2021-02-02
[BUGFIX] : Handle class template with multiple params in Smart Ptr Macro

jturner65:
## Motivation and Context
Our current smart pointer macro breaks if a class template with more than one type parameter is passed to it.  This is because the C Preprocessor sees the type definition, including the list of params, and breaks it into multiple types.

(so `MyType<T,U>` is seen as `MyType<T` and `U>`)

The changes introduced in this small PR fix this issue.  Thanks @mosra!

## How Has This Been Tested
C++ and python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/esp.h
+++ b/src/esp/core/esp.h
@@ -107,19 +107,19 @@ inline std::ostream& operator<<(std::ostream& os, const box3f& bbox) {
 }
 
 // smart pointers macro
-#define ESP_SMART_POINTERS(T)                                 \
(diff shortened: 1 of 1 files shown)


mosra (approved):
To be frank, I have issues with the `ESP_SMART_POINTERS()` macro as a whole.

But the fix is fine :)

Skylion007 (approved):

PR #1072, opened 2021-02-02
--Add managed container template argument to govern object access.

jturner65:
## Motivation and Context
This slight refactor adds a nontype template parameter to govern whether the ManagedContainers will provide copies of managed objects upon get, or the object itself.  Certain potential managed objects do not lend themselves to the current process where a copy is provided, and this change enables the actual object that is tracked to be provided to the user.

## How Has This Been Tested
C++ and python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/ManagedContainer.h
+++ b/src/esp/core/ManagedContainer.h
@@ -16,18 +16,18 @@ namespace esp {
 namespace core {
 
 /**
- * @brief This enum describes how objects held in the ManagedConatainers are
+ * @brief This enum describes how objects held in the @ref ManagedConatainer are
  * accessed.
  */
 enum class ManagedContainerAccess {
   /**
-   * @brief When an object is requested to be retrieved, a copy is made and
+   * When an object is requested to be retrieved, a copy is made and
    * returned.  Modifications to this object will only take place if the object
    * is registered.
    */
   Copy,
   /**
-   * @brief When an object is requested to be retrieved, a reference to the
+   * When an object is requested to be retrieved, a reference to the
    * object itself is returned, and all changes will be tracked without
    * registration.
    */
--- a/src/esp/core/ManagedContainer.h
+++ b/src/esp/core/ManagedContainer.h
@@ -14,17 +14,37 @@
 
 namespace esp {
 namespace core {
+
+/**
+ * @brief This enum describes how objects held in the ManagedConatainers are
+ * accessed.
+ */
+enum class ManagedContainerAccess {
+  /**
+   * @brief When an object is requested to be retrieved, a copy is made and
+   * returned.  Modifications to this object will only take place if the object
+   * is registered.
+   */
+  Copy,
+  /**
+   * @brief When an object is requested to be retrieved, a reference to the
+   * object itself is returned, and all changes will be tracked without
+   * registration.
+   */
+  Share
+};
+
 /**
  * @brief Class template defining responsibilities and functionality for
  * managing @ref esp::core::AbstractManagedObject constructs.
  * @tparam T the type of managed object a particular specialization of
  * this class works with.  Must inherit from @ref
  * esp::core::AbstractManagedObject.
- * @tparam AccessViaCopies Whether the default access (getters) for this
+ * @tparam Access Whether the default access (getters) for this
  * container provides copies of the objects held, or the actual objects
  * themselves.
  */
-template <class T, bool AccessViaCopies>
+template <class T, ManagedContainerAccess Access>
 class ManagedContainer : public ManagedContainerBase {
  public:
   static_assert(std::is_base_of<AbstractManagedObject, T>::value,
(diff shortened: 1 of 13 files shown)


src/esp/core/ManagedContainer.h
```diff
@@ -20,8 +20,11 @@ namespace core {
  * @tparam T the type of managed object a particular specialization of
  * this class works with.  Must inherit from @ref
  * esp::core::AbstractManagedObject.
+ * @tparam AccessViaCopies Whether the default access (getters) for this
+ * container provides copies of the objects held, or the actual objects
+ * themselves.
  */
-template <class T>
+template <class T, bool AccessViaCopies>
```
jturner65, line 27: The reason why I implemented this change via a non-type template argument instead of a variable passed via the constructor is two fold :

1. By making it a template argument, it changes the actual type.  A ManagedContainer<ObjectAttributes, true> is not the same type as a ManagedContainer<ObjectAttributes, false> and so they cannot inherit or otherwise interfere with each other.
2. The impact of the change was minimal (much less code changed than if it was a constructor argument).

Added Bonus : it exposed a [bug in our ESP_SMART_POINTERS macro](https://github.com/facebookresearch/habitat-sim/pull/1071).  🎉

Skylion007 (changes requested):

src/esp/core/ManagedContainer.h
Skylion007: Ah, just wanted to make sure we weren't excessively copying.

src/esp/core/ManagedContainer.h
```diff
@@ -554,7 +565,7 @@ class ManagedContainer : public ManagedContainerBase {
     if (defaultObj_ == nullptr) {
       return nullptr;
     }
-    ManagedPtr res = this->copyObject(defaultObj_);
+    ManagedPtr res = copyObject(defaultObj_);
```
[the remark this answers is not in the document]
jturner65: this-> is required because of class template scoping.  This is also required for internal/c++ access.

Skylion007 (approved):
Looks good to me, but I'd like someone with more C++ experience look at.

jturner65:
> Looks good to me, but I'd like someone with more C++ experience look at.

Yep, I sent up the @mosra signal :bat:

src/esp/core/ManagedContainer.h
mosra, line 27: I don't like the `bool` parameter, it's not very descriptive. What about this?

```cpp
enum class ManagedContainerAccess {
   Copy,
   Share
};

template<class T, ManagedContainerAccess access> class ManagedContainer;
```

and then possibly shorthands

```cpp
template<class T> using ManagedContainerWithCopyAccess = ManagedContainer<T, ManagedContainerAccess::Copy>;
template<class T> using ManagedContainerWithSharedAccess = ManagedContainer<T, ManagedContainerAccess::Copy>;
```

umm, though these are not really shorter and only add further congitive overhead. So maybe without?

jturner65: Cool, I'll modify to consume an enum, very nice ty.

src/esp/core/ManagedContainer.h
```diff
@@ -336,7 +339,11 @@ class ManagedContainer : public ManagedContainerBase {
       return nullptr;
     }
     auto orig = getObjectInternal<T>(objectHandle);
-    return this->copyObject(orig);
+    if (AccessViaCopies) {
```
mosra, line 342: `if constexpr`? (does habitat use c++17 or not?)

Same below.

jturner65: We're not c++17 yet, unfortunately.

src/esp/core/ManagedContainer.h
```diff
@@ -14,14 +14,37 @@
...
 namespace esp {
 namespace core {
+
+/**
+ * @brief This enum describes how objects held in the ManagedConatainers are
+ * accessed.
+ */
+enum class ManagedContainerAccess {
+  /**
+   * @brief When an object is requested to be retrieved, a copy is made and
```
mosra, line 24: FYI a `@brief` for enum values won't be displayed in the docs because &lt;complicated reasons&gt; -- do it without:

```suggestion
   * When an object is requested to be retrieved, a copy is made and
```

Same below.

src/esp/core/ManagedContainer.h
```diff
@@ -14,14 +14,37 @@
 
 namespace esp {
 namespace core {
+
+/**
+ * @brief This enum describes how objects held in the ManagedConatainers are
```
mosra, line 19:
```suggestion
 * @brief This enum describes how objects held in the @ref ManagedConatainer are
```

(personally I'm always making sure that the enum links to the class that uses it so the relation is clear)

PR #1088, opened 2021-02-11
--Semantic Scene Descriptor generalized loading

jturner65:
## Motivation and Context
This PR generalizes the loading mechanism of the Semantic Scene Descriptor files, with the intention that the data, and not the file name, should be used to determine the file type.  

This change is being proposed now mainly because the SceneDataset/Scene Import tutorials and tests are going to describe how to properly configure SceneDataset and SceneAttributes configs to support, among other things, SSD file loading. 

Presented is a working proposal for loading SSD files based on contents (although I don't think it is sufficiently exclusionary to be merged).

In the existing system, the SSD text file loading process is brittle in that it expects users to obey certain file extension or directory hierarchy constraints to successfully load the appropriate SSD files for a particular dataset.   For example, Gibson SSD files are expected to have ".scn" extension (which they are assigned when the gen_gibson_semantics script is executed on the base .npz files), but these files are just json, and the only expectations we have for a valid Gibson SSD file is that it be json and have an "objects" tag, holding object descriptions with an expected format.

The idea behind this PR is to relax some of these arbitrary restrictions placed on SSD file naming/location, and use the file contents to determine what type of file is present.  With the introduction of SceneDataset and SceneAttributes configs, one design goal was to provide the user with as much flexibility as possible for their dataset configuration processes, and the current mechanism for SSD loading runs counter to that goal.

In an effort to make this process more rigorous I am also planning a preprocessing script the user can use that will add a "scene_dataset: <name>" tag to the various SSD files, but I first wanted to see if there was sufficient interest in the proposed change (content-aware SSD loading) to warrant such a change. 

## How Has This Been Tested
C++ and Python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/scene/Mp3dSemanticScene.cpp
+++ b/src/esp/scene/Mp3dSemanticScene.cpp
@@ -91,6 +91,23 @@ bool SemanticScene::loadMp3dHouse(
     return false;
   }
 
+  // open stream and determine house format version
+  std::ifstream ifs = std::ifstream(houseFilename);
+  std::string header;
+  std::getline(ifs, header);
+  if (header != "ASCII 1.1") {
+    LOG(ERROR) << "SemanticScene::loadMp3dHouse : Unsupported Mp3d House "
+                  "format header "
+               << header << " in file name " << houseFilename;
+    return false;
+  }
+
+  return buildMp3dHouse(ifs, scene, rotation);
+}  // SemanticScene::loadMp3dHouse
+
+bool SemanticScene::buildMp3dHouse(std::ifstream& ifs,
+                                   SemanticScene& scene,
+                                   const quatf& rotation) {
   const bool hasWorldRotation = !rotation.isApprox(quatf::Identity());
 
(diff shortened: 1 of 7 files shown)


src/esp/scene/Mp3dSemanticScene.cpp
```diff
@@ -143,7 +142,9 @@ bool SemanticScene::loadMp3dHouse(
   std::string header;
   std::getline(ifs, header);
   if (header != "ASCII 1.1") {
-    LOG(ERROR) << "Unsupported House format header " << header;
+    LOG(ERROR) << "SemanticScene::loadMp3dHouse : Unsupported Mp3d House "
```
erikwijmans, line 145: Will this error get printed if I try to load replica or gibson semantics?  It looks like it will.  We certainly shouldn't log something as an error when all is actually fine.

jturner65: With these updated changes, this error will only trigger if this function is called directly with a filename that is assumed to reference an Mp3d file.

Skylion007 (approved):

PR #1090, opened 2021-02-12
[Sensor] Hierarchical Sensor spec

vauduong:
## Motivation and Context
Currently we only have SensorSpec structure in the simulator. (in Sensor.h). This task is to build a hierarchical sensor spec abstractions to match the new sensor framework. For example, we have “SensorSpec” (base class) containing uuid, sensor type, position and orientation etc., then deriving “VisualSensorSpec”, “NonVisualSensorSpec”, and then “CameraSensorSpec”;

Since we do not have any non visual sensor yet. This task will be focusing on the visual sensor and its subclasses (the left side in the tree structure).

image

The following PR:

- Creates structs SensorSpec, VisualSensorSpec, and CameraSensorSpec with relevant sanityChecks to initialize the default parameters for Sensors
- Adds functionality for SensorSpecs to be initialized (Currently only CameraSensorSpec can be initialized) with `CameraSensorSpec::create()`. After initialization, default parameters can be modified. When a vector of sensorSpecs is passed into `SensorFactory::createSensors()`, SensorFactory will perform checks and automatically initialize the correct sensor for each SensorSpec and return it in a SensorSuite 
- Adds `SensorSubType::None`
- Improved == and != operators for SensorSpecs
- Deprecates height and width in CameraSensor, deprecates parameters in SensorSpec
- `SensorSpec::create()` is no longer used, use correct SensorSpec initialization to create correct sensor, such as `CameraSensorSpec::create()`
- Fixes python code to use `CameraSensorSpec()` to initialize sensorSpecs, adds relevant checks for sensor initialization
- Breaks hab-lab examples, this will be fixed in https://github.com/facebookresearch/habitat-lab/pull/598
## How Has This Been Tested

Build and modify tests

## Types of changes

## Checklist


src/esp/agent/Agent.h
```diff
@@ -60,7 +61,7 @@ struct AgentConfiguration {
   float coefficientOfRestitution = 0.0;
 
   std::vector<sensor::SensorSpec::ptr> sensorSpecifications = {
-      sensor::SensorSpec::create()  // default SensorSpec
+      sensor::CameraSensorSpec::create()  // default CameraSensorSpec
```
bigbike, line 64: @Skylion007 : By default, there is a "default" sensor spec in the list of sensor specifications.
Are we using it directly in the python?

I do not think there is any reason to have such a "default" sensor spec taking up the entry [0] in the vector. This list should be empty at the very beginning. User push sensor specifications to this list from scratch.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -19,6 +19,17 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
     : VisualSensor(cameraNode, spec),
       baseProjMatrix_(Magnum::Math::IdentityInit),
       zoomMatrix_(Magnum::Math::IdentityInit) {
+  // Sanity check
+  CORRADE_ASSERT(std::static_pointer_cast<VisualSensorSpec>(spec) != nullptr,
+                 "Spec given is of type VisualSensor", );
+  CORRADE_ASSERT(spec->sensorType == SensorType::Color,
```
bigbike, line 25: `CameraSensor` can have types such as "depth, semantic".

src/esp/sensor/CameraSensor.h
```diff
@@ -12,6 +12,19 @@
 namespace esp {
 namespace sensor {
 
+struct CameraSensorSpec : public VisualSensorSpec {
+  CameraSensorSpec() : VisualSensorSpec() {
```
bigbike, line 16: would you move the ctor implementation to the cpp?

src/esp/sensor/CameraSensor.h
```diff
@@ -12,6 +12,19 @@
...
+    uuid = "rgba_camera";
+    sensorSubType = SensorSubType::Pinhole;
+    parameters = {{"near", "0.01"},
```
bigbike, line 19: Like I mentioned in the 1:1 with you,
- Would you move the "near" "far" "hfov" etc. out of the parameters?
- Would you put them in the VisualSensor level temporarily, making them member variables? (I suspect we also need it in the fisheye sensor.)

src/esp/sensor/Sensor.h
```diff
@@ -44,17 +44,14 @@ enum class SensorSubType {
...
-  vec3f position = {0, 1.5, 0};
-  vec3f orientation = {0, 0, 0};
-  vec2i resolution = {84, 84};
-  int channels = 4;
+  std::string uuid;
+  SensorType sensorType;
+  SensorSubType sensorSubType;
+  vec3f position;
+  vec3f orientation;
+  vec2i resolution;
```
bigbike, line 52: this should go to visual sensor for now.
If non-visual sensor also need it, we can always lift it up later.

src/esp/sensor/Sensor.h
```diff
@@ -44,17 +44,14 @@ enum class SensorSubType {
...
+  std::map<std::string, std::string> parameters;
+  int channels;
```
bigbike, line 54: make the channels to the visual sensor.

src/esp/sensor/VisualSensor.h
```diff
@@ -18,7 +18,14 @@ class RenderTarget;
 }
 
 namespace sensor {
-
+struct VisualSensorSpec : public SensorSpec {
+  VisualSensorSpec() : SensorSpec() {
```
bigbike, line 22: Would you put the implementation to the cpp?
Also recommend using member initializer lists when giving the initial values.

src/esp/sensor/Sensor.h
vauduong: Ok, I can do that! My reasoning was that position, orientation, and resolution would matter for force and acoustic sensors as well.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -19,6 +19,17 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
     : VisualSensor(cameraNode, spec),
       baseProjMatrix_(Magnum::Math::IdentityInit),
       zoomMatrix_(Magnum::Math::IdentityInit) {
+  // Sanity check
```
bigbike, line 22: Sanity check should be done on each level, not just the CameraSensor. For example, it should be done in the VisualSensor as well.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -19,6 +19,17 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
...
+  CORRADE_ASSERT(std::static_pointer_cast<VisualSensorSpec>(spec) != nullptr,
+                 "Spec given is of type VisualSensor", );
+  CORRADE_ASSERT(spec->sensorType == SensorType::Color,
+                 "Spec given has SensorType Color", );
+  CORRADE_ASSERT(std::static_pointer_cast<CameraSensorSpec>(spec) != nullptr,
+                 "Spec given is of type CameraSensor", );
+  CORRADE_ASSERT(
+      spec->sensorSubType == SensorSubType::Pinhole ||
+          spec->sensorSubType == SensorSubType::Orthographic,
+      "CameraSensorSpec given has SensorSubType Pinhole or Orthographic", );
```
bigbike, line 32: The error msg in the CORRADE_ASSERT always starts with the location, e.g., `CameraSensor::CameraSensor(): `
You need to modify your error msg so that it is informative when being displayed to the user (when an error occurs).
For example here:
`CameraSensor::CameraSensor(): Sensor sub-type" <<spec->sensorSubType <<"is not supported."`

src/esp/sensor/CameraSensor.cpp
```diff
@@ -19,6 +19,17 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
     : VisualSensor(cameraNode, spec),
       baseProjMatrix_(Magnum::Math::IdentityInit),
       zoomMatrix_(Magnum::Math::IdentityInit) {
+  // Sanity check
+  CORRADE_ASSERT(std::static_pointer_cast<VisualSensorSpec>(spec) != nullptr,
```
bigbike, line 23: Are you should static_pointer_cast would work? Usually such static_cast is used when you basically reverse an implicit conversion and it does NOT perform runtime checks. This is used when you are pretty sure you are going to refer to an object of a specific type, and thus a check is not necessary.

Please google static_pointer_cast, and dynamic_pointer_cast.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -19,6 +19,17 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
...
+                 "Spec given is of type VisualSensor", );
+  CORRADE_ASSERT(spec->sensorType == SensorType::Color,
+                 "Spec given has SensorType Color", );
+  CORRADE_ASSERT(std::static_pointer_cast<CameraSensorSpec>(spec) != nullptr,
```
bigbike, line 27: the `!= nullptr` is redundant.

src/esp/sensor/SensorFactory.cpp
```diff
@@ -12,7 +12,16 @@ sensor::SensorSuite SensorFactory::createSensors(
   sensor::SensorSuite sensorSuite = sensor::SensorSuite();
   for (const sensor::SensorSpec::ptr& spec : sensorSetup) {
     scene::SceneNode& sensorNode = node.createChild();
-    sensorSuite.add(sensor::CameraSensor::create(sensorNode, spec));
+    if (spec->sensorType == SensorType::Color) {
+      if (spec->sensorSubType == SensorSubType::Pinhole ||
+          spec->sensorSubType == SensorSubType::Orthographic) {
+        sensorSuite.add(sensor::CameraSensor::create(sensorNode, spec));
```
Skylion007, line 18: We need appropiate errors here.

src/esp/agent/Agent.h
Skylion007: Nah, we don't use this interface I don't think.

src/esp/sensor/Sensor.h
bigbike: My comment was on the resolution, not on position or orientation.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -22,5 +26,31 @@ void VisualSensor::bindRenderTarget(gfx::RenderTarget::uptr&& tgt) {
...
+
+  node().resetTransformation();
+
+  node().translate(Magnum::Vector3(spec_->position));
+  node().rotateX(Magnum::Rad(spec_->orientation[0]));
+  node().rotateY(Magnum::Rad(spec_->orientation[1]));
+  node().rotateZ(Magnum::Rad(spec_->orientation[2]));
+}
+
+void VisualSensor::updateResolution(float x_res, float y_res) {
```
bigbike, line 43: resolution is `vec2i`. `i` means integer. 

Also we do not use _ for the variable by our C++ convention, we use Camel style. Just look around the other functions.

`width` and `height` is better than `xRes` and `yRes` 

CAREFUL!! in computer graphics, we usually w x h, w always comes before the h. HOWEVER here we follow the CV convention, namly h comes first, h x w.

A good design is:
`void VisualSensor::setResolution(int height, int width);`

Also provide an overload function with argument vec2i directly.

src/esp/sensor/Sensor.cpp
```diff
@@ -11,29 +11,19 @@
 namespace esp {
 namespace sensor {
 
+SensorSpec::~SensorSpec() {}
+
 Sensor::Sensor(scene::SceneNode& node, SensorSpec::ptr spec)
-    : Magnum::SceneGraph::AbstractFeature3D{node}, spec_(std::move(spec)) {
+    : Magnum::SceneGraph::AbstractFeature3D{node}, spec_(spec) {
```
bigbike, line 17: What is the reason that you removed the `std::move`?

src/esp/sensor/Sensor.h
```diff
@@ -44,22 +44,15 @@ enum class SensorSubType {
...
-                                                   {"hfov", "90"},
-                                                   {"ortho_scale", ".1"}};
-  vec3f position = {0, 1.5, 0};
-  vec3f orientation = {0, 0, 0};
-  vec2i resolution = {84, 84};
-  int channels = 4;
+  std::string uuid;
+  SensorType sensorType;
+  SensorSubType sensorSubType;
   std::string encoding = "rgba_uint8";
```
bigbike, line 50: there are more stuffs needed to go.

src/esp/sensor/CameraSensor.cpp
vauduong: Thanks, dynamic_pointer_cast seems like the way to go when checking types

src/esp/sensor/Sensor.cpp
vauduong: I noticed that at runtime, keeping `std::move(spec)` would cause VisualSensor's spec to be nullptr, while a separate spec existed in memory. I removed it and it seemed to work. Could it be due to how in `VisualSensor.h`, I have `VisualSensorSpec::ptr spec_ = std::dynamic_pointer_cast<VisualSensorSpec>(spec_);` to make sure that methods can access parameters in VisualSensorSpec but not SensorSpec?

vauduong:
Thanks a lot to everyone for the feedback, it's helping me understand the task more!
TODO: More sanitycheck, fix python bindings and tests

src/esp/sensor/CameraSensor.h
```diff
@@ -245,6 +257,9 @@ class CameraSensor : public VisualSensor {
    */
   gfx::RenderCamera* renderCamera_;
 
+  CameraSensorSpec::ptr spec_ =
```
vauduong, line 260: Is this a valid way to make sure that CameraSensor has access to parameters of CameraSensorSpec, not just SensorSpec?

src/esp/sensor/Sensor.h
```diff
@@ -44,22 +44,12 @@ enum class SensorSubType {
 
 // Specifies the configuration parameters of a sensor
 struct SensorSpec {
```
vauduong, line 46: Any feedback on if this SensorSpec looks as expected would be great, or if there is anything else to move.

src/esp/sensor/Sensor.h
```diff
@@ -44,22 +44,12 @@ enum class SensorSubType {
...
+  SensorSpec() {}
+  std::string uuid;
   vec3f position = {0, 1.5, 0};
   vec3f orientation = {0, 0, 0};
-  vec2i resolution = {84, 84};
-  int channels = 4;
-  std::string encoding = "rgba_uint8";
-  // description of Sensor observation space as gym.spaces.Dict()
-  std::string observationSpace = "";
   std::string noiseModel = "None";
```
vauduong, line 51: I kept noiseModel in SensorSpec because the python bindings had many methods which referred to the noiseModel. From is_valid_sensor_type, it seems like only VisualSensors have noiseModels which are not "None" (but can be). NonVisualSensors can only have noiseModel = "None", so I may move this to VisualSensor. Will need to look more into python bindings @Skylion007 let me know if you have any thoughts!

src/esp/sensor/SensorFactory.cpp
```diff
@@ -12,7 +12,21 @@ sensor::SensorSuite SensorFactory::createSensors(
   sensor::SensorSuite sensorSuite = sensor::SensorSuite();
   for (const sensor::SensorSpec::ptr& spec : sensorSetup) {
     scene::SceneNode& sensorNode = node.createChild();
-    sensorSuite.add(sensor::CameraSensor::create(sensorNode, spec));
+    // VisualSensor Setup
```
vauduong, line 15: Is this a good way to check and initialize sensors? I moved sensorType and sensorSubType out of SensorSpec, so now casting is the only way to tell the type. Would it be better to move sensorType and sensorSubType back in since all sensors will define this?

src/esp/sim/Simulator.cpp
```diff
@@ -1126,7 +1126,7 @@ agent::Agent::ptr Simulator::getAgent(const int agentId) {
 
 esp::sensor::Sensor::ptr Simulator::addSensorToObject(
     const int objectId,
-    esp::sensor::SensorSpec::ptr& sensorSpec) {
+    esp::sensor::SensorSpec::ptr sensorSpec) {
```
vauduong, line 1129: Changed because there was trouble passing in derived classes.
Could not get type esp::sensor::SensorSpec::ptr& from esp::sensor::CameraSensorSpec::ptr

vauduong: However, this bugs clang-tidy because SensorSpec is not modified in addSensorToObject(). I did some googling but was not sure how to fix this

src/esp/sensor/VisualSensor.cpp
```diff
@@ -22,5 +28,21 @@ void VisualSensor::bindRenderTarget(gfx::RenderTarget::uptr&& tgt) {
...
+void VisualSensor::updateResolution(int height, int width) {
+  spec_->resolution = {height, width};
+}
+
+void VisualSensor::updateResolution(vec2i resolution) {
+  spec_->resolution = resolution;
+}
+
+bool operator==(const VisualSensorSpec& a, const VisualSensorSpec& b) {
+  return a.uuid == b.uuid && a.sensorType == b.sensorType && a.near == b.near &&
```
bigbike, line 40: you have the operator == function in base class as well. why not use it to check the variables defined the base class?

src/esp/sim/Simulator.cpp
bigbike: I did not catch this in the previous PR.

`::ptr` is already a pointer so you do not have to pass the value by reference.

Second, which is more important:
The API design in the `Simulator::addSensorToObject` is really needed to be improved.
- you do not need `const` for an atomic int.
- the operation within this function has nothing to do with the ownership. so you do not need to pass a smart pointer. 

Suggested change:
`esp::sensor::Sensor::ptr Simulator::addSensorToObject(int objectId, const esp::sensor::SensorSpec& sensorSPec)`

Please remember to fix it in another PR.

src/tests/SimTest.cpp
```diff
@@ -160,9 +160,7 @@ void SimTest::reset() {
   Simulator simulator(cfg);
   PathFinder::ptr pathfinder = simulator.getPathFinder();
 
-  auto pinholeCameraSpec = SensorSpec::create();
-  pinholeCameraSpec->sensorSubType = esp::sensor::SensorSubType::Pinhole;
-  pinholeCameraSpec->sensorType = SensorType::Color;
```
bigbike, line 165: I understand you have default values. 
But still keep them here because it increases the readability. 
The same everywhere.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -22,5 +28,21 @@ void VisualSensor::bindRenderTarget(gfx::RenderTarget::uptr&& tgt) {
   tgt_ = std::move(tgt);
 }
 
+void VisualSensor::updateResolution(int height, int width) {
```
bigbike, line 31: For the setter and getter functions, we always use setSomething(...) or getSomething(...) as the function name.
Would you change it to `setResolution` 
More importantly, would you please pay attention to the reviewer's comments? I already mentioned it in the previous round of review.

bigbike (changes requested):

src/esp/sensor/VisualSensor.h
```diff
@@ -81,13 +127,19 @@ class VisualSensor : public Sensor {
...
+  void updateResolution(vec2i resolution);
+
   /**
    * @brief Returns RenderCamera
    */
   virtual gfx::RenderCamera* getRenderCamera() = 0;
 
  protected:
   std::unique_ptr<gfx::RenderTarget> tgt_;
+  VisualSensorSpec::ptr spec_ =
```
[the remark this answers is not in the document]
bigbike: Well, I should say this:
you need a `VisualSensorSpec::ptr` for convenience. but do not call it spec_ as it is in the base class. You can call it
`VisualSensorSpec::ptr visualSensorSpec_` for example.
and do the initialization here (as you did it now) or do it in the constructor.

src/esp/sensor/CameraSensor.h
bigbike: I do not get your question. You already cast it to the `CameraSensorSpec`, so within this CameraSensor object, it can access any items in the CameraSensorSpec.

bigbike: Again, call it `cameraSensorSpec_` for example.

src/esp/sensor/VisualSensor.h
```diff
@@ -18,12 +19,57 @@ class RenderTarget;
...
+  VisualSensorSpec();
+  SensorType sensorType = SensorType::Color;
+  float near = 0.01f;
+  float far = 1000.f;
+  Mn::Deg hfov = Mn::Deg{90.f};
+  float ortho_scale = 0.1f;
+  vec2i resolution = {84, 84};
+  std::string encoding = "rgba_uint8";  // For rendering colors in images
+  bool gpu2gpuTransfer = false;         // True for pytorch tensor support
+  void sanityCheck() {
```
bigbike, line 32: Sanitycheck also need to check the value. For example, width or height cannot be <=0.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -28,18 +39,17 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
...
-    spec_->parameters.at(elem.first) = elem.second;
-  }
+  spec_->near = spec->near;
+  spec_->far = spec->far;
 
   width_ = spec_->resolution[1];
   height_ = spec_->resolution[0];
-  near_ = std::atof(spec_->parameters.at("near").c_str());
-  far_ = std::atof(spec_->parameters.at("far").c_str());
+  near_ = spec_->near;
```
bigbike, line 51: all these member variables such as near_, height_, width_ in the CameraSensor can be deprecated. 

Before if you do not have them, you need to convert them from string every time you query these values. It is not the case any more.

src/esp/sensor/SensorFactory.cpp
bigbike: > I moved sensorType and sensorSubType out of SensorSpec

Why are you doing this? They should be in the SensorSpec.

src/esp/sensor/Sensor.h
```diff
@@ -44,22 +44,12 @@ enum class SensorSubType {
 
 // Specifies the configuration parameters of a sensor
 struct SensorSpec {
-  std::string uuid = "rgba_camera";
-  SensorType sensorType = SensorType::Color;
```
bigbike, line 48: `sensorType` and `sensorSubType` should be kept in the base class.

src/esp/sensor/CameraSensor.h
vauduong: My question was if casting was a good way to allow CameraSensor to access items in CameraSensorSpec, not just SensorSpec. (It was not cast previously) Thanks for the answer and I will rename the variable to cameraSensorSpec_!

src/esp/sensor/SensorFactory.cpp
vauduong: Good to know it was a step in the wrong direction. Based on this comment it seemed we wanted to remove it so I tried it out, I will put it back.  (https://github.com/facebookresearch/habitat-sim/pull/1090#discussion_r575594474).

src/esp/sensor/VisualSensor.h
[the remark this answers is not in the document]
vauduong: Thanks!

src/esp/sensor/VisualSensor.h
vauduong: Thanks for the guidance!

src/esp/sensor/VisualSensor.cpp
vauduong: Sorry, I missed it in the previous round of review, but want to let you know that I appreciate each and every comment from reviewers! I wasn't able to address everything I wanted to in the last commit but wanted to get some big parts in to make sure I was on the right track.

src/esp/sim/Simulator.cpp
bigbike: OK. I made a mistake here. 
Here, you are creating sensors from the sensor spec so the ownership of the spec will be transferred. So you need the ptr.

bigbike: You will have to use `esp::sensor::SensorSpec::ptr&`. Cannot change it to `esp::sensor::SensorSpec::ptr`

You will fail the compiling? or just fail the clang-tidy?

vauduong: Ok thanks! I was looking into that but could not make it work so was about to ask. SensorSpec::create() returns a SensorSpec::ptr, so we will rarely have SensorSpec.

vauduong: `esp::sensor::SensorSpec::ptr&` fails compiling but passes clang-tidy
`esp::sensor::SensorSpec::ptr` passes compiling but fails clang-tidy
Not quite sure what to do here - do you have any suggestions?

bigbike: You mentioned this error:
Error was: Could not get type esp::sensor::SensorSpec::ptr& from esp::sensor::CameraSensorSpec::ptr

May I have the location? which file? on which line?

vauduong: Let me push the change, it was when initializing a CameraSensorSpec in SimTest

vauduong: {
	"resource": "/Users/auduong/habitat-sim/src/tests/SimTest.cpp",
	"owner": "C/C++",
	"code": "434",
	"severity": 8,
	"message": "a reference of type \"esp::sensor::SensorSpec::ptr &\" (not const-qualified) cannot be initialized with a value of type \"esp::sensor::CameraSensorSpec::ptr\"",
	"source": "C/C++",
	"startLineNumber": 646,
	"startColumn": 58,
	"endLineNumber": 646,
	"endColumn": 74
}

vauduong:
```
esp::sensor::SensorSuite sensorSuite;
  auto objectSensorSpec = esp::sensor::CameraSensorSpec::create();
  objectSensorSpec->uuid = std::to_string(objectID);
  objectSensorSpec->position = {0, 0, 0};
  objectSensorSpec->orientation = {0, 0, 0};
  objectSensorSpec->resolution = {128, 128};
  sensorSuite.add(simulator->addSensorToObject(objectID, objectSensorSpec));
```

bigbike: I need to go. can you try something like pointer_cast?

vauduong: Ok no problem, thanks for the help!

src/esp/sim/Simulator.cpp
```diff
@@ -1125,7 +1125,7 @@ agent::Agent::ptr Simulator::getAgent(const int agentId) {
 }
 
 esp::sensor::Sensor::ptr Simulator::addSensorToObject(
-    const int objectId,
+    int objectId,
     esp::sensor::SensorSpec::ptr& sensorSpec) {
```
bigbike, line 1129: Oh, just saw your detailed error msg in the afternoon.
This is because one cannot bind a non-const lvalue reference to an rvalue of type.

`esp::sensor::SensorSpec::ptr&` is a non-const reference of type `esp::sensor::SensorSpec::ptr`. However, when you call this function from SimTest, the argument type is actually `esp::sensor::CameraSensorSpec::ptr` which is being converted to `esp::SensorSpec::ptr` so that a temporary, (a rvalue) is created (which you did not realize). 

Solution is to add a `const`, namely, `const esp::sensor::SensorSpec::ptr&`.

Please try it and see if it works.

src/esp/agent/Agent.h
vauduong: Thanks, I changed this so that the default AgentConfiguration has sensorSpecifications empty, and the demo in viewer has the user add a sensorSpec, as we do in SimTest.

src/esp/sim/Simulator.cpp
vauduong: Thank you that works and makes a lot of sense! I understand more about this now

src/esp/sensor/CameraSensor.cpp
```diff
@@ -14,62 +14,76 @@
...
+                 "CameraSensor::CameraSensorSpec(): sensorSpec does not have "
+                 "SensorSubType "
+                 "Pinhole or Orthographic", );
+}
+
+CameraSensorSpec::CameraSensorSpec()
+    : VisualSensorSpec(), channels(4), observationSpace("") {
+  uuid = "rgba_camera";
+  sensorSubType = SensorSubType::Pinhole;
+  sanityCheck();
```
bigbike, line 29: Think about it. What you just created within this constructor is of course legal, and thus sanityCheck() is totally unnecessary.
sanityCheck() should be called, for example, in the CamerSensor constructor, when a new "spec" is passed in as an argument, and then you have to verify it.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -14,62 +14,76 @@
 namespace esp {
 namespace sensor {
 
+void CameraSensorSpec::sanityCheck() {
+  CORRADE_ASSERT(sensorSubType == SensorSubType::Pinhole ||
```
bigbike, line 18: When this sanityCheck is called, it will only check if the variables exclusive to this class are legal. The variables in the parent class is thus ignored.

You will have to call the sanityCheck in the base class first. Here, you should call `VisualSensorSpec::sanityCheck()`.

You may do the same thing in the `VisualSensorSpec`.

src/esp/sensor/VisualSensor.h
```diff
@@ -81,14 +90,51 @@ class VisualSensor : public Sensor {
...
+   */
+  float near_ = 0.001f;
 
+  /** @brief far clipping plane
+   */
+  float far_ = 1000.0f;
+
+  /** @brief field of view
+   */
+  Mn::Deg hfov = Mn::Deg{90.f};
```
bigbike, line 133: you do not have to have this variable. It is redundant, and you need to maintain it every time when width or height is updated, which is rather error-prone.

in the `getFOV()` just compute it every time.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -109,6 +121,14 @@ void initSensorBindings(py::module& m) {
...
           R"(Get the RenderCamera in the sensor (if there is one) for rendering PYTHON DOES NOT GET OWNERSHIP)",
           pybind11::return_value_policy::reference)
+      .def_property_readonly(
+          "near", &VisualSensor::getNear,
+          R"(The distance to the near clipping plane this VisualSensor uses.)")
+      .def_property_readonly(
+          "far", &VisualSensor::getFar,
+          R"(The distance to the far clipping plane this VisualSensor uses.)")
+      .def_property_readonly("hfov", &VisualSensor::getFOV,
+                             R"(The FOV this VisualSensor uses.)")
```
bigbike, line 131: FOV --> Field Of View

src/esp/sensor/VisualSensor.h
bigbike, line 133: hfov_

bigbike: Also you can use literals here: 
You need: `using Mn::Math::Literals::operator""_degf;`
And then here you can use: `hfov_ = 90.0_degf;`

See an example here: https://github.com/facebookresearch/habitat-sim/blob/51b51e94c7fe645514758f28ada04cd49528997a/src/esp/gfx/test/DepthUnprojectionTest.cpp

src/esp/sensor/VisualSensor.h
```diff
@@ -81,14 +93,51 @@ class VisualSensor : public Sensor {
...
+   */
+  float near_ = 0.01f;
+
+  /** @brief far clipping plane
+   */
+  float far_ = 1000.0f;
 
+  /** @brief field of view
+   */
+  Mn::Deg hfov_ = 90.0_degf;
```
bigbike, line 136: kind note: PR would be a breaking change to the habitat-lab. 
You may need another PR to fix it in the habitat-lab repository.

examples/settings.py
```diff
@@ -89,7 +89,7 @@ def make_cfg(settings):
     sensor_specs = []
     for sensor_uuid, sensor_params in sensors.items():
         if settings[sensor_uuid]:
-            sensor_spec = habitat_sim.SensorSpec()
+            sensor_spec = habitat_sim.CameraSensorSpec()
```
bigbike, line 92: Before, we only have one type of sensor specification and you can do it like this in a simple way.
Now, we have different sensor specifications based on sensor type and sensor sub-type, so changes are needed accordingly. Cannot just construct the camera sensor by default.

Which specific sensor spec should be used? 
Check the sensor type first, if it is visual sensor, then check the sensor sub-type, if it is "Pinhole" or "Ortho", then select the `CameraSensorSpec()`

src/esp/sensor/CameraSensor.cpp
```diff
@@ -14,62 +14,81 @@
...
   // refresh size after relevant parameters have changed
   Mn::Vector2 nearPlaneSize_ =
-      Mn::Vector2{1.0f, static_cast<float>(height_) / width_};
+      Mn::Vector2{1.0f, static_cast<float>(cameraSensorSpec_->resolution[0]) /
+                            cameraSensorSpec_->resolution[1]};
   float scale;
-  if (spec_->sensorSubType == SensorSubType::Orthographic) {
-    scale = std::atof(spec_->parameters.at("ortho_scale").c_str());
+  if (cameraSensorSpec_->sensorSubType == SensorSubType::Orthographic) {
+    scale = cameraSensorSpec_->ortho_scale;
```
bigbike, line 80: scale is redundant.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -3,15 +3,60 @@
...
+    : SensorSpec(),
+      ortho_scale(0.1f),
+      resolution({84, 84}),
+      encoding("rgba_uint8"),
+      gpu2gpuTransfer(false) {
+  sensorType = SensorType::Color;
+}
+
+void VisualSensorSpec::sanityCheck() {
+  bool isVisualSensor =
```
bigbike, line 25: Maybe you can have a function `isVisualSensorSpec`, expose it via pybinding, and apply lit when you create sensor configs in the python code.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -14,62 +14,81 @@
...
+}
+
 CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
-                           const SensorSpec::ptr& spec)
+                           const CameraSensorSpec::ptr& spec)
     : VisualSensor(cameraNode, spec),
       baseProjMatrix_(Magnum::Math::IdentityInit),
       zoomMatrix_(Magnum::Math::IdentityInit) {
+  // Sanity check
+  cameraSensorSpec_->sanityCheck();
```
bigbike, line 43: What if cameraSensorSpec_ is nullptr? Then it would segfault here.
You should have a CORRADE_ASSERT and when it happens, give the error saying "the input spec is illegal."

It can occur. For example:

externally:
```
std::shared_ptr<SensorSpec> sp = make_shared<SensorSpec>(...);
std::shared_ptr<CameraSensorSpec> bad_csp = std::static_pointer_cast<CameraSensorSpec>(sp); // CAREFUL: This is already BAD! but bad_csp will NOT be nullptr so far.
// Then someone uses this bad_csp to construct the CameraSensor, and then cameraSensorSpec_ would be nullptr because of the dynamic_pointer_cast.

File: src/esp/sensor/Sensor.cpp [hunk 38]
```diff
@@ -11,17 +11,40 @@
...
+      sensorType(SensorType::None),
+      sensorSubType(SensorSubType::None),
+      position({0, 1.5, 0}),
+      orientation({0, 0, 0}),
+      noiseModel("None") {}
+
 Sensor::Sensor(scene::SceneNode& node, SensorSpec::ptr spec)
     : Magnum::SceneGraph::AbstractFeature3D{node}, spec_(std::move(spec)) {
   node.setType(scene::SceneNodeType::SENSOR);
   if (spec_ == nullptr) {
```
bigbike, line 37: Oh, please help us to change it to use CORRADE_ASSERT. Thanks!

src/esp/sensor/Sensor.cpp
```diff
@@ -11,17 +11,40 @@
...
-
   setTransformationFromSpec();
 }
 
+Sensor::~Sensor() {
+  LOG(INFO) << "Deconstructing Sensor";
+}
+
 void Sensor::setTransformationFromSpec() {
   if (spec_ == nullptr) {
```
bigbike, line 49: Again, CORRADE_ASSERT

src/esp/sensor/Sensor.h
```diff
@@ -38,34 +38,26 @@ enum class ObservationSpaceType {
 };
 
 enum class SensorSubType {
-  Pinhole = 0,
-  Orthographic = 1,
+  None = 0,
```
bigbike, line 41: This is great. Thank you.

src/esp/sensor/Sensor.h
```diff
@@ -38,34 +38,26 @@ enum class ObservationSpaceType {
...
-  bool gpu2gpuTransfer = false;
+  std::string uuid;
+  SensorType sensorType;
+  SensorSubType sensorSubType;
+  vec3f position;
+  vec3f orientation;
+  std::string noiseModel;
+  SensorSpec();
+  virtual ~SensorSpec();
+  bool operator==(const SensorSpec& a);
```
bigbike, line 56: `bool operator==(const SensorSpec& a) const;`
The same for the `!=`.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -3,15 +3,60 @@
...
+
+bool VisualSensorSpec::operator==(const VisualSensorSpec& a) {
+  return SensorSpec::operator==(a) && ortho_scale == a.ortho_scale &&
+         resolution == a.resolution && encoding == a.encoding &&
+         gpu2gpuTransfer == a.gpu2gpuTransfer;
+}
+
+VisualSensor::VisualSensor(scene::SceneNode& node, VisualSensorSpec::ptr spec)
+    : Sensor{node, std::move(spec)}, tgt_{nullptr} {
+  visualSensorSpec_->sanityCheck();
```
bigbike, line 58: The same here. What if `visualSensorSpec_` is nullptr.

src/esp/sensor/VisualSensor.h
```diff
@@ -81,14 +93,51 @@ class VisualSensor : public Sensor {
...
+
   /**
    * @brief Returns RenderCamera
    */
   virtual gfx::RenderCamera* getRenderCamera() = 0;
 
+  /**
+   * @brief Gets near plane distance.
+   */
+  float getNear() { return near_; }
```
bigbike, line 110: by the way, for all the getter functions, help us to mark them const.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -14,62 +14,81 @@
...
 void CameraSensor::recomputeBaseProjectionMatrix() {
   // refresh size after relevant parameters have changed
   Mn::Vector2 nearPlaneSize_ =
-      Mn::Vector2{1.0f, static_cast<float>(height_) / width_};
+      Mn::Vector2{1.0f, static_cast<float>(cameraSensorSpec_->resolution[0]) /
+                            cameraSensorSpec_->resolution[1]};
   float scale;
-  if (spec_->sensorSubType == SensorSubType::Orthographic) {
-    scale = std::atof(spec_->parameters.at("ortho_scale").c_str());
+  if (cameraSensorSpec_->sensorSubType == SensorSubType::Orthographic) {
```
bigbike, line 79: Camera Sensor can only support two mode: Pinhole or Orthographic.
Suggest to add a CORRAD_ASSERT here to verify this. 
Fail immediately when the sub-type is neither of them.

Then remove that fallback below.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -14,62 +14,81 @@
...
         Mn::Matrix4::orthographicProjection(nearPlaneSize_, near_, far_);
   } else {
-    if (spec_->sensorSubType != SensorSubType::Pinhole) {
+    if (cameraSensorSpec_->sensorSubType != SensorSubType::Pinhole) {
       LOG(INFO) << "CameraSensor::setCameraType : Unsupported Camera type val :"
-                << static_cast<int>(spec_->sensorSubType)
+                << static_cast<int>(cameraSensorSpec_->sensorSubType)
                 << " so defaulting to Pinhole.";
-      spec_->sensorSubType = SensorSubType::Pinhole;
+      cameraSensorSpec_->sensorSubType = SensorSubType::Pinhole;
```
bigbike, line 89: This is the fallback which should be removed after adding the CORRADE_ASSERT mentioned earlier.

src/esp/sensor/Sensor.cpp
bigbike: `CORRADE_ASSERT(spec_, "Error msg", );`

src/esp/sensor/Sensor.cpp
```diff
@@ -11,22 +11,63 @@
...
+  return uuid == a.uuid && sensorType == a.sensorType &&
+         sensorSubType == a.sensorSubType && position == a.position &&
+         orientation == a.orientation && noiseModel == a.noiseModel;
+}
+
+bool SensorSpec::operator!=(const SensorSpec& a) const {
+  return !(*this == a);
+}
+
+void SensorSpec::sanityCheck() {
```
vauduong, line 26: Any advice for things to make SanityCheck more robust would be appreciated!

src/esp/sensor/SensorFactory.cpp
```diff
@@ -9,9 +9,27 @@ sensor::SensorSuite SensorFactory::createSensors(
     scene::SceneNode& node,
     const sensor::SensorSetup& sensorSetup) {
   sensor::SensorSuite sensorSuite = sensor::SensorSuite();
-  for (const sensor::SensorSpec::ptr& spec : sensorSetup) {
+  for (const SensorSpec::ptr& spec : sensorSetup) {
     scene::SceneNode& sensorNode = node.createChild();
-    sensorSuite.add(sensor::CameraSensor::create(sensorNode, spec));
+    spec->sanityCheck();
```
vauduong, line 14: SensorFactory now calls the correct SanityCheck before creating Sensors

vauduong:
Thanks everyone for the reviews so far! I still have a lot of work to do to fix all the python examples, but any feedback on the current state of SanityChecks in C++ would be appreciated, I have pointed them out in the 2 comments above.

examples/tutorials/lighting_tutorial.py
```diff
@@ -64,7 +60,7 @@ def make_configuration():
     backend_cfg.enable_physics = True
 
     # agent configuration
```
vauduong, line 62: Example of single CameraSensorSpec initialization

examples/tutorials/nb_python/ECCV_2020_Advanced_Features.py
```diff
@@ -420,43 +421,35 @@ def make_cfg(settings):
             "resolution": [settings["height"], settings["width"]],
```
vauduong, line 421: Example of multiple SensorSpec initialization using a dictionary. sut.make_sensor_specs(sensors) and sut.make_sensor_specs_from_settings(sensors, settings) creates the SensorSpecs under the hood

vauduong: "sensor_subtype": habitat_sim.SensorSubType.PINHOLE is added as we need to make sure that the SensorSpec initialized is a CameraSensor

examples/tutorials/lighting_tutorial.py
```diff
@@ -64,7 +60,7 @@ def make_configuration():
...
-    sensor_cfg = habitat_sim.SensorSpec()
+    sensor_cfg = habitat_sim.CameraSensorSpec()
     sensor_cfg.resolution = [1080, 960]
     agent_cfg = habitat_sim.agent.AgentConfiguration()
     agent_cfg.sensor_specifications = [sensor_cfg]
```
vauduong, line 66: Example of a single CameraSensorSpec initialization

examples/tutorials/nb_python/ECCV_2020_Advanced_Features.py
```diff
@@ -420,43 +421,35 @@ def make_cfg(settings):
...
-            sensor_spec.resolution = sensor_params["resolution"]
-            sensor_spec.position = sensor_params["position"]
-            sensor_spec.orientation = sensor_params["orientation"]
-
-            sensor_specs.append(sensor_spec)
-
     # Here you can specify the amount of displacement in a forward action and the turn angle
     agent_cfg = habitat_sim.agent.AgentConfiguration()
-    agent_cfg.sensor_specifications = sensor_specs
+    agent_cfg.sensor_specifications = sut.make_sensor_specs(sensors)
```
vauduong, line 452: Example of multiple SensorSpec initialization using a dictionary. sut.make_sensor_specs(sensors) and sut.make_sensor_specs_from_settings(sensors, settings) creates the SensorSpecs under the hood

"sensor_subtype": habitat_sim.SensorSubType.PINHOLE is added as we need to make sure that the SensorSpec initialized is a CameraSensor

examples/tutorials/nb_python/ECCV_2020_Interactivity.py
```diff
@@ -121,44 +122,37 @@ def make_cfg(settings):
...
-
-            sensor_specs.append(sensor_spec)
-
     # Here you can specify the amount of displacement in a forward action and the turn angle
     agent_cfg = habitat_sim.agent.AgentConfiguration()
-    agent_cfg.sensor_specifications = sensor_specs
-
+    agent_cfg.sensor_specifications = sut.make_sensor_specs_from_settings(
+        sensors, settings
+    )
```
vauduong, line 155: Example of multiple SensorSpec initialization using a dictionary. sut.make_sensor_specs(sensors) and sut.make_sensor_specs_from_settings(sensors, settings) creates the SensorSpecs under the hood

"sensor_subtype": habitat_sim.SensorSubType.PINHOLE is added as we need to make sure that the SensorSpec initialized is a CameraSensor

vauduong:
Hello @bigbike @Skylion007 @aclegg3 @mosra @jturner65 , thanks everyone for the discussion last week about how to modify python examples with the updated hierarchical SensorSpec. I noticed two cases and have pointed out examples above: single SensorSpec initialization and multiple SensorSpec initialization from a dictionary.

- For single SensorSpec initialization, I changed occurrences of `sensor_cfg = habitat_sim.SensorSpec()` to
    `sensor_cfg = habitat_sim.CameraSensorSpec()`. This closely matches what we're doing in C++ to initialize SensorSpecs, and in the future users can initialize the correct SensorSpec they want by calling something like `sensor_cfg = habitat_sim.FisheyeSensorSpec()`.

- For multiple SensorSpec initialization from a dictionary, I added 2 methods in sim_utils based on our discussion: `make_sensor_specs(sensors)` and `make_sensor_specs_from_settings(sensors, settings)` for when settings specifies when certain sensors are True or False. These functions initialize the correct SensorSpecs and return them in a list.

Let me know what you think of this implementation and looking forward to any more feedback!

habitat_sim/utils/sim_utils.py
```diff
@@ -0,0 +1,103 @@
...
+    :param sensors: Dictionary of sensors to initialize and relevant parameters, sensor_type and sensor_subtype are mandatory for each sensor
+    :param settings: Dictionary modifying parameters and toggling sensors as true/false
+    :return: List[hsim.SensorSpec] of initialized SensorSpecs
+    """
+    sensor_specs = []
+    for sensor_uuid, sensor_params in sensors.items():
+        if settings[sensor_uuid]:
+            # Check if type VisualSensorSpec
+            if (
+                sensor_params["sensor_type"] == habitat_sim.SensorType.COLOR
```
Skylion007, line 29: This would be cleaner to use ```in set syntax```

habitat_sim/utils/sim_utils.py
```diff
@@ -0,0 +1,103 @@
...
+                    sensor_spec.gpu2gpu_transfer = False
+                    if "silent" in settings and not settings["silent"]:
+                        print("==== Initialized Sensor Spec: =====")
+                        print("Sensor uuid: ", sensor_spec.uuid)
+                        print("Sensor type: ", sensor_spec.sensor_type)
+                        print("Sensor position: ", sensor_spec.position)
+                        print("===================================")
+                    sensor_specs.append(sensor_spec)
+                else:
+                    # TODO: Add more checks for other types of sensors
```
Skylion007, line 58: Should raise errors

vauduong: Thanks for the feedback! I'll also write some unit tests so that we get expected errors

habitat_sim/utils/sim_utils.py
```diff
@@ -0,0 +1,105 @@
...
+                )
+            else:
+                # Check if type CameraSensorSpec
+                if sensor_params["sensor_subtype"] not in CameraSensorSubTypeSet:
+                    # TODO: Add more checks for other types of sensors
+                    raise ValueError(
+                        f"""{sensor_params["sensor_subtype"]} is an illegal sensorSubType for a VisualSensor"""
+                    )
+                else:
+                    sensor_spec = habitat_sim.CameraSensorSpec()
```
bigbike, line 51: If the subtype is "pinhole" or "ortho", it will use CameraSensorSpec() to initialize it.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -15,64 +15,80 @@
...
+CameraSensorSpec::CameraSensorSpec() : VisualSensorSpec() {
+  uuid = "rgba_camera";
+  sensorSubType = SensorSubType::Pinhole;
+}
+
+void CameraSensorSpec::sanityCheck() {
+  VisualSensorSpec::sanityCheck();
+  CORRADE_ASSERT(sensorSubType == SensorSubType::Pinhole ||
+                     sensorSubType == SensorSubType::Orthographic,
+                 "CameraSensor::CameraSensorSpec(): sensorSpec does not have "
```
bigbike, line 27: CameraSensorSpec::sanityCheck()
Please check and fix the similar errors everywhere.

bigbike (approved):
This is GREAT! Thanks a lot @vauduong for your hard and important work!!

Just some small comments and I believe it is good to go.

@Skylion007 : Do you need to take a final pass on the Python code and the tutorial code??

habitat_sim/utils/sim_utils.py
vauduong: Yes! That is what we're doing here :)
clang-tidy formatting has it so that the errors should go before the valid cases, that's why we raise the ValueError if subtype is not pinhole or ortho, and use CameraSensorSpec() to initialize otherwise

Skylion007: No reason to havee the else then because raising the Value will terminate the function. No reason to waste white space.

examples/tutorials/lighting_tutorial.py
Skylion007: Now I am starting to see how a SensorFactory could be useful here.

habitat_sim/utils/sim_utils.py
bigbike: @vauduong : no, you do not understand what I am talking about.
In the future there would be `fisheyeSensorSpec`, not just `CameraSensorSpec`. You have to have the if clause to control which one to use.

src/esp/sensor/Sensor.cpp
```diff
@@ -11,22 +11,53 @@
...
+  CORRADE_ASSERT(this, "SensorSpec::sanityCheck(): sensorSpec is illegal", );
+  // Check that all parameters are initalized to legal values
+  CORRADE_ASSERT(!uuid.empty(),
+                 "SensorSpec::sanityCheck(): uuid is unitialized", );
```
bigbike, line 32: you can just say: "Sensor uuid cannot be empty string."

src/esp/sensor/Sensor.cpp
```diff
@@ -11,22 +11,53 @@
...
+      sensorType >= SensorType::None && sensorType <= SensorType::Text,
+      "SensorSpec::sanityCheck(): sensorType is illegal", );
+  CORRADE_ASSERT(sensorSubType >= SensorSubType::None &&
+                     sensorSubType <= SensorSubType::Orthographic,
+                 "SensorSpec::sanityCheck(): sensorSubType is illegal", );
+  CORRADE_ASSERT((abs(position.array()) >= 0).any(),
+                 "SensorSpec::sanityCheck(): position is illegal", );
+  CORRADE_ASSERT((abs(orientation.array()) >= 0).any(),
+                 "SensorSpec::sanityCheck(): orientation is illegal", );
+  CORRADE_ASSERT(!noiseModel.empty(),
```
bigbike, line 43: Are you sure noise model is a must have?

src/esp/sensor/SensorFactory.cpp
bigbike: It is not needed because when you create the sensor, it will be checked within the sensor's constructor.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -3,17 +3,68 @@
...
+void VisualSensorSpec::sanityCheck() {
+  SensorSpec::sanityCheck();
+  bool isVisualSensor =
+      (sensorType == SensorType::Color || sensorType == SensorType::Depth ||
+       sensorType == SensorType::Normal || sensorType == SensorType::Semantic);
+  CORRADE_ASSERT(
+      isVisualSensor,
+      "VisualSensorSpec::sanityCheck(): sensorType must be Color, Depth, "
+      "Normal, or Semantic", );
+  if (noiseModel == "Gaussian" || noiseModel == "Poisson" ||
```
bigbike, line 28: @Skylion007 : is noise model a must have??

src/esp/sensor/VisualSensor.h
```diff
@@ -19,11 +20,24 @@ class RenderTarget;
 
 namespace sensor {
 
+using Mn::Math::Literals::operator""_degf;
+
+struct VisualSensorSpec : public SensorSpec {
+  float ortho_scale = 0.1f;
+  vec2i resolution = {84, 84};
```
bigbike, line 27: add a comment:  // height x width

src/utils/viewer/viewer.cpp
```diff
@@ -600,6 +599,9 @@ void Viewer::switchCameraType() {
       cam.setCameraType(esp::sensor::SensorSubType::Pinhole);
       return;
     }
+    case esp::sensor::SensorSubType::None: {
```
bigbike, line 602: add `CORRADE_INTERNAL_ASSERT_UNREACHABLE();`

src/esp/sensor/VisualSensor.cpp
vauduong: I am not sure if noise model is a must have, but I see in hab-lab that there are sensors which use the noise model (seems like only visual sensors though).  @Skylion007 what do you think we should do here? 

These checks in the VisualSensorSpec mirror checks on they python side.

habitat_sim/utils/sim_utils.py
vauduong: I see, thank you, I will change that then!

src/esp/sensor/VisualSensor.cpp
vauduong: Yili, I think your question is whether we want noiseModel to be an attribute of SensorSpec or VisualSensorSpec, (and not to get rid of noiseModel in all SensorSpecs) right? Since for nonVisualSensorSpecs noiseModel will always be None. Let me know if this is what you're asking.

vauduong: Ok, from discussion in hab-sim dev sync we want to keep the attribute in SensorSpec because we anticipate that future non-visual sensor specs will want to add a noiseModel

vauduong: I think we should keep the checks for noiseModel in SanityCheck because it would be nice to have for C++ and js, but let me know if you would like them removed because of redundancy on the python side.

habitat_sim/utils/sim_utils.py
vauduong: I fixed it based on my understanding in the latest commit. Take a look and let me know if I got everything right.

habitat_sim/utils/sim_utils.py
```diff
@@ -0,0 +1,97 @@
...
+    sensor_specs = []
+    for sensor_uuid, sensor_params in sensors.items():
+        if settings[sensor_uuid]:
+            # Check if type VisualSensorSpec
+            if sensor_params["sensor_type"] not in VisualSensorTypeSet:
+                # TODO: Add more checks for NonVisualSensorSpec
+                raise ValueError(
+                    f"""{sensor_params["sensor_type"]} is an illegal sensorType that is not implemented yet"""
+                )
+            else:
```
Skylion007, line 43:
```suggestion
```
Once again, you don't need this else here.

habitat_sim/utils/sim_utils.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    sensor_specs = []
+    for sensor_uuid, sensor_params in sensors.items():
+        # Check if type VisualSensorSpec
+        if sensor_params["sensor_type"] not in VisualSensorTypeSet:
+            # TODO: Add more checks for NonVisualSensorSpec
+            raise ValueError(
+                f"""{sensor_params["sensor_type"]} is an illegal sensorType that is not implemented yet"""
+            )
+        else:
```
Skylion007, line 81:
```suggestion
```
Don't need this else either

Skylion007 (changes requested):
Else issues in python

src/esp/bindings_js/modules/simenv_embind.js
```diff
@@ -160,12 +160,23 @@ class SimEnv {
   }
 
   createSensorSpec(config) {
-    const converted = new Module.SensorSpec();
-    for (let key in config) {
-      let value = config[key];
-      converted[key] = value;
+    //Check if VisualSensor
+    if (
+      config["sensorType"] == Module.SensorType.COLOR ||
```
Skylion007, line 165: This can also be simplified using a Javascript Set

src/esp/sensor/VisualSensor.cpp
Skylion007: All the noise models are in Pure Python now, it would be a good idea to move them to C++ if performance warrants it though.

habitat_sim/utils/sim_utils.py
vauduong: Oh ok, so we just want to remove the else and keep the errors. Sorry for misunderstanding, I'll fix.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -3,17 +3,68 @@
...
+      noiseModel == "SaltAndPepper" || noiseModel == "Speckle") {
+    CORRADE_ASSERT(
+        sensorType == SensorType::Color,
+        "VisualSensorSpec::sanityCheck(): sensorType must be Color if "
+        "noiseModel is Gaussian, Poisson, SaltAndPepper, or Speckle", );
+  }
+  if (noiseModel == "Redwood") {
+    CORRADE_ASSERT(
```
Skylion007, line 36: Can these noise models even be used from C++?

src/esp/sensor/VisualSensor.h
```diff
@@ -19,11 +20,24 @@ class RenderTarget;
 
 namespace sensor {
 
+using Mn::Math::Literals::operator""_degf;
+
+struct VisualSensorSpec : public SensorSpec {
+  float ortho_scale = 0.1f;
+  vec2i resolution = {84, 84};          // height x width
```
Skylion007, line 27: Why 84, 84 default? Seems like an odd choice to not pick a power of 2.

Skylion007 (changes requested):
More issues.

src/esp/sensor/VisualSensor.h
vauduong: There's real no reason for this besides that this was the default value previously. Happy to change to 128, 128

vauduong:
Thanks @bigbike and @Skylion007 I've addressed all your feedback about the point regarding noiseModel. I understand it's not used in C++ but we are deciding to keep the attribute in SensorSpec, and anticipate that one day we want to include it in C++. Could you confirm/deny that we want to remove the sanityChecks for noiseModel in this PR?

src/esp/sensor/Sensor.h
bigbike: You do not have to give it a name `None`, you can just keep the string empty. 
And keep this term in the SensorSpec in case non-visual sensor also need it. That is it.

src/esp/sensor/VisualSensor.cpp
bigbike: So far I did not see the noise models are used anywhere in the C++, but maybe I am wrong. 

I think we can keep the sanity check like this since it is coded up. It can be easily modified in the future.

But we need to make sure the logic is correct at least. For example, when "redwood" is noise model, the sensor type must be depth sensor.

vauduong: Ok sounds good! I got all the logic for the noiseModels from doing a global search for the noiseModels
Screen Shot 2021-02-25 at 2 19 53 PM

erikwijmans (approved):
This is very nice!  Thank you for all the work @vauduong

Skylion007 (approved):

src/utils/viewer/viewer.cpp
```diff
@@ -523,15 +523,14 @@ Viewer::Viewer(const Arguments& arguments)
...
+  auto pinholeCameraSpec = esp::sensor::CameraSensorSpec::create();
+  pinholeCameraSpec->sensorSubType = esp::sensor::SensorSubType::Pinhole;
+  pinholeCameraSpec->sensorType = esp::sensor::SensorType::Color;
+  pinholeCameraSpec->position = {0.0f, 1.5f, 0.0f};
+  pinholeCameraSpec->orientation = {0, 0, 0};
+  pinholeCameraSpec->resolution = esp::vec2i(viewportSize[1], viewportSize[0]);
+  agentConfig.sensorSpecifications = {pinholeCameraSpec};
 
-  agentConfig.sensorSpecifications[0]->sensorSubType =
-      args.isSet("orthographic") ? esp::sensor::SensorSubType::Orthographic
```
bigbike, line 530: @vauduong :
This is something we did not capture.
You should set the the sub-type based on the input argument.
Please issue any other quick fix when possible! Thanks!

Rename the `pinholeCameraSpec` to `cameraSpec` since it could be ortho camera as well.

vauduong: Ok, I'll make that fix!

src/esp/sensor/CameraSensor.h
```diff
@@ -109,67 +117,72 @@ class CameraSensor : public VisualSensor {
...
   }
-  float getNear() { return near_; }
+  float getNear() const { return near_; }
 
   /**
    * @brief Sets far plane distance.
    */
   void setFar(float _far) {
-    spec_->parameters.at("far") = std::to_string(_far);
+    CORRADE_ASSERT(
```
bigbike, line 179: @vauduong :
This is something we did not capture.
"near" and "far" should be in the `VisualSensor` spec! They were in the spec before refactoring.
 Then member variables, `far_` and `near_` should be removed.
 Please do fix it in another PR.

bigbike: @vauduong : Also, please lift up the 
  `bool getObservation(sim::Simulator& sim, Observation& obs) override;`
  `bool getObservationSpace(ObservationSpace& space) override;`
  from `CameraSensor` to `VisualSensor`

bigbike: move `ortho_scale` to the `CameraSensorSpec`.
move `channel` and `observationSpace` to `VisualSensorSpec`. (@Skylion007 : is `observationSpace` used somewhere in the python side??)

Skylion007: It's replicated entirely on the Python side. The C++ isn't used directly.

bigbike: deprecate the `encoding` (I did not see it was used anywhere, please double check).
deprecate the `observationSpace`.

bigbike: @vauduong : you can see the branch "cubemap". See what I did there for your reference.

PR #1094, opened 2021-02-15
[CI] Improve Clang-Tidy and kill std::binds

Skylion007:
## Motivation and Context

* This fixes a potential memory leak in the Sensor objects and also upgrades clang-tidy to prevent these errors from getting merged in again. The failing test seems unrelated to the PR, oddly enough. 

## How Has This Been Tested
* Tested locally on Clang

## Types of changes

## Checklist


.clang-tidy
```diff
@@ -42,5 +43,10 @@ performance-*,
...
 '*,
 -bugprone-narrowing-conversions,
+-bugprone-reserved-identifier,
 -bugprone-unused-raii,
--clang-diagnostic-deprecated-declarations'
+-clang-diagnostic-deprecated-declarations,
+-clang-analyzer-cplusplus.PlacementNew,
+-modernize-avoid-bind,
+' # TODO fix std::binds later
+# TODO Figure out if PlacementNew is a false positive
```
mosra, line 52: > .../rapidjson/document.h:1143:21: error: Storage type is aligned to 1 bytes but allocated type is aligned to 8 bytes [clang-analyzer-cplusplus.PlacementNew,-warnings-as-errors]

Sigh. Not a false positive, no. Together with the RapidJson assert that prevents us from reliably running AddressSanitzer I'm starting to feel that we should drop RapidJson and replace it with something that is less of a dumpster fire...

src/esp/assets/Mp3dInstanceMeshData.cpp
```diff
@@ -55,7 +55,7 @@ bool Mp3dInstanceMeshData::loadMp3dPLY(const std::string& plyFile) {
     LOG(ERROR) << "Invalid element vertex header line";
     return false;
   }
-  int nVertex;
+  int nVertex = 0;
   iss >> token >> nVertex;
```
mosra, line 59: ... wait, why a custom PLY parser is still used here? :eyes:

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -323,8 +330,9 @@ void StageAttributesManager::setValsFromJSONDoc(
   std::string semanticFName = stageAttributes->getSemanticAssetHandle();
   semanticFName = this->setJSONAssetHandleAndType(
       stageAttributes, jsonConfig, "semantic_asset_type", "semantic_asset",
-      semanticFName,
-      std::bind(&StageAttributes::setSemanticAssetType, stageAttributes, _1));
+      semanticFName, [stageAttributes](auto&& PH1) {
+        stageAttributes->setSemanticAssetType(std::forward<decltype(PH1)>(PH1));
```
mosra, line 334: What's the "PH1"? If anything, doesn't look like it matches our coding conventions :)

src/esp/nav/PathFinder.cpp
```diff
@@ -14,9 +14,10 @@
 #include <Magnum/EigenIntegration/Integration.h>
 
 #include <Corrade/Containers/Optional.h>
+#include <math.h>
 
 #include <cstdio>
-#define _USE_MATH_DEFINES
+#define USE_MATH_DEFINES
```
mosra, line 20: Why this change? What's the macro for?

Why is `<math.h>` included again, if we have `<cmath>` below?

src/esp/sensor/CameraSensor.cpp
```diff
@@ -5,6 +5,7 @@
 #include <Magnum/ImageView.h>
 #include <Magnum/Math/Algorithms/GramSchmidt.h>
 #include <Magnum/PixelFormat.h>
+#include <math.h>
```
mosra, line 8: `<cmath>`, prefer the C++ header wrappers.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -28,6 +29,10 @@ CameraSensor::CameraSensor(scene::SceneNode& cameraNode,
   renderCamera_->setViewport(this->framebufferSize());
 }  // ctor
 
+CameraSensor::~CameraSensor() {
+  renderCamera_ = nullptr;
```
mosra, line 33: Why?

src/esp/metadata/managers/StageAttributesManager.cpp
Skylion007: These fixes were automatically generated by clang-tidy. It's short for std::PlaceHolder::_1, since the std::binds didn't not have any informative args before. @jturner65 Another advantage of lambdas is that you can name your place holders to be more informative.

src/esp/sensor/CameraSensor.cpp
Skylion007: Hmm, was this one also a false positive?  Was to fix: https://clang.llvm.org/docs/analyzer/checkers.html#cplusplus-newdeleteleaks-c

It was orignally complaining about new RenderCamera() not having a corresponding deletion, but calling delete() on RenderCamera was causing tests to fail here so I tried removing the reference from it like we do in Simulator.cpp . Is there a better way to solve this issue? I am not that familiar with C++ memory management.

src/esp/sim/Simulator.cpp
```diff
@@ -577,6 +577,7 @@ void Simulator::reconfigureReplayManager() {
   gfxReplayMgr_->setPlayerCallback(
       std::bind(&esp::sim::Simulator::loadAndCreateRenderAssetInstance, this,
                 std::placeholders::_1, std::placeholders::_2));
+  // TODO: Is there a betterway than std::bind ?
```
erikwijmans, line 580: Like this?
```cpp
  gfxReplayMgr_->setPlayerCallback(
      [this](const assets::AssetInfo& assetInfo,
             const assets::RenderAssetInstanceCreationInfo& creation)
          -> scene::SceneNode* {
        return loadAndCreateRenderAssetInstance(assetInfo, creation);
      });
 ```

File: src/esp/metadata/managers/StageAttributesManager.cpp (hunk 3 above)
Comment (mosra) [reply src/esp/metadata/managers/StageAttributesManager.cpp:334]: Ah, autogenerated, I see :)

File: src/esp/sensor/CameraSensor.cpp (hunk 6 above)
Comment (mosra) [reply src/esp/sensor/CameraSensor.cpp:33]: Assigning a null to a raw pointer in a destructor doesn't achieve anything :) The LGTM "leak" report was a false positive, so no action is needed.

File: src/esp/sim/Simulator.cpp (hunk 7 above)
Comment (Skylion007) [reply src/esp/sim/Simulator.cpp:580]: @erikwijmans Yeah, they typing here won't allow an anonymous function though: https://github.com/facebookresearch/habitat-sim/blob/a3820c4ea7dad72026d8b2a6562349fd31b3ac8b/src/esp/gfx/replay/ReplayManager.h#L40

File: src/esp/sensor/CameraSensor.cpp [hunk 8]
```diff
@@ -55,7 +57,7 @@ void CameraSensor::recomputeBaseProjectionMatrix() {
   // refresh size after relevant parameters have changed
   Mn::Vector2 nearPlaneSize_ =
       Mn::Vector2{1.0f, static_cast<float>(height_) / width_};
-  float scale;
+  float scale = NAN;
```
bigbike, line 60: NAN means not-a-number.
And scale will be used later in `nearPlaneSize_ /= scale;`
I think `scale = 1.0f` would be a reasonable initial value.

src/esp/sim/Simulator.cpp
erikwijmans: That compiles fine for me.  A lambda should be fine for `std::function`.   Did you try the clang-tidy autogen fix and maybe that didn't work because it is using `auto&&` which then won't match?

src/esp/sensor/CameraSensor.cpp
Skylion007: I would prefer NaN, as scale SHOULD be set. The reason to initially set it as NaN is to know something has gone horribly wrong in the constructor or elsewhere.

Scale will only be user later in nearPlaneSize_ /= scale after being parsed from the config value.

src/esp/assets/Mp3dInstanceMeshData.cpp
erikwijmans: That's used just for processing the mp3d semantic ply's into what habitat ends up loading.  IIRC we left this here because it's used just fr that and the mp3d semantic ply's have a whole bunch of weird face attributes instead of our 1 weird one.

src/esp/sim/Simulator.cpp
Skylion007: @erikwijmans Yeah, it compiled fine locally on my mac, but failed on GCC5 on CI.

src/esp/metadata/managers/StageAttributesManager.cpp
jturner65: Very nice, tyvm :)

src/esp/metadata/managers/LightLayoutAttributesManager.cpp
```diff
@@ -98,23 +98,27 @@ void LightLayoutAttributesManager::setLightInstanceValsFromJSONDoc(
   // jsonConfig here holds the JSON description for a single light attributes.
   // set position
   bool posIsSet = io::jsonIntoConstSetter<Magnum::Vector3>(
-      jsonConfig, "position",
-      std::bind(&LightInstanceAttributes::setPosition, lightAttribs, _1));
+      jsonConfig, "position", [lightAttribs](auto&& PH1) {
+        lightAttribs->setPosition(std::forward<decltype(PH1)>(PH1));
+      });
```
jturner65, line 103: This is great! Thanks Aaron. I'm going through these and will rewrite the lamdas to be a bit more semantically meaningful.

jturner65 (approved):
LGTM!

PR #1095, opened 2021-02-16
Dynamic culling

erikwijmans:
## Motivation and Context

Culling for dynamic objects.  This updates the local AABB to be in world coordinates if a scene node doesn't have a precomputed AABB.  Uses caching and lazy computation to be efficient (@mosra I think I have done this correctly, but I would appreciate a double check).

## How Has This Been Tested

Tests pass, in the viewer now you can see the number of culled objects increase as you add objects.

## Types of changes

### DIFF
--- a/src/esp/gfx/RenderCamera.cpp
+++ b/src/esp/gfx/RenderCamera.cpp
@@ -120,21 +120,17 @@ size_t RenderCamera::cull(
                           Mn::Matrix4>& a) {
         // obtain the absolute aabb
         auto& node = static_cast<scene::SceneNode&>(a.first.get().object());
-        Corrade::Containers::Optional<Mn::Range3D> aabb =
-            node.getAbsoluteAABB();
-        if (aabb) {
-          // if it has an absolute aabb, it is a static mesh
-          Cr::Containers::Optional<int> culledPlane =
-              rangeFrustum(*aabb, frustum, node.getFrustumPlaneIndex());
-          if (culledPlane) {
-            node.setFrustumPlaneIndex(*culledPlane);
-          }
-          // if it has value, it means the aabb is culled
-          return (culledPlane != Cr::Containers::NullOpt);
-        } else {
-          // keep the drawable if its node does not have an absolute AABB
-          return false;
+        // This updates the AABB for dynamic objects if needed
+        node.setClean();
+        const Mn::Range3D& aabb = node.getAbsoluteAABB();
+
+        Cr::Containers::Optional<int> culledPlane =
+            rangeFrustum(aabb, frustum, node.getFrustumPlaneIndex());
+        if (culledPlane) {
+          node.setFrustumPlaneIndex(*culledPlane);
         }
+        // if it has value, it means the aabb is culled
+        return (culledPlane != Cr::Containers::NullOpt);
       });
 
   return (newEndIter - drawableTransforms.begin());
--- a/src/esp/scene/SceneNode.cpp
+++ b/src/esp/scene/SceneNode.cpp
@@ -66,6 +66,11 @@ Mn::Vector3 SceneNode::absoluteTranslation() const {
     return absoluteTransformation_.translation();
 }
 
+Mn::Vector3 SceneNode::absoluteTranslation() {
+  setClean();
+  return absoluteTransformation_.translation();
+}
+
 const Mn::Range3D& SceneNode::getAbsoluteAABB() const {
   if (aabb_)
     return *aabb_;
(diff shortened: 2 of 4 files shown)


src/esp/scene/SceneNode.h
```diff
@@ -32,7 +32,8 @@ enum class SceneNodeType {
   OBJECT = 4,  // objects added via physics api
 };
 
-class SceneNode : public MagnumObject {
+class SceneNode : public MagnumObject,
+                  public Magnum::SceneGraph::AbstractFeature3D {
```
bigbike, line 36: My biggest concern comes from this inheritance. SceneNode is just a pure scene graph node, inherited from the MagnumObject, and a "feature" is an attachment on that node.
Logically, a class should not be a node and an attachment to that node at the same time. This is not a good design in my opinion.

Can we do it using "composition method" instead of "inheritance method"? Namely you created a new feature inherited from AbstractFeature3D, and add a "feature pointer" to this new feature in the SceneNode class?

src/esp/scene/SceneNode.cpp
```diff
@@ -47,5 +53,29 @@ const Mn::Range3D& SceneNode::computeCumulativeBB() {
...
+    return absoluteTransformation_.translation();
+}
+
+const Mn::Range3D& SceneNode::getAbsoluteAABB() const {
+  if (aabb_)
+    return *aabb_;
+  else {
+    if (!worldCumulativeBB_)
+      worldCumulativeBB_ = {
+          geo::getTransformedBB(getCumulativeBB(), absoluteTransformation_)};
```
bigbike, line 75: Glad to see `geo::getTransformedBB` becomes useful in this performance-critical project. I remembered I optimized this function so that it became 13x faster.

src/esp/scene/SceneNode.h
erikwijmans: This is the way it was recommended by the magnum docs and I believe it is the only way to do it: https://doc.magnum.graphics/magnum/scenegraph.html#scenegraph-features-caching

src/esp/scene/SceneNode.cpp
erikwijmans: Yeah!  The existence of that function and it being very fast is what makes this work well :)

bigbike:
@mosra: would you take a pass? Thanks!

src/esp/scene/SceneNode.cpp
mosra: > Logically, a class should not be a node and an attachment to that node at the same time. 

That's supported *by design*, and perfectly fine. (Unless your coding guidelines ban multiple inheritance.)

Could be also a member variable, but then everything becomes more verbose, and you'll need another derived class and...

src/esp/scene/SceneNode.h
mosra: > Logically, a class should not be a node and an attachment to that node at the same time. 

That's supported *by design*, and perfectly fine. (Unless your coding guidelines ban multiple inheritance.)

Could be also a member variable, but then everything becomes more verbose, and you'll need another derived class and...

src/esp/scene/SceneNode.cpp
```diff
@@ -47,5 +53,29 @@ const Mn::Range3D& SceneNode::computeCumulativeBB() {
...
+  worldCumulativeBB_ = Cr::Containers::NullOpt;
+
+  absoluteTransformation_ = absoluteTransformation;
+}
+
+Mn::Vector3 SceneNode::absoluteTranslation() const {
+  if (isDirty())
+    return absoluteTransformation().translation();
+  else
+    return absoluteTransformation_.translation();
```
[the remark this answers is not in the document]
erikwijmans: The rationale here was that this method is marked as const (and makes sense to be const).  I can cast around that tho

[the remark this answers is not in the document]
mosra: ooh, right

(add a non-const overload that saves the work at least in 50% of cases? :smile:)

src/esp/gfx/RenderCamera.cpp
```diff
@@ -120,21 +120,17 @@ size_t RenderCamera::cull(
...
+        node.setClean();
+        const Mn::Range3D& aabb = node.getAbsoluteAABB();
+
+        Cr::Containers::Optional<int> culledPlane =
+            rangeFrustum(aabb, frustum, node.getFrustumPlaneIndex());
+        if (culledPlane) {
+          node.setFrustumPlaneIndex(*culledPlane);
         }
+        // if it has value, it means the aabb is culled
+        return (culledPlane != Cr::Containers::NullOpt);
```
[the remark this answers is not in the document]
erikwijmans: Yeah, but I do like the explicitness of this

src/esp/scene/SceneNode.cpp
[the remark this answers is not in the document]
erikwijmans: Yeah, that's a good idea.

Skylion007 (approved):

aclegg3 (approved):
🎉

PR #1099, opened 2021-02-17
--Support stand alone Metadata Mediator, and building Simulator from MM

jturner65:
## Motivation and Context
This PR adds the ability to construct a Metadata Mediator in python, either from an existing Simulator Configuration or using a default, without requiring a Simulator to exist.  It also adds the ability to pass an existing Metadata Mediator into Simulator as a constructor default, as well as to retrieve a reference to the Metadata Mediator currently in use from Simulator.

The purpose of this is that Metadata Mediator maintains all the configuration files that have been loaded into the system, which can be expensive for large datasets.  By enabling the MM to be created and accessed independent of a Simulator instance, this loading process can be performed once, and then maintained through multiple simulator lifetimes.  

## How Has This Been Tested
Existing C++ and python tests.  Simtest.cpp has been modified to test Simulators constructed using MM as well as those constructed from SimulatorConfiguration.  TODO : python test is pending.

## Types of changes

## Checklist

- [ x I have read the **CONTRIBUTING** document.

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -188,10 +188,7 @@ class Simulator(SimulatorBackend):
 
     def _config_backend(self, config: Configuration) -> None:
         if not self._initialized:
-            if config.metadata_mediator is not None:
-                super().__init__(config.metadata_mediator)
-            else:
-                super().__init__(config.sim_cfg)
+            super().__init__(config.sim_cfg, config.metadata_mediator)
             self._initialized = True
         else:
             super().reconfigure(config.sim_cfg)
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -30,6 +30,7 @@ import habitat_sim.errors
 from habitat_sim.agent.agent import Agent, AgentConfiguration, AgentState
 from habitat_sim.bindings import cuda_enabled
 from habitat_sim.logging import logger
+from habitat_sim.metadata import MetadataMediator
 from habitat_sim.nav import GreedyGeodesicFollower, NavMeshSettings, PathFinder
 from habitat_sim.sensor import SensorSpec, SensorType
 from habitat_sim.sensors.noise_models import make_sensor_noise_model
@@ -45,13 +46,15 @@ class Configuration(object):
 
     :property sim_cfg: The configuration of the backend of the simulator
     :property agents: A list of agent configurations
-
+    :property metadata_mediator: (optional) The metadata mediator to build the simulator from
     Ties together a backend config, `sim_cfg` and a list of agent
     configurations `agents`.
     """
 
     sim_cfg: SimulatorConfiguration
     agents: List[AgentConfiguration]
+    # An existing Metadata Mediator can also be used to construct a SimulatorBackend
+    metadata_mediator: MetadataMediator = attr.ib(default=None)
 
 
 @attr.s(auto_attribs=True)
@@ -185,7 +188,10 @@ class Simulator(SimulatorBackend):
 
     def _config_backend(self, config: Configuration) -> None:
         if not self._initialized:
-            super().__init__(config.sim_cfg)
+            if config.metadata_mediator is not None:
+                super().__init__(config.metadata_mediator)
+            else:
+                super().__init__(config.sim_cfg)
             self._initialized = True
         else:
             super().reconfigure(config.sim_cfg)
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -54,7 +54,7 @@ class Configuration(object):
     sim_cfg: SimulatorConfiguration
     agents: List[AgentConfiguration]
     # An existing Metadata Mediator can also be used to construct a SimulatorBackend
-    metadata_mediator: MetadataMediator = attr.ib(default=None)
+    metadata_mediator: Optional[MetadataMediator] = None
 
 
 @attr.s(auto_attribs=True)
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -116,7 +116,7 @@ void initSimBindings(py::module& m) {
           R"(The currently active dataset being used.  Will attempt to load
             configuration files specified if does not already exist.)")
       /* --- Template Manager accessors --- */
-      // We wish a copy of the metadata mediator smart poitner so that we
+      // We wish a copy of the metadata mediator smart pointer so that we
       // increment its ref counter
       .def_property(
           "metadata_mediator", &Simulator::getMetadataMediator,
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -116,6 +116,8 @@ void initSimBindings(py::module& m) {
           R"(The currently active dataset being used.  Will attempt to load
             configuration files specified if does not already exist.)")
       /* --- Template Manager accessors --- */
+      // We wish a copy of the metadata mediator smart poitner so that we
+      // increment its ref counter
       .def_property(
           "metadata_mediator", &Simulator::getMetadataMediator,
           &Simulator::setMetadataMediator, py::return_value_policy::copy,
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -74,6 +74,8 @@ void initSimBindings(py::module& m) {
 
   // ==== Simulator ====
   py::class_<Simulator, Simulator::ptr>(m, "Simulator")
+      // add constructor to build Simulator from existing MetadataMediator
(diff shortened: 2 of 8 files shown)


src/tests/SimTest.cpp
```diff
@@ -129,7 +167,7 @@ SimTest::SimTest() {
             &SimTest::recomputeNavmeshWithStaticObjects,
             &SimTest::loadingObjectTemplates,
             &SimTest::buildingPrimAssetObjectTemplates,
-            &SimTest::addSensorToObject});
+            &SimTest::addSensorToObject}, NumSims );
```
mosra, line 170:
```suggestion
            &SimTest::addSensorToObject}, Cr::Containers::arraySize(SimulationBuilder) );
```

and you don't need the `NumSims` variable at all (sorry, it's still like that in the docs, i have to clean that up)

src/tests/SimTest.cpp
```diff
@@ -261,7 +336,9 @@ void SimTest::getDefaultLightingRGBAObservation() {
 void SimTest::getCustomLightingRGBAObservation() {
   Corrade::Utility::Debug()
       << "Starting Test : getCustomLightingRGBAObservation ";
```
mosra, line 338: Why this? :)

(I'm aware that if you get a crash, it won't print what test case it started, but that shouldn't mean you have to suffer like this ;) On my list of things to fix.)

src/tests/SimTest.cpp
```diff
@@ -152,33 +198,56 @@ void SimTest::reconfigure() {
...
+    auto agent = simulator.addAgent(agentConfig);
+
+    auto stateOrig = AgentState::create();
+    agent->getState(stateOrig);
+
+    simulator.reset();
+
+    auto stateFinal = AgentState::create();
+    agent->getState(stateFinal);
+    CORRADE_VERIFY(stateOrig->position == stateFinal->position);
```
mosra, line 235:
```suggestion
    CORRADE_COMPARE(stateOrig->position, stateFinal->position);
```

assuming those are Magnum vectors and not Eigen vectors (which don't have a debug output), same below

src/tests/SimTest.cpp
```diff
@@ -152,33 +198,56 @@ void SimTest::reconfigure() {
...
+  simulator_mm.reconfigure(cfg_mm);
+  CORRADE_VERIFY(pathfinder_mm == simulator_mm.getPathFinder());
+  SimulatorConfiguration cfg2_mm;
+  cfg2_mm.activeSceneName = skokloster;
+  simulator_mm.reconfigure(cfg2_mm);
+  CORRADE_VERIFY(pathfinder_mm != simulator_mm.getPathFinder());
 }
 
 void SimTest::reset() {
+  auto testReset = [&](Simulator& simulator) {
```
mosra, line 217: Because all test macros (which take care of extracting the function name) are inside the lambda, the function won't be marked correctly in the output. I'd do 

```cpp
void SimTest::reset() {
  CORRADE_VERIFY(true);

  auto testReset = [&](Simulator& simulator) {
```

to work around that :)

mosra (commented):
Just commenting on the test-related things.

src/tests/SimTest.cpp
jturner65: Oh yeah. Ty

jturner65: oops, those are eigen vecs

src/esp/bindings/SimBindings.cpp
```diff
@@ -113,8 +115,12 @@ void initSimBindings(py::module& m) {
           &Simulator::setActiveSceneDatasetName,
           R"(The currently active dataset being used.  Will attempt to load
             configuration files specified if does not already exist.)")
-      /* --- Physics functions --- */
       /* --- Template Manager accessors --- */
+      .def_property(
+          "metadata_mediator", &Simulator::getMetadataMediator,
+          &Simulator::setMetadataMediator, py::return_value_policy::copy,
```
Skylion007, line 121: Interesting, you really want to return a copy?

src/esp/bindings/SimBindings.cpp
```diff
@@ -74,6 +74,8 @@ void initSimBindings(py::module& m) {
 
   // ==== Simulator ====
   py::class_<Simulator, Simulator::ptr>(m, "Simulator")
+      // add constructor to build Simulator from existing MetadataMediator
+      .def(py::init<esp::metadata::MetadataMediator::ptr>())
```
Skylion007, line 78: Mind adding a pytest test case somewhere. It will serve as documentation too.

jturner65: working on it now

src/esp/bindings/SimBindings.cpp
jturner65: copy of smart pointer.  want to increment holder count

src/tests/SimTest.cpp
mosra: [Fixed.](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Tester.html#TestSuite-Tester-instanced). Sorry again, bad decisions from 2014 shouldn't still be preserved seven years later :D

src/tests/SimTest.cpp
mosra: `CORRADE_COMPARE(Mn::Vector3{stateOrig->position}, Mn::Vector3{stateFinal->position});` ? :sweat_smile:

src/esp/bindings/SimBindings.cpp
Skylion007: Please document in a comment.

src/esp/bindings/SimBindings.cpp
```diff
@@ -116,6 +116,8 @@ void initSimBindings(py::module& m) {
           R"(The currently active dataset being used.  Will attempt to load
             configuration files specified if does not already exist.)")
       /* --- Template Manager accessors --- */
+      // We wish a copy of the metadata mediator smart poitner so that we
```
Skylion007, line 119: typo:

habitat_sim/simulator.py
```diff
@@ -45,13 +46,15 @@ class Configuration(object):
...
-
+    :property metadata_mediator: (optional) The metadata mediator to build the simulator from
     Ties together a backend config, `sim_cfg` and a list of agent
     configurations `agents`.
     """
 
     sim_cfg: SimulatorConfiguration
     agents: List[AgentConfiguration]
+    # An existing Metadata Mediator can also be used to construct a SimulatorBackend
+    metadata_mediator: MetadataMediator = attr.ib(default=None)
```
Skylion007, line 57:
```suggestion
    metadata_mediator: Optional[MetadataMediator] = None
```

Skylion007: Can sim_cfg and metadata_mediator override each other? Do we print a warning if that's the case? Or assert that one is None? We may want to add a validator to the effect.

Skylion007 (changes requested):

habitat_sim/simulator.py
jturner65: Thanks!

src/esp/bindings/SimBindings.cpp
erikwijmans: That's what happens by default for smart pointers with pybind

habitat_sim/simulator.py
jturner65: > Can sim_cfg and metadata_mediator override each other? Do we print a warning if that's the case? Or assert that one is None? We may want to add a validator to the effect.

Yes, let's discuss this further in slack.

src/esp/bindings/SimBindings.cpp
jturner65: That's what I thought but I wasn't sure. I'll remove the decorator then.

aclegg3 (approved):
Looks good. 👍

habitat_sim/simulator.py
```diff
@@ -185,7 +188,7 @@ def reset_agent(self, agent_id: int) -> None:
 
     def _config_backend(self, config: Configuration) -> None:
         if not self._initialized:
-            super().__init__(config.sim_cfg)
+            super().__init__(config.sim_cfg, config.metadata_mediator)
```
Skylion007, line 191: We should still give a warning / info via the logger that the MetadataMediator is overriding stuff though. Is this handled on the C++ side?

jturner65: Nothing is being overridden.  The config.sim_cfg is the primary datasource.  The existing MM is passed in, and then config.sim_cfg is queried for the dataset and scene, just as if it wasn't passed at all.  The MM will get its dataset changed to match the dataset specified in the config.sim_cfg.  This is non-destructive, and any loaded dataset information the existing MM will be retained, and can be specified separately.

PR #1100, opened 2021-02-19
Culling for rigid and articulated objects

erikwijmans:
## Motivation and Context

Add culling for articulated and rigid objects.  Also moves away from using btMotionState for syncing positions as this gives us finer control over when the scene graph is dirtied and will be needed for rendering in a background thread.

Also compiles with broadwell vectorized instructions -- broadwell is quite old so unlikely that people don't have support for those but the newer wider AVX instructions don't improve perf.

## How Has This Been Tested

Andrew's interactive benchmark looks identical and I have spent a lot of time in the view watching objects and everything looks correct.


src/esp/physics/bullet/BulletArticulatedObject.cpp
```diff
@@ -285,8 +289,6 @@ void BulletArticulatedObject::setRootState(const Magnum::Matrix4& state) {
   btAlignedObjectArray<btVector3> scratch_m;
   btMultiBody_->forwardKinematics(scratch_q, scratch_m);
   btMultiBody_->updateCollisionObjectWorldTransforms(scratch_q, scratch_m);
-  // sync visual shapes
-  updateNodes();
```
aclegg3, line 289: We return object/link transforms from their SceneNodes. If we query a link transform after this setter but before the dynamics step then the queried transform will be outdated. However, I understand the value of deferring this update whenever possible. Maybe we want to either get those transforms directly from Bullet instead of our SceneGraph, or have a way to set state to dirty and update relevant nodes when the state queries are called.

src/esp/physics/bullet/BulletArticulatedObject.cpp
```diff
@@ -372,8 +374,6 @@ void BulletArticulatedObject::setPositions(
...
```
aclegg3, line 376: Same as previous comment.

src/esp/physics/bullet/BulletArticulatedObject.cpp
```diff
@@ -420,12 +420,11 @@ void BulletArticulatedObject::reset() {
   btMultiBody_->clearConstraintForces();
   btMultiBody_->clearVelocities();
   btMultiBody_->clearForcesAndTorques();
-  // sync visual shapes
-  updateNodes();
```
aclegg3, line 424: Here also.

src/esp/physics/bullet/BulletArticulatedObject.cpp
erikwijmans: Yeah, this is a good point, we should just query those from bullet

src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -36,8 +36,7 @@ namespace physics {
  * state with internal btRigidBody states
  */
 class BulletRigidObject : public BulletBase,
-                          public RigidObject,
-                          public Magnum::BulletIntegration::MotionState {
```
aclegg3, line 40: Same here. RigidObject state getters called after setters and before simulation step may need to manually update the node states since MotionState is not doing this automatically.

erikwijmans:
It's stable with Eric's change are added so we'll just turn it once those are in

PR #1104, opened 2021-02-22
[Potential Speedup] Set render buffer properly based on sensor type

bigbike:
### Impact
Potential rendering speedup.
**Before:**
The renderTarget will contain 1 rgb buffer, 1 depth texture, 1 objectId Buffer (color attachment) no matter what the sensor type is.

**After:**
The renderTarget will have different combination based on sensor type.
**For rgb sensor**: 1 rgb buffer + 1 depth buffer (no objectId buffer, and No depth texture). 
This is based on the fact that usually depth buffer is faster than the depth texture.

**For depth sensor**: 1 depth texture (No rgb buffer, no objectId buffer).
This will speed up the rendering based on the observation [here](https://github.com/facebookresearch/habitat-sim/pull/1011#discussion_r577347437).

**For ObjectId sensor** 1 objectId buffer + 1 depth buffer (No rgb buffer, and no depth texture).
Reason is the same as above.

More test results are coming.
Here is the viewer with rgb sensor:
Before:
before

After:
after

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -61,13 +61,13 @@ struct RenderTarget::Impl {
                               DepthShader::Flag::UnprojectExistingDepth);
     }
 
-    if (!(flags_ & Flag::NoRgbaBuffer)) {
+    if (flags_ & Flag::RgbaBuffer) {
       colorBuffer_.setStorage(Mn::GL::RenderbufferFormat::SRGB8Alpha8, size);
     }
-    if (!(flags_ & Flag::NoObjectIdBuffer)) {
+    if (flags_ & Flag::ObjectIdBuffer) {
       objectIdBuffer_.setStorage(Mn::GL::RenderbufferFormat::R32UI, size);
     }
-    if (!(flags_ & Flag::NoDepthTexture)) {
+    if (flags_ & Flag::DepthTexture) {
       depthRenderTexture_.setMinificationFilter(Mn::GL::SamplerFilter::Nearest)
           .setMagnificationFilter(Mn::GL::SamplerFilter::Nearest)
           .setWrapping(Mn::GL::SamplerWrapping::ClampToEdge)
@@ -80,24 +80,24 @@ struct RenderTarget::Impl {
     }
 
     framebuffer_ = Mn::GL::Framebuffer{{{}, size}};
-    if (!(flags_ & Flag::NoRgbaBuffer)) {
+    if (flags_ & Flag::RgbaBuffer) {
       framebuffer_.attachRenderbuffer(RgbaBuffer, colorBuffer_);
     }
-    if (!(flags_ & Flag::NoObjectIdBuffer)) {
+    if (flags_ & Flag::ObjectIdBuffer) {
       framebuffer_.attachRenderbuffer(ObjectIdBuffer, objectIdBuffer_);
     }
-    if (!(flags_ & Flag::NoDepthTexture)) {
+    if (flags_ & Flag::DepthTexture) {
       framebuffer_.attachTexture(Mn::GL::Framebuffer::BufferAttachment::Depth,
                                  depthRenderTexture_, 0);
     } else {
       framebuffer_.attachRenderbuffer(
           Mn::GL::Framebuffer::BufferAttachment::Depth, unprojectedDepth_);
     }
-    if (!(flags_ & Flag::NoRgbaBuffer)) {
+    if (flags_ & Flag::RgbaBuffer) {
       framebuffer_.mapForDraw({{0, RgbaBuffer}});
     }
 
-    if (!(flags_ & Flag::NoObjectIdBuffer)) {
+    if (flags_ & Flag::ObjectIdBuffer) {
       framebuffer_.mapForDraw({{1, ObjectIdBuffer}});
     }
     CORRADE_INTERNAL_ASSERT(
@@ -107,7 +107,7 @@ struct RenderTarget::Impl {
 
   void initDepthUnprojector() {
     CORRADE_ASSERT(
-        !(flags_ & Flag::NoDepthTexture),
+        flags_ & Flag::DepthTexture,
         "RenderTarget::Impl::initDepthUnprojector(): this render target "
         "was not created with depth texture enabled.", );
 
@@ -133,7 +133,7 @@ struct RenderTarget::Impl {
   void unprojectDepthGPU() {
     CORRADE_INTERNAL_ASSERT(depthShader_ != nullptr);
     CORRADE_ASSERT(
-        !(flags_ & Flag::NoDepthTexture),
+        flags_ & Flag::DepthTexture,
         "RenderTarget::Impl::unporojectDepthGPU(): this render target "
         "was not created with depth texture enabled.", );
     initDepthUnprojector();
@@ -147,10 +147,10 @@ struct RenderTarget::Impl {
 
   void renderEnter() {
     framebuffer_.clearDepth(1.0);
-    if (!(flags_ & Flag::NoRgbaBuffer)) {
+    if (flags_ & Flag::RgbaBuffer) {
       framebuffer_.clearColor(0, Mn::Color4{0, 0, 0, 1});
     }
-    if (!(flags_ & Flag::NoObjectIdBuffer)) {
+    if (flags_ & Flag::ObjectIdBuffer) {
       framebuffer_.clearColor(1, Mn::Vector4ui{});
     }
     framebuffer_.bind();
@@ -161,11 +161,8 @@ struct RenderTarget::Impl {
   void renderExit() {}
 
   void blitRgbaToDefault() {
-    CORRADE_ASSERT(!(flags_ & Flag::NoTextures),
-                   "RenderTarget::Impl::blitRgbaToDefault(): Simulator was "
-                   "initialized with requiresTextures = false", );
     CORRADE_ASSERT(
-        !(flags_ & Flag::NoRgbaBuffer),
+        flags_ & Flag::RgbaBuffer,
         "RenderTarget::Impl::blitRgbaToDefault(): this render target "
         "was not created with rgba render buffer enabled.", );
     CORRADE_ASSERT(
@@ -182,10 +179,7 @@ struct RenderTarget::Impl {
   }
 
   void readFrameRgba(const Mn::MutableImageView2D& view) {
-    CORRADE_ASSERT(!(flags_ & Flag::NoTextures),
-                   "RenderTarget::Impl::readFrameRgba(): Simulator was "
-                   "initialized with requiresTextures = false", );
-    CORRADE_ASSERT(!(flags_ & Flag::NoRgbaBuffer),
+    CORRADE_ASSERT(flags_ & Flag::RgbaBuffer,
                    "RenderTarget::Impl::readFrameRgba(): this render target "
                    "was not created with rgba render buffer enabled.", );
 
@@ -193,7 +187,7 @@ struct RenderTarget::Impl {
   }
 
   void readFrameDepth(const Mn::MutableImageView2D& view) {
-    CORRADE_ASSERT(!(flags_ & Flag::NoDepthTexture),
+    CORRADE_ASSERT(flags_ & Flag::DepthTexture,
                    "RenderTarget::Impl::readFrameDepth(): this render target "
                    "was not created with depth texture enabled.", );
     if (depthShader_) {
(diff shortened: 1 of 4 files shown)


src/esp/gfx/Renderer.h
```diff
@@ -16,6 +16,12 @@ namespace gfx {
 class Renderer {
  public:
   enum class Flag {
+    /**
+     * @brief No textures for the meshes.
```
mosra, line 20:
```suggestion
     * No textures for the meshes.
```

Complex reasons why is it that way, but with the `@brief` the docs won't show up in the output.

Same in `RenderTarget::Flag`.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -112,8 +147,12 @@ struct RenderTarget::Impl {
 
   void renderEnter() {
     framebuffer_.clearDepth(1.0);
-    framebuffer_.clearColor(0, Mn::Color4{0, 0, 0, 1});
-    framebuffer_.clearColor(1, Mn::Vector4ui{});
+    if (!(flags_ & Flag::NoRgbaBuffer)) {
+      framebuffer_.clearColor(0, Mn::Color4{0, 0, 0, 1});
+    }
+    if (!(flags_ & Flag::NoObjectIdBuffer)) {
```
mosra, line 153: TBH I don't like the double negation everywhere. This would be much easier to read and understand:

```suggestion
    if (flags_ & Flag::ObjectIdBuffer) {
```

src/esp/gfx/RenderTarget.h
```diff
@@ -24,6 +25,36 @@ namespace gfx {
...
+  enum class Flag {
+    /**
+     * @brief No textures for the meshes.
+     * Textures take a lot of GPU memory but they are not needed unless we are
+     * doing RGB rendering.
+     * Note: Do NOT set this flag when doing RGB rendering
+     * Currently it is just used to track whether or not @ref readFrameRgba,
+     * @ref blitRgbaToDefault, and @readFrameRgbaGPU are valid calls.
+     */
+    NoTextures = 1 << 0,
```
mosra, line 37: Doesn't `NoRgbaBuffer` imply `NoTextures`? I don't see a use case for one being set but the other not.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -122,28 +161,41 @@ struct RenderTarget::Impl {
   void renderExit() {}
 
   void blitRgbaToDefault() {
-    if (rendererFlags_ & Renderer::Flag::NoTextures)
-      throw std::runtime_error(
```
erikwijmans, line 126: The throws instead of asserts here was intentional.  I expect that this is something python uses could reasonably end up hitting and a throw is much nicer to work with than an assert.  The assert exits hard while the throw can be caught in pdb and more easily debugged.

bigbike: @erikwijmans : 
- Within the current C++ simulator, we do not built exception handling mechanism (no "catch" there) so such throws will not be handled properly in C++. That means it actually fails "silently" (asssuming no gdb, pdb or any debugger, just by system itself.)
- Assertion actually saves debugging time. When a critical error happens, it is meaningless to let the simulator continues running. So if it is to fail, let it fail fast. "It caught quite a lot of mistakes that would take significant time to discover and debug otherwise". quoted from @mosra . I had a lengthy conversation on this topic with him when I wrote the PBR shader.  Somehow I agree with him now on this point. Read the discussion [here](https://github.com/facebookresearch/habitat-sim/pull/840#discussion_r504624575).
- "pdb can catch the throw". Yes, it might be. But the question is why the user would use pdb, gdb or any debugger first? The answer is she will not use it until something is wrong with the system, but the she cannot tell exactly where goes wrong. It starts costing her time to debug already. On the other hand, if the "system exits hard" with the proper error information, she already knows the location, and properly the root cause.

The essence is that to handle the error in a gentle way or a hard way. In the current system, most of the places choose the later so I decide to do the same to make it consistent.

src/esp/gfx/RenderTarget.cpp
bigbike: @mosra : I think about it before I wrote.
The reason is if I use "negative" flags here, yes, it gives me, the developer hard time, and decrease the readability. However it is more friendly in the user's perspective. Because if the user is completely new to the system, she passes an empty flag in, the system by default will give her at least the correct result (It is just a bit slow in terms of performance).

Well, if you use "positive" flags, then the user MUST understand what needs to be set in order to get the correct results. If again, she passes an empty flag in, the system gives here nothing or quits due to the assertions.

I can change it though. Just share what I thought, and I knew you guys would complain it.

src/esp/gfx/RenderTarget.h
bigbike: OK.
First, I just inherited what was there before (the "NoTextures" item).
Second, "NoTextures" honestly does not lead to "NoRgbaBuffer" because in my opinion user can still render the mesh (with no textures) to a "rgbaBuffer" using the flat shaders.

src/esp/gfx/Renderer.h
bigbike: OK. May I ask what we shall do then.

src/esp/gfx/RenderTarget.cpp
erikwijmans: > Within the current C++ simulator, we do not build exception handling mechanism (no "catch" there) so such throws will not be handled properly in C++. That means it actually fails "silently" (asssuming no gdb, pdb or any debugger, just by system itself.)

There seems to be some miss-understanding here.  An uncaught exception will cause the program to exit immediately , not fail silently. What I mean by hard exit is that asserts cause the program to terminate in a way that pdb is not equipped to handle while in can handle cpp exceptions. 

I am by no means advocating that something fails silently.

bigbike: > terminate in a way that pdb is not equipped to handle while it can handle cpp exceptions

Ah, yes, without "try" and "catch", the program will exit. You are right. 

Now it behaves like the CORRADE_ASSERT (it will exit, and print out the error info). 
But CORRADE_ASSERT can give you the file name, line number where it fails, and the reasons. Now you already know where and why. May I understand why pdb is still needed?

erikwijmans: A reason, line number, and file name is nice but only really helpful if you already have an idea of what could go wrong -- awesome for devs, pretty good for power users, but not very useful for normal users.  With a throw you automatically call stack trace on the python side and debuggers are obviously nice to be able to use.

bigbike: OK. backtrace. I see your point.
Let us look at the case here. When an exception happens , it occurs at the rendering stage. So the backtrace will only give you a calling stack related to a serious of rendering calls. The genuine root cause, which is at the initialization stage, will not be shown in this backtrace at all. So I really do not see the benefits here. 

But you reached a very good point here, and we need to consider in the future development.

src/esp/gfx/RenderTarget.h
bigbike: @erikwijmans : May I have your input? Actually this is a question I had for quite a long time.
If the textures are not loaded, the mesh can still be rendered using flat shader, and thus a RgbaBuffer is still needed (and can even be blit to the default FBO)?

erikwijmans: If there are no textures it doesn't make sense to render rgb IMO -- which is why there is that runtime_error if you try to read the rgb buffer when there are no textures.  Currently NoTextures is only set when there isn't an RGB camera on any agent.

Skylion007:
@bigbike merge from master to clear these CI issues.

src/esp/gfx/Renderer.h
mosra: Just don't use `@brief` for enum values, that's it :)

src/esp/gfx/RenderTarget.cpp
mosra: > Because if the user is completely new to the system, she passes an empty flag in

That was a second part of this thought -- what about providing default `flags` parameter in the constructor, with all set? (That's what I usually do, provide a conservative "always correct" option by default, and let the users override that if they want. But no double negations :))

src/esp/gfx/RenderTarget.cpp
mosra: What I do for Magnum python bindings is replicating asserts *outside* of the function and turning them into Python exceptions. So if a C++ developer "holds the function wrong", they get an assert, but if a Python dev does it, the bindings do all needed checks first and raise an exception *before* the actual C++ function gets called.

But I admit this is basically double the work :) (OTOH there's [CORRADE_NO_ASSERT](https://doc.magnum.graphics/corrade/Assert_8h.html#a588d67a03f847239f809eda035d73730) so, in case of magnum at least, the assert needs to be duplicated for python anyway.)

eundersander: My two cents: whatever we decide here (assert or exception), we should wrap it in a macro like HABITAT_RUNTIME_ERROR so we can make this decision in one place and document it.

The HABITAT_RUNTIME_ERROR macro is also a place to customize the behavior for other builds, e.g. the Javascript build.

bigbike:
Updated. Please take another look.

mosra (commented):
Looks good to me, there's just the unresolved conversation about asserts vs exceptions left.

src/esp/gfx/Renderer.cpp
```diff
@@ -63,9 +63,33 @@ struct Renderer::Impl {
           DepthShader::Flag::UnprojectExistingDepth);
     }
 
+    RenderTarget::Flags renderTargetFlags_ = {};
+    switch (sensor.specification()->sensorType) {
+      case sensor::SensorType::Color:
```
erikwijmans, line 68: Can you add an assert here that `NoTextures` isn't set.  Right now a user could make a RGBA RenderTarget when there aren't textures.

src/esp/gfx/Renderer.cpp
```diff
@@ -63,9 +63,33 @@ struct Renderer::Impl {
...
+        break;
+
+      default:
+        // I need this default, since sensor type list is long, and without
+        // default clang-tidy will complain
+        break;
+    }
+    if (flags_ & Flag::NoTextures) {
+      // it means no rgba render buffer
+      renderTargetFlags_ &= ~RenderTarget::Flag::RgbaBuffer;
```
bigbike, line 87: @erikwijmans : not sure if you see this if clause here, if textures are not loaded, if would not create a rgbaBuffer. And if the user tries to get the color observation it would fail within the rendering call. 

But I would change it here to make an assertion.

src/esp/gfx/Renderer.cpp
bigbike: @erikwijmans : Just updated the code to address your concern here. Take a look.

src/esp/gfx/RenderTarget.cpp
bigbike: @erikwijmans : We heard you and we are actively seeking solutions to address your concerns and needs (Please see the docs in the slack channel for more details). Thanks.

bigbike:
Dear reviewers,

Please take a look. If it looks good to you, hope you can stamp it.
Thank you.

src/esp/gfx/Renderer.cpp
mosra: Though ... what if the intent is to render just vertex colors? That can be thought of as some kind of an optimization as well, ignoring textures and using vertex colors instead. Or is that not the intent of this functionality?

src/esp/gfx/Renderer.cpp
erikwijmans: Makes sense to fail sooner IMO.  The renderer wasn't looking at the sensor spec beforehand so it wasn't obvious that we could have things fail here, but it should have in hindsight

src/esp/gfx/Renderer.cpp
erikwijmans: Currently the intent of this is to not load textures when the agent doesn't have and rgb sensor.  AFAIK, we don't have any assets that have both textures and vertex coloring so I think tying theses makes sense, but this is a good idea that we should keep in mind for the future.

bigbike: In computer graphics community, we played with the meshes without textures quite often, e.g., the Stanford bunny. 

But I guess it would not be the case here. :)

bigbike:
Can we ship it like this now?

erikwijmans (approved):
Ship it!

src/esp/gfx/Renderer.cpp
erikwijmans: Agreed that is a legitimate use for rgb rendering without textures in general :)

bigbike:
@erikwijmans . Great! Thanks.

PR #1106, opened 2021-02-22
[AO Prototype] Background rendering thread

erikwijmans:
## Motivation and Context

Background rendering thread to enable overlap of physics and rendering.  This means that sim(a_t) advances the world state to s_{t+1} but returns o_t, not o_{t+1} as it normally would.

## How Has This Been Tested

Things train and rendering time is sub-millisecond.

### DIFF
--- a/src/esp/gfx/RenderCamera.h
+++ b/src/esp/gfx/RenderCamera.h
@@ -43,6 +43,11 @@ class RenderCamera : public MagnumCamera {
   typedef Corrade::Containers::EnumSet<Flag> Flags;
   CORRADE_ENUMSET_FRIEND_OPERATORS(Flags)
 
+  typedef std::vector<
+      std::pair<std::reference_wrapper<Magnum::SceneGraph::Drawable3D>,
+                Magnum::Matrix4>>
+      DrawableTransforms;
+
   /**
    * @brief Constructor
    * @param node, the scene node to which the camera is attached
@@ -151,6 +156,8 @@ class RenderCamera : public MagnumCamera {
    */
   uint32_t draw(MagnumDrawableGroup& drawables, Flags flags = {});
 
+  uint32_t draw(DrawableTransforms& drawableTransforms, Flags flags = {});
+
   /**
    * @brief performs the frustum culling
    * @param drawableTransforms, a vector of pairs of Drawable3D object and its
@@ -161,9 +168,7 @@ class RenderCamera : public MagnumCamera {
    * The preferred way is to enable the frustum culling by calling @ref
    * setFrustumCullingEnabled and then call @ref draw
    */
-  size_t cull(std::vector<
-              std::pair<std::reference_wrapper<Magnum::SceneGraph::Drawable3D>,
-                        Magnum::Matrix4>>& drawableTransforms);
+  size_t cull(DrawableTransforms& drawableTransforms);
 
   /**
    * @brief Cull Drawables for SceneNodes which are not OBJECT type.
@@ -172,10 +177,10 @@ class RenderCamera : public MagnumCamera {
    * absolute transformation
    * @return the number of drawables that are not culled
    */
-  size_t removeNonObjects(
-      std::vector<
-          std::pair<std::reference_wrapper<Magnum::SceneGraph::Drawable3D>,
-                    Magnum::Matrix4>>& drawableTransforms);
+  size_t removeNonObjects(DrawableTransforms& drawableTransforms);
+
+  size_t filterTransforms(DrawableTransforms& drawableTransforms,
+                          Flags flags = {});
 
   /**
    * @brief if the "immediate" following rendering pass is to use drawable ids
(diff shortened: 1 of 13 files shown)


Skylion007 (approved):
The Python part looks good. 

Not sure about the C++ multirthreading. The CBullet API would have also given this to use for free if we used SharedMemory.

Skylion007:
@erikwijmans Also the clang-tidy config in this branch is still broken, we should cherry pick that to make sure we aren't breaking anything performance wise. 👍

erikwijmans:
> The CBullet API would have also given this to use for free if we used SharedMemory. 

It wouldn't have.  There are multiple physics steps that happen during the render (for stability) and there is some python code that happens in each one.  We could reimplement that loop and code in cpp and put it into a background thread but that isn't free.

src/esp/gfx/Renderer.cpp
```diff
@@ -27,16 +34,236 @@ namespace Mn = Magnum;
...
+      auto& job = jobs_[i];
+      sensor::VisualSensor& sensor = std::get<0>(job);
+      scene::SceneGraph& sg = std::get<1>(job);
+      RenderCamera::Flags flags = std::get<3>(job);
+
+      auto* camera = sensor.getRenderCamera();
+      jobTransforms[i].reserve(sg.getDrawableGroups().size());
+      for (auto& it : sg.getDrawableGroups()) {
+        it.second.prepareForDraw(*camera);
+        auto transforms = camera->drawableTransformations(it.second);
```
eundersander, line 124: So you are reading sceneGraph node transforms on this background thread. Isn't Bullet potentially writing those transforms on the main thread? This writing happens via MotionState.cpp MotionState::setWorldTransform, which Bullet calls internally because we provide this motionState when we construct a Bullet rigid body in BulletRigidObject.

eundersander (commented):
I added a comment because I want to understand how Bullet is updating sceneGraph node transforms while the renderer can be reading them on a background thread.

I'm somewhat familiar with STL multithreading API, but not enough to offer a good review of what you've written here, unless you want to go back and add a bunch of comments. At a glance, I would have expected we wouldn't need to write spin loops and other such low-level code; I'd expect we could instead use higher-level STL multithreading APIs. All of this is really only a concern if/when we get to merging this to master.

src/esp/gfx/Renderer.cpp
erikwijmans: I probably should have put that change in this PR, but I stuck that in this one: https://github.com/facebookresearch/habitat-sim/pull/1100 (made testing easier).  That added a defer feature to MotionState so that Bullet won't update the SceneGraph until the call to `updateNodes` which is why where is ` renderer_->waitSG()` before that

aclegg3 (commented):
Overall looks good. Would like to see passing tests before we merge.

erikwijmans:
> Overall looks good. Would like to see passing tests before we merge.

Looks like JS was failing because it doesn't support those assembly instructions.

bigbike:
This is great -- the prototype of the multithreaded renderer, which decouples the sim-thread from the rendering thread so they can do the job in parallel. Finally it is implemented. You are fast @erikwijmans.
 
@aclegg3 and I discussed it a couple of times before. Something we can consider in the future.
- a simple noise model (or provided by the user) to simulate the delay between observation request (time step t) and the time it gets the observation (time step t + k). In your current prototype for example, this model is constant: k = 1.
- Try intel TBB. Often it can beat your code written with pthread, mutex. etc.

erikwijmans:
The atomic_wait example implementation from here compiles on cpp14: https://github.com/ogiroux/atomic_wait. I did a test swap to that it doesn't matter.  Since we are just hitting the mutex once per frame (instead of hundred of time per frame) and it is only the background thread, the time of it doesn't matter.

src/esp/gfx/RenderCamera.h
```diff
@@ -43,6 +43,11 @@ class RenderCamera : public MagnumCamera {
   typedef Corrade::Containers::EnumSet<Flag> Flags;
   CORRADE_ENUMSET_FRIEND_OPERATORS(Flags)
 
+  typedef std::vector<
+      std::pair<std::reference_wrapper<Magnum::SceneGraph::Drawable3D>,
+                Magnum::Matrix4>>
+      DrawableTransforms;
```
jturner65, line 49: Perhaps apply this typedef to [CullingTest.cpp](https://github.com/facebookresearch/habitat-sim/blob/ae2faa8f6d726dd73c4fa5a7b1880b61b2bfab24/src/tests/CullingTest.cpp#L180) as well?

jturner65 (approved):
Very impressive; One thing, though - I personally would like to see some comments, at least in BackgroundRenderThread struct, it would be good if this was extensively commented to facilitate future support/extension of the functionality (if this is ever going to be merged into master).

erikwijmans:
Yeah, this needs some work before it would go into master.  I removed the spin lock since that is the largest culprit (that is only fine here because of how balanced all the timing is isn't a good idea in general)

PR #1107, opened 2021-02-23
--[Bug Fix]Remake activateCollisionIsland with bitset

jturner65:
This PR changes BulletRigidObject::activateCollisionIsland() to use a Magnum::BoolVector instead of a std::set.  The std::set.insert function can throw an exception due to allocation, but BulletRigidObject::activateCollisionIsland is called from the BulletRigidObject destructor, which cannot allow exceptions.  Instead of bypassing a useful clang-tidy check, this uses a Magnum::BoolVector that does not throw exceptions.

Magnum::BoolVector also has a smaller memory footprint per element ([an integer set requires approximately 32 bytes per 32 bit element](https://lemire.me/blog/2016/09/15/the-memory-usage-of-stl-containers-can-be-surprising/)); the BoolVector<65536> requires 8kb storage space, which is equivalent to a std::set<int> containing approx 200 elements.

A potential issue is if we have more than 65536 island ids, this will overflow unless the template param size is increased.

## How Has This Been Tested
c++ and python tests

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletRigidObject.cpp
+++ b/src/esp/physics/bullet/BulletRigidObject.cpp
@@ -43,9 +43,6 @@ BulletRigidObject::BulletRigidObject(
       RigidObject(rigidBodyNode, objectId, resMgr),
       MotionState(*rigidBodyNode) {}
 
-// clang-tidy assumes that the set insert in activateCollisionIsland() may
-// potentially throw an exception
-// NOLINTNEXTLINE(bugprone-exception-escape)
 BulletRigidObject::~BulletRigidObject() {
   if (!isActive()) {
     // This object may be supporting other sleeping objects, so wake them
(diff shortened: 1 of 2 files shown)


src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -459,29 +456,30 @@ void BulletRigidObject::activateCollisionIsland() {
   // first query overlapping pairs of the current object from the most recent
   // broadphase to collect relevant simulation islands.
 
-  std::set<int> overlappingSimulationIslands = {thisColObj->getIslandTag()};
+  // each index represents an island tag present - default is -1, so add one.
+  std::bitset<65536> overlappingSimIslands;
```
Skylion007, line 460: Magic Number! Can we derive it somehow? Or at the very least put some detailed comments on why this number was chosen (if arbitrary).

Skylion007 (approved):
Neat! Just one minor complaint about the magic number.

src/esp/physics/bullet/BulletRigidObject.cpp
jturner65: Yeah, I hate it too.  It's a magic # in bullet - not even a constant there that I could harvest. :(

mosra: Arbitrary :) [64k should be enough for everybody!](https://quoteinvestigator.com/2011/09/08/640k-enough/)

I'd write `64*1024` personally instead and add a note that this is just 8kB of memory and should be sufficient even for huge scenes (assuming that's true).

jturner65: [Here it is in bullet btCollisionObject.cpp constructor](https://github.com/bulletphysics/bullet3/blob/2a0f811c3bb3c65810eb2c095724b20b2f8b8dd4/src/BulletCollision/CollisionDispatch/btCollisionObject.cpp#L31)

jturner65: Oh, you mean the size of the bitset.  I'll add a comment saying that we expect to never have more than 65536 collision islands.

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -459,29 +456,30 @@ void BulletRigidObject::activateCollisionIsland() {
...
+                            1] = true;
     } else if (pairCache.at(i).m_pProxy1->m_clientObject == thisColObj) {
-      overlappingSimulationIslands.insert(
-          static_cast<btCollisionObject*>(
-              pairCache.at(i).m_pProxy0->m_clientObject)
-              ->getIslandTag());
+      overlappingSimIslands[static_cast<btCollisionObject*>(
+                                pairCache.at(i).m_pProxy0->m_clientObject)
+                                ->getIslandTag() +
+                            1] = true;
```
mosra, line 474: My obsession with not using unnecessarily heavy APIs probably shows here, but anyway -- using `.set()` might be nicer to the compiler (and debug build speed) as it doesn't have to create a temporary "reference" object that performs the bit assignment.

The `[]` query below is fine, tho.

mosra (commented):
> 2048 bytes

Wait, how did you get to that number? :) For 64k bits it should be 8 kB, I think.

src/esp/physics/bullet/BulletRigidObject.cpp
jturner65: Tried set first, apparently it will potentially throw an exception if out of range 👯

jturner65: [] just pukes

mosra: ugh STL

alternatively you could use my [BoolVector](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1BoolVector.html), that one isn't riddled with exceptions :))

aclegg3 (approved):
Cool. When tests pass, let's ship it.

PR #1115, opened 2021-03-01
[AO Prototype] Fix leak in BG renderer

erikwijmans:
## Motivation and Context

Something somewhere is leaking GPU memory because we create and destroy the background render thread all the time.  Ideally we'd fix that, but we currently have no idea how.  This is a workaround to not destroy the context or renderer on a close call.

Just putting this here for visibility and to note the caveats: If you change flags related to the renderer, those won't be propagated.  This is fine ATM and fixable in the long run, but not needed currently.

## How Has This Been Tested

Ran the training script for a long time with lots of scene swaps and no leaks!

Planning to land if the CI is happy

## Types of changes

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -117,7 +117,7 @@ class Simulator(SimulatorBackend):
         self._sanitize_config(self.config)
         self.__set_from_config(self.config)
 
-    def close(self) -> None:
+    def close(self, destroy: bool = False) -> None:
         if self.renderer is not None:
             self.renderer.acquire_gl_context()
 
@@ -136,7 +136,7 @@ class Simulator(SimulatorBackend):
 
         self.__last_state.clear()
 
-        super().close()
+        super().close(destroy)
 
     def __enter__(self) -> "Simulator":
         return self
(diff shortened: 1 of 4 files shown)


aclegg3 (approved):
Thanks for the fix. Should do for now. 👍 
I'll address the clang-tidy shortly.

jturner65:
@erikwijmans  If a sim is built with a renderer and then reconfigure in simulator is called with `config_.createRenderer = false`, this will still maintain a renderer and context, right?  Do we want to kill them in that case?

mosra:
Hmm, does it still leak if you do the change only for the `context_` and not to `renderer_`?

erikwijmans:
> Hmm, does it still leak if you do the change only for the `context_` and not to `renderer_`?

Yeah, it does.  So there is some state the binding the context in the BG thread creates and that is just never freed

erikwijmans:
> @erikwijmans If a sim is built with a renderer and then reconfigure in simulator is called with `config_.createRenderer = false`, this will still maintain a renderer and context, right? Do we want to kill them in that case?

Yeah, this isn't perfect currently, but this unblocks experiments and I won't have the cycles to think through making this perfect for a while.

PR #1119, opened 2021-03-01
Bullet perf fixes and better Bullet debugging

eundersander:
## Motivation and Context

See individual commit messages.

There's some code here that should really be cleaned up. This isn't suitable to merge to Hab-sim master as-is.

## How Has This Been Tested

I tested this a couple weeks ago in my perf branch. Unfortunately, I'm unable to run/test latest pvizplan play.py with latest AO branch due to some error about missing sensor config.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/CollisionGroupHelper.cpp
+++ b/src/esp/physics/CollisionGroupHelper.cpp
@@ -26,36 +26,45 @@ int CollisionGroupHelper::getMaskForGroup(CollisionGroup group) {
   switch (group) {
     case CollisionGroup::Default:
       // everything
-      return -1;
+      return -1 & ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::Static:
-      // everything but kinematic
-      return int(-1) & ~int(CollisionGroup::Kinematic);
+      // everything but [kinematic, static]
+      return int(-1) & ~int(CollisionGroup::Kinematic) &
+             ~int(CollisionGroup::Static) & ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::Kinematic:
-      // everything but static
-      return int(-1) & ~int(CollisionGroup::Static);
+      // everything but [kinematic, static]
+      return int(-1) & ~int(CollisionGroup::Kinematic) &
+             ~int(CollisionGroup::Static) & ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::FreeObject:
       // everything
-      return int(-1);
+      return int(-1) & ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::GraspedObject:
       // everything but robot
-      return int(-1) & ~int(CollisionGroup::Robot);
+      return int(-1) & ~int(CollisionGroup::Robot) &
+             ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::Robot:
       // everything but grasped object
-      return int(-1) & ~int(CollisionGroup::GraspedObject);
+      return int(-1) & ~int(CollisionGroup::GraspedObject) &
+             ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::EeMargin:
       // everything but grasped object or the robot
-      return int(-1) &
-             ~(int(CollisionGroup::GraspedObject) | int(CollisionGroup::Robot));
+      return int(-1) & ~int(CollisionGroup::GraspedObject) &
+             ~int(CollisionGroup::Robot) & ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::SelObj:
       // everything but the ee margin
-      return int(-1) & ~(int(CollisionGroup::EeMargin));
+      return int(-1) & ~(int(CollisionGroup::EeMargin)) &
+             ~int(CollisionGroup::Noncollidable);
+
+    case CollisionGroup::Noncollidable:
+      // nothing
+      return 0;
 
     default:
       CORRADE_ASSERT(
--- a/src/esp/physics/bullet/BulletArticulatedObject.cpp
+++ b/src/esp/physics/bullet/BulletArticulatedObject.cpp
@@ -34,6 +34,13 @@ BulletArticulatedObject::~BulletArticulatedObject() {
   // Corrade::Utility::Debug() << "deconstructing ~BulletArticulatedObject";
   if (motionType_ != MotionType::KINEMATIC) {
     // KINEMATIC objects have already been removed from the world.
+
+    if (bFixedObjectRigidBody_) {
+      bWorld_->removeRigidBody(bFixedObjectRigidBody_.get());
+      bFixedObjectRigidBody_ = nullptr;
+      bFixedObjectShape_ = nullptr;
+    }
+
     bWorld_->removeMultiBody(btMultiBody_.get());
   }
   // remove link collision objects from world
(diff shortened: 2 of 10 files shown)


src/esp/physics/bullet/BulletArticulatedObject.cpp
```diff
@@ -132,6 +139,47 @@ bool BulletArticulatedObject::initializeFromURDF(
...
+      if (col->getBroadphaseHandle()->m_collisionFilterGroup ==
+          int(CollisionGroup::Noncollidable)) {
+        bFixedObjectShape_.get()->addChildShape(col->getWorldTransform(),
+                                                col->getCollisionShape());
+      }
+
+      for (int m = 0; m < mb->getNumLinks(); m++) {
+        btMultiBodyLinkCollider* col = mb->getLink(m).m_collider;
+        if (col) {
+          if (col->getBroadphaseHandle()->m_collisionFilterGroup ==
```
eundersander, line 159: It would be nice to remove the link altogether here, but that's a more complicated change. My use of CollisionGroup::Noncollidable was the least-intrusive way to disable the link.

src/esp/physics/bullet/BulletDebugManager.cpp
```diff
@@ -114,5 +132,55 @@ std::string BulletDebugManager::getStepCollisionSummary(
...
+          s << "  " << name1 << std::endl;
+          foundAny = true;
+        }
+      }
+
+      if (!foundAny) {
+        s << "  (none)" << std::endl;
+      }
+
+      LOG(INFO) << s.str();
```
eundersander, line 177: This is pretty sloppy. Conceptually, all this verbose output should just be appended to s2 and returned as part of the return-value string. Instead, I'm logging directly here, because I hit an issue where printing a very long string to stdout wasn't working very well with VS Code's terminal.

src/esp/physics/CollisionGroupHelper.cpp
```diff
@@ -26,36 +26,45 @@ int CollisionGroupHelper::getMaskForGroup(CollisionGroup group) {
...
+      // everything but [kinematic, static]
+      return int(-1) & ~int(CollisionGroup::Kinematic) &
+             ~int(CollisionGroup::Static) & ~int(CollisionGroup::Noncollidable);
 
     case CollisionGroup::Kinematic:
-      // everything but static
-      return int(-1) & ~int(CollisionGroup::Static);
+      // everything but [kinematic, static]
+      return int(-1) & ~int(CollisionGroup::Kinematic) &
+             ~int(CollisionGroup::Static) & ~int(CollisionGroup::Noncollidable);
```
jturner65, line 39: Should kinematic objects not collide with other kinematic objects? Otherwise, why have both kinematic group and static group?

src/esp/physics/bullet/BulletURDFImporter.cpp
```diff
@@ -646,9 +654,17 @@ Mn::Matrix4 BulletURDFImporter::ConvertURDF2BulletInternal(
...
+      int collisionFilterMask = CollisionGroupHelper::getMaskForGroup(
+          CollisionGroup(collisionFilterGroup));
+// We don't like overriding the mask in the URDF; we disable support for this.
+// We prefer to only override the group, while still using getMaskForGroup
+// (above) for mask computation.
+#if 0
       if (collisionFlags & io::URDF::HAS_COLLISION_MASK) {
         collisionFilterMask = colMask;
       }
+#endif
```
jturner65, line 667: I can envision experimental situations where one might want to manipulate the potential collisions of a single linkage in an explicit manner without changing its group membership.  I think having the mask override being supported in the xml would be useful and much less arcane than determining the proper group that has the desired target group membership.  How this mask input is consumed could perhaps be decided by another collision tag-level input (say "mask_type" : additive, subtractive, override or something like that)

src/esp/physics/CollisionGroupHelper.cpp
eundersander: kinematic objects shouldn't collide with other kinematic objects, because neither require a collision response.

Since they are conceptually different (kinematic objects move; static objects don't), I'm inclined to keep them as separate groups. I can imagine a day when we want selective collision, for example, a raycast that only hits static objects or only kinematic.

src/esp/physics/bullet/BulletURDFImporter.cpp
eundersander: First, beware that the URDF importer always seems to set both HAS_COLLISION_GROUP and HAS_COLLISION_MASK, even if you've only specified one in the URDF. So that would need to change if we're considering re-enabling this code here.

>much less arcane than determining the proper group
I don't consider this arcane. The group is conceptual and usually well-named.

eundersander: >I can envision experimental situations where one might want to manipulate the potential collisions of a single linkage in an explicit manner without changing its group membership. 

Can you give an explicit example?

jturner65 (approved):
LGTM, let's get this updated!

src/esp/physics/bullet/BulletURDFImporter.cpp
jturner65: > I don't consider this arcane. The group is conceptual and usually well-named.
The names of the groups don't reflect what collisions they allow.  They also don't come close to covering all the possible masking groups. For instance, using the example above, what if i want to record kinematic-kinematic collision point locations? Is there a group that allows for all but static + noncollidable?  Kinematic may not require a response, but the information for the location could be very useful.

> Can you give an explicit example?
Record collision point locations of a kinematically-posed linkage component to see if/where it is hitting the stage.    


I guess my point in bringing this up is that we can hardcode masks to 32 common, common-sense, collision groups describing the majority of feasible behavior, but we should design around supporting research-driven situations where "uncommon" behavior may be desirable, particularly for diagnostic purposes.  

Why would making this functionality available to the user be undesirable?

src/esp/physics/CollisionGroupHelper.cpp
jturner65: I think being able to have collision location information is useful on demand regardless of the nature of the object, but i agree that it would be sub-optimal as default behavior for k-k since, as you said, no force response is appropriate.

src/esp/physics/bullet/BulletURDFImporter.cpp
eundersander: This is probably better discussed in-person just for time's sake. I am just giving my two cents here; ultimately @aclegg3  owns the physics API. Also, the AO branch is still just a branch so what we merge here doesn't represent a final decision.

eundersander:
I was unable to test this with latest everything due to the sensorSpec error. I also see that lots of CI tests are failing. It's hard for me to know the status of this branch. I suggest @aclegg3 click merge on this whenever he's comfortable.

src/esp/physics/bullet/BulletArticulatedObject.cpp
```diff
@@ -132,6 +139,47 @@ bool BulletArticulatedObject::initializeFromURDF(
...
+
+      if (bFixedObjectShape_.get()->getNumChildShapes()) {
+        btRigidBody::btRigidBodyConstructionInfo info =
+            btRigidBody::btRigidBodyConstructionInfo(0.f, nullptr,
+                                                     bFixedObjectShape_.get());
+        bFixedObjectRigidBody_ = std::make_unique<btRigidBody>(info);
+        BulletDebugManager::get().mapCollisionObjectTo(
+            bFixedObjectRigidBody_.get(),
+            "URDF, " + u2b.getModel()->m_name + ", fixed body");
+        bWorld_->addRigidBody(
```
aclegg3, line 175: We'll also want to add this new fixed base object to `collisionObjToObjIds_` map for collision queries. It should then be removed in the destructor.

aclegg3 (commented):
Looks good overall. A couple of comments.

aclegg3:
> Looks good overall. A couple of comments.

I'll add tweaks in a follow-up.

PR #1123, opened 2021-03-03
Update Magnum to a version that's definitely not slower than before

mosra:
## Motivation and Context

Some minor breaking changes happened recently so I'm submitting a dependency-updating PR to avoid others potentially hitting unexpected issues down the line. A very rough overview of what this brings (though a subset of these might have been already pulled in with previous updates by other people without being properly advertised, so just FYI):

- Corrade got a bunch of crazy new containers that are being used by the Vulkan backed, namely [ArrayTuple](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1ArrayTuple.html) for coalescing allocations from multiple arrays and [BigEnumSet](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BigEnumSet.html) fo enum sets that need more than 64 values
- The [StringView](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BasicStringView.html) container now supersedes C++17 `std::string_view` in terms of usability and flexibility and is great for allocation-less string processing if that sounds useful somewhere
- Memory use and performance optimization in `Platform::GLContext` creation thanks to removing all STL use (no `std::vector`, `std::string` or `std::sort()` there anymore; this is just "backporting" what was already done for Vulkan right from the start)
- Reduced the amount of internal allocations in `Platform::GLContext` from ~9237 to about 3, which may result in faster startup times
- There's *finally* a [non-annoying programmatic way](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Context_1_1Configuration.html) to suppress the engine startup log (or configure driver workarounds, GPU validation, ...) if you'd ever need that
- Removed one function pointer indirection from *all* `setUniform()` calls (except for Mac; sorry Mac users), might result in minor frame time improvements
- Various fixes for multithreaded OpenGL contexts, in particular to avoid race conditions on default framebuffer state and APIs allowing to correctly release a context when moving it to another thread (needed by @erikwijmans)
- New `--magnum-gpu-validation no-error` CLI and environment option that creates GL contexts without error reporting and as such *may* result in faster execution on certain drivers at the risk of instability when actual GL errors happen (well, except for Mac again; sorry² Mac users)
- 16-bit-per-channel PNG image import in `StbImageImporter`
- python bindings don't use `from` as a method name anymore, that was a silly thing from me to do (requested by @aclegg3 a decade ago, but seems the change didn't land here yet)
- and a ton of Vulkan additions, as usual

I have another batch of breakages scheduled, especially related to shaders, expect a PR again at the end of March.

## How Has This Been Tested

It was a pain but my CIs pass. ~~Yours?~~ Conda build fails on `hdf5.h` not found, I am pretty sure I didn't break that :scream: 

## Types of changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,7 @@ commands:
                   zip \
                   libbullet-dev\
                   lcov\
+                  libhdf5-dev \
                   unzip || true
       - run:
           name: Install Headless Chrome dependencies


erikwijmans:
h5py why: https://github.com/facebookresearch/habitat-lab/pull/607

aclegg3 (approved):

Skylion007 (approved):

bigbike (approved):

PR #1118, opened 2021-03-01
[Sensor] SensorSuite in each SceneNode

vauduong:
## Motivation and Context
Due to new functionality from https://github.com/facebookresearch/habitat-sim/pull/1064, there needs to be more functionality to keep track of sensors.

In this PR:
- Change the sensors_ in SensorSuite from std::map<std::string, Sensor::ptr> to std::map<std::string, reference_wrapper<Sensor>>
- Make SensorSuite inherited from Magnum::SceneGraph::AbstractFeature3D
- Add raw pointers nodeSensorSuite and subtreeSensorSuite in SceneNode. These features will be released upon SceneNode's deletion.
- In the Sensor constructor, add this sensor to its node's nodeSensorSuite and subtreeSensorSuite. Traverse all the way up to the root node, add this sensor to the subtreeSensorSuite of every ancestor node. Remove the Sensor from  subtreeSensorSuite of every ancestor node in the destructor.
- Overload setParent so that ancestor subtrees are updated 
- Write bindings for getter functions and SensorFactory methods
- Write robust unit tests for creating sensors, deleting sensors, and setting parents

Copy of SensorCameraWeeklyUpdates 8 pptx

## How Has This Been Tested

Build and test, write additional unit tests

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1131, opened 2021-03-07
[Sensor] Fisheye Sensor  

bigbike:
## Motivation and Context
Introduce a brand-new sensor type, the fisheye sensor to the system. 
In this diff:
- developed a flexible framework that can support different fisheye sensor models.
- implemented one fisheye sensor model: the double sphere fisheye camera model. (see here: https://vision.in.tum.de/research/vslam/double-sphere) 
- (Temporarily) adjusted a couple of sensor specs. My intern will write another diff to handle them. Before that happens, these changes will be in this PR temporarily. 

## Types of changes

## Checklist


dhruvbatra:
Cool! Can we get a couple of visuals to illustrate what the images look like?

Skylion007:
Can we also get some benchmarks between doing this a PyTorch observation transformer? Would be cool to see how much of a speed up we get.

src/esp/gfx/FisheyeShader.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "FisheyeShader.h"
+
+#include <Corrade/Utility/Assert.h>
+#include <Corrade/Utility/FormatStl.h>
+#include <Corrade/Utility/Resource.h>
+#include <Magnum/GL/Shader.h>
+#include <Magnum/GL/Texture.h>
+#include <Magnum/GL/Version.h>
+
+#include <sstream>
```
mosra, line 13: Not needed here (sorry, I have a special sense developed for these kinds of includes :sweat_smile:)

src/esp/gfx/FisheyeShader.h
```diff
@@ -0,0 +1,118 @@
...
+   * the constructor
+   */
+  virtual void cacheUniforms() = 0;
+
+  /**
+   * @brief set texture binding points in the shader
+   * NOTE: Subclass must implement this function, and call it at the end of
+   * the constructor
+   */
+  virtual void setTextureBindingPoints() = 0;
```
mosra, line 113: I don't understand why these need to exist at all... the base class has no way of invoking these from its own constructor anyway. To me it feels like `DoubleSphereCameraShader` should just put the contents of these functions inline into the constructor, like all other shaders do.

src/esp/sensor/FisheyeSensor.cpp
```diff
@@ -0,0 +1,189 @@
...
+}
+
+void FisheyeSensorDoubleSphereSpec::sanityCheck() {
+  FisheyeSensorSpec::sanityCheck();
+  CORRADE_ASSERT(fisheyeModelType == FisheyeSensorModelType::DoubleSphere,
+                 "FisheyeSensorDoubleSphereSpec::sanityCheck(): fisheye model "
+                 "type is wrong. It should be DoubleSphere.", );
+  CORRADE_ASSERT(alpha >= 0 && alpha <= 1.0,
+                 "FisheyeSensorDoubleSphereSpec::sanityCheck(): alpha"
+                     << alpha << "is illegal.", );
```
mosra, line 40: Should any of these be `ESP_CHECK()` instead? AFAICT the alpha or focal length could be settable from python and thus a great candidate for a python exception.

src/esp/sensor/FisheyeSensor.cpp
```diff
@@ -0,0 +1,189 @@
...
+      (*shader)
+          .setFocalLength(actualSpec.focalLength)
+          .setPrincipalPointOffset(actualSpec.principalPointOffset)
+          .setAlpha(actualSpec.alpha)
+          .setXi(actualSpec.xi);
+      drawWith(*shader);
+    } break;
+
+      // TODO:
+      // The other FisheyeSensorModelType
```
mosra, line 172: `default: CORRADE_INTERNAL_ASSERT_UNREACHABLE();` maybe? :) just so it properly slaps you when you forget, haha

src/esp/sensor/FisheyeSensor.h
```diff
@@ -0,0 +1,166 @@
...
+
+  gfx::RenderCamera* getRenderCamera() = delete;
+
+ protected:
+  FisheyeSensorSpec::ptr fisheyeSensorSpec_ =
+      std::dynamic_pointer_cast<FisheyeSensorSpec>(spec_);
+  // raw pointer only, we can create it but let magnum to handle the memory
+  // recycling when releasing it.
+  gfx::CubeMapCamera* cubeMapCamera_;
+  std::unique_ptr<esp::gfx::CubeMap> cubeMap_ = nullptr;
```
mosra, line 129: This could possibly be an `Optional`? You're including the header anyway and it'd be one allocation less.

src/utils/viewer/viewer.cpp
```diff
@@ -956,12 +1038,19 @@ void Viewer::drawEvent() {
...
-                     : "Pinhole"));
+    ImGui::Text(
+        "%s camera",
+        (fisheyeMode_ != 0
+             ? "Fisheye"
+             : (cam.getCameraType() == esp::sensor::SensorSubType::Orthographic
+                    ? "Orthographic"
+                    : cam.getCameraType() == esp::sensor::SensorSubType::Pinhole
+                          ? "Pinhole"
+                          : "Fisheye")));
```
mosra, line 1053: Could *maybe* make sense to rewrite this to a cascaded if :)

src/shaders/doubleSphereCamera.frag
```diff
@@ -0,0 +1,50 @@
...
+
+void main(void) {
+  vec3 m;
+  m.xy = (gl_FragCoord.xy - PrincipalPointOffset) / FocalLength;
+  // MUST flip the x axis
+  m.x = -m.x;
+  float r2 = dot(m.xy, m.xy);
+  float sq1 = 1.0 - (2 * Alpha - 1.0) * r2;
+  if (sq1 < 0.0) {
+    discard;
```
mosra, line 32: Question -- how would the shader perform without the `discard`s? As in, GPU time spent. If my knowledge isn't outdated, the `discard`s inhibit certain optimizations in the driver and have a similar effect as if blending would be unnecessarily enabled.

If you're worried about `sqrt()` called with a negative value, that's fine, you just get a NaN, which on most drivers would result in a black color output. Assuming the edges on the output don't affect the consumer, I think even random noise on the edges would be okay.

*Alternatively*, to avoid the need for a discard and to have a control over what's the "outside color", instead of a fullscreen triangle you could render a double circle. This is a common optimization in VR reprojection where the cost of those few extra fragment shader invocations on the edges is high enough that rendering a more complex mesh usually pays off. Not sure here, you'd have to profile, but I don't think it would be worse than with the discards :)

mosra (commented):
Got any screenshots? :eyes:

src/esp/sensor/FisheyeSensor.cpp
bigbike: I do not have preference, either way is OK.
Yes, we can try it.

src/shaders/doubleSphereCamera.frag
bigbike: Update:
Tried. Yes, visually it will not affect. 
Perf does not not improve. I believe this is because these corner cases are so rare.

src/esp/gfx/FisheyeShader.h
mosra: My primary concern was with the `virtual`ity, there's no point in having these virtual, as the base constructor can't call them anyway, and it just makes the class heavier for no reason. Yes, making them pure virtual is _a_ way to enforce the subclass author to provide implementations for those, but it ends there, and the author still has to look around and check *what* the functions need to do and *where* these need to be called, and at that point it could as well be just a comment saying 

> Implement a subclass the same way as `DoubleSphereFisheyeShader` please or @bigbike will get upset.

Apart from the `virtual` aspect -- I know your position on this and I remember we were discussing this quite extensively in the past. However, with shader construction, the steps have to be done in a very rigid manner and I feel this just increases potential for bugs instead of reducing it, as it's easier to:

- forget calling them at the end of the constructor after a larger refactor
- implement them wrong just because they live in a different portion of the file and so the mistakes that you'd immediately see when the code was directly inlined are now non-obvious
- call them before the program was linked -- "she can do it without needing to know any details" is simply not true, the developer *has to know* what these functions exactly do to call them at the right place
- call them outside of the constructor again, repeatedly, in each setter, for no reason at all, just because it seemed to fix some bug

but on the other hand it's unnecessarily hard to transfer any state from the constructor to these (for example, *what if* the constructor has some complex calculated state which affects what textures are fetched, or the uniform names are dynamic etc? now the class has to save that to a private memeber that's not used otherwise...).

Looking at it from a completely different angle, [init functions can be considered as an antipattern](https://softwareengineering.stackexchange.com/q/334970).

- - -

I probably shared this link already, this is John Carmack's view on the topic and he sums up the pros and cons of both approaches better than I could ever do: http://number-none.com/blow/john_carmack_on_inlined_code.html

src/esp/sensor/FisheyeSensor.cpp
mosra: Depends if you're going to expose those to python or not. If not, doesn't matter :)

src/shaders/doubleSphereCamera.frag
mosra: Good, then it probably doesn't make sense to try the double circle mesh either :)

src/esp/sensor/FisheyeSensor.cpp
```diff
@@ -0,0 +1,190 @@
...
+    : VisualSensor(cameraNode, spec) {
+  switch (fisheyeSensorSpec_->fisheyeModelType) {
+    case FisheyeSensorModelType::DoubleSphere: {
+      specSanityCheck<FisheyeSensorDoubleSphereSpec>(fisheyeSensorSpec_.get());
+    } break;
+  };
+
+  // initialize a cubemap
+  auto& res = fisheyeSensorSpec_->resolution;
+  int size = res[0] < res[1] ? res[0] : res[1];
```
eundersander, line 63: We talked about this in a recent meeting and I'm thinking about this more. What if the fisheye FoV is very small (i.e., large focal length)? It seems like the cubemap needs to be higher resolution when the fisheye FoV is smaller.

I would love to see some sample images at different focal lengths.

src/esp/sensor/FisheyeSensor.h
```diff
@@ -0,0 +1,166 @@
...
+      "fisheye-model-type={}-flags={}";
+
+  gfx::RenderCamera* getRenderCamera() = delete;
+
+ protected:
+  FisheyeSensorSpec::ptr fisheyeSensorSpec_ =
+      std::dynamic_pointer_cast<FisheyeSensorSpec>(spec_);
+  // raw pointer only, we can create it but let magnum to handle the memory
+  // recycling when releasing it.
+  gfx::CubeMapCamera* cubeMapCamera_;
```
eundersander, line 128: Does this mean, if I create and delete lots of FisheyeSensor instances, the cubeMapCamera_ instances will hang around unless/until the related scene nodes are deleted?

src/shaders/doubleSphereCamera.frag
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  // unproject to get the ray direction
+  vec3 ray = (m.z * Xi + sqrt(sq2)) / (mz2 + r2) * m - vec3(0.0, 0.0, Xi);
+
+#if defined(COLOR_TEXTURE)
+  fragmentColor = texture(ColorTexture, normalize(ray));
+#endif
+#if defined(DEPTH_TEXTURE)
+  gl_FragDepth = texture(DepthTexture, normalize(ray)).r;
```
eundersander, line 48: Isn't there a depth conversion that needs to happen? The depth in the cubemap texture is the distance along the given cubemap side axis (right?). Shouldn't the depth rendered for this sensor be distance along the camera look axis?

Can you attach some depth renders converted to grayscale? Especially looking in the direction of cubemap corners.

src/utils/viewer/viewer.cpp
```diff
@@ -533,6 +536,62 @@ Viewer::Viewer(const Arguments& arguments)
   cameraSensorSpec->resolution = esp::vec2i(viewportSize[1], viewportSize[0]);
   agentConfig.sensorSpecifications = {cameraSensorSpec};
 
+  // add the new fisheye sensor
```
eundersander, line 539: Could you wrap all this code in a helper function, to keep viewer.cpp a bit more readable?

src/utils/viewer/viewer.cpp
```diff
@@ -878,16 +937,41 @@ void Viewer::drawEvent() {
     // reset timeSinceLastSimulation, accounting for potential overflow
     timeSinceLastSimulation = fmod(timeSinceLastSimulation, 1.0 / 60.0);
   }
-  uint32_t visibles = 0;
-  if (flyingCameraMode_) {
+
```
eundersander, line 940: It'd be nice to see these ~25 lines encapsulated to a helper function, just to keep viewer.cpp a bit more readable.

eundersander (commented):
Cool feature! I left a few comments and requested a few sample renders.

src/esp/sensor/FisheyeSensor.cpp
bigbike: > It seems like the cubemap needs to be higher resolution when the fisheye FoV is smaller.

May I understand why.

> I would love to see some sample images at different focal lengths.

I will try to create them.

src/esp/sensor/FisheyeSensor.h
bigbike: Yes and no. :)
Yes, that is correct. `cubeMapCamera_` is a feature attaching to a node and will be destroyed when the node is gone. If the node is still there, this feature will be there.

No. it will not happen after this [PR](https://github.com/facebookresearch/habitat-sim/pull/1118). If you would like to delete a sensor, you actually delete the node it directly attaches to. We enforce it. And such a "sensor node" will be a "leaf node" when created, which means, when you delete it, you do not need to worry if there is a subtree being connected to it.

src/shaders/doubleSphereCamera.frag
bigbike: > Isn't there a depth conversion that needs to happen? The depth in the cubemap texture is the distance along the given cubemap side axis (right?). Shouldn't the depth rendered for this sensor be distance along the camera look axis?

The depth is a scale, not a vector. The value is like the value in the ordinary depth buffer in an ordinary image (see [here](https://github.com/facebookresearch/habitat-sim/blob/08f5dcae8cb9e94e0985f8134affdb0e8ce1cc2c/src/esp/gfx/CubeMap.cpp#L178))
(A good tutorial is here: https://learnopengl.com/Advanced-OpenGL/Depth-testing)

>Can you attach some depth renders converted to grayscale?

This will depend on this [diff](https://github.com/facebookresearch/habitat-sim/pull/1066), which needs big changes. But I will see what I can do.

>Especially looking in the direction of cubemap corners.

The cubemap is fixed in the camera space, NOT in the world space. You can imagine camera is locating in the center of this cube, and always looking at the -Z direction (Y is up, X is right). If camera rotates (in the world space), the cubemap rotates. In other words, cubemap needs to be rendered in every frame when camera is looking around.

src/utils/viewer/viewer.cpp
bigbike: OK. will do.

src/utils/viewer/viewer.cpp
bigbike: Hello, welcome to our discussion about "to inline or not inline". :)

In `FisheyeShader.h`:
https://github.com/facebookresearch/habitat-sim/pull/1131/files#r589273848

eundersander: @mosra 's concerns in that other thread were around unnecessary use of virtual. Here, I'm just asking for a helper method in the Viewer class. I don't think this is a controversial ask.

src/shaders/doubleSphereCamera.frag
eundersander: >The cubemap is fixed in the camera space, NOT in the world space. 

Ok, makes sense. My concern remains. Let's see a depth render with a wide enough FOV such that some cubemap edges/corners are "in frame".

src/esp/sensor/FisheyeSensor.cpp
eundersander: >May I understand why.

If the FOV were very small, the fisheye sensor would be reading all the pixels for its sensor from a very small portion of the cubemap. If there aren't enough texels in the cubemap, the fisheye render will look blurry.

src/utils/viewer/viewer.cpp
bigbike: When I gave the link, I simply provided a pointer to the beginning of the lengthy conversation. You need to read through it. "virtual" is simply one part but not the main topic in it. Most of the discussion was focusing on "inline or not inline".

src/shaders/doubleSphereCamera.frag
bigbike: Yes, usually there are 8 edges (LT, LF, LB, FT, FB, RT, RF, RB) in the rendering frame. (L = left, R = Right, T = Top, B = bottom) It happens for example in the demo shown in the meeting.

src/esp/sensor/FisheyeSensor.cpp
bigbike: @Skylion007 : You are using the fisheye sensor for training. Do you have any opinions here?

src/esp/sensor/FisheyeSensor.cpp
bigbike: No, these functions will not be exposed directly to python. They will only be called in the sensor's constructor.

src/utils/viewer/viewer.cpp
eundersander: Ok, I skimmed the Carmack article link. The situation here is basically a state machine (fisheye, flying, else) where I'm arguing that each state's logic should be encapsulated. That's not the same as Carmack's code (below), so I don't think Carmack would particularly advocate for inlining all these different states' logic.
```
void MajorFunction( void ) {
        MinorFunction1();
        MinorFunction2();
        MinorFunction3();
}
```

Anyway, I don't feel strongly. This is my two cents. I won't block the PR on this point.

src/esp/sensor/FisheyeSensor.cpp
bigbike: >If the FoV were very small, the fisheye sensor would be reading all the pixels for its sensor from a very small portion of the cubemap. If there aren't enough texels in the cubemap, the fisheye render will look blurry.

Now I understand your concern. Thanks. Here are my thoughts.

First, fisheye sensor is designed to handle the large FOV. If FOV is small, we use pinhole camera model directly. 

Second, let us look into this problem:
The process to render the observation of a fisheye sensor is similar to the ray tracing. For each pixel (u,v) in the image space, we try to unproject it to a ray (2D -> 3D, all points on this ray will be projected to this (u, v)). We sample the cubemap using this ray, a unit vector, and thus find the "light source" pixel on the cubemap.

Blur is more likely to happen to the pixels that are closer to the principal axis. Because a neighboring pixel pair closer to the principal axis will be unprojected to 2 rays that have smaller included angle. That means it has higher chance to hit the same pixel on the cubemap.

**One solution I was thinking** (I did not implement it in this PR due to complexity.):
We know (or compute) the `FOV`, which is an angle. And we can compute the smallest included `angel` between rays, which are projected from the pixels `(h/2 + 1, w/2)` and `(h/2, w/2)` (assuming the principle axis goes through the image center, `(h/2, w/2)`) Then the resolution of the cubemap can be computed as: . It guarantees 1 pixel offset on the image leads to at least 1 pixel offset on the cubemap.

I was asking why, because I never worried about the resolution of cubemap would not be high enough. Actually I was concerned the opposite case, that is, the current resolution is more than enough in most cases, which slows down the rendering.

I can try what I proposed above. Since in the "double sphere" camera model, the `FOV` is not given, I need some extra time to derive it. It will further delay this diff. So I prefer to try it in another PR as future optimization.

**One practical solution to address your concern:**
- I will add a `cubemapSize` in the `FisheyeSensorSpec`, and a flag `CustomizedCubeMapSize` in the `FisheyeSensor` class. When the flag is set, the program will adopt the user input.
- If the flag is not set, by default, the cubemap will use `min(w, h)` as I implemented here.

src/utils/viewer/viewer.cpp
bigbike: >I'm arguing that each state's logic should be encapsulated.

I agree.

@eundersander :  I was actually on your side. :) 
I remembered I did the similar thing when I developed the object picking functions in the viewer and @mosra requested me to inline everything. See [here](https://github.com/facebookresearch/habitat-sim/pull/661#discussion_r463000522). 

But I guess @mosra would require me to rollback if I do it. :grin:

src/shaders/doubleSphereCamera.frag
bigbike: @mosra : I tried it WITHOUT the `discard`. 
>you just get a NaN, which on most drivers would result in a black color output.

No, on my machine, it gives me random colors (sometimes it is light blue, sometimes it is brown etc.) when I resize the window.
However, if I have the "discard" in place, it gives me "black".
With or without the `discard`, the performance is almost the same in this case.

So I put the `discard` back.

**No** `disgard`:
no-disgard

With `disgard`:
discard

bigbike:
These are the pictures (rgb, depth) in different resolution settings. @eundersander

bigbike:
This diff is ready to be reviewed again.

Thanks.

src/shaders/doubleSphereCamera.frag
mosra: Are cases with such a huge border around common? Because then rendering just a (double) circle could lead to pretty significant savings. If not (which I assume), then it's not worth caring about I think.

src/utils/viewer/viewer.cpp
mosra: This is not my code, I don't *require* anybody to do anything ;) 

Just stating my opinion and sharing past experience.

src/shaders/doubleSphereCamera.frag
bigbike: >Are cases with such a huge border around common?

@mosra: No, in terms of training, these are rare cases. The most common case is rendered in a square, no border. (see [here](https://user-images.githubusercontent.com/931093/111849867-e5331180-88cb-11eb-9ed4-605eabc9bfd5.png).)

mosra (commented):
One remaining thing -- does the shader work on webgl? So it doesn't break the web build again like when PBR got introduced.

You might want to validate it with `magnum-shaderconverter` [as I mentioned previously](https://cvmlp.slack.com/archives/CFN5TAUSD/p1606765426042700?thread_ts=1606149864.021100&cid=CFN5TAUSD): 

```sh
magnum-shaderconverter --validate file.glsl \
    --input-version "300 es" --output-version opengl # and possibly a few -Ds as well
```

bigbike:
Not easy. Finally make the `shaderconverter` work on ubuntu with the help of @mosra. (Cannot make it work on Mac as there is something wrong with the latest `brew install`.)

```
magnum-shaderconverter --validate doubleSphereCamera.frag --input-version "300 es" --output-version opengl
```

Fixed the errors in the shader so it is now compatible with the WebGL.

bigbike:
I believe this PR is ready to go.

mosra (approved):

src/esp/gfx/CubeMap.cpp
```diff
@@ -155,6 +155,7 @@ void CubeMap::recreateTexture() {
   // color texture
   if (flags_ & Flag::ColorTexture) {
     auto& colorTexture = texture(TextureType::Color);
```
Skylion007, line 157: Should this line be removed or the line above it be deleted? It seems weird to overwrite colorTexture and depthTexture immediately like this.

bigbike: > overwrite

Yes, it overrides whatever existing in the variable. This function is called `recreateTexture()`. It is its job.
The function `texture` is to make sure the type is legal, and return the texture container properly.

Skylion007:
Looks good to me. Let's merge.

PR #1134, opened 2021-03-08
Support for shader_type specification in stage and object configurations - Part 1 of 2

jturner65:
## Motivation and Context
This PR is the first of 2 PRs intended to introduce configuration-file level support for independent shader types being specified for the stage and objects in a scene. Supported shader types that the user can specify are Flat, Phong and PBR (case insensitive), and these values can be specified in either stage and object configurations (to specify a default value for a particular stage or object TYPE), or in the scene instance configurations (which, if specified, will override the default values given in stage or object configs for a single instance only).

To support this functionality, stage and object creation had to be slightly modified in order to accept a copy of the appropriate attributes as a parameter, which is then moved to the initialization Attributes of the resultant rigidStage/rigidObject.

Part 1 of 2 : This PR only implements the framework required for this change, but lacks the functionality to consume the configuration-specified shader type in ResourceManager.  The lacking functionality will be added in Part 2 of 2.

## How Has This Been Tested
C++ and Python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -493,21 +493,20 @@ scene::SceneNode* ResourceManager::loadAndCreateRenderAssetInstance(
       sceneID = creation.isSemantic() ? activeSceneIDs[1] : activeSceneIDs[0];
     }
   }
-
-  auto& sceneGraph = sceneManagerPtr->getSceneGraph(sceneID);
-  auto& rootNode = sceneGraph.getRootNode();
-  auto& drawables = sceneGraph.getDrawables();
-
   const bool fileIsLoaded = resourceDict_.count(assetInfo.filepath) > 0;
   if (!fileIsLoaded) {
     if (!loadRenderAsset(assetInfo)) {
       return nullptr;
     }
   }
-
   ASSERT(assetInfo.filepath == creation.filepath);
+
+  auto& sceneGraph = sceneManagerPtr->getSceneGraph(sceneID);
+  auto& rootNode = sceneGraph.getRootNode();
+  auto& drawables = sceneGraph.getDrawables();
+
   return createRenderAssetInstance(creation, &rootNode, &drawables);
-}
+}  // ResourceManager::loadAndCreateRenderAssetInstance
 
 bool ResourceManager::loadRenderAsset(const AssetInfo& info) {
   bool meshSuccess = false;
@@ -570,7 +569,7 @@ scene::SceneNode* ResourceManager::createRenderAssetInstance(
   }
 
   return newNode;
-}
+}  // ResourceManager::createRenderAssetInstance
 
 bool ResourceManager::loadStageInternal(
     const AssetInfo& info,
@@ -658,7 +657,7 @@ bool ResourceManager::buildStageCollisionMeshGroup(
 
 bool ResourceManager::loadObjectMeshDataFromFile(
     const std::string& filename,
-    const std::string& objectTemplateHandle,
+    const metadata::attributes::ObjectAttributes::ptr& objectAttributes,
     const std::string& meshType,
     const bool requiresLighting) {
   bool success = false;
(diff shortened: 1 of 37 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -2258,8 +2257,17 @@ void ResourceManager::createConvexHullDecomposition(
...
   if (resourceDict_.count(filename) == 0) {
+    // retrieve existing, or create new, object attributes corresponding to
+    // passed filename
+    auto objAttributes =
+        getObjectAttributesManager()->getObjectCopyByHandle(filename);
+    if (objAttributes == nullptr) {
+      objAttributes =
+          getObjectAttributesManager()->createObject(filename, false);
+    }
+
```
jturner65, line 2268: Note this change - the object being decomposed needs to have an attributes, at least to pass to loadObjectMeshDataFromFile, since this function is going to process shader_type directives from configs.  If no attributes for the source object exists, one is made, but not registered.

src/esp/sim/Simulator.cpp
```diff
@@ -419,28 +430,50 @@ bool Simulator::createSceneInstance(const std::string& activeSceneName) {
...
-                      "attributes whose handle contains "
-                   << objInst->getHandle()
-                   << " as specified in object instance attributes, so unable "
-                      "to instance object; skipping. ";
-      continue;
+      LOG(ERROR) << errMsgTmplt
+                 << "Unable to find objectAttributes whose handle contains "
+                 << objInst->getHandle()
+                 << " as specified in object instance attributes.";
+      return false;
```
jturner65, line 446: This and other warnings in this loop were changed to errors so that scene instancing would stop and inform the user if there were any issues processing the scene object instance configs specified.  This would indicate errors in the JSON that would result in objects being poorly formed, or otherwise not implemented. The thinking is that it would be better to stop with an error during training than just not have an important object in the scene.

src/esp/core/ManagedContainerBase.cpp
```diff
@@ -159,9 +161,9 @@ bool ManagedContainerBase::verifyLoadDocument(const std::string& filename,
     return true;
   } else {
     // by here always fail
-    LOG(ERROR) << objectType_
-               << "ManagedContainerBase::verifyLoadDocument : File " << filename
-               << " does not exist";
+    !Cr::Utility::Error{} << objectType_
```
mosra, line 164: FYI there are slightly shorter `Mn::Debug`, `Mn:Error` aliases to save you typing.

(These are not really visible in the docs, I'll fix that.)

src/esp/metadata/MetadataMediator.cpp
```diff
@@ -312,7 +321,7 @@ attributes::SceneAttributes::ptr MetadataMediator::getSceneAttributesByName(
         // 4.  Existing stage config/asset on disk, but not in current dataset.
         //    In this case, a stage attributes is loaded and registered, then
         //    added to current dataset, and then 3. is performed.
-        LOG(INFO)
+        !Cr::Utility::Debug{}
```
mosra, line 324: About the `!` that prepends the message with file/line info, just a personal opinion -- I personally find it to be unnecessary noise, given that the error message contains the function name already. Consequently this is also a reason why there's `CORRADE_ASSERT()` with a message but no file/line info and `CORRADE_INTERNAL_ASSERT()` with file/line info but no message:

- user-facing messages should contain enough info to allow the user fix the root cause that caused the message and shouldn't need file/line info that's just noise to the user (if I get a bugreport in Magnum with certain error and it's not immediately clear from the message, it's enough to just `grep` for the message text, plus file/line info gets outdated *very soon* if users tend to prefer stable releases)
- on the other hand, file/line info is very useful for immediate "printf debugging" and developer-facing messages, where knowing the exact location where the error happened is most useful, not matter how beautifully crafted the human-readable error message is

mosra (commented):
Big diff, anything I should pay closer attention to? :)

src/esp/metadata/MetadataMediator.cpp
jturner65: Thanks.  So perhaps I'll just use the ! messages for debug, and change the warning and error messages to not have the bang?

mosra: I would personally not have it anywhere, given that there's always `MetadataMediator::getSceneAttributesByName :` and the like, to me it looks like the message alone tells the location quite well on its own :)

jturner65: Ok.  I was thinking about getting rid of that message prefix (I manually add that everywhere heh), but the bang output is a bit long, so i'll get rid of the bang :)

jturner65: Ooof.  This log messaging refactor really shouldn't be a part of this PR anyway.  I accidentally merged my shader type branch onto my message refactor branch instead of master.  I am going to remake this PR properly. Sorry!

PR #1135, opened 2021-03-09
[Sensor] Remove logging on SensorSpec destructor

erikwijmans:
## Motivation and Context

SensorSpec can't have a print in its destructor since they are created/destroyed on episode swap in habitat-lab.  Having one makes logging really noisy and, as a byproduct, kill perf -- while most of the time training is done with logging off, leaving it on can be useful for debugging, so we should not overly spam INFO.

## How Has This Been Tested

In prod.

## Types of changes

### DIFF
--- a/src/esp/sensor/Sensor.cpp
+++ b/src/esp/sensor/Sensor.cpp
@@ -11,10 +11,6 @@
 namespace esp {
 namespace sensor {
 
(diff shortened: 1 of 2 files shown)


bigbike (approved):

bigbike:
Ship it.

Skylion007 (approved):

PR #1140, opened 2021-03-13
[1/3] Voxel Framework integration

ScottCSteinhauser:
## Motivation and Context

I've build a framework that can voxelize a scene or object to a specified resolution, creating a boundary voxelization (voxels are booleans - either filled or unfilled), and construct other voxelizations within the same VoxelGrid class (signed distance fields, Vector fields, etc). It includes various ways of visualizing the fields constructed which can be tested in Viewer.

The following is an example of a rendered vector-field.
Screen Shot 2021-04-07 at 10 48 52 AM

Going in viewer and using Ctrl + Right Click will voxelize a selected object
Screen Shot 2021-03-25 at 3 57 25 PM
Screen Shot 2021-03-25 at 3 57 41 PM

The voxelization framework has been built in a flexible way such that users can employ the framework for their own purposes. 

## How Has This Been Tested

A simple test is found in src/tests/VoxelGridTest.cpp, which tests voxelizing a scene, retrieving voxels, and converting voxel indices into global coordinates and vice versa.

## Types of changes

- Implemented a voxelization framework in VoxelGrid.cpp and VoxelWrapper.cpp
- Added a voxel field visualizer in viewer (Ctrl + Right Click)
- Added Simulator functions for voxelizing a scene/object, retrieving the voxelization, and visualizing the voxelization mesh.

## What else needs to be done
For this PR:
 - I've built functionality for generating Distance Gradient fields and Signed distance fields, and will be adding those to a utility class in a follow up PR. 
 - I will also do a PR with python bindings and a tutorial.


src/utils/viewer/viewer.cpp
```diff
@@ -1076,7 +1143,7 @@ void Viewer::mousePressEvent(MouseEvent& event) {
     // also, setup the color attachment for rendering, and remove the
     // visualizer for the previously picked object
     objectPickingHelper_->prepareToDraw();
-
+    !Mn::Debug();
```
Skylion007, line 1146: Is this suppose to be here?

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+        } else if (getBoolVoxelByIndex(Mn::Vector3i(ind, j, k))) {
+          hit = true;
+          setBoolVoxelByIndex(Mn::Vector3i(ind, j, k), true, "negXShadow");
+        }
+      }
+      // fill posX
+      hit = false;
+      ind = -1;
+      while (++ind < m_voxelGridDimensions[0]) {
+        if (hit) {
```
Skylion007, line 402: This could be simplified with something like ``` hit || (hit = getBoolVoxelByIndex(Mn::Vector3i(ind, j, k)) ``` or maybe 
```c++
hit |= getBoolVoxelByIndex(Mn::Vector3i(ind, j, k));
if (hit)  setBoolVoxelByIndex(Mn::Vector3i(ind, j, k), true, "posXShadow");
```

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+  addBoolGrid("negYShadow");
+  addBoolGrid("posYShadow");
+  addBoolGrid("negZShadow");
+  addBoolGrid("posZShadow");
+  // fill each grids with ray cast
+  bool hit = false;
+  int ind = 0;
+  std::string gridName = "InteriorExterior";
+  // X axis ray casts
+  for (int j = 0; j < m_voxelGridDimensions[1]; j++) {
```
Skylion007, line 385: Really should abstract away these raycasts into a little function. A lot of repeated code here where only the axises and string vary.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+  int hash = 0;
+  // fill in int grid with voting approach
+  for (int i = 0; i < m_voxelGridDimensions[0]; i++) {
+    for (int j = 0; j < m_voxelGridDimensions[1]; j++) {
+      for (int k = 0; k < m_voxelGridDimensions[2]; k++) {
+        if (getBoolVoxelByIndex(Mn::Vector3i(i, j, k))) {
+          setIntVoxelByIndex(Mn::Vector3i(i, j, k), 0, gridName);
+          continue;
+        }
+        nX = !getBoolVoxelByIndex(Mn::Vector3i(i, j, k), "negXShadow");
```
Skylion007, line 478: Why not just reuse the vector here?

Skylion007 (changes requested):

Skylion007:
FYI, there are lot of clang warnings about what look like unintentional use of Magnum:Vector3 instead of Magnum:Vector3i

src/esp/geo/CMakeLists.txt
```diff
@@ -6,9 +6,19 @@ add_library(
...
+  VoxelGrid.h
+  VoxelWrapper.cpp
+  VoxelWrapper.h
 )
 
 target_link_libraries(
   geo
   PUBLIC core glog io gfx
 )
+
```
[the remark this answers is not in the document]
ScottCSteinhauser: Yes! Thank you for pointing this out.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+  std::shared_ptr<bool> boundary_grid(b_grid);
+  boolGrids_.insert(std::make_pair("Boundary", boundary_grid));
+  int num_filled = 0;
+  // Transfer data from Volume to VoxelGrid
+  for (int i = 0; i < dims[0]; i++) {
+    for (int j = 0; j < dims[1]; j++) {
+      for (int k = 0; k < dims[2]; k++) {
+        if (vhacdVolume->GetVoxel(i, j, k) >= 2) {
+          num_filled++;
+          setBoolVoxelByIndex(Mn::Vector3i(i, j, k), true);
```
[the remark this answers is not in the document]
ScottCSteinhauser: I'm avoiding doing that right now since setBoolVoxelByIndex has to retrieve the relevant dict, hash the voxel index, and then set the boolean at that index. However I think I can do this more efficiently looking at it.

src/esp/assets/ResourceManager.h
```diff
@@ -966,6 +1007,14 @@ class ResourceManager {
    */
   int nextMaterialID_ = 0;
 
+  /**
+   * @brief Storage for precomuted voxel grids. Useful for when multiple objects
+   * in a scene are using the same VoxelGrid.
+   *
+   * Maps absolute path keys to VoxelGrid.
+   */
+  std::map<std::string, std::shared_ptr<esp::geo::VoxelGrid>> voxelGridDict_;
```
eundersander, line 1016: What is the reasoning for having ResourceManager keep a map of voxel grids?

Is a voxel grid a "resource"? What is a "resource"? What is the job of ResourceManager? @aclegg3

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+void VoxelGrid::addIntGrid(const std::string& gridName) {
+  if (intGrids_.find(gridName) != intGrids_.end()) {
+    // grid exists, simply overwrite
+    Mn::Debug() << "Grid" << gridName << "exists, overwriting.";
+    std::memset(intGrids_[gridName].get(), 0,
+                m_voxelGridDimensions[0] * m_voxelGridDimensions[1] *
+                    m_voxelGridDimensions[2] * sizeof(int));
+    return;
+  }
+  const int gridSize = m_voxelGridDimensions[0] * m_voxelGridDimensions[1] *
```
eundersander, line 112: use gridSize in the above memset, too

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+                       m_voxelGridDimensions[2];
+  float* n_grid = new float[gridSize];
+  std::shared_ptr<float> new_grid(n_grid);
+  std::memset(new_grid.get(), 0, gridSize * sizeof(float));
+  floatGrids_.insert(std::make_pair(gridName, new_grid));
+}
+
+void VoxelGrid::addVector3Grid(const std::string& gridName) {
```
eundersander, line 137: a lot of code duplication in addIntGrid, addFloatGrid, addVector3Grid. Can they all share a single template helper function?

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+  Mn::Vector3* n_grid = new Mn::Vector3[gridSize];
+  std::shared_ptr<Mn::Vector3> new_grid(n_grid);
+  std::memset(new_grid.get(), 0, gridSize * sizeof(Mn::Vector3));
+  vector3Grids_.insert(std::make_pair(gridName, new_grid));
+}
+
+// Currently naive hashing. TODO: Look into hibert curves and Z-curve
+// alternatives for mappings that maximize memory locality of adjacent cells.
+int VoxelGrid::hashVoxelIndex(const Mn::Vector3i& coords) {
+  int hashed_voxel =
```
eundersander, line 157: This is probably a good spot to assert valid coords (not out of range).

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+      coords[0] + coords[1] * m_voxelGridDimensions[0] +
+      coords[2] * m_voxelGridDimensions[0] * m_voxelGridDimensions[1];
+  return hashed_voxel;
+}
+
+Mn::Vector3i VoxelGrid::reverseHash(const int hash) {
+  return Mn::Vector3i(
```
eundersander, line 164: Assert on valid hash (not out of range)

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+                                       const Mn::Vector3& val,
+                                       const std::string& gridName) {
+  assert(vector3Grids_.find(gridName) != vector3Grids_.end());
+  int hashedVoxelIndex = hashVoxelIndex(coords);
+  vector3Grids_[gridName].get()[hashedVoxelIndex] = val;
+}
+
+std::shared_ptr<Mn::Trade::MeshData> VoxelGrid::getMeshData(
+    const std::string& gridName) {
+  if (meshDataDict_[gridName] == nullptr)
```
eundersander, line 237: What happens if I get the mesh, change the underlying grid, then get the mesh again? Is there logic somewhere to ensure the mesh gets updated to reflect the changes to the underlying grid?

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+int VoxelGrid::generateBoolGridFromFloatGrid(const std::string& floatGridName,
+                                             std::string boolGridName,
+                                             float startRange,
+                                             float endRange) {
+  assert(floatGrids_.find(floatGridName) != floatGrids_.end());
+  int num_filled = 0;
+  if (boolGridName == "")
+    boolGridName = floatGridName;
+  addBoolGrid(boolGridName);
+  for (int i = 0; i < m_voxelGridDimensions[0] * m_voxelGridDimensions[1] *
```
eundersander, line 306: This expression `m_voxelGridDimensions[0] * m_voxelGridDimensions[1] * m_voxelGridDimensions[2]` is used pretty often so let's have a helper `cellCount()` method or similar.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+}
+
+Mn::GL::Mesh& VoxelGrid::getMeshGL(const std::string& gridName) {
+  if (meshDataDict_[gridName] == nullptr)
+    generateMesh(gridName);
+  return meshGLDict_[gridName];
+}
+
+Mn::Vector3 VoxelGrid::getGlobalCoords(const Mn::Vector3i& coords) {
+  Mn::Vector3 global_coords((coords[0]) * m_voxelSize[0],
```
eundersander, line 249: I'm always wary of off-by-one and similar bugs for this kind of conversion.

Would you say this is getting the world-space position of a particular corner of the voxel? Or the position of the center of the voxel?

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,451 @@
...
+   * @return A shared pointer to the MeshData.
+   */
+  std::shared_ptr<Mn::Trade::MeshData> getMeshData(
+      const std::string& gridName = "Boundary");
+
+  /**
+   * @brief Retrieves the MeshGL used for rendering for a particular voxelGrid.
+   * If it does not exist, it will generate the mesh for that grid.
+   * @param gridName The key underwhich the desired voxel grid is registered.
+   * @return A shared pointer to the MeshGL.
```
eundersander, line 226: the @return description looks wrong

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,451 @@
...
+   * @brief Generates a Vector3 field where each vector of a cell points away
+   * from it's closest Boundary cell.
+   * @param gridName The name underwhich to register the newly created distance
+   * flow field.
+   */
+  void generateDistanceFlowField(
+      const std::string& gridName = "DistanceFlowField");
+
+  /**
+   * @brief Saves a particular grid to a svx file at a specified directory.
```
eundersander, line 375: A link to info on the web about svx format would be nice.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,923 @@
...
+
+  // VHACD computes a axis-aligned bounding box; we need to offset the voxelgrid
+  // by the minimum corner of the AABB
+  VHACD::Vec3<double> minBB = vhacdVolume->getMinBB();
+  m_offset = Mn::Vector3(minBB[0], minBB[1], minBB[2]);
+
+  // create empty VoxelGrid
+  const int gridSize = dims[0] * dims[1] * dims[2];
+  bool* b_grid = new bool[gridSize];
+  std::shared_ptr<bool> boundary_grid(b_grid);
```
eundersander, line 52: This looks shady. How does shared_ptr know the bool* is a pointer to a C-style array of bools and not a pointer to a single bool allocated via new?

I did a quick test in coderpad and I hit `==16==ERROR: AddressSanitizer: alloc-dealloc-mismatch (operator new [] vs operator delete) on 0x606000000020`.

src/esp/geo/VoxelWrapper.h
```diff
@@ -0,0 +1,72 @@
...
+#define ESP_GEO_VOXEL_WRAPPER_H_
+
+#include "VoxelGrid.h"
+#include "esp/assets/ResourceManager.h"
+#include "esp/scene/SceneNode.h"
+
+namespace esp {
+namespace geo {
+
+class VoxelWrapper {
```
eundersander, line 11: How about a comment explaining this class?

src/tests/VoxelGridTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  // Voxelize the scene with resolution = 1,000,000 and make asserts
+  const int resolution = 1000000;
+  simulator_->createSceneVoxelization(resolution);
+  auto voxelWrapper = simulator_->getSceneVoxelization().get();
+
+  // Verify coordinate conversion works in both directions
+  Mn::Vector3i voxelIndex(2, 1, 7);
+  Mn::Vector3 globalCoords =
+      voxelWrapper->getGlobalCoordsFromVoxelIndex(voxelIndex);
```
eundersander, line 49: Let's also check that the global coords are correct.

src/tests/VoxelGridTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+// TODO: Add tests for different Sensors
+struct VoxelGrid : Cr::TestSuite::Tester {
+  explicit VoxelGrid();
+
+  void testVoxelGrid();
+};
+
+VoxelGrid::VoxelGrid() {
+#ifdef ESP_BUILD_WITH_VHACD
+  addTests({&VoxelGrid::testVoxelGrid});
```
eundersander, line 25: Are you planning to add more tests? There's a lot of functionality in this PR that isn't tested here.

src/utils/viewer/viewer.cpp
```diff
@@ -1302,6 +1382,70 @@ void Viewer::keyPressEvent(KeyEvent& event) {
     case KeyEvent::Key::V:
       invertGravity();
       break;
+#ifdef ESP_BUILD_WITH_VHACD
```
eundersander, line 1385: How about helper functions in viewer.cpp to encapsulate this logic? Just to keep viewer.cpp more readable.

eundersander (commented):
I left some more comments. The new/shared_ptr comment is the most concerning.

src/esp/geo/VoxelGrid.cpp
mosra: FYI this is getting switched to `Containers::Array`, we were discussing this on Slack and the code just didn't land here yet.

About the deallocation, it could be fixed with `std::shared_ptr<bool[]>`, but the `shared_ptr` has no notion of the size which is why we're switching to an Array.

src/esp/assets/ResourceManager.h
ScottCSteinhauser: Right now I'm registering each VoxelGrid in ResourceManager so that multiple objects with the same RenderAssetHandle can share voxel grids. Do you think it'd make more sense to register the Voxel Grids in a dict in PhysicsManager?

src/esp/geo/VoxelGrid.cpp
ScottCSteinhauser: I considered giving each mesh a boolean that would keep track of whether or not it's in sync with the underlying grid. The problem I think is that each time a voxel is changed, I'd have to set this boolean. Since a user could potentially set millions of voxels, I think it makes sense to leave it up to them to manually update the mesh (by calling generateMesh(gridName)) again when they need to. Does that answer your question, or do you think there is a better way?

src/esp/geo/VoxelGrid.cpp
ScottCSteinhauser: This gets the position of the center of the voxel.

src/tests/VoxelGridTest.cpp
ScottCSteinhauser: Yes definitely. I think that a lot of the tests can be encapsulated in the saving & loading to SVX format, so I'll be implementing those once I'm done with the saving/loading.

src/esp/geo/VoxelGrid.cpp
ScottCSteinhauser: Made the change to Containers::Array<char> :)

src/esp/assets/ResourceManager.h
aclegg3: For this purpose the primary motivation is to keep asset related resources together. Since voxel fields are registered per-asset keyed by that asset's handle, shared per-instance, and require access to the mesh geometry this was a most straightforward place for the underlying data to live. 

At a high-level I agree this is debatable and ResourceManager owns too much overall. Let's chat more about this.

src/esp/geo/VoxelGrid.cpp
eundersander: Manual/explicit calls to generateMesh sounds reasonable. As a nit, I'd prefer it to *always* be manual, even for the first generate, for the sake of consistency. As-is, you risk confusing a user who never discovered the "generateMesh" API and is surprised to be getting a stale mesh from getMeshData in some cases.

ScottCSteinhauser: I see - would a different naming scheme be more appropriate? I thought about doing updateMesh() instead, but the intitial creation of the mesh isn't updating anything since the mesh hasn't been made.

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,515 @@
...
+   * @brief Checks to see if a given 3D voxel index is valid and does not go out
+   * of bounds.
+   * @param coords The voxel index.
+   * @return True if the voxel index is valid, false otherwise.
+   */
+  bool isValidIndex(const Mn::Vector3i& coords) const {
+    return coords[0] >= 0 && coords[1] >= 0 && coords[2] >= 0 &&
+           coords[0] < m_voxelGridDimensions[0] &&
+           coords[1] < m_voxelGridDimensions[1] &&
+           coords[2] < m_voxelGridDimensions[2];
```
mosra, line 143: Most pperations done on scalars can be done on the vectors as well :) In this case:

```cpp
return (coords >= Mn::Vector3i{0}).all() && (coords < m_voxelGridDimensions).all();
```

(I was trying to figure out if there was a way to not need this function but failed. So at least this.)

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,515 @@
...
+   * @param gridName The voxel grid.
+   * @param value The new value.
+   */
+  template <typename T>
+  void setVoxel(Mn::Vector3i index,
+                const std::string& gridName,
+                const T& value) {
+    auto arrayView3D = getGrid<T>(gridName);
+    arrayView3D[index[0]][index[1]][index[2]] = value;
+  }
```
mosra, line 159: Honestly I would just remove this, and the `getVoxel()` function as well, because it encourages slow (= bad) use of the class, and if it wouldn't be here, it would force all users to think deeper about accessing the data :) I'll comment on the relevant cases in `VoxelGrid.cpp` to show what I mean.

I also see a TODO for myself here which I should be able to implement soon -- being able to do `arrayView3D[index]` directly without this verbose indexing. (Looking at your code, you don't actually need it yet it seems, as you have separate `i`/`j`/`k` variables in most cases anyway.)

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,515 @@
...
+   * @return The has value of the voxel index.
+   */
+  int hashVoxelIndex(const Mn::Vector3i& coords);
+
+  /**
+   * @brief Converts a hash value into a 3 dimensional coordinate.
+   * @param hash The hash value.
+   * @return The voxel index derived from the given hash value.
+   */
+  Mn::Vector3i reverseHash(const int hash);
```
mosra, line 131: As with `setVoxelByHash()` / `getVoxelByHash()`, I don't see a reason for these to exist anymore :)

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,515 @@
...
+                                 const Mn::Vector3i& local_coords,
+                                 const Mn::Vector3& vec);
+
+  /**
+   * @brief Gets the length of the voxel grid.
+   * @return The length of the 1 dimensional array voxel grid.
+   */
+  int gridSize() {
+    return m_voxelGridDimensions[0] * m_voxelGridDimensions[1] *
+           m_voxelGridDimensions[2];
```
mosra, line 481: `return m_voxelGridDimensions.product();` ;)

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,877 @@
...
+    }
+  }
+
+  // 1st sweep
+  for (int i = 0; i < m_voxelGridDimensions[0]; i++) {
+    for (int j = 0; j < m_voxelGridDimensions[1]; j++) {
+      for (int k = 0; k < m_voxelGridDimensions[2]; k++) {
+        int i_behind = INT_MAX, j_behind = INT_MAX, k_behind = INT_MAX;
+        if (isValidIndex(Mn::Vector3i(i - 1, j, k))) {
+          i_behind = abs(std::max(sdfGrid[i - 1][j][k], -2147483646));
```
mosra, line 429:
```suggestion
          i_behind = abs(std::max(sdfGrid[i - 1][j][k], -INT_MAX));
```

? :)

Same in all the occurences below.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,877 @@
...
+  auto sdfGrid = getGrid<int>(gridName);
+
+  for (int i = 0; i < m_voxelGridDimensions[0]; i++) {
+    for (int j = 0; j < m_voxelGridDimensions[1]; j++) {
+      for (int k = 0; k < m_voxelGridDimensions[2]; k++) {
+        // copy value
+        sdfGrid[i][j][k] = intExtGrid[i][j][k];
+      }
+    }
+  }
```
mosra, line 421: [`Cr::Utility::copy(intExtGrid, sdfGrid);`](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#a6cd0c04124a0d7c4879d4cf6da3556a2)

(Note the include.) With magic containers come magic tools, haha :sparkles:

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,877 @@
...
+  addGrid<Mn::Vector3>("ClosestBoundaryCell");
+  auto intExtGrid = getGrid<int>("InteriorExterior");
+  auto closestCellGrid = getGrid<Mn::Vector3>("ClosestBoundaryCell");
+
+  Mn::Debug() << m_voxelGridDimensions;
+  for (int i = 0; i < m_voxelGridDimensions[0]; i++) {
+    for (int j = 0; j < m_voxelGridDimensions[1]; j++) {
+      for (int k = 0; k < m_voxelGridDimensions[2]; k++) {
+        Mn::Vector3i index = Mn::Vector3i(i, j, k);
+        int label = getVoxel<int>(index, "InteriorExterior");
```
mosra, line 517: And this is why I was suggesting to just remove the `getVoxel()` / `setVoxel()` APIs altogether -- to avoid using them by accident. (Yes, I know they had their use before you introduced strided array views, but now they don't anymore.) Because this means you do a redundant string lookup for *every voxel*, while you could do just *one*, above when you called the `getGrid()`, and then you can index the grid directly:

```suggestion
        int label = intExtGrid[i][j][k];
```

Same in all other cases where `getVoxel()` / `setVoxel()` is used -- fetch the grid once, index the view instead :)

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,877 @@
...
+          voxelSet.push_back(Mn::Vector3i(i, j, k));
+        }
+      }
+    }
+  }
+}
+
+// 6D SDF - labels each cell as interior (-inf), exterior (+inf), or boundary
+void VoxelGrid::generateInteriorExteriorVoxelGrid() {
+  // create 6 bool grids
```
mosra, line 267: To me it looks like one int grid could be more efficient, and then use the bits of that int as +X, -X, ... Or, for more type safety and easier use, add a possibility to have a `Math::BoolVector<6>` grid and use that :)

Seeing [Aaron's comment below](https://github.com/facebookresearch/habitat-sim/pull/1140/files#r593914384), it may also help with the code duplication. 

A bit more advanced solution could be massaging the view via `transpose()` / `flip()` so you then always go through the first two dimensions and the third is used for raycasts.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,877 @@
...
+      }
+    }
+  }
+  // remove the transient utility grids
+  removeGrid("negXShadow");
+  removeGrid("posXShadow");
+  removeGrid("negYShadow");
+  removeGrid("posYShadow");
+  removeGrid("negZShadow");
+  removeGrid("posZShadow");
```
mosra, line 399: Probably could be worth to have a way to create a grid but don't add it to the map (which then means it doesn't need to have any string name, can get removed/destructed implicitly at the end of scope etc...)

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,515 @@
...
+    // set type depending on the type of T
+    if (std::string(typeid(T).name()) == "i") {
+      type = "int";
+    } else if (std::string(typeid(T).name()) == "f") {
+      type = "float";
+    } else if (std::string(typeid(T).name()) == "b") {
+      type = "bool";
+    } else if (std::string(typeid(T).name()) == "N6Magnum4Math7Vector3IfEE") {
+      type = "vector3";
+    }
```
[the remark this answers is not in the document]
ScottCSteinhauser: Thank you!

src/esp/bindings/GeoBindings.cpp
```diff
@@ -64,6 +66,112 @@ void initGeoBindings(py::module& m) {
            1.0 is a chordal Catmull-Rom spline)",
       "key_points"_a, "num_interpolations"_a, "alpha"_a = .5f);
 
+  // ==== VoxelWrapper ====
+  py::class_<VoxelWrapper, VoxelWrapper::ptr>(m, "VoxelWrapper")
+      // TODO: Replace with magnum binded Strided Array views (Not sure how to
+      // do this yet..)
```
mosra, line 72: I'll look into this the next days -- *ideally*, when it's fixed, you should be able to directly return the typed views.

src/esp/bindings/GeoBindings.cpp
```diff
@@ -64,6 +66,112 @@ void initGeoBindings(py::module& m) {
...
+          &VoxelWrapper::generateBoolGridFromIntGrid, "int_grid_name"_a,
+          "bool_grid_name"_a, "start_range"_a, "end_range"_a,
+          R"(Takes in the name of an existing integer grid and produces a boolean grid where each voxel is set to true if the value of that voxel in the integer grid falls within the specified range. The newly created boolean grid is registered under bool_grid_name.)")
+      .def(
+          "generate_bool_grid_from_float_grid",
+          &VoxelWrapper::generateBoolGridFromFloatGrid, "float_grid_name"_a,
+          "bool_grid_name"_a, "start_range"_a, "end_range"_a,
+          R"(Takes in the name of an existing float grid and produces a boolean grid where each voxel is set to true if the value of that voxel in the float grid falls within the specified range. The newly created boolean grid is registered under bool_grid_name.)")
+      // TODO: put voxel set filling functions here (need to figure out how to
+      // pass in function ptr as arg)
```
mosra, line 130: A `std::function` works, but a plain function pointer won't (to make it work, pybind attaches a lot of state to each function).

However, an API in this form will be very slow due to all the jumps between C++ and Python in the callback, so it'll make more sense to do that processing directly in python over the returned view (...once i implement that).

src/esp/sim/Simulator.cpp
```diff
@@ -790,6 +790,22 @@ Magnum::Vector3 Simulator::getAngularVelocity(const int objectID,
...
+  // If the objectId > 0, then pull from the Object template. Otherwise, return
+  // the scene handle.
+  auto objHandle =
+      objectId >= 0
+          ? getObjectInitializationTemplate(objectId)->getRenderAssetHandle()
+          : getStageInitializationTemplate()->getRenderAssetHandle();
+
+  // get and return mesh
+  return resourceManager_->createJoinedCollisionMesh(objHandle);
+}
```
aclegg3, line 807: What is this used for?

examples/tutorials/voxel_framework_tutorial.py
```diff
@@ -0,0 +1,137 @@
+# [setup]
```
eundersander, line 1: This is already a very large PR. Can this tutorial be a separate PR? The problem with large PRs is that we hit reviewer fatigue and not everything gets properly reviewed.

src/esp/bindings/GeoBindings.cpp
```diff
@@ -64,6 +66,112 @@ void initGeoBindings(py::module& m) {
            1.0 is a chordal Catmull-Rom spline)",
       "key_points"_a, "num_interpolations"_a, "alpha"_a = .5f);
 
+  // ==== VoxelWrapper ====
```
eundersander, line 69: Could all these new bindings be a separate PR? This PR with core VoxelGrid implementation is already very large.

src/esp/bindings/SimBindings.cpp
```diff
@@ -73,97 +73,102 @@ void initSimBindings(py::module& m) {
...
-                    esp::metadata::MetadataMediator::ptr>())
-      .def("get_active_scene_graph", &Simulator::getActiveSceneGraph,
-           R"(PYTHON DOES NOT GET OWNERSHIP)",
-           py::return_value_policy::reference)
-      .def("get_active_semantic_scene_graph",
-           &Simulator::getActiveSemanticSceneGraph,
-           R"(PYTHON DOES NOT GET OWNERSHIP)",
-           py::return_value_policy::reference)
-      .def_property_readonly("semantic_scene", &Simulator::getSemanticScene, R"(
+  py::
```
eundersander, line 76: Did the code formatter really introduce this line break? Strange.

src/esp/geo/VoxelWrapper.h
```diff
@@ -3,11 +3,18 @@
 
 #include "VoxelGrid.h"
 #include "esp/assets/ResourceManager.h"
+#include "esp/gfx/DrawableGroup.h"
 #include "esp/scene/SceneNode.h"
 
 namespace esp {
 namespace geo {
 
+// When an object is voxelized, it is given a VoxelWrapper which points to an
```
eundersander, line 12: Will a class comment like this get picked up correctly by the doc-generator?

E.g. https://aihabitat.org/docs/habitat-sim/classesp_1_1assets_1_1ResourceManager.html

src/esp/geo/VoxelGrid.cpp
```diff
@@ -198,129 +106,143 @@ Mn::Vector3 VoxelGrid::getGlobalCoords(const Mn::Vector3i& coords) {
 }
 
 int VoxelGrid::generateBoolGridFromIntGrid(const std::string& intGridName,
-                                           std::string boolGridName,
+                                           const std::string& boolGridName,
```
eundersander, line 127: Could this be implemented using VoxelGrid's public interface? If so, consider making this a free function, to avoid class bloat. https://ict.senecacollege.ca/~oop244/pages/content/nonme.html

src/esp/geo/VoxelGrid.cpp
```diff
@@ -198,129 +106,143 @@ Mn::Vector3 VoxelGrid::getGlobalCoords(const Mn::Vector3i& coords) {
...
+          voxelSet.push_back(Mn::Vector3i(i, j, k));
+        }
+      }
     }
   }
+  return voxelSet;
 }
 
 // 6D SDF - labels each cell as interior (-inf), exterior (+inf), or boundary
 void VoxelGrid::generateInteriorExteriorVoxelGrid() {
```
eundersander, line 256: This is such a specific algorithm and specific application of VoxelGrid. From the class name, I would expect VoxelGrid to be a "dumb" container. Will this class grow and grow as we add more algorithms for operating on voxel grids? Can we instead make some kind of split here between the container and the algorithms? @aclegg3

eundersander (commented):
I made another pass here and I'm realizing I'm not a big fan of this monolithic VoxelGrid class implementing not just the container itself but also tons of algorithms and utilities that operate on it. For reference, consider STL containers and their very minimal public interfaces. I left some in-line comments along these lines.

I'm also not a fan of this monolithic PR. Can we break it up? I made some specific suggestions in in-line comments.

examples/tutorials/voxel_framework_tutorial.py
ScottCSteinhauser: Yeah that makes sense - I'll split up the tutorial and python bindings into another PR.

src/esp/geo/VoxelGrid.cpp
ScottCSteinhauser: I think transfering all algorithms to the voxel wrapper would make sense, since I can't really imagine the case when someone would have a VoxelGrid without it's wrapper and need to use these algorithms.

I'm a bit hesitant about making the algorithms free functions, since it might make the python API a bit less intuitive (right now all you do is get the object's voxelWrapper and can call all functions using that). Although I'm open to an alternative design. Thoughts @aclegg3 @mosra ?

src/esp/geo/VoxelGrid.h
eundersander: Bool vectors! This is neat.

src/esp/geo/VoxelGrid.cpp
aclegg3: @ScottCSteinhauser and I chatted about this a bit. Agreed that reducing bloat in the interface class would be ideal. We're thinking a voxel utilities module may be the best way to handle these sorts of applications/algorithms which can be implemented through the interface. Scott will break these out into a separate, incremental PR and we can focus on the interface first.

src/esp/sim/Simulator.h
```diff
@@ -501,6 +501,85 @@ class Simulator {
...
+   *
+   * If a voxel grid for the scene has not been created, it will make one with
+   * default arguments using createObjectVisualization().
+   *
+   * @param drawV Whether or not the render the voxel grid.
+   * @param sceneID !! Not used currently !! Specifies which physical scene of
+   * the object.
+   */
+  void setSceneVoxelizationDraw(bool drawV,
+                                const std::string& gridName = "Boundary");
```
aclegg3, line 566: `gridName` param undocumented

src/tests/VoxelGridTest.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "esp/geo/VoxelWrapper.h"
+#include "esp/sim/Simulator.h"
+
+#include "configure.h"
+
+namespace Cr = Corrade;
+namespace Mn = Magnum;
+
+// TODO: Add tests for different Sensors
```
aclegg3, line 16: remove?

src/tests/VoxelGridTest.cpp
```diff
@@ -0,0 +1,83 @@
...
+};
+
+VoxelGrid::VoxelGrid() {
+#ifdef ESP_BUILD_WITH_VHACD
+  addTests({&VoxelGrid::testVoxelGridWithVHACD});
+#endif
+}
+
+#ifdef ESP_BUILD_WITH_VHACD
+void VoxelGrid::testVoxelGridWithVHACD() {
```
aclegg3, line 30: Some of this PRs functionality (e.g. getters/setters, creating/remove grids, wrapper conversions, global grids, etc...) could be tested without VHACD. Would love to see these added to validate the code as we iterate.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,381 @@
...
+  }
+}
+
+void VoxelGrid::addVectorToMeshPrimitives(std::vector<Mn::Vector3>& positions,
+                                          std::vector<Mn::Vector3>& normals,
+                                          std::vector<Mn::Color3>& colors,
+                                          std::vector<Mn::UnsignedInt>& indices,
+                                          const Mn::Vector3i& local_coords,
+                                          const Mn::Vector3& vec) {
+  Mn::Trade::MeshData cubeData = Mn::Primitives::cubeSolid();
```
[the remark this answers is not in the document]
ScottCSteinhauser: Yeah I considered this - although I was unsure how to properly rotate the cone & cylinder to be in line with the vector? I can look more into that, but do you know of any shortcuts off the top of your head?

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+  return VoxelGridType::Float;
+}
+template <>
+VoxelGridType VoxelGrid::voxelGridTypeFor<Mn::Vector3>() {
+  return VoxelGridType::Vector3;
+}
+
+//  --== GETTERS AND SETTERS FOR VOXELS ==--
+
+std::vector<std::pair<std::string, std::string>> VoxelGrid::getExistingGrids() {
```
mosra, line 100: Is it really needed to have the type as a string? :) Why not return a `std::pair<std::string, VoxelGridType>`?

If you need a string representation for python bindings, there's a way to do it with an enum also.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+Mn::Vector3 VoxelGrid::getGlobalCoords(const Mn::Vector3i& coords) {
+  return Mn::Vector3(coords) * m_voxelSize + m_offset;
+}
+
+void VoxelGrid::fillBoolGridNeighborhood(std::vector<bool>& neighbors,
+                                         const std::string& gridName,
+                                         const Mn::Vector3i& index) {
+  std::vector<Mn::Vector3i> increments = {{0, 0, 1},  {1, 0, 0},  {0, 1, 0},
+                                          {0, 0, -1}, {0, -1, 0}, {-1, 0, 0}};
```
mosra, line 142: Can be just `Mn::Vector3i increments[] = {...}`, no need for a vector. Furthermore, it can be `constexpr`.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+  // VHACD computes a axis-aligned bounding box; we need to offset the voxelgrid
+  // by the minimum corner of the AABB
+  VHACD::Vec3<double> minBB = vhacdVolume->getMinBB();
+  m_offset = Mn::Vector3(minBB[0], minBB[1], minBB[2]);
+  m_BBMaxOffset = m_offset + Mn::Vector3(m_voxelGridDimensions) * scale;
+
+  // create empty VoxelGrid
+  addGrid<bool>("Boundary");
+  auto boundaryGrid = getGrid<bool>("Boundary");
```
mosra, line 52: I'm a caveman, but could you use explicit types where possible, instead of `auto`? Makes the code easier to read. (Here I know it's some sort of a strided view, but that's only because we discussed this code thoroughly before. If I would it see it again after a longer while I wouldn't know anymore.)

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+  Mn::Vector3 pos3 = mid + orthog2.normalized() * m_voxelSize * 1 / 15;
+  Mn::Vector3 pos4 = mid - orthog1.normalized() * m_voxelSize * 1 / 15;
+  Mn::Vector3 pos5 = mid - orthog2.normalized() * m_voxelSize * 1 / 15;
+
+  unsigned int sz = positions.size();
+  positions.push_back(pos1);
+  positions.push_back(pos2);
+  positions.push_back(pos3);
+  positions.push_back(pos4);
+  positions.push_back(pos5);
```
mosra, line 215: General comment for all the code in this function: it's possible to do 

```cpp
positions.insert(positions.end(), {
    pos1,
    pos2,
    ...
})
``` 

and it's also much faster that way.

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,392 @@
...
+        static_cast<std::ptrdiff_t>(sizeof(T))};
+
+    if (grids_.find(gridName) != grids_.end()) {
+      // grid exists, simply overwrite
+      Mn::Debug() << gridName << "exists, overwriting.";
+
+      grids_[gridName].data = Corrade::Containers::Array<char>(
+          Corrade::Containers::ValueInit, gridSize() * sizeof(T));
+
+      Cr::Containers::StridedArrayView<3, void> view{grids_[gridName].data,
```
mosra, line 102: FYI you can find&replace all `Cr::Containers::StridedArrayView<3, T>` with `Cr::Containers::StridedArrayView3D<T>`.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+#include "esp/assets/ResourceManager.h"
+
+namespace Mn = Magnum;
+namespace Cr = Corrade;
+
+namespace esp {
+namespace geo {
+
+#ifdef ESP_BUILD_WITH_VHACD
+VoxelGrid::VoxelGrid(const std::unique_ptr<assets::MeshData>& meshData,
```
mosra, line 22: The  `meshData` can't be null, right? Then maybe it would make sense to accept just 

```suggestion
VoxelGrid::VoxelGrid(const assets::MeshData& meshData,
```

instead.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+template <>
+VoxelGridType VoxelGrid::voxelGridTypeFor<Mn::Vector3>() {
+  return VoxelGridType::Vector3;
+}
+
+//  --== GETTERS AND SETTERS FOR VOXELS ==--
+
+std::vector<std::pair<std::string, std::string>> VoxelGrid::getExistingGrids() {
+  std::vector<std::pair<std::string, std::string>> existingGrids =
+      std::vector<std::pair<std::string, std::string>>();
```
mosra, line 102:
```suggestion
  std::vector<std::pair<std::string, std::string>> existingGrids;
```

The explicit initialization is only needed for trivial (builtin) types, such as int or float, to avoid having them uninitialized. All classes have their constructor called implicitly. (This inconsistency is something I started to really hate on C++.)

src/esp/geo/VoxelGrid.h
```diff
@@ -0,0 +1,392 @@
...
+#include "esp/geo/geo.h"
+#include "esp/gfx/Drawable.h"
+#include "esp/gfx/magnum.h"
+
+#ifdef ESP_BUILD_WITH_VHACD
+#include "VHACD.h"
+#endif
+
+namespace Mn = Magnum;
+namespace Cr = Corrade;
```
mosra, line 32: Hm, AFAIK we so far avoided these in the headers and used the full names to avoid alias pollution, only doing this in cpp files.

Opinions? @aclegg3

aclegg3: Yeah, that makes sense to me.

src/esp/geo/VoxelWrapper.h
```diff
@@ -0,0 +1,243 @@
...
+   * @param gridName The key underwhich the desired voxel grid is registered.
+   * @return A reference to the MeshGL.
+   */
+  Mn::GL::Mesh& getMeshGL(const std::string& gridName = "Boundary") {
+    return voxelGrid->getMeshGL(gridName);
+  }
+
+  /**
+   * @brief Sets the SceneNode of the voxel grid.
+   * @param coords The new sceneNode.
```
aclegg3, line 199: copy typo

src/esp/physics/PhysicsManager.h
```diff
@@ -1007,6 +1060,16 @@ class PhysicsManager {
                                      const std::string& handle,
                                      scene::SceneNode* objectNode);
 
+  /** @brief Set the voxelization visualization for a scene node to be true or
+   * false.
+   * @param drawables The drawables group with which to render the voxelization.
+   * @param drawVoxelization Set rendering of the voxelization to true or false.
+   */
+  void setVoxelizationDraw(const std::string& gridName,
+                           esp::physics::RigidBase* rigidBase,
```
aclegg3, line 1069: Missing param docstrings for "gridName" and "rigidBase"

src/esp/sim/Simulator.h
```diff
@@ -501,6 +501,83 @@ class Simulator {
...
+
+  /**
+   * @brief Returns the VoxelWrapper for a particular object.
+   * @return A shared ptr to the object's VoxelWrapper .
+   */
+  std::shared_ptr<esp::geo::VoxelWrapper> getStageVoxelization();
+
+  /**
+   * @brief Registers a voxel wrapper in a a dictionary.
```
aclegg3, line 573:
```suggestion
   * @brief Registers a voxel wrapper in a dictionary.
```

aclegg3, line 573: Maybe the tutorial obviates this, but consider some additional information about the purpose of the function here. Is it part of a user workflow for creating new global voxel fields? What does registration get you, why do it? Does it do something more?

src/utils/viewer/viewer.cpp
```diff
@@ -1093,7 +1164,23 @@ void Viewer::mousePressEvent(MouseEvent& event) {
...
+      esp::physics::RaycastResults raycastResults = simulator_->castRay(ray);
 
+      if (raycastResults.hasHits()) {
+        auto objID = raycastResults.hits[0].objectId;
+#ifdef ESP_BUILD_WITH_VHACD
+        displayVoxelField(objID);
+#endif
+      }
+    }
+  }
```
aclegg3, line 1183: Can we merge this with the other right click functionality below to avoid code duplication? Consider checking for the ctrl key and diverging after line 1192.

aclegg3 (commented):
A few more typos and suggestions. I'd still like to see a couple additional test cases related to manipulating and verifying VF field values through the getter/setter interface.
Edit: alos consider updating the vector visualization on the PR description with the arrows you added. 🙂

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,323 @@
...
+        gridName, Mn::MeshTools::compile(*meshDataDict_[gridName])));
+  }
+}
+
+void VoxelGrid::generateMesh(const std::string& gridName) {
+  assert(grids_.find(gridName) != grids_.end());
+  std::vector<Mn::UnsignedInt> indices;
+  std::vector<Mn::Vector3> positions;
+  std::vector<Mn::Vector3> normals;
+  std::vector<Mn::Color3> colors;
```
mosra, line 292: Okay let's do some fancy advanced C++ again and save a bunch of overhead that way :) Since there's the same count of positions, normals and colors and MeshData wants them interleaved for the best GPU performance, we could use that data layout right from the start. 

Step 1:

```cpp
// define it outside the function (or maybe in the header?) since addVector...() will need it too
struct VoxelVertex {
  Mn::Vector3 position;
  Mn::Vector3 normal;
  Mn::Vector3 color;
};
Cr::Containers::Array<VoxelVertex> vertices;
```

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,323 @@
...
+    meshGLDict_[gridName] = Mn::MeshTools::compile(*meshDataDict_[gridName]);
+  } else {
+    meshGLDict_.insert(std::make_pair(
+        gridName, Mn::MeshTools::compile(*meshDataDict_[gridName])));
+  }
+}
+
+void VoxelGrid::generateMesh(const std::string& gridName) {
+  assert(grids_.find(gridName) != grids_.end());
+  std::vector<Mn::UnsignedInt> indices;
```
mosra, line 289: Step 2 -- basically we just need to not use a std::vector because MeshData can't use it directly:

```cpp
Cr::Containers::Array<Mn::UnsignedInt> indices;
```

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,323 @@
...
+      positions,
+      {Mn::Trade::MeshAttributeData{Mn::Trade::MeshAttribute::Position,
+                                    Cr::Containers::arrayView(positions)}}};
+  meshDataDict_[gridName] =
+      std::make_shared<Mn::Trade::MeshData>(Mn::MeshTools::interleave(
+          positionMeshData,
+          {Mn::Trade::MeshAttributeData{Mn::Trade::MeshAttribute::Color,
+                                        Cr::Containers::arrayView(colors)},
+           Mn::Trade::MeshAttributeData{Mn::Trade::MeshAttribute::Normal,
+                                        Cr::Containers::arrayView(normals)}}));
```
mosra, line 275: Step 3, instead of all this, we can populate the mesh data directly:

```cpp
void VoxelGrid::generateMeshDataAndMeshGL(const std::string& gridName,
                                          Cr::Containers::Array<Mn::UnsignedInt>&& indexData, // note the &&s (!)
                                          Cr::Containers::Array<VoxelVertex>&& vertexData) {
  // need to make the index/attribute views first, before calling std::move() on the data
  Mn::Trade::MeshIndexData indices{indexData}; 
  Mn::Trade::MeshAttributeData positions{ 
      Mn::Trade::MeshAttribute::Position,
      Cr::Containers::stridedArrayView(vertexData).slice(&VoxelVertex::position)}; // view on just the position member
  Mn::Trade::MeshAttributeData colors{ 
      Mn::Trade::MeshAttribute::Color,
      Cr::Containers::stridedArrayView(vertexData).slice(&VoxelVertex::color)}; // view on just the color member
  Mn::Trade::MeshAttributeData normals{ 
      Mn::Trade::MeshAttribute::Normal,
      Cr::Containers::stridedArrayView(vertexData).slice(&VoxelVertex::normal)}; // view on just the normal member
      
  // now move the data into a MeshData, together with all metadata describing indices and attributes
  Mn::Trade::MeshData data{
      Mn::MeshPrimitive::Triangles,
      // the cast takes an Array<UnsignedInt> and turns it into an Array<char> without having to copy
      // everything again
      Containers::arrayAllocatorCast<char>(std::move(indexData)), indices,
      Containers::arrayAllocatorCast<char>(std::move(vertexData)), {positions, colors, normals}};
      
  // and save those, again move because the neither the Array nor MeshData are copyable
  meshDataDict_[gridName] =
      std::make_shared<Mn::Trade::MeshData>(std::move(data));
```

The comments are there mainly for you to understand better what's going on, feel free to drop them / rewrite as needed.

- [Docs for `arrayAllocatorCast()`](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Containers.html#a46c91612607a6ce5a823322d05666879)
- [Docs for the weird member `slice()`](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1StridedArrayView.html#ababb10360346c92e016188824b54d61d)

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,323 @@
...
+    positions.push_back(
+        vecRotation.transformVector(ind * Mn::Vector3(0.02, 0.04, 0.02) +
+                                    Mn::Vector3(0, 0.025, 0)) +
+        mid);
+    colors.push_back(Mn::Color3(0, .3, 1));
+  }
+
+  for (const auto& norm : coneNormals) {
+    normals.push_back(norm);
+  }
```
mosra, line 220: Step 4a -- assuming you change the function to take `Cr::Containers::Array<VoxelVertex>& vertices` instead of the three vectors, these three loops can be merged. The arrays have a slightly different "growing" behavior and API than `std::vector` ([docs]()), and you need to `#include <Corrade/Containers/GrowableArray.h>` for that:

```cpp
for(std::size_t i = 0; i != coneData.vertexCount(); ++i) {
    arrayAppend(vertices, Containers::InPlaceInit, 
        vecRotation.transformVector(conePosition[i] * Mn::Vector3(0.02, 0.04, 0.02) +
                                    Mn::Vector3(0, 0.025, 0)) + mid,
        coneNormals[i],
        Mn::Color3{0.0f, 0.3f, 1.0f}
    );
}
```

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+  for (const auto&& index : coneIndices) {
+    indices.push_back(sz + index[0]);
+  }
```
mosra, line 224: Step 4b -- and assuming you change the function to take `Cr::Containers::Array<Mn::UnsignedInt>& indices` instead of the vector, this can be 

```cpp
  for (const UnsignedInt index : coneIndices) {
    arrayAppend(indices, sz + index);
  }
```

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+  for (const auto& norm : cylinderNormals) {
+    Mn::Vector3 transformedNorm =
+        vecRotation.transformVector(norm * Mn::Vector3(0.01, 0.04, 0.01));
+    normals.push_back(transformedNorm);
+  }
+
+  for (const auto&& index : cylinderIndices) {
+    indices.push_back(sz + index[0]);
+  }
```
mosra, line 250: Same as the Step 4a / 4b here as well.

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,323 @@
...
+    colors.push_back(color);
+  }
+
+  for (int i = 0; i < 6; i++) {
+    if (!neighbors[i]) {
+      for (int j = 0; j < 6; j++) {
+        indices.push_back(sz + cubeIndices[i * 6 + j][0]);
+      }
+    }
+  }
```
mosra, line 180: And this also changed similarly as the Step 4a/4b mentioned before, together with the function signature update.

mosra (commented):
Ok here's the promised suggestion about populating the MeshData. Have fun! (I hope it's fun, haha.)

aclegg3 (approved):
When the tests pass, let's ship it. 👍

PR #1141, opened 2021-03-15
[Sensor] Equirectangular sensor

vauduong:
## Motivation and Context

Given a cubemap, implement an Equirectangular Sensor and shader
Screen Shot 2021-03-15 at 3 48 53 AM

## How Has This Been Tested

Build and test

## Types of changes

## Checklist


bigbike:
I see a lot of code duplications between the `Equirectangular` and the `FisheyeSensor`. 
Maybe `Equirectangular` should be sub-classed from `FisheyeSensor`?

Skylion007:
In Habitat-Lab, we had them use a shared base class called CameraProjection.

Skylion007:
https://github.com/facebookresearch/habitat-lab/blob/f89d53ea47e784fdf6e78fadccded645bb5913a7/habitat_baselines/common/obs_transformers.py#L226

vauduong:
> I see a lot of code duplications between the `Equirectangular` and the `FisheyeSensor`.
> Maybe `Equirectangular` should be sub-classed from `FisheyeSensor`?

That's a good point, only the frag shader and the uniforms are different. I think it would make more sense for EquirectangularSensor and FisheyeSensor to be subclassed from `CubeMapSensor` though, since that is the underlying connection. What do you think?

bigbike:
`FisheyeSensor` is already a base class.
Let `EquirectangularSensor` inherit from `FisheyeSensor` (we can think about the name in the future, but leave it like this for now).

vauduong:
> `FisheyeSensor` is already a base class.
> Let `EquirectangularSensor` inherit from `FisheyeSensor` (we can think about the name in the future, but leave it like this for now).

Ok can do! Thanks for the clarification!

src/shaders/equirectangular.frag
```diff
@@ -0,0 +1,38 @@
...
+#endif
+
+// ------------ output -----------------------
+#if defined(COLOR_TEXTURE)
+layout(location = OUTPUT_ATTRIBUTE_LOCATION_COLOR) out vec4 fragmentColor;
+#endif
+
+void main(void) {
+  vec3 m;
+  float PI = 3.1415926535897932384626433832795;
```
Skylion007, line 25: Shouldn't this be a const?

Skylion007:
Would appreciate a pass by @mosra at some point to double check we aren't doing anything silly with Magnum and/or Shaders. Also @vauduong your test case .png file is empty which is probably not what you wanted.

vauduong:
Thanks for the catch @Skylion007 ! That's been fixed. 

Making EquirectangularSensor subclass FisheyeSensor will take some more refactoring.

mosra:
I'm confused. I see a lot of extremely similar code between this and #1131. What's the relation of those two? Don't they mutually conflict with each other?

bigbike:
@mosra: #1131 will be checked in first. Then this PR will rebase on top it.

Skylion007:
@vauduong #1131 is merged in. Please rebase.

bigbike:
The `viewer` really needs to be refactored.
Will let #1066 be committed in first.

bigbike:
close it since we have #1179

PR #1146, opened 2021-03-19
Physics bindings for JS build

eundersander:
## Motivation and Context

I added physics bindings, including support for adding/removing objects, setting their transforms, and stepping physics.

I also added a hacky test framework for testing the JS bindings. See below.

## How Has This Been Tested

Ug, what a mess:
- I'm unable to properly/cleanly test the JS bindings. I didn't see any obvious way for yarn/jest tests to invoke our webassembly emscripten-built Module.
- ~I'm also unable to build JS with bullet enabled in CI, see [build output](https://app.circleci.com/pipelines/github/facebookresearch/habitat-sim/5450/workflows/9bd6dbba-78b8-4fdb-afd3-f513429ab150/jobs/24505/parallel-runs/0/steps/0-121)~ Update: I've updated CI to build JS with bullet, including newer CMake
 
We should fix these issues, but they're out of scope for this PR.

For this PR, I added a hacky framework for testing the JS bindings. See load_test_page.test.js.

Examples of CI output for this test [failing](https://app.circleci.com/pipelines/github/facebookresearch/habitat-sim/5449/workflows/46770d11-2d85-4ba4-8ec5-08aef422fb1e/jobs/24503/parallel-runs/0/steps/0-123) and [passing](https://app.circleci.com/pipelines/github/facebookresearch/habitat-sim/5451/workflows/4d91d36a-f776-4f21-b45f-ec7f07ca8b3c/jobs/24511/parallel-runs/0/steps/0-123).

I've found that Puppateer crashes intermittently in my local testing so I'm curious if it will be robust in CI. I'll rerun the tests a few times before merging.

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -476,7 +476,7 @@ jobs:
               . activate habitat
               nvm use v11.9.0
               . ~/emsdk/emsdk_env.sh
-              ./build_js.sh
+              ./build_js.sh --bullet
       - run:
           name: Run sim benchmark
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -476,7 +476,7 @@ jobs:
               . activate habitat
               nvm use v11.9.0
               . ~/emsdk/emsdk_env.sh
-              ./build_js.sh --bullet
+              ./build_js.sh
       - run:
           name: Run sim benchmark
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,14 +22,18 @@ commands:
           no_output_timeout: 5m
           # tinyply requires CMake 3.10, so stay at that version. Ubuntu 16.04
           # has 3.5, so once tinyply is removed, we could go lower (well, if
-          # anybody actually needs that)
+          # anybody actually needs that). We also download CMake 3.20, which
+          # we'll use later for the JS build only.
           command: |
               echo $(git ls-remote https://github.com/facebookresearch/habitat-lab.git HEAD | awk '{ print $1}') > ./hablab_sha
               cat ./hablab_sha
               wget https://cmake.org/files/v3.10/cmake-3.10.3-Linux-x86_64.sh
-              sudo mkdir /opt/cmake
-              sudo sh ./cmake-3.10.3-Linux-x86_64.sh --prefix=/opt/cmake --skip-license
-              sudo ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake
+              wget https://cmake.org/files/v3.20/cmake-3.20.1-linux-x86_64.sh
+              sudo mkdir /opt/cmake310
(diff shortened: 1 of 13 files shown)


Skylion007 (changes requested):
The tests are failing because this adds a hard dependency on Bullet. The bindings need to support building the emscriptem build without Bullet.

Also some tests of the Bullet functions would be nice (even just to make sure they don't error).

aclegg3:
> Also some tests of the Bullet functions would be nice (even just to make sure they don't error).

Should be enough to `addObjectByHandle("cubeSolid")` to create a cube and then `stepWorld()`. 
You may also want to add `getWorldTime()` to the bindings for sanity checks against expected time-stepping.

aclegg3 (approved):
Looks good once CI build issues are resolved.

src/esp/bindings_js/index.js
```diff
@@ -23,7 +23,7 @@ function preload(url) {
   if (url.indexOf("http") === -1) {
     let file_parents = splits.slice(0, splits.length - 1);
     for (let i = 0; i < splits.length - 1; i += 1) {
-      file_parents_str += file_parents[i] + "/";
+      file_parents_str += file_parents[i];
```
eundersander, line 26: This was not failing on my local Ubuntu but was failing for CI (FS runtime error), maybe due to differences in emscripten version or browser.

eundersander:
I will add some units tests here before merging.

src/esp/bindings_js/index.js
Skylion007: Hold on, this will break any assets in subfolders if you make this change. (It will work if the asset is not in  subfolder because the splits is 0. Are you sure the physics JSON is copied correctly during the Build Process? Will assets in subfolder still work? I don't think we actually test that in the CI so that breakage wouldn't be detected by this "fix". Won't this just instead create a single folder that mashes together the asset path names? Ie: dataphysicsconfig.json instead of data/physicsconfig.json?

eundersander:
See updated PR description for notes about the tests added here.

Skylion007 (approved):

Skylion007 (changes requested):
One last issue, CircleCI never actually builds the Web build with Bullet so the Bullet tests aren't run. We need to test it without/wo physics like the python build. Once that is taken care of this PR should be good to go.

.circleci/config.yml
```diff
@@ -476,7 +476,7 @@ jobs:
               . activate habitat
               nvm use v11.9.0
               . ~/emsdk/emsdk_env.sh
-              ./build_js.sh --bullet
+              ./build_js.sh
```
Skylion007: Requiring a newer CMake for Bullet tests seem fine. Let's ask in Slack though.

Skylion007 (approved):

Issue #1081, opened 2021-02-09
Access semantic scene mesh directly


rfalcon100:
## ❓ Questions and Help
Hi,

Is it possible to get access to the full scene mesh with semantic annotations, instead of just rendering images from the sensors?
In particular, I am using the replica dataset, and I want to parse the scene into a list of surfaces (preferably triangles), where each triangle is assigned to a semantic class.

I have tried reading the `mesh.ply` and `/habitat/mesh_semantic.ply` files with a few mesh loading tools like [https://github.com/nschloe/meshio](url) or [https://github.com/intel-isl/Open3D](url), but I can only get point clouds and not a list of surfaces. 
Is there a way to do this directly from Habitat-sim?

aclegg3:
Unfortunately we don't provide an mesh-level access in the API at this time. 

If you are comfortable with C++ you could try adding this support and writing a PR or forking and adapting our code to fit your needs. You may be most interested in [this file](https://github.com/facebookresearch/habitat-sim/blob/master/src/esp/assets/GenericInstanceMeshData.cpp) where .ply parsing is done and semantic meshes are split by object id as you are hoping to do.

Skylion007:
We have some bindings to access the NavMesh directly, feel free to look
there for examples about how this could be done.

rfalcon100:
Thanks for the info, I'll see what I can do.
But I am a bit confused, should parse the `mesh_semantic.ply` or the navmesh?

Skylion007:
@rfalcon100 Referring to add new bindings to access the underling semantic mesh from the MeshData like so: https://github.com/facebookresearch/habitat-sim/blob/da03230101ca7b7b77ea5df90b8a74806d413942/src/esp/bindings/ShortestPathBindings.cpp#L100

rfalcon100:
It seems, I am am able to get the vertices , surface, and semantic object data that I need.
 However, I would like to get the surface normals too. 
 It seems like normals are not store in the GenericInstanceMeshData object:
 https://github.com/facebookresearch/habitat-sim/blob/87364fd0f7f36589f055b61fa41a7199d8ee4c97/src/esp/assets/GenericInstanceMeshData.cpp#L161
 
Would it make sense to store and expose the normals too?

neepacmu:
@rfalcon100 can you share exactly how did you get the vertices , surface, and semantic object data? Which files did you parse?

mahnoor-fatima-saad:
[...]

PR #1149, opened 2021-03-22
Access semantic scene mesh directly

rantahar:
## Motivation and Context

We have implemented the ability to directly access mesh data as discussed in issue #1081.

This required adding methods to the ResourceManager class and the Simulator class. We added python bindings
to the methods in the Simulator class.

## How Has This Been Tested

We added two new tests to tests/test_simulator.py, test_mesh_point_extraction and 
test_mesh_point_extraction_when_not_instance. The first one uses a datafile not 
present in the repository and is marked as skipped. It tests that the methods run and
return vectors with the correct length.

The other test check that the methods throw an exception when the meshes are not
of the GenericInstanceMeshData type.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.h
+++ b/src/esp/assets/ResourceManager.h
@@ -275,35 +275,58 @@ class ResourceManager {
   /**
    * @brief Return a vector of the vertex position in a given mesh.
    *
-   * Given the integer ID of a mesh, get the vertex positions in that mesh.
+   * Given the integer ID of a semantic mesh, get the vertex positions in that
+   * mesh.
    * @param id The number index of a mesh.
    * @return A vector of vertex positions.
    */
   std::vector<vec3f> getVertices(int id) {
-    const std::vector<vec3f> vertexPositions =
-          dynamic_cast<GenericInstanceMeshData&>(*meshes_.at(id))
-              .getVertexBufferObjectCPU();
-    return vertexPositions;
+    BaseMesh &mesh = *meshes_.at(id);
+    if(mesh.getMeshType() == SupportedMeshType::INSTANCE_MESH){
+      // It is an instance mesh and it is possible to return the vertex data
+      const GenericInstanceMeshData &instMesh = dynamic_cast<GenericInstanceMeshData&>(mesh);
+      const std::vector<vec3f> vertexPositions = instMesh.getVertexBufferObjectCPU();
+      return vertexPositions;
+
+    } else {
+      // The mesh is not an instance mesh
+      throw "Error in getVertices. The mesh is not an instance mesh.";
+      const std::vector<vec3f> vertexPositions;
+    }
   }
 
   /**
    * @brief Return a vector of the surface indexes for a specific mesh in the scene.
    */
   std::vector<uint32_t> getSurfIndexes(int id) {
-    const std::vector<uint32_t> surfIndexes =
-          dynamic_cast<GenericInstanceMeshData&>(*meshes_.at(id))
-              .getIndexBufferObjectCPU();
-    return surfIndexes;
+    BaseMesh &mesh = *meshes_.at(id);
+    if(mesh.getMeshType() == SupportedMeshType::INSTANCE_MESH){
+      // It is an instance mesh and it is possible to return the vertex data
+      const GenericInstanceMeshData &instMesh = dynamic_cast<GenericInstanceMeshData&>(mesh);
+      const std::vector<uint32_t> surfIndexes = instMesh.getIndexBufferObjectCPU();
+      return surfIndexes;
+
+    } else {
+      // The mesh is not an instance mesh
+      throw "Error in getSurfIndexes. The mesh is not an instance mesh.";
+    }
   }
 
   /**
    * @brief Return a vector of color per vertex for a specific mesh in the scene.
    */
   std::vector<vec3uc> getVerticesColor(int id) {
-    const std::vector<vec3uc> colors =
-          dynamic_cast<GenericInstanceMeshData&>(*meshes_.at(id))
-              .getColorBufferObjectCPU();
-    return colors;
+    BaseMesh &mesh = *meshes_.at(id);
+    if(mesh.getMeshType() == SupportedMeshType::INSTANCE_MESH){
+      // It is an instance mesh and it is possible to return the vertex data
+      const GenericInstanceMeshData &instMesh = dynamic_cast<GenericInstanceMeshData&>(mesh);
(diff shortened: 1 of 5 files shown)


tests/test_simulator.py
```diff
@@ -246,3 +247,49 @@ def test_object_template_editing():
...
             assert obj_init_template.render_asset_handle.endswith("sphere.glb")
+
+
+@pytest.mark.skip(reason="the test requires data not present on the repository")
+def test_mesh_point_extraction():
+    """Test the interface for retrieving mesh data from instance meshes"""
+    cfg_settings = examples.settings.default_sim_settings.copy()
+    cfg_settings["scene"] = "office_0/habitat/mesh_semantic.ply"
+    hab_cfg = examples.settings.make_cfg(cfg_settings)
+    sim = habitat_sim.Simulator(hab_cfg)
```
Skylion007, line 258: NIT: sim should be constructed using a with context so it is properly closed.

tests/test_simulator.py
```diff
@@ -246,3 +247,49 @@ def test_object_template_editing():
...
+    assert len(surface_ids) == 76908
+
+
+def test_mesh_point_extraction_when_not_instance():
+    """Run the mesh extraction methods when the mesh is not
+    an instance mesh. This should result in an exception."""
+    cfg_settings = examples.settings.default_sim_settings.copy()
+    cfg_settings["scene"] = "data/test_assets/scenes/simple_room.glb"
+    hab_cfg = examples.settings.make_cfg(cfg_settings)
+    sim = habitat_sim.Simulator(hab_cfg)
```
Skylion007, line 281: Same as above.

Skylion007 (changes requested):
Please fix the two minor issues in the test cases.

src/esp/assets/ResourceManager.h
```diff
@@ -271,6 +272,106 @@ class ResourceManager {
...
 
+  /**
+   * @brief Return a vector of the vertex position in a given mesh.
+   *
+   * Given the integer ID of a semantic mesh, get the vertex positions in that
+   * mesh.
+   * @param id The number index of a mesh.
+   * @return A vector of vertex positions.
+   */
+  std::vector<vec3f> getVertices(int id) {
```
erikwijmans, line 283: Please move the function implementations to the `.cpp` file

src/esp/assets/ResourceManager.h
```diff
@@ -271,6 +272,106 @@ class ResourceManager {
...
+    }
+  }
+
+  /**
+   * @brief Return a vector of the integer IDs of existing meshes.
+   *
+   * Return a vector of the integer IDs of existing meshes.
+   * @return a vector of existing mesh IDs.
+   */
+  std::vector<int> getMeshKeys() {
```
erikwijmans, line 366: This will return the mesh keys for every mesh type, which is fine when there is just 1 mesh type, but for mp3d and gibson semantics, those have two mesh types.  Perhaps this should take a parameter to return keys for only a specific mesh type?  Alternatively, this could be renamed `getInstanceMeshKeys` and return just the keys for instance meshes.

src/esp/assets/ResourceManager.h
```diff
@@ -271,6 +272,106 @@ class ResourceManager {
...
+   * @brief Return a vector of object IDs for each vertex in a given mesh.
+   *
+   * Given the integer ID of a mesh, get the object IDs of each vertex in
+   * the mesh.
+   * @param id The number index of a mesh.
+   * @return A vector of object IDs.
+   */
+  std::vector<uint16_t> getObjectIds(int id) {
+    BaseMesh& mesh = *meshes_.at(id);
+    if (mesh.getMeshType() == SupportedMeshType::INSTANCE_MESH) {
```
erikwijmans, line 347: Use `ESP_CHECK` here instead of throwing

src/esp/bindings/SimBindings.cpp
```diff
@@ -323,12 +323,27 @@ void initSimBindings(py::module& m) {
...
           "object_id"_a, "light_setup_key"_a, "scene_id"_a = 0,
           R"(Modify the LightSetup used to the render all components of an object by setting the LightSetup key referenced by all Drawables attached to the object's visual SceneNodes.)")
-
       .def(
           "get_num_active_contact_points",
           &Simulator::getNumActiveContactPoints,
-          R"(The number of contact points that were active during the last step. An object resting on another object will involve several active contact points. Once both objects are asleep, the contact points are inactive. This count can be used as a metric for the complexity/cost of collision-handling in the current scene.)");
-  ;
+          R"(The number of contact points that were active during the last step. An object resting on another object will involve several active contact points. Once both objects are asleep, the contact points are inactive. This count can be used as a metric for the complexity/cost of collision-handling in the current scene.)")
+      .def("get_vertices", &Simulator::getVertices, "ID"_a,
```
erikwijmans, line 330:
```suggestion
      .def("get_vertices", &Simulator::getVertices, "id"_a,
```

erikwijmans: Lower case, or `mesh_key` or something if you are trying to avoid the builtin `id`

erikwijmans (changes requested):

rantahar:
Sorry for loosing track of this for a long time. I'm updating it now to the current main.

src/esp/assets/ResourceManager.h
rantahar: Renamed to getInstanceMeshKeys and changed to only return instance mesh keys.

rantahar: Renamed to getInstanceMeshKeys and changed to only return instance mesh keys.

Skylion007:
You are going to need pull / rebase from master. Assigning @jturner65 to review since we now support more than just semantic  meshes, but semantic textures / vertexes etc.

jturner65:
> I rebased on main before making the requested changes.
> 
> Unfortunately I see some changes in the submodules in my PR branch. Any ideas on what I can do about this?

When you rebased you committed your local changes to the submodules, which reverted them to an older state.  Always discard submodule changes when you rebase unless you explicitly are changing the submodules :).

cuizhuyefei:
@rantahar Hello, I also encounter the problem of accessing mesh directly. After cloning your repo https://github.com/AaltoRSE/habitat-sim, unfortunately, I cannot use APIs like `sim.get_instance_mesh_keys()` shown in test_simulator.py. I wonder how the pr is going(it is not in the master branch yet) and also how I can use your code to access mesh. I will really appreciate it if you can provide some help!

PR #1151, opened 2021-03-23
[Sensor][WIP] Python SensorSuite in each SceneNode

vauduong:
## Motivation and Context
PR #1118 introduces SensorSuites in each SceneNode. 
In this PR:

- Deprecate SensorSuite and Sensor wrapper in python simulator code
- Note, this builds locally but not on ci, currently trying to debug

## How Has This Been Tested

Run tests

## Types of changes

## Checklist


src/esp/bindings/SensorBindings.cpp
```diff
@@ -37,6 +38,94 @@ esp::scene::SceneNode* nodeGetter(T& self) {
...
+// Get the torch tensor or numpy array buffer of the Sensor, initialize if it
+// does not exist yet
+template <class T>
+
+auto buffer(T& self, int gpuDevice) {
+  py::handle handle = py::cast(self);
+  if (!py::hasattr(handle, "__buffer")) {
+    if (self.specification()->gpu2gpuTransfer) {
+#ifdef ESP_BUILD_WITH_CUDA
+      auto torch = py::module_::import("torch");
```
Skylion007, line 50: Torch isn't a hard dependency even with CUDA enabled. Not sure if throws a graceful error, can you check?

src/esp/bindings/SensorBindings.cpp
```diff
@@ -37,6 +38,94 @@ esp::scene::SceneNode* nodeGetter(T& self) {
...
+                                 py::int_(self.specification()->resolution[1]),
+                                 py::int_(self.specification()->channels)),
+                                "dtype"_a = torch.attr("uint32"),
+                                "device"_a = torch.attr("device")(
+                                    py::str("cuda"), py::int_(gpuDevice))));
+      }
+#endif
+    } else {
+      if (self.specification()->sensorType == SensorType::Semantic) {
+        auto pyBuffer = py::array(py::buffer_info(
```
Skylion007, line 81: This part isn't even necessary, you can return the array plainly by reference. @mosra Thoughts? Also not sure setting the buffer here like this is the appropriate way to do it in pyBind.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -125,12 +243,15 @@ void initSensorBindings(py::module& m) {
...
+      .def_property_readonly("render_target", &VisualSensor::renderTarget)
+      .def(
+          "buffer", buffer<VisualSensor>,
+          R"(Get the torch tensor or numpy array buffer of the Sensor, initialize if it does not exist yet)");
 
   // === CameraSensor ====
   py::class_<CameraSensor, Magnum::SceneGraph::PyFeature<CameraSensor>,
              VisualSensor, Magnum::SceneGraph::PyFeatureHolder<CameraSensor>>(
-      m, "CameraSensor")
+      m, "CameraSensor", py::dynamic_attr())
```
Skylion007, line 254: Why do we need dynamic_attrs here, still?

Skylion007:
This changes some whitehot code in habitat_sim so we need to make sure this doesn't cause any serious performance regressions.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -37,6 +38,94 @@ esp::scene::SceneNode* nodeGetter(T& self) {
...
+                                                 for each dimension */
+            {sizeof(uint8_t), sizeof(uint8_t), sizeof(uint8_t)}
+            /* Strides for each dimension */
+            ));
+        py::setattr(handle, "__buffer", pyBuffer);
+      }
+    }
+  }
+  return py::getattr(handle, "__buffer");
+};
```
mosra, line 127: Wait, is this all new code? If it is, how was this handled previously? I'm confused.

In my opinion Habitat isn't a place to do such low level buffer protocol operations, there needs to be *extensive* testing for such pieces of code because a lot of things can go wrong -- and that's why it's all already implemented in magnum bindings. Zero-copy array transfer via a buffer protocol was originally done through magnum's image views, did that change now? Why does the old way not work anymore?

Skylion007:
It's an attempt to deduplicate and generalize this Python class into one
that uses pure C++ calls:
https://github.com/facebookresearch/habitat-sim/blob/3c7d566c00d6c0c15165cbdf9c4fc791c44397de/habitat_sim/simulator.py#L476

src/esp/bindings/SensorBindings.cpp
Skylion007: It's an attempt to deduplicate and generalize this Python class into one
that uses pure C++ calls:
https://github.com/facebookresearch/habitat-sim/blob/3c7d566c00d6c0c15165cbdf9c4fc791c44397de/habitat_sim/simulator.py#L476

src/esp/bindings/SensorBindings.cpp
vauduong: Thanks @Skylion007 , it builds without errors even without the #ifdef

src/esp/bindings/SensorBindings.cpp
vauduong: Thanks for the catch!

src/esp/bindings/SensorBindings.cpp
vauduong: Hmm, if we return the array by reference, this would mean that the array would be created by this function every time get_observation is called in python right? We're trying to store it as an attribute to avoid this.

habitat_sim/simulator.py
```diff
@@ -472,166 +472,51 @@ def __del__(self) -> None:
...
-        tgt = self._sensor_object.render_target
-
-        if self._spec.gpu2gpu_transfer:
-            with torch.cuda.device(self._buffer.device):  # type: ignore[attr-defined]
-                if self._spec.sensor_type == SensorType.SEMANTIC:
-                    tgt.read_frame_object_id_gpu(self._buffer.data_ptr())  # type: ignore[attr-defined]
-                elif self._spec.sensor_type == SensorType.DEPTH:
-                    tgt.read_frame_depth_gpu(self._buffer.data_ptr())  # type: ignore[attr-defined]
+    def get_observation(self, sensor: Sensor) -> Union[ndarray, "Tensor"]:
+        self.renderer.bind_render_target(sensor)
```
erikwijmans, line 476: This cannot be called here as that allocates and binds the render target, it should be called once and only once per sensor.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -188,8 +189,102 @@ void initSensorBindings(py::module& m) {
           R"(The distance to the near clipping plane for this CameraSensor uses.)")
       .def_property(
           "far_plane_dist", &CameraSensor::getFar, &CameraSensor::setFar,
-          R"(The distance to the far clipping plane for this CameraSensor uses.)");
-
+          R"(The distance to the far clipping plane for this CameraSensor uses.)")
```
erikwijmans, line 192: Given that we still need something on the python side for noise models, this can also be done in python.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -188,8 +189,102 @@ void initSensorBindings(py::module& m) {
...
+                } else if (self.specification()->sensorType ==
+                           esp::sensor::SensorType::Depth) {
+                  py::setattr(
+                      handle, "__buffer",
+                      torch.attr("empty")(
+                          (py::int_(self.specification()->resolution[0]),
+                           py::int_(self.specification()->resolution[1])),
+                          "dtype"_a = torch.attr("float32"),
+                          "device"_a = torch.attr("device")(py::str("cuda"),
+                                                            py::int_(0))));
```
erikwijmans, line 219: The device ID is wrong here.  Is should be `sim.gpu_device`

erikwijmans (changes requested):

src/esp/bindings/SensorBindings.cpp
erikwijmans: To follow up on this, given that we still need something on the python side for noise models, this can also be done in python.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -188,8 +190,102 @@ void initSensorBindings(py::module& m) {
...
+            py::handle handle = py::cast(self);
+            if (!py::hasattr(handle, "__buffer")) {
+              if (self.specification()->gpu2gpuTransfer) {
+                auto torch = py::module_::import("torch");
+                if (self.specification()->sensorType ==
+                    esp::sensor::SensorType::Semantic) {
+                  py::setattr(
+                      handle, "__buffer",
+                      torch.attr("empty")(
+                          (py::int_(self.specification()->resolution[0]),
```
Skylion007, line 206: huh this works? you don't need an explicit py::make_tuple?

src/esp/bindings/SensorBindings.cpp
```diff
@@ -188,8 +190,102 @@ void initSensorBindings(py::module& m) {
           R"(The distance to the near clipping plane for this CameraSensor uses.)")
       .def_property(
           "far_plane_dist", &CameraSensor::getFar, &CameraSensor::setFar,
-          R"(The distance to the far clipping plane for this CameraSensor uses.)");
-
+          R"(The distance to the far clipping plane for this CameraSensor uses.)")
+      .def(
+          "buffer",
+          [](CameraSensor& self, int gpuDevice) {
+            py::handle handle = py::cast(self);
```
Skylion007, line 197: Why do you need to cast this back to a python object? It should automatically be cast when called on any relevant python function. Also the naming is a tad nondescript.

PR #1152, opened 2021-03-23
--Scene Dataset Config paths glob/wildcard handling

jturner65:
## Motivation and Context
This PR adds the ability to provide glob wildcards in the scene dataset configuration for the paths to specific configs, enabling wild-card-based searches and loads for configs that may be spread across many directories, such as per-stage paths that exist in Mp3d and HM3D datsets.  See the newly added config file data/test_assets/dataset_tests/dataset_1/test_dataset_1.scene_dataset_config.json for usage examples.

## How Has This Been Tested
C++ and python tests pass

## Types of changes

## Checklist


src/esp/io/io.h
```diff
@@ -5,6 +5,7 @@
 #ifndef ESP_IO_IO_H_
 #define ESP_IO_IO_H_
 
+#include <glob.h>
```
mosra, line 8: Can be in the cpp instead :)

src/esp/io/io.h
```diff
@@ -19,6 +20,14 @@ std::string removeExtension(const std::string& file);
 
 std::string changeExtension(const std::string& file, const std::string& ext);
 
+/**
+ * @brief This function will perform globbing on a passed path + pattern,
+ * returning all the files and directories that match the pattern.
+ * @param pattern The pattern to match
```
mosra, line 26: Maybe mention what is allowed in the pattern? As in, it's just wildcards and not full regular expressions.

mosra (commented):
The glob() API usage is simpler than I expected, nice.

src/esp/io/io.h
jturner65: Is there a way in the docs to put an external html link? (I was thinking of linking wiki :D )

mosra: It's a standard Markdown, so `[title](url)` works. Or just put an URL there directly.

Skylion007 (approved):
Looks pretty good to me.

PR #1153, opened 2021-03-24
Add custom background clear color for COLOR sensors.

aclegg3:
## Motivation and Context

Some applications (e.g. rendering an object mask in the empty scene) require a custom background color for particular sensors. This change adds the `clearColor`(C++)|`clear_color`(Py) property to `VisualSensorSpec` and the necessary tooling to consume it in C++ and python rendering pathways.

## How Has This Been Tested

Local testing in viewer and python:
Example: `sensor_spec.clear_color = mn.Color4(0.4,0,0.6,1)` in physics_benchmarking.py
Screen Shot 2021-03-24 at 2 10 03 PM

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -18,7 +18,7 @@
 #include <Magnum/PixelFormat.h>
 
 #include "RenderTarget.h"
-#include "esp/sensor/CameraSensor.h"
+#include "esp/sensor/VisualSensor.h"
 #include "magnum.h"
 
 #include "esp/gfx/DepthUnprojection.h"
@@ -46,7 +46,8 @@ struct RenderTarget::Impl {
   Impl(const Mn::Vector2i& size,
        const Mn::Vector2& depthUnprojection,
        DepthShader* depthShader,
-       Flags flags)
+       Flags flags,
+       const sensor::VisualSensor* visualSensor)
       : colorBuffer_{},
         objectIdBuffer_{},
         depthRenderTexture_{},
@@ -56,7 +57,8 @@ struct RenderTarget::Impl {
         unprojectedDepth_{Mn::NoCreate},
         depthUnprojectionMesh_{Mn::NoCreate},
         depthUnprojectionFrameBuffer_{Mn::NoCreate},
-        flags_{flags} {
+        flags_{flags},
+        visualSensor_{visualSensor} {
     if (depthShader_) {
       CORRADE_INTERNAL_ASSERT(depthShader_->flags() &
                               DepthShader::Flag::UnprojectExistingDepth);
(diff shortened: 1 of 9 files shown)


Skylion007 (approved):

src/esp/gfx/RenderTarget.cpp
```diff
@@ -145,10 +145,14 @@ struct RenderTarget::Impl {
...
-  void renderEnter() {
+  void renderEnter(const Cr::Containers::Optional<Mn::Color4>& clearColor) {
     framebuffer_.clearDepth(1.0);
     if (flags_ & Flag::RgbaBuffer) {
-      framebuffer_.clearColor(0, Mn::Color4{0, 0, 0, 1});
+      if (clearColor) {  // custom clear color
+        framebuffer_.clearColor(0, *clearColor);
+      } else {  // default clear color
+        framebuffer_.clearColor(0, Mn::Color4{0, 0, 0, 1});
+      }
```
mosra, line 155: What about, instead of the optional and the extra branch, have `Mn::Color4{0, 0, 0, 1}` as a default value for the clear color argument?

src/esp/sensor/CameraSensor.h
```diff
@@ -14,6 +14,7 @@ namespace sensor {
 
 struct CameraSensorSpec : public VisualSensorSpec {
   float orthoScale = 0.1f;
+  Mn::Color4 clearColor = {0, 0, 0, 1};
```
jturner65, line 17:
```suggestion
  Mn::Color4 clearColor{0, 0, 0, 1};
```

mosra: It's more consistent this way, haha :trollface:

src/esp/gfx/RenderTarget.cpp
erikwijmans: One additional option is to pass a reference to the sensor for a given render target.  The render targets are per sensor and this would let it just grab the clear color from the spec.  This would also cleanly resolve the double clear issue pointed out on the python side.

aclegg3:
Thanks for the feedback. New design as @erikwijmans suggested: explicit RenterEnter(CameraSensor&) to consume the configured clear color and avoid the double clear.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -337,8 +340,8 @@ RenderTarget::RenderTarget(const Mn::Vector2i& size,
                                             depthShader,
                                             flags)) {}
 
-void RenderTarget::renderEnter() {
-  pimpl_->renderEnter();
+void RenderTarget::renderEnter(sensor::CameraSensor& cameraSensor) {
```
erikwijmans, line 343: I meant in the constructor itself.  This works but it kinda invalidates the context manager on the python side.  If there is some reason I am missing for why we can't have it be in the constructor, the we should get rid of the context manager in python

aclegg3: 🤔  Looks like there are a number of tests which create RenderTargets for direct consumption with a RenderCamera and have no associated CameraSensor. I guess an optional CameraSensor member could be used, but we'll need a default color anyway in that case.

erikwijmans: Yeah, optional with a default color is fine by me.  Adds an if, but it's a pretty tame one

aclegg3: Using your suggested constructor approach, agreed it's cleaner. Since RenderTarget is constructed in VisualSensor I elevated the clearColor to the VisualSensorSpec.

PR #1154, opened 2021-03-29
[2/3] Voxel Utility Functions

ScottCSteinhauser:
## Motivation and Context
This is part 2 in the Voxel Framework integration project.
I implemented a set of utility functions for generating signed distance fields (Euclidean & Manhattan) and refactored some code to make the framework work with these functions.

## How Has This Been Tested
Going into viewer and pressing 'L' iterates through and displays slices of the Euclidean signed distance field. Additionally, pressing 'G' generates a vector field of the signed distance field distance gradient which is then displayed.

https://user-images.githubusercontent.com/56770849/112901747-c5a99f00-90aa-11eb-816f-1bfdd76491d8.mov

## Types of changes

 - Added VoxelUtils.h and VoxelUtils.cpp in src/esp/geo
 - Refactored some code in VoxelGrid.h and VoxelGrid.cpp to work properly with the new utility functions
 - Added a few ways of interacting with and visualizing the utility functions in Viewer ('G' and 'L')


src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+VoxelGrid::VoxelGrid(const Mn::Vector3& voxelSize,
+                     const Mn::Vector3i& voxelGridDimensions) {
+  m_voxelSize = voxelSize;
+  m_voxelGridDimensions = voxelGridDimensions;
+  m_offset = Mn::Vector3(0.0, 0.0, 0.0);
+  addGrid<bool>("Boundary");
+}
+
+template <>
```
Skylion007, line 81: There has to be a better way to do this casting with template, right?

src/esp/geo/VoxelGrid.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+  // cube indices
+  sz = positions.size();
+  Mn::Vector3 pos6 = mid + orthog1.normalized() * m_voxelSize * 1 / 40;
+  Mn::Vector3 pos7 = mid + orthog2.normalized() * m_voxelSize * 1 / 40;
+  Mn::Vector3 pos8 = mid - orthog1.normalized() * m_voxelSize * 1 / 40;
+  Mn::Vector3 pos9 = mid - orthog2.normalized() * m_voxelSize * 1 / 40;
+
+  Mn::Vector3 pos10 = mid + orthog1.normalized() * m_voxelSize * 1 / 40 -
+                      vec.normalized() * m_voxelSize * 1 / 4;
```
Skylion007, line 245: A lot of wasted cycles on renomralized ortho1, ortho2, and vec. Should just cache their normalized forms.

src/esp/geo/VoxelUtils.cpp
```diff
@@ -0,0 +1,451 @@
...
+  }
+
+  // create new intGrid and copy data from interior/exterior grid
+  v_grid->addGrid<int>(gridName);
+  auto intExtGrid = v_grid->getGrid<int>("InteriorExterior");
+  auto sdfGrid = v_grid->getGrid<int>(gridName);
+
+  Cr::Utility::copy(intExtGrid, sdfGrid);
+
+  // 1st sweep
```
aclegg3, line 146: sweep duplication logic can probably be collapsed

src/esp/geo/VoxelUtils.cpp
```diff
@@ -0,0 +1,451 @@
...
+        if (boolGrid[i][j][k]) {
+          voxelSet.push_back(Mn::Vector3i(i, j, k));
+        }
+      }
+    }
+  }
+  return voxelSet;
+}
+
+std::vector<Mn::Vector3i> getVoxelSetFromIntGrid(
```
aclegg3, line 404: Looks like this could be templated to merge with float variant. 

Additionally a templated comparator could merge all 3. Python bindings could supply explicitly named variants with comparators so you don't need to bind the template or comparator functions.

src/esp/geo/VoxelUtils.h
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * @brief Generates a Vector3 field where each vector of a cell is the
+ * gradient of how the SDF changes.
+ * @param voxelWrapper The voxelization for the gradient field.
+ * @param gridName The name underwhich to register the newly created distance
+ * gradient field.
+ */
+void generateDistanceGradientField(
+    std::shared_ptr<esp::geo::VoxelWrapper>& voxelWrapper,
+    const std::string& gridName = "DistanceGradientField");
```
aclegg3, line 57: Suggestion: maybe a more general gradient field computation would be more useful. Instead of implicitly acting on the SDF field, allow the user to provide a grid name for any float or int field and compute the gradient directly on that.

src/tests/VoxelGridTest.cpp
```diff
@@ -105,6 +108,41 @@ void VoxelGrid::testVoxelGridWithVHACD() {
...
+
+  auto simulator_ = esp::sim::Simulator::create_unique(simConfig);
+
+  // Voxelize the scene with resolution = 1,000,000 and make asserts
+  const int resolution = 1000000;
+  simulator_->createStageVoxelization(resolution);
+  auto voxelization = simulator_->getStageVoxelization();
+
+  // Generate Euclidean SDF for the voxelization
+  esp::geo::generateEuclideanDistanceSDF(voxelization, "ESDF");
```
aclegg3, line 129: I'd like to see some verification of the results here in case the logic changes. Ideally an import/export would be used, but a few golden values would be good enough for now.

aclegg3 (commented):
Mostly ready. A couple of suggestions to reduce duplication and improve generality and testing.

src/esp/geo/VoxelUtils.h
ScottCSteinhauser: generateScalarGradientField(voxelization, inputGrid, outputGrid)

src/esp/geo/VoxelUtils.cpp
```diff
@@ -0,0 +1,451 @@
...
+                      static_cast<std::size_t>(m_voxelGridDimensions[1]),
+                      static_cast<std::size_t>(m_voxelGridDimensions[2])};
+  Corrade::Containers::Array<char> cr_grid{
+      Corrade::Containers::ValueInit,
+      v_grid->gridSize() * sizeof(Mn::Math::BoolVector<6>)};
+  auto shadowGrid_ =
+      Cr::Containers::StridedArrayView<3, Mn::Math::BoolVector<6>>{
+          Cr::Containers::arrayCast<Mn::Math::BoolVector<6>>(cr_grid), dims};
+
+  // fill each grids with ray cast
```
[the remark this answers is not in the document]
aclegg3: Took a pass at this. Not super elegant, but halves the code 🤷 
```
  // fill each grid with ray cast
  bool hit = false;
  int ind = 0;
  int increment = 0;
  Mn::Vector3i indices;
  int shadowGridIndex = 0;
  for (int castAxis = 0; castAxis<3; ++castAxis){
    int a1 = castAxis ? 0 : 1;
    int a2 = castAxis == 2 ? 1 : 2;
    for (int j = 0; j < m_voxelGridDimensions[a1]; j++) {
      for (int k = 0; k < m_voxelGridDimensions[a2]; k++) {
        indices[a1] = j;
        indices[a2] = k;
        //fill from front and back of each 1D slice
        for(int direction=-1; direction<2; direction+=2){ //-1 and 1
          shadowGridIndex = castAxis*2 + (direction < 0 ? 0 : 1);
          hit = false;
          ind = (direction > 0 ? 0 : m_voxelGridDimensions[castAxis]-1);
          while(ind >= 0 && ind < m_voxelGridDimensions[castAxis]){
            indices[castAxis] = ind;
            hit = (hit || boundaryGrid[indices[0]][indices[1]][indices[2]]);
            if (hit)
              shadowGrid_[indices[0]][indices[1]][indices[2]].set(castAxis*2 + (direction < 0 ? 0 : 1), true);
            ind += direction;
          }
        }
      }
    }
  }
```

aclegg3 (approved):
When the tests pass, let's ship this.

PR #1156, opened 2021-04-01
[Bugfix] Set Stage Instance default lighting appropriately.

jturner65:
This small bugfix sets the Scene Instance default lighting to be "DEFAULT_LIGHTING_KEY" instead of "NO_LIGHT_KEY", so that unspecified scenes, particularly those constructed internally to support the direct loading of a stage, will have reasonable light setups.
## How Has This Been Tested
All C++ and python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/attributes/SceneAttributes.cpp
+++ b/src/esp/metadata/attributes/SceneAttributes.cpp
@@ -34,8 +34,8 @@ const std::map<std::string, managers::SceneInstanceTranslationOrigin>
 };
 SceneAttributes::SceneAttributes(const std::string& handle)
     : AbstractAttributes("SceneAttributes", handle) {
-  // defaults to no lights
-  setLightingHandle(NO_LIGHT_KEY);
+  // defaults to default lighting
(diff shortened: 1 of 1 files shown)


Skylion007:
Looks like the WebGL test is failing. We may need to regenerate the snapshot or change the scene config used for that example.

aclegg3 (approved):
👍 

Looks like you'll need to check the js viewer test output image and possibly update it.

eundersander:
It's not clear to me that we want the default to be lit instead of unlit/flat. More discussion here: https://cvmlp.slack.com/archives/G0131KVLBLL/p1617324615060900?thread_ts=1617220202.041900&cid=G0131KVLBLL

jturner65:
> 
> 
> 👍
> 
> Looks like you'll need to check the js viewer test output image and possibly update it.

Yeah, that image was probably captured with no lights.

jturner65:
Upon consideration we have decided that "no lights" is a more appropriate default state (i.e. flat shaded) for stages being loaded without any configuration support.

PR #1159, opened 2021-04-04
[Sensor] upgrade draw_observation in python Sensor class

bigbike:
## Motivation and Context
This diff is trying to address the following issues:
- **Address the code duplication.** The old `draw_observation` in `Sensor` (python) has the same code logic as the `drawObservation()` in the `CameraSensor` class (C++). It is a type of code duplication.
- **Such code logic does not work for the FisheyeSensor introduced recently.** It works ONLY for the "Pinhole" or "Ortho" Camera.
- **The old dependency is wrong.**  Before the actual rendering logic was implemented in the `Renderer` class (C++). And `Sensor` class (both in C++ and python) depended on it. The correct dependency should be: Each **sensor type** implements the actually rendering logic (because different sensors (Pinhole, Fisheye etc.) have different, specific logic), and `Renderer` depends on the sensors. 
- **Upgrade Fisheye sensor and Camera sensor accordingly.**

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -553,45 +553,7 @@ class Sensor:
                  (has it been detached from a scene node?)"
             )
 
-        # get the correct scene graph based on application
-        if self._spec.sensor_type == SensorType.SEMANTIC:
-            if self._sim.semantic_scene is None:
-                raise RuntimeError(
-                    "SemanticSensor observation requested but no SemanticScene is loaded"
-                )
-            scene = self._sim.get_active_semantic_scene_graph()
-        else:  # SensorType is DEPTH or any other type
-            scene = self._sim.get_active_scene_graph()
-
-        # now, connect the agent to the root node of the current scene graph
-
-        # sanity check is not needed on agent:
-        # because if a sensor is attached to a scene graph,
-        # it implies the agent is attached to the same scene graph
-        # (it assumes backend simulator will guarantee it.)
-
-        agent_node = self._agent.scene_node
-        agent_node.parent = scene.get_root_node()
-
-        render_flags = habitat_sim.gfx.Camera.Flags.NONE
-
-        if self._sim.frustum_culling:
-            render_flags |= habitat_sim.gfx.Camera.Flags.FRUSTUM_CULLING
-
-        with self._sensor_object.render_target:
-            self._sim.renderer.draw(self._sensor_object, scene, render_flags)
-
-        # add an OBJECT only 2nd pass on the standard SceneGraph if SEMANTIC sensor with separate semantic SceneGraph
-        if (
-            self._spec.sensor_type == SensorType.SEMANTIC
-            and self._sim.get_active_scene_graph()
-            is not self._sim.get_active_semantic_scene_graph()
-        ):
-            agent_node.parent = self._sim.get_active_scene_graph().get_root_node()
-            render_flags |= habitat_sim.gfx.Camera.Flags.OBJECTS_ONLY
-            self._sim.renderer.draw(
-                self._sensor_object, self._sim.get_active_scene_graph(), render_flags
-            )
+        self._sim.renderer.draw(self._sensor_object, self._sim)
 
     def get_observation(self) -> Union[ndarray, "Tensor"]:
 
--- a/src/esp/bindings/GfxBindings.cpp
+++ b/src/esp/bindings/GfxBindings.cpp
@@ -20,6 +20,7 @@
 #include "esp/gfx/Renderer.h"
 #include "esp/scene/SemanticScene.h"
 #include "esp/sensor/CameraSensor.h"
+#include "esp/sim/Simulator.h"
 
 namespace py = pybind11;
 using py::literals::operator""_a;
@@ -97,6 +98,12 @@ void initGfxBindings(py::module& m) {
           },
           R"(Draw given scene using the camera)", "camera"_a, "scene"_a,
           "flags"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
+      .def(
+          "draw",
+          [](Renderer& self, sensor::VisualSensor& visualSensor,
+             sim::Simulator& sim) { self.draw(visualSensor, sim); },
+          R"(Draw the active scene in current simulator using the visual sensor)",
+          "visualSensor"_a, "sim"_a)
       .def("bind_render_target", &Renderer::bindRenderTarget);
 
   py::class_<RenderTarget>(m, "RenderTarget")
--- a/src/esp/bindings/GfxBindings.cpp
+++ b/src/esp/bindings/GfxBindings.cpp
@@ -59,12 +59,13 @@ void initGfxBindings(py::module& m) {
   render_camera
       .def(py::init_alias<std::reference_wrapper<scene::SceneNode>,
                           const vec3f&, const vec3f&, const vec3f&>())
-      .def("set_projection_matrix",
-           static_cast<RenderCamera& (RenderCamera::*)(int, int, float, float,
-                                                       Mn::Deg)>(
-               &RenderCamera::setProjectionMatrix),
-           R"(Set this `Camera`'s projection matrix.)", "width"_a, "height"_a,
-           "znear"_a, "zfar"_a, "hfov"_a)
+      .def(
+          "set_projection_matrix",
+          [](RenderCamera& self, int w, int h, float n, float f, Mn::Degd fov) {
+            self.setProjectionMatrix(w, h, n, f, Mn::Deg(fov));
+          },
+          R"(Set this `Camera`'s projection matrix.)", "width"_a, "height"_a,
+          "znear"_a, "zfar"_a, "hfov"_a)
       .def("set_orthographic_projection_matrix",
            &RenderCamera::setOrthoProjectionMatrix,
            R"(Set this `Orthographic Camera`'s projection matrix.)", "width"_a,
--- a/src/esp/bindings/GfxBindings.cpp
+++ b/src/esp/bindings/GfxBindings.cpp
@@ -84,17 +84,17 @@ void initGfxBindings(py::module& m) {
       .def(
           "draw",
           [](Renderer& self, sensor::VisualSensor& visualSensor,
-             scene::SceneGraph& sceneGraph, RenderCamera::Flags flags) {
-            self.draw(visualSensor, sceneGraph, flags);
+             scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
+            self.draw(visualSensor, sceneGraph, RenderCamera::Flags{flags});
           },
           R"(Draw given scene using the visual sensor)", "visualSensor"_a,
           "scene"_a,
-          "flags"_a = RenderCamera::Flags{RenderCamera::Flag::FrustumCulling})
+          "flags"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
       .def(
           "draw",
           [](Renderer& self, RenderCamera& camera,
-             scene::SceneGraph& sceneGraph, RenderCamera::Flags flags) {
-            self.draw(camera, sceneGraph, flags);
+             scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
+            self.draw(camera, sceneGraph, RenderCamera::Flags{flags});
           },
           R"(Draw given scene using the camera)", "camera"_a, "scene"_a,
           "flag"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
--- a/src/esp/bindings/GfxBindings.cpp
+++ b/src/esp/bindings/GfxBindings.cpp
@@ -83,15 +83,6 @@ void initGfxBindings(py::module& m) {
       .def(py::init(&Renderer::create<>))
       .def(
           "draw",
-          [](Renderer& self, sensor::VisualSensor& visualSensor,
-             scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
-            self.draw(visualSensor, sceneGraph, RenderCamera::Flags{flags});
-          },
-          R"(Draw given scene using the visual sensor)", "visualSensor"_a,
-          "scene"_a,
-          "flags"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
-      .def(
-          "draw",
           [](Renderer& self, RenderCamera& camera,
              scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
             self.draw(camera, sceneGraph, RenderCamera::Flags{flags});
--- a/src/esp/bindings/GfxBindings.cpp
+++ b/src/esp/bindings/GfxBindings.cpp
@@ -84,17 +84,17 @@ void initGfxBindings(py::module& m) {
       .def(
           "draw",
           [](Renderer& self, sensor::VisualSensor& visualSensor,
-             scene::SceneGraph& sceneGraph, RenderCamera::Flag flag) {
-            self.draw(visualSensor, sceneGraph, RenderCamera::Flags{flag});
+             scene::SceneGraph& sceneGraph, RenderCamera::Flags flags) {
+            self.draw(visualSensor, sceneGraph, flags);
           },
           R"(Draw given scene using the visual sensor)", "visualSensor"_a,
           "scene"_a,
-          "flag"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
+          "flags"_a = RenderCamera::Flags{RenderCamera::Flag::FrustumCulling})
       .def(
           "draw",
           [](Renderer& self, RenderCamera& camera,
-             scene::SceneGraph& sceneGraph, RenderCamera::Flag flag) {
-            self.draw(camera, sceneGraph, RenderCamera::Flags{flag});
+             scene::SceneGraph& sceneGraph, RenderCamera::Flags flags) {
+            self.draw(camera, sceneGraph, flags);
           },
           R"(Draw given scene using the camera)", "camera"_a, "scene"_a,
           "flag"_a = RenderCamera::Flag{RenderCamera::Flag::FrustumCulling})
--- a/src/esp/bindings/GfxBindings.cpp
+++ b/src/esp/bindings/GfxBindings.cpp
@@ -84,8 +84,8 @@ void initGfxBindings(py::module& m) {
       .def(
           "draw",
           [](Renderer& self, sensor::VisualSensor& visualSensor,
-             scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
-            self.draw(visualSensor, sceneGraph, RenderCamera::Flags{flags});
+             scene::SceneGraph& sceneGraph, RenderCamera::Flag flag) {
+            self.draw(visualSensor, sceneGraph, RenderCamera::Flags{flag});
           },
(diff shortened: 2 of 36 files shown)


src/esp/gfx/Renderer.h
```diff
@@ -38,11 +40,22 @@ class Renderer {
             scene::SceneGraph& sceneGraph,
             RenderCamera::Flags flags = {RenderCamera::Flag::FrustumCulling});
 
+  // TODO: deprecate the following function: draw(visualSensor, sceneGraph,
+  // flags);
   // draw the scene graph with the visual sensor provided by user
   void draw(sensor::VisualSensor& visualSensor,
```
bigbike, line 46: My opinion: This function is really obsolete and thus should be removed.

bigbike:
Tested it with the `example.py` and it gave the correct result. See the attached picture:
test rgba 00000

habitat_sim/simulator.py
```diff
@@ -553,45 +553,7 @@ def draw_observation(self) -> None:
...
-        with self._sensor_object.render_target:
-            self._sim.renderer.draw(self._sensor_object, scene, render_flags)
-
-        # add an OBJECT only 2nd pass on the standard SceneGraph if SEMANTIC sensor with separate semantic SceneGraph
-        if (
-            self._spec.sensor_type == SensorType.SEMANTIC
-            and self._sim.get_active_scene_graph()
-            is not self._sim.get_active_semantic_scene_graph()
-        ):
-            agent_node.parent = self._sim.get_active_scene_graph().get_root_node()
```
Skylion007, line 590: Doesn't this need to be replicated somewhere in the C++? We have a couple of the tests are segfault with the new C++ function.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -115,20 +124,42 @@ bool CameraSensor::drawObservation(sim::Simulator& sim) {
...
+      node().setTransformation(absTransform);
+      draw(sim.getActiveSemanticSceneGraph(), flags);
+      // after drawing, put the node back to the regular scene graph.
+      // MUST use the setParent function! Otherwise the sensorSuite in the node
+      // will be wrong!!!
+      node().setParent(p);
+      node().setTransformation(relativeTransform);
+    } else {
+      draw(sim.getActiveSemanticSceneGraph(), flags);
+    }
```
bigbike, line 154: @Skylion007 : I just updated the code as I replicated the logic here, and did it in the correct way with `setParent()`.

habitat_sim/simulator.py
bigbike: @Skylion007 : Yeah, this is the root cause I found last night.  See my previous comments.
Now I believe it is resolved in the `CameraSensor` class. Let us see what the CI would say.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -115,20 +124,42 @@ bool CameraSensor::drawObservation(sim::Simulator& sim) {
...
+
+    if (twoSceneGraphs) {
       flags |= gfx::RenderCamera::Flag::ObjectsOnly;
```
Skylion007, line 157: BTW, this cludge we are doing is going to hurt our FPS a lot in scenes without the semantic mesh, but only added objects (such as synthetic datasets.) One thing on the agenda is to find a better way to resolve this long term so we don't have to do this double pass all the time. @aclegg3 I think you mentioned something about this during our last meeting?

Skylion007: This can of course be saved for a different PR, just wanted to bring it up.

Skylion007 (approved):

habitat_sim/simulator.py
```diff
@@ -553,45 +553,7 @@ def draw_observation(self) -> None:
                  (has it been detached from a scene node?)"
             )
 
-        # get the correct scene graph based on application
-        if self._spec.sensor_type == SensorType.SEMANTIC:
-            if self._sim.semantic_scene is None:
```
Skylion007, line 558: Do we need to add an ESPCHECK for this in the C++ or are we good?

src/esp/bindings/GfxBindings.cpp
```diff
@@ -83,20 +84,26 @@ void initGfxBindings(py::module& m) {
       .def(
           "draw",
           [](Renderer& self, sensor::VisualSensor& visualSensor,
-             scene::SceneGraph& sceneGraph, RenderCamera::Flag flags) {
-            self.draw(visualSensor, sceneGraph, RenderCamera::Flags{flags});
+             scene::SceneGraph& sceneGraph, RenderCamera::Flags flags) {
```
bigbike, line 87: @aclegg3 : 
Would you mind taking a close look at it? 
The original code was introduced in #679 . I believe it is a bug. Can you help me to double verify it?
Thanks.

src/esp/core/ManagedContainer.h
```diff
@@ -703,7 +703,7 @@ class ManagedContainer : public ManagedContainerBase {
   ManagedPtr defaultObj_ = nullptr;
 
  public:
-  ESP_SMART_POINTERS(ManagedContainer<T, Access>);
+  ESP_SMART_POINTERS(ManagedContainer<T, Access>)
```
bigbike, line 706: this ";" gives us a lot of compiling warnings. Get rid of it in this PR as well.

habitat_sim/simulator.py
bigbike: @Skylion007 : ESP_CHECK is only used in the C++, NO?

Skylion007: Yeah, but this check was removed. I just want to make sure we throw the error somewhere.

bigbike: OK, good catch. I will see if I can do something in the `getActiveSemanticSceneGraph`.

src/esp/gfx/Renderer.cpp
```diff
@@ -51,6 +53,16 @@ struct Renderer::Impl {
     draw(*visualSensor.getRenderCamera(), sceneGraph, flags);
   }
 
+  void draw(sensor::VisualSensor& visualSensor, sim::Simulator& sim) {
+    if (visualSensor.specification()->sensorType ==
+        sensor::SensorType::Semantic) {
+      ESP_CHECK(sim.semanticSceneExists(),
```
bigbike, line 59: @Skylion007 : I added a sanity check here to address your concern.

src/esp/sensor/FisheyeSensor.cpp
```diff
@@ -155,12 +155,36 @@ bool FisheyeSensor::drawObservation(sim::Simulator& sim) {
...
-    cubeMap_->renderToTexture(*cubeMapCamera_,
-                              sim.getActiveSemanticSceneGraph(), flags);
-    if (&sim.getActiveSemanticSceneGraph() != &sim.getActiveSceneGraph()) {
+    bool twoSceneGraphs =
+        (&sim.getActiveSemanticSceneGraph() != &sim.getActiveSceneGraph());
+
+    if (twoSceneGraphs) {
+      // *assume* the sensor is always in the regular scene graph.
+      // backup first
+      Mn::Matrix4 relativeTransform = node().transformation();
```
Skylion007, line 164: Perhaps this should be abstracted out if possible. Disappointing that this hack is necessary, but it is required for every VisualSensor, right?

Skylion007: That can also be saved for a later PR though.

bigbike: Yes, a bit code duplication here. I did not figure out a good way to extract it so far. Like I said in the description, different visual sensors has it is own rendering logic. Read this two pieces of code. They are a bit different.

bigbike: Allow me to think about it.
This is problem that we need to solve.

Otherwise it will become a big problem as more sensors are coming in the future.

@mosra @aclegg3 @erikwijmans any ideas here?

bigbike: I am thinking the ways to address this. It must be solved before committing this diff in.

src/esp/sensor/FisheyeSensor.cpp
```diff
@@ -147,12 +147,41 @@ bool FisheyeSensor::drawObservation(sim::Simulator& sim) {
...
-  cubeMap_->renderToTexture(*cubeMapCamera_, sim.getActiveSceneGraph(), flags);
+  if (fisheyeSensorSpec_->sensorType == SensorType::Semantic) {
+    bool twoSceneGraphs =
+        (&sim.getActiveSemanticSceneGraph() != &sim.getActiveSceneGraph());
+
+    if (twoSceneGraphs) {
+      moveSemanticSensorToSemanticSceneGraph(sim);
+      cubeMap_->renderToTexture(*cubeMapCamera_,
+                                sim.getActiveSemanticSceneGraph(), flags);
+      moveSemanticSensorBackToRegularSceneGraph(sim);
```
bigbike, line 165: @Skylion007 : I reduced the code duplication by extracting 2 helper functions applied here.

The challenge is that the code looks like a "sandwich". The "bread" is the same but the "food" in between varies.

Skylion007:
Similar to render_enter / exit semantics  though, right?  It would need to
be split between two functions. Also could be done with a functional
programming lambda interface, although not as cleanly.

src/esp/sensor/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ add_library(
 target_link_libraries(
   sensor
   PUBLIC core gfx scene
+  PRIVATE sim
```
erikwijmans, line 27: Why private?

src/esp/sensor/CameraSensor.cpp
```diff
@@ -107,6 +106,16 @@ gfx::RenderCamera* CameraSensor::getRenderCamera() const {
   return renderCamera_;
 }
 
+void CameraSensor::draw(scene::SceneGraph& sceneGraph,
+                        gfx::RenderCamera::Flags flags) {
+  for (auto& it : sceneGraph.getDrawableGroups()) {
+    // TODO: remove || true
```
erikwijmans, line 112: This TODO is getting carried around more and more places, are we ever going to actually do it?

src/esp/sim/Simulator.h
```diff
@@ -73,6 +73,8 @@ class Simulator {
     return semanticScene_;
   }
 
+  bool semanticSceneExists() { return (semanticScene_ != nullptr); }
```
erikwijmans, line 76:
```suggestion
  bool semanticSceneExists() const { return (semanticScene_ != nullptr); }
```

src/esp/sensor/VisualSensor.cpp
```diff
@@ -136,5 +137,39 @@ bool VisualSensor::getObservation(sim::Simulator& sim, Observation& obs) {
...
+    sim::Simulator& sim) {
+  CORRADE_INTERNAL_ASSERT(visualSensorSpec_->sensorType ==
+                          SensorType::Semantic);
+
+  // check if the sensor is already in this semantic scene graph
+  if (node().scene() == sim.getActiveSceneGraph().getRootNode().scene()) {
+    return;
+  }
+
+  node().setParent(semanticSensorParentNodeBackup_);
```
erikwijmans, line 170: Can you add an assert that the pointer is not null and the optional is not empty and then null/empty them after?

src/esp/sensor/CameraSensor.cpp
```diff
@@ -115,20 +124,30 @@ bool CameraSensor::drawObservation(sim::Simulator& sim) {
...
-  gfx::Renderer::ptr renderer = sim.getRenderer();
   if (cameraSensorSpec_->sensorType == SensorType::Semantic) {
     // TODO: check sim has semantic scene graph
-    renderer->draw(*this, sim.getActiveSemanticSceneGraph(), flags);
-    if (&sim.getActiveSemanticSceneGraph() != &sim.getActiveSceneGraph()) {
+    bool twoSceneGraphs =
+        (&sim.getActiveSemanticSceneGraph() != &sim.getActiveSceneGraph());
+
+    if (twoSceneGraphs) {
+      moveSemanticSensorToSemanticSceneGraph(sim);
```
erikwijmans, line 137: This seems a bit risky as it's easy to forget to call the `moveSemanticSensorBackToRegularSceneGraph` function.  Worth considering RAII here.

Skylion007:
@bigbike Merging from master should solve the CI issues caused by doc building

src/esp/sensor/CameraSensor.cpp
bigbike: I believe the purpose is to build a filter function on drawables, which can be useful in the material editing.

src/esp/sensor/CameraSensor.cpp
bigbike: Can you elaborate a bit how you are going to do RAII here?

if people forgets to call this function `moveSemanticSensorBackToRegularSceneGraph`, she will know immediately as in the following (L145), she needs it to draw the "objects". If the above function is not called, the program will hit an assertion in Magnum (camera is not in the same SG) and quit.

src/esp/sensor/VisualSensor.cpp
bigbike: OK. Will do.

src/esp/sensor/CMakeLists.txt
bigbike: It was set (in different ways) to help me better understand why the CI fails.
It is a sad story. I was taking PTO on Mon. but I spent whole day trying to solve the CI failure, which is still there.

I found the `ResourceMangerTest` cannot be compiled. It complains `libsensor.a` is missing some definitions from here and there, even though it does not need anything from `sensor`.

src/esp/sensor/CameraSensor.cpp
erikwijmans: Yeah, the `it.second.prepareForDraw(*renderCamera_)` part of the if statement seems fine and I can see a use for that returning false and something not getting drawn, my concern is that we are simply ignoring the return value of that function via the `|| true`.  If we ever do want to do something with that function, we have a bunch of ` || true`'s that will make that harder.  

I don't know the history of the `|| true` (I think it's been there forever and thus likely a debugging thing that never got removed), but it currently does nothing since the only implementation for `prepareForDraw` just returns true.

src/esp/sensor/CameraSensor.cpp
erikwijmans: Just a helper class that calls `moveSemanticSensorToSemanticSceneGraph` in its constructor and `moveSemanticSensorBackToRegularSceneGraph` in its destructor.

Skylion007:
Okay weird, only 1 test is running.

src/esp/sensor/VisualSensor.cpp
```diff
@@ -136,5 +137,58 @@ bool VisualSensor::getObservation(sim::Simulator& sim, Observation& obs) {
...
+  CORRADE_INTERNAL_ASSERT(visualSensor_.specification()->sensorType ==
+                          SensorType::Semantic);
+  scene::SceneNode& node = visualSensor_.node();
+  CORRADE_ASSERT(
+      !scene::SceneGraph::isRootNode(node),
+      "VisualSensor::moveSemanticSensorToSemanticSceneGraph(): the semantic "
+      "sensor is attached to the root node, and thus cannot be moved.", );
+
+  // check if the sensor is already in this semantic scene graph
+  if (node.scene() == sim.getActiveSemanticSceneGraph().getRootNode().scene()) {
```
erikwijmans, line 153: I think this case will always cause a crash?  Since the destructor will always hit `CORRADE_INTERNAL_ASSERT(semanticSensorParentNodeBackup_);`

src/esp/sensor/VisualSensor.h
```diff
@@ -195,6 +199,34 @@ class VisualSensor : public Sensor {
   std::unique_ptr<gfx::RenderTarget> tgt_;
   VisualSensorSpec::ptr visualSensorSpec_ =
       std::dynamic_pointer_cast<VisualSensorSpec>(spec_);
+
+  class MoveSemanticSensorNodeHelper {
```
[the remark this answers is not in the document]
erikwijmans: Yeah, this looks nice.  Left one question about implementation

src/esp/sensor/VisualSensor.cpp
bigbike: Well, the basic assumption is that at the very beginning, the semantic sensor is attached to the regular scene graph, not the semantic scene graph. Your comment makes me think if it is always true.

If this is true, then you do not need to worry about it here.

bigbike: If there are NO two scene graphs, e.g., only the semantic scene graph, then user should never ever use `MoveSemanticSensorNodeHelper` at all.

src/esp/sensor/CameraSensor.cpp
```diff
@@ -115,20 +123,29 @@ bool CameraSensor::drawObservation(sim::Simulator& sim) {
...
+    if (twoSceneGraphs) {
       flags |= gfx::RenderCamera::Flag::ObjectsOnly;
-      renderer->draw(*this, sim.getActiveSceneGraph(), flags);
+      draw(sim.getActiveSceneGraph(), flags);
     }
   } else {
-    // SensorType is Depth or any other type
-    renderer->draw(*this, sim.getActiveSceneGraph(), flags);
+    // SensorType is Color, Depth or any other type
+    draw(sim.getActiveSceneGraph(), flags);
```
bigbike, line 148: @erikwijmans :
If there is only one scene graph, we assume it is the regular scene graph.
Is there a chance that we will only have semantic scene graph?
also cc: @aclegg3

src/esp/sensor/CameraSensor.cpp
aclegg3: 🤔  I'm not sure that this is the case. Here if we only have a single scene graph we'll just render it with semantic ids, if we have 2 scene-graphs (e.g. a specific separate semantic asset) then we render that asset and then a pass over dynamic objects. I don't see the overhead you are worried about. Am I missing something?

src/esp/sensor/CameraSensor.cpp
aclegg3: If we only have a one scene graph we'll use it for both RGB and SEMANTIC and all object will be on the correct scene graph for both renders.

src/esp/sensor/VisualSensor.cpp
erikwijmans: Should this be an assert then?  Right now it looks like a feature where this won't move the sensor if it doesn't need to.  However, the destructor assumes that if the sensor isn't part of the active scene graph, then the constructor moved it to the semantic scene graph and will try to move it back.  This will cause a crash however.

src/esp/sensor/CameraSensor.cpp
bigbike: @aclegg3 : It means the absolute transformation for all nodes will be computed twice (one per each SG).

bigbike:
Dear reviewers,

If you do not have more concerns, I am going to merge it in this afternoon. Thanks.
@erikwijmans : hope the latest version addressed all your concerns.

mosra (approved):
Just a random comment that doesn't affect the approval status.

erikwijmans (approved):
🚢

Skylion007:
Looks good. Let's merge this in.

PR #1168, opened 2021-04-12
[Sensor][PYTHON][GFX] FishEyeSensor Bindings

Skylion007:
## Motivation and Context

* Adds Python bindings for FishEye Sensor

## Types of changes

## Checklist

Need to lock in generation of ground truth.


src/esp/bindings/SensorBindings.cpp
```diff
@@ -54,7 +55,11 @@ void initSensorBindings(py::module& m) {
   py::enum_<SensorSubType>(m, "SensorSubType")
       .value("NONE", SensorSubType::None)
       .value("PINHOLE", SensorSubType::Pinhole)
-      .value("ORTHOGRAPHIC", SensorSubType::Orthographic);
+      .value("ORTHOGRAPHIC", SensorSubType::Orthographic)
+      .value("FISHEYE", SensorSubType::Fisheye);
```
Skylion007, line 59: @bigbike Not sure this sensorsubtype is necessary right? I just overloaded the sensor_subtype enum on FishEyeSensorSpecs with the FisheyeSensorModelType.

examples/settings.py
```diff
@@ -95,6 +98,23 @@ def make_cfg(settings):
...
+        fisheye_sensor_spec.sensor_type = habitat_sim.SensorType.COLOR
+        fisheye_sensor_spec.sensor_subtype = (
+            habitat_sim.FisheyeSensorModelType.DOUBLESPHERE
+        )
+        fisheye_sensor_spec.resolution = [settings["height"], settings["width"]]
+        fisheye_sensor_spec.focal_length = (
+            np.ones(2) * min(settings["height"], settings["width"]) * 0.5
+        )
+        fisheye_sensor_spec.principle_point_offset = np.ones(2, dtype=np.int32) * (
+            min(settings["height"], settings["width"]) // 2
```
bigbike, line 113: I think the principle offset is not correct. 
(For Principal Point Offset, see here: http://ksimek.github.io/2013/08/13/intrinsic/)
This is what I did in the viewer:
https://github.com/facebookresearch/habitat-sim/blob/master/src/utils/viewer/viewer.cpp#L455-L457

bigbike: Also you need to set the 
    spec->xi = -0.18;
    spec->alpha = 0.59;
   somewhere.

Skylion007: These are already the defaults in the sensorspec so no need to adjust the values though?

Skylion007: https://github.com/facebookresearch/habitat-sim/blob/fc7baba4f208a7589a7af752ab686e39cccf28c4/src/esp/sensor/FisheyeSensor.h#L77

bigbike: Yeah, I would still encourage you to explicitly set them here. We are writing examples for the external users. So people know these are the parameters that they can tune and reset.

src/esp/bindings/SensorBindings.cpp
bigbike: Yes, I believe we need the sub-type of the sensor. You are doing the correct thing here.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -54,7 +55,11 @@ void initSensorBindings(py::module& m) {
...
+
+  py::enum_<FisheyeSensorModelType>(m, "FisheyeSensorModelType")
+      .value("DOUBLESPHERE", FisheyeSensorModelType::DoubleSphere);
```
bigbike, line 62: Maybe DOUBLE_SPHERE?

examples/settings.py
```diff
@@ -95,6 +98,23 @@ def make_cfg(settings):
         ortho_sensor_spec.sensor_subtype = habitat_sim.SensorSubType.ORTHOGRAPHIC
         sensor_specs.append(ortho_sensor_spec)
 
+    if settings["fisheye_sensor"]:
```
bigbike, line 101: `fisheye_color_sensor`? or `fisheye_rgb_sensor`? 
Later we can have `fisheye_depth_sensor` `fisheye_semantic_sensor`.

examples/settings.py
```diff
@@ -95,6 +98,23 @@ def make_cfg(settings):
...
+        fisheye_sensor_spec = habitat_sim.FisheyeSensorDoubleSphereSpec()
+        fisheye_sensor_spec.uuid = "fisheye_sensor"
```
bigbike, line 103: I would change the uuid as well.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -93,11 +98,29 @@ void initSensorBindings(py::module& m) {
...
       .def_readwrite("ortho_scale", &CameraSensorSpec::orthoScale);
 
+  // ====FisheyeSensorSpec ====
+  py::class_<FisheyeSensorSpec, FisheyeSensorSpec::ptr, VisualSensorSpec>(
+      m, "FisheyeSensorSpec", py::dynamic_attr())
+      .def(py::init(&FisheyeSensorSpec::create<>))
+      .def_readwrite("focal_length", &FisheyeSensorSpec::focalLength)
+      .def_readwrite("principal_point_offset",
+                     &FisheyeSensorSpec::principalPointOffset)
+      .def_readwrite("cubemap_size", &FisheyeSensorSpec::cubemapSize)
```
bigbike, line 113: We need to leave a comment here for the cubemapSize. By default, if the user does not set it, the system will adopt `min(resolution[0], resolution[1])`. If this value is set by the user, it will override the default value.

src/esp/bindings/SensorBindings.cpp
```diff
@@ -93,11 +98,29 @@ void initSensorBindings(py::module& m) {
...
+      .def_readwrite("sensor_subtype", &FisheyeSensorSpec::fisheyeModelType);
+
+  // ====FisheyeSensorDoubleSphereSpec ====
```
bigbike, line 116: You can consider leaving a comment on the parameters and pointing to the paper:
  /**
   * alpha and xi are specific to "double sphere" camera model.
   * see details (value ranges) in:
   * Vladyslav Usenko, Nikolaus Demmel and Daniel Cremers: The Double Sphere
   * Camera Model, The International Conference on 3D Vision (3DV), 2018
   */

examples/settings.py
Skylion007: Yeah, but I kinda want to redo this entire system as it is right now because we also should do the same with the ortho view.

src/esp/bindings/SensorBindings.cpp
Skylion007: Yeah, but we don't end up using it because I am overriding SensorType to be the modelType for the FishEye Sensor, should I avoid doing that?

tests/test_sensors.py
```diff
@@ -54,6 +55,8 @@ def _render_and_load_gt(sim, scene, sensor_type, gpu2gpu):
             "{}-{}.npy".format(osp.basename(osp.splitext(scene)[0]), sensor_type),
         )
     )
+    if not osp.exists(gt_obs_file):
```
erikwijmans, line 58: Please add ground truth or make a separate test.  Currently this test is a half-decent integration test for the full system and I don't like making part of it just a smoke test.

examples/settings.py
Skylion007: Fixed it. It was a typo that wasn't caught due to the fact that we allow dynamic attrs. I thought for sure it would crash if that value wasn't set since it wasn't initialized.

@bigbike Should we make principal point set automatically from resolution like CubeMap size? I don't see many cases where you would want the principal point of center, right?

examples/settings.py
bigbike: Yeah, I would really appreciate it if you can refactor the code here. You can still make this one right in this PR though.
That means when you refactor, you would have an example to follow, and one less place to fix.

Skylion007: Came up with a nice clean way of doing it.

src/esp/bindings/SensorBindings.cpp
bigbike: Correct, if I were you, I would avoid doing that. This causes inconsistency between the sim engine and the py interface.
Also in terms of design, I think it is enough at this level that user understands it is a fisheye sensor (without knowing the actual camera model). 
And if the user does not set the actual camera type, by default, the simulator will adopt "doubleSphere" model (assuming we have a couple of options in the future).

examples/settings.py
bigbike: Seriously, you can change it to `fisheye_color_sensor`. You do not have to wait until you refactor it in the future. It must be corrected anyway, sooner or later.

bigbike: Oh, I see. you are passing the parameters in.

examples/settings.py
```diff
@@ -95,6 +99,38 @@ def make_cfg(settings):
...
 
+    # TODO Figure out how to implement copying of specs
+    def create_fisheye_spec(**kw_args):
+        fisheye_sensor_spec = habitat_sim.FisheyeSensorDoubleSphereSpec()
+        fisheye_sensor_spec.uuid = "fisheye_sensor"
+        fisheye_sensor_spec.sensor_type = habitat_sim.SensorType.COLOR
+        fisheye_sensor_spec.sensor_model_type = (
+            habitat_sim.FisheyeSensorModelType.DOUBLE_SPHERE
+        )
+        fisheye_sensor_spec.alpha = 0.59  # default
```
bigbike, line 110: I would add a comment here:
/*
The default value (alpha, xi) is set to match the lens "BF2M2020S23" found in Table 3 of this paper:
    Vladyslav Usenko, Nikolaus Demmel and Daniel Cremers: The Double Sphere
    Camera Model, The International Conference on 3D Vision (3DV), 2018

 You can find the intrinsic parameters for the other lenses in the same table as well.
*/

bigbike (approved):

bigbike:
Also, it would be awesome if we can add this new fisheye sensor to the colab tutorial, e.g., in the navigation section.

Skylion007:
BTW, wanted to give a heads up to @matsuren about this if you want to help implement any other sensors/shaders. This looks like it will be at least 3X as fast the ObservationTransformer hack. :)

matsuren:
Hi @Skylion007, thank you for the information! It's awesome that it's written in C++ with shaders!

Skylion007:
@matsuren If there are any other shaders, you would be interested in contributing like (Omnidirectional Stereo or such), let us know. Would be happy to get it merged.

matsuren:
@Skylion007 Thank you! I will let you know when I have something in my mind. (recently, I'm swamped with other projects, so I hope I can contribute in the future;)

PR #1172, opened 2021-04-15
[BUGFIX] Fix Voxel grid test failing when --vhacd not enabled.

jturner65:
## Motivation and Context
VoxelGridTest was failing if --vhacd was not enabled (with no actual tests to run, the test suite considers this a failure), so this PR does not compile the test unless it has --vhacd is enabled.  All compiler directives in the test file were removed as well, since they were not necessary.

## How Has This Been Tested
C++ and python tests with and without flag passed.

## Types of changes

## Checklist

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -57,18 +57,20 @@ target_include_directories(GfxReplayTest PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 test(ResourceManagerTest assets)
 target_include_directories(ResourceManagerTest PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
-corrade_add_test(
-  VoxelGridTest
-  VoxelGridTest.cpp
-  LIBRARIES
-  sim
-  assets
-  geo
-  Magnum::DebugTools
-  Magnum::AnyImageConverter
-  MagnumPlugins::StbImageImporter
-)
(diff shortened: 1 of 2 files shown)


aclegg3:
I didn't realize this was failing without `--vhacd`, thanks for catching this. However, I think there is still value to testing non-vhacd dependent functionality. Users can still create and access voxel grids, they just can't auto-compute a boundary grid.

jturner65:
> I didn't realize this was failing without `--vhacd`, thanks for catching this. However, I think there is still value to testing non-vhacd dependent functionality. Users can still create and access voxel grids, they just can't auto-compute a boundary grid.

Unfortunately none of the authored tests can run without VHACD, so without a valid test the test suite will always fail.

aclegg3:
> > I didn't realize this was failing without `--vhacd`, thanks for catching this. However, I think there is still value to testing non-vhacd dependent functionality. Users can still create and access voxel grids, they just can't auto-compute a boundary grid.
> 
> Unfortunately none of the authored tests can run without VHACD, so without a valid test the test suite will always fail.

Yeah, I think some of that non-vhacd functionality could be split out into a separate test but for now this fix seems fine. The CI will test with vhacd anyway, so we won't miss issues there.

aclegg3 (approved):

Skylion007 (approved):

PR #1173, opened 2021-04-15
Material ID refactor

aclegg3:
## Motivation and Context

`ResourceManager::shaderManager_` holds materials keyed by `Magnum::ResourceKey`. The current data import pipeline assumes these materials are keyed by int and that all materials for an asset are keyed contiguously. This is a limiting paradigm as it does not allow us to use arbitrary string keying for programmatic or user-defined materials.

Concretely an example of this case is material overriding in URDF requiring on-the-fly material registration and editing existing MeshMetaData.

This change refactors away from int ids for material keys and the notion of local material ids in `MeshMetaData` to instead directly store global string keys. The previous method of creating new unique ids during import by incrementing a counter remains.

## How Has This Been Tested

All current tests pass after the refactor.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/MeshMetaData.h
+++ b/src/esp/assets/MeshMetaData.h
@@ -28,8 +28,8 @@ struct MeshTransformNode {
   /** @brief Local mesh index within @ref MeshMetaData::meshIndex. */
   int meshIDLocal;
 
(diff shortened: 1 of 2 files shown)


eundersander (approved):
LGTM

Skylion007 (approved):

PR #1175, opened 2021-04-16
[BUGFIX] Add hfov to CameraSensor config settings

Skylion007:
## Motivation and Context

During a refactor, we lost the ability to set the initial hfov via a sensor config. This restores the ability to do that via the CameraSensorSpec. See the corresponding Habitat-Lab bug: https://github.com/facebookresearch/habitat-lab/issues/627. This should add the necessary bindings to set the initial hfov of a CameraSensorSpec.

## Types of changes

## Checklist

### DIFF
--- a/examples/settings.py
+++ b/examples/settings.py
@@ -13,6 +13,7 @@ default_sim_settings = {
     "height": 480,
     "default_agent": 0,
     "sensor_height": 1.5,
+    "hfov": 90,
     "color_sensor": True,  # RGB sensor (default: ON)
     "semantic_sensor": False,  # semantic sensor (default: OFF)
     "depth_sensor": False,  # depth sensor (default: OFF)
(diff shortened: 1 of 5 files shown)


aclegg3 (approved):
Looks good.

mosra:
I wasn't able to review this sooner, but with the "Allow FOV to take in floats and ints" commit the setter is no longer unit-safe. The main intention of the Deg/Rad type was to ensure the people don't accidentally put degrees where radians are expected and vice versa.

Or was it because it was not possible to construct `mn.Deg` from an integer? If that's the case, I can fix that in the bindings.

Skylion007:
It was mostly to simplify config construction for the Sensors via Habitat-Lab. Converting from a float to a mn.Deg would require explicitly setting the type and overwriding since YAML can't deduce basic types. 

Mainly to help with this PR so it can be constructed automatically: https://github.com/facebookresearch/habitat-lab/pull/628 . It would still be unit safe if we only allow mn.Deg to be constructed this way right? A float would also e interpreted as a degree unless explicitly specified as a Radian? We seem to use degrees a lot in Habitat-Lab so that might be a safe assumption to make for Lab side of the PR.

mosra:
> It would still be unit safe if we only allow mn.Deg to be constructed this way right? A float would also e interpreted as a degree unless explicitly specified as a Radian?

It wouldn't, because it would allow you to do `fov = math.pi` and then would happily interpret that as 3.1°. And I'm not sure if degrees can be safely assumed, you wouldn't usually pass degrees to `math.sin()` for example (but of course [`mn.math.sin(mn.Deg(35))`](https://doc.magnum.graphics/python/magnum/math/#sin) works and does the right thing, because of automatic type-aware conversions).

- - - 

But I see your point, Hab-Lab doesn't depend on / expose Magnum for various reasons and so exposing/requiring Magnum-specific types there probably doesn't make sense. Just something to be aware of.

PR #1179, opened 2021-04-19
[Sensor] Improved Equirectangular sensor

bigbike:
## Motivation and Context
- Introduce equirectangular sensor. Unlike #1141, which is a prototype and contains a large amount of code duplication, this one refactors the existing class hierarchy to accommodate this new sensor.
- Introduces an intermediate class called `CubeMapSensorBase` as the immediate parent class for `FisheyeSensor` and `Equirectangular` sensor.
- Introduces another intermediate class named `CubeMapShaderBase` to be the parent class for specific sensor shaders.
- Small fixes in both the C++ code and shader code.
- Shaders has been verified using:
```
magnum-shaderconverter --validate equirectangular.frag --input-version "300 es" --output-version opengl

```

Results:
equirect-1
equirect-depth

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -391,7 +391,7 @@ Key Commands:
     Equirectangular,
     VisualSensorModeCount,
   };
-  uint8_t sensorMode_ = 0;
+  VisualSensorMode sensorMode_ = VisualSensorMode::Camera;
 
   void bindRenderTarget();
 };
@@ -1124,9 +1124,9 @@ void Viewer::drawEvent() {
   if (depthMode_) {
     // ================ Depth Visualization ==================================
     std::string sensorId = "depth";
-    if (sensorMode_ == (uint8_t)VisualSensorMode::Fisheye) {
+    if (sensorMode_ == VisualSensorMode::Fisheye) {
       sensorId = "depth_fisheye";
-    } else if (sensorMode_ == (uint8_t)VisualSensorMode::Equirectangular) {
+    } else if (sensorMode_ == VisualSensorMode::Equirectangular) {
       sensorId = "depth_equirectangular";
     }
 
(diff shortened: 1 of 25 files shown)


bigbike:
Just updated the visual results in the description.

src/utils/viewer/viewer.cpp
```diff
@@ -1088,27 +1124,19 @@ void Viewer::drawEvent() {
   if (depthMode_) {
     // ================ Depth Visualization ==================================
     std::string sensorId = "depth";
-    if (fisheyeMode_) {
+    if (sensorMode_ == (uint8_t)VisualSensorMode::Fisheye) {
```
Skylion007, line 1127: NIT: Avoid C Style Casts

src/utils/viewer/viewer.cpp
```diff
@@ -1117,55 +1145,77 @@ void Viewer::drawEvent() {
...
+        renderCamera_->draw(objectPickingHelper_->getDrawables(), flags);
 
-      Mn::GL::Renderer::disable(Mn::GL::Renderer::Feature::Blending);
+        Mn::GL::Renderer::disable(Mn::GL::Renderer::Feature::Blending);
+      }
+      sensorRenderTarget->blitRgbaToDefault();
+    } else {
+      // ================ NON-regular RGB sensor ==================
+      std::string sensorId = "";
+      if (sensorMode_ == (uint8_t)VisualSensorMode::Fisheye) {
```
Skylion007, line 1202: See above

src/utils/viewer/viewer.cpp
```diff
@@ -56,13 +56,13 @@
 #include "esp/geo/VoxelUtils.h"
 #endif
 
+#include "esp/physics/configure.h"
```
Skylion007, line 59: Will the viewer build without Bullet enabled?

src/utils/viewer/viewer.cpp
mosra: I'm usually doing

```suggestion
    if (sensorMode_ == uint8_t(VisualSensorMode::Fisheye)) {
```

Which *technically* is a C-style cast as well I think(?), but `-Wold-style-cast` doesn't complain and it's short enough to not be annoying. Only using it for builtin types though.

mosra: On the other hand, I'm wondering why not just have `sensorMode_` as `VisualSensorMode` and do the casting on the *single* place where `+` and `%` is used on it, instead of casting all other cases? :)

PR #1181, opened 2021-04-20
[Sensor][Bugfix] Fix: depth info cannot be visualized correctly due to macOS GL driver bugs

bigbike:
## Motivation and Context
As titled.
The problem is that macOS GL driver has bugs. Even the depth test is disabled (depth buffer is supposed to be read only), it will still be overridden during the runtime. We do not have such problems in Linux (checked, no GL errors).
See the picture:
image (1)

## Types of changes

## Checklist


bigbike:
With this fix, the correct image is as follows. It is robust as it can produce correct results when window size changes.
Screen Shot 2021-04-19 at 11 12 00 PM

src/esp/gfx/Renderer.cpp
```diff
@@ -79,7 +82,27 @@ struct Renderer::Impl {
                 esp::gfx::Renderer::Impl::RendererShaderType::
                     DepthTextureVisualizer);
 
+#ifdef CORRADE_TARGET_APPLE
```
Skylion007, line 85: Please add a comment pointing as to why this is needed. IE pointing to this PR or containing a relevant explanation / bug report.

Skylion007 (changes requested):
@mosra

src/esp/gfx/Renderer.cpp
eundersander: Nit: you could define a preprocessor define like ENABLE_DEPTH_VISUALIZATION_WORKAROUND and use that everywhere in this file instead of CORRADE_TARGET_APPLE. It would make it easier to (1) test/investigate the old codepath on Mac, or (2) use the workaround on other platforms. You would define ENABLE_DEPTH_VISUALIZATION_WORKAROUND based on CORRADE_TARGET_APPLE.

mosra: @eundersander this is a good point :+1: 

On the engine side I'm having named workarounds that can be disabled at runtime, but that's probably an overkill here. (Though when we figure out the root cause, I'm planning to add such Apple-specific workaround.)

src/esp/gfx/Renderer.cpp
```diff
@@ -79,7 +82,27 @@ struct Renderer::Impl {
...
+        Mn::Image2D image = tgt.getDepthTexture().image(
+            0, {Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float});
```
mosra, line 87: Like we discussed last week (after the call i think?), this and `setSubImage()` below should be instead something like

```cpp
// after setStorage(), of course
tgt.getFramebuffer().copySubImage({{}, tgt.framebufferSize()}, *visualizedTex_, 0, {});
```

Or, architecture-wise, wrap this in a `RenderTarget::copyFrameDepth()` maybe? [Docs for copySubImage().](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1AbstractFramebuffer.html#a7669d9c7749e58c0462644194433d566)

This will do a much more efficient GPU-GPU copy without going through the CPU. I hope copy from a depth format to a R32F texture works, tho :O If it doesn't work, go through a `BufferImage` instead, which will also make the copy stay on the GPU.

src/esp/gfx/Renderer.cpp
```diff
@@ -79,7 +82,27 @@ struct Renderer::Impl {
...
+        Mn::ImageView2D imgView{image.storage(), Mn::PixelFormat::R32F,
+                                image.size(), image.data()};
+
+        if ((visualizedTex_ == Cr::Containers::NullOpt) ||
+            (visualizedTex_ != Cr::Containers::NullOpt &&
+             visualizedTex_->imageSize(0) != image.size())) {
```
mosra, line 93: Can be just 

```suggestion
        if (visualizedTex_ == Cr::Containers::NullOpt ||
             visualizedTex_->imageSize(0) != image.size()) {
```

due to the short-circuit evaluation of the boolean operators. Or even

```suggestion
        if (!visualizedTex_ ||
             visualizedTex_->imageSize(0) != image.size()) {
```

mosra (commented):
Nice that the workaround can be done in such a localized way :)

Skylion007 (approved):

Skylion007 (changes requested):
Address Reviewer Concerns

Skylion007:
@bigbike Anything block this or can we merge it?

src/esp/gfx/Renderer.cpp
```diff
@@ -79,7 +82,26 @@ struct Renderer::Impl {
                 esp::gfx::Renderer::Impl::RendererShaderType::
                     DepthTextureVisualizer);
 
+#ifdef CORRADE_TARGET_APPLE
+        Mn::Image2D image = tgt.getDepthTexture().image(
+            0, {Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float});
+        Mn::ImageView2D imgView{image.storage(), Mn::PixelFormat::R32F,
+                                image.size(), image.data()};
```
mosra, line 89:
```suggestion
        Mn::GL::BufferImage2D image = tgt.getDepthTexture().image(
            0, {Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float}, GL::BufferUsage::StaticRead);
        Mn::Vector2i size = image.size(); // because release() clears the size/storage, it has to be fetched first
        Mn::PixelStorage storage = image.storage();
        // This takes `image` (which is depth) and "reinterprets" it as R32F
        Mn::GL::BufferImage2D imgView{ // shouldn't be named imgView anymore I guess?
            storage, Mn::PixelFormat::R32F, size, image.release());
```

... I hope it can work directly like this, worst-case with minor corrections.

src/esp/gfx/RenderTarget.h
```diff
@@ -110,6 +110,11 @@ class RenderTarget {
    */
   void renderExit();
 
+  /**
+   * @brief get framebuffer
+   */
+  Magnum::GL::Framebuffer& getFramebuffer();
```
mosra, line 116: Not needed anymore, I guess?

src/esp/gfx/Renderer.cpp
```diff
@@ -79,7 +83,33 @@ struct Renderer::Impl {
                 esp::gfx::Renderer::Impl::RendererShaderType::
                     DepthTextureVisualizer);
 
+#ifdef CORRADE_TARGET_APPLE
+        Mn::GL::BufferImage2D image = tgt.getDepthTexture().image(
+	            0, {Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float}, Mn::GL::BufferUsage::StaticRead);
```
mosra, line 88: To save a bit more perf, you could keep the buffer image around instead of recreating and reallocating it every frame (so using the `image()` variant that takes a reference, not this one that returns it by-value).

~~But to actually make that work, you'd have to `img.release()` the buffer back to this depth format~~ wow, what a dumpster fire of an idea, disregard

So, make `image` a member to avoid the creation/allocation every frame, and then instead of the `release()` below do

```cpp
                Mn::GL::BufferImage2D imageAsRed{ // better name?
                    image.storage(), Mn::PixelFormat::R32F, image.size(), Mn::GL::Buffer::wrap(image.buffer().id(), Mn::ObjectFlag::Created), image.dataSize()};
```

... which basically creates a "buffer view" and should avoid all this verbose complicated ownership transfer.

Skylion007:
@bigbike Pull from master to solve the linting bugs

src/esp/gfx/RenderTarget.h
Skylion007: Bring out the dead code!

src/esp/gfx/Renderer.cpp
```diff
@@ -79,7 +87,37 @@ struct Renderer::Impl {
...
+            0, {Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float},
+            Mn::GL::BufferUsage::StaticRead);
+
+        // This takes the above output image (which is depth) and "reinterprets"
+        // it as R32F. In other words, the image below serves as an "image
+        // view".
+        Mn::GL::BufferImage2D clonedDepthImage{
+            depthBufferImage_->storage(), Mn::PixelFormat::R32F,
+            depthBufferImage_->size(),
+            Mn::GL::Buffer::wrap(depthBufferImage_->buffer().id(),
```
bigbike, line 101: FYI @mosra: the latest optimization attempt did not give us further performance gain. :)

bigbike:
This diff is ready to be reviewed again. 

The latest result is as follows.
Screen Shot 2021-04-27 at 6 01 09 PM

src/esp/gfx/Renderer.cpp
mosra: Yeah, not like that, you're still allocating a new image in every call to `image()` above. It has to be instead

```cpp
        if(!depthBufferImage_) // create a BufferImage instance, if not already
          depthBufferImage_.emplace(Mn::GL::PixelFormat::DepthComponent, Mn::GL::PixelType::Float);
        tgt.getDepthTexture().image(0, *depthBufferImage_, Mn::GL::BufferUsage::StaticRead);
```

src/esp/gfx/Renderer.cpp
```diff
@@ -143,8 +181,14 @@ struct Renderer::Impl {
...
   const Flags flags_;
-  Corrade::Containers::Optional<Mn::GL::Mesh> mesh_;
-  Magnum::ResourceManager<Mn::GL::AbstractShaderProgram> shaderManager_;
+  Cr::Containers::Optional<Mn::GL::Mesh> mesh_;
+  Mn::ResourceManager<Mn::GL::AbstractShaderProgram> shaderManager_;
+#ifdef ENABLE_VISUALIZATION_WORKAROUND_ON_MAC
+  Cr::Containers::Optional<Mn::GL::Texture2D> visualizedTex_ =
+      Cr::Containers::NullOpt;
+  Cr::Containers::Optional<Mn::GL::BufferImage2D> depthBufferImage_ =
+      Cr::Containers::NullOpt;
```
mosra, line 190: doesn't need to have the NullOpt initializer, that's the default

src/esp/gfx/Renderer.cpp
bigbike: OK. Thanks. Changed.
Haha, maybe 1 FPS improvement. 
Screen Shot 2021-04-28 at 11 15 05 AM

Skylion007 (approved):

src/esp/gfx/Renderer.cpp
mosra: Heh, okay. Well, who cares, it's just MacGL anyway :)

PR #1182, opened 2021-04-20
Managed rigid objects

jturner65:
## Motivation and Context
This PR provides direct user of physics objects via Managed Containers.  Users will be able to name objects on creation, and query objects by name.  The user will access objects directly to change their properties, instead of having to go through simulator and then physics manager to access an object.

Wrapper constructs, which own a weak pointer to the underlying physical object, will be given to the user to enable access to individual objects.  The wrapper object will verify the underlying object exists before attempting to access its functions; if the underlying object does not exist and a user attempts to access it, a warning message will be given to the user that this object no longer exists (and optionally, the object manager will be queried to verify that no copy of the invalid wrapper is stored).

The existing [Rigid Object Tutorial](https://github.com/facebookresearch/habitat-sim/blob/master/examples/tutorials/nb_python/rigid_object_tutorial.py) and commensurate [documentation](https://aihabitat.org/docs/habitat-sim/rigid-object-tutorial.html) have been modified to use [Object Managers and the Object wrappers](https://github.com/facebookresearch/habitat-sim/blob/ca2cb3e939b7c4362e0f3398ae02aa54b58e1ee5/examples/tutorials/nb_python/managed_rigid_object_tutorial.py)( [documentation](https://github.com/facebookresearch/habitat-sim/blob/ca2cb3e939b7c4362e0f3398ae02aa54b58e1ee5/docs/pages/managed-rigid-object-tutorial.rst) ) to accomplish the functionality that did go through Simulator before.

## How Has This Been Tested
Local c++ and python tests pass. Added tutorial successfully passes python test.

## Types of changes

## Checklist


Skylion007:
Clang-Tidy error: ```/home/circleci/project/habitat-sim/src/esp/bindings/PhysicsObjectBindings.cpp:39:49: error: parameter 'objType' is unused [misc-unused-parameters,-warnings-as-errors]```

Skylion007 (changes requested):

src/esp/physics/PhysicsObjectBase.h
```diff
@@ -0,0 +1,125 @@
...
+      : Magnum::SceneGraph::AbstractFeature3D(*bodyNode),
+        objectId_(objectId),
+        resMgr_(resMgr) {}
+
+  ~PhysicsObjectBase() override = default;
+
+  /**
+   * @brief Get the scene node being attached to.
+   */
+  scene::SceneNode& node() { return object(); }
```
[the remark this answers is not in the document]
jturner65: Object() is a function call about 4 lines down, which casts the feature container to be holding a scene node.

src/esp/physics/PhysicsManager.h
```diff
@@ -893,6 +892,11 @@ class PhysicsManager {
   /** @overload */
   scene::SceneNode& getObjectSceneNode(int physObjectID);
 
+  void setObjectLightSetup(const int objectID,
```
aclegg3, line 895: docstring

aclegg3 (commented):
Supporting changes look good. Maybe split into a prep PR?

Skylion007:
Woot broken doc strings!

jturner65:
> Woot broken doc strings!
Can we separate the sim and doc builds in that CI cell?  

Also seems like some warnings got promoted to errors too recently.  How come these don't fail when the CI changes are introduced (they're in files I haven't touched)?

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
```diff
@@ -0,0 +1,560 @@
...
+    agent_transform = place_agent(sim)
+
+    # get the primitive assets attributes manager
+    prim_templates_mgr = sim.get_asset_template_manager()
+
+    # get the physics object attributes manager
+    obj_templates_mgr = sim.get_object_template_manager()
+
+    # get the rigid object manager
+    rigid_obj_mgr = sim.get_rigid_object_manager()
```
erikwijmans, line 155: Based on naming, it seems like the plan is to have a different manager for rigid objects and articulated objects?  If possible having a single manager for both would be nice.  I'm not sure if this is feasible tho

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
```diff
@@ -0,0 +1,560 @@
...
+        boxes[b].translation = box_positions[b]
+        boxes[b].rotation = box_orientation
+
+    # get the object's initialization attributes (all boxes initialized with same mass)
+    object_init_template = boxes[0].creation_attributes
+    # anti-gravity force f=m(-g)
+    anti_grav_force = -1.0 * sim.get_gravity() * object_init_template.mass
+
+    # throw a sphere at the boxes from the agent position
+    sphere_template = obj_templates_mgr.get_template_by_ID(sphere_template_id)
```
erikwijmans, line 218: Looks like sometimes `ID` is capitalized and other times it isn't.  We should be consistent, my vote is uncapitalized to better fit with python style.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
```diff
@@ -0,0 +1,560 @@
...
+
+    # load some object templates from configuration files
+    sphere_template_id = obj_templates_mgr.load_configs(
+        str(os.path.join(data_path, "test_assets/objects/sphere"))
+    )[0]
+
+    # add a sphere to the scene, returns the object
+    sphere_obj = rigid_obj_mgr.add_object_by_id(sphere_template_id)
+    # move sphere
+    sphere_obj.translation = np.array([2.50, 0, 0.2])
```
erikwijmans, line 169: This is very pretty :)

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
```diff
@@ -0,0 +1,560 @@
...
+    # get the object's initialization attributes (all boxes initialized with same mass)
+    object_init_template = boxes[0].creation_attributes
+    # anti-gravity force f=m(-g)
+    anti_grav_force = -1.0 * sim.get_gravity() * object_init_template.mass
+
+    # throw a sphere at the boxes from the agent position
+    sphere_template = obj_templates_mgr.get_template_by_ID(sphere_template_id)
+    sphere_template.scale = np.array([0.5, 0.5, 0.5])
+
+    obj_templates_mgr.register_template(sphere_template)
```
erikwijmans, line 221: To me it seems like this should create a new template internally.  i.e. I would have expected
```
scaled_sphere_template_id = obj_templates_mgr.register_template(sphere_template)
```

Perhaps `update_template`?

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
```diff
@@ -0,0 +1,560 @@
...
+    observations += simulate(sim, dt=2.0, get_frames=make_video)
+
+    vel_control.controlling_ang_vel = False
+    vel_control.controlling_lin_vel = False
+
+    # simulate settling
+    observations += simulate(sim, dt=3.0, get_frames=make_video)
+
+    # remove the agent's body while preserving the SceneNode
+    rigid_obj_mgr.remove_object_by_ID(locobot.ID, delete_object_node=False)
```
erikwijmans, line 426: I assume you are not deleting the object node as otherwise that would delete the node for the agent?  If yes, I think we should have `add_object_by_id` take the parent node, not the node, as otherwise this seems like it'll be a frequent gotcha

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
```diff
@@ -0,0 +1,560 @@
...
+
+    # %%
+    # [embodied_agent_navmesh]
+
+    # load the lobot_merged asset
+    locobot_template_id = obj_templates_mgr.load_configs(
+        str(os.path.join(data_path, "objects/locobot_merged"))
+    )[0]
+    # add robot object to the scene with the agent/camera SceneNode attached
+    locobot = rigid_obj_mgr.add_object_by_id(
```
erikwijmans, line 448: Hmmm I see why it works that way.  This way you can easily attach a rigid object to the agent's scene node and then control it and its sensors will follow.

src/esp/physics/objectManagers/PhysicsObjectBaseManager.h
```diff
@@ -0,0 +1,158 @@
...
+  }  // PhysicsObjectBaseManager::registerObjectFinalize
+
+  /**
+   * @brief return a reference to physicsManager_, or null ptr if it does not
+   * exist anymore.  This is necessary since a reference of this manager may
+   * linger in python after simulator/physicsManager get clobbered/rebuilt.
+   */
+  std::shared_ptr<esp::physics::PhysicsManager> getPhysicsManager() const {
+    auto sp = weakPhysManager_.lock();
+    if (!sp) {
```
erikwijmans, line 117: Let's make this an ESP_CHECK that way we can throw an error.  It's good that keeping an object after the simulator is deleted won't cause a segfault, but I can't think of a reason why we shouldn't loudly inform the user that this object reference is  no longer controlling anything.  Can you also also expose an `isAlive()` method (or similar) that checks to make sure the object wrapper's physic manager is still alive and the ID still exists in the physics manager (if it's alive).  That way someone can easily do

```python
my_objs = list(filter(lambda obj: not obj.is_alive, my_objs))
```
to get rid of any dead objects?

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
erikwijmans: How hard would it be to track internally if the object wrapper owns the scene node?  If `add_object_by_id` was given a scene node it is assumed that the object wrapper doesn't own that node (and thus doesn't delete it by default).  Then `delete_object_node` has a default value of `None` where that gets inferred as `True` if the wrapper is an owner and `False` if it isn't.  The user can override this inference by setting the value explicitly.

@mosra IIRC `Corrade::Containers::Optional` doesn't have python bindings right?  pybind11 treats pointers as optionals so a `bool*` would work as the signature that get's bound to python and we can bind a lambda so that the cpp API doesn't have to be ugly :)

erikwijmans (commented):
Yay this is really nice!  Left some comments on API design to try and avoid sharp edges

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
mosra: It has them since https://github.com/mosra/magnum-bindings/commit/8e00226da3b01364064d29ba64b594ea567d79b7 (@Skylion007 requested this), but I think it's not pulled in by Habitat yet.

src/esp/physics/objectManagers/PhysicsObjectBaseManager.h
jturner65: Great idea re:isAlive(), thanks!

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
jturner65: > How hard would it be to track internally if the object wrapper owns the scene node? If `add_object_by_id` was given a scene node it is assumed that the object wrapper doesn't own that node (and thus doesn't delete it by default). Then `delete_object_node` has a default value of `None` where that gets inferred as `True` if the wrapper is an owner and `False` if it isn't. The user can override this inference by setting the value explicitly.
> 

Right now the wrapper owns nothing but a weak ref to the object itself.  The manager just passes the arguments for add and remove into the physics manager.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -3,16 +3,28 @@
...
+      rigidObjectManager_(RigidObjectManager::create()) {}
+
+bool PhysicsManager::initPhysics(
+    scene::SceneNode* node,
+    const std::shared_ptr<esp::physics::PhysicsManager>& physMgr) {
   physicsNode_ = node;
-
+  // set the rigidObjectManager's weak reference to physics manager tp be based
+  // on the same shared pointer that Simulator is using.
+  rigidObjectManager_->setPhysicsManager(physMgr);
```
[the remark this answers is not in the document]
jturner65: I was passing Simulator's pointer around because I wanted to always have that specific pointer (to stay synched with Simulator).  I don't think having 2 shared pointers pointing at same PM would behave the same way, would it? If sim->physMgrPtr starts pointing at another physMgr, rigidObjectMgr->weakPhysMgr will not be pointing at the same thing, will it?

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
Skylion007: @erikwijmans id is a python builtin though, so when naming args as id it will cause issues. :(

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
Skylion007: I don't think the np.array is even necessary here (as long as all values are float). Can you check?

src/esp/physics/objectManagers/PhysicsObjectBaseManager.h
mosra: (Apologies if my comment is totally out of place, I don't know much about this area of Habitat's code.) Isn't it possible to ensure the originating object stays around, instead of emitting an error or managing "dead objects"? I'm doing this for a lot of Magnum objects already (array views keep the originating memory owner around, meshes keep their buffers around etc.).

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
Skylion007: I don't like the overloading of the term ID anyhow, key or something would be better in the longrun, no? or alternatively uuid to be consistent with Sensors etc?

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
mosra: Was about to comment on the same :) If `translation` is a magnum vector, assigning a float 3-tuple / list should work.

src/esp/physics/objectManagers/PhysicsObjectBaseManager.h
Skylion007: @mosra I think the main issue is memory reload/restarts, right? Also you could theoretically hotswap managers so those errors would be needed if I understand correctly, right?

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
mosra: `index`? :)

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
mosra: Was about to comment on the same, if `translation` is a magnum vector, assigning a float 3-tuple / list should be enough.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
jturner65: @aclegg3 and I have been discussing how we can sync the underlying rigid object/articulated object functional overlap, so this may be realizable, but from what I know now, it will probably not be possible in the first pass of the Managed AO wrapper functionality.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
```diff
@@ -0,0 +1,560 @@
...
+    # build multiple object initial positions
+    box_positions = [
+        np.array([2.39, -0.37, 0]),
+        np.array([2.39, -0.64, 0]),
+        np.array([2.39, -0.91, 0]),
+        np.array([2.39, -0.64, -0.22]),
+        np.array([2.39, -0.64, 0.22]),
+    ]
+    box_orientation = mn.Quaternion.rotation(
+        mn.Rad(math.pi / 2.0), np.array([-1.0, 0, 0])
```
mosra, line 203:
```suggestion
        mn.Deg(90.0), np.array([-1.0, 0, 0])
```

Works the same as in C++, degrees and radians convert implicitly to each other.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
jturner65: Thanks for the suggestion, this looks a lot better.

src/esp/physics/PhysicsManager.h
```diff
@@ -1045,6 +1095,23 @@ class PhysicsManager {
...
+   * @brief returns the wrapper manager for the currently created rigid objects.
+   * @return RigidObject wrapper manager.
+   */
+  std::shared_ptr<RigidObjectManager> getRigidObjectManager() {
+    return rigidObjectManager_;
+  }
+
+  /**
+   * @brief Check if @p ohysObjectID represents an existing object.
+   * @param ohysObjectID Object ID to check
```
Skylion007, line 1108: typo

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
erikwijmans: Looks like I got terminology wrong here.  If you swap "object wrapper" for "object" does that make sense?

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
erikwijmans: Yeah, for arg names and properties `object_id` makes sense, but I don't think we need to be that explicit in method names.  I don't have a strong preference between id and uuid.  uuid wouldn't be confused with python's builtin `id()` (but no one uses that in our world AFAIK) but no one uses uuid in our world either -- e.g. we reference to things by their instance id in instance segmentation, not their instance uuid -- so uuid isn't a natural choice.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
aclegg3: Agreed it is ugly and leads to easy pitfalls. There is a minor complication here. Physics will update the object's node, so if we create a child internally and attach a RigidObject feature to the child, the agent node with the sensors won't be affected by changes to the object such as dynamics.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
aclegg3: @jturner65 Correct me if I'm wrong here. I think it serves a dual purpose, right? You can get a copy of an attributes template (or a fresh one), modify it, and then register it with a new handle as a new cached template. Alternatively you can just update an existing attributes template using the default arguments. Do we want to separate these? In my mind this is very similar to a C++ std::map [] operator.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
aclegg3: Agreed we should pick one consistent naming scheme. In this case the id's refer to unique integer values rather than string keys/handles, so we should keep that distinction. Overlap with python id() is unfortunate, but semantically correct. uuid, object_id, hab_id, etc... would be my preference.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
aclegg3: Both types of objects already share the same uuid pool, but will have different wrapper class types and interfaces. I agree it would be great if these interfaces overlap as much as possible and to have a unified "object_manager". I think this is possible and agree it would be ideal. Perhaps this will come with AO merge, but may need to be a follow-up. Either way it should be a minimal refactor since it only affects create/get/remove of objects and not manipulation.

examples/tutorials/nb_python/managed_rigid_object_tutorial.py
jturner65: I'd rather not separate them - that would be a fairly substantial refactor to the managed container.  All the functionality is accomplished currently based on context.

src/esp/physics/PhysicsManager.cpp
```diff
@@ -3,16 +3,28 @@
...
+    : resourceManager_(_resourceManager),
+      physicsManagerAttributes_(_physicsManagerAttributes),
+      rigidObjectManager_(RigidObjectManager::create()) {}
+
+bool PhysicsManager::initPhysics(
+    scene::SceneNode* node,
+    const std::shared_ptr<esp::physics::PhysicsManager>& physMgr) {
   physicsNode_ = node;
-
+  // set the rigidObjectManager's weak reference to physics manager tp be based
```
aclegg3, line 25:
```suggestion
  // set the rigidObjectManager's weak reference to physics manager to be based
```

src/esp/sim/Simulator.h
```diff
@@ -719,6 +731,25 @@ class Simulator {
    */
   bool isNavMeshVisualizationActive();
 
+  /**
+   * @brief Return a ref to a new drawables in the currently active scene, for
+   * object creation.  Eventually support multi-scene ID
+   * @param sceneID The scene to get the drawables for.  Currently not used.
+   */
+  inline esp::gfx::DrawableGroup& getDrawables(
```
aclegg3, line 739: Minor naming suggestion.
```suggestion
  inline esp::gfx::DrawableGroup& getDrawableGroup(
```

src/esp/physics/objectManagers/PhysicsObjectBaseManager.h
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * @brief Class template defining responsibilities and functionality for
+ * managineg object wrappers specializing @ref
+ * esp::physics::AbstractManagedPhysicsObject template.
+ * @tparam T the type of managed physics object wrapper a particular
+ * specialization of this class works with.  Must inherit from @ref
+ * esp::physics::AbstractManagedPhysicsObject
+ * @tparam Access Whether the default access (getters) for this
+ * container provides copies of the objects held, or the actual objects
+ * themselves.
```
aclegg3, line 32:
```suggestion
 * @brief Class template defining responsibilities and functionality for
 * managing object wrappers specializing @ref
 * esp::physics::AbstractManagedPhysicsObject template.
 * @tparam T the type of managed physics object wrapper a particular
 * specialization of this class works with.  Must inherit from @ref
 * esp::physics::AbstractManagedPhysicsObject
```

src/esp/physics/objectManagers/RigidBaseManager.h
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * @brief Class template defining responsibilities and functionality shared for
+ * managing all @ref esp::physics::ManagedRigidBase wrappers.
+ * @tparam T the type of managed physics object a particular specialization
+ * of this class works with.  Must inherit from @ref
+ * esp::physics::ManagedRigidBase
+ * @tparam Access Whether the default access (getters) for this
+ * container provides copies of the objects held, or the actual objects
+ * themselves.
```
aclegg3, line 21:
```suggestion
```

src/esp/physics/objectManagers/RigidObjectManager.h
```diff
@@ -0,0 +1,133 @@
...
+
+#ifndef ESP_PHYSICS_RIGIDOBJECTMANAGER_H
+#define ESP_PHYSICS_RIGIDOBJECTMANAGER_H
+
+#include "RigidBaseManager.h"
+#include "esp/physics/objectWrappers/ManagedRigidObject.h"
+namespace esp {
+namespace physics {
+
+class RigidObjectManager
```
aclegg3, line 13: class doc-string

src/esp/physics/objectManagers/RigidObjectManager.h
```diff
@@ -0,0 +1,133 @@
...
+ public:
+  RigidObjectManager()
+      : esp::physics::RigidBaseManager<ManagedRigidObject>::RigidBaseManager(
+            "RigidObject") {
+    this->buildCtorFuncPtrMaps();
+  }
+
+  /** @brief Instance a physical object from an object properties template in
+   * the @ref esp::metadata::managers::ObjectAttributesManager.  This method
+   * calls the physicsManager method with the same signature that queries for an
```
aclegg3, line 24:
```suggestion
   * calls the physicsManager method with the same signature that queries for a
```

src/esp/physics/objectWrappers/ManagedRigidObject.h
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef ESP_PHYSICS_MANAGEDRIGIDOBJECT_H_
+#define ESP_PHYSICS_MANAGEDRIGIDOBJECT_H_
+
+#include "ManagedRigidBase.h"
+#include "esp/physics/RigidObject.h"
+
+namespace esp {
+namespace physics {
+class ManagedRigidObject
```
aclegg3, line 13: class doc-string

aclegg3 (approved):
Overall looks great for initial PR. Some minor typos and suggestions.

Skylion007 (approved):

Skylion007:
@jturner65 Nit: Could these gifs be encoded as WebMs instead? Would take up a lot less space and be higher quality.

src/esp/core/managedContainers/ManagedContainerBase.cpp
```diff
@@ -124,5 +124,34 @@ ManagedContainerBase::getObjectHandlesBySubStringPerType(
...
+  int incr = 0;
+  if (resVals.size() != 0) {
+    // handles exist with passed substring.  Find highest handle increment, add
+    // 1 and use for new name 1, build new handle
+    for (const std::string& s : resVals) {
+      // split string on underscore, last value will be string of highest incr
+      // value existing.
+      std::vector<std::string> vals = Cr::Utility::String::split(s, '_');
+      // if any exist, all are expected to end
+      int new_incr = std::stoi(vals.back());
```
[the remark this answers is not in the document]
jturner65: I can use a pivot token that will never come up in a file name due to being illegal - maybe "/:" to catch both linux and mac.  After all, this is just the name of the instance of the object, and will never go out to disk.  I have to handle 2 different cases - the very first object from a template, which will use <template_title> to build a name <template_title>/:0000,  and the subsequent objects <template_title>/:0000 -> <template_title>/:0001.

jturner65:
> @jturner65 Nit: Could these gifs be encoded as WebMs instead? Would take up a lot less space and be higher quality.

That should probably be done with all the doc gifs.  I can certainly do that, but I'd rather do that in a separate PR, since it will hit all the images.

src/esp/physics/PhysicsManager.cpp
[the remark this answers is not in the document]
erikwijmans: `shared_from_this()` will grab that same shared pointer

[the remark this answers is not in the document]
jturner65: Cool, I didn't know that.

PR #1183, opened 2021-04-20
Support for shader_type specification in stage and object configurations - Part 1 of 2

jturner65:
## Motivation and Context
[(Replaces this PR which included wrong commit)](https://github.com/facebookresearch/habitat-sim/pull/1134)
This PR is the first of 2 PRs intended to introduce configuration-file level support for independent shader types being specified for the stage and objects in a scene. Supported shader types that the user can specify are Flat, Phong and PBR (case insensitive), and these values can be specified in either stage and object configurations (to specify a default value for a particular stage or object TYPE), or in the scene instance configurations (which, if specified, will override the default values given in stage or object configs for a single instance only).

To support this functionality, stage and object creation had to be slightly modified in order to accept a copy of the appropriate attributes as a parameter, which is then moved to the initialization Attributes of the resultant rigidStage/rigidObject.

Part 1 of 2 : This PR only implements the framework required for this change, but lacks the functionality to consume the configuration-specified shader type in ResourceManager. The lacking functionality will be added in Part 2 of 2.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/attributes/ObjectAttributes.h
+++ b/src/esp/metadata/attributes/ObjectAttributes.h
@@ -14,6 +14,31 @@ namespace metadata {
 namespace attributes {
 
 /**
+ * @brief This enum class defines the possible shader options for rendering
+ * instances of objects or stages in Habitat-sim.
+ */
+enum class ObjectInstanceShaderType {
+  /**
+   * Represents an unknown/unspecified value for the shader type to use. Resort
+   * to defaults for object type.
+   */
+  Unknown = -1,
+  /**
+   * Refers to flat shading, pure color and no lighting.  This is often used for
+   * textured objects
+   */
+  Flat = 0,
+  /**
+   * Refers to phong shading with pure diffuse color.
+   */
+  Phong = 1,
+  /**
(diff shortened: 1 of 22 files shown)


src/esp/metadata/attributes/ObjectAttributes.h
```diff
@@ -13,6 +13,31 @@ namespace esp {
...
+  /**
+   * Represents an unknown/unspecified value for the shader type to use. Resort
+   * to defaults for object type.
+   */
+  Unknown = -1,
+  /**
+   * Refers to flat shading, pure color and no lighting.  This is often used for
+   * textured objects
+   */
+  Flat = 0,
```
eundersander, line 30: Beware, our "flat-shaded" objects actually use a special material along with the Phong shader, if I'm not mistaken. ResourceManager::buildFlatShadedMaterialData .

aclegg3: I think the purpose of calling these "flat" is to get away from specifying "no-lights" even though that is closer to the actual implementation.

aclegg3 (approved):
This looks good to me.

PR #1184, opened 2021-04-20
Change LOG messages in Managers and MM to Corrade Debugs

jturner65:
## Motivation and Context
This PR will change the various log messages to use the Corrade provided Debug constructs.  This cuts down on the console output stream during simulator creation and execution.

## How Has This Been Tested

All c++ and python tests pass
## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/managedContainers/ManagedContainer.h
+++ b/src/esp/core/managedContainers/ManagedContainer.h
@@ -119,8 +119,8 @@ class ManagedContainer : public ManagedContainerBase {
                      const std::string& objectHandle = "",
                      bool forceRegistration = false) {
     if (nullptr == managedObject) {
-      LOG(ERROR) << "ManagedContainer::registerObject : Invalid "
-                    "(null) managed object passed to registration. Aborting.";
+      Mn::Error{} << "ManagedContainer::registerObject : Invalid "
+                     "(null) managed object passed to registration. Aborting.";
       return ID_UNDEFINED;
     }
     if ("" != objectHandle) {
@@ -129,9 +129,9 @@ class ManagedContainer : public ManagedContainerBase {
     }
     std::string handleToSet = managedObject->getHandle();
     if ("" == handleToSet) {
-      LOG(ERROR) << "ManagedContainer::registerObject : No "
-                    "valid handle specified for "
-                 << objectType_ << " managed object to register. Aborting.";
+      Mn::Error{} << "ManagedContainer::registerObject : No "
+                     "valid handle specified for "
+                  << objectType_ << " managed object to register. Aborting.";
       return ID_UNDEFINED;
     }
     return registerObjectFinalize(managedObject, handleToSet,
@@ -495,9 +495,9 @@ class ManagedContainer : public ManagedContainerBase {
       return getObjectInternal<T>(objectHandle)->getID();
     } else {
       if (!getNext) {
-        LOG(ERROR) << "ManagedContainer::getObjectIDByHandleOrNew : No "
-                   << objectType_ << " managed object with handle "
-                   << objectHandle << "exists. Aborting";
+        Mn::Error{} << "ManagedContainer::getObjectIDByHandleOrNew : No "
+                    << objectType_ << " managed object with handle "
+                    << objectHandle << "exists. Aborting";
         return ID_UNDEFINED;
       } else {
         return getUnusedObjectID();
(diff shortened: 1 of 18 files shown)


erikwijmans:
I'm not sure if I like the motivation for doing this.  The logging from Magnum can be very important to us, e.g. for OpenGL/EGL context debugging, so keeping Magnum logs as Magnum focused as possible seems beneficial.  While I like the idea of moving some of our noisier logging into a different scope, I'm not sure if throwing it into the Magnum scope is a good idea in the long run.

jturner65:
> I'm not sure if I like the motivation for doing this. The logging from Magnum can be very important to us, e.g. for OpenGL/EGL context debugging, so keeping Magnum logs as Magnum focused as possible seems beneficial. While I like the idea of moving some of our noisier logging into a different scope, I'm not sure if throwing it into the Magnum scope is a good idea in the long run.

Thanks for your input Erik.  I would extend your excellent point to include customizable logging output in general - I know there are many times I would prefer to isolate, or at least be able to easily separate, output logs based on "subsystem" (which could perhaps be easily accomplished by having independent, assignable logging output per namespace).  Perhaps something as simple as being able to assign output to file, console, or none based on "subsystem".

That being said, aren't the Magnum Debugs just an alternative to the LOG macro outputs?  They have already been added to the voxelization functionality, and I was under the impression we were migrating to them across the system.  

@aclegg3 and I discussed this some, and we will be bringing it up at our Thursday Sim Infra meeting, to get some input on this.

mosra:
> The logging from Magnum can be very important to us, e.g. for OpenGL/EGL context debugging, so keeping Magnum logs as Magnum focused as possible seems beneficial.

Interesting view, I never looked at it like this. From my side, I *do* want applications and user code to use Magnum's debug output, as compared to any 3rd party solution it works much better with Magnum's builtin types, and the whole log management and capturing works better when there's a single system for all. When I get out of the "oh god late with everything" state, I'd like to implement various missing functionality like output prefixes or timestamps. This particular concern seems like it could get solved with an ability to specify (and filter) "message source", right? Like, "keep all engine-level messages but ignore application-level" etc.

> @aclegg3 and I discussed this some, and we will be bringing it up at our Thursday Sim Infra meeting

@jturner65 sorry about hijacking the whole meeting with the shader presentation, let's discuss that this week instead? :)

jturner65:
> This particular concern seems like it could get solved with an ability to specify (and filter) "message source", right? Like, "keep all engine-level messages but ignore application-level" etc.

I think it would be great if you could assign messages to a particular subsystem, and then filter on this assignment.

> @jturner65 sorry about hijacking the whole meeting with the shader presentation, let's discuss that this week instead? :)

Hey that presentation was definitely worth it. :)

mosra:
As I mentioned on the call, here's the quick idea for Habitat-side automagical subsystem-based output redirection. It relies on `__FILE__`, ~~because that's AFAIK the only real way to know which file we're in at the preprocessor level~~ (see below), the other option (if you wouldn't want a macro) would be `std::source_location` internals that `!Debug` is using, unfortunately those are not *that* portable yet (older Clangs and MSVC don't have those, which also means `!Debug` doesn't produce file/line info there).

Relying on a file means the namespaces have to correspond to filesystem layout, but I think Habitat follows that pretty well. Plus for *really pathological cases* you can always do some advanced string parsing.

```cpp
namespace esp { namespace logging {
    
enum class Subsystem {
    Asset,
    Gfx,
    Scene,
    // ...
};

// whether logging for a particular subsystem is enabled
bool isEnabled(Subsystem);
    
// using a StringView because it has contains(), std::string has that only 
// since C++23 (!!!)
Magnum::Debug outputFor(Corrade::Containers::StringView file) {
    std::ostream* output;
    const char* prefix;
    if(file.contains("/esp/gfx/") && isEnabled(Subsystem::Gfx)) {
        output = somewhere;
        prefix = "[gfx]";
    } else if(file.contains("/esp/scene/") && isEnabled(Subsystem::Gfx)) {
        output = somewhere;
        prefix = "[scene]";
    // ...
    } else {
        return Magnum::Debug{nullptr};
    } 
    
    Magnum::Debug out{output};
    out << prefix;
    return out;
}
    
}}

#define LOG() esp::logging::outputFor(__FILE__)
```

Usage is simple:

```cpp
LOG() << "ha!";
```

------

EDIT: Actually I just got another idea, which doesn't cause every debug output call to embed the full filename (which Emscripten would certainly love) and instead of potentially heavy string parsing it relies on some (I hope robust enough) function resolution trickery:

```cpp
namespace esp { 
    
// EVERY namespace you want to use LOG() in has to have this
namespace gfx { 
    logging::Subsystem loggingSubsystem() { return logging::Subsystem::Gfx; }
}
namespace scene { 
    logging::Subsystem loggingSubsystem() { return logging::Subsystem::Scene; }
}
// ...

namespace logging {

Magnum::Debug outputFor(logging::Subsystem subsystem) {
    std::ostream* output;
    const char* prefix;
    if(isEnabled(subsystem)) {
        Magnum::Debug out{outputForSubsystem(subsystem)};
        out << prefixForSubsystem(subsystem);
        return out;
    } 
    
    return Magnum::Debug{nullptr};
}
    
}
   
}

#define LOG() esp::logging::outputFor(loggingSubsystem())
```

jturner65:
Replaced by https://github.com/facebookresearch/habitat-sim/pull/1382

PR #1189, opened 2021-04-22
Managed rigid object prep

jturner65:
## Motivation and Context
This PR sets up some necessary prep work for the Managed rigid objects work and is [split off from this PR](https://github.com/facebookresearch/habitat-sim/pull/1182). 

This PR introduces a new base class for RigidBase that will hold required information to be used by the Managed Containers.  These constructs will be wrapped in a wrapper class that will reference the underlying object via a weak pointer, to facilitate safe access from python.

## How Has This Been Tested
Existing python and c++ tests pass.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletRigidObject.cpp
+++ b/src/esp/physics/bullet/BulletRigidObject.cpp
@@ -306,7 +306,12 @@ void BulletRigidObject::setCollisionFromBB() {
 }  // setCollisionFromBB
 
 void BulletRigidObject::setMotionType(MotionType mt) {
-  if ((mt == MotionType::UNDEFINED) || (mt == objectMotionType_)) {
+  if (mt == MotionType::UNDEFINED) {
+    LOG(WARNING) << "BulletRigidObject::setMotionType : Cannot set motion type "
+                    "to MotionType::UNDEFINED.  Aborting.";
+    return;
+  }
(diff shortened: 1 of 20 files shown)


src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -305,25 +305,27 @@ void BulletRigidObject::setCollisionFromBB() {
   }
 }  // setCollisionFromBB
 
-bool BulletRigidObject::setMotionType(MotionType mt) {
+void BulletRigidObject::setMotionType(MotionType mt) {
   if (mt == MotionType::UNDEFINED) {
-    return false;
+    LOG(WARNING) << "BulletRigidObject::setMotionType : Cannot set motion type "
```
Skylion007, line 310: Shouldn't this be an ERROR level logging?

jturner65: I don't think so - this would only be accessed by a user directly attempting to set the motion type to an undefined value.  Originally, the process only returned "false" if attempting to set the motion type to UNDEFINED, and if the user was not explicitly checking for a return value, the issue was totally missed.

aclegg3 (approved):
Looks good once the tests pass.

PR #1190, opened 2021-04-22
[bugfix] - Update the broad-phase aabb of an object when manually setting state

aclegg3:
## Motivation and Context

Setting world transform of an object in Bullet does not update the broadphase AABB automatically. This results in a bug where collision queries which do not perform a full discrete collision check (most prominently ray casting) can miss objects with manual kinematic updates since the last simulation or collision check. 

## How Has This Been Tested

`test_raycast` was missing this bug because the test ray originated at the world origin where objects are spawned, so the broad-phase was triggered for moved objects anyway. This has been addressed and golden numbers updated.

## Types of changes

## Checklist


eundersander (approved):
LGTM. Nice catch.

PR #1194, opened 2021-04-24
[RECAST] Recast Navigation update

Skylion007:
## Motivation and Context

* This updates the RecastNavigation submodule with new performance improvements/bugfixes (as well as a few new C++ APIs). 
* Biggest change is a significant decrease in memory allocations needed while recomputing the NavMesh. Should be very useful since we are recomputing the navmesh dynamically more often especially with inserted objects.

## How Has This Been Tested
* By CircleCI / existing unit tests.

## Types of changes

## Checklist


Skylion007:
Edit: Also found a bug in our CMake and merged that into this PR.

src/cmake/dependencies.cmake
```diff
@@ -73,7 +73,7 @@ if(BUILD_ASSIMP_SUPPORT)
   if(NOT USE_SYSTEM_ASSIMP)
     set(ASSIMP_BUILD_ASSIMP_TOOLS OFF CACHE BOOL "ASSIMP_BUILD_ASSIMP_TOOLS" FORCE)
     set(ASSIMP_BUILD_TESTS OFF CACHE BOOL "ASSIMP_BUILD_TESTS" FORCE)
-    set(BUILD_SHARED_LIBS OFF CACHE BOOL "ASSIMP_BUILD_TESTS" FORCE)
+    set(BUILD_SHARED_LIBS OFF CACHE BOOL "BUILD_SHARED_LIBS" FORCE)
```
Skylion007, line 76: This was the typo I found in the CMake.

src/cmake/dependencies.cmake
```diff
@@ -102,7 +102,8 @@ endif()
 set(RECASTNAVIGATION_DEMO OFF CACHE BOOL "RECASTNAVIGATION_DEMO" FORCE)
 set(RECASTNAVIGATION_TESTS OFF CACHE BOOL "RECASTNAVIGATION_TESTS" FORCE)
 set(RECASTNAVIGATION_EXAMPLES OFF CACHE BOOL "RECASTNAVIGATION_EXAMPLES" FORCE)
-set(RECASTNAVIGATION_STATIC ON CACHE BOOL "RECASTNAVIGATION_STATIC" FORCE)
+set(BUILD_SHARED_LIBS OFF CACHE BOOL "BUILD_SHARED_LIBS" FORCE)
```
Skylion007, line 105: Recast now uses the more standard BUILD_SHARE_LIBS CMake config to determine whether to do a static build.

Now also requires the GNUInstallDirs to be set.

mosra: Just a note -- on older CMake, Bullet *has to be* built as shared because the static subproject build is broken beyond repair: https://github.com/facebookresearch/habitat-sim/blob/e821de1fbb4198353a1f7f1f1db940942fa5ba6c/src/cmake/dependencies.cmake#L164-L178 

Which points to a problem with the global `BUILD_SHARED_LIBS` -- there's no way to set it differently for different subprojects, which is why Magnum has its own variable as well. Is it possible to control this with some Recast-specific variable? Otherwise we force it to `OFF` in one place and to `ON` in another and it's ... a mess :)

Skylion007: @mosra @aclegg3 Unfortunately, no the RECAST specific variable was removed in an update. Can we store it to a temp variable and reset it somewhere? This already is an issue for the ASSIMP submodule anyway so it's not like it's really introducing a new problem. Thoughts?

Skylion007: @mosra Relevant PR: https://github.com/recastnavigation/recastnavigation/pull/439

src/cmake/dependencies.cmake
```diff
@@ -102,7 +102,8 @@ endif()
...
+include(GNUInstallDirs)
 add_subdirectory("${DEPS_DIR}/recastnavigation/Recast")
 add_subdirectory("${DEPS_DIR}/recastnavigation/Detour")
```
mosra, line 108:
```suggestion
# Temp BUILD_SHARED_LIBS override for Recast, has to be reset back after to avoid affecting Bullet (which needs shared) and potentially other submodules
set(_PREV_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
set(BUILD_SHARED_LIBS OFF)
include(GNUInstallDirs)
add_subdirectory("${DEPS_DIR}/recastnavigation/Recast")
add_subdirectory("${DEPS_DIR}/recastnavigation/Detour")
set(BUILD_SHARED_LIBS ${_PREV_BUILD_SHARED_LIBS})
```

Let me know if this works. I hope it does, if not, then we may need to go through the cache instead (the only difference is added `CACHE BOOL "" FORCE` in two places):

```suggestion
# Temp BUILD_SHARED_LIBS override for Recast, has to be reset back after to avoid affecting Bullet (which needs shared) and potentially other submodules
set(_PREV_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
set(BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
include(GNUInstallDirs)
add_subdirectory("${DEPS_DIR}/recastnavigation/Recast")
add_subdirectory("${DEPS_DIR}/recastnavigation/Detour")
set(BUILD_SHARED_LIBS ${_PREV_BUILD_SHARED_LIBS} CACHE BOOL "" FORCE)
```

Skylion007: It worked.

mosra (approved):

aclegg3 (approved):

PR #1195, opened 2021-04-26
Managed file based container for Managed Container File IO

jturner65:
## Motivation and Context
This PR moves the file-io based functionality for ManagedContainers to a ManagedFIleBasedContainer, and it moves the ManagedContainer code to its own directory in core.  Some managed objects, such as the PhysicsObjectWrappers, have no need of file i/o functionality as part of their creation process,  since they will never be loaded/created from configuration files.  By moving the functionality out of ManagedContainer, the containers that manage these non-file-based objects do not need to implement useless functions by specializing ManagedContainer directly.

## How Has This Been Tested
All existing C++ and Python tests pass

## Types of changes

## Checklist


src/esp/bindings/AttributesBindings.cpp
```diff
@@ -43,9 +44,14 @@ void initAttributesBindings(py::module& m) {
   // NOLINTNEXTLINE(bugprone-unused-raii)
   py::class_<AbstractManagedObject, AbstractManagedObject::ptr>(
       m, "AbstractManagedObject");
+  // ==== AbstractFileBasedManagedObject ====
+  // NOLINTNEXTLINE(bugprone-unused-raii)
+  py::class_<AbstractFileBasedManagedObject, esp::core::AbstractManagedObject,
```
Skylion007, line 49: Hold on. Are you sure don't want this inherit from the Python ```AbstractManagedObject``` binding? That should remove the need for NoLintNextLine too (it won't complain about unused raii if it's used in inheritance.)

src/esp/bindings/AttributesBindings.cpp
```diff
@@ -43,9 +44,14 @@ void initAttributesBindings(py::module& m) {
...
+             AbstractFileBasedManagedObject::ptr>(
+      m, "AbstractFileBasedManagedObject");
 
   // ==== AbstractAttributes ====
-  py::class_<AbstractAttributes, esp::core::AbstractManagedObject,
+  py::class_<AbstractAttributes, esp::core::AbstractFileBasedManagedObject,
```
Skylion007, line 54: Likewise for here. Why are you trying so hard to hide the inheritance from Python?

jturner65: Not trying hard at all :).  As I understood it when I first coded this last summer, it could be done either way.  I saw no benefit exposing them to python.

src/esp/bindings/AttributesBindings.cpp
Skylion007: Ah, apparently this is a VALID way of binding the inheritance of PyBind11 as it will autoresolve, but if the class is already bound (which it is above), it's probably better to have the parent be the already PyBinded class. I think it will remove the need for the NoLint as an added bonus.

src/esp/bindings/AttributesBindings.cpp
Skylion007: Ah nvm, just confused because one is bound with the namespace and the other is not, which is a bit odd.

src/esp/bindings/AttributesBindings.cpp
Skylion007: disregard.

src/esp/bindings/AttributesBindings.cpp
jturner65: Would there be some benefit that I am overlooking, other than getting rid of the 2nd NoLintNextLine?  If you think this would be a good change, I'm certainly open to it. :)

aclegg3 (approved):

PR #1198, opened 2021-04-26
Data download utility

aclegg3:
## Motivation and Context

As more Habitat-ready data is generated and system updates require changes to existing data, we want to provide a straightforward, unified utility for users (and automated tests such as CI) to download and prepare correctly versioned assets and datasets with minimal fuss. 

**This PR introduces a python utility script which defines unique ids and data-prep conventions for available example assets  tied to a particular version of the codebase.** 

Features/conventions: (feedback welcome)
- Each data-source has a unique_id defined in the util file.
- Data-sources can be grouped with a new unique_id to simplify multi-package downloads.
- Each data-source must have a self-contained target directory which cannot be shared with another data-source (to facilitate versioning and cleaning).
  - This choice will result in changes to existing example data setup:
    - E.g. `data/objects` will now have sub-directories `data/objects/example_objects/` and `data/objects/locobo_merged/`
- To prevent accidental deletion any existing data directories will require user input or command-line argument `--replace` to be deleted and overwritten by new versions.
- Versioning conventions:
  - A new directory data/versioned_data/ will contain separate directories for each version of each datasource with a tag configured directly in the util.
  - Directory symlinks will be generated in the specified data path (e.g. `data/objects/example_objects`) to the current version directory to allow quick swapping between data versions without re-downloading.
  - existing data-sources with unspecified versions are now `1.0`

All example/test assets can now be downloaded/unpacked with 1 line (e.g. on CI):
`python habitat_sim/utils/datasets_download.py --uids ci_test_assets --replace`
and removed with:
`python habitat_sim/utils/datasets_download.py --uids ci_test_assets --clean`
a custom data directory can be specified:
`python habitat_sim/utils/datasets_download.py --uids ci_test_assets --data_path "my/custom/path"`

**Future Work:**
- Add test_assets to this framework?
- Support external representation of data-source config? (e.g. JSON?)
- Support workflow for protected data (e.g. MP3D) requiring external password or URL.

## How Has This Been Tested
Local testing and integrated with CI.

## Types of changes

## Checklist


habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+import os
+
+# TODO: is there a better way to do this more generaly? Maybe pybind a cmake compiled DATA_DIR?
```
erikwijmans, line 10: We probably want to let users specify a path and have the default path just be relative to their current working directory.  Doing relative to this file won't work for the conda install

dhruvbatra:
Super valuable utility! Thanks for making this @aclegg3!

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+dir_path = os.path.dirname(os.path.realpath(__file__))
+data_dir = os.path.abspath(os.path.join(dir_path, "../../data/")) + "/"
+
+# dict keyed by uids with download specs for various test assets and datasets
+data_sources = {
+    # "example-uid": {
+    #
+    # }
+    "habitat-test-scenes": {
```
Skylion007, line 19: Instead of a nested dicts, would an attr.s class make more sense here?

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def clean_data(uid):
+    if uid not in data_sources:
+        print("Data download failed, no datasource named " + uid)
+        return
+    print(
+        "Cleaning datasource " + uid + ' directory: "' + data_sources[uid]["root"] + '"'
+    )
+    os.system("rm -r " + data_sources[uid]["root"])
```
Skylion007, line 93: Use shutil.rmtree instead here. Very bad to use rm like this. What could go wrong... 0_0

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+        if "download_pre_args" not in data_sources[uid]
+        else data_sources[uid]["download_pre_args"]
+    )
+    download_post_args = (
+        ""
+        if "download_post_args" not in data_sources[uid]
+        else data_sources[uid]["download_post_args"]
+    )
+    download_command = (
+        "wget --continue "
```
Skylion007, line 168: Why call wget? Pythonhas plenty of ways to handle downloads itself like requests. No reason to add an dependency like this. (And no properly shell escape it).

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+    # print(download_command)
+    os.system(download_command)
+    assert os.path.exists(
+        os.path.join(data_dir, data_sources[uid]["package_name"])
+    ), "Download failed, no package found."
+
+    # unpack
+    package_name = data_sources[uid]["package_name"]
+    if package_name.endswith(".zip"):
+        os.system(
```
Skylion007, line 184: Python has builtin unzip support too: https://stackoverflow.com/a/3451150/2444240

habitat_sim/utils/datasets_download.py
Skylion007: Requests or url lib should get you what you need here iwth better error handling

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+    assert os.path.exists(
+        os.path.join(data_dir, data_sources[uid]["root"])
+    ), "Unpacking failed, no root directory."
+
+    # write version
+    with open(version_filepath, "w"):
+        print("Wrote version file: " + version_filepath)
+
+    # clean-up
+    os.system("rm " + data_dir + package_name)
```
Skylion007, line 206: ```os.remove``` instead

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--uid",
+        type=str,
+        help="Unique ID of the data to download.",
+    )
+    parser.add_argument(
```
Skylion007, line 223: Fuzzing: "clean and replace should conflict"

Skylion007 (changes requested):
A lot of error handling and functionality can be done by stdlib here. Will make your job a lot easier. Also rm should never be used in bare strings like this outside of BASH where we don't have a choice.

Skylion007:
Any reason we aren't putting this in the tools folder? Unless you want to integrate deeply in habitat_sim so scripts can call it programatically if an expected test asset is missing.

habitat_sim/utils/datasets_download.py
Skylion007: See the notebooks, we already have a very general way that gets the path of the current GitHook.

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+    )
+    if os.path.exists(data_sources[uid]["root"]):
+        resolved_existing = False
+        replace_existing = False
+        for file in os.listdir(data_sources[uid]["root"]):
+            if "DATA_VERSION" in file:
+                existing_version = file[13:]
+                if existing_version == data_sources[uid]["version"]:
+                    print(
+                        "Existing data source ("
```
Skylion007, line 114: f-string formatting would make this a lot cleaner.

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+            # found root directory with unspecified version
+            found_version_message = (
+                "("
+                + uid
+                + ") Found unknown data version. Replace with requested version ("
+                + data_sources[uid]["version"]
+                + ")?"
+            )
+            replace_existing = (
+                replace if replace else prompt_yes_no(found_version_message)
```
Skylion007, line 148: invalid answer doesn't error or return. I would make it loop until they pick yes / no.

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+        if replace_existing:
+            clean_data(uid)
+        else:
+            print("Not replacing data, aborting download.")
+            return
+
+    # download
+    download_pre_args = (
+        ""
+        if "download_pre_args" not in data_sources[uid]
```
Skylion007, line 159: use the dict.get(key, "") instead. It's much cleaner.

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,246 @@
...
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+import os
+
+# TODO: is there a better way to do this more generaly? Maybe pybind a cmake compiled DATA_DIR?
+dir_path = os.path.dirname(os.path.realpath(__file__))
```
Skylion007, line 11: You are doing a lot of os.path.joins. You may want to look into PathLib (now standard as of Python 3.6)

habitat_sim/utils/datasets_download.py
aclegg3: The prompt_yes_no function loops until a valid input is provided.

habitat_sim/utils/datasets_download.py
aclegg3: Added: `--data_path` default from git root.

aclegg3:
> Any reason we aren't putting this in the tools folder? Unless you want to integrate deeply in habitat_sim so scripts can call it programmatically if an expected test asset is missing.

Exactly, we can discuss whether or not we want this kind of automation behavior (e.g. when running tutorials).

habitat_sim/utils/datasets_download.py
```diff
@@ -179,12 +179,12 @@ def download_and_place(uid, replace=False):
     # download
     download_pre_args = (
         ""
-        if "download_pre_args" not in data_sources[uid]
+        if not data_sources[uid].get("download_pre_args")
```
Skylion007, line 163: What I meant was:
```
download_pre_args = data_sources[uid].get("download_pre_args", "")
```

Skylion007 (changes requested):

habitat_sim/utils/datasets_download.py
```diff
@@ -103,8 +103,8 @@ def prompt_yes_no(message):
 
 def clean_data(uid):
     r"""Deletes the "root" directory for the named data-source."""
-    if uid not in data_sources:
-        print("Data download failed, no datasource named " + uid)
+    if not data_sources.get(uid):
```
Skylion007, line 106: Wait, this works? The old method was much more readable, see the other comment for what I meant.

habitat_sim/utils/datasets_download.py
Skylion007: @aclegg3 Actually, if you want the class to be serializable to checkout Python dataclasses (there is a small backport we can add to requirements easily). This seems like a good case to use them here.

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,249 @@
...
+    )
+    # print(download_command)
+    os.system(download_command)
+    assert os.path.exists(
+        os.path.join(data_path, data_sources[uid]["package_name"])
+    ), "Download failed, no package found."
+
+    # unpack
+    package_name = data_sources[uid]["package_name"]
+    if package_name.endswith(".zip"):
```
Skylion007, line 180: Tar has an almost identical API in Python too. Might as well just add it in here too: https://docs.python.org/3/library/tarfile.html#tarfile.open

habitat_sim/utils/datasets_download.py
Skylion007:
```
import urllib.request
import shutil
...
# Download the file from `url` and save it locally under `file_name`:
with urllib.request.urlopen(url) as response, open(file_name, 'wb') as out_file:
    shutil.copyfileobj(response, out_file)
```

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,249 @@
...
+dir_path = repo.working_tree_dir
+# Root data directory. Optionaly overridden by input argument "--data_path".
+data_path = os.path.join(dir_path, "data")
+
+data_sources = {}
+data_groups = {}
+
+
+def initialize_test_data_sources(data_path):
+    global data_sources
```
Skylion007, line 24:
```
Declaring as global variables is unnecessary. Just do a .clear() and a dict.update() on the original references.
```

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,249 @@
...
+def clean_data(uid):
+    r"""Deletes the "root" directory for the named data-source."""
+    if not data_sources.get(uid):
+        print(f"Data clean failed, no datasource named {uid}")
+        return
+    root_path = data_sources[uid]["root"]
+    print(f"Cleaning datasource ({uid}) directory: {root_path}")
+    try:
+        shutil.rmtree(root_path)
+    except OSError as e:
```
Skylion007, line 113: Instead of printing error, you could use the traceback module to generate a nicer error message and traceback. Also weird to output errors to stdout, no?

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,249 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+import os
+import shutil
+import zipfile
+
+import git
+
+repo = git.Repo(".", search_parent_directories=True)
```
Skylion007, line 14: Meta Point: if we want this to live in habitat-sim/utils the git repo trick will break for a global/conda install, no? Also putting it in the global namespace liek this will prevent this code from being imported or run outside of a git repo even if that path is specified, which is less than ideal. 

Should all these print statements use the habitat_sim logger?

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,249 @@
...
+    else:
+        # TODO: support more compression types as necessary
+        print(f"Data unpack failed for {uid}. Unsupported filetype: {package_name}")
+        return
+    assert os.path.exists(
+        os.path.join(data_path, data_sources[uid]["root"])
+    ), "Unpacking failed, no root directory."
+
+    # write version
+    with open(version_filepath, "w"):
```
Skylion007, line 193: Pathlib has a touch command, which seems like what you want here btw.

habitat_sim/utils/datasets_download.py
erikwijmans: My 2 cents here is that `wget` is worth it.  The ability to resume partial downloads can be important when on flaky internet connections.

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,252 @@
...
+        #   "download_pre_args": (optional)(wget) commands preceding filename
+        #   "download_post_args": (optional)(wget) commands follow filename
+        #   "unpack_to": destination path for uncompression of the download package
+        #   "link": symlink to the data directory pointing to the active version directory
+        #   "version": data version tag
+        # }
+        "habitat_test_scenes": {
+            "source": "http://dl.fbaipublicfiles.com/habitat/habitat-test-scenes.zip",
+            "package_name": "habitat-test-scenes.zip",
+            "unpack_to": data_path + "../",
```
eundersander, line 44: Doesn't this break if the user overrides the `data_path`? Since that particular zip file hard-codes `data`.

I would be a fan of changing that zip file to not contain `data/scene_datasets/habitat-test-scenes`. These dataset zip files shouldn't be making assumptions about where they might live in the larger project. And then maybe this whole idea of separate `unpack_to` and `link` paths could go away (there'd just be `link`).

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,252 @@
...
+    print("=======================================================")
+    print(f"Dataset ({uid}) successfully downloaded.")
+    print(f"Source: {version_dir}.")
+    print(f"Symlink: {link_path}")
+    print("=======================================================")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
```
eundersander, line 207: Nit: would be cool to have an option to print the list of available data sources.

eundersander (commented):
I left some comments. Looks promising!

Skylion007:
Pull from master to fix the linting stuff.

README.md
```diff
@@ -314,7 +318,7 @@ Load a specific MP3D or Gibson house: `examples/example.py --scene path/to/mp3d/
 
 Additional arguments to `example.py` are provided to change the sensor configuration, print statistics of the semantic annotations in a scene, compute action-space shortest path trajectories, and set other useful functionality. Refer to the `example.py` and `demo_runner.py` source files for an overview.
 
-To reproduce the benchmark table from above run `examples/benchmark.py --scene /path/to/mp3d/17DRP5sb8fy/17DRP5sb8fy.glb`.
+To reproduce the benchmark table from above run `examples/benchmark.py --scene /path/to/mp3d_test/17DRP5sb8fy/17DRP5sb8fy.glb`.
```
erikwijmans, line 321:
```suggestion
To reproduce the benchmark table from above run `examples/benchmark.py --scene /path/to/mp3d_example/17DRP5sb8fy/17DRP5sb8fy.glb`.
```

Please use example instead of test for examples.  This is already somewhat confusing with the habitat-test-scenes naming, but will be even more confusing with MP3D as an actual test split exists.

habitat_sim/utils/datasets_download.py
Skylion007: No external dependency needed for this: @erikwijmans  https://gist.github.com/tobiasraabe/58adee67de619ce621464c1a6511d7d9

README.md
Skylion007: Also this script will try to download the test_scene if it's not where we tell it is. May want to make sure that logic is affected.

README.md
```diff
@@ -206,29 +206,33 @@ Read [build instructions and common build issues](BUILD_FROM_SOURCE.md).
 
 ## Testing
 
-1. Download the test scenes from this [link](http://dl.fbaipublicfiles.com/habitat/habitat-test-scenes.zip) and extract locally.
+1. Run our python data download utility to retrieve the test assets:
+   ```bash
+   python /path/to/habitat-sim/habitat_sim/utils/datasets_download.py --uid habitat_test_scenes --data_path /path/to/habitat-sim/data/
```
erikwijmans, line 211:
```suggestion
   python -m habitat_sim.utils.datasets_download --uid habitat_test_scenes --data_path /path/to/habitat-sim/data/
```

This will work with both a local install and a conda install

erikwijmans: Additionally, maybe use `/path/to/data` or something to signify that nothing needs to get in a `habitat-sim/data` folder?

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,310 @@
...
+        help="Unique ID of the data to download.",
+    )
+    arg_group1.add_argument(
+        "--list",
+        action="store_true",
+        help="List available datasource uid options and exit.",
+    )
+
+    parser.add_argument(
+        "--data_path",
```
erikwijmans, line 229:
```suggestion
        "--data-path",
```
We've been using `-` for spacing in the rest of the codebase

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,310 @@
...
+    def error(self, message):
+        sys.stderr.write("error: %s\n" % message)
+        self.print_help()
+        sys.exit(2)
+
+
+def main(args):
+    parser = AutoHelpParser()
+    arg_group1 = parser.add_mutually_exclusive_group()
+    arg_group1.add_argument(
```
erikwijmans, line 217: Worth considering making this `--uids` (add `nargs="*", default=None` to enable this taking multiple args, `args.uids` will then be a list if one or more is provided).  This would then let you do 
```
python -m habitat_sim.utils.datasets_download --uids habitat_example_objects locobot_merged --data_path path/to/habitat-sim/data/
```

README.md
erikwijmans: Looks like there is a bunch of `python /path/to/habitat-sim/habitat_sim/utils/datasets_download.py`, all can/should be `python -m habitat_sim.utils.datasets_download`

habitat_sim/utils/datasets_download.py
```diff
@@ -0,0 +1,310 @@
...
+data_groups = {}
+
+
+def initialize_test_data_sources(data_path):
+    global data_sources
+    global data_groups
+    r"""Initializes data_sources and data_groups dicts with a variable data_path path.
+    """
+    # dict keyed by uids with download specs for various individual test assets and datasets
+    data_sources = {
```
erikwijmans, line 25: Probably for a different PR as this will require discussion, but we should add the ability to have this work with mp3d and Gibson, but require the user to provide a download URL

README.md
```diff
@@ -206,29 +206,33 @@ Read [build instructions and common build issues](BUILD_FROM_SOURCE.md).
...
-   Use W/A/S/D keys to move forward/left/backward/right and arrow keys to control gaze direction (look up/down/left/right).
+   Use W/A/S/D keys to move forward/left/backward/right and arrow keys or mouse to control gaze direction (look up/down/left/right).
    Try to find the picture of a woman surrounded by a wreath.
    Have fun!
 
-1. **Physical interactions**: If you would like to try out habitat with dynamical objects, first download our pre-processed object data-set from this [link](http://dl.fbaipublicfiles.com/habitat/objects_v0.2.zip) and extract as `habitat-sim/data/objects/`.
+1. **Physical interactions**: If you would like to try out habitat with dynamical objects using the interactive viewer:
+   First setup the test object assets by running the data download utility:
+   ```bash
+   python -m habitat_sim.utils.datasets_download --uids habitat_example_objects --data-path /path/to/data/
```
eundersander, line 227: I guess users must download to `data/objects` because that's the path you specify later for `object-dir`.

eundersander: Overall, it feels a little dissatisfying to always say `/path/to/data/` and leave it up to the user. A n00b user is usually not in a position to make a good decision on where to put data (inside the hab-sim root folder? inside the hab-lab root folder? Somewhere outside both of these folders?)

Some of your changes in this PR are changing clear recommended data locations (`path/to/habitat-sim/data/objects/`) to the more ambiguous `/path/to/data/`.

eundersander (approved):
Left some minor comments. LGTM.

README.md
aclegg3: Ideally, users with a standard install provide either no path (defaults to `github_root/data/`) or directly provide `/path/to/habitat-sim/data/`. The script takes care of placing the symlinks in `data/objects/` and in both of these cases tests and examples should proceed correctly. However, users with a more complex setup will need to install the data and then manually point the simulator to their data directory or create necessary symlinks.

PR #1203, opened 2021-04-29
[Sensor] Use texture instead of render buffer for semantic info rendering

bigbike:
## Motivation and Context
As titled.
The benefit:
The texture is much easier to be visualized.

Should have no performance lose.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -299,6 +299,8 @@ struct RenderTarget::Impl {
         "RenderTarget::Impl::readFrameObjectIdGPU(): this render target "
         "was not created with objectId render texture enabled.", );
 
+    // We replaced the render buffer by the render texture in #1203, and
+    // expected NO performance loss. Let us know if it is not the case.
     if (objecIdBufferCugl_ == nullptr)
       checkCudaErrors(cudaGraphicsGLRegisterImage(
           &objecIdBufferCugl_, objectIdTexture_.id(), GL_TEXTURE_2D,
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -226,8 +226,12 @@ struct RenderTarget::Impl {
         flags_ & Flag::ObjectIdTexture,
         "RenderTarget::Impl::readFrameObjectId(): this render target "
         "was not created with objectId render texture enabled.", );
-
+#ifdef MAGNUM_TARGET_WEBGL
+    framebuffer_.mapForRead(ObjectIdTextureColorAttachment)
+        .read(framebuffer_.viewport(), view);
+#else
     objectIdTexture_.image(0, view);
+#endif
   }
 
   Mn::Vector2i framebufferSize() const {
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -36,11 +36,11 @@ namespace Mn = Magnum;
 namespace esp {
 namespace gfx {
 
-const Mn::GL::Framebuffer::ColorAttachment RgbaBuffer =
+const Mn::GL::Framebuffer::ColorAttachment RgbaBufferAttachment =
     Mn::GL::Framebuffer::ColorAttachment{0};
 const Mn::GL::Framebuffer::ColorAttachment ObjectIdTextureColorAttachment =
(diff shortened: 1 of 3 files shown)


src/esp/gfx/RenderTarget.cpp
```diff
@@ -215,18 +223,23 @@ struct RenderTarget::Impl {
...
         "RenderTarget::Impl::readFrameObjectId(): this render target "
-        "was not created with objectId render buffer enabled.", );
-
-    framebuffer_.mapForRead(ObjectIdBuffer).read(framebuffer_.viewport(), view);
+        "was not created with objectId render texture enabled.", );
+#ifdef MAGNUM_TARGET_WEBGL
+    framebuffer_.mapForRead(ObjectIdTextureColorAttachment)
+        .read(framebuffer_.viewport(), view);
+#else
+    objectIdTexture_.image(0, view);
```
bigbike, line 233: @mosra: I used the `image(...)` here. Is it necessary?

erikwijmans (commented):
I'm in principle fine with this.  I don't think there is any appreciable perf concerns but I will let @mosra comment as he knows this better than I.

bigbike:
@erikwijmans : thanks. Pay attention to the cuda function. See if all the necessary changes are made.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -37,7 +38,7 @@ namespace gfx {
 
 const Mn::GL::Framebuffer::ColorAttachment RgbaBuffer =
     Mn::GL::Framebuffer::ColorAttachment{0};
-const Mn::GL::Framebuffer::ColorAttachment ObjectIdBuffer =
+const Mn::GL::Framebuffer::ColorAttachment ObjectIdTextureColorAttachment =
     Mn::GL::Framebuffer::ColorAttachment{1};
 const Mn::GL::Framebuffer::ColorAttachment UnprojectedDepthBuffer =
```
mosra, line 43: Why rename just one and not all? `ColorAttachment` seems like a clearer name, regardless of whether the underlying object is a renderbuffer or a texture.

src/esp/gfx/RenderTarget.h
```diff
@@ -38,11 +38,11 @@ class RenderTarget {
...
-     * create objectId buffer
+     * create objectId texture
      * No need to set it for color sensor, depth sensor etc. as it makes the
      * rendering slower
      */
-    ObjectIdBuffer = 1 << 1,
+    ObjectIdTexture = 1 << 1,
     /**
      * @brief use depth texture, it must be set for the depth sensor.
      * No need to set it for color sensor, objectId sensor etc. as it makes the
```
mosra, line 48: Same here, rename them all to `*Attachment` to not need to differentiate between renderbuffers and textures?

src/esp/gfx/RenderTarget.cpp
mosra: Why not have the same code path for both, like it was before? I don't think there should be any practical difference between the two, so in that case I'd say less code is better code :D

mosra (commented):
> Should have no performance lose.

I know that's what I suggested based on my experience with native rendering. But just to be sure, did you (did anybody) verify this has no negative effect on the CUDA interop and WebGL performance? These are two cases where I'm not 100% certain.

src/esp/gfx/RenderTarget.cpp
```diff
@@ -83,34 +88,37 @@ struct RenderTarget::Impl {
...
-    if (flags_ & Flag::RgbaBuffer) {
-      framebuffer_.mapForDraw({{0, RgbaBuffer}});
+    if (flags_ & Flag::RgbaAttachment) {
+      framebuffer_.mapForDraw(
+          {{Mn::Shaders::Generic3D::ColorOutput, RgbaBufferAttachment}});
     }
-
-    if (flags_ & Flag::ObjectIdBuffer) {
-      framebuffer_.mapForDraw({{1, ObjectIdBuffer}});
+    if (flags_ & Flag::ObjectIdAttachment) {
```
bigbike, line 109: @mosra:
There is a big bug hidden here that even you did not capture it.
When I need to visualize the semantic info, both `flags_ & Flag::RgbaAttachment` and `flags_ & Flag::ObjectIdAttachment` will be true.
That means `mapForDraw` will be called twice in two different if clauses. So the first attempt is overidden by the second. Sigh...

Skylion007 (approved):

PR #1206, opened 2021-04-30
[AO] - ContactTest for ArticulatedObject

aclegg3:
## Motivation and Context

Extends `contactTest` function to support `ArticulatedObjects` by id. Individually checks each link and base collision shape against the world and screens self collisions if necessary returning a boolean indicator of whether or not any multibody collision shape is in contact with any other collision object in the scene.

Also modified the `contactTest` setup to use configured collision group and mask filters by default. A side-effect of this change is that `STATIC` objects (including articulated fixed base) will not report contact with other `STATICs` (such as the stage). An option override is provided by setting `static_as_stage` to false which then treats the queried `STATIC` as a dynamic object.

## How Has This Been Tested

Local testing of AO. Modified some tests for rigids.

## Types of changes

### DIFF
--- a/src/esp/physics/bullet/BulletArticulatedObject.cpp
+++ b/src/esp/physics/bullet/BulletArticulatedObject.cpp
@@ -714,8 +714,77 @@ void BulletArticulatedObject::updateKinematicState() {
   }
 }
 
+/**
+ * @brief Specific callback function for ArticulatedObject::contactTest to
+ * screen self-collisions.
+ */
+struct AOSimulationContactResultCallback
+    : public SimulationContactResultCallback {
+  btMultiBody* mb_ = nullptr;
+  btRigidBody* fixedBaseColObj_ = nullptr;
+
+  /**
+   * @brief Constructor taking the AO's btMultiBody as input to screen
+   * self-collisions.
+   */
+  AOSimulationContactResultCallback(btMultiBody* mb,
+                                    btRigidBody* fixedBaseColObj)
+      : mb_(mb), fixedBaseColObj_(fixedBaseColObj) {
+    bCollision = false;
+  }
+
+  /**
+   * @brief Called when a contact is detected.
+   *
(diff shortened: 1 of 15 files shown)


src/esp/physics/bullet/BulletArticulatedObject.cpp
```diff
@@ -714,5 +714,103 @@ void BulletArticulatedObject::updateKinematicState() {
...
+      const btMultiBodyLinkCollider* mblc1 =
+          btMultiBodyLinkCollider::upcast(colObj0Wrap->getCollisionObject());
+      const btMultiBodyLinkCollider* mblc2 =
+          btMultiBodyLinkCollider::upcast(colObj1Wrap->getCollisionObject());
+      if (mblc1 && mblc2) {
+        if (mblc1->m_multiBody == mblc2->m_multiBody) {
+          // link self-collision
+          return 0;
+        }
+      } else if (colObj0Wrap->getCollisionObject() == fixedBaseColObj_) {
```
eundersander, line 763: Shouldn't "multibody link/base versus fixedBaseColObj" be already getting filtered out in the broadphase via use of CollisionGroups?

I'm not sure if getCollisionWorld()->contactTest uses the same CollisionGroup logic for collision filtering that stepSimulation uses. I thought it did.

eundersander (approved):
I added a comment. LGTM.

src/esp/physics/bullet/BulletArticulatedObject.cpp
aclegg3: Good point, I dug into this a bit. We can configure the mask and group for the filter on a per-object basis. I added this change. However, we also need to separately handle self-collision between links. 

This also means that STATIC objects and fixed base will not report contact with other STATICs such as the stage by default. I added an option to override that behavior for common use cases. For example placing a fixed base ArticulatedObject in the stage via rejection sampling.

PR #1207, opened 2021-04-30
[Sensor] Visualize the semantic observations

bigbike:
## Motivation and Context
As titled.
This PR adds a brand new functionality to visualize the observations from the semantic sensor.

The results are as follows:
rgb-1
semantic-1
rgb-2
semantic-2

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/TextureVisualizerShader.cpp
+++ b/src/esp/gfx/TextureVisualizerShader.cpp
@@ -92,10 +92,14 @@ TextureVisualizerShader::TextureVisualizerShader(Flags flags) : flags_(flags) {
   const Mn::Vector2i size{int(map.size()), 1};
   colorMapTexture_.setMinificationFilter(Mn::GL::SamplerFilter::Linear)
       .setMagnificationFilter(Mn::GL::SamplerFilter::Linear)
-      .setWrapping(Mn::GL::SamplerWrapping::Repeat)
       .setStorage(1, Mn::GL::TextureFormat::SRGB8Alpha8, size)
       .setSubImage(0, {},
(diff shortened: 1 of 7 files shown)


src/esp/gfx/TextureVisualizerShader.cpp
```diff
@@ -99,9 +101,10 @@ TextureVisualizerShader::TextureVisualizerShader(Flags flags) : flags_(flags) {
   // set default offset, scale based on flags
   if (flags_ & Flag::DepthTexture) {
     setColorMapTransformation(1.0f / 512.0f, 1.0f / 1000.0f);
+  } else if (flags & Flag::ObjectIdTexture) {
+    setColorMapTransformation(1.0f / 512.0f,
```
bigbike, line 105: @mosra:
I was thinking the wrapping of the colorMapTexture found [here](https://github.com/facebookresearch/habitat-sim/pull/1207/files#diff-1f42624b53785fbfad33558bea811592548000ef11dedc0533d69086ed35db23R95).
I believe for depth rendering it should be set to `ClampToEdge` while for semantic rendering it should be set to `Repeat`. How do you think?

bigbike: @mosra:
Another question: we can clear the depth, clear the color, but how can we clear the objectId (another color attachment)

mosra: use the `clearDepth()` / `clearColor()` APIs on the framebuffer, instead of `clear()`, those allow you to clear also non-integer attachments

mosra (approved):

Skylion007 (approved):

PR #1009, opened 2021-01-06
Change sensor enum from ALL_CAPS to Pascal style

vauduong:
## Motivation and Context
Changing the style of the enumeration in Sensor.h

## How Has This Been Tested

Built and compiled
## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1210, opened 2021-05-03
Support specifying position model/frame of reference for lighting instances

jturner65:
## Motivation and Context
This PR introduces JSON support, and appropriate testing and documentation, for setting the desired position model/frame of reference for a lighting instance.  With this setting, the user can specify whether they wish for a particular light info/instance to be placed in the scene relative to the global origin or the camera.  Object-relative positioning is also fully supported in the configuration, but only as a stub currently in the underlying lighting code.

The enum describing the position model value has also been refactored to follow our naming conventions, from  ALL_CAPS to pascal case, in the c++ and python code.
 

## How Has This Been Tested
All existing and added c++ and python test cases pass

## Types of changes

## Checklist

### DIFF
--- a/docs/pages/attributesJSON.rst
+++ b/docs/pages/attributesJSON.rst
@@ -230,3 +230,6 @@ The Light Setup attributes JSON should contain a single cell named "lights" that
 "type"
 	- string
 	- The type of the light.  "point" and "directional" are currently supported.
+"position_model"
+  - string
+  - They origin to use to place the light. "global", meaning stage's origin, and "camera" meaning place relative to a (potentially moving) camera, are currently supported.
--- a/docs/pages/attributesJSON.rst
+++ b/docs/pages/attributesJSON.rst
@@ -232,4 +232,4 @@ The Light Setup attributes JSON should contain a single cell named "lights" that
 	- The type of the light.  "point" and "directional" are currently supported.
 "position_model"
   - string
-  - They origin to use to place the light. "global", meaning stage's origin, and "camera" meaning place relative to a (potentially moving) camera, are currently supported.
+  - They frame to use to place the light. "global", meaning stage's origin, and "camera", meaning place relative to a (potentially moving) camera, are currently supported.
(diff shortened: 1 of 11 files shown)


aclegg3 (approved):

src/esp/gfx/LightSetup.cpp
```diff
@@ -28,11 +28,11 @@ Magnum::Vector4 getLightPositionRelativeToCamera(
                         "w == 1 for a point light");
 
   switch (light.model) {
-    case LightPositionModel::OBJECT:
+    case LightPositionModel::Object:
```
Skylion007, line 31: Why the change in style?

jturner65: This was a refactor in other enums through Habitat-Sim we decided [upon a few months ago](https://github.com/facebookresearch/habitat-sim/pull/1009).  Guess this enum got missed.

docs/pages/attributesJSON.rst
```diff
@@ -230,3 +230,6 @@ The Light Setup attributes JSON should contain a single cell named "lights" that
 "type"
 	- string
 	- The type of the light.  "point" and "directional" are currently supported.
+"position_model"
+  - string
+  - They frame to use to place the light. "global", meaning stage's origin, and "camera", meaning place relative to a (potentially moving) camera, are currently supported.
```
bigbike, line 235: Oh, the "object" mode cannot be supported at the moment.

bigbike (commented):
This is great! Thank you very much!

docs/pages/attributesJSON.rst
jturner65: yes, this would require more configuration specifications at the scene instance level, to specify which object instance the light should "orbit"

bigbike: Not "orbit". Just stay fixed relative to the object.
In that case, the "light" should be an object in the scene graph, and be attached to a child node of that object.

jturner65: Fixed orbit :).

PR #1211, opened 2021-05-03
[CMake][Bullet] Let Sim be built without Bullet

Skylion007:
## Motivation and Context

* Habitat-Sim could not be built without Bullet if Bullet wasn't installed due to a CMake issue.
Without having Bullet installed on your system, if you tried to build Habitat-Sim without Bullet, you would get this error:
```
CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Bullet (missing: Bullet_Dynamics_LIBRARY
  Bullet_Collision_LIBRARY Bullet_LinearMath_LIBRARY Bullet_SoftBody_LIBRARY
  Bullet_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  deps/magnum-integration/modules/FindBullet.cmake:193 (find_package_handle_standard_args)
  deps/magnum-integration/src/Magnum/BulletIntegration/CMakeLists.txt:31 (find_packag
```
This solves that issue.

## How Has This Been Tested
* I removed the system wide Bullet from CircleCI since we don't need it anymore. This way this issue should repro on CI. I tested it locally.

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -462,11 +462,6 @@ jobs:
             export PATH=$HOME/miniconda/bin:$PATH
             . activate habitat;
             ccache --show-stats
-      - save_cache:
-          key: ccache-{{ arch }}-{{ .Branch }}
-          background: true
-          paths:
-            - /home/circleci/.ccache
       - run: *download_test_data
       - run:
           name: Build Javascript bindings
@@ -537,6 +532,11 @@ jobs:
               nvm use v11.9.0
               # Generate JS CodeConv
               npm run test_with_coverage
+      - save_cache:
+          key: ccache-{{ arch }}-{{ .Branch }}
+          background: true
+          paths:
+            - /home/circleci/.ccache
       - run:
           name: Upload test coverage
           command: |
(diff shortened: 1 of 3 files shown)


Skylion007:
I also went and optimized the JS build to be twice as fast by increasing the number of parallel jobs being used. By switching it over to use Ninja instead of Make, the memory issues no longer occur allowing higher levels of parallelism.

jturner65 (approved):
LGTM

mosra:
Dang, again too late for a review, sorry about that :sweat_smile: 

> By switching it over to use Ninja instead of Make, the memory issues no longer occur allowing higher levels of parallelism.

I don't think that's related to Make vs Ninja. If it works now, it doesn't mean it'll work always -- according to my investigation, CircleCI gives you anywhere between 8 and 32 GB of RAM (I suppose depending on the load, so when things build during the US daytime, you get less, while during the European daytime you get more (haha)).

> This solves that issue.

Is that the `set(WITH_BULLET OFF)`? I fail to see how that is the fix. This option is not enabled by default by Magnum itself, which means *something else* is setting this and so you're not fixing the root cause, only a consequence.

Skylion007:
Yeah, something else is enabling it then. Not sure what though, this was a quick and dirty fix.

PR #1213, opened 2021-05-04
[Clang-Tidy] enable modernize-pass-by-value

Skylion007:
## Motivation and Context

* I was looking for easy ways to improve performance with clang-tidy, and this was one check we hadn't enabled yet. When I first tried it last year, it only caught some copies of trivially copyable types (vec3) and therefore wasn't terribly useful. Now though with the new attributemanagers and increased use of inheritance throughout the codebase it is getting rid of a lot of potentially unecessary copies (one of even a std::vector).
* This has a few advantages. One it makes the function definition clear about whether an argument will be copied to a class variable or simply stored in it (which a const reference does not give you).
* If it's the later, it prevents the copying the class variable from happening at all. It only does this for trivially movable types of course. If it can't be moved, std::move will copy it. This also prevents unnecessary overloads of the functions to prevent copies for rvalue vs lvalue types.
* This post gives a great explanation about why this is best modern C++ practice: https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference
* The other check I would like to enable at some point is https://clang.llvm.org/extra/clang-tidy/checks/modernize-loop-convert.html as that can reduce the number of checks in common statically allocated for loops and therefore help make loops more vectorizable. That has more caveats than this one though.

* Mainly a performance + code quality one

## How Has This Been Tested
* Locally. Everything compiled. Tests Passed.

## Types of changes

## Checklist


jturner65 (commented):
I'm undecided about this change, but it seems to my neophyte eye :)  that most of the time we are passing lvalues, so this change is actually impacting performance negatively, somewhat (bind + copy vs copy + move).  

As always, I could be wrong though :).

src/esp/assets/RenderAssetInstanceCreationInfo.h
```diff
@@ -29,10 +29,10 @@ struct RenderAssetInstanceCreationInfo {
   RenderAssetInstanceCreationInfo() = default;
 
   RenderAssetInstanceCreationInfo(
-      const std::string& _filepath,
-      const Corrade::Containers::Optional<Magnum::Vector3>& _scale,
+      std::string _filepath,
+      Corrade::Containers::Optional<Magnum::Vector3> _scale,
```
mosra, line 33: Here I'm afraid of another ["analysis paralysis"](https://en.wikipedia.org/wiki/Analysis_paralysis) problem, where users would need to think how big each trivially-copyable type is and then decide whether to pass by value or by `const&`, and then clang tidy having a slightly different opinion in many cases, fighting with the user.

Which is why I instead decided for consistency (at the cost of *some* perf disadvantage in certain cases on certain platforms), and with all trivially copyable types in Magnum I'm:

- passing all scalar types by value (`float`, but also `Deg`)
- passing `[Static]ArrayView`, `[Mutable]StringView` and `Reference` by value, as those are just one or two pointers
- passing `StridedArrayView` by `const&`, as it's quite big
- passing all math types by `const&`, even though e.g. `Vector2` is smaller than an `ArrayView`

This way my brain is freed from deciding this again for every function signature.

PR #1218, opened 2021-05-05
[AO] - convex compound collision shapes

aclegg3:
## Motivation and Context

Primary changes refactors the convex collision shape generation pipeline to separate joined and compound variants and support compounds for ArticulatedObjects. 

Secondary is a change to default behavior for fixed base ArticulatedObjects. We now only create proxy rigid objects for static links when the URDF explicitly overrides the collision shape filter group with `NonCollidable (256)`. This change prioritizes stability since neighboring links in the kinematic chain often have overlapping collision shapes (see the fridge.urdf asset). Using the rigid object proxy in these cases requires the dynamic links to mask contacts with STATIC objects such as the stage with a manual group override in URDF or they will become unstable in simulation with fixed base. As such, fixed base rigid proxy should be manually configured for specific cases in order to maximize performance for STATIC furniture objects. An example asset `fridge_static.urdf` is provided to demonstrate this option. While cheaper to simulate when the base is touching the stage, the doors will not register collisions with the walls or other STATIC furniture, making it useful only in targeted situations.

Tertiary Changes:
- viewer adds a fixed fridge.urdf facing the agent with '5'
- temporarily adjusted a physics test in which an object just tunneling through the floor. TODO: this test should be modified in master to test functionality rather than resulting stability.

## How Has This Been Tested

Local testing and verification with existing demos in `URDF_robotics_tutorial.py`.

## Types of changes

### DIFF
--- a/data/test_assets/urdf/fridge/fridge_static.urdf
+++ b/data/test_assets/urdf/fridge/fridge_static.urdf
@@ -1,0 +1,134 @@
+<?xml version="1.0" ?>
+<robot name="fridge">
+  <link name="root">
+  </link>
+  <joint name="root_rotation" type="fixed">
+    <origin rpy="1.565 0 0" xyz="0 0 0"/>
+    <parent link="root"/>
+    <child link="body"/>
+  </joint>
+  <link name="body">
+    <inertial>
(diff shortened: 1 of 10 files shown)


eundersander (approved):
LGTM

PR #1226, opened 2021-05-10
[WIP] Articulated Objects Support

aclegg3:
## Motivation and Context

WIP PR to track remaining work on merging ArticulatedObject support into Habitat-sim master branch. 

Feature TODO:
- URDF support in SceneInstanceAttributes structure
- ManagedArticulatedObject wrapper API and refactor
- Visual primitive shapes from URDF
- Attaching sensors to articulated agents (blocked by #1151 )
- Query Link maximal velocity

## How Has This Been Tested

`examples/tutorials/URDF_robotics_tutorial.py` demonstrates and applies new feature set.

TODO: better testing of individual features.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #1228, opened 2021-05-10
[AO migration] ResourceManager

aclegg3:
## Motivation and Context

ArticulatedObject migration PR split from #1226.

This PR modifies existing asset loading functionality in ResourceManager allowing simple import and attachment of assets referenced from URDF including custom Phong material generation and caching. Also supports COLLADA (.dae) import with standard coordinate frame. 

## How Has This Been Tested

See the [ao-no-async branch](https://github.com/facebookresearch/habitat-sim/tree/ao-no-async) for example of AO import using this approach.

## Types of changes

## Checklist


src/esp/bindings/SceneBindings.cpp
```diff
@@ -40,7 +40,8 @@ void initSceneBindings(py::module& m) {
       .value("EMPTY", SceneNodeType::EMPTY)
       .value("SENSOR", SceneNodeType::SENSOR)
       .value("AGENT", SceneNodeType::AGENT)
-      .value("CAMERA", SceneNodeType::CAMERA);
+      .value("CAMERA", SceneNodeType::CAMERA)
+      .value("OBJECT", SceneNodeType::OBJECT);
```
jturner65, line 44: Perhaps this PR is not the time for this, but we need to refactor our enums to pascal capitalization from all CAPs, for consistency.  Would be nice to have this done before 2.0 release.

jturner65 (approved):
LGTM

src/esp/assets/ResourceManager.cpp
```diff
@@ -1985,6 +1989,112 @@ void ResourceManager::removePrimitiveMesh(int primitiveID) {
...
+
+    std::vector<StaticDrawableInfo> staticDrawableInfo;
+    addComponent(meshMetaData, node, DEFAULT_LIGHTING_KEY, drawables,
+                 meshMetaData.root, visNodeCache, false, staticDrawableInfo);
+    // compute the full BB hierarchy for the new tree.
+    node.computeCumulativeBB();
+  } else {
+    return false;
+  }
+  return true;
```
Skylion007, line 2095: Isn't this statement unreachable?

Skylion007: Ah my bad. Wouldn't the better style here be to negate the condition and do return False at the top so that all the corresponding logic would be grouped together? Also would remove the need for the else statement.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1985,6 +1989,112 @@ void ResourceManager::removePrimitiveMesh(int primitiveID) {
...
+    // first register the copied metaData
+    resourceDict_.emplace(modifiedAssetName,
+                          LoadedAssetData(resourceDict_.at(filename)));
+  }
+
+  // create/set a new PhongMaterialData for the asset
+  auto& meshMetaData = resourceDict_.at(modifiedAssetName).meshMetaData;
+
+  std::ostringstream matHandleStream;
+  matHandleStream << "phong_amb_" << matColor.toSrgbAlphaInt() << "_spec_"
```
eundersander, line 2043: It would be really nice if this properties-to-string mapping went both ways. This would address my long-standing concerns about storing scene-specific info in ResourceManager.

As-is, a renderer in another process (such as a batched renderer) won't know what to do when it encounters a material called "my_material_phong_amb_01234567_spec_87654321" referenced by a SceneGraph/Drawable.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1985,6 +1989,112 @@ void ResourceManager::removePrimitiveMesh(int primitiveID) {
...
+  return meshSuccess;
+}
+
+std::string ResourceManager::setupMaterialModifiedAsset(
+    const std::string& filename,
+    const Magnum::Color4& matColor,
+    const Magnum::Color3& specularColor) {
+  assert(resourceDict_.count(filename) > 0);
+
+  std::string modifiedAssetName = filename + "_MatMod";
```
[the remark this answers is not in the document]
aclegg3: Another good point. 

Currently, this is setup to support only the URDF case where every time we load an asset we know the "override" material. We adjust the MeshMetaData material keys and then immediately create a Drawable. The Drawable never references back to the resourceDict_ which only stores metadata, so it doesn't matter if the entries are changed after that point to another material override. 

We also key the resourceDict_ by contrived keys in other cases (E.g. primitives, convex decomps, voxel meshes). The filepath keys should end with a file ending (e.g. .glb), so "my_file.glb_MatMod" should not be possible to load from disk.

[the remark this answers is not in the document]
Skylion007: @aclegg3 Can we just store a filesystem invalid seperator? Like ?, *, and "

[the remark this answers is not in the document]
aclegg3: Sure, easy enough. @jturner65 has been using this method in other places also I think.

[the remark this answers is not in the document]
jturner65: Yep. See [this link](https://github.com/facebookresearch/habitat-sim/blob/9bd03ecefde715de505a193ab454c1fb380a477a/src/esp/core/managedContainers/ManagedContainerBase.cpp#L135)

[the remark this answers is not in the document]
eundersander: @aclegg3 and I discussed offline and we had an idea to embed material-override info in our AssetInfo struct and conceptually handle this override at asset-load time, instead of later. This addresses/sidesteps my long-standing concern about modifying assets after loading them (and how to propagate those changes).

src/esp/assets/ResourceManager.cpp
```diff
@@ -1984,6 +1988,119 @@ void ResourceManager::removePrimitiveMesh(int primitiveID) {
   primitive_meshes_.erase(primitiveID);
 }
 
+bool ResourceManager::loadAsset(
```
eundersander, line 1991: This function is very similar to the existing loadAndCreateRenderAssetInstance. Can you consolidate so we don't have code duplication?

src/esp/assets/ResourceManager.cpp
```diff
@@ -1984,6 +1988,119 @@ void ResourceManager::removePrimitiveMesh(int primitiveID) {
...
+    const esp::assets::PhongMaterialColor* materialOverride,
+    scene::SceneNode* parent,
+    DrawableGroup* drawables,
+    std::vector<scene::SceneNode*>* visNodeCache) {
+  std::string finalAssetName = assetName;
+  bool meshSuccess = true;
+  if (resourceDict_.count(assetName) == 0) {
+    esp::assets::AssetInfo meshinfo{AssetType::UNKNOWN, assetName};
+    meshinfo.requiresLighting = requiresLighting;
+    meshSuccess = loadRenderAssetGeneral(meshinfo);
```
eundersander, line 2003: Is there a good reason to use the more narrow loadRenderAssetGeneral here versus the more broad loadRenderAsset?

eundersander (commented):
I left some comments.

src/esp/io/JsonEspTypes.h
```diff
@@ -71,6 +90,10 @@ inline JsonGenericValue toJsonValue(const esp::assets::AssetInfo& x,
   addMember(obj, "virtualUnitToMeters", x.virtualUnitToMeters, allocator);
   addMember(obj, "requiresLighting", x.requiresLighting, allocator);
   addMember(obj, "splitInstanceMesh", x.splitInstanceMesh, allocator);
+  if (x.overridePhongMaterial != Cr::Containers::NullOpt) {
```
eundersander, line 93: You can directly serialize the Corrade::Optional! Thanks to some template magic in src/esp/io/JsonMagnumTypes.h.

`addMember(obj, "overridePhongMaterial", x.overridePhongMaterial, allocator);` should just work, and similar for readMember.

src/tests/GfxReplayTest.cpp
```diff
@@ -108,11 +125,23 @@ TEST(GfxReplayTest, recorder) {
   ASSERT(keyframes[0].stateUpdates[0].first == instanceKey);
 
   // verify frame #1 has our translation and semantic Id
-  ASSERT(keyframes[1].stateUpdates.size() == 1);
+  ASSERT(keyframes[1].stateUpdates.size() == 2);
```
eundersander, line 128: Can you add a comment explaining why we expect 2 state updates on this keyframe? The reason is because (1) we moved `node`, and (2) we created `node2`, which gets a stateUpdate to capture its starting transform after creation.

eundersander (approved):
Updates look good! See minor comments.

PR #1229, opened 2021-05-11
[AO migration] async rendering

aclegg3:
## Motivation and Context

This PR introduces async rendering in a background thread allowing overlap of CPU physics simulation and GPU rendering operations resulting in a significant performance improvement. For training scenarios, this does result in delayed observation (the visual observations passed to the model will be one step behind the physical state) which is, however, analogous with sensor latency on physical systems.

Note: this feature is not enabled by default due to instability on some systems and potential for memory leaks if not properly used. See `examples/tutorials/async_rendering.py` for details.

Thanks @erikwijmans for building the vast majority of this feature.

## How Has This Been Tested

Adds a python example with CI testing. 

## Types of changes

## Checklist


src/deps/atomic_wait/atomic_wait.cpp
```diff
@@ -0,0 +1,31 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+*/
+
+#include <atomic_wait.h>
+
+contended_t contention[256];
```
eundersander, line 27: How about some comments here, and let's avoid unnamed magic numbers like 256 and 255.

src/esp/gfx/Renderer.cpp
```diff
@@ -39,17 +50,247 @@ namespace Mn = Magnum;
 namespace esp {
 namespace gfx {
 
+#if !defined(CORRADE_TARGET_EMSCRIPTEN)
+
+struct BackgroundRenderThread {
```
eundersander, line 55: This is a big struct with lots of code so let's move it to its own file.

src/esp/gfx/Renderer.cpp
```diff
@@ -39,17 +50,247 @@ namespace Mn = Magnum;
...
+  void run() {
+    context_->makeCurrentPlatform();
+    threadContext_ = new Mn::Platform::GLContext{Mn::NoCreate};
+    if (!threadContext_->tryCreate())
+      Mn::Fatal{} << "BackgroundRenderThread: Failed to create OpenGL context";
+
+    Mn::GL::Context::makeCurrent(threadContext_);
+    threadOwnsContext_ = true;
+
+    Mn::GL::Renderer::enable(Mn::GL::Renderer::Feature::DepthTest);
```
eundersander, line 214: Is this bit of render-state logic copy-pasted from `Renderer::Impl` constructor? I'm worried about maintaining this. Can we avoid the code duplication?

src/esp/gfx/WindowlessContext.cpp
```diff
@@ -60,13 +60,30 @@ struct WindowlessContext::Impl {
...
 
-  void makeCurrent() { windowlessGLContext_.makeCurrent(); }
+  void makeCurrent() {
+    makeCurrentPlatform();
+    Mn::GL::Context::makeCurrent(&magnumGLContext_);
+  }
+
+  bool makeCurrentPlatform() {
+    if (!windowlessGLContext_.makeCurrent())
+      Mn::Fatal{} << "Failed to make platform current";
```
eundersander, line 76: As far as detecting/throwing fatal errors, we also have `ESP_CHECK` and `CORRADE_ASSERT`. This style here is new to me. Is it correct?

eundersander (commented):
I still consider it an open question whether we should merge this to `master`. It would be nice if the PR description listed the known caveats/issues with this approach. (@aclegg3 mentioned that it was unsafe to reset the scene or something like that).

Anyway I left some comments.

src/deps/atomic_wait/atomic_wait.cpp
erikwijmans: This is a reference implementation for cpp20's atomic wait, it isn't part of a library so that's why it's not a submodule (it isn't intended to be compiled by another project) and it is not versioned (as this is the only version).

eundersander:
Have we tested this on Mac?

erikwijmans:
@eundersander @aclegg3 @bigbike I refactored this and tried to make the API a little more user friendly, i.e. force people to set physics immediately after starting the async draw.

I also made initialization of the background render thread lazy, so that if people never make async render calls, it will never be initialized and everything will work exactly like before.

aclegg3:
@erikwijmans Thanks for the work on this. I think it can land if we add a short python smoke test / tutorial script which can be checked on CI by `test_examples.py` and show users how to use this feature.

aclegg3:
@erikwijmans Thanks for adding the example! Agreed with @eundersander that some docstrings for new functions and python bindings would be ideal, but I think this is ready.

eundersander:
@erikwijmans thanks for adding `async_rendering.py`. I tried running this on my Mac, Fedora, and Ubuntu machines just out of curiosity. It ran fine on Mac but crashed on both Fedora and Ubuntu. Both machines have Nvidia GPUs. Segfault in `glDrawElements` in the background renderer for `apartment_1.glb`. Any ideas?

I'm surprised this passed our CI (which is also Ubuntu, I think). How confident are we that this python actually ran as part of https://app.circleci.com/pipelines/github/facebookresearch/habitat-sim/6961/workflows/26578358-7348-4169-ad02-d840053f266c/jobs/34364 ? Is it possible the CI machine ran it with a single hardware thread or otherwise somehow avoided the crash I hit?

eundersander (approved):
LGTM

PR #1235, opened 2021-05-12
Disable some useless 3rd party code

mosra:
## Motivation and Context

While looking through the build log for #1233, I found some other things that don't have to be there.

- Don't enable any export code in Assimp. Magnum doesn't have an Assimp-based exporter (yet) and no mesh export functionality is used in Habitat either, so this is not needed.
- Don't enable tests, docs and examples for GLFW. We don't need any of that.

This removes about 120 targets from my local Ninja build (1228 before, 1105 after). Not sure about impact on compile times, since a majority of that is C code and Assimp isn't *that* crazy C++ either, but it could remove a few seconds I think.

## How Has This Been Tested

I'm not expecting the CI to fail.

## Types of changes


Skylion007 (approved):

Skylion007:
Thank you @mosra looked into disabling a while back when we were building the package from scratch on Colab, never merged it back end though. LGTM.

jturner65 (approved):

PR #1236, opened 2021-05-12
Fix some documentation warnings and issues

jturner65:
## Motivation and Context
The new documentation for the object wrappers was incorrectly marking certain functions as static.  This PR addresses this.

Also, there are nearly 4 thousand warnings indicative of missing docstrings, improper references, or other issues with the documentation when the docs are built.  This PR attempts to address some of these, although many thousand issues still remain.

## How Has This Been Tested
C++ and python tests pass local. Docs build locally (with thousands of warnings).

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/bindings.cpp
+++ b/src/esp/bindings/bindings.cpp
@@ -102,17 +102,17 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
   // TODO(msb) sim and sensor should not cross-depend
   esp::initEspBindings(m);
   esp::core::initCoreBindings(m);
-  esp::metadata::initAttributesBindings(m);
-  esp::metadata::initMetadataMediatorBindings(m);
-  esp::metadata::managers::initAttributesManagersBindings(m);
   esp::geo::initGeoBindings(m);
-  esp::physics::initPhysicsBindings(m);
-  esp::physics::initPhysicsObjectBindings(m);
-  esp::physics::initPhysicsWrapperManagerBindings(m);
   esp::scene::initSceneBindings(m);
   esp::gfx::initGfxBindings(m);
   esp::gfx::replay::initGfxReplayBindings(m);
   esp::sensor::initSensorBindings(m);
   esp::nav::initShortestPathBindings(m);
+  esp::metadata::initAttributesBindings(m);
+  esp::metadata::initMetadataMediatorBindings(m);
+  esp::metadata::managers::initAttributesManagersBindings(m);
+  esp::physics::initPhysicsBindings(m);
+  esp::physics::initPhysicsObjectBindings(m);
+  esp::physics::initPhysicsWrapperManagerBindings(m);
   esp::sim::initSimBindings(m);
 }
--- a/src/esp/bindings/bindings.cpp
+++ b/src/esp/bindings/bindings.cpp
@@ -111,6 +111,7 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
   esp::metadata::initAttributesBindings(m);
   esp::metadata::initMetadataMediatorBindings(m);
   esp::metadata::managers::initAttributesManagersBindings(m);
+  // These depend on SceneNode bindings
   esp::physics::initPhysicsBindings(m);
   esp::physics::initPhysicsObjectBindings(m);
   esp::physics::initPhysicsWrapperManagerBindings(m);
--- a/src/esp/core/managedContainers/ManagedContainer.h
+++ b/src/esp/core/managedContainers/ManagedContainer.h
@@ -51,8 +51,16 @@ class ManagedContainer : public ManagedContainerBase {
                 "ManagedContainer :: Managed object type must be derived from "
                 "AbstractManagedObject");
 
+  /**
+   * @brief Alias for shared pointer to the @ref
+   * esp::core::AbstractManagedObject this container manages.
+   */
   typedef std::shared_ptr<T> ManagedPtr;
 
+  /**
+   * @brief Constructor
+   * @param metadataType The name of the managed object type.
+   */
   explicit ManagedContainer(const std::string& metadataType)
       : ManagedContainerBase(metadataType) {}
 
@@ -221,7 +229,7 @@ class ManagedContainer : public ManagedContainerBase {
   }
 
   /**
-   * @brief  Remove the managed object referenced by the passed string handle.
+   * @brief Remove the managed object referenced by the passed string handle.
    * Will emplace managed object ID within deque of usable IDs and return the
    * managed object being removed.
    * @param objectHandle the string key of the managed object desired.
@@ -282,7 +290,7 @@ class ManagedContainer : public ManagedContainerBase {
 
   /**
    * @brief Get a reference to, or a copy of, the managed object identified by
-   * the @p managedObjectID, depending on @ref Access value.  This is the
+   * the @p managedObjectID, depending on @p Access value.  This is the
    * function that should be be accessed by the user for general object
    * consumption by ID.
    *
@@ -298,7 +306,7 @@ class ManagedContainer : public ManagedContainerBase {
 
   /**
    * @brief Get a reference to, or a copy of, the managed object identified by
-   * the @p objectHandle, depending on @ref Access value.  This is the function
+   * the @p objectHandle, depending on @p Access value.  This is the function
    * that should be be accessed by the user for general object consumption by
    * Handle.
    *
@@ -316,7 +324,7 @@ class ManagedContainer : public ManagedContainerBase {
 
   /**
    * @brief Get a reference to, or a copy of, the managed object identified by
-   * the @p managedObjectID, depending on @ref Access value, and casted to the
+   * the @p managedObjectID, depending on @p Access value, and casted to the
    * appropriate derived managed object class. This is the version that should
    * be accessed by the user for type-casted object consumption by ID.
    *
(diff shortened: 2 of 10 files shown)


src/esp/bindings/bindings.cpp
```diff
@@ -102,17 +102,17 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
...
   esp::gfx::initGfxBindings(m);
   esp::gfx::replay::initGfxReplayBindings(m);
   esp::sensor::initSensorBindings(m);
   esp::nav::initShortestPathBindings(m);
+  esp::metadata::initAttributesBindings(m);
+  esp::metadata::initMetadataMediatorBindings(m);
+  esp::metadata::managers::initAttributesManagersBindings(m);
+  esp::physics::initPhysicsBindings(m);
+  esp::physics::initPhysicsObjectBindings(m);
+  esp::physics::initPhysicsWrapperManagerBindings(m);
```
mosra, line 116: I'd suggest adding a comment explaining what depends on what so it doesn't accidentally get reordered again

src/esp/core/managedContainers/ManagedContainer.h
```diff
@@ -282,7 +290,7 @@ class ManagedContainer : public ManagedContainerBase {
 
   /**
    * @brief Get a reference to, or a copy of, the managed object identified by
-   * the @p managedObjectID, depending on @ref Access value.  This is the
+   * the @p managedObjectID, depending on @p Access value.  This is the
```
mosra, line 293: Doc and API usability tip: have the template parameter accessible also through a `constexpr static` member (or a `typedef`) that you can then `@ref` to, so people have an actual possibility to know what the `Access` is. (Yes, it would be better if we could just link to template parameter docs but Doxygen can do only so much.)

For example the Magnum [Vector type](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Vector.html#pub-types) has a `size` and `T` parameter, which are then exposed as `Size` and `Type`.

src/esp/gfx/RenderCamera.h
```diff
@@ -104,14 +104,30 @@ class RenderCamera : public MagnumCamera {
    */
   ~RenderCamera() override = default;
 
-  // Get the scene node being attached to.
+  /**
+   * @brief Get the scene node being attached to.
+   * @return the current scene node
```
mosra, line 109: FYI the `@return` doesn't have to be there, and I don't specify it unless it actually adds some new information (here it doesn't). Same for `@param`, you either have to specify all of them or you can specify none, if the meaning is obvious.

aclegg3 (approved):

src/esp/gfx/RenderCamera.h
```diff
@@ -104,14 +104,30 @@ class RenderCamera : public MagnumCamera {
...
+   * @return ref to current scene node
+   */
   scene::SceneNode& object() {
     return static_cast<scene::SceneNode&>(MagnumCamera::object());
   }
+
+  /**
+   * @brief Overloads to avoid confusion
+   * @return ref to current scene node
+   */
```
mosra, line 130: I'd just do `/** @overload */` for the two here. Less typing and Doxygen puts some prefabricated docs to these.

src/esp/gfx/RenderCamera.h
```diff
@@ -154,7 +180,7 @@ class RenderCamera : public MagnumCamera {
   /**
    * @brief Overload function to render the drawables
    * @param drawables, a drawable group containing all the drawables
-   * @param frustumCulling, whether do frustum culling or not, default: false
```
mosra, line 157: Not sure about the `,` after `drawables`, I suppose it shows as a stray comma in the output?

src/esp/physics/RigidObject.h
```diff
@@ -114,8 +115,8 @@ class RigidObject : public RigidBase {
                       initAttributes) override;
 
   /**
-   * @brief Finalize the creation of @ref RigidObject or @ref RigidScene that
-   * inherits from this class.
+   * @brief Finalize the creation of @ref esp::physics::RigidObject or @ref
+   * esp::physics::RigidScene that inherits from this class.
```
aclegg3, line 119:
```suggestion
   * @brief Finalize the creation of the RigidObject.
```

src/esp/physics/PhysicsObjectBase.h
```diff
@@ -81,17 +81,20 @@ class PhysicsObjectBase : public Magnum::SceneGraph::AbstractFeature3D {
...
   MotionType getMotionType() const { return objectMotionType_; }
 
   /**
-   * @brief Set the @ref MotionType of the object. If the object is @ref
-   * ObjectType::SCENE it can only be @ref MotionType::STATIC. If the object is
-   * @ref ObjectType::OBJECT is can also be set to @ref MotionType::KINEMATIC.
-   * Only if a dervied @ref PhysicsManager implementing dynamics is in use can
-   * the object be set to @ref MotionType::DYNAMIC.
+   * @brief Set the @ref MotionType of the object. If the construct is a @ref
+   * esp::physics::RigidScene, it can only be @ref
```
aclegg3, line 92:
```suggestion
   * esp::physics::RigidStage, it can only be @ref
```

aclegg3 (commented):
A couple minor things I noticed.

src/esp/physics/PhysicsManager.h
```diff
@@ -420,17 +422,17 @@ class PhysicsManager : public std::enable_shared_from_this<PhysicsManager> {
   // ============ Object Transformation functions =============
 
   /** @brief Set the 4x4 transformation matrix of an object kinematically.
-   * Calling this during simulation of a @ref MotionType::DYNAMIC object is not
-   * recommended.
+   * Calling this during simulation of a @ref esp::physics::MotionType::DYNAMIC
```
mosra, line 425: Strange that you have to fully-qualify these. Does the warning disappear after doing that, at least? If it does, then it's a shitty Doxygen bug (I get these sometimes also).

But generally you shouldn't need to fully-qualify -- if you're referring to something in `esp::physics` from the `esp::physics` namespace, it should be enough to use a relative name, like in C++ itself.

src/esp/physics/RigidBase.h
```diff
@@ -32,20 +32,26 @@ namespace physics {
...
+   */
   RigidBase(scene::SceneNode* rigidBodyNode,
             int objectId,
             const assets::ResourceManager& resMgr)
       : PhysicsObjectBase(rigidBodyNode, objectId, resMgr),
         visualNode_(&rigidBodyNode->createChild()) {}
 
   /**
-   * @brief Virtual destructor for a @ref RigidBase.
+   * @brief Virtual destructor for a @ref esp::physics::RigidBase.
```
mosra, line 48: ... for example here the error is because `RigidBase` itself is not documented (and fully qualifying *didn't* fix the error). 

Same is I suppose with `RigidObject` and `RigidStage` below -- no full qualification should be needed.

src/esp/physics/RigidBase.h
```diff
@@ -159,8 +166,8 @@ class RigidBase : public esp::physics::PhysicsObjectBase {
 
   /** @brief Virtual angular velocity setter for an object.
    *
-   * Does nothing for default @ref MotionType::KINEMATIC or @ref
-   * MotionType::STATIC objects.
+   * Does nothing for default @ref esp::physics::MotionType::KINEMATIC or @ref
+   * esp::physics::MotionType::::STATIC objects.
```
mosra, line 170:
```suggestion
   * esp::physics::MotionType::STATIC objects.
```

:eyes:

src/esp/physics/RigidBase.h
```diff
@@ -263,33 +270,33 @@ class RigidBase : public esp::physics::PhysicsObjectBase {
   /**
    * @brief Virtual linear velocity setter for an object.
    *
-   * Does nothing for default @ref MotionType::KINEMATIC or @ref
-   * MotionType::STATIC objects.
+   * Does nothing for default @ref esp::physics::MotionType::KINEMATIC or @ref
+   * esp::physics::MotionType::::STATIC objects.
```
mosra, line 274: :eyes:

src/esp/physics/RigidBase.h
```diff
@@ -238,22 +245,22 @@ class RigidBase : public esp::physics::PhysicsObjectBase {
...
    */
   virtual void setLinearDamping(CORRADE_UNUSED const double linDamping) {}
 
   /**
    * @brief Virtual linear velocity getter for an object.
    *
-   * Returns zero for default @ref MotionType::KINEMATIC or @ref
-   * MotionType::STATIC objects.
+   * Returns zero for default @ref esp::physics::MotionType::KINEMATIC or @ref
+   * esp::physics::MotionType::::STATIC objects.
```
mosra, line 263: :eyes:

Skylion007 (approved):

src/esp/physics/PhysicsManager.h
jturner65: > But generally you shouldn't need to fully-qualify -- if you're referring to something in esp::physics from the esp::physics namespace, it should be enough to use a relative name, like in C++ itself.

That's what I thought, but it did seem to address some of the warnings.

src/esp/physics/RigidBase.h
jturner65: Where should the doc be? Right about "class RigidBase...."?

mosra: Right above the class, yep.

PR #1240, opened 2021-05-13
[Flake8] Remove Object Inheritance

Skylion007:
## Motivation and Context

Flake8 simplify has been upgraded again and so this fixes most of its valid, but petty complaints.

## How Has This Been Tested
* Locally

## Types of changes

## Checklist


habitat_sim/utils/data/pose_extractor.py
```diff
@@ -10,7 +10,7 @@
 from habitat_sim.utils.common import quat_from_two_vectors
 
 
-class TopdownView(object):
+class TopdownView:  # noqa: SIM119
```
Skylion007, line 13: This complains that it really should be a data/attrs class, which is probably true, but I don't want to refactor it now.

aclegg3 (approved):

PR #1241, opened 2021-05-14
[CI] Improved mypy typing

Skylion007:
## Motivation and Context

* With typeguard removed, we should rely more on static type checking. This fixes a bunch of type issues to get mypy properly working with habitat_sim. Mainly we install mypy along with all the requirements so that it can compare the types of imported libraries like Numpy and Torch. We then install it and add it the python-linter checks. We ran it as a pre-commit before, but that would run in a virtualenv without dependencies of the project and therefore could miss things.

## How Has This Been Tested
* CI

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)sting functionality to change)

## Checklist


jturner65 (approved):
LGTM

PR #1242, opened 2021-05-15
Fix incorrectly configured LTO

Skylion007:
## Motivation and Context

IMPORTANT: This will significantly increase compile systems unless we figure out how to enable "ThinLTO" for Clang and will currently always increase compile time on GCC. For local development, you may want to use the newly added "--no-lto" flag to disable LTO. Alternatively, you can manually set the CMAKE variable CMAKE_INTERPROCEDURAL_OPTIMIZATION to enable or disable LTO.
* This enable LTO properly for all the submodules of Habitat_Sim. TLDR: LTO must be enabled for all compilation / sub-compilation units for it to be effect. We were relying on PyBind to do this for us, but it only did LTO on the final step (which is very important as it needs to optimize across cross the binding sub-units). However, this did not propogate the LTO down to Bullet or other submodules since they were already compiled without emiting the necessary bitcode. This changes it so the bitcode is always properly emitted. 
* LTO is slow, so I only enabled it for Release so far. Thoughts?
* I further optimized CircleCI.
* I added a config var in setup.py to specify the release. 
* We have always done LTO on the PyBind endpoints, but the additional information needed for better LTO optimization was rarely generated by the C++ frontends for the various submodules (including CMake). This PR attempts to use CMake features to force the enabling of LTO on submodules.
* Learned a lot about CMake and Clang/GCC in this front end.
Update:
* Some benchmarks with binary size:
* 25MB -> 20MB viewer build size (with symbols for Linux). (20% binary size reduction!)

## How Has This Been Tested
* Locally with on a CI with a lot of CI builds. I manually inspected the subprocess commands to ensure the proper '-flto' args were being passed around by CMake.

## Types of changes
* Performance improvement. I am seeing about a 30 - 50 FPS improvement on 1 GPU via the CI. I only really expect performance improvement in CPU bound tasks (Navigating on the navmesh a lot of doing complex physics). I haven't benchmarked physics yet, but I expect that might even benefit further since we are more CPU bound by bullet physics than graphics currently.
* Downside: This will increase compilation time slightly, hence why I defaulted to only doing full LTO in Release mode. We may want to make Anaconda start using this mode @mathfac or otherwise ensure LTO is enabled for our Anaconda builds so that we can get the added performance.
* 

## Checklist


Skylion007:
Weirdly the viewer and magnum-bindings link find using Thin_LTO, only the habitat_sim PyBind11 give it any trouble.

src/CMakeLists.txt
```diff
@@ -63,6 +66,17 @@ if(NOT CMAKE_BUILD_TYPE)
           FORCE
   )
 endif()
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  # Policy that enable INTERPROCEDURAL_OPTIMIZATIOn
+  set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)
+  cmake_policy(SET CMP0069 NEW)
```
mosra, line 72: As you pinged me on Slack, I'll enable this across all Magnum projects... but you still need this settings for other subprojects (assimp, bullet) anyway :)

src/CMakeLists.txt
```diff
@@ -63,8 +64,25 @@ if(NOT CMAKE_BUILD_TYPE)
...
+  # Policy that enables INTERPROCEDURAL_OPTIMIZATION
+  set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)
+  cmake_policy(SET CMP0069 NEW)
+  # Clang has some issues with thinLTO (segfaults on final linking)
+  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+endif()
 # avoid ld visibility warnings, should be done by CMAKE_CXX_VISIBILITY_PRESET
 # but need cmake_policy(SET CMP0063 NEW) also which seems to not work
+# TODO Figure out how to remove the visibility flags from Bullet
+#set(CMAKE_POLICY_DEFAULT_CMP0063 NEW)
```
Skylion007, line 82: Saving this for a later time, Figured out how to properly enable the CMake policy, but we need to work through our dependencies an make sure Bullet is built properly with old CMake versions in Shared Lib mode. Saving that for a future PR.

src/CMakeLists.txt
```diff
@@ -63,8 +63,27 @@ if(NOT CMAKE_BUILD_TYPE)
           FORCE
   )
 endif()
+if(CMAKE_BUILD_TYPE MATCHES "Release")
+  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    # Clang has some issues with thinLTO (segfaults on final linking)
+    # This enables full LTO through CMake
+    # FullLTO is slow, but may yield performance improvements.
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -flto")
```
jturner65, line 71: Do we want to revisit this at some time to see if performance is improved?

jturner65 (approved):
From what I can tell, this looks good (when the tests pass) :)

src/CMakeLists.txt
Skylion007: I can confirm it does fix a bug where LTO wasn't being properly set on Emscriptem 2.X
The thing I want to ask @jturner65 is what the default config should be. LTO can slow things down a lot, but it is almost guaranteed to perform a minor performance improvement (particularly with the cyclic dependencies we have in our Codebase now). 

Should calling setup.py install a nice a debug friendly Release? Or should it install one optimized as best as possible. We can try to do both, but I think GCC may have some issue with symbols mangling with LTO enabled for GCC <= 7. Not to mention it does slow linking quite a bit on my underpowered Mac.

setup.py
```diff
@@ -89,7 +89,12 @@ def build_parser():
         action="store_true",
         help="Don't update git submodules",
     )
-
+    parser.add_argument(
+        "--config",
```
erikwijmans, line 93:
```suggestion
        "--build-type",
```
`config` is not a very descriptive name

setup.py
```diff
@@ -89,7 +89,12 @@ def build_parser():
         action="store_true",
         help="Don't update git submodules",
     )
-
+    parser.add_argument(
+        "--build_type",
```
erikwijmans, line 93:
```suggestion
        "--build-type",
```
`-` to match the other args

PR #1249, opened 2021-05-19
[Clang-Tidy] Enable modernize-make checks

Skylion007:
## Motivation and Context

* Enabling another clang-tidy check. This one just enforces the usage of make_[shared,unique] whenever appropriate. This is more efficient calling new and then the constructor for shared pointer (fewer allocations). Individually, this isn't a big deal, but by automating as a check, hopefully it will prevent such minor performance bugs from slipping through. Shared_Ptrs are slow enough so no reason to make them even slower.

## How Has This Been Tested
* Locally and with clang-tidy. I ran it on the code base and it found one instance where it could be improved.

## Types of changes

## Checklist


src/esp/assets/ResourceManager.cpp
```diff
@@ -148,8 +150,8 @@ void ResourceManager::initPhysicsManager(
   if (isEnabled) {
     if (physicsManagerAttributes->getSimulator().compare("bullet") == 0) {
 #ifdef ESP_BUILD_WITH_BULLET
-      physicsManager.reset(
-          new physics::BulletPhysicsManager(*this, physicsManagerAttributes));
+      physicsManager = std::make_shared<physics::BulletPhysicsManager>(
+          *this, physicsManagerAttributes);
```
jturner65, line 154: Pretty sure we need this and the assignment below to be a reset, to retain the same ptr.  @aclegg3?

aclegg3: Yeah, won't this re-assign the particular shared_ptr instance to a new PhysicsManager while leaving any other references alive? We don't want that. We want to change the data all shared_ptrs are referencing here.

src/esp/assets/ResourceManager.cpp
```diff
@@ -165,8 +168,8 @@ void ResourceManager::initPhysicsManager(
   // if the desired simulator is not supported reset to "none" in metaData
   if (defaultToNoneSimulator) {
     physicsManagerAttributes->setSimulator("none");
-    physicsManager.reset(
-        new physics::PhysicsManager(*this, physicsManagerAttributes));
+    physicsManager = std::make_shared<physics::PhysicsManager>(
```
Skylion007, line 171: @jturner65 @aclegg3  Do we want the same behavior down here with the reset?

src/esp/assets/ResourceManager.cpp
erikwijmans: If you want to change the data all shared_ptrs are referencing you need to do `*physicsManager = ...`.  Reset only impacts this specific shared_ptr instance, as does swap.

erikwijmans:
```cpp
int main() {
    auto a = std::make_shared<int>(5);
    auto b = a;

    a.reset(new int{3});

    std::cout << *a << std::endl << *b << std::endl;
}
```
```
3
5
```

Skylion007: https://en.cppreference.com/w/cpp/memory/shared_ptr/operator%3D

Skylion007: Yeah @erikwijmans I was pretty sure it was equivalent behavior. @aclegg3 Given this info, what is the diff suppose to be?

Skylion007: doing `*physics_manager=` works fine for BulletPhysicsManager, but when I construct a normal physics manager and try to set it to that I get errors.

jturner65: The wrapper object managers maintain a weak ptr built from the original shared physicsManager ptr.  Will these weak pointers still track the physicsManager::ptr after its re-assignment?  This was why I was thinking the reset might still be needed.

jturner65 (approved):
LGTM

src/esp/assets/ResourceManager.cpp
aclegg3: Sorry, my mistake. This was not originally meant to reset all smart pointer references. Only Simulator has an active reference to the object. We use shared_ptr because the object managers will have a weak_ptr to the PhysicsManager to do their work. It should be fine to re-assign this referenced smart pointer as Aaron originally suggested.

src/esp/assets/ResourceManager.cpp
aclegg3: Yeah, whatever we end up doing should be the same in both places.

PR #1247, opened 2021-05-18
[WIP][AO migration] collision groups, Bullet debugging, contact queries, RigidObject changes

aclegg3:
## Motivation and Context

ArticulatedObject migration PR split from #1226.

This PR wraps all significant physics related changes which are not directly related to ArticulatedObjects:
- Bullet debug queries (contact points and step summary)
- custom CollisionGroup configuration
- contactTest with configurable STATIC filter
- refactoring object activation API
- refactoring collision compound construction from mesh hierarchy

**These changes are being further broken into individual PRs for review iteration/polish and this PR will be closed.**

## How Has This Been Tested

Should be covered by current tests. Expanded a PhysicsTest to check new contactTest functionality.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1257, opened 2021-05-21
[AO migration] query contact points

aclegg3:
## Motivation and Context

ArticulatedObject migration PR split from #1226.
Contains all contact point query related changes from #1247 originally implemented by @eundersander.

Note: I extracted functionality from BulletDebugManager to keep this isolated.

## How Has This Been Tested

Added a C++ unit test.

## Types of changes

## Checklist


src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -272,13 +310,93 @@ int BulletPhysicsManager::getNumActiveContactPoints() {
...
+        LOG(WARNING)
+            << "getContactPoints: additional warnings will be suppressed.";
+      }
+    }
+  }
+
+  std::vector<ContactPointData> contactPoints;
+
+  // sloppy: assume fixedTimeStep_ hasn't changed since last call to stepPhysics
+  const float recentSubstepDt = fixedTimeStep_;
```
eundersander, line 345: Just to be clear on the bug described by this comment: if you call stepPhysics, then change fixedTimeStep_, then call getContactPoints, quantities like normalForce will be wrong. It would be accurate if we cached the actual recent substepDt somewhere.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -272,13 +310,93 @@ int BulletPhysicsManager::getNumActiveContactPoints() {
...
+  return count;
+}
+
+std::vector<ContactPointData> BulletPhysicsManager::getContactPoints() const {
+  if (m_recentNumSubStepsTaken != 1) {
+    if (m_recentNumSubStepsTaken == -1) {
+      LOG(WARNING) << "getContactPoints: no previous call to stepPhysics";
+    } else {
+      // todo: proper logging-throttling API
+      static int count = 0;
```
eundersander, line 328: This static local is super hacky. And this is a ton of code here just to avoid spamming this warning every frame. I didn't intend this code to merge to master as-is when I wrote it. As I've said before, we should be aiming to clean up these hacks as we're merging AO to master.

eundersander (commented):
A couple hacks here that I'd rather not see merged to master. If you are under a time crunch and need to merge as-is, let's at least document the tech debt somewhere.

src/esp/physics/bullet/BulletPhysicsManager.cpp
aclegg3: Maybe we don't log anything here and just document the function. This only scrapes contacts from the most recent substep or call to discrete collision detection, right? That is pretty straightforward. I guess we could also add a query for the `recentNumSubStepsTaken_` so users can pivot or check this manually if they want to.

eundersander (approved):
Your plan to remove the logging sounds good. The rest looks good, too.

PR #1258, opened 2021-05-21
New flag to enable a Simulator that doesn't create a Magnum/OpenGL context

eundersander:
## Motivation and Context

I added `SimulatorConfiguration::createMagnumRenderer` to enable creation of a Simulator that doesn't create a Magnum/OpenGL context. This flag can be a useful if you're using the Simulator but not doing rendering, e.g. only querying physics or producing gfx-replay keyframes.

In addition to skipping context creation, I also don't load textures or upload meshes to the GPU, but I do still load all meshes (from disk to CPU memory). I would like to see a related flag in the future that disables loading of render-only meshes (which general aren't needed if you're not rendering), but that's a separate PR.

This is a draft PR and needs some cleanup:
- The hacky use of a global `g_createMagnumRenderer` var.
- Probably not all cases of graphics context use have been properly disabled/guarded, e.g. PbrDrawable.
- Let's add a test of creating a non-renderer sim and using various APIs like `drawObservation`.
- We should handle primitives correctly or at least fail gracefully. If we need to disable primitive-loading/use, we should issue a proper warning to users if they try to use a primitive that's unavailable.
- We need to reconcile with the existing `SimulatorConfiguration::createRenderer`. From discussion with @jturner65 , it sounds like the existing functionality of the existing flag (including disabling physics manager) isn't needed at all. So we instead hook up the `createRenderer` flag to the new functionality introduced in this PR (and discard the no-physics-manager code path altogether?).
- Decide whether to create a stub Renderer object (discussed below).
- reconcile two related features: disabling usage of OpenGL (this PR) and disabling loading of render assets (separate feature). After team discussion on 6/1, we agree this can be implemented separately. However, one suggestion is to have an enum for all the combinations of usage, instead of two independent flags. However, we all agreed that this PR doesn't need to implement the second feature, disabling loading of render assets (that can be a future PR).

I show two example usages here: (1) our `PhysicsTest` unit tests, which no longer require a graphics context, and (2) `replay_tutorial.py`, which uses a non-renderer sim to produce gfx-replay keyframes, then later creates a sim-with-renderer to render the replay.

I made a decision here to not create the Renderer object at all. Thus some fixes are required in Hab python programs, e.g., you have to avoid using sim.renderer if it's `None`. This is a bit more invasive than my other choices and might be wrong. ~It might be better to create Renderer as a stub so that python programs don't have to be modified. E.g. `get_observations` can just return all black pixels or perhaps some debug pattern. If we opted for the stub route, we might rename the flag to `createStubRenderer` or `createNullRenderer`.~ Update: after team discussion on 6/1, we don't need to bother with a stub renderer. So the PR can stay as-is in this respect.

## How Has This Been Tested

See PhysicsTest, GfxReplayTest, and replay_tutorial.py.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1088,7 +1088,7 @@ scene::SceneNode* ResourceManager::createRenderAssetInstanceIMesh(
     // That means One CANNOT query the data like e.g.,
     // meshes_.at(iMesh)->getMeshData()->hasAttribute(Mn::Trade::MeshAttribute::Tangent)
     // It will SEGFAULT!
-    createDrawable(*(meshes_.at(iMesh)->getMagnumGLMesh()),  // render mesh
+    createDrawable(meshes_.at(iMesh)->getMagnumGLMesh(),  // render mesh
                    meshAttributeFlags,                 // mesh attribute flags
                    node,                               // scene node
                    creation.lightSetupKey,             // lightSetup key
@@ -1939,7 +1939,12 @@ void ResourceManager::addComponent(
   // Add a drawable if the object has a mesh and the mesh is loaded
   if (meshIDLocal != ID_UNDEFINED) {
     const int meshID = metaData.meshIndex.first + meshIDLocal;
-    Magnum::GL::Mesh& mesh = *meshes_.at(meshID)->getMagnumGLMesh();
+    Magnum::GL::Mesh* mesh = meshes_.at(meshID)->getMagnumGLMesh();
+    if (g_createMagnumRenderer) {
+      CORRADE_INTERNAL_ASSERT(mesh);
+    } else {
+      CORRADE_INTERNAL_ASSERT(!mesh);
+    }
     Mn::ResourceKey materialKey = meshTransformNode.materialID;
 
     gfx::Drawable::Flags meshAttributeFlags{};
@@ -1991,12 +1996,12 @@ void ResourceManager::addPrimitiveToDrawables(int primitiveID,
   // so do not need to worry about the tangent or bitangent.
   // it might be changed in the future.
   gfx::Drawable::Flags meshAttributeFlags{};
-  createDrawable(*primitive_meshes_.at(primitiveID),  // render mesh
-                 meshAttributeFlags,                  // meshAttributeFlags
-                 node,                                // scene node
-                 NO_LIGHT_KEY,                        // lightSetup key
-                 WHITE_MATERIAL_KEY,                  // material key
-                 drawables);                          // drawable group
+  createDrawable(primitive_meshes_.at(primitiveID).get(),  // render mesh
+                 meshAttributeFlags,                       // meshAttributeFlags
+                 node,                                     // scene node
+                 NO_LIGHT_KEY,                             // lightSetup key
+                 WHITE_MATERIAL_KEY,                       // material key
+                 drawables);                               // drawable group
 }
 
 void ResourceManager::removePrimitiveMesh(int primitiveID) {
@@ -2004,7 +2009,7 @@ void ResourceManager::removePrimitiveMesh(int primitiveID) {
   primitive_meshes_.erase(primitiveID);
 }
 
-void ResourceManager::createDrawable(Mn::GL::Mesh& mesh,
+void ResourceManager::createDrawable(Mn::GL::Mesh* mesh,
                                      gfx::Drawable::Flags& meshAttributeFlags,
                                      scene::SceneNode& node,
                                      const Mn::ResourceKey& lightSetupKey,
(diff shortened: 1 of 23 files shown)


erikwijmans:
This is just a draft so just flagging this:

Seems like `createRenderer` does the wrong thing when there is bullet enabled since the intent of that flag was to do allow physics sim (just navmesh stuff when that flag was added).  I don't think we need to add another flag as `createRenderer == false` should still load whatever is needed for physics.

jturner65:
> This is just a draft so just flagging this:
> 
> Seems like `createRenderer` does the wrong thing when there is bullet enabled since the intent of that flag was to do allow physics sim (just navmesh stuff when that flag was added). I don't think we need to add another flag as `createRenderer == false` should still load whatever is needed for physics.

This certainly simplifies how the code should be organized in Simulator::reconfigure.  When I was recoding reconfigure to load stages, and subsequently to create scene instances, the physics objects were only made when a renderer existed, and so I retained that implied partition of functionality.  I also was under the possibly erroneous impression that a renderer was required to instantiate Magnum primitives (the use of which we support for non-implicit collisions).

With this new info, however, it seems like the forking based on whether a renderer is created or not should be performed in RM, at least with regard to scene instance creation, instead of Simulator::reconfigure.

src/esp/assets/ResourceManager.cpp
```diff
@@ -124,6 +126,10 @@ void ResourceManager::buildImporters() {
 }  // buildImporters
 
 void ResourceManager::initDefaultPrimAttributes() {
+  if (!g_createMagnumRenderer) {
```
eundersander, line 129: Todo: issue a warning when users try to use primitives that didn't get loaded due to this code

mosra:
Some unordered high-level thoughts on this, hopefully more in line with the planned multithreaded renderer work. This is assuming there's an intention to eventually merge this and have it maintainable (instead of being a quick throwaway experiment):

- Instead of a global bool flag that creeps into the whole codebase and makes it impossible to verify that all cases are indeed handled, I'd split this on an architectural level -- a renderer part (which requires either being created from an external GPU context or creating its own), and a simulation part (which can exist completely without). All unit tests (physics, simulation alone) that don't need to render anything would then not even instantiate the renderer part and given sufficient coverage that would act as a check that we're not using a GPU context when we shouldn't. This split (I think) also reflects how @bigbike's multithreaded renderer would behave -- if we want to parallelize rendering and simulation, and GPU resources have to be on the rendering thread (to stay compatible with webgl at the very least), then the simulation thread can't really be accessing them as well.
- Because the simulator part shouldn't access any GPU resources, it shouldn't contain things like `GL::Mesh&` or `Shaders::FlatGL*`. Not sure what should be the "easy enough" short-term solution for this PR, but this is something I want to handle in the upcoming `RenderGraph` abstraction as well. I'm thinking about some typed API-independent handles that get allocated by the renderer on request and passed back to the simulator. In the case of builtin primitives for example, this would mean the simulator part creates the desired `Trade::MeshData` for each primitive, asks the renderer to create a mesh from it and remembers the handle. (Or, more in line with the multidraw workflow, it concatenates all desired primitives to one giant `Trade::MeshData`, remembers vertex offsets/counts into it for each primitive and then asks the renderer to allocate just one mesh for all.)
- Based on my (possibly outdated?) understanding of Habitat code and the `uploadBuffersToGPU()` method in particular, it looks like the mesh data are kept forever CPU-side? (Or the data are freed after that, which means `uploadBuffersToGPU()` can't be called twice?) In any case, I want this to change and so it shouldn't be something to rely on -- mesh data used for collision detection and for rendering should get decoupled and the rendering data should be only transient, kept for only as long as needed until GPU upload, and not forever in some member variable.
- Right now, the SceneGraph "draw" process is tightly coupled with calling GL APIs that set all uniforms, bind meshes and textures and so on. This is changing with the UBO work (which is also a prerequisite for the multithreaded renderer) -- in the initial porting step, instead of directly calling into GL, drawables would only populate CPU-side buffers with material references and mesh views (which range of vertices to draw), and those then get passed to the renderer (thread) and uploaded to the GPU all at once right before submitting draws for all meshes combined. In other words, what might be exceptionally hard to split in the current codebase, should become rather easy when the UBO-focused multithreaded renderer work starts getting integrated into the codebase.
- When splitting the responsibilities between the renderer and simulator, you can assume that opening a file using Magnum's importer APIs is relatively cheap and the actual expensive work is done only when you actually ask for image or mesh data. So it's fine to for example open a glTF file, extract just five meshes out of 100 for collision detection purposes and skip the rest because there's currently no renderer attached. There shouldn't be any need to do this differently to make "loading just collision information" more efficient. (There's a bunch of leftover tasks I have listed on the next SoW to optimize this even further, but the main point is that it's designed this way to make exactly these use cases possible.)

eundersander:
Replaced by https://github.com/facebookresearch/habitat-sim/pull/1308

PR #1261, opened 2021-05-24
[CMake Improvements] Visibility, LTO, and setup.py improvements.

Skylion007:
## Motivation and Context

* I went and double checked all the visibility tags across CMake. This seems to work fine for all the libraries we build. It also allows any library to override the fvisibility args the old fashioned way.
* Improves some bug for LTO in the setup.py
* Properly allows LTO option to be overriden by CMake args.
* Also updates GLOG and GTEST (90% CMake + Performance Improvements. No Breaking Changes in Either).

## Types of changes

## Checklist


mosra:
The pace of proposed changes is faster than I can review them and every change has a lot of potential nasty interactions with 3rd party projects we bundle. Could we just revert everything related to LTO _for now_ and retry again later when I'm able to fully focus on this to ensure nothing gets broken?

Thanks.

PR #1262, opened 2021-05-24
1 version of retention for nightly packages

erikwijmans:
## Motivation and Context

Given our average binary size of ~150 MB and building 12 different binaries on the CI, we can't retain any and stay under the 3 GB limit.

## How Has This Been Tested

Just changing a number

### DIFF
--- a/conda-build/common/delete_old_night_packages.py
+++ b/conda-build/common/delete_old_night_packages.py
@@ -9,10 +9,7 @@ import re
 import subprocess
 import sys
 
-# We currently can't have any retention as our average binary
-# size is ~150 MB, so even 1 day of retention would require
-# storing 24 binaries, which is over the 3 GB limit
-MAX_NUMBER_OF_PACKAGES = 0
+MAX_NUMBER_OF_VERSIONS = 1
 
 parser = argparse.ArgumentParser(
     "A tool for removing conda nightly builds in chronical order to reduce conda repo storage usage"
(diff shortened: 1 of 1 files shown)


aclegg3 (approved):

Skylion007 (approved):

erikwijmans:
Hmmm, no retention has the issue that, for the ~2 hours it takes to build everything there are no conda packages in the nightly channel

Skylion007:
> Hmmm, no retention has the issue that, for the ~2 hours it takes to build everything there are no conda packages in the nightly channel

Exactly why we need to enable LTO for Conda-binaries.

Skylion007:
@erikwijmans According to the nightly channel, we only do 8 builds a day: https://anaconda.org/aihabitat-nightly/habitat-sim/files

erikwijmans:
Ah that's right, we aren't building bullet == true headless == false on the CI for some reason.  Although we do need to fix that because that is gonna cause a ton of issues for hab2

conda-build/common/delete_old_night_packages.py
```diff
@@ -9,7 +9,10 @@
 import subprocess
 import sys
 
-MAX_NUMBER_OF_PACKAGES = 30
+# We currently can't have any retention as our average binary
+# size is ~150 MB, so even 1 day of retention would require
+# storing 24 binaries, which is over the 3 GB limit
+MAX_NUMBER_OF_PACKAGES = 0
```
Skylion007, line 15: Change to 6 for now.

erikwijmans: A number that isn't an integer multiple of the number of packages we upload seems strange as we are just keeping an arbitrary selection

Skylion007: Typo: Meant 8

erikwijmans: This feels like we are just leaving ourselves a very hard to catch bug for once we fix the headless == false bullet == true build

erikwijmans: It actually is by upload date (well version).  A package according to this name is a version.  I changed the name to be less confusing.  I'll make a version that does the delete right before uploading everything and the new can do no retention later, but that is way more involved while just retaining 1 version can be pushed tonight

mathfac:
@erikwijmans thank you for fixing the freshness of nightly builds.  Looks like we can increase at least to 2 versions.

Screen Shot 2021-05-27 at 7 49 56 PM

Skylion007:
Yep, only after enabling LTO. Without it we don't have room for even 2.

PR #1263, opened 2021-05-24
[LTO] Change default build config to not use LTO unless requested

Skylion007:
## Motivation and Context
* LTO is no longer done by default. It now needs only done in Release, MinSizeRel or when specified manually with the LTO flag. 
* Adds flag to enable LTO in addition to disabling it.

## How Has This Been Tested
* Locally and with the CI

## Types of changes
* Change default LTO

-
## Checklist


setup.py
```diff
@@ -97,9 +97,18 @@ def build_parser():
     )
     parser.add_argument(
         "--no-lto",
+        dest="lto",
+        default=None,
+        action="store_false",
+        help="Disables Link Time Optimization: faster compile times but worse performance.",
+    )
```
mosra, line 104: Why do we need both `--lto` and `--no-lto`? To the user that might look like there would be a third state, where LTO is neither enabled nor disabled.

src/CMakeLists.txt
```diff
@@ -79,12 +81,12 @@ if(CMAKE_INTERPROCEDURAL_OPTIMIZATION)
...
     set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)
     cmake_policy(SET CMP0069 NEW)
-    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
       # Clang has some issues with thinLTO (segfaults on final linking)
       # This enables full LTO through CMake
       # FullLTO is slow, but may yield performance improvements.
-      add_compile_options(-flto=full)
+      set(CMAKE_CXX_COMPILE_OPTIONS_IPO "-flto")
+      #add_compile_options(-flto=full)
```
mosra, line 89: Why is this commented out?

src/CMakeLists.txt
```diff
@@ -79,12 +81,12 @@ if(CMAKE_INTERPROCEDURAL_OPTIMIZATION)
     # Policy that enables INTERPROCEDURAL_OPTIMIZATION
     set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)
     cmake_policy(SET CMP0069 NEW)
-    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
     if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
       # Clang has some issues with thinLTO (segfaults on final linking)
       # This enables full LTO through CMake
       # FullLTO is slow, but may yield performance improvements.
-      add_compile_options(-flto=full)
+      set(CMAKE_CXX_COMPILE_OPTIONS_IPO "-flto")
```
mosra, line 88: Did you investigate parallel LTO? `-flto=8` for example. GCC has it, not sure if Clang as well. It makes everything faster but also eats a lot more RAM that way, so may cause issues on memory-constrained systems.

(This is exactly why I was hesitating to enable LTO everywhere, because it's either annoyingly slow or memory-hungry. A solution would be to ask the buildsystem to know how many parallelization slots are available (`-flto=jobserver`), which works with Make, but not with Ninja yet.)

src/CMakeLists.txt
```diff
@@ -63,7 +63,9 @@ if(NOT CMAKE_BUILD_TYPE)
           FORCE
   )
 endif()
-if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION AND CMAKE_BUILD_TYPE MATCHES "Rel")
+if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION
+   AND (CMAKE_BUILD_TYPE MATCHES "Release" OR CMAKE_BUILD_TYPE MATCHES "MinSizeRel")
+)
```
mosra, line 68: Wait what. So this enables LTO when `CMAKE_INTERPROCEDURAL_OPTIMIZATION` is not passed on the command line? That's again the wrong default.

I'm not using `setup.py` personally because it does things I don't want, and I'm building directly using CMake and Ninja -- and in that case it would implicitly enable LTO for me.

I'd say, remove this whole `if`?

setup.py
Skylion007: @mosra Yeah, there is where its' picked automatically by the build_type in the CMake file. I wasn't sure how to take care of handling the cached LTO otherwise.

src/CMakeLists.txt
Skylion007: There is not a standard way to force FAT LTO building. They are both hacky. The add_compile_option adds it after the thinLTO so it override replacement, where the CMAKE_CXX_COMPILE_OPTIONS_IPO is an undocumented variable to replace it. I am not sure which is better. Thoughts? I guess the original way might be better since it might preserve other OPTIONS_IPO flags that may eventually get set by CMake if they get added to clang in the future.

setup.py
Skylion007: Also, I think of this as a way forcing whether LTO is enabled / disabled or letting the default take over. (Otherwise, you would need to clean the build folder to switch between LTO and non-LTO builds. I also don't want to force subprojects that may have LTO enabled by default for whatever reason (like PyBind) from disabling the LTO at that stage if it's not set.

src/CMakeLists.txt
Skylion007: I was just trying to mimic how PyBInd1sets LTO and do the same for all the subprojects so if it is set, it's propogated down to the subprojects properly.

setup.py
Skylion007: @mosra In other words, the states are force LTO on for all compilation units, force it off, or let the subproject's respective CMake decide (as was the case before we added the flag to force enable it).

src/CMakeLists.txt
Skylion007: @mosra Enabling LTO is also the default Release config for the emscriptem toolchain though. It would enable in full release mode.

setup.py
aclegg3: This makes sense. I suppose these could be added to a mutually exclusive arg group.

src/CMakeLists.txt
Skylion007: @mosra It's not enabled by default in CMake, but I went ahead and enabled it.

setup.py
Skylion007: @aclegg3 In this case, the last positional arg is the one that takes precedence.

src/CMakeLists.txt
Skylion007: Ah yeah, it breaks Clang-Tidy which is annoying: https://app.circleci.com/pipelines/github/facebookresearch/habitat-sim/6367/workflows/5d3ff9c9-d972-4913-ba17-17f8e91d68ee/jobs/30212

Skylion007:
Ping @mosra

aclegg3 (approved):

bigbike:
Please do not merge. Let Mosra take a final look first.

bigbike:
@mosra

setup.py
mosra: I see, makes sense, thanks for the clarification.

src/CMakeLists.txt
```diff
@@ -84,12 +81,20 @@ if(CMAKE_INTERPROCEDURAL_OPTIMIZATION)
...
       # Clang has some issues with thinLTO (segfaults on final linking)
       # This enables full LTO through CMake
       # FullLTO is slow, but may yield performance improvements.
-      add_compile_options(-flto=full)
+      #set(CMAKE_{lang}_COMPILE_OPTIONS_IPO ${CMAKE_CXX_COMPILE_OPTIONS_IPO} -flto)
+      add_compile_options(-flto) # alternate way of setting LTO.
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+      # If we are on GCC, override CMake do parallel LTO.
+      include(ProcessorCount)
+      ProcessorCount(N_LTO_JOBS)
```
mosra, line 93: As I mentioned previously, the problem with this is that you'll suddenly need 30 or more GB RAM for the build -- on 8 cores it can happen that a 8-parallel LTO job will run together with 7 pybind compilations and each of those alone can easily take two gigabytes of memory. On the CI it could result in flaky builds, where they randomly fail due to running out of memory.

Two options:

- Make the job count configurable by passing a number to the option (so people could do `--lto=$(nproc)` or `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=$(nproc)`, if they have enough RAM) and then you extract that value here and pass it to `-flto=`; `--lto` alone or `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` would mean serial LTO.
- Drop this special case altogether for now.

Another way to fix this would be using [Ninja job pools](https://cmake.org/cmake/help/latest/prop_gbl/JOB_POOLS.html), where the memory-hungry jobs wouldn't run in parallel, but I didn't investigate this thoroughly yet and fear that it would slow down the regular non-LTO build too much.

src/CMakeLists.txt
```diff
@@ -84,12 +81,20 @@ if(CMAKE_INTERPROCEDURAL_OPTIMIZATION)
...
+      # Flag must only be passed to linker to prevent clang-tidy errors
+      string(REPLACE "<LINK_FLAGS>" "<LINK_FLAGS> -flto=${N_LTO_JOBS}"
+                     CMAKE_CXX_LINK_EXECUTABLE "${CMAKE_CXX_LINK_EXECUTABLE}"
```
mosra, line 96: I don't recommend doing this, it's very brittle and consequences of changing this mid-build are unknown. It might work with some compilers / on some platforms and fail spectacularly on other. It should only get changed in the toolchain files, not in the build scripts themselves.

(What's the use case of having both Clang Tidy and LTO enabled? Is that common? I'd say the cases where Clang Tidy is enabled (locally and on the CI) don't need LTO as well, and where LTO is used you don't want to have your build slowed down even further by Clang Tidy.)

Skylion007: We run Clang-Tidy separately as part with run-clang-tidy not as part of the CMake, you just need to generate another compile.json without the flto=N on the compiler args (since it has no affect there anyhow). @mosra We aren't changing it mid build though? This setting it for downstream builds?

src/CMakeLists.txt
Skylion007: Dropping this is fine by my. I added it per your earlier suggestion.

mosra (approved):

PR #1265, opened 2021-05-25
MetadataManager: Add binding to remove dataset functionality.

jturner65:
## Motivation and Context
This PR adds a python binding to the existing remove dataset functionality, as well as a c++ test to verify functionality.

## How Has This Been Tested
Existing c++ and python tests pass. C++ test added to verify functionality.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/MetadataMediatorBindings.cpp
+++ b/src/esp/bindings/MetadataMediatorBindings.cpp
@@ -28,7 +28,7 @@ void initMetadataMediatorBindings(py::module& m) {
       /* --- Methods --- */
       .def(
           "dataset_exists", &MetadataMediator::sceneDatasetExists,
-          R"(Checks whether the passed name references an existing scene dataset)",
+          R"(Returns whether the passed name references an existing scene dataset or not.)",
           "dataset_name"_a)
       .def(
           "remove_dataset", &MetadataMediator::removeSceneDataset,
--- a/src/esp/bindings/MetadataMediatorBindings.cpp
+++ b/src/esp/bindings/MetadataMediatorBindings.cpp
@@ -25,6 +25,11 @@ void initMetadataMediatorBindings(py::module& m) {
           R"(The currently active dataset being used.  Will attempt to load
             configuration files specified if does not already exist.)")
 
+      /* --- Methods --- */
+      .def(
+          "remove_dataset", &MetadataMediator::removeSceneDataset,
+          R"(Remove the given dataset from MetadataMediator.  If specified dataset is currently active, this will fail.)",
+          "dataset_name"_a)
       /* --- Template Manager accessors --- */
       .def_property_readonly(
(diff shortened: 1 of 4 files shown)


Skylion007 (approved):

src/esp/bindings/MetadataMediatorBindings.cpp
```diff
@@ -25,6 +25,15 @@ void initMetadataMediatorBindings(py::module& m) {
...
             configuration files specified if does not already exist.)")
 
+      /* --- Methods --- */
+      .def(
+          "dataset_exists", &MetadataMediator::sceneDatasetExists,
+          R"(Checks whether the passed name references an existing scene dataset)",
+          "dataset_name"_a)
+      .def(
+          "remove_dataset", &MetadataMediator::removeSceneDataset,
+          R"(Remove the given dataset from MetadataMediator.  If specified dataset is currently active, this will fail.)",
```
Skylion007, line 35: Fail how? return false, throw an exception?

src/esp/bindings/MetadataMediatorBindings.cpp
```diff
@@ -25,6 +25,15 @@ void initMetadataMediatorBindings(py::module& m) {
           R"(The currently active dataset being used.  Will attempt to load
             configuration files specified if does not already exist.)")
 
+      /* --- Methods --- */
+      .def(
+          "dataset_exists", &MetadataMediator::sceneDatasetExists,
+          R"(Checks whether the passed name references an existing scene dataset)",
```
Skylion007, line 31: and returns a bool or throws an exception?

src/esp/bindings/MetadataMediatorBindings.cpp
jturner65: Returns false, with a message.

src/esp/bindings/MetadataMediatorBindings.cpp
jturner65: Returns boolean.

Skylion007: I would say returns whether it exists instead of check since check ambiguous.

src/esp/bindings/MetadataMediatorBindings.cpp
Skylion007: ... and return false. Otherwise true. @jturner65 Is this actually a good candidate for an ESP_CHECK?

jturner65: I don't think so.  This is just unloading a particular dataset from MetadataMediator, and returning a success code reflecting whether the deletion happened or not.  This does not verify the existence of a particular disk resource, nor is it intended to do any kind of file management on disk, which is what I would think an ESP_CHECK would be for.

PR #1267, opened 2021-05-25
[Clang-Tidy] Readabillity implicit bool convert

Skylion007:
## Motivation and Context

* Tiny Clang-Tidy check that explicitly expands implicit boolean casts. This is generally good style and can be very useful for finding some very subtle bugs. Enabling this since the string-compare-readability stuff already caught a bug so hopefully this will catch one in the useful. This should be one of my last clang-tidy PRs since we are running out of useful features to enable. Even though this doesn't affect performance or catch any bugs this time around, it will hopefully make the code more readable. 
* I added an option so that nullptrs checks don't have to be expanded. I could do the same for if int = 0 checks, but these are useful to have IMO.

## Types of changes

## Checklist


jturner65 (approved):
LGTM

PR #1272, opened 2021-05-26
[Bugfix] Scene Dataset Attributes Managers access enforcement

jturner65:
## Motivation and Context
This PR makes all scene dataset attributes managers accessed by const refs.  This required a slight change to how the current active dataset is queried - now there will never be an occasion for a null ptr return for any of the managers - if somehow a non-existent dataset has become active (which is not currently possible) then the default dataset will be called upon to provide the managers.  

The scene dataset attributes also has been changed so that its destructor will kill its various managers.  This is so that any persistent external references to the managers will give an error, and not leak, if the dataset they reference has been removed, addressing an extant possible bug in the scene dataset removal process.

Ideally these managers would be accessed by weak pointers, which would address this issue more politely, but such support would necessitate a substantial refactor of how the managers work; this can be addressed at a later date.

## How Has This Been Tested
C++ and python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/attributes/AttributesBase.h
+++ b/src/esp/metadata/attributes/AttributesBase.h
@@ -28,7 +28,7 @@ class AbstractAttributes : public esp::core::AbstractFileBasedManagedObject,
     AbstractAttributes::setHandle(handle);
   }
 
-  virtual ~AbstractAttributes() override = default;
+  ~AbstractAttributes() override = default;
   /**
    * @brief Get this attributes' class.  Should only be set from constructor.
    * Used as key in constructor function pointer maps in AttributesManagers.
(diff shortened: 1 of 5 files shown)


src/esp/metadata/attributes/AttributesBase.h
```diff
@@ -28,7 +28,7 @@ class AbstractAttributes : public esp::core::AbstractFileBasedManagedObject,
     AbstractAttributes::setHandle(handle);
   }
 
-  ~AbstractAttributes() override = default;
+  virtual ~AbstractAttributes() override = default;
```
Skylion007, line 31: redundant. Surprise clang tidy didn't catch this.

Skylion007 (approved):

src/esp/metadata/attributes/AttributesBase.h
jturner65: Is it? will inheriting destructors treat it as virtual? I was unclear.

Skylion007:
override strictly implies virtual.

src/esp/metadata/attributes/AttributesBase.h
jturner65: Heh. well clang-tidy agrees with you.  Groovy :)

PR #1277, opened 2021-05-27
Derived class able to use base class's shared_from_this properly

jturner65:
## Motivation and Context
Small PR that adds functionality so that a derived class's shared_from_this is appropriately downcast when the parent class inherits from std::enable_shared_from_this<parent>.

## Types of changes
All c++ and python tests pass locally

## Checklist

### DIFF
--- a/src/esp/core/esp.h
+++ b/src/esp/core/esp.h
@@ -122,6 +122,24 @@ inline std::ostream& operator<<(std::ostream& os, const box3f& bbox) {
     return std::make_unique<__VA_ARGS__>(std::forward<Targs>(args)...); \
   }
 
+/**
+ * shim function helper for derived class of std::enable_shared_from_this<Base>
+ */
+template <typename Base>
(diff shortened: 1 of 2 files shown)


src/esp/core/esp.h
```diff
@@ -122,6 +122,24 @@ inline std::ostream& operator<<(std::ostream& os, const box3f& bbox) {
     return std::make_unique<__VA_ARGS__>(std::forward<Targs>(args)...); \
   }
 
+/**
```
Skylion007, line 125: Is this really the best way to do this?

Skylion007 (approved):

src/esp/core/esp.h
jturner65: I've seen two refs (SO and elsewhere) using this process to accomplish the desired result. If there's a better way, I'm hoping you guys can point it out so I can improve it. :)

PR #1260, opened 2021-05-24
[WIP][AO migration] collision groups and modified contactTest

aclegg3:
## Motivation and Context

ArticulatedObject migration PR split from #1226.
Contains collision group related changes from #1247 originally implemented by @eundersander.

WIP: adapt to provide better user-end behavior with some default protected behaviors (e.g. STATIC, KINEMATIC, and free objects with locked behaviors). 

## How Has This Been Tested

TODO: add some specific tests for group filter behavior

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1280, opened 2021-05-28
[AO migration] collision groups

aclegg3:
## Motivation and Context

Expanded collision group implementation spun off from #1260.

Primary new feature is an interface for configuring collision groups and masks `CollisionGroupHelper` enabling users to:
- assign custom names to collision groups
- manipulate group masks
- assign custom groups to objects via wrapper with `object.override_collision_group`

Secondary Changes:
- deprecated `Simulator::ContactTest` and added wrapper bindings for `object.contact_test`
- Cleaned up BulletRigidObject creation regarding MotionTypes and related test (may slightly change existing interactions between Kinematic and Dynamic objects).

Future work:
- add optional collision group override to contact queries (e.g. contact_test and raycast) to provide more flexible querying options (e.g. cast a ray which only reports hits against STATIC objects).

## How Has This Been Tested

- Added more C++ testing for contact filtering between standard groups (Dynamic, Kinematic, Static).
- Added python tests for new features in `tests/test_physics.py`.
- Refactored MotionType tests to account for new changes.

## Types of changes

## Checklist


src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * @brief Defined available collision groups with bitwise integer correspondance
+ * for use in collision filtering masks.
+ *
+ * Contains several named default groups used by Habitat-sim and a set of unused
+ * user groups for custom behavior.
+ */
+enum class CollisionGroup {
+  //! Default group for unspecified object (e.g. raycast)
+  Default = 1,
```
Skylion007, line 51: Nit: but maybe you should do 2<<N syntax to make it more ear they are flags.

src/esp/physics/objectWrappers/ManagedPhysicsObjectBase.h
```diff
@@ -112,6 +112,20 @@ class AbstractManagedPhysicsObject : public esp::core::AbstractManagedObject {
     }
   }  // setLightSetup
 
+  bool contactTest() {
+    if (auto sp = this->getObjectReference()) {
+      return sp->contactTest();
+    } else {
```
Skylion007, line 118: else return is discouraged by style guide.

Skylion007 (approved):

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  /**
+   * @brief Get a list of all configured collision group names for easy
+   * iteration over groups.
+   *
+   * @return List of configured group names
+   */
+  static std::vector<std::string> getAllGroupNames() {
+    std::vector<std::string> groupNames;
```
Skylion007, line 196: reverse std::vector here

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,236 @@
...
+   * corresponding to whether a mask bit is set indicating interaction with each
+   * other group.
+   *
+   * @param group The collision group for which the mask will be queried.
+   * @return The group mask split into per-group booleans indicating interaction
+   * with each group.
+   */
+  static std::map<CollisionGroup, bool> getSplitMask(
+      const CollisionGroup& group) {
+    std::map<CollisionGroup, bool> splitMask;
```
Skylion007, line 218: reserve space

src/esp/physics/CollisionGroupHelper.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include "CollisionGroupHelper.h"
+
+namespace esp {
+namespace physics {
+
+// initialize the default collision group names
+std::map<std::string, CollisionGroup>
+    CollisionGroupHelper::collisionGroupNames = {
+        {"Default", CollisionGroup::Default},
```
Skylion007, line 13: Optional: This makes it clear this could be a Corrade EnumSet https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html with all the convience functions you get from that.

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,236 @@
...
+*/
+
+/**
+ * @brief Defined available collision groups with bitwise integer correspondance
+ * for use in collision filtering masks.
+ *
+ * Contains several named default groups used by Habitat-sim and a set of unused
+ * user groups for custom behavior.
+ */
+enum class CollisionGroup {
```
Skylion007, line 49: Optional: This makes it clear this could be a Corrade EnumSet https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html with all the convience functions you get from that.

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * @brief A convenience class with all static functions providing an interface
+ * for customizing collision masking behavior for simulated objects.
+ */
+class CollisionGroupHelper {
+  //! maps custom names to collision groups
+  static std::map<std::string, CollisionGroup> collisionGroupNames;
+  //! maps collision groups to collision group filter masks
+  static std::map<CollisionGroup, int> collisionGroupMasks;
```
Skylion007, line 88: Collision groups are effectively an unsigned in. Doesn't that mean this should really just be an array?

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,236 @@
...
+   * corresponding to whether a mask bit is set indicating interaction with each
+   * other group.
+   *
+   * @param group The collision group for which the mask will be queried.
+   * @return The group mask split into per-group booleans indicating interaction
+   * with each group.
+   */
+  static std::map<CollisionGroup, bool> getSplitMask(
+      const CollisionGroup& group) {
+    std::map<CollisionGroup, bool> splitMask;
```
Skylion007, line 218: Since collision group is an unsigned int, this makes more sense as a https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1BoolVector.html

Skylion007: Actually, a bool vector is overkill, it's just it's own unsigned int, right?

Skylion007 (changes requested):
We can use the fact that each collision group is itself a number to further remove a whole bunch of std::map for check collision groups. We can also exploit some Corrade containers for further speed up.

src/esp/physics/CollisionGroupHelper.h
Skylion007: At the very least, it can inherit from an unsigned int.

src/esp/physics/CollisionGroupHelper.cpp
```diff
@@ -0,0 +1,74 @@
...
+        {"UserGroup7", CollisionGroup::UserGroup7},
+        {"UserGroup8", CollisionGroup::UserGroup8},
+        {"UserGroup9", CollisionGroup::UserGroup9},
+        {"AllFilter", CollisionGroup::AllFilter},
+};
+
+// initialize the default collision group masks
+std::map<CollisionGroup, int> CollisionGroupHelper::collisionGroupMasks = {
+    // everything except Noncollidable
+    {CollisionGroup::Default, -1 & ~int(CollisionGroup::Noncollidable)},
```
Skylion007, line 34: Corrade's EnumSet would clean this code up a lot.

Skylion007:
If you want an example of how to expose these types to Python: https://github.com/facebookresearch/habitat-sim/blob/84fb827038b98a3641e1194c85798298f13cf19c/src/esp/bindings/GfxBindings.cpp#L52

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,236 @@
...
+  //! Free (i.e., DYNAMIC) objects are integrated by dynamic simulation
+  FreeObject = 1 << 3,
+  //! convenience groups to separate robots from free objects
+  Robot = 1 << 4,
+  //! convenience group for objects which should not collide with anything
+  Noncollidable = 1 << 5,
+
+  // User configurable groups with "CollisionGroup::Default" behavior unless
+  // modified
+  UserGroup1 = 1 << 6,
```
Skylion007, line 65: I just realized. No UserGroup0? -_- :P

Skylion007 (approved):
Looks good to me! Assuming Clang-Tidy and all tests pass.

src/esp/physics/CollisionGroupHelper.cpp
```diff
@@ -0,0 +1,86 @@
...
+        // all but Static and Kinematic
+        {CollisionGroup::Kinematic,
+         ~CollisionGroups(CollisionGroup::Static | CollisionGroup::Kinematic |
+                          CollisionGroup::Noncollidable)},
+        // everything except Noncollidable
+        {CollisionGroup::FreeObject,
+         ~CollisionGroups(CollisionGroup::Noncollidable)},
+        // everything except Noncollidable
+        {CollisionGroup::Robot,
+         ~CollisionGroups(CollisionGroup::Noncollidable)},
```
mosra, line 50: Just `~CollisionGroup::Noncollidable` and `~(blah|blah)` elsewhere doesn't work? It should. :thinking:

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,207 @@
...
+   */
+  static std::vector<std::string> getAllGroupNames() {
+    std::vector<std::string> groupNames(collisionGroupNames.size());
+    for (std::map<std::string, CollisionGroup>::iterator it =
+             collisionGroupNames.begin();
+         it != collisionGroupNames.end(); ++it) {
+      groupNames.push_back(it->first);
+    }
+    return groupNames;
+  }
```
mosra, line 202: I would de-inline all methods into the cpp file, standard containers generate amazing amounts of code, making compilation slower.

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * @brief Get a list of all configured collision group names for easy
+   * iteration over groups.
+   *
+   * @return List of configured group names
+   */
+  static std::vector<std::string> getAllGroupNames() {
+    std::vector<std::string> groupNames(collisionGroupNames.size());
```
mosra, line 195: wait, doesn't this create a vector that has double the items and the first half is empty strings?

(I think you meant to either call `reserve()` instead, or use `[]` instead of `push_back()`?)

src/esp/physics/CollisionGroupHelper.cpp
Skylion007: ```invalid argument type 'esp::physics::CollisionGroup' to unary expression``` was the error.

mosra: But that's what `CORRADE_ENUMSET_OPERATORS` should take care of :O

aclegg3: Hm, VSCode doesn't like this, but it does compile.

src/esp/bindings/PhysicsBindings.cpp
```diff
@@ -69,6 +72,69 @@ void initPhysicsBindings(py::module& m) {
...
+          "get_mask_for_group",
+          [](const std::string& group_name) {
+            return CollisionGroup(
+                uint32_t(CollisionGroupHelper::getMaskForGroup(group_name)));
+          },
+          "group"_a,
+          R"(Get the mask for a collision group describing its interaction with other groups.)")
+      .def_static(
+          "set_mask_for_group",
+          [](CollisionGroup group, CollisionGroup mask) {
```
eundersander, line 125: I am puzzled by `CollisionGroup mask`. Shouldn't the type of mask be `int` or `unsigned int`? When you do bitwise-or of a few CollisionGroup enum values, the resulting value is *not* a valid CollisionGroup enum value, right?

src/esp/physics/CollisionGroupHelper.h
```diff
@@ -0,0 +1,166 @@
...
+ * user groups for custom behavior.
+ */
+enum class CollisionGroup : unsigned int {
+  //! Default group for unspecified object (e.g. raycast)
+  Default = 1,
+  //! STATIC objects should not change state
+  Static = 1 << 1,
+  //! KINEMATIC objects are updated manually
+  Kinematic = 1 << 2,
+  //! Free (i.e., DYNAMIC) objects are integrated by dynamic simulation
```
eundersander, line 57: Why not call it "Dynamic"? Does "free" convey anything useful?

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -463,10 +470,33 @@ Magnum::Vector3 BulletRigidObject::getCOM() const {
...
+  src.m_collisionFilterMask =
+      bObjectRigidBody_->getBroadphaseHandle()->m_collisionFilterMask;
+  // TODO: custom collision group/mask override support
+  // if (!staticAsStage && objectMotionType_ == MotionType::STATIC) {
+  // override collision filter for STATIC object to include other STATICs such
+  // as the stage
+  //  src.m_collisionFilterGroup = int(CollisionGroup::FreeObject);
+  //  src.m_collisionFilterMask =
+  //      CollisionGroupHelper::getMaskForGroup(CollisionGroup::FreeObject);
+  //}
```
eundersander, line 484: Let's clean up this comment block above.

tests/test_physics.py
```diff
@@ -434,3 +434,102 @@ def test_raycast():
...
+            sim.get_physics_simulation_library()
+            != habitat_sim.physics.PhysicsSimulationLibrary.NoPhysics
+        ):
+            cgh = habitat_sim.physics.CollisionGroupHelper
+            cg = habitat_sim.physics.CollisionGroups
+
+            # test group naming
+            assert cgh.get_group_name(cg.UserGroup1) == "UserGroup1"
+            assert cgh.get_group("UserGroup1") == cg.UserGroup1
+            cgh.set_group_name(cg.UserGroup1, "my_custom_group_1")
```
eundersander, line 473: Nit: sometimes we use unit tests to demonstrate functionality, almost like a tutorial. It would be nice if you showed a real robotics use case here with meaningful group names. As-is, users will probably wonder, why bother renaming enum UserGroup1 to "my_custom_group_1"? What am I supposed to do with this?

eundersander (approved):
LGTM (looks good to me). I left some minor comments for you to consider.

src/esp/bindings/PhysicsBindings.cpp
aclegg3: I spoke to Mosra about this. Essentially the CollisionGroup enum variable here will hold the uint mask value on the python side without an explicit category name (e.g. as produced by the set operators) and this lambda casts that value to a C++ EnumSet at this layer. I know it looks a bit odd.

src/esp/physics/CollisionGroupHelper.h
aclegg3: I'll switch it. Dynamic would be more in-line with other naming conventions.

tests/test_physics.py
aclegg3: Good point. This may make more sense once the AOs are merged in and we can do a grasping example and/or the override parameters to raycast/contactTest allow custom queries for things like sensor region objects.

PR #1284, opened 2021-05-30
add --no-web-apps for JS build; update JS build documentation

eundersander:
## Motivation and Context

This flag helps define the division between the Emscripten-built `hsim_bindings.wasm` binary and the demo web apps that showcase it. I also updated documentation to convey this.

The flag simplifies and speeds up the build if you're just building the binary and not using the demo web apps.

This is a small change but there's a lot of diff spam due to indenting changes.

## How Has This Been Tested

I built locally with `--no-web-apps`. (Is a new CI step warranted? To build with this flag?)

## Types of changes

## Checklist

### DIFF
--- a/build_js.sh
+++ b/build_js.sh
@@ -4,10 +4,12 @@
 set -e
 
 BULLET=false
+WEB_APPS=true
 
 while [[ "$#" -gt 0 ]]; do
     case $1 in
         --bullet) BULLET=true ;;
+        --no-web-apps) WEB_APPS=false ;;
         *) echo "Unknown parameter passed: $1"; exit 1 ;;
     esac
     shift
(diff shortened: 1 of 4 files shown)


build_js.sh
```diff
@@ -4,10 +4,12 @@
 set -e
 
 BULLET=false
+WEB_APPS=true
 
 while [[ "$#" -gt 0 ]]; do
     case $1 in
         --bullet) BULLET=true ;;
+        --no-web-apps) WEB_APPS=false ;;
```
Skylion007, line 12: Seems like an odd request as their are other ways to speed up compilation (like not compiling with Release or disabling LTO). It would be nice to have tests for this, but I don't know if any of them would be useful besides does it compile with current infra.

Skylion007 (approved):

build_js.sh
```diff
@@ -41,19 +43,24 @@ cmake ../src \
     -DCMAKE_PREFIX_PATH="$EMSCRIPTEN" \
     -DCMAKE_TOOLCHAIN_FILE="../src/deps/corrade/toolchains/generic/Emscripten-wasm.cmake" \
     -DCMAKE_INSTALL_PREFIX="." \
-    -DCMAKE_CXX_FLAGS="-s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1" \
+    -DCMAKE_CXX_FLAGS="-s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1 -s ASSERTIONS=0" \
```
Skylion007, line 46: @mosra Is this not already covered by the Magnum toolchain? Should it be added for the default Release flags?

eundersander: ASSERTIONS=0 is the default behavior so this change has no effect. I just wanted it to be easier for folks (including myself) to turn on asserts.

This is admittedly unrelated to the rest of the changes in this PR.

PR #1289, opened 2021-06-02
Fix vertex colored flat shading

ldcWV:
## Motivation and Context

Currently, it is impossible to view a mesh that is flat shaded because this case is being handled incorrectly. Flat shaded meshes show up completely black. This PR fixes that issue.

Before:
Screen Shot 2021-06-04 at 12 56 16 PM

After:
Screen Shot 2021-06-04 at 12 59 13 PM

In addition, the property `vertexColored` is now removed from `PhongMaterialData`, as this is not technically a property of a material because it depends on the geometry of the mesh. It has now been changed to a flag for `Drawable`.

## How Has This Been Tested

A flat shaded mesh was loaded into the viewer successfully. If more testing is necessary, please give suggestions.

## Types of changes

## Checklist


src/esp/assets/ResourceManager.cpp
```diff
@@ -1608,7 +1608,7 @@ gfx::PhongMaterialData::uptr ResourceManager::buildFlatShadedMaterialData(
     finalMaterial->ambientTexture =
         textures_.at(textureBaseIndex + material.diffuseTexture()).get();
   } else {
-    finalMaterial->ambientColor = material.ambientColor();
+    finalMaterial->ambientColor = Mn::Color4(1.f, 1.f, 1.f, 1.f);
```
bigbike, line 1611: May I understand the reason why you would like to change this?

bigbike (commented):
Have you tried to load any vertex colored mesh?
Post the result image in the comments, will you? :)

ldcWV:
Screen Shot 2021-06-02 at 5 00 31 PM
Screenshot requested by @bigbike of a vertex colored mesh. Better before/after pictures will be uploaded later.

bigbike:
Interesting. This mesh is vertex colored mesh for the Replica dataset. We can display it before even without your change.

Can you try the "master" branch on this mesh? I suspect you can display it already.

src/esp/assets/ResourceManager.cpp
ldcWV: I'm not sure the exact reason because @eundersander wrote this part, so maybe he can give an explanation. However, without this change, everything appears black in the Replica dataset mesh example.

ldcWV:
Just tried it on the master branch, and it's all black:
Screen Shot 2021-06-03 at 8 52 48 AM

eundersander:
>Interesting. This mesh is vertex colored mesh for the Replica dataset. We can display it before even without your change.

Context on this test model: I provided it to Lawrence. It is a GLB that I created (not the Replica source PLY). I agree that we'd like to see another vertex-colored test model here.

erikwijmans:
You'll need to add the HasVertexColor flag here https://github.com/facebookresearch/habitat-sim/blob/8ec2ef573a8367670fc7c7ed678f0c8c8ff8ae49/src/esp/assets/ResourceManager.cpp#L1171

src/esp/assets/ResourceManager.cpp
bigbike: No, this is the ambientColor (environment color). Do not set it unless the material says so.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2234,7 +2239,6 @@ void ResourceManager::initDefaultMaterials() {
   shaderManager_.set<gfx::MaterialData>(WHITE_MATERIAL_KEY, whiteMaterialData);
   auto* perVertexObjectId = new gfx::PhongMaterialData{};
   perVertexObjectId->perVertexObjectId = true;
-  perVertexObjectId->vertexColored = true;
```
bigbike, line 2237: Interesting. 
Then the instance mesh will not be displayed properly because before we set this vertexColored in the "material" to info the simulator that any mesh chunk that uses this material is vertexColored.

InstanceMesh is such a case. (mesh_semantic.ply in any Replica dataset model, in `/habitat/` sub-folder)

src/esp/assets/ResourceManager.cpp
eundersander: The ambient color defined in the material is often going to be black, but if we're going to render as flat-shaded, we need ambient color to be set to white. Notice how this function sets ambient color to white by default at the top.

Remember how our "flat-shading" works: we run the phong shader in a special way with no lights, making sure the ambient contribution is appropriate to give the flat-shaded appearance.

We do need this fix for non-textured models to be flat-shaded correctly. We can't leave the code as-is. A cleaner fix would be to simply remove this else block altogether (thus leaving ambient color as white).

ldcWV:
> You'll need to add the HasVertexColor flag here
> 
> https://github.com/facebookresearch/habitat-sim/blob/8ec2ef573a8367670fc7c7ed678f0c8c8ff8ae49/src/esp/assets/ResourceManager.cpp#L1171

I pushed a change that does this, although I'm not sure if this is actually a good idea. It will apply to all instance meshes because currently, there's no way to determine whether it's actually supposed to be vertex colored or not (see the warning comment under the changed line). This is a bigger issue than this PR is trying to address.

erikwijmans:
Yep, not the job of this PR to address that issue, but currently we default to instance meshes have vertex coloring, so we shouldn't change that unintentionally.

src/esp/assets/ResourceManager.cpp
eundersander: Note that Lawrence and I have made more progress here. We have a better solution that what I described above, coming soon.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1607,9 +1608,15 @@ gfx::PhongMaterialData::uptr ResourceManager::buildFlatShadedMaterialData(
...
         textures_.at(textureBaseIndex + material.diffuseTexture()).get();
-  } else {
+  }
+
+  if (material.hasAttribute(Mn::Trade::MaterialAttribute::DiffuseColor)) {
+    finalMaterial->ambientColor = material.diffuseColor();
+  } else if (material.hasAttribute(
+                 Mn::Trade::MaterialAttribute::AmbientColor)) {
     finalMaterial->ambientColor = material.ambientColor();
   }
```
bigbike, line 1618: If it does NOT have either the diffuse color or the ambient color, it will set the `finalMaterial->ambientColor = Mn::Vector3(1.0f, 1.0f, 1.0f)`.

Do you think this is necessary?

src/esp/assets/ResourceManager.cpp
```diff
@@ -1168,7 +1168,8 @@ scene::SceneNode* ResourceManager::createRenderAssetInstanceIMesh(
 
     // Instance mesh does NOT have normal texture, so do not bother to
     // query if the mesh data contain tangent or bitangent.
-    gfx::Drawable::Flags meshAttributeFlags{};
+    gfx::Drawable::Flags meshAttributeFlags{
+        gfx::Drawable::Flag::HasVertexColor};
```
bigbike, line 1172: Yes, this is very necessary. Thanks.

bigbike (approved):

bigbike:
Good job, @ldcWV ! 

I approved this PR but left a question for you to consider.

Also please wait the CI tests passes.

src/esp/assets/ResourceManager.cpp
bigbike: I agree with you and @eundersander : If there is no ambient color, the mesh will not be visible. 

See the code here:
https://github.com/mosra/magnum/blob/dd3ce93888ada70074d63d72c5998d950fa2eba6/src/Magnum/Shaders/Phong.frag#L250

ldcWV: The alternative is to set it to `material.ambientColor()`. If `material` has an ambient texture, this will set it to white, which is the same result. Otherwise, it will be set to black. I think it's more useful for it to be set to white so that vertex colored meshes are viewable.
Screen Shot 2021-06-04 at 2 53 17 PM

bigbike: I did not understand your comment. Our goal is to set the `finalMaterial` and `material` is just the input.

What I suggested is:

```
  if (material.hasAttribute(Mn::Trade::MaterialAttribute::DiffuseColor)) {
    finalMaterial->ambientColor = material.diffuseColor();
  } else if (material.hasAttribute(
                 Mn::Trade::MaterialAttribute::AmbientColor)) {
    finalMaterial->ambientColor = material.ambientColor();
  } else {
  // set finalMaterial explicitly to some color.e.g, (1.0f, 1.0f, 1.0f);
}
```
If there is NO light, yes, we need to set the ambientColor in the final Material. Cannot let it black. Otherwise you will see nothing.

ldcWV: Oh I see what you mean. I think the code already does this though on the previous line
```  finalMaterial->ambientColor = 0xffffffff_rgbaf;```
If `material` has neither diffuseColor nor ambientColor, then `finalMaterial->ambientColor` will stay as white.

bigbike: Oh, OK. I see. 
`finalMaterial->ambientColor = 0xffffffff_rgbaf;`
This is hided by the github.

Then I do not have any questions. Make sure the CI tests passed. Then merge it. Thanks.

Skylion007:
Pull from master to fix the Habitat-Lab issue.

src/esp/assets/ResourceManager.cpp
eundersander: @ldcWV have you tested this change?

What I'm imagining: 1. find a suitable test model that uses this codepath (I can provide one, let me know). 2. On your local system, put a breakpoint or print statement here to sanity-check that you're testing this codepath (createRenderAssetInstanceIMesh). 3. Run locally and verify the model is rendered correctly.

ldcWV: Yep, Yili gave me a .ply vertex colored model to test this. Without the flag everything is white, but adding the flag makes the model colors viewable.
Screen Shot 2021-06-03 at 1 43 50 PM
Screen Shot 2021-06-03 at 2 32 12 PM

eundersander: Is it supposed to be all purple? Can we find another test model or otherwise verify this is correct? @bigbike thoughts?

bigbike:
Wait. Why I cannot see any comments from @eundersander ? I received a lot of comments in email notification. But cannot see any here?

To answer the "invisible" question:
Yes, that model is purple. I confirm it. Sorry I happened to give him a model that looks a bit strange.

src/esp/assets/ResourceManager.cpp
bigbike: Oh, I see. The issue is marked as resolved. That is why I did not see it. 

Yes, it is supposed to be all purple.
Yes, we can find another one to test.

ldcWV:
> Wait. Why I cannot see any comments from @eundersander ? I received a lot of comments in email notification. But cannot see any here?
> 
> To answer the "invisible" question:
> Yes, that model is purple. I confirm it. Sorry I happened to give him a model that looks a bit strange.

He replied to an already resolved conversation. The thread is the second one in your approval comments.

eundersander (approved):
Nice work! LGTM!

PR #1290, opened 2021-06-03
Basis Compression Tool

erikwijmans:
## Motivation and Context

* Add a tool to create basis compressed basis GLBs
* Compile the basis compressor, conditionally
* Add support for using cmake from `pip install cmake` that way people can easily get a more recent version of cmake as compiling requires a newer version

## Testing

Compressing various datasets!


Skylion007:
@erikwijmans Are we still against making Basis a submodule? Because now we are adding quite a few files.

tools/create_basis_compressed_glbs.py
```diff
@@ -0,0 +1,253 @@
...
+    assert osp.exists(gltf_name)
+    with open(gltf_name, "r") as f:
+        json_data = json.load(f)
+
+    for material in json_data["materials"]:
+        assert "pbrMetallicRoughness" in material
+
+        # Drop everything except base color and base color texture
+        pbrMetallicRoughness = material["pbrMetallicRoughness"]
+        for key in list(pbrMetallicRoughness.keys()):
```
Skylion007, line 113: Why the list conversion?

erikwijmans: No reason, good catch.

src/deps/basis-universal/basisu.sln
```diff
@@ -0,0 +1,31 @@
...
+		{59586A07-8E7E-411D-BC3D-387E039AA423}.Release|x86.ActiveCfg = Release|Win32
+		{59586A07-8E7E-411D-BC3D-387E039AA423}.Release|x86.Build.0 = Release|Win32
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {5E583429-7830-4B3A-9DDE-F01B115CE0D8}
+	EndGlobalSection
+EndGlobal
```
jturner65, line 31: Any reason why we're including a VS project file?

jturner65 (commented):
I would think a submodule would be preferable for future support and bug fixes.

aclegg3:
Thanks for packaging this up and publishing it here, @erikwijmans. Should be very useful for all of us and the community to have an easy way to do this conversion!

I agree with other comments that it would be best to avoid pulling all of this code into Habitat-sim if we can find another way (dependency or submodule) to do so.

erikwijmans:
The download for the basis git repo is ~200 MB, which is 100x larger than taking just what we need.  I pushed the submodule so just an FYI

Skylion007:
200mb is not unreasonable for a submodule, and similar to the size of some other submodules we have like Bullet. I vote to keep it as submodule for easy upgrades, bugfixes, etc.

Skylion007:
If we are worried about this being cloned a lot even when it's a relatively rarely used submodule, we could make a CMAKE external project: https://cmake.org/cmake/help/git-stage/module/ExternalProject.html

erikwijmans:
That's a new 3.20 feature and pretty buggy from what I've heard

Skylion007:
My vote remains keeping it in as a submodule.

mosra:
I vote for including just the files, as you did originally, especially since this dependency is not needed for 90% users (unlike Bullet or Assimp). Updating it would mean just replacing the files, I see no problem there (this is how I do it for Homebrew and Arch packages already anyway) -- it's actually more likely that I would need to apply manual patches to get rid of various silly warnings, and a submodule makes that harder.

Basis has a "I hate Git"-style maintenance and that they don't care doesn't mean we should be careless too ;) Bundling it as a submodule would also act as a dangerous precedent for other dependencies ("hey, it's fine to bundle this gigabyte dataset, the other submodules are 500 MB already anyway") :stuck_out_tongue_closed_eyes:

src/CMakeLists.txt
```diff
@@ -2,7 +2,22 @@
...
+  "Wether or not to build the basis compressor.  Loading basis compressed meshes does NOT require this."
+  OFF
+)
+
+if(BUILD_BASIS_COMPRESSOR)
+  if(CMAKE_VERSION VERSION_LESS "3.13")
+    message("Building the basis compressor requires a newer version of cmake.")
+    message("You can get one with pip install cmake")
+  endif()
+  cmake_minimum_required(VERSION 3.13)
```
mosra, line 16: This command could be directly in `utils/imageconverter/CMakeLists.txt`, I think, that would make it more modular -- the minimum CMake version could change based on what all subdirs are added, while the top-level file requires still just 3.10.

src/CMakeLists.txt
```diff
@@ -2,7 +2,22 @@
...
+  cmake_policy(SET CMP0079 NEW)
```
mosra, line 17: If you require 3.13, this is set implicitly I think.

src/cmake/dependencies.cmake
```diff
@@ -227,13 +227,19 @@ if(NOT USE_SYSTEM_MAGNUM)
   # submodule we bundle just the transcoder files, and only a subset of the
   # formats (BC7 mode 6 has > 1 MB tables, ATC/FXT1/PVRTC2 are quite rare and
   # not supported by Magnum).
-  set(BASIS_UNIVERSAL_DIR "${DEPS_DIR}/basis-universal")
+  set(BASIS_UNIVERSAL_DIR "${DEPS_DIR}/basis_universal")
```
mosra, line 230: why? :D (i hated the underscore so I renamed it, originally)

mosra (commented):
About the submodule, see my comment above :)

erikwijmans:
@aclegg3 @jturner65 @Skylion007 -- Mosra made some good points wrt. keeping things as files instead of a submodule.  It also doesn't seem like basis is keeping their API very stable (they are changing functions, args, etc) so upgrading likely isn't trivial.

mosra:
> doesn't seem like basis is keeping their API very stable

Yeah, one reason why Magnum is still stuck on the pre-UASTC version of Basis (and backporting various patches to it). The APIs changed significantly since, certain things that worked previously are crashing now and I didn't find the courage to look into the changes yet.

jturner65:
> 
> 
> > doesn't seem like basis is keeping their API very stable
> 
> Yeah, one reason why Magnum is still stuck on the pre-UASTC version of Basis (and backporting various patches to it). The APIs changed significantly since, certain things that worked previously are crashing now and I didn't find the courage to look into the changes yet.

Is there a way we could access this functionality through Magnum?

mosra:
What do you mean? The functionality is exposed and used through Magnum's [BasisImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1BasisImporter.html) and [BasisImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1BasisImageConverter.html) plugins, yes.

tools/create_basis_compressed_glbs.py
```diff
@@ -0,0 +1,253 @@
...
+    osp.join(
+        osp.dirname(__file__),
+        "..",
+        "src/deps/magnum-plugins/src/MagnumPlugins/TinyGltfImporter/Test",
+    )
+)
+
+IMAGE_CONVERTER = osp.realpath(
+    osp.join(
+        osp.dirname(__file__), "..", "build/utils/imageconverter/magnum-imageconverter"
```
eundersander, line 25: In my experience using this script, I have to locally modify here because I don't usually name my build directory "build". It would be nice to offer a friendly error message like "build directory not found" for users like me.

tools/create_basis_compressed_glbs.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def img_name_to_basis(img: str) -> str:
+    return osp.splitext(img)[0] + ".basis"
+
+
+def convert_image_to_basis(img: str) -> None:
+    basis_img = img_name_to_basis(img)
+
+    settings = "threads=4,mip_gen=true,compression_level=2"
```
eundersander, line 94: What's `compression_level`? Can we document here? Is this an interesting choice for the user?

eundersander (approved):
LGTM. I left some minor comments.

I don't have a strong opinion on submodule versus duplicated files. I do wish there was a workflow for "optional submodules". 90% of Hab users will never use this. Similarly, 90% of Hab users will never use our VHACD functionality. I wish a user could just init/clone specific submodules as needed.

mosra:
> I wish a user could just init/clone specific submodules as needed.

There is a way, but as far as i know, it's not possible for the repo itself to specify that (would be great if it could be said in `.gitmodules`), the users have to know what submodules they want and what not, which isn't really user-friendly. This is what I use locally -- as I have system-wide Assimp, GLFW, Bullet, Magnum, Eigen... installation, instead of `git submodule update --init` I init just the modules that are actually used:

```sh
cd src/deps
git submodule init glog
git submodule init googletest
git submodule init imgui
git submodule init recastnavigation
git submodule init Sophus
git submodule init tinyobjloader
git submodule init tinyxml2
git submodule init v-hacd
git submodule update # clones / fetches only the submodules that were init'd
```

tools/create_basis_compressed_glbs.py
erikwijmans: Basis doesn't document it with anything more descriptive than "the overall effort the compressor uses on a 1-5 scale".  I'll expose it, but there isn't a great rule of thumb for setting it when you can't manually verify if there are compression artifacts (higher may reduce them).  I just put its as high as I could while still keeping things to a reasonable amount of time and checked a few meshes for artifacts.

erikwijmans:
I switched back to just including the files since upgrading basis is a more involved process than just changing a commit, so we don't get the biggest benefit of a submodule but get all its downsides.

mosra (approved):
The `basisu_tool.cpp` doesn't need to be included, as that's just the source of their `basisu` console tool, which the plugin doesn't use in any way. For the record, [here's the set of cpp files Magnum uses](https://github.com/mosra/magnum-plugins/blob/ae1278de39a2b8b8f46dd590d39a15d0b410f18e/modules/FindBasisUniversal.cmake#L121-L135), I think all other files besides this one are needed.

Besides that, this looks good to me :)

PR #1291, opened 2021-06-03
[GFX][FisheyeBindings] Allow principal point to default to center of sensor

Skylion007:
## Motivation and Context

* We currently have to specify the Principal Point which is a big annoyance on the Habitat-Lab side. Most of the time, the principal point should lie in the center of the image. If not, the user can still specify it. Using the new fixed Corrade::Optional bindings to make this a sensible and easy to compute default.

## How Has This Been Tested
* PyTest and locally

## Types of changes

## Checklist

### DIFF
--- a/examples/settings.py
+++ b/examples/settings.py
@@ -1,7 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-import magnum as mn
 
 import habitat_sim
 import habitat_sim.agent
@@ -153,10 +152,8 @@ def make_cfg(settings):
 
         fisheye_sensor_spec.resolution = [settings["height"], settings["width"]]
         # The default principal_point_offset is the middle of the image
-        fisheye_sensor_spec.principal_point_offset = mn.Vector2(
-            settings["height"] // 2,
-            settings["width"] // 2,
-        )
+        fisheye_sensor_spec.principal_point_offset = None
+        # default: fisheye_sensor_spec.principal_point_offset = [i/2 for i in fisheye_sensor_spec.resolution]
         fisheye_sensor_spec.position = [0, settings["sensor_height"], 0]
         for k in kw_args:
             setattr(fisheye_sensor_spec, k, kw_args[k])
--- a/src/esp/sensor/FisheyeSensor.cpp
+++ b/src/esp/sensor/FisheyeSensor.cpp
@@ -2,6 +2,7 @@
 // This source code is licensed under the MIT license found in the
 // LICENSE file in the root directory of this source tree.
 #include "FisheyeSensor.h"
+#include "Corrade/Containers/Containers.h"
 #include "esp/core/Check.h"
 #include "esp/gfx/DoubleSphereCameraShader.h"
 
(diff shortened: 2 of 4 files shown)


jturner65 (approved):
LGTM

src/esp/sensor/FisheyeSensor.cpp
```diff
@@ -48,6 +50,14 @@ void specSanityCheck(FisheyeSensorSpec* spec) {
   actualSpec->sanityCheck();
 }
 
+Magnum::Vector2 computePrincipalPointOffset(const FisheyeSensorSpec& spec) {
+  if (spec.principalPointOffset != Corrade::Containers::NullOpt) {
```
bigbike, line 54: I think you can just use: `Corrade::Containers::Optional<T>::operator bool() const explicit`
so if (spec.principalPointOffset)

src/esp/sensor/FisheyeSensor.cpp
```diff
@@ -48,6 +50,14 @@ void specSanityCheck(FisheyeSensorSpec* spec) {
...
+    return *spec.principalPointOffset;
+  }
+  auto res = spec.resolution;
+  return Mn::Vector2{Mn::Vector2i{res}} * 0.5f;
```
bigbike, line 58: `return Mn::Vector2{spec.resolution[0], spec.resolution[1]} * 0.5f;`

src/esp/sensor/FisheyeSensor.h
```diff
@@ -46,9 +46,11 @@ struct FisheyeSensorSpec : public CubeMapSensorBaseSpec {
   Magnum::Vector2 focalLength;
   /**
    * @brief Principal Point Offset in pixel, cx, cy, location of the principal
-   * point relative to the image plane's origin.
+   * point relative to the image plane's origin. None will place it in the
+   * middle of the image (height/2, width/2).
    */
-  Magnum::Vector2 principalPointOffset;
+  Corrade::Containers::Optional<Magnum::Vector2> principalPointOffset =
```
bigbike, line 52: The default for the `Optional` is `NullOpt` by the way.

bigbike (commented):
Check if you need to upgrade the viewer.cpp as well.

examples/settings.py
```diff
@@ -153,10 +152,8 @@ def create_fisheye_spec(**kw_args):
 
         fisheye_sensor_spec.resolution = [settings["height"], settings["width"]]
         # The default principal_point_offset is the middle of the image
-        fisheye_sensor_spec.principal_point_offset = mn.Vector2(
-            settings["height"] // 2,
-            settings["width"] // 2,
-        )
+        fisheye_sensor_spec.principal_point_offset = None
+        # default: fisheye_sensor_spec.principal_point_offset = [i/2 for i in fisheye_sensor_spec.resolution]
```
bigbike, line 156: typo: 1/2 not i/2

Skylion007: No, this intended: 
```
[i/2 for i in fisheye_sensor_spec.resolution]
``` 
is a valid list comprehension.

bigbike:
@Skylion007 : 
Last thing: check the viewer.cpp. 
Then you are good to merge it. :)

Thanks a lot!

PR #1294, opened 2021-06-04
[bugfix] update VHACD tutorial script

aclegg3:
## Motivation and Context

Apply recent API changes and add the tutorial to test_examples so future issues will be caught by CI.

## How Has This Been Tested

New CI coverage.

## Types of changes

## Checklist

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -57,6 +57,7 @@ def powerset(iterable):
             "--no-make-video",
         ),
         ("examples/tutorials/semantic_id_tutorial.py", "--no-show-images"),
+        ("examples/tutorials/VHACD_tutorial.py", "--no-show-video", "--no-make-video"),
     ],
 )
 def test_example_modules(args):
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -3,6 +3,7 @@ from os import path as osp
 
 import pytest
 
+import habitat_sim
 from utils import run_main_subproc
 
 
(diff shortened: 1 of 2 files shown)


tests/test_examples.py
```diff
@@ -57,6 +57,7 @@ def powerset(iterable):
             "--no-make-video",
         ),
         ("examples/tutorials/semantic_id_tutorial.py", "--no-show-images"),
+        ("examples/tutorials/VHACD_tutorial.py", "--no-show-video", "--no-make-video"),
```
Skylion007, line 60: Thanks for adding this. Will this error if vhacd is not enabled though? IF so we may need an additional skipif.

Skylion007 (approved):

tests/test_examples.py
aclegg3: Yeah, but CI should always hit this with --vhacd currently. 

Considering adding a binding like [this one for bullet in js](https://github.com/facebookresearch/habitat-sim/blob/f568b6564e59269e69f995ad54263b845dac1e22/src/esp/bindings_js/bindings_js.cpp#L127) as suggested by @eundersander.

Skylion007: @aclegg3 It already exists: https://github.com/facebookresearch/habitat-sim/blob/f568b6564e59269e69f995ad54263b845dac1e22/habitat_sim/__init__.py#L62

aclegg3: Thanks 👍

tests/test_examples.py
```diff
@@ -63,6 +64,16 @@ def test_example_modules(args):
     run_main_subproc(args)
 
 
+@pytest.mark.skipif(
+    not habitat_sim.vhacd_enabled,
+    reason="Requires Habitat-sim to be built with VHACD (--vhacd)",
+)
+def test_vhacd_example():
```
Skylion007, line 71: Nit can we fully paramterize with pytest.mark.parameterize so that we can add more VHACD tests easily in the future.

PR #1300, opened 2021-06-08
[AO migration] URDF Parser + Model

aclegg3:
## Motivation and Context

This change introduces the `esp::io::URDF` namespace containing a `Parser` for xml based URDF files (using tinyxml) and a `Model` struct (plus supporting datastrctures) to cache articulated object metadata before instancing. No physics simulation or Bullet implementation is included at this stage.

## How Has This Been Tested

Added a new C++ test `IOTest::parseURDF` with URDF parsing and validation of a test file. 

## Types of changes

## Checklist


src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,596 @@
...
+  float m_restitution{0};
+  float m_inertiaScaling{1};
+  float m_contactCfm{0};
+  float m_contactErp{0};
+  float m_contactStiffness{1e4};
+  float m_contactDamping{1};
+
+  int m_flags;
+
+  LinkContactInfo() { m_flags = CONTACT_HAS_LATERAL_FRICTION; }
```
Skylion007, line 227: Why not use just make this the default initialization? Then the struct would be trivially constructible?

src/esp/io/URDFParser.cpp
```diff
@@ -0,0 +1,1058 @@
...
+  sourceFilePath_ = filename;
+  urdfModel->m_sourceFile = filename;
+
+  std::string xmlString = Corrade::Utility::Directory::readString(filename);
+
+  XMLDocument xml_doc;
+  xml_doc.Parse(xmlString.c_str());
+  if (xml_doc.Error()) {
+    Mn::Debug{}
+        << "Parser::parseURDF - XML parse error, aborting URDF parse/load.";
```
eundersander, line 43: How about print the name of the file that failed to parse? And maybe this is Mn::Warn or Mn::Error, not Mn::Debug?

src/esp/io/URDFParser.cpp
```diff
@@ -0,0 +1,1058 @@
...
+  }
+
+  // Get all link elements including shapes
+  for (XMLElement* link_xml = robot_xml->FirstChildElement("link"); link_xml;
+       link_xml = link_xml->NextSiblingElement("link")) {
+    std::shared_ptr<Link> link = std::make_shared<Link>();
+
+    if (parseLink(urdfModel, *link, link_xml)) {
+      if (urdfModel->m_links.count(link->m_name) != 0u) {
+        Mn::Debug{} << "E - Link name is not unique, link names "
```
eundersander, line 85: Again, these error messages could specify the relevant file and also print to an error stream, not Mn::Debug.

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,596 @@
...
+  // scaling values which can be applied to the model after parsing
+  //! Global euclidean scaling applied to the model's transforms, asset scales,
+  //! and prismatic joint limits. Does not affect mass.
+  float m_globalScaling = 1.0;
+
+  //! Mass scaling of the model's Link inertias.
+  float m_massScaling = 1.0;
+
+  //! Scale the transformation and parameters of a Shape
+  void scaleShape(Shape& shape, float scale) {
```
eundersander, line 417: Should this be in the cpp, not the header?

Counterpoint: I can appreciate not changing this class too much, in case we ever want to diff against the version inside Bullet.

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,596 @@
...
+  std::string sourceFilePath_;
+
+  /**
+   * @brief Parse a transform into a Matrix4.
+   *
+   * @param tr The transform to fill.
+   * @param xml The source xml element to parse (e.g. <origin>).
+   * @return Success or failure.
+   */
+  bool parseTransform(Magnum::Matrix4& tr, tinyxml2::XMLElement* xml) const;
```
eundersander, line 455: Nit: it would make more sense to me if these xml elements in these function signatures were pointers-to-const instead of pointers-to-nonconst.

src/esp/io/URDFParser.h
```diff
@@ -0,0 +1,596 @@
...
+   * @return Filepath is valid or invalid.
+   */
+  bool validateMeshFile(std::string& filename);
+
+ public:
+  Parser() = default;
+
+  // parse a loaded URDF string into relevant general data structures
+  // return false if the string is not a valid urdf or other error causes abort
+  bool parseURDF(std::shared_ptr<Model>& model,
```
eundersander, line 585: Nit: kinda a strange interface to pass in a reference to a shared_ptr to be populated. How about just return a valid shared_ptr or null on failure?

eundersander (approved):
I know this class is adapted from Bullet code so I expect we don't want to modify it too much. Still, I added some comments.

src/esp/io/URDFParser.h
aclegg3: This one is my function encapsulating all the changes necessary to modify the scale after parsing. Sure we can move it into the cpp.

Skylion007 (approved):

PR #1304, opened 2021-06-09
[AO] pytest-benchmark integration

aclegg3:
## Motivation and Context

Add a `pytest-benchmark` integration for benchmarking AO functionality. 
Initial PoC function benchmarks URDF parse->instance->remove loop with and without `force_reload` caching option.

Also added `habitat_sim.built_with_bullet` flag to easily check if Habitat-sim was compiled with Bullet physics from python.

Note: requires local installation with `pip install pytest-benchmark`.

## How Has This Been Tested

Local testing and CI integration. Example output of `pytest tests/benchmark_physics.py`:
image

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,7 +129,7 @@ commands:
                 conda create -y -n habitat python=3.6
                 . activate habitat
                 conda install -q -y -c conda-forge ninja numpy pytest pytest-cov ccache hypothesis
-                pip install pytest-sugar pytest-xdist
+                pip install pytest-sugar pytest-xdist pytest-benchmark
               fi
       - run:
           name: Install emscripten
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -476,6 +476,9 @@ jobs:
               export PATH=$HOME/miniconda/bin:/usr/local/cuda/bin:$PATH
               . activate habitat; cd habitat-sim
               python examples/example.py --scene data/scene_datasets/habitat-test-scenes/van-gogh-room.glb --silent --test_fps_regression $FPS_THRESHOLD
+
+              #run the marked pytest-benchmark tests and print the results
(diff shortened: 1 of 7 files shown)


Skylion007 (approved):

Skylion007:
May want to add it here: https://github.com/facebookresearch/habitat-sim/blob/ea1288f8ffdfae3e1556ed86635277f7b755809a/setup.py#L423 and remove pytest-cov (we no-longer run it by default except on CI).

.circleci/config.yml
```diff
@@ -129,7 +129,7 @@ commands:
                 conda create -y -n habitat python=3.6
                 . activate habitat
                 conda install -q -y -c conda-forge ninja numpy pytest pytest-cov ccache hypothesis
-                pip install pytest-sugar pytest-xdist
+                pip install pytest-sugar pytest-xdist pytest-benchmark
```
Skylion007, line 132: This is neat, but does it actually run on CI? I think benchmarking is disabled when parallel xdist is run so if you want it to actually print out you need to select the benchmark tests and run them without parallelism.

Skylion007: You can do this easily with a pytest.mark decorator for benchmark tests.

aclegg3: Thanks for bringing this up. I haven't gone as far as to hook this up to mean something in CI. It would be ideal to have some cached results to compare against for automated regression testing, but printing the results would at least provide a standard environment for manual comparison.

Skylion007: @aclegg3 PyTest-Benchmark has the ability to do that, but that can wait for another PR to do that and just print out the stats similar to above.

eundersander (approved):
Cool!

aclegg3:
Closing in favor of master branch PR #1323

PR #1305, opened 2021-06-09
Update to Magnum with UBO and multidraw support

mosra:
## Motivation and Context

This is a huge (22kLOC) Magnum update with [uniform buffer, multidraw and texture array support](https://doc.magnum.graphics/magnum/shaders.html#shaders-usage-multidraw), which, when used the right way, can result in:

- **3x–10x** reduction in CPU overhead natively, depending on driver used
- up to **2x-3x** reduction in GPU time natively on some drivers (while *finally* not being worse even on Intel)
- **15x–20x** reduction in CPU overhead on WebGL

image

The UBO structure layout and multidraw design is prepared for multiview as well, which will land later. The data-oriented SceneData import will be a topic of another PR --  the main benefits of it will depend on quality of the input dataset, which will thus require some dataset pipeline additions. The documentation should be complete enough, and the accompanying example will appear in the documentation (and on the website) soon as well.

Switching Habitat to the multidraw workflow will be the initial step towards the multithreaded renderer work by Yili, so the above speedup isn't achieved just by merging this PR. Nevertheless, *some* speedup for the "classic" workflow is there:

- Various microoptimizations in the Phong shader resulting in **5%–10%** speed improvements depending on the number of lights used
- New [Flag::NoSpecular](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1PhongGL.html#a1e76742078247ee85732757e0e0bd23da4c3d2fb6c484fc5759ee1f125078bbdc) that speeds up the Phong shader by **10%–20%** in scenes that don't need specular highlights. I didn't enable it anywhere in Habitat tho, not sure where it would make sense.

## How Has This Been Tested

The new shaders got tested on Mesa Intel, Mesa AMD, NV Windows, Intel Windows, ARM Mali, macOS, ANGLE-on-GL, ANGLE-on-D3D and SwiftShader.

## Types of changes


bigbike (approved):
There is no code to be reviewed. So I will approve it directly. :)
I read the stats in the figures, which is amazing. 

Thank you.

p.s. I am thinking we really need an efficient way to measure the rendering performance boost in our end (the simulator) to justify  the impact of such awesome work. :)

aclegg3 (approved):

jturner65 (approved):
LGTM!

eundersander (approved):
LGTM

mosra:
Merging, if you see any measurable perf impact from just this PR alone (and not the `NoSpecular` flag), please let me know :)

PR #1306, opened 2021-06-10
[Sensor] reset the orientation of a cubemap

bigbike:
## Motivation and Context
This diff is to:
- reset the orientation of a cubemap so it aligns to the official definition found here:
https://www.khronos.org/opengl/wiki/Cubemap_Texture
(It is a must-have change as I will use the cubemap heavily in the PBR IBL)
- make changes in the fisheye shader and equirectangular shader accordingly;
- fix 2 places that caused the compiler warnings;

Cubemap before and after:
They are the same if the orientation is ignored.
But when you do consider the orientation, one is just transformed by rotating the other cube 180 degrees around the Y-axis.
The one after this fix is the "standard, correct" cubemap that we should use in the system.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/CubeMapCamera.cpp
+++ b/src/esp/gfx/CubeMapCamera.cpp
@@ -54,6 +54,12 @@ Mn::Matrix4 CubeMapCamera::getCameraLocalTransform(
   Mn::Vector3 eye{0.0, 0.0, 0.0};
   Mn::Vector3 yUp{0.0, 1.0, 0.0};
   Mn::Vector3 zUp{0.0, 0.0, 1.0};
+
+  // Careful: the coordinate system for cubemaps is left-handed.
+  // The following implementation is based on:
+  // https://www.khronos.org/opengl/wiki/Cubemap_Texture
+  // check the diagram which shows how each face in the cubemap is oriented
+  // relative to the cube being defined.
   switch (cubeSideIndex) {
     case Mn::GL::CubeMapCoordinate::PositiveX:
       return Mn::Matrix4::lookAt(eye, Mn::Vector3{1.0, 0.0, 0.0}, -yUp);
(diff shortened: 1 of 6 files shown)


src/shaders/doubleSphereCamera.frag
```diff
@@ -35,8 +35,6 @@ out highp uint fragmentObjectId;
 void main(void) {
   vec3 m;
   m.xy = (gl_FragCoord.xy - PrincipalPointOffset) / FocalLength;
-  // MUST flip the x axis
-  m.x = -m.x;
```
bigbike, line 39: Like I said, the difference between the "old and new" cubemaps is 180 degrees around Y-axis.
The orientation of the "new" cubemap is actually the "correct" one we should adopt. (+Z points inwards)

Before I directly flipped the x which does 2 things:
- convert the coordinates from right-handed to left handed.
- rotate the frame round Y-axis by 180 degrees (making the +Z from outwards to inwards direction)
That is why the results are consistent and correct "before and after" this change.

src/esp/gfx/CubeMapCamera.cpp
```diff
@@ -49,38 +49,47 @@ CubeMapCamera& CubeMapCamera::switchToFace(unsigned int cubeSideIndex) {
   return *this;
 }
 
-CubeMapCamera& CubeMapCamera::switchToFace(Mn::GL::CubeMapCoordinate cubeSide) {
+Mn::Matrix4 CubeMapCamera::getCameraLocalTransform(
```
bigbike, line 52: Create a new function to replace the internal lambda since it will be used in the future in the PBR IBL.

bigbike:
Results are consistent and correct before and after this change.

Before:
fisheye:
before-1
equirectangular:
before-2

After:
fisheye:
after-1
equirectangular:
after-3

Skylion007 (approved):

PR #1308, opened 2021-06-10
Cleanup and development of flag to enable/disable rendering

ldcWV:
## Motivation and Context

This PR is a continuation of Eric's PR here: #1258

Here is the todo list in the PR and how I addressed them:

> The hacky use of a global g_createMagnumRenderer var.

The flag is now stored in SimulatorConfiguration and passed around using getter methods. GenericDrawable also has a copy of the flag that is set in its constructor.

> Probably not all cases of graphics context use have been properly disabled/guarded, e.g. PbrDrawable.

Turns out the PbrDrawable codepath is impossible to reach when `createRenderer=false`. See https://github.com/facebookresearch/habitat-sim/pull/1308#discussion_r652931814

> Let's add a test of creating a non-renderer sim and using various APIs like drawObservation.

See SimTest.cpp.

> We should handle primitives correctly or at least fail gracefully. If we need to disable primitive-loading/use, we should issue a proper warning to users if they try to use a primitive that's unavailable.

This was referring to the skipping of the `ResourceManager::initDefaultPrimAttributes` method if the flag is off. The concern was that because of this, primitives would not be loaded correctly and thus should not be used. However, it appears that the only thing disabling this method affects is the visualization of bounding boxes, which doesn't matter when not using a renderer anyways.

> We need to reconcile with the existing SimulatorConfiguration::createRenderer. From discussion with @jturner65 , it sounds like the existing functionality of the existing flag (including disabling physics manager) isn't needed at all. So we instead hook up the createRenderer flag to the new functionality introduced in this PR (and discard the no-physics-manager code path altogether?).

The old createRenderer flag and its codepath have been deleted. The new flag has been renamed to createRenderer.

> Decide whether to create a stub Renderer object (discussed below).

Decided in meeting against this approach.

> reconcile two related features: disabling usage of OpenGL (this PR) and disabling loading of render assets (separate feature). After team discussion on 6/1, we agree this can be implemented separately. However, one suggestion is to have an enum for all the combinations of usage, instead of two independent flags. However, we all agreed that this PR doesn't need to implement the second feature, disabling loading of render assets (that can be a future PR).

Done (first feature). GL::Mesh and Shaders::FlatGL are no longer created when the flag is set.

## How Has This Been Tested

A test case demonstrating the functionality (what still works, what doesn't work) when the flag is disabled has been added in `SimTest.cpp`.

## Types of changes

## Checklist


examples/tutorials/nb_python/replay_tutorial.py
```diff
@@ -81,6 +81,7 @@ def make_configuration():
     # Enable gfx replay save. See also our call to sim.gfx_replay_manager.save_keyframe()
     # below.
     backend_cfg.enable_gfx_replay_save = True
+    backend_cfg.create_magnum_renderer = make_video_during_simulation
```
eundersander, line 84: You haven't added a Python binding for the flag so I don't think this is going to do anything. Python doesn't complain when you assign to non-existent members (unfortunate in this case).

See https://github.com/facebookresearch/habitat-sim/pull/1258/files#diff-08c325c91ce6aa20a6124d796c72d68934251a35f1cbc6e24421e5c23617de69 .

src/esp/assets/ResourceManager.cpp
```diff
@@ -123,7 +123,15 @@ void ResourceManager::buildImporters() {
 
 }  // buildImporters
 
+bool ResourceManager::getCreateRenderer() {
+  return metadataMediator_->getCreateRenderer();
+}
+
 void ResourceManager::initDefaultPrimAttributes() {
+  if (!getCreateRenderer()) {
```
eundersander, line 131: What is the consequence of this? For example, I would expect the `CHECK` in `addPrimitiveToDrawables` to fail. Are we providing good enough error-handling in case a user tries to set createRenderer to false and also tries to use some primitive? @jturner65

src/esp/gfx/Drawable.h
```diff
@@ -80,7 +80,7 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
   virtual void setLightSetup(
       CORRADE_UNUSED const Magnum::ResourceKey& lightSetup){};
 
-  Magnum::GL::Mesh& getMesh() { return mesh_; }
+  Magnum::GL::Mesh& getMesh() { return *mesh_; }  // todo: assert non-null
```
eundersander, line 83: Let's address this todo. We don't ever want to return a null reference.

src/esp/gfx/Drawable.h
```diff
@@ -91,7 +91,7 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
    * NOTE: sub-class should override this function if the "visualizer mesh" is
    * different from mesh_ (check the example in the PTexMeshDrawable class)
    */
-  virtual Magnum::GL::Mesh& getVisualizerMesh() { return mesh_; }
+  virtual Magnum::GL::Mesh& getVisualizerMesh() { return *mesh_; }
```
eundersander, line 94: Same here. We don't ever want to return a null reference.

src/esp/gfx/Drawable.h
```diff
@@ -110,7 +110,7 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
   uint64_t drawableId_;
 
   scene::SceneNode& node_;
-  Magnum::GL::Mesh& mesh_;
+  Magnum::GL::Mesh* mesh_ = nullptr;
```
eundersander, line 113: Can you post a reply here and explain to other reviewers why we did this change?

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -147,10 +150,14 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
     shader_->bindNormalTexture(*(materialData_->normalTexture));
   }
 
-  shader_->draw(mesh_);
+  shader_->draw(*mesh_);
```
eundersander, line 153: We should assert non-null before de-referencing here.

Maybe add a `getMesh()` in Drawable that does the assert (and make mesh_ private?). It makes me nervous that we're de-referencing this pointer in lots of places scattered throughout lots of files. Casual users won't understand that this pointer may be null for the obscure case of `createRenderer==false`.

src/esp/gfx/GenericDrawable.h
```diff
@@ -49,6 +50,8 @@ class GenericDrawable : public Drawable {
   Magnum::Resource<LightSetup> lightSetup_;
 
   Magnum::Shaders::PhongGL::Flags flags_;
+
+  bool createRenderer_;
```
eundersander, line 54: We can probably name this better. `isRendererCreated`?

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -156,6 +159,10 @@ Mn::ResourceKey PbrDrawable::getShaderKey(Mn::UnsignedInt lightCount,
 }
 
 PbrDrawable& PbrDrawable::updateShader() {
+  if (!createRenderer_) {
```
eundersander, line 162: How about a comment explaining what's going on here? Why are we early-out returning for this case?

src/esp/gfx/PbrDrawable.h
```diff
@@ -85,6 +86,8 @@ class PbrDrawable : public Drawable {
   Magnum::Resource<Magnum::GL::AbstractShaderProgram, PbrShader> shader_;
   Magnum::Resource<MaterialData, PbrMaterialData> materialData_;
   Magnum::Resource<LightSetup> lightSetup_;
+
+  bool createRenderer_;
```
eundersander, line 90: I guess this flag could live in the base Drawable class instead of repeated in several derived Drawable classes.

src/esp/sim/Simulator.h
```diff
@@ -99,8 +99,10 @@ class Simulator {
    * --headless mode on linux
    */
   int gpuDevice() const {
-    CORRADE_ASSERT(context_ != nullptr,
-                   "Simulator::gpuDevice: no OpenGL context.", 0);
+    if (!context_) {  // todo: be more strict and verbose here. test
```
eundersander, line 102: Let's resolve this todo. Simulator has access to the createRenderer flag, so it should know for sure whether we expect to have a context here. We should still be able to do some kind of assert.

src/tests/SimTest.cpp
```diff
@@ -797,6 +799,61 @@ void SimTest::addSensorToObject() {
...
+  simConfig.enablePhysics = true;
+  simConfig.physicsConfigFile = physicsConfigFile;
+  simConfig.overrideSceneLightDefaults = true;
+  simConfig.createRenderer = false;
+  simConfig.sceneLightSetup = esp::NO_LIGHT_KEY;
+
+  // check that creating a simulator and adding a primitive object works
+  auto simulator = Simulator::create_unique(simConfig);
+  auto objectAttribsMgr = simulator->getObjectAttributesManager();
+  int objectID = simulator->addObjectByHandle("cubeSolid");
```
eundersander, line 817: Can we also test loading a GLB? (you don't have to add another raycast, though)

src/tests/SimTest.cpp
```diff
@@ -797,6 +799,61 @@ void SimTest::addSensorToObject() {
...
+      esp::geo::Ray({10.0, 9.0, 10.0}, {0.0, 1.0, 0.0}), 100.0, 0);
+  CORRADE_VERIFY(raycastresults.hits[0].objectId == objectID);
+  auto point = raycastresults.hits[0].point;
+  CORRADE_VERIFY(distanceBetween(point, {10.0, 9.9, 10.0}) < 0.001);
+  raycastresults = simulator->castRay(
+      esp::geo::Ray({10.0, 11.0, 10.0}, {0.0, -1.0, 0.0}), 100.0, 0);
+  CORRADE_VERIFY(raycastresults.hits[0].objectId == objectID);
+  point = raycastresults.hits[0].point;
+  CORRADE_VERIFY(distanceBetween(point, {10.0, 10.1, 10.0}) < 0.001);
+
```
eundersander, line 837: Can we also test bounding-box functionality?

eundersander (commented):
Good progress here. I left some comments.

src/tests/SimTest.cpp
eundersander, line 817: Can you confirm whether adding this object tests the PbrDrawable codepath? I guess you could verify this via the VS code debugger or by adding a print statement locally.

src/esp/assets/ResourceManager.cpp
ldcWV: It seems like `ResourceManager::initDefaultPrimAttributes` is named like this because it used to be necessary for loading in all primitives. But now only `cubeWireframe` is referenced here because it is being used to draw bounding boxes, which somehow makes it a special case. Skipping this method causes bounding boxes to not show up in the viewer, but this shouldn't matter if we're not rendering anything anyways.

IIRC when I tried adding `cubeWireframe` and raycasting at it, the results were still fine. But I can add that part of the test back in.

I don't understand this very well, so I agree that some additional insight from @jturner65 would be helpful here.

src/esp/gfx/Drawable.h
ldcWV: The `createRenderer` flag prevents the function `GenericMeshData::uploadBuffersToGPU` from running, which is the function that creates `GL::Mesh`. This means that the `GenericMeshData` that are inserted into `ResourceManager::meshes_` will not have set `GL::Mesh` fields.

However, `getMagnumGLMesh` is still being called on elements of `meshes_`. These calls will return NULL because no `GL::Mesh` objects were ever created. Yet we are still trying to do things with the returned values, which is unsafe. 

Therefore, we instead store pointers to the `GL::Mesh` objects so that we must explicitly dereference them, which requires first checking that they are not `nullptr`. This is a cleaner way of handling errors.

src/esp/sim/Simulator.h
ldcWV: Pushed something to address this; let me know what you think!

examples/tutorials/nb_python/replay_tutorial.py
```diff
@@ -245,14 +247,14 @@ def simulate_with_moving_agent(
...
-    get_frames=make_video,
+    get_frames=make_video_during_simulation,
 )
 
 # %% [markdown]
 # ## End the episode. Render the episode observations to a video.
 # %%
 
-if make_video:
+if make_video_during_simulation:
```
Skylion007, line 257: nit: that's an annoying long variable name. Even shortening to make_video_during_sim would help

Skylion007 (commented):
@bigbike Should be the one to look at this considering how it intertwines with the GFX system.

src/esp/physics/bullet/BulletPhysicsManager.h
```diff
@@ -56,6 +56,11 @@ class BulletPhysicsManager : public PhysicsManager {
       : PhysicsManager(_resourceManager, _physicsManagerAttributes) {
     collisionObjToObjIds_ =
         std::make_shared<std::map<const btCollisionObject*, int>>();
+    if (_resourceManager.getCreateRenderer()) {
+      debugDrawer_ = std::make_unique<Magnum::BulletIntegration::DebugDraw>();
+    } else {
+      LOG(INFO) << "WTF";
```
eundersander, line 62: ^

src/tests/SimTest.cpp
ldcWV: This does not; currently it seems like making `createMagnumRenderer=true` is the difference between `PbrDrawable` being tested or not, which defeats the purpose. I'll keep looking to see if there is a way for the codepath to be executed when the flag is set to `false`. If it turns out not to be possible, I'll remove the changes in `PbrDrawable.cpp`.

src/esp/physics/bullet/BulletPhysicsManager.h
ldcWV: OOPS...

src/tests/SimTest.cpp
ldcWV: Yep, I did some backtracking and turns out it's impossible to trigger the PbrDrawable codepath when `createRenderer=false`.

This is because in `ResourceManager.cpp`, the only time a Pbr material can possibly be added to `shaderManager_` is for `finalMaterial` in `loadMaterials`.

However, `loadMaterials` cannot possibly run because it requires `requiresTextures_=true`, but in `Simulator.cpp` we have
```
if (!config_.createRenderer) {
  config_.requiresTextures = false;
}
```

eundersander (approved):
LGTM. Let's get at least one more approval from someone else before merging.

habitat_sim/simulator.py
```diff
@@ -529,6 +530,8 @@ def __init__(self, sim: Simulator, agent: Agent, sensor_id: str) -> None:
         )
 
     def draw_observation(self) -> None:
+        assert self._sim.renderer
```
Skylion007, line 533:
```suggestion
        assert self._sim.renderer is not None
```

habitat_sim/simulator.py
```diff
@@ -542,6 +545,7 @@ def draw_observation(self) -> None:
         self._sim.renderer.draw(self._sensor_object, self._sim)
 
     def get_observation(self) -> Union[ndarray, "Tensor"]:
+        assert self._sim.renderer
```
Skylion007, line 548:
```suggestion
        assert self._sim.renderer is not None
```

src/esp/assets/ResourceManager.cpp
jturner65: Initially all primitives (bounding boxes and navmesh) were managed ad hoc in ResourceManager.  When I added support for primitive objects, I made this function to still support bounding boxes using the old mechanism without necessitating a complete redesign of that process.  

I would suggest disposing of the function completely and using the AssetAttributesManager to provide the template for the bounding box, as is stated in the comment below this block. :)  Barring that, it should be fine to avoid this function if rendered bounding boxes are not required.

jturner65 (approved):
A minor request, but otherwise LGTM.  Once all the tests pass, this should be good to go.

Skylion007 (changes requested):
Please resolve all the merge conflicts and style issues. `pre-commit run -a` should solve many of them.

src/esp/assets/ResourceManager.cpp
```diff
@@ -123,7 +123,15 @@ void ResourceManager::buildImporters() {
 
 }  // buildImporters
 
+bool ResourceManager::getCreateRenderer() {
```
erikwijmans, line 126:
```suggestion
bool ResourceManager::getCreateRenderer() const {
```

src/esp/assets/ResourceManager.h
```diff
@@ -83,6 +83,8 @@ namespace assets {
  */
 class ResourceManager {
  public:
+  bool getCreateRenderer();
```
erikwijmans, line 86:
```suggestion
  bool getCreateRenderer() const;
```

src/esp/metadata/MetadataMediator.cpp
```diff
@@ -386,5 +386,9 @@ attributes::SceneAttributes::ptr MetadataMediator::makeSceneAndReferenceStage(
   return sceneAttributes;
 }  // MetadataMediator::makeSceneAndReferenceStage
 
+bool MetadataMediator::getCreateRenderer() {
```
erikwijmans, line 389:
```suggestion
bool MetadataMediator::getCreateRenderer() const {
```

src/esp/metadata/MetadataMediator.h
```diff
@@ -343,6 +343,8 @@ class MetadataMediator {
         sceneDatasetName);
   }
 
+  bool getCreateRenderer();
```
erikwijmans, line 346:
```suggestion
  bool getCreateRenderer() const;
```

src/esp/gfx/Drawable.h
```diff
@@ -80,7 +80,10 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
   virtual void setLightSetup(
       CORRADE_UNUSED const Magnum::ResourceKey& lightSetup){};
 
-  Magnum::GL::Mesh& getMesh() { return mesh_; }
+  Magnum::GL::Mesh& getMesh() {
+    ASSERT(mesh_ != nullptr);
```
erikwijmans, line 84:
```suggestion
     CORRADE_INTERNAL_ASSERT(mesh_ != nullptr);
```

src/esp/gfx/Drawable.h
```diff
@@ -91,7 +94,10 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
    * NOTE: sub-class should override this function if the "visualizer mesh" is
    * different from mesh_ (check the example in the PTexMeshDrawable class)
    */
-  virtual Magnum::GL::Mesh& getVisualizerMesh() { return mesh_; }
+  virtual Magnum::GL::Mesh& getVisualizerMesh() {
+    ASSERT(mesh_ != nullptr);
```
erikwijmans, line 98:
```suggestion
     CORRADE_INTERNAL_ASSERT(mesh_ != nullptr);
```

erikwijmans (commented):
`const` is what the CI is complaining about.  Also, use one of Corrade's asserts or ESP_CHECK, not ASSERT

examples/tutorials/nb_python/replay_tutorial.py
eundersander: Hmmm, I still don't see a Python binding so I feel like this is still doing nothing. Can you add the binding and test that the flag is working with python scripts?

eundersander: We should add a Python unit test. You can test `sim.renderer is None` or similar.

eundersander: Oh wait, I forgot. This is just re-using the existing flag that already has a binding. Up to you if you want to bother with a Python unit test.

habitat_sim/simulator.py
```diff
@@ -475,7 +475,8 @@ def __init__(self, sim: Simulator, agent: Agent, sensor_id: str) -> None:
 
         self._spec = self._sensor_object.specification()
 
-        self._sim.renderer.bind_render_target(self._sensor_object)
+        if self._sim.renderer:
```
Skylion007, line 478: Should this also be is not None. If so better than implicit boolean conversion.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2056,7 +2080,12 @@ void ResourceManager::addComponent(
   // Add a drawable if the object has a mesh and the mesh is loaded
   if (meshIDLocal != ID_UNDEFINED) {
     const int meshID = metaData.meshIndex.first + meshIDLocal;
-    Magnum::GL::Mesh& mesh = *meshes_.at(meshID)->getMagnumGLMesh();
+    Magnum::GL::Mesh* mesh = meshes_.at(meshID)->getMagnumGLMesh();
+    if (getCreateRenderer()) {
+      CORRADE_INTERNAL_ASSERT(mesh);
+    } else {
+      CORRADE_INTERNAL_ASSERT(!mesh);
```
bigbike, line 2087: Use CORRADE_ASSERT instead in order to output the error message.
e.g., 
"ResourceManager::addComponent(): Simulator is not created with createRenderer is enabled. Check your simulator config."

It will significantly save developer's debugging time.

src/esp/gfx/Drawable.h
```diff
@@ -94,14 +94,16 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
...
-
   /**
    * @brief the the scene node
    */
   virtual scene::SceneNode& getSceneNode() const { return node_; }
 
   /** @brief get the GL mesh */
-  Magnum::GL::Mesh& getMesh() const { return mesh_; }
+  Magnum::GL::Mesh& getMesh() const {
+    CORRADE_INTERNAL_ASSERT(mesh_ != nullptr);
```
bigbike, line 104: Same logic:
Use CORRADE_ASSERT instead.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -147,10 +150,16 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
     shader_->bindNormalTexture(*(materialData_->normalTexture));
   }
 
-  shader_->draw(mesh_);
+  shader_->draw(getMesh());
```
bigbike, line 153: I looked into the `getMesh()` which has a sanity check there. Great.
However, better solution is to add such sanity check at the beginning of this function, and make it fail even faster (without even setting the shader, because it does not make any sense.)

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -147,10 +150,16 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
...
 }
 
 void GenericDrawable::updateShader() {
+  // Block the function if there is no renderer. Otherwise this will segfault.
+  // We don't have use for a shader anyways if we are not rendering anything.
+  if (!isRendererCreated) {
```
bigbike, line 159: Not just silently return, just let if fail. For example, CORRADE_INTERNAL_ASSERT. 

On big, critical functions, such as draw(...) I will put CORRADE_ASSERT there and for small, helper functions, I would put CORRADE_INTERNAL_ASSERT

src/esp/gfx/GenericDrawable.h
```diff
@@ -49,6 +50,8 @@ class GenericDrawable : public Drawable {
   Magnum::Resource<LightSetup> lightSetup_;
 
   Magnum::Shaders::PhongGL::Flags flags_;
+
+  bool isRendererCreated;
```
bigbike, line 54: Here is what I think:
When renderer is NOT created, we do not need to create drawables. In other words, we will create them when the renderer is created as usual. Then why shall we change all the drawables? Is it because this is a temporary solution, and it is the easiest way to hack it in such a manner?

src/esp/gfx/MeshVisualizerDrawable.cpp
```diff
@@ -27,7 +27,7 @@ void MeshVisualizerDrawable::draw(const Magnum::Matrix4& transformationMatrix,
   shader_.setProjectionMatrix(camera.projectionMatrix())
       .setTransformationMatrix(transformationMatrix);
 
-  shader_.draw(mesh_);
+  shader_.draw(getMesh());
```
bigbike, line 30: Fail fast, same logic for every type of drawable. See my previous comments.

src/esp/sim/Simulator.cpp
```diff
@@ -143,28 +145,23 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
         flags |= gfx::Renderer::Flag::NoTextures;
       renderer_ = gfx::Renderer::create(flags);
     }
-
-    // (re) create scene instance
-    success = createSceneInstance(config_.activeSceneName);
   } else {
-    // (re) create scene instance without renderer
-    success = createSceneInstanceNoRenderer(config_.activeSceneName);
+    CORRADE_INTERNAL_ASSERT(!Magnum::GL::Context::hasCurrent());
```
bigbike, line 149: CORRADE_ASSERT to output an error message:
"Simulator::reconfigure(): ..."

bigbike (changes requested):

src/esp/assets/ResourceManager.cpp
eundersander: I think you misunderstand the assert. For the else case, this assert is verifying that we haven't created a GL mesh, as expected. The code elsewhere that avoids calling `uploadBuffersToGPU` ensures this. A user isn't going to hit this assert. This is an internal assert.

With that said, there's no harm in changing both of these to `CORRADE_ASSERT` and adding some explanation text. For the second one, it would be something like "encountered unexpected GL mesh with createRenderer==false".

src/esp/gfx/GenericDrawable.h
eundersander: We're creating a Drawable because this flag is meant to be minimally-invasive. The fewer lines of code this PR changes, the easier it will be to ensure this flag's correctness and maintain it.

If we disabled creating a Drawable, we'd have to make lots of other changes in the code, for example, places where we call `Drawable::setLightSetup`.

This flag is not meant to aggressively avoid loading as much data as possible for the case of no rendering. We expect that will be a separate flag, something like "loadRenderAssets".

bigbike: Make sense. You explanation aligned with my guess. Thank you.

src/esp/gfx/GenericDrawable.cpp
ldcWV: For the "sanity check" here, I blocked `draw` when `isRendererCreated==false` instead of having an assert. This goes along with Eric's comments about minimal invasiveness. Otherwise I would have to block every draw call elsewhere in the code.

src/esp/gfx/MeshVisualizerDrawable.cpp
ldcWV: Only the `GenericDrawable` codepath is possible when `createRenderer==true`. Therefore, `getMesh()` should definitely always succeed when it is called from here. So I don't think it's necessary to add some fail fast code here.

src/esp/gfx/GenericDrawable.cpp
bigbike: My comments will not require you to do massive refactoring. So it aligns with your design rules -- minimal changes. 
>I would have to block every draw call elsewhere in the code.

You misunderstood "the beginning of this function". I did not ask you to add sanity check everywhere when you call this function. 
I am asking you to add sanity check at the start of this function definition. 
(Line 119 before updateShader()).

src/esp/gfx/MeshVisualizerDrawable.cpp
bigbike: See my previous comments. 

>Only the GenericDrawable codepath is possible when createRenderer==true.

A Typo? createRenderer!=true ?

Please read my previous comments first (about sanity check at the beginning of the draw function).

You need the same sanity check here simply because you modified the original code here even you understand this code path will never be reached. This is because other developers do not know this fact. If they do refactoring in the future, they can easily introduce bugs into the system.

src/esp/gfx/GenericDrawable.cpp
ldcWV: Sorry, I should've been more clear. I meant that I added
```if (!isRendererCreated) {
    return;
  }```
before `updateShader()`, rather than an assert statement (which is what I think you meant by sanity check). I have not pushed this change yet.

File: src/esp/gfx/MeshVisualizerDrawable.cpp (hunk 29 above)
Comment (ldcWV) [reply src/esp/gfx/MeshVisualizerDrawable.cpp:30]: Oops yeah, that was a typo. Should be `createRenderer==false`.

This makes sense, I'll add in some earlier checks that `mesh_` exists. Thanks for the feedback!

Comment (ldcWV) [reply src/esp/gfx/MeshVisualizerDrawable.cpp:30]: It's a bit inelegant because `mesh_` is private in order to prevent unsafe behavior since it's no longer a pointer. So the only possible way to sanity check would be to try running `getMesh()` and seeing if it errors.

**Review by Skylion007 — APPROVED** [review]

File: src/esp/assets/ResourceManager.cpp [hunk 31]
```diff
@@ -1237,7 +1251,10 @@ bool ResourceManager::loadRenderAssetGeneral(const AssetInfo& info) {
   assimpmetadata->configuration().setValue("ImportColladaIgnoreUpDirection",
                                            "true");
 #endif
-  {
+
+  if (!requiresTextures_) {
```
Skylion007, line 1255: Would it just be cleaner to remove the else and rephrase the comment here?

mosra:
FYI the `GL::Mesh` can be constructed as `GL::Mesh{NoCreate}` without a GL context present. So you could keep passing `GL::Mesh&` around, which could reduce the amount of changes and additional logic in this PR quite a bit. 

Same is for textures (`GL::Texture2D{NoCreate}` won't need a GL context to exist).

(Apologies if this suggestion came too late, was too busy to check PR reviews regularly.)

eundersander (approved):
LGTM. Let's get the unit tests passing and get this merged.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -113,6 +118,12 @@ void GenericDrawable::updateShaderLightingParameters(
 
 void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
                            Mn::SceneGraph::Camera3D& camera) {
+  if (!isRendererCreated) {
+    return;
+  }
+  // Check that getMesh() doesn't error, i.e. that a mesh exists.
+  getMesh();
```
bigbike, line 125: No, you do not need to call this function here (I know you have sanity check within, but do not call it simply because you would like to do the santity check.)
You should use "CORRADE_ASSERT" directly.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -113,6 +118,12 @@ void GenericDrawable::updateShaderLightingParameters(
 
 void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
                            Mn::SceneGraph::Camera3D& camera) {
+  if (!isRendererCreated) {
```
bigbike, line 121: Remove this if clause. Instead use CORRADE_ASSERT here to let it fail fast.

The if clause you put here just let the function return silently, without any further info given back to the developer.

Developer may just get a black screen and spend even more time to debug.

bigbike (commented):
Please fix all the places that I mentioned, before committing it in.

eundersander:
>FYI the GL::Mesh can be constructed as GL::Mesh{NoCreate} without a GL context present. 
As you say, this suggestion is coming a bit late. :) For this PR, we removed GL::Mesh creation as you originally requested in our Slack discussions. This PR has been under review for a while and we'd like to get it merged. Let's do the switch to `GL::Mesh{NoCreate}` as a separate refactor PR.

bigbike:
> FYI the GL::Mesh can be constructed as GL::Mesh{NoCreate} without a GL context present. 
@mosra : Great. Good to know. However, I guess one reason they changed them to pointer is that they can easily tell if the simulator is initialized with the renderer (just checking if the pointer is NULL).
In the way you suggested, how can this be done?

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -118,11 +116,8 @@ void GenericDrawable::updateShaderLightingParameters(
 
 void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
                            Mn::SceneGraph::Camera3D& camera) {
-  if (!isRendererCreated) {
```
ldcWV: It turns out that this block is not necessary because the `GenericDrawable::draw` codepath is impossible to reach when `createRenderer==false`.

src/esp/assets/ResourceManager.cpp
ldcWV: Do you mean to move everything into `if(requiresTextures_)`?

src/esp/gfx/GenericDrawable.h
```diff
@@ -19,12 +19,13 @@ class GenericDrawable : public Drawable {
...
   explicit GenericDrawable(scene::SceneNode& node,
-                           Magnum::GL::Mesh& mesh,
+                           Magnum::GL::Mesh* mesh,
                            Drawable::Flags& meshAttributeFlags,
                            ShaderManager& shaderManager,
                            const Magnum::ResourceKey& lightSetupKey,
                            const Magnum::ResourceKey& materialDataKey,
-                           DrawableGroup* group = nullptr);
+                           DrawableGroup* group = nullptr,
+                           bool createRenderer = true);
```
bigbike, line 28: No, you do not need an extra boolean here. See my comments above.

src/esp/gfx/MeshVisualizerDrawable.cpp
bigbike: I saw the "CORRADE_ASSERT(glMeshExists(), ..." on the top of this function. It is great.

src/esp/sim/Simulator.cpp
```diff
@@ -143,28 +145,25 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
...
+                   "createRenderer==false", );
   }
 
+  // (re) create scene instance
+  success = createSceneInstance(config_.activeSceneName);
+
   LOG(INFO) << "::reconfigure : createSceneInstance success == "
             << (success ? "true" : "false")
-            << " for active scene name : " << config_.activeSceneName
-            << (config_.createRenderer ? " with" : " without") << " renderer.";
```
bigbike, line 157: Why is this info removed from the output?

src/esp/gfx/Drawable.h
```diff
@@ -94,14 +94,17 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
...
   /**
    * @brief the the scene node
    */
   virtual scene::SceneNode& getSceneNode() const { return node_; }
 
   /** @brief get the GL mesh */
-  Magnum::GL::Mesh& getMesh() const { return mesh_; }
+  Magnum::GL::Mesh& getMesh() const {
+    CORRADE_ASSERT(mesh_ != nullptr,
+                   "Attempting to get the GL mesh when none exists", *mesh_);
```
bigbike, line 105: You are missing the class name and function name in the error message.
"Drawable::getMesh(): ..."

src/esp/assets/ResourceManager.cpp
```diff
@@ -1508,6 +1527,9 @@ int ResourceManager::loadNavMeshVisualization(esp::nav::PathFinder& pathFinder,
   if (!pathFinder.isLoaded())
     return navMeshPrimitiveID;
 
+  if (!getCreateRenderer())
+    return navMeshPrimitiveID;
```
bigbike, line 1531: should have "{}" even it is one line of code within the if clause based on our code convention.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1508,6 +1527,9 @@ int ResourceManager::loadNavMeshVisualization(esp::nav::PathFinder& pathFinder,
   if (!pathFinder.isLoaded())
     return navMeshPrimitiveID;
```
bigbike, line 1528: should have "{}" even it is one line of code within the if clause based on our code convention.
Please help us to fix it.

src/esp/assets/ResourceManager.cpp
bigbike: I think you understand @Skylion007  correctly. 
Would you change it?

src/tests/PhysicsTest.cpp
```diff
@@ -41,9 +41,10 @@ class PhysicsManagerTest : public testing::Test {
   void SetUp() override {
     // set up a default simulation config to initialize MM
     auto cfg = esp::sim::SimulatorConfiguration{};
+    cfg.createRenderer = false;
     metadataMediator_ = MetadataMediator::create(cfg);
     resourceManager_ = std::make_unique<ResourceManager>(metadataMediator_);
-    context_ = esp::gfx::WindowlessContext::create_unique(0);
```
bigbike, line 46: Wait, if you change this, then we lose the protection from the unit test when the context needs to be created. 

In your case, you need to create a brand new test function to test when the Renderer is not created.

src/tests/PhysicsTest.cpp
```diff
@@ -97,9 +98,6 @@ class PhysicsManagerTest : public testing::Test {
     return objectWrapper;
   }
 
-  // must declare these in this order due to avoid deallocation errors
-  esp::gfx::WindowlessContext::uptr context_;
```
bigbike, line 101: See my previous comment.

bigbike (changes requested):

src/esp/assets/ResourceManager.cpp
```diff
@@ -2056,7 +2080,14 @@ void ResourceManager::addComponent(
   // Add a drawable if the object has a mesh and the mesh is loaded
   if (meshIDLocal != ID_UNDEFINED) {
     const int meshID = metaData.meshIndex.first + meshIDLocal;
-    Magnum::GL::Mesh& mesh = *meshes_.at(meshID)->getMagnumGLMesh();
+    Magnum::GL::Mesh* mesh = meshes_.at(meshID)->getMagnumGLMesh();
+    if (getCreateRenderer()) {
+      CORRADE_ASSERT(mesh, "::addComponent : GL mesh expected but not found", );
```
bigbike, line 2085: "::addComponent :" is incomplete. 
You need:
"ClassName::functionName()"
Please fix this problem in this PR (a couple of places needs to be fixed.)

src/esp/assets/ResourceManager.cpp
```diff
@@ -2056,7 +2080,14 @@ void ResourceManager::addComponent(
...
+    } else {
+      CORRADE_ASSERT(!mesh,
+                     "addComponent : encountered unexpected GL mesh with "
```
bigbike, line 2088: fix this as well.

src/esp/gfx/MeshVisualizerDrawable.cpp
```diff
@@ -16,18 +16,21 @@ MeshVisualizerDrawable::MeshVisualizerDrawable(
...
     Magnum::GL::Mesh& mesh,
     DrawableGroup* group)
-    : Drawable{node, mesh, DrawableType::MeshVisualizer, group},
+    : Drawable{node, &mesh, DrawableType::MeshVisualizer, group},
       shader_(shader) {}
 
 void MeshVisualizerDrawable::draw(const Magnum::Matrix4& transformationMatrix,
                                   Magnum::SceneGraph::Camera3D& camera) {
+  CORRADE_ASSERT(glMeshExists(),
+                 "MeshVisualizerDrawable::draw : GL mesh doesn't exist", );
```
bigbike, line 25: MeshVisualizerDrawable::draw()

src/esp/gfx/PTexMeshDrawable.cpp
```diff
@@ -44,6 +44,9 @@ PTexMeshDrawable::PTexMeshDrawable(scene::SceneNode& node,
 
 void PTexMeshDrawable::draw(const Magnum::Matrix4& transformationMatrix,
                             Magnum::SceneGraph::Camera3D& camera) {
+  CORRADE_ASSERT(glMeshExists(),
+                 "PTexMeshDrawable::draw : GL mesh doesn't exist", );
```
bigbike, line 48: PTexMeshDrawable::draw():

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -80,6 +80,8 @@ void PbrDrawable::setLightSetup(const Mn::ResourceKey& lightSetupKey) {
 
 void PbrDrawable::draw(const Mn::Matrix4& transformationMatrix,
                        Mn::SceneGraph::Camera3D& camera) {
+  CORRADE_ASSERT(glMeshExists(), "PbrDrawable::draw : GL mesh doesn't exist", );
```
bigbike, line 83: PbrDrawable::draw()

src/esp/gfx/MeshVisualizerDrawable.cpp
```diff
@@ -21,6 +21,9 @@ MeshVisualizerDrawable::MeshVisualizerDrawable(
 
 void MeshVisualizerDrawable::draw(const Magnum::Matrix4& transformationMatrix,
                                   Magnum::SceneGraph::Camera3D& camera) {
+  // Check that getMesh() doesn't error, i.e. that a mesh exists.
+  getMesh();
```
bigbike: Just put it here for the record:
You should not call this function here simply because it contains a sanity check within.
Instead, you can create a function called e.g., `glMeshExists()` which would tell you if the mesh is nullptr or not.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -17,12 +17,13 @@ namespace esp {
...
 GenericDrawable::GenericDrawable(scene::SceneNode& node,
-                                 Mn::GL::Mesh& mesh,
+                                 Mn::GL::Mesh* mesh,
                                  Drawable::Flags& meshAttributeFlags,
                                  ShaderManager& shaderManager,
                                  const Mn::ResourceKey& lightSetupKey,
                                  const Mn::ResourceKey& materialDataKey,
-                                 DrawableGroup* group /* = nullptr */)
+                                 DrawableGroup* group /* = nullptr */,
+                                 bool isRendererCreated /* = true */)
```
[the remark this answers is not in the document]
ldcWV: Yeah, that would be its only purpose. I agree this way is cleaner.

src/esp/sim/Simulator.cpp
ldcWV: Oops, I think this was an accident from resolving merge conflicts. I'll add it back in.

src/tests/PhysicsTest.cpp
ldcWV: Doing this shows that you don't need a context in order to do physics when `createRenderer==false`, which is good to know. Maybe I can modify the tests to try both with and without context?

I already added a non-physics test with `createRenderer=false` in SimTest.cpp by the way.

eundersander: Sorry for late reply here. I want to see this change stay as-is.

I actually love that we're now guaranteed not to use OpenGL in our Physics unit tests. Conceptually, physics tests shouldn't rely on rendering. Let's imagine a future day where we've completely decoupled physics and rendering code--it'll be nice that these tests just port over directly, without having to be fixed up. By never creating a context here, we're guaranteed that our physics test code is "clean"--not accidentally relying on some rendering feature.

>we lose the protection from the unit test when the context needs to be created.
If you have in mind a unit test that involves physics and rendering, the best place for that is SimTest. Simulator is where we integrate physics and rendering. You can add objects, make them interact with physics, then call draw() or whatever and check the result.

src/esp/sim/Simulator.h
```diff
@@ -101,7 +101,12 @@ class Simulator {
    * --headless mode on linux
    */
   int gpuDevice() const {
-    CORRADE_ASSERT(context_ != nullptr, "::gpuDevice: no OpenGL context.", 0);
+    if (context_ == nullptr) {
```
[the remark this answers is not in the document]
eundersander: FWIW, I found it useful that `gpuDevice` can silently return 0 and do nothing for the case of `createRenderer==false`. I encountered this when running some of our python programs with `createRenderer=false`.

However, if Yili is adamant here, we can make his change as suggested. The result is just that python users who try `createRenderer==false` will have to also guard/avoid their calls to `gpuDevice`.

Either is fine with me.

src/tests/PhysicsTest.cpp
eundersander: Yep, let's resolve the discussion about `context_` here: https://github.com/facebookresearch/habitat-sim/pull/1308/files#r658137220

src/tests/PhysicsTest.cpp
bigbike: This has nothing to do with the visualization verification. We do that in the SimTest. No problem.

My opinion is that by doing both (with or without renderer) in the physicsTest, we have solid proof that physics simulation and the rendering are totally orthogonal to each other. Both tests should pass.

bigbike: Do we have a test in the SimTest that creates a GL context and enabled physics? No, we do not have.
We used to count on this physicsTest here until this change.

src/esp/sim/Simulator.h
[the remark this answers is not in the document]
bigbike: > I found it useful that gpuDevice can silently return 0 and do nothing for the case of createRenderer==false.

Good to know. May I ask in which case you find it useful when it returns silently?
I was thinking assertion (again "fail fast") would be more useful because user would immediately know she made a mistake in the code.

src/tests/PhysicsTest.cpp
eundersander: >My opinion is that by doing both (with or without gl context) in the physicsTest, we have solid proof that physics simulation and the rendering are totally orthogonal to each other. Both tests should pass.
Ok, fair enough. We'll try to find a way to run both versions of the physics tests, hopefully without duplicating a lot of code.

bigbike: Yeah, I hear you. 

Too bad we are using the google test, not the corrade test (used in the SimTest).
Corrade test supports instanced tests. See [here](https://github.com/facebookresearch/habitat-sim/blob/master/src/tests/SimTest.cpp#L157).

This diff is to add a new feature which can enable/disable the rendering. I saw Lawrence added tests in the SimTest. So I assume this feature is well covered. 

We also know the current solution is a temporary solution as we will refactor this code in the near future. So I am totally fine if we do not touch the PhysicsTest.cpp at all in this PR. I think it is the easiest way to unblock it.

Are you comfortable with this solution?

ldcWV: I've made a commit to run PhysicsTest both with and without the flag set.

bigbike (approved):

aclegg3 (approved):
LGTM.

PR #1311, opened 2021-06-12
[Bugfix]Shader type config value "material" option

jturner65:
## Motivation and Context
This small PR introduces an accepted value of "material" for the "shader_type" field within the various configuration files describing objects and stages.  This value will override any previously set shader_type values for the particular asset to instead use whatever shader_type is specified by the materials within in asset's file.  This is necessary because there is a hierarchy of configuration specification in place, and without this option, the user would be unable to override any shader type specification set at a lower level configuration.  

For instance, a user may specify that the desired shader_type for all bananas (in their copy of banana.object_config.json) should be Flat.  However, within a certain scene, they may wish for a particular instance of banana to be rendered using the shader suggested by its own materials.  They can use the `"shader_type" : "material"` specification within the object instance describing the desired banana's instantiation in the scene instance config describing the scene in question, and this will override the default flat shader choice, but only for that one banana.

This is a bugfix because without this additional choice, undoing a config-specified shader type in the above manner (for a single instance, for example) would be impossible.

EDIT : This PR also reverts the change from [this PR](https://github.com/facebookresearch/habitat-sim/pull/1307) that specified that shader_type default for stages be flat. This PR changes this so that the default is Unknown, which means the stage material dictates its shader type.  

## How Has This Been Tested
All existing c++ and python tests pass.  Simple tests were added to verify specification works correctly for both template config level and scene object instance config level

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/MetadataUtils.cpp
+++ b/src/esp/metadata/MetadataUtils.cpp
@@ -31,12 +31,12 @@ int getShaderTypeFromJsonDoc(const io::JsonGenericValue& jsonDoc) {
         attributes::AbstractObjectAttributes::ShaderTypeNamesMap.end()) {
       shader_type = static_cast<int>(found->second);
     } else {
-      LOG(WARNING) << "getShaderTypeFromJsonDoc : "
-                      "motion_type value in json  : `"
-                   << tmpShaderType << "|" << strToLookFor
-                   << "` does not map to a valid "
-                      "SceneInstanceTranslationOrigin value, so defaulting "
-                      "motion type to SceneInstanceTranslationOrigin::Unknown.";
+      LOG(WARNING)
+          << "getShaderTypeFromJsonDoc : `shader_type` value in json  : `"
+          << tmpShaderType << "` -> `" << strToLookFor
+          << "` does not map to a valid "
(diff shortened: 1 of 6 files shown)


Skylion007 (approved):

bigbike:
Please do not commit in this diff which needs further discussion on Mon.

jturner65:
> 
> 
> Please do not commit in this diff which needs further discussion on Mon.

Regardless of what we make the default value for the stage shader type, this needs to be added to the code base, to enable undoing a specified shader-type in configs, such as the example I mentioned above.

aclegg3 (approved):
This looks good to me. Simply adds the option to use the material default instead of manually choosing a shader. 👍

bigbike:
OK,At the beginning it seems a bit weird to me because the other items in that enum is `pbr`, `flat`, `phong`, `material` is just not a `shader_type` after all.
I understand you add this term, meaning "material driven", at the beginning, in such a case, user can directly specify "flat", "phong" or "pbr" in terms of shader_type.
(But 2nd thought is what if user does not know which should use as well? So yeah, you can add a term called "material".)

PR #1326, opened 2021-06-17
[AO migration] JointMotor API

aclegg3:
## Motivation and Context

Add JointMotor API for articulated object PD control via Bullet physics. Supports single dof joint types (Revolute and Prismatic) and Spherical joints.

Note: also added getters for joint and link names to improve user experience.

TODO: Spherical joint motor velocity targets are not behaving as expected. This may be misunderstanding of parameters or error within Bullet. May need to log an issue and circle back on this in a later PR. Position targets are working as expected.

## How Has This Been Tested

New pytests for the changes.

Example of spherical JointMotor position control for humanoid holding a T-pose position:
 

https://user-images.githubusercontent.com/1445143/122348713-4b95f300-cf00-11eb-836e-10128fa1acff.mp4

## Types of changes

## Checklist


src/esp/bindings/PhysicsBindings.cpp
```diff
@@ -44,6 +44,35 @@ void initPhysicsBindings(py::module& m) {
...
+      .def_readwrite("position_target", &JointMotorSettings::positionTarget)
+      .def_readwrite("spherical_position_target",
+                     &JointMotorSettings::sphericalPositionTarget)
+      .def_readwrite("position_gain", &JointMotorSettings::positionGain)
+      .def_readwrite("velocity_target", &JointMotorSettings::velocityTarget)
+      .def_readwrite("spherical_velocity_target",
+                     &JointMotorSettings::sphericalVelocityTarget)
+      .def_readwrite("velocity_gain", &JointMotorSettings::velocityGain)
+      .def_readwrite("max_impulse", &JointMotorSettings::maxImpulse)
+      .def_readwrite("motor_type", &JointMotorSettings::motorType);
```
jturner65, line 74: how about some docstrings?

src/esp/physics/ArticulatedObject.h
```diff
@@ -46,6 +47,94 @@ enum class JointType : int {
...
+                     double _velocityGain,
+                     double _maxImpulse) {
+    positionTarget = _positionTarget;
+    positionGain = _positionGain;
+    velocityTarget = _velocityTarget;
+    velocityGain = _velocityGain;
+    maxImpulse = _maxImpulse;
+  }
+
+  //! constructor for spherical motor settings
```
jturner65, line 78: Do docs with this format get picked up in our documentation?  If not I think it would be worth the effort to refactor them to appropriate docs.

tests/test_physics.py
```diff
@@ -980,3 +990,232 @@ def test_articulated_object_fixed_base_proxy():
...
+    cfg_settings["enable_physics"] = True
+    hab_cfg = examples.settings.make_cfg(cfg_settings)
+    with habitat_sim.Simulator(hab_cfg) as sim:
+        art_obj_mgr = sim.get_articulated_object_manager()
+        # parse URDF and add an ArticulatedObject to the world
+        robot = art_obj_mgr.add_articulated_object_from_urdf(filepath=robot_file)
+        assert robot.is_alive
+        # When URDF joint damping is defined, we generate a set of motors automatically
+        # get a map of joint motor ids to starting DoF indices
+        existing_joint_motors = robot.get_existing_joint_motor_ids()
```
jturner65, line 1013: `robot.get_existing_joint_motor_ids()`
Could/Should this be a read-only property?

jturner65 (approved):
LGTM : Left a few comments, mostly about comments :).

src/esp/physics/ArticulatedObject.h
aclegg3: Yeah, these get picked up. A bit lazy, but fine IMO for small doc briefs.

src/esp/physics/ArticulatedObject.h
```diff
@@ -46,6 +47,94 @@ enum class JointType : int {
...
+  JointMotorSettings(const Mn::Quaternion& _sphericalPositionTarget,
+                     double _positionGain,
+                     const Mn::Vector3& _sphericalVelocityTarget,
+                     double _velocityGain,
+                     double _maxImpulse)
+      : sphericalPositionTarget(_sphericalPositionTarget),
+        sphericalVelocityTarget(_sphericalVelocityTarget) {
+    positionGain = _positionGain;
```
Skylion007, line 86: These should all be member list inits, no?

src/esp/physics/ArticulatedObject.h
```diff
@@ -46,6 +47,94 @@ enum class JointType : int {
...
+ public:
+  JointMotorSettings() = default;
+
+  //! constructor for single dof motor settings
+  JointMotorSettings(double _positionTarget,
+                     double _positionGain,
+                     double _velocityTarget,
+                     double _velocityGain,
+                     double _maxImpulse) {
+    positionTarget = _positionTarget;
```
Skylion007, line 71: likewise here.

tests/test_physics.py
```diff
@@ -593,7 +593,17 @@ def getRestPositions(articulated_object):
 
 def getRandomPositions(articulated_object):
     r"""Constructs a random pose vector for an ArticulatedObject with unit quaternions for spherical joints."""
-    rand_pose = np.random.uniform(-1, 1, len(articulated_object.joint_positions))
+    lower_limits = articulated_object.get_joint_position_limits(upper_limits=False)
+    upper_limits = articulated_object.get_joint_position_limits(upper_limits=True)
+    for ix, val in enumerate(lower_limits):
```
Skylion007, line 598:
```suggestion
[val if val >= -1 else -1 for val in lower_limits]
```
or
```suggestion
lower_limits = np.maximum(lower_limits, -1)

File: tests/test_physics.py [hunk 7]
```diff
@@ -593,7 +593,17 @@ def getRestPositions(articulated_object):
 
 def getRandomPositions(articulated_object):
     r"""Constructs a random pose vector for an ArticulatedObject with unit quaternions for spherical joints."""
-    rand_pose = np.random.uniform(-1, 1, len(articulated_object.joint_positions))
+    lower_limits = articulated_object.get_joint_position_limits(upper_limits=False)
+    upper_limits = articulated_object.get_joint_position_limits(upper_limits=True)
```
Skylion007, line 597: This seems like a counter_intuitive return for me. You could return a tuple that includes both, right?

src/esp/physics/ArticulatedObject.h
```diff
@@ -553,6 +670,102 @@ class ArticulatedObject : public esp::physics::PhysicsObjectBase {
...
+    CHECK(jointMotors_.count(motorId) > 0);
+    CHECK(jointMotors_.at(motorId)->settings.motorType == settings.motorType);
+    jointMotors_.at(motorId)->settings = settings;
+  }
+
+  /**
+   * @brief Query a map of motorIds -> links/joints for all active JointMotors.
+   */
+  virtual std::map<int, int> getExistingJointMotors() {
+    std::map<int, int> motorIdsToLinkIds;
```
Skylion007, line 721: Reserve size of map

Skylion007 (changes requested):

src/esp/physics/ArticulatedObject.h
aclegg3: std::map has no reserve without custom allocator.

Skylion007: Actually, wait jointMotors is already a std::map right? just use the copy constructor.

tests/test_physics.py
aclegg3: Yeah, that makes sense. I was trying to avoid extra work if user doesn't want both, but simplicity is probably best.

src/esp/physics/ArticulatedObject.h
Skylion007: @aclegg3 Went ahead and did it with an iterator constructor which should be faster.

Skylion007 (approved):

tests/test_physics.py
```diff
@@ -593,14 +593,9 @@ def getRestPositions(articulated_object):
...
     r"""Constructs a random pose vector for an ArticulatedObject with unit quaternions for spherical joints."""
-    lower_limits = articulated_object.get_joint_position_limits(upper_limits=False)
-    upper_limits = articulated_object.get_joint_position_limits(upper_limits=True)
-    for ix, val in enumerate(lower_limits):
-        if val < -1:
-            lower_limits[ix] = -1
-    for ix, val in enumerate(upper_limits):
-        if val > 1:
-            upper_limits[ix] = 1
+    joint_limits = articulated_object.joint_position_limits
```
Skylion007, line 596:
```suggestion
    lower_limits, upper_limits = articulated_object.joint_position_limits
```

src/esp/physics/ArticulatedObject.h
```diff
@@ -593,6 +803,9 @@ class ArticulatedObject : public esp::physics::PhysicsObjectBase {
   //! link object for the AO base
   ArticulatedLink::uptr baseLink_;
 
+  //! map motorId to JointMotor
+  std::map<int, JointMotor::uptr> jointMotors_;
```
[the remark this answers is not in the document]
eundersander: Or if not vector, could this at least be `unordered_map`? I guess it's faster than `map` (fewer memory allocations, anyway).

tests/test_physics.py
aclegg3: Is this more efficient given the lines below with limiting?

eundersander (commented):
I left some comments. I'm worried about the use of `CHECK` everywhere and honestly wondering if you mixed up `CHECK` and `ESP_CHECK`.

src/esp/physics/ArticulatedObject.h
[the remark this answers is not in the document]
Skylion007: @eundersander Whole heartily agree. We should submit a PR to replace as many maps as possible with unordered maps or even faster ones like https://github.com/martinus/robin-hood-hashing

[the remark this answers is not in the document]
aclegg3: Good point. I think `unordered_map` is probably better here.

tests/test_physics.py
Skylion007: It will just error if we ever increase or decrease the number of elements returned by the function and is useful for that purpose. Also means you don't have to worry about index out of bound accesses for the tuple.

src/esp/physics/ArticulatedObject.h
```diff
@@ -553,6 +667,102 @@ class ArticulatedObject : public esp::physics::PhysicsObjectBase {
...
+  virtual void removeJointMotor(const int motorId) {
+    CHECK(jointMotors_.count(motorId) > 0);
+    jointMotors_.erase(motorId);
+  }
+
+  /**
+   * @brief Get a copy of the JointMotorSettings for an existing motor.
+   */
+  virtual JointMotorSettings getJointMotorSettings(const int motorId) {
+    CHECK(jointMotors_.count(motorId) > 0);
```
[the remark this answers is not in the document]
aclegg3: `CHECK` is fatal from the glog:

```
// CHECK dies with a fatal error if condition is not true.  It is *not*
// controlled by DCHECK_IS_ON(), so the check will be executed regardless of
// compilation mode.  Therefore, it is safe to do things like:
//    CHECK(fp->Write(x) == 4)
#define CHECK(condition)  \
      LOG_IF(FATAL, GOOGLE_PREDICT_BRANCH_NOT_TAKEN(!(condition))) \
             << "Check failed: " #condition " "
```

A sanity check, adding `CHECK(jointMotors_.count(motorId) == 0);` here produces:

```
F0617 13:17:18.694432 399650304 ArticulatedObject.h:703] Check failed: jointMotors_.count(motorId) == 0 
*** Check failure stack trace: ***
Fatal Python error: Aborted

Current thread 0x0000000117d22e00 (most recent call first):
  File "...habitat-sim/tests/test_physics.py", line 1188 in test_articulated_object_joint_motors
... more stack trace
```

src/esp/physics/ArticulatedObject.h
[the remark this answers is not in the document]
Skylion007: @aclegg3 The only case where you may want to use an non-unordered map it looks like if you want the returned keys to be sorted.

src/esp/physics/ArticulatedObject.h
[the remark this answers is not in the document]
aclegg3: The macro override in `esp/core/logging.h`, however may not be fatal but appears to be unused. Perhaps this is not in line with our expectations. Happy to change this to ESP_CHECK. Should probably refactor several places in the code for consistency at some point.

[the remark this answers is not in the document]
eundersander: Ok, I didn't notice that `CHECK` was fatal. But yes, for a fatal user error, we settled on `ESP_CHECK`. One nice thing is the way it bubbles up errors to python if you're running python (so you get a python callstack instead of just aborting the whole process).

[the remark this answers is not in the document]
eundersander: See also comment at top of src/esp/core/Check.h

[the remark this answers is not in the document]
eundersander: >it actually doesn't link the C++ line of the error. 
We could certainly add __FILE__ and __LINE__ inside the ESP_CHECK macro. At the time I implemented ESP_CHECK, I guess I figured any user could search for the error string to find the C++ source. Additionally, in theory a user shouldn't care about the source C++ code since she generally doesn't have access to it.

eundersander (approved):
LGTM

PR #1327, opened 2021-06-17
fix for KINEMATIC objects being moved by Bullet unexpectedly

eundersander:
## Motivation and Context

Alex and I have both seen behavior where a Bullet body with kinematic motion will get moved by Bullet stepWorld. This is avoided by setting zero mass. Alex speculated that collision-resolution code isn't checking the kinematic flag and is thus moving the kinematic body when the mass is non-zero.

## How Has This Been Tested

I tested locally in a scene with kinematic and dynamic objects. This change fixed the behavior bug for me.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/bullet/BulletRigidObject.cpp
+++ b/src/esp/physics/bullet/BulletRigidObject.cpp
@@ -332,7 +332,7 @@ void BulletRigidObject::constructAndAddRigidBody(MotionType mt) {
 
   double mass = 0;
   btVector3 bInertia = {0, 0, 0};
-  if (mt != MotionType::STATIC) {
+  if (mt == MotionType::DYNAMIC) {
     mass = tmpAttr->getMass();
     bInertia = btVector3(tmpAttr->getInertia());
     if (bInertia == btVector3{0, 0, 0}) {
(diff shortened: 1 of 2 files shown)


aclegg3 (approved):

src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -394,7 +394,7 @@ void BulletRigidObject::constructAndAddRigidBody(MotionType mt) {
         bObjectRigidBody_->getCollisionFlags() |
         btCollisionObject::CF_KINEMATIC_OBJECT);
     CORRADE_INTERNAL_ASSERT(bObjectRigidBody_->isKinematicObject());
-    CORRADE_INTERNAL_ASSERT(!bObjectRigidBody_->isStaticObject());
+    CORRADE_INTERNAL_ASSERT(bObjectRigidBody_->isStaticObject());
```
jturner65, line 397: Can the rigid body be both kinematic and static?

aclegg3:
For the failing test, I adjusted the golden number when I made the change you are reverting. Essentially a dynamic box is dropping on a moving kinematic platform and the exact response of the box (and thus the final transform) is different when mass is 0.

src/esp/physics/bullet/BulletRigidObject.cpp
aclegg3: The Bullet static flag is set when you provide 0 mass, so yes it can be both.

aclegg3: The won't change the Habitat MotionType of the object or what we allow it to do.

aclegg3:
Remaining test failure is fixed on master.

PR #1057, opened 2021-01-28
add offline-render utils

eundersander:
## Motivation and Context

The goal of these utils is to help a user produce renders (images) given a user-specified model, camera intrinsics and camera transforms.

Steps shown in the demo function:
1. Write a Habitat replay file that stores (a) a reference to our desired model to be rendered, and (b) a camera transform (position and rotation) for each render.
2. Initialize the simulator for offline rendering, specifying camera intrinsics.
3. Load the replay and produce renders (saved as png files).

The use of a Habitat replay file is slightly overkill, but it'll help us extend to more complex scenes (multiple models) and other renderers (Unity, Blender) in the future.

## Texture Downsampling
Update March 2: support for texture downsampling, see images in comments below. Think of texture resolution as "texels per square meter" in world space. downsample_factor=0 is default and full resolution. 1 means downsampling 2x, e.g. a MP3D 2048x2048 texture gets downsampled to 1024x1024. As the downsample_factor increases and the downsampled texture resolutions approach 1x1, the the visuals start to look like solid-colored shapes with no texture.

## Mesh Simplification
Update April 16: support for mesh simplification at load-time, see images below. This uses the Magnum MeshOptimizer plugin. We reduce mesh triangle count to some fraction of the original count. As mesh_simplification_fraction drops below 0.5, we see more and more severe artifacts like gaps in the mesh.

## How Has This Been Tested

todo: add python unit tests for these utilities

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1329, opened 2021-06-20
Update to latest Magnum with configuration iteration and support for new meshoptimizer

mosra:
## Motivation and Context

This brings mostly:

- Support for [configuration value/group iteration](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Configuration.html#Utility-Configuration-iteration), requested by @jturner65
- Support for MeshOptimizer 0.16 and its updated simplifier, reuqested by @eundersander. Should help the artifacts in https://github.com/facebookresearch/habitat-sim/pull/1057.

But also:

 * Support for OpenEXR 3.0 (so the package installed from Homebrew can work with it, which is also slowly inching towards an ability to use it as a CMake subproject)
 * Support for Glslang 11.4 (which fixes the *really strange* linker bug @bigbike had when attempting to use `magnum-shaderconverter` on macOS some time ago, don't even get me started on the aspects of this project buildsystem quirks)
  * Support for SPIRV-Tools 2021.1 

## How Has This Been Tested

I was very angry for a few days, but eventually the CIs stopped giving me a red light.

## Types of changes


Skylion007 (approved):

Skylion007:
@mosra Wow, you buried the lead with the new Corrade::Pair class. We use std::pair everywhere currently, is it ready to swap over all our std::pair to Corrade::Pair in the codebase?

jturner65 (approved):
LGTM.  I can't wait to use this stuff. Thanks!

mosra:
> We use std::pair everywhere currently

@Skylion007 ... and you can continue to do so, the corrade `Pair` is convertible to the STL equivalent so in most cases you won't even realize it's a different type. Same is with string (view)s and `std::unique_ptr` already, I don't think Habitat directly uses the `Pointer` anywhere and yet it works :)

Skylion007:
@mosra Yeah, I was just referring to the perf advantages with it being trivially copyable and all. It would be good to change our code over, no?

Skylion007:
This is probably the only struct that can benefit though: https://github.com/facebookresearch/habitat-sim/blob/34ec4f664742e762107ef606208e66259e051775/src/esp/assets/MeshMetaData.h#L59

mosra:
> It would be good to change our code over, no?

I don't think it'd matter in cases where just single instances are used. In Magnum I plan to use it in hashmap internals and the RenderGraph and there I expect a significant difference on reallocation, copying 10k pairs via memcpy will definitely be faster than looping over and calling copy constructors because `std::is_trivially_copyable` says no.

I see no other complaints, so merging :P (Did anybody witness any perf improvements from the last PR?)

PR #1330, opened 2021-06-21
[Perf] Use reserve() on unordered_map calls

Skylion007:
## Motivation and Context
* Performance Optimization. Unordered maps are often faster than std::map with one notable exception, when they have to be resized. Unlike std::maps though, the size can be reserved ahead of time. We should do that whenever possible prevent the O(n(log(n))). Even if we aren't sure of the exact size, we often have an upperbound.

## How Has This Been Tested
* CI

## Types of changes

## Checklist


jturner65 (approved):
LGTM

PR #1333, opened 2021-06-22
[fix CI] Update one test image screenshot

bigbike:
## Motivation and Context
We recently upgraded the testing models (found [here](http://dl.fbaipublicfiles.com/habitat/habitat-test-scenes_v1.0.zip)), making them using "flattened" materials so that our simulator can detect and render them using built-in flat shader. (These models do not have meaningful pbr materials, thus it does not make any sense to render them with PBR shader.)

However, our reference images (for CI tests) are rendered using the PBR shader before which causes the CI test failures in recent 2 days. This diff is trying to fix this problem.

The file in this diff is newly generated. 

## HOW to generate the new test images?? 
- Build your branch with the `--build-tests`. For example `./build.sh -j8 --no-lto --build-tests`;
- Run the test with `--save-diagnostic  [folder]` , For example, I created a new folder called `new-tests` in the habitat-sim;
- Images are named by the system automatically, which is very handy. (for example, here system named it `SimTestExpectedSceneWithLighting.png`)

## Types of changes

## Checklist


jturner65 (approved):
LGTM.  Fixed the issue when I tested it locally for Simtest.cpp

PR #1337, opened 2021-06-23
[Clang-Tidy] Enables readability-avoid-const-decl

Skylion007:
## Motivation and Context
* Removes useless const predicates from header declarations where they don't have an effect (ie they are primitives). Const ref or const pointers will remain unaffected. They can still be declared const in the function definition.

## How Has This Been Tested
* Locally and CI

## Types of changes

I also wanted to enable a check that made sure the naming of the params was consistent between the headers and c++, but that changed quite a lot of functions so I am holding off on that change for now.

## Checklist


jturner65 (approved):
LGTM.

PR #1338, opened 2021-06-23
[ObjectInstances] Remove platform dependence on pivot char; add checks for validity

jturner65:
## Motivation and Context
This small PR removes the platform-dependent pivot character used to partition the count component of the object instance name, so that users can write platform-independent code referencing object instances by name.  Checks were also added to verify that candidate names were formatted appropriately.

## How Has This Been Tested
All c++ and python tests pass locally.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/managedContainers/ManagedContainerBase.cpp
+++ b/src/esp/core/managedContainers/ManagedContainerBase.cpp
@@ -199,7 +199,7 @@ std::string ManagedContainerBase::getUniqueHandleFromCandidatePerType(
       const std::string digitStr = vals.back();
       if (digitStr.empty() ||
           (std::find_if(digitStr.begin(), digitStr.end(), [](unsigned char c) {
-             return !std::isdigit(c);
+             return std::isdigit(c) == 0;
            }) != digitStr.end())) {
         // if string is not a valid representation of an integer, skip this name
         // candidate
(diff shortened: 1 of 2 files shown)


src/esp/core/managedContainers/ManagedContainerBase.cpp
```diff
@@ -171,25 +171,43 @@ std::string ManagedContainerBase::getObjectInfoCSVString(
...
-  if (resVals.size() != 0) {
+
+  if (objHandles.size() != 0) {
     // handles exist with passed substring.  Find highest handle increment, add
     // 1 and use for new name 1, build new handle
-    for (const std::string& s : resVals) {
+    for (const std::string& objName : objHandles) {
+      // make sure pivot character is found in name
+      if (objName.find(pivotChar) == std::string::npos) {
+        continue;
```
eundersander, line 191: What's going on here? Why would this happen? (or can we just assert?)

eundersander (approved):
LGTM

src/esp/core/managedContainers/ManagedContainerBase.cpp
jturner65: This function is also intended to be used by containers that do not normally manage numbered instances (such as for the attributes).  Intent is to allow for same-named configs to be auto-incremented before save, if the user doesn't wish to overwrite the existing config.

As far as the object wrappers/instances go (which is what you are interested in) this is superfluous - there will never be a handle that does not contain the pivot.

PR #1341, opened 2021-06-24
Robot wrapper

aclegg3:
## Motivation and Context

End goal is to provide an extensible convenience interface for common robot functionality. This PR addresses immediate need for wheeled mobile manipulator and specifically Fetch robot interface.

Migrated from @ASzot's implementation.

Some open questions are listed TODO in the code for future consideration.

Secondary:
- Adds `"hab_fetch"` option to `datasets_download.py`
- Fixes a bug with ArticulatedLink bounding boxes
- Adds Bullet debug draw toggle to viewer with `,`

## How Has This Been Tested

Added a pytest.
Output: 

https://user-images.githubusercontent.com/1445143/123559802-281a4600-d753-11eb-8661-a41530aede6a.mp4

## Types of changes

## Checklist


habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,322 @@
+from typing import List, Optional, Tuple
+
+import attr
+import magnum as mn
+import numpy as np
+
+import habitat_sim
```
erikwijmans, line 7:
```suggestion
from habitat_sim.simulator import Simulator
```

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+class MobileManipulator(RobotInterface):
+    """Robot with a controllable base and arm."""
+
+    def __init__(
+        self,
+        params: MobileManipulatorParams,
+        urdf_path: str,
+        sim: habitat_sim.simulator.Simulator,
```
erikwijmans, line 86:
```suggestion
        sim: Simulator,
```

These two should make it so that you don't have to import simulator before robots.

habitat_sim/robots/mobile_manipulator.py
erikwijmans: Looks like you'll also need to grab other stuff in a similar fashion as the top-level module is being used in other places.

habitat_sim/robots/fetch_robot.py
```diff
@@ -0,0 +1,159 @@
+import magnum as mn
+import numpy as np
+import pybullet as p
```
erikwijmans, line 3: Consider making this soft.  You can do a similar thing like was done with torch where you import once it's needed.

erikwijmans (commented):
Just some quick comments

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,322 @@
...
+        limit_robo_joints: bool = True,
+    ):
+        """Constructor
+        :param limit_robo_joints: If true, joint limits of robot are always
+            enforced.
+        """
```
Skylion007, line 92: Good practice to call super constructor.

habitat_sim/robots/fetch_robot.py
```diff
@@ -0,0 +1,159 @@
...
+        return np.array(world_ee)
+
+    def calculate_ee_ik(self, ee_targ):
+        joint_pos = self.get_arm_pos()
+        joint_vel = self.get_arm_vel()
+
+        self._ik.set_arm_state(joint_pos, joint_vel)
+
+        des_joint_pos = self._ik.calc_ik(ee_targ)
+        des_joint_pos = list(des_joint_pos)
```
Skylion007, line 141: Why list then array? You are iterating twice for no reason?

habitat_sim/robots/fetch_robot.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def set_arm_state(self, joint_pos, joint_vel=None):
+        if joint_vel is None:
+            joint_vel = np.zeros((len(joint_pos),))
+        for i in range(7):
+            p.resetJointState(
+                self.robo_id, i, joint_pos[i], joint_vel[i], physicsClientId=self.pc_id
+            )
+
+    def calc_ik(self, targ_ee):
```
Skylion007, line 53: Annotate some return types and you should see a lot of these np.array calss are unnecessary.

aclegg3:
Looks like any `import habitat_sim` will error unless we manually defer the `robots` import in `__init__.py`. 🤔

erikwijmans:
I think that is because there is no `__init__.py` in the robots folder.  Without one of those, even when they are just a blank file, python often does the wrong thing.

aclegg3:
> I think that is because there is no `__init__.py` in the robots folder. 

😆  Apparently I forgot to `git add` that.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def close_gripper(self) -> None:
+        """Set gripper to the close state"""
+        self.set_gripper_state(self.params.gripper_closed_state)
+
+    def open_gripper(self) -> None:
+        """Set gripper to the open state"""
+        self.set_gripper_state(self.params.gripper_open_state)
+
+    def is_gripper_open(self) -> bool:
```
mathfac, line 268:
```suggestion
    @property
    def is_gripper_open(self) -> bool:
```

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+        # Give some threshold for the open state.
+        grip_dist = np.abs(self._gripper_state - self.params.gripper_open_state)
+        return grip_dist < 0.001
+
+    def set_arm_pos(self, ctrl: List[float]):
```
mathfac, line 273: Maybe make `arm_pos` property as well, with getter and setter.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+        return [self._robot.joint_positions[i] for i in arm_pos_indices]
+
+    def get_arm_vel(self) -> np.ndarray:
+        """Get the velocity of the arm joints."""
+        arm_dof_indices = list(
+            map(lambda x: self.joint_dof_indices[x], self.params.arm_joints)
+        )
+        return [self._robot.joint_velocities[i] for i in arm_dof_indices]
+
+    def set_arm_mtr_pos(self, ctrl: List[float]) -> None:
```
mathfac, line 295:
```suggestion
    def set_arm_motor_pos(self, ctrl: List[float]) -> None:
```

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+        self._robot.joint_positions = joint_positions
+
+    def get_arm_pos(self) -> np.ndarray:
+        """Get the current arm joint positions in radians."""
+        arm_pos_indices = list(
+            map(lambda x: self.joint_pos_indices[x], self.params.arm_joints)
+        )
+        return [self._robot.joint_positions[i] for i in arm_pos_indices]
+
+    def get_arm_vel(self) -> np.ndarray:
```
mathfac, line 288:
```suggestion
    def get_arm_velocity(self) -> np.ndarray:
```

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+    # WHEEL RELATED
+    #############################################
+
+    # TODO: add some functions for easy wheel control
+
+    #############################################
+    # BASE RELATED
+    #############################################
+
+    def set_base_position(self, position):
```
mathfac, line 328: We should use `pos` or `position` in the methods/properties.

habitat_sim/robots/fetch_robot.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+        super().__init__(fetch_params, urdf_path, sim, limit_robo_joints)
+        self.back_joint_id = 6
+        self.head_rot_jid = 8
+        self.head_tilt_jid = 9
+
+    def reconfigure(self) -> None:
+        super().reconfigure()
+        # remove any default damping motors
+        for motor_id in self._robot.existing_joint_motor_ids:
```
mathfac, line 53: If this all isn't specific for the Fetch robot only we can move it to Mobile Manipulator `reconfigure`.

mathfac (approved):
Looks great. Left some naming and styling comment that can be easily addressed. Some harder to fix comments can be addressed in follow up PRs. Thank you @ASzot and @aclegg3  for contributing the wrapper.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+        self.set_gripper_state(self.params.gripper_closed_state)
+
+    def open_gripper(self) -> None:
+        """Set gripper to the open state"""
+        self.set_gripper_state(self.params.gripper_open_state)
+
+    def is_gripper_open(self) -> bool:
+        # Give some threshold for the open state.
+        grip_dist = np.abs(self._gripper_state - self.params.gripper_open_state)
+        return grip_dist < 0.001
```
mathfac, line 271: `0.001` should be named constant like `GRIPPER_CLOSED_EPS` or be part of the parameters.

habitat_sim/robots/fetch_robot.py
aclegg3: Depends on whether we want to use the default damping motors from URDF or clear them for custom settings. This seemed like a top level decision so we could call super methods. 🤔 
Let's see how we feel once we try a couple more robots.

habitat_sim/robots/mobile_manipulator.py
aclegg3: Simulation will change the arm state, so the property may be out of date unless explicitly updated. Maybe better to let the user manage that since we'll otherwise risk a stale state in observation, reward, termination, etc... depending on when update is called. If it ends up becoming inefficient for some reason we can add it later.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+        for link_id in self._robot.get_link_ids():
+            self.joint_pos_indices[link_id] = self._robot.get_link_joint_pos_offset(
+                link_id
+            )
+            self.joint_dof_indices[link_id] = self._robot.get_link_dof_offset(link_id)
+        self.joint_limits = self._robot.joint_position_limits
+
+        # print relevant joint/link info for debugging
+        for link_id in self._robot.get_link_ids():
+            print(
```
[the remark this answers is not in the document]
aclegg3: Moved this to a method return str instead.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+        self._robot.joint_positions = js
+        return self.get_end_effector_transform().translation
+
+    def calculate_ee_ik(self, ee_targ: np.ndarray) -> np.ndarray:
+        """Gets the joint states necessary to achieve the desired end-effector
+        configuration.
+        """
+        raise NotImplementedError("Currently no implementation for generic IK.")
+
+    def get_end_effector_transform(self) -> mn.Matrix4:
```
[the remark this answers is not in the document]
aclegg3: I'll refactor all to "ee" for consistency rather than risk being too verbose.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,391 @@
...
+import magnum as mn
+import numpy as np
+
+from habitat_sim.physics import JointMotorSettings, ManagedBulletArticulatedObject
+from habitat_sim.robots.robot_interface import RobotInterface
+from habitat_sim.simulator import Simulator
+
+
+@attr.s(auto_attribs=True, slots=True)
+class MobileManipulatorParams:
```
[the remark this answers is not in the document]
aclegg3: Completely agree. Config makes sense here.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,396 @@
...
+    @arm_joint_pos.setter
+    def arm_joint_pos(self, ctrl: List[float]):
+        """Kinematically sets the arm joints and sets the motors to target."""
+        joint_positions = self._robot.joint_positions
+        for i, jidx in enumerate(self.params.arm_joints):
+            self._set_mtr_pos(jidx, ctrl[i])
+            joint_positions[self.joint_pos_indices[jidx]] = ctrl[i]
+        self._robot.joint_positions = joint_positions
+
+    def get_arm_velocity(self) -> np.ndarray:
```
mathfac, line 293: This can be the property as well.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,396 @@
...
+        """Get the velocity of the arm joints."""
+        arm_dof_indices = list(
+            map(lambda x: self.joint_dof_indices[x], self.params.arm_joints)
+        )
+        return [self._robot.joint_velocities[i] for i in arm_dof_indices]
+
+    def set_arm_motor_pos(self, ctrl: List[float]) -> None:
+        """Set the desired target of the arm joints for the controller to
+        achieve.
+        """
```
mathfac, line 303: Same here.

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,464 @@
...
+
+        self._update_motor_settings_cache()
+
+    #############################################
+    # ARM RELATED
+    #############################################
+    @property
+    def arm_joint_limits(self) -> Tuple[np.ndarray, np.ndarray]:
+        """Get the arm joint limits in radians"""
+        arm_pos_indices = list(
```
Skylion007, line 222: This probably would have been a bit faster using a dedicated getter and a list comprehension. apologies for not getting to this pr in time*

habitat_sim/robots/mobile_manipulator.py
```diff
@@ -0,0 +1,464 @@
...
+            map(lambda x: self.joint_pos_indices[x], self.params.arm_joints)
+        )
+
+        lower_lims = [self.joint_limits[0][i] for i in arm_pos_indices]
```
Skylion007, line 226:
```suggestion
        lower_lims, upper_lims = zip(*self.joint_limits)
```

Skylion007: or 
```
lower_lims, upper_lims = self.joint_limits
```

Skylion007, line 226:
```suggestion
lower_lims, upper_lims = self.joint_limits[:, arm_pos_indices]
```
Will be, much, much faster.

PR #1349, opened 2021-06-28
Add DebugLineRender utility

eundersander:
## Motivation and Context

Singleton utility class for on-the-fly rendering of lines (e.g. every frame). This is intended for debugging or simple UX for prototype apps. The API prioritizes ease-of-use over maximum runtime performance.

https://user-images.githubusercontent.com/6557808/123686446-aaaf0e00-d804-11eb-8253-1ddd97d4ce0c.mp4

## How Has This Been Tested

integrated into replay_tutorial.py; also ad-hoc testing in viewer.cpp

## Types of changes

## Checklist


src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+  glDisable(GL_LINE_SMOOTH);  // anti-aliased lines cause artifacts
+
+  glLineWidth(_internalLineWidth);
+
+  // We draw lines multiple times, with pixel offsets, to produce a single
+  // thick, visually-appealing line. Todo: implement thick lines using
+  // triangles, for example https://mattdesl.svbtle.com/drawing-lines-is-hard.
+  // 1.2 is hand-tuned for Nvidia hardware to be just small enough so we don't
+  // see gaps between the individual offset lines.
```
eundersander, line 125: To be clear: the current thick-line implementation here is sloppy in the sense of not producing perfectly-rendered thick lines. We can upgrade in the future without changing the public API, if we wish.

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+  // restore to a reasonable default
+  Mn::GL::Renderer::setDepthFunction(Mn::GL::Renderer::DepthFunction::Less);
+
+  // Clear _verts to receive new data
+  arrayResize(_verts, 0);
+
+  // restore blending state if necessary
+  if (doToggleBlend) {
+    Mn::GL::Renderer::disable(Mn::GL::Renderer::Feature::Blending);
+    // Note: because we are disabling blending, we don't need to restore
```
eundersander, line 178: @mosra any concern here?

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+}
+
+void DebugLineRender::flushLines(const Magnum::Matrix4& camMatrix,
+                                 const Magnum::Matrix4& projMatrix,
+                                 const Magnum::Vector2i& viewport) {
+  CORRADE_ASSERT(_glResources,
+                 "DebugLineRender::flushLines: no GL resources; see "
+                 "also releaseGLResources", );
+
+  bool doToggleBlend = !glIsEnabled(GL_BLEND);
```
eundersander, line 101: @mosra @bigbike I don't love calling `glIsEnabled` but I didn't see a Magnum getter.

What I really need here is a way to "push" a render state and later pop it (then I wouldn't need the getter).

I'm open to other suggestions.

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+    Mn::GL::Renderer::setBlendEquation(Mn::GL::Renderer::BlendEquation::Add);
+  }
+
+  // Update buffer with new data
+  _glResources->buffer.setData(_verts, Mn::GL::BufferUsage::DynamicDraw);
+
+  // Update shader
+  _glResources->mesh.setCount(_verts.size());
+
+  glDisable(GL_LINE_SMOOTH);  // anti-aliased lines cause artifacts
```
eundersander, line 117: Feels weird calling OpenGL directly here.

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+Mn::Color4 remapAlpha(const Mn::Color4& src) {
+  // Because we render lines multiple times additively in flushLines, we need to
+  // remap alpha (opacity). This is an approximation.
+  constexpr float exponent = 2.f;
+  return Mn::Color4(src.r(), src.g(), src.b(), std::pow(src.a(), exponent));
+}
+
+// return false if segment is entirely clipped
+bool scissorSegmentToOutsideCircle(Mn::Vector3* pt0,
```
eundersander, line 30: This is used for `drawPathWithEndpointCircles` as seen in the video.

src/esp/sim/Simulator.cpp
```diff
@@ -72,6 +72,13 @@ void Simulator::close(const bool destroy) {
 
   resourceManager_ = nullptr;
 
+  if (debugLineRender_) {
+    // Python may keep around other shared_ptrs to this object, but we need
+    // to release GL resources here.
+    debugLineRender_->releaseGLResources();
+    debugLineRender_ = nullptr;
```
eundersander, line 79: This is weird. What I really want to do is delete the DebugLineRender instance here. But I can't because I had to share `shared_ptr`s with Python when doing the bindings. And Python tends to keep around the pointers indefinitely. So we don't actually control when this object gets deleted.

aclegg3 (commented):
Overall looks good. Cool feature.

src/esp/sim/Simulator.cpp
aclegg3: We used a weak_ptr wrapper paradigm to get around this with ManagedObjects. The wrapper does a lock and forwards the command if it succeeds, otherwise warning about the stale reference.

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+    Mn::Deg angle = Mn::Deg(360.f * float(seg) / numSegments);
+    Mn::Vector3 pt(Mn::Math::cos(angle), Mn::Math::sin(angle), 0.f);
+    if (seg > 0) {
+      drawTransformedLine(prevPt, pt, color);
+    }
+    prevPt = pt;
+  }
+
+  popTransform();
+}
```
mosra, line 264: No need to reinvent anything :) 

I was thinking you could have a generic `drawPrimitive()` utility:

```cpp
void DebugLineRender::drawPrimitive(const Mn::Trade::MeshData& mesh, const Mn::Matrix4& transformation, const Mn::Color4& color) {
  /* Turn the mesh into nonindexed lines first */
  if(mesh.primitive() != Mn::MeshPrimitive::Lines) {
    CORRADE_INTERNAL_ASSERT(mesh.primitive() == Mn::MeshPrimitive::LineLoop ||
                            mesh.primitive() == Mn::MeshPrimitive::LineStrip);
    return drawPrimitive(Mn::MeshTools::generateIndices(mesh), transformation, color);
  }
  if(mesh.isIndexed())
    return drawPrimitive(Mn::MeshTools::duplicate(mesh), transformation, color);

  /* Steal all the positions */
  for(const Mn::Vector3& position: mesh.attribute<Mn::Vector3>(Mn::Trade::MeshAttribute::Position))
    arrayAppend(_verts, Cr::InPlaceInit, transformation.transformPoint(position), color);
}
```

and then just reuse whatever existing primitive code (didn't you also have a debug renderer with colored axes? there's [`Primitives::axis3D()`](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Primitives.html#a4c8248d7f389b072c9d1a6a25ee363b0) for that too):

```cpp
void drawBox(const Mn::Vector3& min, const Mn::Vector3& max, const Mn::Color4& color) {
  drawPrimitive(Mn::Primitives::cubeWireframe(),
    _cachedInputTransform*
    Mn::Matrix4::translation((max + min)/2.0f)* /* i hope this is correct? */
    Mn::Matrix4::scaling((max - min)/2.0f),
    color);
}

void drawCircle(const Mn::Vector3& pos, float radius, int segments, const Mn::Color4& color, const Mn::Vector3& normal) {
  ...
  drawPrimitive(Mn::Primitives::circle3DWireframe(segments),
    _cachedInputTransform*
    Mn::Matrix4::lookAt(pos, pos + normal, randomPerpVec)*
    Mn::Matrix4::scaling(Mn::Vector3(radius, radius, 0.f)));
}
```

Add `#include`s as needed. The `Primitives` library should be sufficiently lightweight to have negligible perf impact when used this way.

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+  glLineWidth(_internalLineWidth);
+
+  // We draw lines multiple times, with pixel offsets, to produce a single
+  // thick, visually-appealing line. Todo: implement thick lines using
+  // triangles, for example https://mattdesl.svbtle.com/drawing-lines-is-hard.
+  // 1.2 is hand-tuned for Nvidia hardware to be just small enough so we don't
+  // see gaps between the individual offset lines.
+  const float x = _internalLineWidth * 1.2f;
+  constexpr float sqrtOfTwo = 1.4142f;
```
mosra, line 127: [`Mn::Constants::sqrt2()`](https://doc.magnum.graphics/magnum/structMagnum_1_1Math_1_1Constants.html#adeb45f98b1d92e393bf8bcbeacb05af1)

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+      drawTransformedLine(prevPt, pt, color);
+    }
+    prevPt = pt;
+  }
+
+  popTransform();
+}
+
+void DebugLineRender::drawPathWithEndpointCircles(
+    const std::vector<Mn::Vector3>& points,
```
mosra, line 267: FYI you could take an `Containers::ArrayView<const Vector3>`, and then if you `#include <Corrade/Containers/ArrayViewStl.h>`, the `std::vector` is implicitly convertible to it, but it allows you to use any other contiguous container (`std::array`, a C array...) as well.

Could be worth it if you're often submitting paths with known sizes and can thus have e.g. `Vector3 path[5] { ... }` instead of populating a vector, if it's always a dynamic vector then probably not.

src/esp/gfx/DebugLineRender.cpp
mosra: I thought this was some ancient relic and so this isn't even exposed anywhere.

If you want a "less" direct way and something that should be more futureproof, `GL::Renderer::disable(GL::Renderer::Feature(GL_LINE_SMOOTH))` and then please add a TODO for me.

src/esp/gfx/DebugLineRender.cpp
mosra: Yeah sorry, a nasty pain point. 

This should _ideally_ eventually be handled by a scoped renderer state, where you enable what you need and Magnum's state tracker then takes care of cleaning that up again at the end of scope, _ideally_ also not calling into GL more than necessary. Practically we'll probably switch to Vulkan sooner than I get to implement such a feature.

src/esp/gfx/DebugLineRender.cpp
mosra: Same as [in the other comment](https://github.com/facebookresearch/habitat-sim/pull/1349#discussion_r660108712), this should be something Magnum eventually takes care of. 

For now, this is fine.

src/esp/gfx/DebugLineRender.cpp
eundersander: I feel like your suggestion here is more complicated than what I have. From my perspective, a bit of code to create a circle or a box is simple and readable, and users can easily adapt it later to their own shapes (tetrahedrons, helices, etc.).

src/esp/gfx/DebugLineRender.cpp
eundersander: I also hit an issue with this not even compiling for Emscripten. I'm starting to think it's pretty unlikely that anyone is going to set this setting to true. I'm just going to remove this line and not worry about it.

src/esp/gfx/DebugLineRender.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+  // Update buffer with new data
+  _glResources->buffer.setData(_verts, Mn::GL::BufferUsage::DynamicDraw);
+
+  // Update shader
+  _glResources->mesh.setCount(_verts.size());
+
+  glDisable(GL_LINE_SMOOTH);  // anti-aliased lines cause artifacts
+
+  glLineWidth(_internalLineWidth);
```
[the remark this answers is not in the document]
eundersander: FWIW, this feature is working fine on my Macbook with AMD Radeon Pro 5500M.

>(Yes, yes, I need to add a wide line renderer to Magnum, this is ridiculous.)
:+1:

src/esp/gfx/DebugLineRender.cpp
eundersander: I support all efforts to reduce our per-frame dynamic allocations!

eundersander: Hmmmm, in fact I was calling this in one test with a fixed list, but I hit this:

```
  lineRender->drawPathWithEndpointCircles(
    {{0.f, 0.f, 0.f}, {0.1f, 0.f, 0.f}}, radius, color);
```

`viewer.cpp:1156:55: error: cannot convert ‘<brace-enclosed initializer list>’ to ‘Corrade::Containers::ArrayView<const Magnum::Math::Vector3<float> >’`

src/esp/sim/Simulator.cpp
eundersander: The `weak_ptr` solution is interesting but feels like overkill for my situation. If a Python user keeps a reference to DebugLineRender after the sim has been closed, and she tries to use it, nothing bad happens.

src/esp/gfx/DebugLineRender.cpp
mosra: Ah, *that's* why I thought it's an ancient relic, because GLES/WebGL doesn't even define such a thing :D

src/esp/gfx/DebugLineRender.cpp
mosra: The code has a lot of extra API calls with long names to massage the mesh data, yes, but the essence of it is a single loop that copies line endpoints from a large gallery of existing primitives, where adding another one (such as a cone for visualizing spotlights) is extremely trivial :)

I commented also because my past experience is that any primitive generation code that wasn't thoroughly tested tended to rot over time, getting increasingly broken during minor refactorings. Since I suppose you won't want to spend time writing tests ensuring the mesh is generated correctly, this would delegate that responsibility to the engine.

src/esp/gfx/DebugLineRender.cpp
mosra: C++ is unintuitive. [Lengthy answer why is it like that](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1ArrayView.html#Containers-ArrayView-stl) -- see the yellow box titled "Conversion from std::initializer_list" there.

IOW, the API is designed to avoid very common use-after-free issues, which unfortunately makes this one use case annoying. What I commonly do on the engine side is adding a `std::initializer_list` overload for every function that takes an `ArrayView`. Like

```cpp
void DebugLineRender::drawPathWithEndpointCircles(const Cr::Containers::ArrayView<const Mn::Vector3> points, ...) { ... }

void DebugLineRender::drawPathWithEndpointCircles(const std::initializer_list<Mn::Vector3> points, ...) {
    drawPathWithEndpointCircles(Containers::arrayView(points), ...);
}
```

Without an overload, in the case above you'd have to do a bit more typing:

```cpp
  lineRender->drawPathWithEndpointCircles(
    Containers::arrayView<Vector3>({{0.f, 0.f, 0.f}, {0.1f, 0.f, 0.f}}), radius, color);
```

eundersander: Ok. I'm not too worried about the ugly callsite I gave above; it was just some temp testing code. I don't think the `initializer_list` version is necessary.

I've switched to ArrayView and also added a wrapper version that takes a std::vector; this was needed for the Python binding.

src/esp/gfx/DebugLineRender.cpp
eundersander: For now, rather than drag out the discussion, let's agree to disagree. :)

I'll add a comment in the file pointing to this discussion, in case folks in the future want to go this route.

src/esp/bindings/GfxBindings.cpp
```diff
@@ -205,6 +206,46 @@ void initGfxBindings(py::module& m) {
...
+      .def(
+          "set_line_width", &DebugLineRender::setLineWidth,
+          R"(Set global line width for all lines rendered by DebugLineRender.)")
+      .def(
+          "push_transform", &DebugLineRender::pushTransform,
+          R"(Push (multiply) a transform onto the transform stack, affecting all line-drawing until popped. Must be paired with popTransform().)")
+      .def("pop_transform", &DebugLineRender::popTransform,
+           R"(See push_transform.)")
+      .def("draw_box", &DebugLineRender::drawBox,
+           R"(Draw a box in world-space or local-space (see pushTransform).)")
```
aclegg3, line 220: No color for box bindings?

aclegg3 (approved):
LGTM. One comment on the bindings.

src/esp/bindings/GfxBindings.cpp
eundersander: draw_box takes two Vector3 and a color. This binding here just takes all the C++ function parameters as-is without explicitly declaring them all here.

aclegg3: I guess this just makes it harder for python only user seeing the doc-string API reference to know the parameters. Also, does implicit conversion allow keyword (named) arguments?

PR #1350, opened 2021-06-28
Bump version to 0.2.0

aclegg3:
## Motivation and Context

Bump sim version for release.


Skylion007:
There is also a version in package.json that we are ignoring. This should be updated as well.

erikwijmans (approved):
Ship it!

PR #1357, opened 2021-06-30
HM3D

erikwijmans:
## Motivation and Context

Add support for HM3D to the download script.  This behaves a little differently than other things as 1) the download links require authorization, an API token ID and token key (passed in as a user name/password) and HM3D is quite large so there are many slices of it to allow users to only download what they need.

## How Has This Been Tested

By downloading stuff!  Reviewers, if you want to try this but your account hasn't been enabled send me a ping.


habitat_sim/utils/datasets_download.py
```diff
@@ -101,6 +120,59 @@ def initialize_test_data_sources(data_path):
         },
     }
 
+    data_sources.update(
+        {
+            f"hm3d_{split}_{data_format}": {
+                "source": "https://api.matterport.com/resources/habitat/hm3d-{split}-{data_format}.tar{ext}".format(
```
Skylion007, line 126: Wouldn't this jsut be cleaner with fstrrings.

Skylion007 (commented):
Nit

habitat_sim/utils/datasets_download.py
erikwijmans: IMO no because of the need to add .gz to the obj+mtl tarballs

Skylion007: f strings can have nested ifs. *shrugs*

Skylion007 (approved):

erikwijmans:
Landing this so it can go out with conda tonight

PR #1358, opened 2021-06-30
Scale Phong materials to better match PBR intensities

aclegg3:
## Motivation and Context

Phong and PBR materials respond very differently to that same lighting. We recently updated the light setups to support our transition to PBR primary materials which caused Phong materials (primarily primitives and URDF with color overrides) to be very bright. This PR adds a multiplier hack to balance the color intensities of Phong and PBR.

## How Has This Been Tested
Used Viewer to balance w/ default lights:

Before: 
image

After: 
image

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -461,8 +461,10 @@ std::string ResourceManager::createColorMaterial(
     gfx::PhongMaterialData::uptr phongMaterial =
         gfx::PhongMaterialData::create_unique();
     phongMaterial->ambientColor = materialColor.ambientColor;
-    phongMaterial->diffuseColor = materialColor.diffuseColor;
-    phongMaterial->specularColor = materialColor.specularColor;
+    // NOTE: This multiplication is a hack to roughly balance the Phong and PBR
+    // light intensity reactions.
+    phongMaterial->diffuseColor = materialColor.diffuseColor * 0.175;
+    phongMaterial->specularColor = materialColor.specularColor * 0.175;
 
     std::unique_ptr<gfx::MaterialData> finalMaterial(phongMaterial.release());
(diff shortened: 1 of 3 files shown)


eundersander:
I'm not familiar with ResourceManager::createColorMaterial.

As I recall, p-viz-plan uses Flag::BuildPhongFromPbr to force gfx::buildPhongFromPbrMetallicRoughness. Should we also change that codepath? How will folks reproduce the visuals from p-viz-plan?

mathfac (approved):
Looks good, unblocking for the release.

src/esp/assets/ResourceManager.cpp
```diff
@@ -461,8 +461,10 @@ std::string ResourceManager::createColorMaterial(
     gfx::PhongMaterialData::uptr phongMaterial =
         gfx::PhongMaterialData::create_unique();
     phongMaterial->ambientColor = materialColor.ambientColor;
-    phongMaterial->diffuseColor = materialColor.diffuseColor;
-    phongMaterial->specularColor = materialColor.specularColor;
+    // NOTE: This multiplication is a hack to roughly balance the Phong and PBR
+    // light intensity reactions.
+    phongMaterial->diffuseColor = materialColor.diffuseColor * 0.175;
+    phongMaterial->specularColor = materialColor.specularColor * 0.175;
```
mosra, line 467: Hm, wouldn't it make sense to do this in the PBR shader directly instead? Like, the will go through many other changes for IBL anyway, and that would keep the hack in a single place.

aclegg3: Yeah, agreed. It makes sense to scale up in the PBR intesnties or scale down the Phong intensities in the shader, but I needed to get something in for the release visuals.

PR #1365, opened 2021-07-01
Add way to download data for the hand demo

ldcWV:
## Motivation and Context

Very small PR that adds a way to download the data needed for the WebXR Hand Demo, which is here: https://github.com/facebookresearch/habitat-sim/pull/1366

The command used to download the data will be
```
python -m habitat_sim.utils.datasets_download --uids webxr_hand_demo --data-path examples/web_apps/webxr_hand_demo/data
```
This will create `examples/web_apps/webxr_hand_demo/data/webxr_hand_demo_dataset` with the necessary files inside.

## How Has This Been Tested

I ran the dataset downloader and it worked.

## Types of changes

## Checklist


Skylion007 (approved):

habitat_sim/utils/datasets_download.py
```diff
@@ -88,6 +88,12 @@ def initialize_test_data_sources(data_path):
             "link": data_path + "scene_datasets/coda",
             "version": "1.0",
         },
+        "webxr_hand_demo_data": {
+            "source": "https://dl.fbaipublicfiles.com/habitat/data/scene_datasets/webxr_hand_demo_data.zip",
+            "package_name": "webxr_hand_demo_data.zip",
+            "link": data_path + "data",
```
eundersander, line 94: It's kinda weird that this dataset doesn't have a meaningful subfolder name once downloaded. It seems like the other datasets try to use a unique subfolder name, e.g. `"link": data_path + "replica_cad"`.

@aclegg3 thoughts?

aclegg3: Yeah, agreed. This will be used to generate a symlink, so best if it's specific. Maybe: 
```suggestion
            "link": data_path + "webxr_hand_demo",
```

ldcWV: @aclegg3 Note that in this case, `data_path` will not be the same as usual. I intend for the user to run
```
python -m habitat_sim.utils.datasets_download --uids webxr_hand_demo_data --data-path examples/web_apps/webxr_hand_demo
```
This means that the directory folder `data` will be created inside the directory `examples/web_apps/webxr_hand_demo`. Your suggestion will make the data path for the hand demo become `examples/web_apps/webxr_hand_demo/webxr_hand_demo`, which doesn't make much sense. Maybe I could name it `webxr_hand_demo_data` instead? I thought that would be redundant. But also I'm not sure what the purpose of the symlink is so I don't know if this would be better.

eundersander: >Your suggestion will make the data path for the hand demo become examples/web_apps/webxr_hand_demo/webxr_hand_demo

The convention that Alex has set up here is for a project to have a `data` directory and then various named datasets within there. In your case, your project only really needs this one dataset, but you should probably still stick to this convention. Also consider: what if someone wants to modify this hand demo to also use some objects from another dataset? where would that go?

I propose `examples/web_apps/webxr_hand_demo/data` as your project's data directory, and then advise users to do:
```
python -m habitat_sim.utils.datasets_download --uids webxr_hand_demo_data --data-path examples/web_apps/webxr_hand_demo/data
```
... and then the final path resolves to `examples/web_apps/webxr_hand_demo/data/webxr_hand_demo` or perhaps `examples/web_apps/webxr_hand_demo/data/webxr_hand_demo_dataset`.

eundersander (approved):
LGTM

PR #1367, opened 2021-07-01
Update Magnum to latest.

mosra:
## Motivation and Context

Mainly just because of one bugfix and buildsystem updates that might get useful soon, nothing major.

 * Supports OpenEXR as a CMake subproject, needed for @bigbike's PBR HDR and IBL support (though this doesn't integrate OpenEXR into Habitat yet as there's nothing using it right now)
 * Growable arrays now respect 64- and 128-bit type alignment, fixing Emscripten `SAFE_HEAP` runtime error discovered by @eundersander 
 * `AnyImageConverter` etc. plugins now propagate configuration options to the target plugins instead of silently ignoring them, improving usability
 * `StbImageConverter` can now detect desired file format from the filename like a proper grownup
 * Performance improvements in String APIs (btw, did you know that each call to `std::tolower()` involves a mutex lock and virtual dispatch? wonderful, isn't it)
 * Some string APIs got renamed for clarity, but this time nothing affects Habitat 
 * (Community work) Animation import in AssimpImporter, supporting FBX and COLLADA

## How Has This Been Tested

Suffered through some new and exciting CMake `try_compile()` issues involving Assimp and CMake subprojects. Hopefully will go fine on the third attempt. (The lab build failure seems to be something not caused by me.)

## Types of changes


Skylion007:
The Habitat-Lab CI is currently broken. So this should be fine.

Skylion007 (approved):

jturner65 (approved):
LGTM

PR #1368, opened 2021-07-01
Viewer Hot-swap scenes

jturner65:
## Motivation and Context
This PR adds the ability to cycle through all the SceneInstances specified in the current dataset, starting with the chosen scene on the command line.  By pressing Tab/Shift-Tab, the user can cycle to the next/previous SceneInstance without having to reload any configuration data or restart (about 5 seconds per ReplicaCAD scene).

## How Has This Been Tested
Locally - all changes are in viewer.cpp

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -264,7 +264,7 @@ Key Commands:
   esc: Exit the application.
   'H': Display this help message.
   'm': Toggle mouse mode.
-  TAB/Shift-TAB : Cycle to next/previous scene in scene datsaet (if --cycle-scenes argument specified)
+  TAB/Shift-TAB : Cycle to next/previous scene in scene datsaet
 
   Agent Controls:
   'wasd': Move the agent's body forward/backward, left/right.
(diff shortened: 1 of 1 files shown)


src/utils/viewer/viewer.cpp
```diff
@@ -633,66 +676,27 @@ Viewer::Viewer(const Arguments& arguments)
   Mn::GL::Renderer::enable(Mn::GL::Renderer::Feature::DepthTest);
   Mn::GL::Renderer::enable(Mn::GL::Renderer::Feature::FaceCulling);
 
-  sceneFileName = args.value("scene");
-  bool useBullet = args.isSet("enable-physics");
-  if (useBullet && (args.isSet("debug-bullet"))) {
+  // whether to cycle scene instances
+  cycleSceneInstances_ = args.isSet("cycle-scenes");
```
aclegg3, line 680: Can't we cycle on any scene dataset even if we specify to start with a particular one? Maybe the cycle-scenes replaces the default `-- scene` as a mutually exclusive option to override the initial, but otherwise behaves the same once loaded?

jturner65: That would require reloading a different scene-dataset and all the constituent metadata.  May as well restart.

jturner65: Oops, misunderstood.  This is a good idea - make it always available.

aclegg3 (approved):
Very useful tool for checking datasets, thanks!

PR #1369, opened 2021-07-01
datasets_download.py - change default data_path 

aclegg3:
## Motivation and Context

Change default data_path from git to ./data/ so the script has a reasonable default outside of git repo but is also compatible with previous default behavior in habitat-sim directory and doesn't flood an existing user directory. 

## How Has This Been Tested

Locally

## Types of changes

## Checklist

### DIFF
--- a/habitat_sim/utils/datasets_download.py
+++ b/habitat_sim/utils/datasets_download.py
@@ -505,21 +505,21 @@ def main(args):
     args = parser.parse_args(args)
     replace = args.replace
 
-    # get a default data_path from git
+    # get a default data_path "./data/"
     data_path = args.data_path
     if not data_path:
         try:
-            import git
-
-            repo = git.Repo(".", search_parent_directories=True)
-            dir_path = repo.working_tree_dir
-            # Root data directory. Optionaly overridden by input argument "--data-path".
-            data_path = os.path.join(dir_path, "data")
+            data_path = os.path.abspath("./data/")
+            print(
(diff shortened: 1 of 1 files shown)


habitat_sim/utils/datasets_download.py
```diff
@@ -505,21 +505,21 @@ def main(args):
     args = parser.parse_args(args)
     replace = args.replace
 
-    # get a default data_path from git
+    # get a default data_path "./data/"
     data_path = args.data_path
     if not data_path:
         try:
-            import git
```
mathfac, line 512: Maybe, we can do both, check git repo first, and then use `./data/` as fall back?

mathfac (approved):

habitat_sim/utils/datasets_download.py
```diff
@@ -505,21 +505,21 @@ def main(args):
...
-
-            repo = git.Repo(".", search_parent_directories=True)
-            dir_path = repo.working_tree_dir
-            # Root data directory. Optionaly overridden by input argument "--data-path".
-            data_path = os.path.join(dir_path, "data")
+            data_path = os.path.abspath("./data/")
+            print(
+                f"No data-path provided, default to: {data_path}. Use '--data-path' to specify another location."
+            )
+            if not os.path.exists(data_path):
```
Skylion007, line 516: This can be one line. os.makedirs(data_path, exists_ok=True)

PR #1376, opened 2021-07-08
[Web] Add flag to enable SIMD instructions in WASM

Skylion007:
## Motivation and Context
* Adds a flag that can enable building the Web build of Habitat to use SIMD. This should speed up physics and other linalg heavy vectorization quite significantly. The browser must support SIMD in WASM like modern versions of Chrome though.
* To enable pass the --simd flag to build_js.sh
* Also fixes a small bug with pre-commit that prevented shell check from running on build.sh and build_js.sh .

## How Has This Been Tested
* Locally and with CI.

## Types of changes

## Checklist

### DIFF
--- a/build_js.sh
+++ b/build_js.sh
@@ -11,8 +11,8 @@ while [[ "$#" -gt 0 ]]; do
     case $1 in
         --bullet) BULLET=true ;;
         --no-web-apps) WEB_APPS=false ;;
-	--simd) USE_SIMD=true ;;
-	*) echo "Unknown parameter passed: $1"; exit 1 ;;
+        --simd) USE_SIMD=true ;;
+        *) echo "Unknown parameter passed: $1"; exit 1 ;;
     esac
     shift
 done
--- a/build_js.sh
+++ b/build_js.sh
@@ -36,7 +36,7 @@ cd build_js
 EXE_LINKER_FLAGS="-s USE_WEBGL2=1"
 HAB_C_FLAGS="-s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1 -s ASSERTIONS=0"
 if ${USE_SIMD} ;
-    then HAB_C_FLAGS="${HAB_CXX_FLAGS} -msimd128"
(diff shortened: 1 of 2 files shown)


mosra:
> quite significantly

(With my skeptical hat on.) Do you have some numbers to back this? I'm interested in how much this helps in a codebase of this size.

Skylion007:
@mosra Probably not on Magnum, but WASM doesn't even enable SSE2 instructions without this flag for Bullet so it should give a speed up there.

mosra:
It's not as simple as "enabling SSE2" since WASM has to work on ARM as well -- and that's why I'm skeptical, because different platforms have different instructions and what could directly map to a SSE instruction might have to be emulated on NEON and vice versa.

But in any case I *really want to know* how this helps, seriously :) Did you try it out? I know from certain projects that hand-coded WASM SIMD can be four to six times times faster than scalar code, but have no idea about autovectorization, especially when combined with everything else we're running here. Is it 1%? 10%? 2x faster?

build_js.sh
```diff
@@ -4,12 +4,14 @@
 set -e
 
 BULLET=false
+USE_SIMD=false
```
eundersander, line 7: Some documentation would be nice; even just linking back to this PR.

eundersander (approved):
LGTM, thanks for doing this! I'm also curious to evaluate the speedup. @ldcWV It would be cool to try this with your upcoming JS physics benchmark.

ldcWV:
I tried this on my webxr hand demo benchmark, which drops a lot of objects in a big pile and tries to step physics 60 times per second (or 16.7 ms per stepWorld() call). I repeated the benchmark 3 times with and without the --simd flag and here were the results:

without --simd:
78.76
72.60
82.87

with --simd:
79.75
84.07
83.48

These numbers are the average ms between stepWorld() calls. Note that it's trying to achieve 16.67ms per stepWorld() call but cannot keep up. So it doesn't seem like this SIMD optimization has helped much in this case.

PR #1382, opened 2021-07-13
Subsystem Logging

erikwijmans:
## Motivation and Context

In #1184 we discussed having different logging subsystems for habitat-sim so that we can filter logs based on different subsystems.  This PR takes an attempt at implementing this.  The most important user facing part of this is how different logging filters are constructed with the environment variable (named `HABITAT_SIM_LOG`).  This simply applies a series of changes to the logging level for each system or a subset of systems, letting you construct an arbitrary filter.

The filter has a simple grammar
```
Filter: SetLevelCommand (COLON SetLevelCommand)*
SetLevelCommand: (SUBSYSTEM (COMMA SUBSYSTEM)* EQUALS)? LOGGING_LEVEL
```

So to have verbose logging on just the physics subset, you can do
```
quiet:physics=verbose
```
The "quiet" sets all subsets to subsystem and "physics=verbose" sets the just the physics subsystem to verbose.  If you want physics and gfx logging,
```
quiet:physics,gfx=verbose
```

Currently I have a couple subsystems added and a catch all "Other" subsystem.  The logging levels are "verbose,debug,warning,quiet,error" where `verbose` includes everything and `quiet` includes errors or worse.  The other names correspond to a given logger type.  A log level is printed if its logger type has a higher level than the current logging level for a subsystem.  So and `ESP_DEBUG()` get's printed if the logging level is "debug" or lower in the given subsystem.

This isn't 100% done yet, i.e. if folks like this we can just ditch glog, but I thought I would open this up for feedback.

Open questions:
[x] ~Should we ditch glog?~  Yes, in the next PR tho.
[x] ~What should the default level be?  Just warnings?~  Everything but very verbose things.
[x] ~Should we inject the subsystem name into the logging statement?  I think yes as that'll make it easier to turn noisy things off.~ Yes
[x] ~Would something like an `ESP_DEBUG_ALWAYS()` be useful?  That way you can just disable all logging and have rise above while debugging?~ `Mn::Debug{}` or `!Mn::Debug{}` does this.
[x] ~Should we change the default log level based on `GLOG_minloglevel` to enable backwards compatibility?~ No

## How Has This Been Tested

With the unit test.

## Types of changes

### DIFF
--- a/src/esp/core/logging.cpp
+++ b/src/esp/core/logging.cpp
@@ -44,7 +44,7 @@ LoggingLevel levelFromName(const Corrade::Containers::StringView name) {
   if (lowerCaseName == #name##_s) \
   return LoggingLevel::level
 
-  CASE(Verbose1, verbose1);
+  CASE(VeryVerbose, veryverbose);
   CASE(Verbose, verbose);
   CASE(Debug, debug);
   CASE(Warning, warning);
--- a/src/esp/core/logging.cpp
+++ b/src/esp/core/logging.cpp
@@ -1,0 +1,142 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "logging.h"
+#include "Check.h"
+
+#include <algorithm>
+#include <cstdlib>
+
+#include <Corrade/Containers/Array.h>
+#include <Corrade/Containers/StaticArray.h>
+#include <Corrade/Containers/String.h>
+#include <Corrade/Utility/Format.h>
+
+namespace Cr = Corrade;
+using Cr::Containers::Literals::operator""_s;
+
+namespace esp {
+namespace logging {
+
+static_assert(uint8_t(Subsystem::NumSubsystems) ==
+                  (sizeof(subsystemNames) / sizeof(subsystemNames[0])),
+              "Missing a subsystem name.");
+
+Subsystem subsystemFromName(const Corrade::Containers::StringView name) {
+  const Cr::Containers::String lowerCaseName =
+      Cr::Utility::String::lowercase(name);
+  for (uint8_t i = 0; i < uint8_t(Subsystem::NumSubsystems); ++i)
+    if (lowerCaseName == Cr::Utility::String::lowercase(
+                             Cr::Containers::StringView{subsystemNames[i]}))
+      return Subsystem(i);
+
+  CORRADE_ASSERT_UNREACHABLE("Unknown subsystem '"
+                                 << Cr::Utility::Debug::nospace << name
+                                 << Cr::Utility::Debug::nospace << "'",
+                             {});
+}
+
+LoggingLevel levelFromName(const Corrade::Containers::StringView name) {
+  const Cr::Containers::String lowerCaseName =
+      Cr::Utility::String::lowercase(name);
+#define CASE(level, name)         \
+  if (lowerCaseName == #name##_s) \
+  return LoggingLevel::level
+
+  CASE(Verbose1, verbose1);
+  CASE(Verbose, verbose);
+  CASE(Debug, debug);
+  CASE(Warning, warning);
+  CASE(Quiet, quiet);
+  CASE(Error, error);
+
+#undef CASE
+  CORRADE_ASSERT_UNREACHABLE("Unknown logging level name '"
+                                 << Cr::Utility::Debug::nospace << name
+                                 << Cr::Utility::Debug::nospace << "'",
+                             {});
+}
+
+namespace {
+#ifdef CORRADE_BUILD_MULTITHREADED
+CORRADE_THREAD_LOCAL
+#endif
+#if defined(MAGNUM_BUILD_STATIC_UNIQUE_GLOBALS) && \
+    !defined(CORRADE_TARGET_WINDOWS)
+/* On static builds that get linked to multiple shared libraries and then used
+   in a single app we want to ensure there's just one global symbol. On Linux
+   it's apparently enough to just export, macOS needs the weak attribute.
+   Windows handled differently below. */
+CORRADE_VISIBILITY_EXPORT
+#ifdef __GNUC__
+__attribute__((weak))
+#else
+/* uh oh? the test will fail, probably */
+#endif
+#endif
+LoggingContext* currentLoggingContext = nullptr;
+}  // namespace
+
+bool LoggingContext::hasCurrent() {
+  return currentLoggingContext != nullptr;
+}
+
+LoggingContext& LoggingContext::current() {
+  ESP_CHECK(hasCurrent(),
+            "esp::logging::LoggingContext: No current logging context.");
+
+  return *currentLoggingContext;
+}
+
+LoggingContext::LoggingContext(Corrade::Containers::StringView envString)
+    : loggingLevels_{Cr::DirectInit, uint8_t(Subsystem::NumSubsystems),
+                     DEFAULT_LEVEL},
+      prevContext_{currentLoggingContext} {
+  currentLoggingContext = this;
+  processEnvString(envString);
+}
+
+LoggingContext::LoggingContext()
+    : LoggingContext{std::getenv(LOGGING_ENV_VAR_NAME)} {}
+
+LoggingContext::~LoggingContext() {
+  currentLoggingContext = prevContext_;
+}
+
+void LoggingContext::processEnvString(
+    const Cr::Containers::StringView envString) {
+  for (const Cr::Containers::StringView setLevelCommand :
+       envString.split(':')) {
+    if (setLevelCommand.contains("=")) {
+      const auto parts = setLevelCommand.partition('=');
+      LoggingLevel lvl = levelFromName(parts[2]);
+
+      for (const Cr::Containers::StringView subsystemName : parts[0].split(','))
+        loggingLevels_[uint8_t(subsystemFromName(subsystemName))] = lvl;
+    } else {
+      std::fill(loggingLevels_.begin(), loggingLevels_.end(),
+                levelFromName(setLevelCommand));
+    }
+  }
+}
+
+void LoggingContext::reinitializeFromEnv() {
+  std::fill(loggingLevels_.begin(), loggingLevels_.end(), DEFAULT_LEVEL);
+  processEnvString(std::getenv(LOGGING_ENV_VAR_NAME));
+}
+
+LoggingLevel LoggingContext::levelFor(Subsystem subsystem) const {
+  return loggingLevels_[uint8_t(subsystem)];
+}
+
+bool isLevelEnabled(Subsystem subsystem, LoggingLevel level) {
+  return level >= LoggingContext::current().levelFor(subsystem);
+}
+
+Cr::Containers::String subsystemPrefix(Subsystem subsystem) {
+  return ""_s.join({"["_s, subsystemNames[uint8_t(subsystem)], "] "_s});
+}
+
+}  // namespace logging
+}  // namespace esp
--- a/src/esp/core/logging.h
+++ b/src/esp/core/logging.h
@@ -208,19 +208,21 @@ class LogMessageVoidify {
  * would work for this overload.  That way the logging statements get evaluated
  * first, then the voidify, then the ternary.
  */
-#define ESP_LOG_IF(condition, output) \
-  !(condition) ? static_cast<void>(0) \
-               : esp::logging::impl::LogMessageVoidify{} & (output)
+
+#define ESP_LOG_IF(condition, output)                      \
+  !(condition) ? static_cast<void>(0)                      \
+               : esp::logging::impl::LogMessageVoidify{} & \
+                     (output) /* NOLINT(bugprone-macro-parentheses) */
 
 // This ends with a nospace since the space is baked in to subsystemPrefix for
 // the case that the logger was created with a nospace flag.
-#define ESP_SUBSYS_LOG_IF(subsystem, level, output)                  \
-  ESP_LOG_IF(esp::logging::isLevelEnabled(subsystem, level), output) \
-      << esp::logging::subsystemPrefix(subsystem)                    \
+#define ESP_SUBSYS_LOG_IF(subsystem, level, output)                        \
+  ESP_LOG_IF(esp::logging::isLevelEnabled((subsystem), (level)), (output)) \
+      << esp::logging::subsystemPrefix((subsystem))                        \
       << Corrade::Utility::Debug::nospace
 
 #define ESP_LOG_LEVEL_ENABLED(level) \
-  esp::logging::isLevelEnabled(espLoggingSubsystem(), level)
+  esp::logging::isLevelEnabled(espLoggingSubsystem(), (level))
 
 #define ESP_VERBOSE_1(...)                                              \
   ESP_SUBSYS_LOG_IF(                                                    \
@@ -283,7 +285,7 @@ class LogMessageVoidify {
 #define ASSERT(x, ...)                                                   \
   do {                                                                   \
     if (!(x)) {                                                          \
-      ESP_ERROR(Corrade::Utility::Debug::Flag::NoSpace)                  \
+      Corrade::Utility::Error{Corrade::Utility::Debug::Flag::NoSpace}    \
           << "Assert failed: " #x << "," << __FILE__ << ":" << __LINE__; \
       exit(-1);                                                          \
     }                                                                    \
--- a/src/esp/core/logging.h
+++ b/src/esp/core/logging.h
@@ -209,10 +209,11 @@ class LogMessageVoidify {
  * first, then the voidify, then the ternary.
  */
 
-#define ESP_LOG_IF(condition, output) \
-  !(condition)                        \
-      ? static_cast<void>(0)          \
-      : esp::logging::impl::LogMessageVoidify{} & (output) /* NOLINT */
+// No formatting on this macro as the NOLINT for clang
+#define ESP_LOG_IF(condition, output)                                         \
+  !(condition)                                                                \
+      ? static_cast<void>(0) /* NOLINTNEXTLINE(bugprone-macro-parentheses) */ \
+      : esp::logging::impl::LogMessageVoidify{} & (output)
 
 // This ends with a nospace since the space is baked in to subsystemPrefix for
 // the case that the logger was created with a nospace flag.
--- a/src/esp/core/logging.h
+++ b/src/esp/core/logging.h
@@ -209,10 +209,10 @@ class LogMessageVoidify {
  * first, then the voidify, then the ternary.
  */
 
-#define ESP_LOG_IF(condition, output)                      \
-  !(condition) ? static_cast<void>(0)                      \
-               : esp::logging::impl::LogMessageVoidify{} & \
-                     (output) /* NOLINT(bugprone-macro-parentheses) */
+#define ESP_LOG_IF(condition, output) \
+  !(condition)                        \
+      ? static_cast<void>(0)          \
+      : esp::logging::impl::LogMessageVoidify{} & (output) /* NOLINT */
 
 // This ends with a nospace since the space is baked in to subsystemPrefix for
 // the case that the logger was created with a nospace flag.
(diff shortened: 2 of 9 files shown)


src/esp/sim/Simulator.cpp
```diff
@@ -197,28 +198,32 @@ Simulator::setSceneInstanceAttributes(const std::string& activeSceneName) {
...
-              << (pfSuccess ? "Navmesh Loaded." : "Navmesh load error.");
+    ESP_DEBUG() << "::setSceneInstanceAttributes : "
+                << (pfSuccess ? "Navmesh Loaded." : "Navmesh load error.");
   } else {
-    LOG(WARNING) << "::setSceneInstanceAttributes : Navmesh file not found, "
-                    "checked at filename : '"
-                 << navmeshFileLoc << "'";
+    ESP_DEBUG() << "::setSceneInstanceAttributes : Navmesh file not found, "
+                   "checked at filename : '"
+                << navmeshFileLoc << "'";
```
mosra, line 226:
```suggestion
                   "checked at filename : '" << Mn::Debug::nospace
                << navmeshFileLoc << Mn::Debug::nospace << "'";
```

src/esp/sim/Simulator.cpp
```diff
@@ -248,10 +253,10 @@ Simulator::setSceneInstanceAttributes(const std::string& activeSceneName) {
     // semantic scene descriptor might not exist, so
     semanticScene_ = nullptr;
     semanticScene_ = scene::SemanticScene::create();
-    LOG(INFO) << "::setSceneInstanceAttributes : SceneInstance : "
-              << activeSceneName
-              << " proposed Semantic Scene Descriptor filename : "
-              << semanticSceneDescFilename;
+    ESP_DEBUG() << "::setSceneInstanceAttributes : SceneInstance : "
+                << activeSceneName
+                << " proposed Semantic Scene Descriptor filename : "
```
mosra, line 258:
```suggestion
    ESP_DEBUG() << "::setSceneInstanceAttributes : SceneInstance :"
                << activeSceneName
                << "proposed Semantic Scene Descriptor filename :"
```

src/esp/sim/Simulator.cpp
```diff
@@ -266,16 +271,16 @@ Simulator::setSceneInstanceAttributes(const std::string& activeSceneName) {
...
       }
     }  // if given SSD file name specifiedd exists
-    LOG(WARNING) << "::setSceneInstanceAttributes : All attempts to load "
-                    "SSD with SceneAttributes-provided name "
-                 << semanticSceneDescFilename << " : exist : " << fileExists
-                 << " : loaded as expected type : " << success;
+    ESP_WARNING() << "::setSceneInstanceAttributes : All attempts to load "
+                     "SSD with SceneAttributes-provided name "
+                  << semanticSceneDescFilename << " : exist : " << fileExists
+                  << " : loaded as expected type : " << success;
```
mosra, line 283:
```suggestion
    ESP_WARNING() << "::setSceneInstanceAttributes : All attempts to load "
                     "SSD with SceneAttributes-provided name"
                  << semanticSceneDescFilename << ": exist :" << fileExists
                  << ": loaded as expected type :" << success;
```

src/esp/sim/Simulator.cpp
```diff
@@ -315,15 +320,15 @@ bool Simulator::createSceneInstance(const std::string& activeSceneName) {
...
+                << lightSetupKey << "-";
   } else {
     lightSetupKey = metadataMediator_->getLightSetupFullHandle(
         curSceneInstanceAttributes->getLightingHandle());
-    LOG(INFO) << "::createSceneInstance : Using scene instance-specified "
-                 "Light key : -"
-              << lightSetupKey << "-";
+    ESP_DEBUG() << "::createSceneInstance : Using scene instance-specified "
+                   "Light key : -"
+                << lightSetupKey << "-";
```
mosra, line 331:
```suggestion
                   "Light key : -" << Mn::Debug::nospace
                << lightSetupKey << Mn::Debug::nospace << "-";
```

src/esp/sim/Simulator.cpp
```diff
@@ -397,9 +402,9 @@ bool Simulator::createSceneInstance(const std::string& activeSceneName) {
     throw std::invalid_argument("::createSceneInstance : Cannot load: " +
                                 stageAttributesHandle);
   } else {
-    LOG(INFO) << "::createSceneInstance : Successfully loaded stage "
-                 "named : "
-              << stageAttributes->getHandle();
+    ESP_DEBUG() << "::createSceneInstance : Successfully loaded stage "
+                   "named : "
```
mosra, line 406:
```suggestion
                   "named :"
```

src/esp/sim/Simulator.cpp
```diff
@@ -883,9 +888,9 @@ int Simulator::addTrajectoryObject(const std::string& trajVisName,
     return ID_UNDEFINED;
   }
   auto trajObj = getRigidObjectManager()->getObjectCopyByID(trajVisID);
-  LOG(INFO) << "::showTrajectoryVisualization : Trajectory "
-               "visualization object created with ID "
-            << trajVisID;
+  ESP_DEBUG() << "::showTrajectoryVisualization : Trajectory "
+                 "visualization object created with ID "
```
mosra, line 892:
```suggestion
                 "visualization object created with ID"
```

src/esp/core/test/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+find_package(Corrade REQUIRED Utility)
+
+corrade_add_test(coreLoggingTest LoggingTest.cpp LIBRARIES core Corrade::Utility)
```
mosra, line 3: Linking `Corrade::Utility` explicitly won't hurt, it's just done implicitly (it's a dependency of `TestSuite`).

mosra (commented):
Thanks, this is great!

mosra:
> Should we ditch glog?

YES PLEASE :grin: Also see #121 :)

> What should the default level be?

I'd say print everything, otherwise we quickly get to a point where the logs become uselessly verbose. A valid point is also that the printing is not completely zero-cost even if disabled, so printing 1000 disabled verbose messages every frame will slow things down and you wouldn't know why.

> Should we inject the subsystem name into the logging statement?

I think that'd be a good idea.

> Would something like an `ESP_DEBUG_ALWAYS()` be useful?

I'd say directly use `!Debug{}` instead for that purpose? That gives you also file and line and you can use `!Debug{};` alone to perform easy "it got here"-style printf debugging :)

> Should we change the default log level based on `GLOG_minloglevel` to enable backwards compatibility?

Personally I never used this, and I think there's a lot of people on the team that don't even know this was a thing. So I'd say don't bother.

src/esp/core/Check.h
```diff
@@ -63,15 +63,16 @@ extern void (*throwInPython)(const char*);
...
-          << "ESP_CHECK failed:" << __VA_ARGS__;                  \
-      esp::core::throwIfInPythonOtherwiseAbort(out.str().data()); \
-    }                                                             \
+#define ESP_CHECK(condition, ...)                                            \
+  do {                                                                       \
+    if (!(condition)) {                                                      \
+      std::ostringstream out;                                                \
+      Corrade::Utility::Debug{                                               \
+          &out, Corrade::Utility::Debug::Flag::NoNewlineAtTheEnd}            \
+          << "ESP_CHECK(" << Corrade::Utility::Debug::nospace << #condition  \
```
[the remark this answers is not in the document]
erikwijmans: Ah yes, I added this as I like it (made it easy to make sure the exact assert I thought should fire does without having to remember what exact error language I used) but it's not that great for users

src/esp/core/logging.cpp
```diff
@@ -0,0 +1,128 @@
...
+  std::fill(loggingLevels_.begin(), loggingLevels_.end(), DEFAULT_LEVEL);
+
+  for (const Cr::Containers::StringView setLevelCommand :
+       envString.split(';')) {
+    if (setLevelCommand.contains("=")) {
+      const auto parts = setLevelCommand.partition('=');
+      LoggingLevel lvl = levelFromName(parts[2]);
+
+      for (const Cr::Containers::StringView subsystemName : parts[0].split(','))
+        loggingLevels_[uint8_t(subsystemFromName(subsystemName))] = lvl;
```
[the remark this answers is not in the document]
erikwijmans: They are great!  Made doing this as trivial as it is in python :)

src/esp/core/logging.h
```diff
@@ -7,6 +7,117 @@
...
+
+enum class LoggingLevel : uint8_t {
+  // Always first
+  verbose,
+  debug,
+  warning,
+  // Quiet mode disables debug and warning
+  quiet,
+  error,
+};
```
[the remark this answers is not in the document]
erikwijmans: The intent here is that verbose always gives you everything and quiet always silences everything besides errors or worse.  That way we can add new levels without changing things downstream.  verbose and quiet are the two I expect users to use while the others may be useful in specific circumstances.

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -195,16 +195,16 @@ int BulletPhysicsManager::addArticulatedObjectFromURDF(
               .first)
           .first;
 
-  Magnum::Debug{} << "BulletPhysicsManager::addArticulatedObjectFromURDF: "
-                     "simpleObjectHandle : "
-                  << simpleArtObjHandle;
+  ESP_DEBUG() << "BulletPhysicsManager::addArticulatedObjectFromURDF: "
```
erikwijmans, line 198:
```suggestion
  ESP_DEBUG() << "BulletPhysicsManager::addArticulatedObjectFromURDF:"
```

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -195,16 +195,16 @@ int BulletPhysicsManager::addArticulatedObjectFromURDF(
...
+                 "simpleObjectHandle : "
+              << simpleArtObjHandle;
 
   std::string newArtObjectHandle =
       articulatedObjectManager_->getUniqueHandleFromCandidate(
           simpleArtObjHandle);
-  Magnum::Debug{} << "BulletPhysicsManager::addArticulatedObjectFromURDF: "
-                     "newArtObjectHandle : "
-                  << newArtObjectHandle;
+  ESP_DEBUG() << "BulletPhysicsManager::addArticulatedObjectFromURDF: "
```
erikwijmans, line 205:
```suggestion
  ESP_DEBUG() << "BulletPhysicsManager::addArticulatedObjectFromURDF:"
```

src/esp/core/logging.cpp
```diff
@@ -0,0 +1,128 @@
...
+                             {});
+}
+
+CORRADE_THREAD_LOCAL LoggingSubsystemTracker*
+    LoggingSubsystemTracker::instance = nullptr;
+LoggingSubsystemTracker& LoggingSubsystemTracker::Instance() {
+  if (!instance)
+    instance = new LoggingSubsystemTracker{};
+  return *instance;
+}
```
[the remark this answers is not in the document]
erikwijmans: Gave this a try

src/esp/core/logging.h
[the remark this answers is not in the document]
mosra: In other words, just so I understand correctly, the following would be functionally equivalent, right?

```cpp
enum class LoggingLevel : uint8_t {
  debug,
  warning,
  error,
  
  // Verbose is debug and all higher levels
  verbose = debug,
  
  // Quiet is only errors and nothing else
  quiet = error
};
```

src/esp/core/logging.h
```diff
@@ -7,6 +7,141 @@
...
+// to add an appropriate loggingSubsystem function (see the ADD_SUBSYSTEM_FN
+// helper bellow) and add it to @ref subsystemFromName()
+enum class Subsystem : uint8_t {
+  gfx,
+  scene,
+  sim,
+  physics,
+  nav,
+  // This is the catch all subsystem
+  Default,
```
jturner65, line 28: Can we add a metadata subsystem?

src/esp/core/logging.h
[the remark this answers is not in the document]
jturner65: This is excellent! A few questions/comments

1. I would like to see an "info" level - debug is, to me, a loaded term.  An example of an "info"-level message would be some output of a successful transaction that would have resulted in a warning/error otherwise, for example - perhaps not as detailed as "debug" info, and probably not used in the same contexts (with "debug" documenting potential errors vs "info" giving non-negative feedback).
2. How hard would it be to add subsystem-level/logging-level stream routing (i.e. send certain subsystem and/or log-level output to certain log files instead of the console)? I know this would probably be beyond the scope of this PR.  I just would think something like this would be useful by giving folks a way to check on continuing processes without having to search the console output.

jturner65 (approved):
This is really great Erik, thanks so much for taking this on.  My thoughts on your open questions :     

> - Should we ditch glog?

Yes please.  The system presented here seems sufficient to make glog redundant.

> -    What should the default level be? Just warnings?

Depends on how we want to present this to folks. On the one hand, if we make it obnoxious, say setting "verbose" as the default, this will motivate most folks to put in a few minutes to learn the system to set it the way they want, and in the process gain some understanding of the logging system's capabilities. On the other hand, if we make it convenient and comfortable for folks, then they may not even be aware that such capabilities exist.

> -    Should we inject the subsystem name into the logging statement? I think yes as that'll make it easier to turn noisy things off.

Yes.  

> -    Would something like an ESP_DEBUG_ALWAYS() be useful? That way you can just disable all logging and have rise above while debugging?

I'm not sure I understand the purpose of this. Can you explain?

> -    Should we change the default log level based on GLOG_minloglevel to enable backwards compatibility?

Will this still be around if we dispose of glog? If we don't dispose of glog, then yes, I think this should be consumed by the new system as well.
______________

I have a few comments below, but I think this is great, and I'd like to see this added as soon as possible.

src/esp/core/logging.h
erikwijmans: Yep

src/esp/core/logging.h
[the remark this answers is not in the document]
erikwijmans: @mosra, yeah, those are equivalent.  I'll change it to look like that as that makes it clearer.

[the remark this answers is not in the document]
erikwijmans: @jturner65 I am currently matching Magnum names that what it is clear which of the Corrade's loggers gets returned by the macro (since they can be configured differently).  Given @mosra's point that even turned off logging isn't free, we should be wary of adding a level that is turned off by default to not inadvertently tank perform due to adding lots of logging statements.

It should be pretty simple to route to a file.  Corrade's loggers just take an ostream point, which can easily be a file.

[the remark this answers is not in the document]
jturner65: > Given @mosra's point that even turned off logging isn't free, we should be wary of adding a level that is turned off by default to not inadvertently tank perform due to adding lots of logging statements.

That's a valid point. Do we want to add compiler directives around non-warning/error log messages, so they are not even compiled if we do not wish to support them?  or is that overkill do you think?

src/esp/core/test/LoggingTest.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  void envVarTest();
+};
+
+constexpr const struct {
+  const char* envString;
+  const char* expected;
+} EnvVarTestData[]{
+    {nullptr,
+     "[Subsystem: Default] DebugDefault\n[Subsystem: Default] WarningDefault\n"
```
eundersander, line 60: The `Subsystem: ` prefix seems awfully verbose. I also wonder if `[Subsystem: Default]` is worth omitting altogether for brevity's sake.

src/esp/sim/Simulator.cpp
```diff
@@ -95,6 +96,8 @@ void Simulator::close(const bool destroy) {
 }
 
 void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
+  if (!loggingContext_)
```
eundersander, line 99: I don't see braceless if statements much in our codebase. I would hope they are taboo.

eundersander (approved):
Left some minor comments but otherwise LGTM! Thanks for doing this!

src/esp/core/test/LoggingTest.cpp
erikwijmans: Yeah, I made it `[<SubsystemName>]` in my current working branch.

src/esp/core/test/LoggingTest.cpp
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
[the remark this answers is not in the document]
erikwijmans: I like one file per test class.  I don't have a strong preference over putting things in `src/tests` vs a folder in the part of the source tree.  We currently do both in the code base and I slightly prefer keeping the tests next to what they test.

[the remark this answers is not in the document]
mosra: I personally prefer smaller tests located next to the code they test as well. The corresponding test is easier to localize and it's also clearer to see what parts are tested and what not.

> We currently do both

I suppose I'm to blame for this inconsistency, as I started putting certain tests into `src/esp/gfx/test/` :) If I would be in charge, I'd relocate all existing tests to be next to the code they test.

src/esp/core/esp.h
```diff
@@ -106,6 +137,14 @@ inline std::ostream& operator<<(std::ostream& os, const box3f& bbox) {
   return os << "{min:" << bbox.min() << ",max:" << bbox.max() << "}";
 }
 
+inline Corrade::Utility::Debug& operator<<(Corrade::Utility::Debug& out,
+                                           const box3f& bbox) {
+  using Corrade::Utility::Debug;
+  return out << "{min:" << Debug::nospace << bbox.min() << Debug::nospace
+             << ",max:" << Debug::nospace << bbox.max() << Debug::nospace
+             << "}";
+}
```
mosra, line 146: Oh, Eigen has AABBs? Interesting, missed those.

Continuing with the suggestion from above -- as of mosra/magnum-integration@789f8132842536f077548bc4f2d2c3685e9f491e, converting to a `Range3D` in order to print it with Debug should work as well :)

mosra (commented):
I'm thinking that maybe the introduction of logging macros and glog replacement/removal should be separate PRs because the actual essence of the PR is kinda hiden among all the dry replacement, making it hard to see what needs attention and what not :)

src/esp/core/logging.h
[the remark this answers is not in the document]
mosra: @jturner65 we figured out a way how to make them ~free, after all, with some *cursed* code :)

erikwijmans:
Removed the glog removal changes.  Will make a new PR with those.

src/esp/core/logging.h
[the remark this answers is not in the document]
erikwijmans: See `_ESP_LOG_IF` for the craziness :)

src/esp/core/logging.h
```diff
@@ -62,8 +284,8 @@ class LogMessageVoidify {
 #define ASSERT(x, ...)                                              \
   do {                                                              \
     if (!(x)) {                                                     \
-      LOG(ERROR) << "Assert failed: " #x << ", " << __FILE__ << ":" \
-                 << __LINE__;                                       \
+      ESP_ERROR() << "Assert failed: " #x << "," << __FILE__ << ":" \
+                  << __LINE__;                                      \
       exit(-1);                                                     \
```
mosra: That reminds me, earlier we went and replaced most (all?) `ASSERT()`s with `CORRADE_ASSERT()` and/or `ESP_CHECK()`, what should we do with those? Should they stay or should they get reworked on top of this new logging system as well?

Personally I'd stay with the current implementation of `CORRADE_ASSERT()` / `ESP_CHECK()`. Since it's not really meant for logging, I think we don't need any of the features like subsystem prefix or output redirection -- conversely, we want it to *never* be redirected, and that's what the "vanilla" macros do. (Plus you'd have to rework all the `INTERNAL`/`UNREACHABLE`/... variants here, I don't think that's worth the time.)

erikwijmans: Yep, my thought was to stick with `CORRADE_ASSERT()` / `ESP_CHECK()` as is since those should never be silenced

mosra (approved):
Happy with how this shaped up, and thanks for the detailed documentation of how the internals work :+1:

This might cause some new doxygen warnings about undocumented classes and such, so you might want to add docblocks there. OTOH a lot of existing code has the same issues, so fixing those would be just a drop in the ocean anyway? :)

PR #1390, opened 2021-07-16
[PBR][IBL] render an equirectangular image to cubemap

bigbike:
## Motivation and Context
As titled.
Given an HDRi image, we would like to use it as the "envionment" to light up the scene.
The 1st step is to render it to a cubemap.

This shader is to do it, and it will be used later in the `PbrImageBasedLighting` class.

## Types of changes

## Checklist


bigbike:
lythwood_room

Picture is from: https://polyhaven.com/

CC0 license, completely free. see details here:
https://polyhaven.com/license

The cubemap results (Yes, there are up-side-down, which is totally CORRECT) :
environment rgba mip_2 -Z
environment rgba mip_2 +Z
environment rgba mip_2 -Y
environment rgba mip_2 +Y
environment rgba mip_2 -X
environment rgba mip_2 +X

All the functions are fully tested in the "ibl" branch.
see #1389

src/esp/gfx/PbrEquiRectangularToCubeMapShader.cpp
```diff
@@ -0,0 +1,91 @@
...
+  Mn::GL::Shader frag{glVersion, Mn::GL::Shader::Type::Fragment};
+
+  // Add macros
+  vert.addSource("#define OUTPUT_UV\n").addSource(rs.get("bigTriangle.vert"));
+
+  std::stringstream outputAttributeLocationsStream;
+  outputAttributeLocationsStream << Cr::Utility::formatString(
+      "#define OUTPUT_ATTRIBUTE_LOCATION_COLOR {}\n", ColorOutput);
+
+  frag.addSource(outputAttributeLocationsStream.str())
```
mosra, line 60: :eyes: why

```suggestion
  frag.addSource(Cr::Utility::formatString(
      "#define OUTPUT_ATTRIBUTE_LOCATION_COLOR {}\n", ColorOutput))
```

src/esp/gfx/PbrEquiRectangularToCubeMapShader.h
```diff
@@ -0,0 +1,75 @@
...
+#include <Magnum/GL/AbstractShaderProgram.h>
+#include <Magnum/GL/CubeMapTexture.h>
+#include <Magnum/GL/Texture.h>
+#include <Magnum/Shaders/GenericGL.h>
+
+#include <string>
+
+namespace esp {
+namespace gfx {
+class PbrEquiRectangularToCubeMapShader
```
mosra, line 17: ~~Docs? What is this for? How is it meant to be used?~~

Only after seeing the PR description I was able to understand what this does and why it is needed. And now I wonder, do we need to have this code *in the Habitat runtime* at all? 

As I see it, this shader is for offline one-shot processing of 3rd party images, we never have a need to use it at runtime, because at runtime we are able to generate the envmap as a cubemap directly, without having to go through an equirectangular image. 

For the 3rd party images, a lot of the HDR envmaps are provided as either a cubemap or as an equirectangular map already anyway, and if not, there are existing tools to convert from one to the other -- one of them is `exrenvmap`, for example. In conclusion, I don't think we need any of this code at all?

mosra (changes requested):

src/esp/gfx/PbrEquiRectangularToCubeMapShader.h
bigbike: Careful: when I say equirectangular image here, I mean the HDRi image that is directly downloaded from website such as: https://polyhaven.com/. It has nothing to do with the "equirectangular" image related to our equirectangular sensor. (Should I change the class name to `PbrHDRiToCubeMapShader`?)
I am doing: HDRi --> Cubemap here and I am NOT doming HDRi --> Equirectangular --> Cubemap.

We need this functionality to provide the out-of-box experience for the user, the same as the blender.
From a user's perspective: the most common case is that she only have a HDRi image on hand, and she would like to use it in our simulator.
The ideal, easiest way is: specify the file location of the image (a HDRi image) in the scene_instance config. That is it. DONE. The simulator will handle from there. And my diff is working towards this direction.

> As I see it, this shader is for offline one-shot processing of 3rd party images, we never have a need to use it at runtime

We need it at run-time, so that people can directly use the HDRi images she has, try them quickly without any pre-processing (HDRi -> 6 sides of cubemap).

> For the 3rd party images, a lot of the HDR envmaps are provided as either a cubemap or as an equirectangular map already anyway

Usually (99% cases) the HDR envmaps are in equirectangular map. And we need to convert it to a cubemap in the run-time as you mentioned.

mosra: (Just for the record: we discussed this extensively over a call and came to a conclusion that preprocessing to a cubemap first is the easier solution.)

mosra (approved):

PR #1407, opened 2021-07-26
Move physics off the main thread in WebXR Hand Demo

ldcWV:
## Motivation and Context

This PR is an optimization for the WebXR Hand Demo. It uses a Web Worker to do all the physics simulation on a separate thread. The results of the simulation are constantly sent to the main thread in the form of a GfxReplay Keyframe. The main thread renders frames as fast as possible (taking into account changes in the user's pose), and whenever a new Keyframe is received, it uses it to update the positions of the objects.

~~A side effect of these changes is that the user can no longer use their hands to spawn/manipulate objects. This feature can be readded by using postMessage to communicate hand grab/release/spawn requests to the worker.~~
Update: hand functionality has been added back.

## How Has This Been Tested

A [benchmark](https://github.com/facebookresearch/habitat-sim/pull/1387) was used to evaluate the speedup.
Time between frame renders
Time between stepWorld calls

Side by side comparison of before and after the optimization:

https://user-images.githubusercontent.com/4172272/127229360-85699310-27f2-4a84-a0e4-ff6f92f8e38d.mp4

## Types of changes

## Checklist


examples/web_apps/webxr_hand_demo/js/physics_worker_setup.js
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+/* global FS, importScripts, PhysicsWorker */
+
+function preloadFunc(url) {
```
ldcWV, line 7: As far as I know, there is no easy way to get this function other than copy paste. This is located in utils.js but as a module export, but since physics_worker_setup.js cannot be a module file, it cannot use that function. So I just copy pasted it here instead.

examples/web_apps/webxr_hand_demo/js/vr_demo.js
```diff
@@ -65,39 +61,67 @@ export class VRDemo {
...
-      preloadFunc(DataUtils.getObjectCollisionGlbFilepath(name));
-      preloadFunc(DataUtils.getObjectConfigFilepath(name));
+      doPreload(DataUtils.getObjectFilepath(name));
+      doPreload(DataUtils.getObjectCollisionGlbFilepath(name));
+      doPreload(DataUtils.getObjectConfigFilepath(name));
     }
   }
 
+  applyKeyframe(jsonKeyframe) {
+    let keyframe = this.player.keyframeFromString(jsonKeyframe);
```
eundersander, line 97: How about a version of `applyKeyframe` that takes a JSON keyframe? Then we never have to construct a Javascript Keyframe object, right? Maybe it'll faster.

With that said, I don't want you to entirely remove the Keyframe bindings and related functionality you've shown here. I can imagine a future use where the app wants to manipulate the keyframe in Javascript.

src/esp/gfx/replay/Player.h
```diff
@@ -101,10 +107,14 @@ class Player {
    */
   const std::vector<Keyframe>& debugGetKeyframes() const { return keyframes_; }
 
+  /**
+   * @brief Applies a keyframe to the player.
+   */
+  void applyKeyframe(const Keyframe& keyframe);
```
eundersander, line 113: I'm flexible on all the JS code in this PR, but for the Player class, we need to rethink your API changes.

It's dangerous to give users the freedom to manage and apply their own keyframes, because keyframes are "incremental"--they generally only store the changes since the previous frame.

src/esp/gfx/replay/ReplayManager.h
```diff
@@ -47,6 +47,11 @@ class ReplayManager {
    */
   std::shared_ptr<Player> readKeyframesFromFile(const std::string& filepath);
 
+  /**
+   * @brief Returns an empty Player object.
```
eundersander, line 51: How about a better comment explaining why you'd want an empty player object?

eundersander (commented):
Good progress here. I left some comments. I played this on Quest 2 and appreciated the improved framerate!

examples/web_apps/webxr_hand_demo/js/physics_worker_setup.js
Skylion007: You can use Webpack (which we already use for bundling the JS) to copy and paste it the worker as an additional target. That would be the best way of doing it @ldcWV . I think you can also use requires / imports across webworkers.

Skylion007: You can also just eval / import other JS scripts to prevent duplicating the function.

src/esp/gfx/replay/Player.cpp
```diff
@@ -19,6 +19,14 @@ void Player::readKeyframesFromJsonDocument(const rapidjson::Document& d) {
   esp::io::readMember(d, "keyframes", keyframes_);
 }
 
+Keyframe Player::keyframeFromString(const std::string& keyframe) {
+  Keyframe res;
+  rapidjson::Document d;
+  d.Parse<0>(keyframe.c_str());
+  esp::io::readMember(d, "keyframe", res);
```
Skylion007, line 26: Is d trivial to copy? Otherwise, might  want to be:
```suggestion
  esp::io::readMember(std:move(d), "keyframe", res);
```

Skylion007 (changes requested):

examples/web_apps/webxr_hand_demo/js/physics_worker_setup.js
eundersander: >You can use Webpack
I'd prefer to avoid Webpack for this example app. It adds complexity and makes iteration slower (due to build time). This is just an example app so optimizing browser load-time isn't a priority.

src/esp/gfx/replay/Recorder.h
```diff
@@ -66,6 +66,11 @@ class Recorder {
    */
   void saveKeyframe();
 
+  /**
+   * @brief Returns the last saved keyframe.
+   */
+  Keyframe getLatestKeyframe();
```
eundersander, line 72: How about a const reference instead? This is a large object with several child containers, so a copy is expensive. I guess the python binding will eventually have to make a deep copy (that can't be helped), but we shouldn't encourage copying here in the C++ api.

src/esp/gfx/replay/Player.h
ldcWV: That makes sense, I have changed it to better demonstrate this "incremental" idea.

ldcWV:
> What about doing the web worker as a module? https://stackoverflow.com/questions/44118600/web-workers-how-to-import-modules

Unfortunately, this isn't possible as I also need to use `importScripts` for `hsim_bindings.js`. This command is not available for JS modules.

examples/web_apps/webxr_hand_demo/js/vr_demo.js
```diff
@@ -65,39 +61,67 @@ export class VRDemo {
...
+    let preloadInfo = {
+      physicsConfigFilepath: DataUtils.getPhysicsConfigFilepath(),
+      stageFilepath: DataUtils.getStageFilepath(Module.stageName),
+      objectBaseFilepath: DataUtils.getObjectBaseFilepath(),
+      preloadedFiles: preloadedFiles
+    };
+    this.workerThread.postMessage({ type: "preloadInfo", value: preloadInfo });
+    this.workerThread.onmessage = function(e) {
+      if (e.data.type == "ready") {
+        setup();
```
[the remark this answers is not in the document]
ldcWV: Oops yeah, this isn't needed.

examples/web_apps/webxr_hand_demo/js/vr_demo.js
```diff
@@ -175,9 +180,25 @@ export class VRDemo {
 
     this.webXRSession.requestAnimationFrame(this.drawVRScene.bind(this));
 
-    this.physicsStepFunction = setInterval(() => {
-      this.sim.stepWorld(1.0 / 60);
-    }, 1000.0 / 60);
+    const handFilepathsByHandIndex = [
```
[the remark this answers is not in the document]
ldcWV: The reason for this is that I am relying on the renderer to dictate everything to do with the objects files, such as what the stage is, what the object spawn order is, and in this case, what the hand models are. Stuff like the stage and object spawn order MUST be decided by the renderer because it needs to preload them first.

Indeed, the hand models are a different case and they can be done within the worker, but I don't like that it would be inconsistent with how the other objects are loaded. Also, the renderer would still need to preload the hand models, so in any case it needs to reference "hand_l_open" anyways.

examples/web_apps/webxr_hand_demo/js/physics_worker_setup.js
ldcWV: > You can also just eval / import other JS scripts to prevent duplicating the function.

Hey, I'm not sure what you mean by this. Do you mind elaborating?

As far as I know, `eval` is a function that takes in JS code as a string and executes it. Are you suggesting there is some way to get this code as a string without copy paste?

Skylion007: Yeah, the code can be hosted remotely too.

ldcWV: Wouldn't that require duplication anyways? The issue is that this `preload` function is already in `utils.js` but I cannot use it because it has an `export`. I don't see how making another JS file would resolve this.

src/esp/gfx/replay/Player.cpp
```diff
@@ -172,6 +172,14 @@ void Player::applyKeyframe(const Keyframe& keyframe) {
   }
 }
 
+void Player::pushKeyframe(const Keyframe& keyframe) {
```
eundersander, line 175: Let's avoid the keyframe copy here. Let's move it into the vector. How about `Keyframe&& keyframe` and then `keyframes_.emplace_back(std::move(keyframe))`? I'm guessing a user doesn't need to keep around their keyframe after calling this.

src/esp/gfx/replay/Player.h
```diff
@@ -108,11 +108,17 @@ class Player {
   const std::vector<Keyframe>& debugGetKeyframes() const { return keyframes_; }
 
   /**
-   * @brief Applies a keyframe to the player.
+   * @brief Appends a Keyframe to the keyframe list.
    */
-  void applyKeyframe(const Keyframe& keyframe);
+  void pushKeyframe(const Keyframe& keyframe);
```
eundersander, line 113: `appendKeyframe` instead? I'm not sure `push` conveys adding to the back (stl has push_front, for example).

examples/web_apps/webxr_hand_demo/js/physics_worker.js
```diff
@@ -61,16 +79,21 @@ class PhysicsWorker {
...
     };
 
-    // physics step
+    // Step physics every 1/60s.
     this.physicsStepFunction = setInterval(() => {
       if (this.handInfo) {
         this.operateHands();
       }
+
       this.sim.stepWorld(0.003);
```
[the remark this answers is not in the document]
eundersander: If you prefer, you could wait and do this as a separate PR.

eundersander (commented):
I left a few more comments.

eundersander (approved):
LGTM. Check with @Skylion007 and make sure CI is passing before merging.

PR #1412, opened 2021-07-29
[PBR][IBL] Upgrade the cubemap

bigbike:
## Motivation and Context
Refactor the cubemap so that it can be used in the PBR IBL.
- Allow the developer to compute and customize the mipmap of each level by herself.
- Enable multiple drawable groups (with different names as index) in the scene graph.
- Upgrade the sensor, the CubeMapCamera accordingly.

## Types of changes

## Checklist


src/esp/scene/SceneGraph.h
```diff
@@ -27,12 +27,11 @@ class SceneGraph {
   SceneNode& getRootNode() { return rootNode_; }
   const SceneNode& getRootNode() const { return rootNode_; }
 
-  // TODO: remove this
-  gfx::DrawableGroup& getDrawables() {
-    return drawableGroups_.at(std::string{});
+  gfx::DrawableGroup& getDrawables(const char* groupName = "") {
```
Skylion007, line 30: Why char pointers as args instead of std::string?

bigbike: I am passing a string literal to this function, so I use `const char*` here. It would be faster if `std::string` can be avoided. 
But here it will be converted to the `std::string` anyway, so I guess you are right, I can just use string directly.

src/esp/scene/SceneGraph.h
```diff
@@ -27,11 +27,11 @@ class SceneGraph {
...
   const SceneNode& getRootNode() const { return rootNode_; }
 
-  gfx::DrawableGroup& getDrawables(const char* groupName = "") {
-    return drawableGroups_.at(std::string{groupName});
+  gfx::DrawableGroup& getDrawables(std::string groupName = "") {
+    return drawableGroups_.at(groupName);
   }
-  const gfx::DrawableGroup& getDrawables(const char* groupName = "") const {
-    return drawableGroups_.at(std::string{groupName});
+  const gfx::DrawableGroup& getDrawables(std::string groupName = "") const {
```
Skylion007: Actually, shoudln't these be const refs?

src/esp/scene/SceneGraph.h
Skylion007: @bigbike Couldn't you use corrade::stringview or something to avoid a copy?

bigbike: >corrade::stringview 

@Skylion007 : No, that is exactly for the C string literals. To avoid the copy then the data structure should not use `std::string` as the key which I doubt.

src/esp/scene/SceneGraph.h
bigbike: I do not think so since reference is rvalue.

bigbike: Wait, it is `const string&` so it would work (compile).
But it will not help since it will construct a string anyway (because the input parameter is a string literal.)

src/esp/gfx/CubeMap.cpp
```diff
@@ -34,14 +34,17 @@ const Mn::GL::Framebuffer::ColorAttachment rgbaAttachment =
...
+    Mn::GL::Framebuffer::ColorAttachment{2};
 
 /**
  * @brief check if the class instance is created with corresponding texture
  * enabled
  */
-void textureTypeSanityCheck(CubeMap::Flags& flag,
-                            CubeMap::TextureType type,
-                            const std::string& functionNameStr) {
+void textureTypeSanityCheck(const std::string& functionNameStr,
```
mosra, line 45: I might be overdoing the optimizations, but since the string is used only for assertions and thus should be as zero-overhead as possible, maybe go with just

```suggestion
void textureTypeSanityCheck(const char* functionNameStr,
```

instead?

src/esp/gfx/CubeMap.cpp
```diff
@@ -64,12 +67,44 @@ void textureTypeSanityCheck(CubeMap::Flags& flag,
...
+      break;
+
     case CubeMap::TextureType::Count:
       break;
   }
   CORRADE_INTERNAL_ASSERT_UNREACHABLE();
 }
 
+/** @brief do a couple of sanity checks based on mipLevel value */
+void mipLevelSanityCheck(const std::string& msgPrefix,
```
mosra, line 85: Same as above, maybe this instead?

```suggestion
void mipLevelSanityCheck(const char* msgPrefix,
```

src/esp/gfx/CubeMap.cpp
```diff
@@ -120,6 +158,9 @@ Mn::PixelFormat getPixelFormat(CubeMap::TextureType type) {
     case CubeMap::TextureType::ObjectId:
       return Mn::PixelFormat::R32UI;
       break;
+    case CubeMap::TextureType::VarianceShadowMap:
+      return Mn::PixelFormat::RG32F;
+      break;
     case CubeMap::TextureType::Count:
       break;
   }
```
mosra, line 166: While you're at it, I think everything above including the attachment constants could be wrapped in a `namespace {` to avoid potential symbol clashes and help the linker a bit?

src/esp/gfx/CubeMap.cpp
```diff
@@ -201,64 +265,84 @@ void CubeMap::recreateTexture() {
...
 
-void CubeMap::recreateFramebuffer() {
-  Mn::Vector2i viewportSize{imageSize_, imageSize_};
-  for (int iFbo = 0; iFbo < 6; ++iFbo) {
-    frameBuffer_[iFbo] = Mn::GL::Framebuffer{{{}, viewportSize}};
-  }
+void CubeMap::recreateFramebuffer(unsigned int cubeSideIndex,
+                                  unsigned int framebufferSize) {
+  Mn::Vector2i viewportSize{int(framebufferSize),
+                            int(framebufferSize)};  // at mip level 0
```
mosra, line 271: Code golf -- the single-argument constructor is there exactly for this:

```suggestion
  Mn::Vector2i viewportSize{int(framebufferSize)};  // at mip level 0
```

About the casts, I ended up using `Vector2i` instead of `Vector2ui` for sizes because the extra annoyance with unsigned types wasn't worth it. So maybe here just `int framebufferSize` could be fine as well.

src/esp/gfx/CubeMap.cpp
```diff
@@ -201,64 +265,84 @@ void CubeMap::recreateTexture() {
...
 
 void CubeMap::prepareToDraw(unsigned int cubeSideIndex,
-                            RenderCamera::Flags renderCameraFlags) {
+                            RenderCamera::Flags renderCameraFlags,
+                            unsigned int mipLevel) {
+  mipLevelSanityCheck("CubeMap::prepareToDraw():", flags_, mipLevel,
+                      mipmapLevels_);
+
+  if ((flags_ & Flag::ManuallyBuildMipmap) && (flags_ & Flag::ColorTexture)) {
+    int size = imageSize_ / pow(2, mipLevel);
```
mosra, line 325:
```suggestion
    int size = imageSize_ / (1 << mipLevel);
```

(I used to have `Math::pow2(n)` for this but ... it felt weird to have that wrapping just a bit shift, so it's no longer there.)

src/esp/gfx/CubeMap.cpp
```diff
@@ -201,64 +265,84 @@ void CubeMap::recreateTexture() {
...
     mapForDraw(cubeSideIndex);
   }
 
   if (flags_ & CubeMap::Flag::ColorTexture &&
       renderCameraFlags & RenderCamera::Flag::ClearColor) {
     frameBuffer_[cubeSideIndex].clearColor(
-        0,                      // color attachment
-        Mn::Color4{0, 0, 0, 1}  // clear color
+        0,                              // color attachment
+        Mn::Color4{0.0, 0.0, 0.0, 1.0}  // clear color
```
mosra, line 345: If you go the extra way, why not extra extra :D

```suggestion
        Mn::Color4{0.0f, 0.0f, 0.0f, 1.0f}  // clear color
```

(I wonder, did the compiler warn here? My assumption is that it shouldn't have for the integer literals, but it should for the double literals?)

src/esp/gfx/CubeMap.cpp
```diff
@@ -430,20 +533,83 @@ void CubeMap::loadTexture(TextureType type,
...
+
+  if (mipLevel > 0) {
+    mipLevelSanityCheck("CubeMap::CopyToTexture2D():", flags_, mipLevel,
+                        mipmapLevels_);
+  }
+
+  int size = imageSize_ / pow(2, mipLevel);
+
+#ifndef MAGNUM_TARGET_WEBGL
+  CORRADE_ASSERT(texture.imageSize(0) == Mn::Vector2i(size, size),
```
mosra, line 580: FYI, the single-argument constructor does exactly this:

```suggestion
  CORRADE_ASSERT(texture.imageSize(0) == Mn::Vector2i{size},
```

src/esp/scene/SceneGraph.h
```diff
@@ -27,12 +27,12 @@ class SceneGraph {
   SceneNode& getRootNode() { return rootNode_; }
   const SceneNode& getRootNode() const { return rootNode_; }
 
-  // TODO: remove this
-  gfx::DrawableGroup& getDrawables() {
-    return drawableGroups_.at(std::string{});
+  gfx::DrawableGroup& getDrawables(const std::string& groupName = "") {
```
mosra, line 30:
```suggestion
  gfx::DrawableGroup& getDrawables(const std::string& groupName = {}) {
```

`std::string` initialization from an empty C string literal involves a lot of nasty code including a `strlen()`, calling the default constructor not.

Same below.

mosra (commented):
Sorry If I'm going too overboard with those microoptimizations, spending a lot of time in low-level code lately and it shows :sweat_smile:

src/esp/gfx/CubeMap.cpp
bigbike: Ha, thanks! Good suggestions.

src/esp/scene/SceneGraph.h
bigbike: Yeah, good to know.
I was thinking the empty string was OK, since `strlen()` returned 0.

mosra (approved):

PR #1414, opened 2021-07-29
[Perf] AOWrapper pass settings by const ref instead of by value

Skylion007:
## Motivation and Context
* Static analysis suggested we were passing this by value when there was no reason to do so add the object was 80 bytes large. This should remove a copy of those 80 bytes with no bad effects. Surprised clang-tidy didn't pick up on this. Found by checking the lgtm.com static analysis.

## How Has This Been Tested
* Locally

## Types of changes

## Checklist


jturner65 (approved):
LGTM

PR #1423, opened 2021-08-04
ManagedContainers - Change underlying data structure to unordered map

jturner65:
## Motivation and Context
This PR changes the underlying data structure for the Managed Containers from std::map to std::unordered_map, for better performance with containers holding many elements.  

The only downside to this process is that formerly, any string-based query of the managed container would return results in sorted order. This PR will sort the results from such a query as default behavior, which can also be overridden on demand if sorting is not required. Since generally the results set is small (where the query is matching a substring of the handles of objects held in the map), the overhead for this sort will also generally be small.  

## How Has This Been Tested
C++ and python tests

## Types of changes

## Checklist

### DIFF
--- a/examples/tutorials/colabs/ECCV_2020_Advanced_Features.ipynb
+++ b/examples/tutorials/colabs/ECCV_2020_Advanced_Features.ipynb
@@ -839,7 +839,7 @@
     "    if not HAS_WIDGETS:\n",
     "        sel_file_obj_handle = file_obj_handles[0]\n",
     "        sel_prim_obj_handle = prim_obj_handles[0]\n",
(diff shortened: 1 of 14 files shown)


Skylion007 (approved):
Thanks, was considering making this change for a while as well, but never got around to doing it.

jturner65:
> Thanks, was considering making this change for a while as well, but never got around to doing it.

Had to consider it a bit, since the unordered map is slightly larger for small containers (less than 14 elements from what I have seen), but we don't have a lot of small containers so this isn't really a concern.  And apparently, running some tests on my machine, accessing the unordered map will be around 2-3x faster.

Issue #1401, opened 2021-07-22
AgentState.linear_velocity and AgentState.angular_velocity return 0-vectors


ericchen321:
## Habitat-Sim version
v0.2.0

## Docs and Tutorials
Did you read the docs? Yes

Did you check out the tutorials? Yes

### What I was trying to do
Hello, while playing around with the rigid object tutorial, I was trying to access the embodied agent's linear and angular velocity after setting them to some non-zero values. Specifically I have the following code:
```
# [embodied_agent]
# load the lobot_merged asset
locobot_template_id = obj_templates_mgr.load_configs(
    str(os.path.join(data_path, "objects/locobot_merged"))
)[0]
...
vel_control = sim.get_object_velocity_control(id_1)
vel_control.linear_velocity = np.array([0, 0, -1.0])
vel_control.angular_velocity = np.array([0.0, 2.0, 0])
...
vel_control.controlling_lin_vel = True
vel_control.controlling_ang_vel = True
vel_control.lin_vel_is_local = True
vel_control.ang_vel_is_local = True

# simulate forward and turn
observations += simulate(sim, dt=1.0, get_frames=make_video)
print(f"linear velocity: {sim.agents[0].get_state().velocity}")
print(f"angular velocity: {sim.agents[0].get_state().angular_velocity}")
```
Some of the API calls in the snippet are from Sim V1 (`Simulator.get_object_velocity_control()` for example), but with the V2 API calls I had the same problem.
### Expected behavior
I thought since  `simulate()` does not modify the agent's velocity, the print statements should show the velocities as the exact same values as the ones I set. 
### Actual behavior
I got the following values printed out:
```
linear velocity: [0. 0. 0.]
angular velocity: [0. 0. 0.]
```
So I am wondering why the velocities are both zero. Thanks in advance for the help.

Skylion007:
This is a good catch, those two properties are from an earlier API and are currently unused. We need to either add a pass through to the underlying physics velocity or remove those and calculate the velocity from the attached object to the agent. https://github.com/facebookresearch/habitat-sim/blob/55cf55827e21aa6816332ab3cae5b644b348f0a0/habitat_sim/agent/agent.py#L84 We are also working on merging the duplicated Python and C++ classes to reduce this code duplication. In other words, this info is not available in the agent_state class, but it should be.  

TLDR: Don't use that velocity property on agent_state, we are not updating that value, but we should be.
@aclegg3 @eundersander flagging this bug.

[...]

PR #1426, opened 2021-08-05
Remove unused and potentially confusing agent dynamic properties

aclegg3:
## Motivation and Context

As pointed out by user issues (e.g. #1401 ) Habitat Agents have unused dynamic properties: velocity, mass, acceleration, friction, restitution. This PR removes those properties to reduce confusion. When an Agent is attached to a dynamic object (e.g. robots and rigid objects) dynamic properties should be queried from the physics object interface directly ([ManagedBulletRigidObject](https://aihabitat.org/docs/habitat-sim/habitat_sim.physics.ManagedBulletRigidObject.html) or [ManagedBulletArticulatedObject](https://aihabitat.org/docs/habitat-sim/habitat_sim.physics.ManagedBulletArticulatedObject.html)).

## How Has This Been Tested

I couldn't find any references to these in our code, but if anyone knows of any existing uses which would be negatively impacted by this cleanup we can potentially think of an alternative to removing them.

## Types of changes

## Checklist


Skylion007 (approved):
It would be nice to pass through some this info through Bullet if it supports it though.

aclegg3:
> It would be nice to pass through some this info through Bullet if it supports it though.

Sure, but a base Habitat Agent does not have a `Bullet` embodiment and won't know it has one attached unless we add more sophisticated extensions of Agent to track this. I propose we add these back if/when we make that change rather than keeping them as part of the base agent.

Skylion007:
Looks like the Habitat-Lab configs currently do specify these properties so we should link it up to the physics manager if appropriate.

mathfac (approved):
Yes, as @Skylion007 said we need to send PR to Habitat Lab first to remove field from the [HabLab default config and others](https://github.com/facebookresearch/habitat-lab/blob/368007d067119518124e9d632f4769ee1d5898e4/habitat/config/default.py#L353-L357).

PR #1430, opened 2021-08-06
Build OSX Conda Nightly

Skylion007:
## Motivation and Context
* Build Nightly OSX Binaries
@mathfac Putting this here so I can get some help testing and deploying the CI command.

## Types of changes

## Checklist


.circleci/config.yml
```diff
@@ -6,6 +6,43 @@ gpu: &gpu
   environment:
     FPS_THRESHOLD: 900
 commands:
+  build_osx_conda_pkg:
+    description: "Build OSX Conda-Package"
+    parameters:
```
mathfac, line 11: It's better to have same parameter signature like [the linux building job](https://github.com/facebookresearch/habitat-sim/blob/f216a1732a2d33747c56df9a4d07485ba70cda3a/.circleci/config.yml#L617-L626):

```
    parameters:
      AIHABITAT_CONDA_CHN:
        type: string
        default: "aihabitat"
      AIHABITAT_CONDA_CHN_PWD_VAR:
        type: string
        default: "AIHABITAT_CONDA_PWD"
      NIGHTLY_FLAG:
        type: string
        default: ""
```
 Maybe, add one more parameter `upload` for cases when no upload required.

.circleci/config.yml
```diff
@@ -6,6 +6,43 @@ gpu: &gpu
...
+      - run:
+          name: Conda OSX build
+          command: |
+            export PATH=$HOME/miniconda/bin:$PATH
+            source ~/.bash_profile
+            conda activate
+            conda install -y anaconda-client ccache cmake git ninja conda-build pip
+            echo $(which -a python)
+            pip install gitpython
+            <<^ parameters.test>> anaconda login --username $AIHABITAT_CONDA_CHN --password $AIHABITAT_CONDA_CHN_PWD
```
mathfac, line 39: Here channel is known and nonsensitive information, while password is taken from environment variable and parameter is used only for the variable name:

```suggestion
            anaconda login --username << parameters.AIHABITAT_CONDA_CHN >>  --password $<< parameters.AIHABITAT_CONDA_CHN_PWD_VAR >>
```

mathfac (changes requested):
Ideally the one job should serve 3 use-cases:
- CI test build with limited python versions support without upload that will run on every commit
- CI nightly that will build nightly package from master and upload it
- Version release that will build the packages and upload it to main Habitat conda channel.

mathfac:
The error you have right now is not related to the anaconda, but pre installed dependencies:
`cp: cannot stat '/usr/include/EGL': No such file or directory`

erikwijmans:
> @erikwijmans The conda nightly builds are currently being built, after they are done I will add them as a dependency on build_conda_binaries so they are run if the Linux conda build succeeds.

I guess the dependency is since the linux one is responsible for removing old binaries? If so, we can factor that out in the future, but no need to do that now.

Skylion007:
Exactly

> 
> 
> > @erikwijmans The conda nightly builds are currently being built, after they are done I will add them as a dependency on build_conda_binaries so they are run if the Linux conda build succeeds.
> 
> I guess the dependency is since the linux one is responsible for removing old binaries? If so, we can factor that out in the future, but no need to do that now.

Exactly

erikwijmans (approved):
Ship it!

PR #1431, opened 2021-08-10
[Logging] Expand logging output information

jturner65:
## Motivation and Context
The new logging functionality printed log messages that were somewhat arcane, lacking source information outside of what subsystem/namespace the log was being printed from : 

`[Metadata] <Dataset>::createFromJsonOrDefaultInternal : Proposing JSON name : default.scene_dataset_config.json from original name : default | This file  does not exist.`

This PR improves this by adding the source file (without path), the line number and the source function within which the logging macro call was executed  :

`[Metadata] AttributesManagerBase.h(360)::createFromJsonOrDefaultInternal : <Dataset>: Proposing JSON name : default.scene_dataset_config.json from original name : default | This file  does not exist.` 

The majority of the other file changes are just removing the now redundant references to calling function names that were ubiquitous in the logging calls, particularly in the metadata subsystem, having been added to provide this functionality that was lacking in glog also.

## How Has This Been Tested
Existing C++ and python tests pass locally.  Logging test was modified to match new output format.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/logging.cpp
+++ b/src/esp/core/logging.cpp
@@ -11,6 +11,7 @@
 #include <Corrade/Containers/Array.h>
 #include <Corrade/Containers/StaticArray.h>
 #include <Corrade/Containers/String.h>
+#include <Corrade/Utility/Directory.h>
 #include <Corrade/Utility/Format.h>
 
 namespace Cr = Corrade;
@@ -121,8 +122,14 @@ bool isLevelEnabled(Subsystem subsystem, LoggingLevel level) {
   return level >= LoggingContext::current().levelFor(subsystem);
 }
 
-Cr::Containers::String subsystemPrefix(Subsystem subsystem) {
-  return ""_s.join({"["_s, subsystemNames[uint8_t(subsystem)], "] "_s});
+Cr::Containers::String buildMessagePrefix(Subsystem subsystem,
+                                          const std::string& filename,
+                                          const std::string& function,
+                                          int line) {
+  auto baseFileName = Cr::Utility::Directory::filename(filename);
+  return ""_s.join({"["_s, subsystemNames[uint8_t(subsystem)], "] "_s,
+                    baseFileName.c_str(), "("_s, std::to_string(line).c_str(),
+                    ")::"_s, function.c_str(), " : "_s});
 }
 
 }  // namespace logging
--- a/src/esp/metadata/attributes/SceneDatasetAttributes.cpp
+++ b/src/esp/metadata/attributes/SceneDatasetAttributes.cpp
@@ -25,8 +25,7 @@ SceneDatasetAttributes::SceneDatasetAttributes(
 bool SceneDatasetAttributes::addNewSceneInstanceToDataset(
     const attributes::SceneAttributes::ptr& sceneInstance) {
   // info display message prefix
-  const std::string infoPrefix("::addNewSceneInstanceToDataset : Dataset : '" +
-                               getSimplifiedHandle() + "' : ");
+  const std::string infoPrefix("Dataset : '" + getSimplifiedHandle() + "' : ");
 
   const std::string sceneInstanceName = sceneInstance->getHandle();
   // verify stage in sceneInstance (required) exists in SceneDatasetAttributes,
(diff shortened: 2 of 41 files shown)


src/esp/core/logging.cpp
```diff
@@ -121,8 +122,14 @@ bool isLevelEnabled(Subsystem subsystem, LoggingLevel level) {
...
 
-Cr::Containers::String subsystemPrefix(Subsystem subsystem) {
-  return ""_s.join({"["_s, subsystemNames[uint8_t(subsystem)], "] "_s});
+Cr::Containers::String buildMessagePrefix(Subsystem subsystem,
+                                          const std::string& filename,
+                                          const std::string& function,
+                                          int line) {
+  auto baseFileName = Cr::Utility::Directory::filename(filename);
+  return ""_s.join({"["_s, subsystemNames[uint8_t(subsystem)], "] "_s,
+                    baseFileName.c_str(), "("_s, std::to_string(line).c_str(),
```
erikwijmans, line 131: If you include `#include <Corrade/Containers/StringStl.h>` you can get rid of the `c_str()` calls which will avoid a `std::strlen` call.

src/esp/metadata/attributes/SceneDatasetAttributes.cpp
```diff
@@ -25,8 +25,7 @@ SceneDatasetAttributes::SceneDatasetAttributes(
 bool SceneDatasetAttributes::addNewSceneInstanceToDataset(
     const attributes::SceneAttributes::ptr& sceneInstance) {
   // info display message prefix
-  const std::string infoPrefix("::addNewSceneInstanceToDataset : Dataset : '" +
-                               getSimplifiedHandle() + "' : ");
+  const std::string infoPrefix("Dataset : '" + getSimplifiedHandle() + "' : ");
```
Skylion007, line 28: Nit: This is going to require 3 reallocations of the prefix in the constructor of the String. Should all be std::string

erikwijmans (approved):
This seems reasonable to me.  Adding this extra detail should be worth it.

aclegg3 (approved):
LGTM

src/esp/core/logging.cpp
```diff
@@ -121,8 +123,14 @@ bool isLevelEnabled(Subsystem subsystem, LoggingLevel level) {
...
-Cr::Containers::String subsystemPrefix(Subsystem subsystem) {
-  return ""_s.join({"["_s, subsystemNames[uint8_t(subsystem)], "] "_s});
+Cr::Containers::String buildMessagePrefix(Subsystem subsystem,
+                                          const std::string& filename,
+                                          const std::string& function,
+                                          int line) {
+  auto baseFileName = Cr::Utility::Directory::filename(filename);
+  return ""_s.join({"["_s, subsystemNames[uint8_t(subsystem)], "] "_s,
+                    baseFileName, "("_s, std::to_string(line), ")::"_s,
+                    function, " : "_s});
```
mosra, line 133: @jturner65 I was just browsing through some code and came across this -- the following should be significantly more performant, without temporary string allocations for `filename`, `function` and `line` (it returns a `std::string` now instead of a `String`, yes, making that consistent is on my tech debt task pile):

```cpp
std::string buildMessagePrefix(Subsystem subsystem,
                                          const char* filename,
                                          const char* function,
                                          int line) {
  auto baseFileName = Cr::Utility::Directory::filename(filename);
  return Utility::formatString("[{}] {}({})::{} : ", subsystemNames[uint8_t(subsystem)],
                    baseFileName, line, function);
}
```

FYI there's also [CORRADE_FUNCTION](https://doc.magnum.graphics/corrade/Macros_8h.html#a35bb07eaded9ca76d8a2c72135243ded) and [CORRADE_LINE_STRING](https://doc.magnum.graphics/corrade/Macros_8h.html#a074226a9c37ee5816e828a4cf823381d), in case you'd ever need a portable function name or stringified line number.

**Also**, certain terminals and IDEs (KDE's Konsole, VS Code, ...) allow you to click on the filename if you include a reasonable part of the path, and even can jump directly to given line number if you format it as `<filename>:<line>` (which is the format used by GCC messages, GDB, CMake and other tools). So if you want an extra bit of convenience you might want to keep the path and reformat the output slightly. For example this I have in TestSuite, the failure location leads me directly to the line in the editor if I click on it:

image

PR #1433, opened 2021-08-11
esp::core::Configuration changed to use unordered_map of tagged union instead of Magnum::ConfigurationGroups

jturner65:
## Motivation and Context
Currently Habitat-sim uses Magnum::ConfigurationGroups as the backing datastructure for all metadata loaded via JSON configuration files.  A ConfigurationGroup saves all data as strings, and the original type of that data is lost once it is loaded into the system.  This is particularly problematic with the user-defined attributes that habitat now supports. Upon read, the type of the user-defined value is inferred by the format of the data in the source JSON.  With the upcoming functionality to save JSON configs to disk, including user_defined values, without having the type known will cause all data to be saved to disk as strings.

This PR replaces the ConfigurationGroup as the container for esp::core::Configuration and instead uses a  ( _collection of typed std::unordered_maps._  ) single unordered map of ConfigValues, which use a tagged union/variant structure to retain type information and safety.

One question to consider : do we wish to remove the ConfigurationGroup as the backing structure for esp::core::Configuration.  A different structure can be easily designed to replace esp::core::Configuration as the base class for the various configuration attributes that Habitat uses.

If nobody is using an esp::core::Configuration currently outside of the Metadata attributes/configuration subsystem, then the changes proposed by this PR can probably proceed with impunity; if, however, the ConfigurationGroup functionality backing habitat's Configuration is desired to remain, then I should instead create another class to back the attributes.

## How Has This Been Tested
Locally, all c++ and python tests pass.

## Types of changes

## Checklist


Skylion007:
Really would like to hear @mosra's input on this. Might be a better solution.

jturner65:
Discussing this refactor with @mosra, he suggested I investigate a tagged union storage container, so this PR introduces this.

src/esp/metadata/attributes/PhysicsManagerAttributes.cpp
```diff
@@ -14,6 +14,7 @@ PhysicsManagerAttributes::PhysicsManagerAttributes(const std::string& handle)
   setTimestep(0.008);
   setGravity({0, -9.8, 0});
   setFrictionCoefficient(0.4);
+  setRestitutionCoefficient(0.1);
```
eundersander, line 17: Separate PR?

src/esp/core/Configuration.h
```diff
@@ -9,184 +9,619 @@
...
+      valueMap_.erase(key);
     }
-    return 0;
   }
 
-  /**@brief Add a string to a group and return the resulting group size. */
-  int addStringToGroup(const std::string& key, const std::string& value) {
-    cfg.addValue(key, value);
-    return cfg.valueCount(key);
+  bool removeBool(const std::string& key) {
```
eundersander, line 330: Why do we need these type-specific removes?

eundersander (commented):
Overall seems good. It is quite verbose, e.g. lots of switch statements... hopefully we aren't adding new types often.

I spotted a number of changes that seemed to be unrelated to the config stuff. Let's move them to separate PRs so they get properly reviewed.

Side note: we should move to C++17 soon! I guess a lot of this is re-inventing std::any/std::variant.

src/esp/core/Configuration.cpp
```diff
@@ -0,0 +1,280 @@
...
+    v.~Vector3();
+  } else if (type == ConfigStoredType::MN_QUAT) {
+    q.~Quaternion();
+  } else if (type == ConfigStoredType::MN_RAD) {
+    r.~Rad();
+  }
+  return false;
+}
+
+ConfigValue& ConfigValue::operator=(const ConfigValue& otr) {
```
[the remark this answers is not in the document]
jturner65: I know. I'll investigate an improvement

aclegg3 (commented):
I'd also like Mosra's opinion before merge. Overall, I like this approach.

I chatted with @jturner65 and came up with a couple of other nice-to-have features:
1. a query for all keys as a map/dict (key -> type enum) and supporting type enum bindings.
2. a get(key, type) python function for easy access out of the dict returned by (1.)
3. sub configuration accessor bindings: get a sub configuration as a separate Configuration object (nested hierarchal structure) with option to get copy or ref and to register modified copies back into a subconfiguration
4. query list of sub configuration keys

src/esp/core/Configuration.h
jturner65: wanted to provide a remove w/return object removed.

src/esp/metadata/attributes/PhysicsManagerAttributes.cpp
jturner65: Need to have initial values for all supported values for this PR, since there's no default value available/provided.

src/esp/core/Configuration.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+void ConfigValue::deleteCurrentValue(const std::string& src) {
+  switch (type) {
+    case ConfigStoredType::Unknown:
+    case ConfigStoredType::Boolean:
+    case ConfigStoredType::Integer:
+    case ConfigStoredType::Double:
+      // trivially destructible
+      break;
+    case ConfigStoredType::String:
```
Skylion007, line 62: This really could be handled better with some clever templating and std::is_trivially_destructible, no?

src/esp/core/Configuration.cpp
```diff
@@ -0,0 +1,246 @@
...
+      return "Undefined value/Unknown type";
+    case ConfigStoredType::Boolean:
+      return (b ? "True" : "False");
+    case ConfigStoredType::Integer:
+      return std::to_string(i);
+    case ConfigStoredType::Double:
+      return std::to_string(d);
+    case ConfigStoredType::String:
+      return s;
+    case ConfigStoredType::MagnumVec3: {
```
Skylion007, line 182: Doesn't Magnum have some sort of serializer for these types @mosra?

src/esp/core/Configuration.cpp
jturner65: Yeah, I have been iteratively refactoring to simplify, and doing that is on my list, actually :)

jturner65: Actually, I'm open to suggestions here.  How would i get the correct value to destroy (and its type) based on the type enum, which is all I know about the value when this is called?

src/esp/bindings/CoreBindings.cpp
```diff
@@ -14,25 +14,177 @@ namespace esp {
...
+            return self.checkMapForKeyAndType(key,
+                                              ConfigStoredType::MagnumVec3);
+          },
+          R"(Returns true if specified key references a Magnum::Vector3 value in this configuration.)")
+      .def(
+          "has_rad",
+          [](Configuration& self, const std::string& key) {
+            return self.checkMapForKeyAndType(key, ConfigStoredType::MagnumRad);
+          },
+          R"(Returns true if specified key references a Magnum::Rad value in this configuration.)")
```
mosra, line 159: Not sure about usefulness of the `has_*` APIs, now that you have everything stored in a single map there isn't a possibility of the same key being used for two different types and so this could be expressed simply as this and there's no need to have a dedicated overload for each and every type.

```py
foo.get_type('bar') == mn.Rad
```

(assuming the above suggestion with `get_type()` returning a Python type object, which makes it *a lot* shorter than `ConfigStoredType.MagnumRad`)

src/esp/bindings/CoreBindings.cpp
```diff
@@ -14,25 +14,177 @@ namespace esp {
...
+      .def(
+          "has_subconfig", &Configuration::hasSubconfig,
+          R"(Returns true if specified key references an existing subconfiguration within this configuration.)")
```
mosra, line 162: In this particular case, it'd be for example

```py
foo.get_type('bar') == Configuration

# or
type(foo.bar) == Configuration
```

src/esp/bindings/CoreBindings.cpp
```diff
@@ -14,25 +14,177 @@ namespace esp {
...
+          R"(Removes and returns Magnum::Vector3 value corresponding to passed key, if found. Gives warning otherwise.)")
+      .def(
+          "remove_quat", &Configuration::removeQuat,
+          R"(Removes and returns Magnum::Quaternion value corresponding to passed key, if found. Gives warning otherwise.)")
+      .def(
+          "remove_rad", &Configuration::removeRad,
+          R"(Removes and returns Magnum::Rad value corresponding to passed key, if found. Gives warning otherwise.)")
+      .def(
+          "remove_subconfig", &Configuration::removeSubconfig,
+          R"(Removes and returns subconfiguration corresponding to passed key, if found. Gives warning otherwise.)");
```
mosra, line 187: I think these overloads aren't needed either anymore now that you have a single map for all -- just `foo.remove('bar')`. Or, in a more Pythonic way again, implement `__delitem__` / `__delattr__`:

```py
del foo['bar']
# or
del foo.bar
```

src/esp/metadata/attributes/LightLayoutAttributes.h
```diff
@@ -143,23 +143,20 @@ class LightInstanceAttributes : public AbstractAttributes {
...
-        .append(cfg.value("color"))
+        .append(getAsString("color"))
         .append(1, ',')
-        .append(cfg.value("intensity"))
+        .append(std::to_string(getIntensity()))
         .append(1, ',')
         .append(getCurrLightTypeName())
         .append(1, ',')
         .append(getCurrLightPositionModelName())
         .append(1, ',');
```
mosra, line 157: Ow, the allocations... Use [`Utility::formatString()`](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#a5b85466788fc3d71b143909dda1ef44a), please :)

```cpp
return Cr::Utility::formatString("{},{},{},{},{},{}", getAsString("position"), getAs...);
```

For even less temporary string allocations format the Vector3 components directly, instead of taking them as a string separately.

src/esp/metadata/attributes/ObjectAttributes.cpp
```diff
@@ -54,19 +54,19 @@ std::string AbstractObjectAttributes::getObjectInfoInternal() const {
...
+      .append(std::to_string(getUnitsToMeters()))
       .append(1, ',')
-      .append(cfg.value("friction_coefficient"))
+      .append(std::to_string(getFrictionCoefficient()))
       .append(1, ',')
-      .append(cfg.value("restitution_coefficient"))
+      .append(std::to_string(getRestitutionCoefficient()))
       .append(1, ',')
       .append(getCurrShaderTypeName())
       .append(1, ',')
```
mosra, line 72: Same here, `formatString()` or [`Utility::String::join()`](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1String.html#a9af3018407a9cac247c09da9ac8e577d) at least :P

src/esp/metadata/attributes/ObjectAttributes.cpp
```diff
@@ -96,17 +96,17 @@ ObjectAttributes::ObjectAttributes(const std::string& handle)
...
-      .append(cfg.value("angular_damping"))
+      .append(std::to_string(getAngularDamping()))
       .append(1, ',')
-      .append(cfg.value("linear_damping"))
+      .append(std::to_string(getLinearDamping()))
       .append(1, ',')
-      .append(cfg.value("semantic_id"));
+      .append(std::to_string(getSemanticId()));
 }
```
mosra, line 110: Also :)

src/esp/metadata/attributes/ObjectAttributes.h
```diff
@@ -406,11 +402,11 @@ class StageAttributes : public AbstractObjectAttributes {
...
-        .append(cfg.value("gravity"))
+        .append(getAsString("gravity"))
         .append(1, ',')
-        .append(cfg.value("origin"))
+        .append(getAsString("origin"))
         .append(1, ',')
-        .append(cfg.value("light_setup"));
+        .append(getLightSetup());
     return res;
   }
```
mosra, line 411: Here too.

src/esp/metadata/attributes/PhysicsManagerAttributes.h
```diff
@@ -64,15 +62,15 @@ class PhysicsManagerAttributes : public AbstractAttributes {
...
         .append(1, ',')
-        .append(cfg.value("gravity"))
+        .append(getAsString("gravity"))
         .append(1, ',')
-        .append(cfg.value("friction_coefficient"))
+        .append(std::to_string(getFrictionCoefficient()))
         .append(1, ',')
-        .append(cfg.value("restitution_coefficient"));
+        .append(std::to_string(getRestitutionCoefficient()));
   }
```
mosra, line 74: Here as well :P

src/esp/metadata/attributes/SceneAttributes.cpp
```diff
@@ -102,8 +103,8 @@ std::string SceneAOInstanceAttributes::getSceneObjInstanceInfoInternal() const {
     initJointVels.append(std::to_string(it.second)).append(1, ',');
   }
   initJointVels.append("]");
-  return cfg.value("fixed_base").append(1, ',') + initJointPose.append(1, ',') +
-         initJointVels.append(1, ',');
+  return getAsString("fixed_base").append(1, ',') +
+         initJointPose.append(1, ',') + initJointVels.append(1, ',');
 }  // SceneAOInstanceAttributes::getSceneObjInstanceInfoInternal()
```
mosra, line 108: Also

src/esp/physics/PhysicsObjectBase.h
```diff
@@ -427,13 +427,17 @@ class PhysicsObjectBase : public Magnum::SceneGraph::AbstractFeature3D {
...
-   * @brief This function will completely overwrite this object's
-   * user-defined attributes.
+   * @brief This function will overwrite or merge this object's existing
+   * user-defined attributes with @p attr , based on value of @p merge
    * @param attr A ptr to the user defined attributes specified for this object.
    * merge into them.
    */
-  void setUserAttributes(core::Configuration::ptr attr) {
-    userAttributes_ = std::move(attr);
+  void setOrMergeUserAttributes(core::Configuration::ptr attr, bool merge) {
```
mosra, line 435: [Not a fan of booleans in APIs](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap), why can't it be two different functions? Shorter to type, easier to understand, and the two functions don't really have any common code to begin with so putting them together doesn't make anything simpler.

src/tests/AttributesManagersTest.cpp
```diff
@@ -355,15 +354,15 @@ class AttributesManagersTest : public testing::Test {
...
+      double double_val,
       Magnum::Vector3 vec_val,
       Magnum::Quaternion quat_val) {
     // user defined attributes from light instance
     ASSERT_NE(nullptr, userConfig);
     ASSERT_EQ(userConfig->getString("user_string"), str_val);
     ASSERT_EQ(userConfig->getBool("user_bool"), bool_val);
     ASSERT_EQ(userConfig->getInt("user_int"), int_val);
-    ASSERT_EQ(userConfig->getFloat("user_float"), float_val);
+    ASSERT_EQ(userConfig->getDouble("user_double"), double_val);
```
mosra, line 365: Why doesn't the new API have 32-bit floats anymore? Those are used for physics, for the transformations, for most of everything.

mosra (commented):
Went through the python bindings and the "other code" first, will focus on the internals in a separate review.

src/esp/bindings/AttributesBindings.cpp
```diff
@@ -66,28 +66,71 @@ void initAttributesBindings(py::module& m) {
           R"(System-generated ID for template.  Will be unique among templates
           of same type.)")
       .def("get_user_config_bool",
-           &AbstractAttributes::getUserConfigValue<bool>)
+           [](AbstractAttributes& self, const std::string& key) {
+             const auto ptr = self.editUserConfiguration();
+             return ptr->getBool(key);
+           })
```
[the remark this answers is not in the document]
jturner65: Ah, I tried to do that, but I didn't know about the py::cast.  Ty!

src/tests/AttributesManagersTest.cpp
jturner65: Now, there's no need to distinguish between floats and doubles (Same map and all that) so I will reintroduce them.  

But at the time when I made the above change, I was using separate maps for floats and doubles, and since there's no way to distinguish between doubles and floats in python, I found this to be a potential problem. Granted, there's no way to distinguish them in json either (rapidjson just does a cast from double to float if you want something as a float) so i thought there was minimal loss.

src/esp/metadata/attributes/LightLayoutAttributes.h
jturner65: Thanks! forgot about that!

src/esp/bindings/CoreBindings.cpp
```diff
@@ -14,25 +14,177 @@ namespace esp {
...
 void initCoreBindings(py::module& m) {
+  py::enum_<ConfigStoredType>(m, "ConfigStoredType")
+      .value("Unknown", ConfigStoredType::Unknown)
+      .value("Boolean", ConfigStoredType::Boolean)
+      .value("Integer", ConfigStoredType::Integer)
+      .value("Double", ConfigStoredType::Double)
+      .value("String", ConfigStoredType::String)
+      .value("MagnumVec3", ConfigStoredType::MagnumVec3)
+      .value("MagnumQuat", ConfigStoredType::MagnumQuat)
+      .value("MagnumRad", ConfigStoredType::MagnumRad);
```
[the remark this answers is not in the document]
Skylion007: `py::type::of` is what you are looking for: https://github.com/pybind/pybind11/pull/2364

src/esp/bindings/CoreBindings.cpp
```diff
@@ -14,25 +14,177 @@ namespace esp {
...
+                           const double val) { self.set(key, val); })
+      .def("set_bool", [](Configuration& self, const std::string& key,
+                          const bool val) { self.set(key, val); })
+      .def("set_quat",
+           [](Configuration& self, const std::string& key,
+              const Magnum::Quaternion& val) { self.set(key, val); })
+      .def("set_vec3", [](Configuration& self, const std::string& key,
+                          const Magnum::Vector3& val) { self.set(key, val); })
+      .def("set_rad", [](Configuration& self, const std::string& key,
+                         const Magnum::Rad& val) { self.set(key, val); })
```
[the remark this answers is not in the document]
jturner65: Back-compat with the original "set" name - was trying to be as compatible as possible, since I wasn't sure how much the original "set" verbiage was expected.  
My preferences were, when I pushed the most recent commit, to have the type explicitly named in all the accessors, so there as no confusion or mistakes.  While this is exceedingly verbose, this was a first pass at being as safe as possible.  Since this is only pertinent for configurations that the user is customizing and creating on the fly, either in python or in JSON files (not the built ins, which will always be accessed directly as properties), I wanted to provide as many guardrails as possible.

src/esp/bindings/CoreBindings.cpp
jturner65: I did have them separate so that I could "remove and return" but with your suggestion of py::cast above, I can now do this in a single method.

src/esp/physics/PhysicsObjectBase.h
jturner65: I had this because I wanted to support a process where I was not sure which call would be first.  That is probably not a reasonable concern, really - there's a call right after the object is created, and then another when it is placed.  There's not going to be confusion about this.

src/esp/bindings/CoreBindings.cpp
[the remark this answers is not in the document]
jturner65: I do need to have an equivalent to 'ConfigStoredType::unknown' to handle the edge case where the value has not been properly initialized.  Is there a python type equivalent i could use?

src/tests/AttributesManagersTest.cpp
jturner65: Due to both the python and the json not distinguishing between floats and doubles, I am going to keep the Configuration to only support doubles for now, unless someone has some other suggestions.

src/esp/metadata/attributes/LightLayoutAttributes.h
jturner65: I'm going to strip out all of the offending functionality from this PR and refactor all of these to use the formatString process in another PR.

src/esp/metadata/attributes/SceneAttributes.cpp
jturner65: I'm going to strip out all of the offending functionality from this PR and refactor all of these to use the formatString process in another PR.

src/esp/metadata/attributes/PhysicsManagerAttributes.h
jturner65: I'm going to strip out all of the offending functionality from this PR and refactor all of these to use the formatString process in another PR.

src/esp/metadata/attributes/ObjectAttributes.h
jturner65: I'm going to strip out all of the offending functionality from this PR and refactor all of these to use the formatString process in another PR.

src/esp/metadata/attributes/ObjectAttributes.cpp
jturner65: I'm going to strip out all of the offending functionality from this PR and refactor all of these to use the formatString process in another PR.

src/esp/metadata/attributes/ObjectAttributes.cpp
jturner65: I'm going to strip out all of the offending functionality from this PR and refactor all of these to use the formatString process in another PR.

jturner65 (commented):
I'm going to strip out all the string concatenation changes and put them in another PR.  (The only modifications will be what are required to support the underlying changes in the Configuration).

src/esp/core/Configuration.cpp
Skylion007: https://stackoverflow.com/questions/17328454/calling-destructor-with-decltype-and-or-stdremove-reference You can generically call the destructor and trivial ones will get optimized away, no?

mosra (commented):
Another batch, about the internals, I still need to look deeper into the breadcrumb parts. 

Sorry in advance for all the comments! :)

src/esp/metadata/attributes/LightLayoutAttributes.h
```diff
@@ -207,6 +204,15 @@ class LightLayoutAttributes : public AbstractAttributes {
    * @brief Add a light instance to this lighting layout
    */
   void addLightInstance(const LightInstanceAttributes::ptr& _lightInstance) {
+    // set id
+    if (availableLightIDs_.size() > 0) {
```
Skylion007, line 208:
```suggestion
    if (!availableLightIDs_.empty()) {
```

src/esp/bindings/CoreBindings.cpp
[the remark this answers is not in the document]
mosra: `py::none` would be the thing, I'd say? Similar semantics to a null pointer.

src/esp/core/Configuration.cpp
mosra: It has, one for debug output (where the goal is human readability) and one for (INI-style, space-delimited) configuration values. Neither of those produces what's supposedly needed here, tho (JSON-like formatting).

Skylion007:
>>  A std::variant/std::visitor system would have been safer and more streamlined, but these require c++17, which we do not currently support. 
This would be nice to have and PyBind11 has native support for std::variant. Everything seems to compile just fine under C++17 so I would in favor of upgrading as well. https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#c-17-library-containers. There are also some minor improvements in PyBind11 you get when using C++17 thanks to fold expressions and better constexpr support.

jturner65:
> > A std::variant/std::visitor system would have been safer and more streamlined, but these require c++17, which we do not currently support.
 
> This would be nice to have and PyBind11 has native support for std::variant. Everything seems to compile just fine under C++17 so I would in favor of upgrading as well. https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#c-17-library-containers. There are also some minor improvements in PyBind11 you get when using C++17 thanks to fold expressions and better constexpr support.

Just FYI, when I said safer and more streamlined, I meant as compared to the former tagged-union refactor that this PR originally introduced.  The system that I replaced it with (with substantial guidance from @mosra ) does not have the same safety concerns.

mosra:
Re `std::variant`, to clarify my view and repeat to what I said to John privately -- I think what we created here does everything we need, is safe, memory- and runtime-efficient, compiles fast and is reasonably easy to maintain / extend for new types.

On the other hand, `std::variant<bool, int, double, Vector3, Quaternion, Rad, std::string, Configuration, ...>` would have [notorious issues with compile times](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85771), where the compile-time cost of accessing a value scales linearly with the number of types a variant can store, and it can quickly become unbearable to the point where people have to "optimize" by storing e.g. a `float` as a `Vector3` to reduce the amount of types used. (Not to mention the ["fun" one has to go through](https://bitbashing.io/std-visit.html) with `std::visit()`.) Yes, I'm skeptical about C++17 features.

image

About python bindings, I think we're fine here, exposing all types in a Pythonic way wasn't too complicated.

src/esp/core/Configuration.cpp
jturner65: I think the new design addresses the concerns you bring up here, but if not please comment :)

aclegg3 (approved):
Looks good. Once the last couple review changes are addressed I think we can ship this.

tests/test_configs.py
```diff
@@ -17,39 +17,36 @@ def test_config_eq():
 def test_core_configuration():
```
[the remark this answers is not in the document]
jturner65: Now that python access is accomplished solely by 'get' and 'set', and type is inferred/queried behind the scenes, there's no way from python to forcibly query the wrong type for a specific field by name.

src/esp/physics/PhysicsObjectBase.h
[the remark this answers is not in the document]
jturner65: I have separated the set and merge to two different functions, but retained them to support @aclegg3's efforts with clutter generation.

src/esp/core/Configuration.h
```diff
@@ -9,184 +9,649 @@
...
+   * configuration holds and the types of these values.
+   */
+  std::unordered_map<std::string, ConfigStoredType> getValueTypes() const {
+    std::unordered_map<std::string, ConfigStoredType> res{};
+    res.reserve(valueMap_.size());
+    for (const auto& elem : valueMap_) {
+      res[elem.first] = elem.second.getType();
+    }
+    return res;
+  }
```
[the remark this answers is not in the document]
jturner65: This function was added for informational purposes, primarily to provide a listing of all the available keys.  Since, with the user-defined attributes configuration, the user may not remember all the keys that are currently available in the Configuration.  Originally, this was a vector of strings, but the types were added to provide extra information to the user.

src/esp/metadata/attributes/AttributesBase.h
```diff
@@ -174,46 +175,35 @@ class AbstractAttributes : public esp::core::AbstractFileBasedManagedObject,
...
-   * be available to the user via python bindings for each object.
+   * @brief Gets a smart pointer reference to the actual user-specified
+   * configuration data from config file. Habitat does not parse or process this
+   * data, but it will be available to the user via python bindings for each
+   * object.  This method is for editing the configuration.
    */
-
-  std::shared_ptr<Configuration> getUserConfiguration() const {
-    return getConfigSubgroupAsPtr("user_defined");
+  std::shared_ptr<Configuration>& editUserConfiguration() {
```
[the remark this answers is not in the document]
jturner65: Most of the issues brought up here are obviated with the removal of the typed getter/setter bindings.

PR #1435, opened 2021-08-11
add definition for LoggingContext::DEFAULT_LEVEL 

eundersander:
## Motivation and Context

to fix link error in debug build

## How Has This Been Tested

built locally

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/logging.cpp
+++ b/src/esp/core/logging.cpp
@@ -19,6 +19,7 @@ using Cr::Containers::Literals::operator""_s;
 namespace esp {
 namespace logging {
 
+constexpr const char* LoggingContext::LOGGING_ENV_VAR_NAME;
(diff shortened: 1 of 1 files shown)


Skylion007:
Seems like we need to update the apt-get install command if the Habitat-Lab build keeps flaking like this. Odd since the normal habitat-sim build worked fine.

erikwijmans (approved):
Interesting.  I guess that on the debug build the constant doesn't get optimized out and thus requires a definition?  While you are here, can you add a definition for `LOGGING_ENV_VAR_NAME` in case that stops getting optimized out?

Skylion007:
@erikwijmans I wanted to change CircleCI to use the Debug Build, but I don't think all our tests currently pass in Debug build.

PR #1439, opened 2021-08-13
[BUGFIX] Do not set orientation based on file ext if default attributes provided, overrides default config for orientation.

jturner65:
## Motivation and Context
This PR fixes an issue when someone loads an asset with a particular extension that we have hardcoded internally to have a certain orientation (for backwards compatibility).  If the desired asset does not have that orientation, it is impossible to set the orientation via default attributes settings in scene dataset configuration file.  

This fix is a stopgap measure for now, and should work mostly, but the only foolproof solution to this is going to be getting rid of the hardcoded asset setting based on extension.  As we get more dataset assets, we cannot assume anything orientation-wise about assets solely on their extension.  To do this, we need to rework the tests to use config-backed assets.

## How Has This Been Tested
Locally c++ and python.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/managers/StageAttributesManager.cpp
+++ b/src/esp/metadata/managers/StageAttributesManager.cpp
@@ -164,7 +164,8 @@ StageAttributes::ptr StageAttributesManager::initNewObjectInternal(
   // copy
   StageAttributes::ptr newAttributes =
       this->constructFromDefault(attributesHandle);
-  if (nullptr == newAttributes) {
+  bool createNewAttributes = (nullptr == newAttributes);
+  if (createNewAttributes) {
     newAttributes = StageAttributes::create(attributesHandle);
   }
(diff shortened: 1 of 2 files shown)


Skylion007 (approved):

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -164,7 +164,8 @@ StageAttributes::ptr StageAttributesManager::initNewObjectInternal(
   // copy
   StageAttributes::ptr newAttributes =
       this->constructFromDefault(attributesHandle);
-  if (nullptr == newAttributes) {
+  bool createNewAttributes = (nullptr == newAttributes);
```
Skylion007, line 167:
```suggestion
  const bool createNewAttributes = (nullptr == newAttributes);
```

PR #1447, opened 2021-08-23
Bundle OpenEXR and enable import & conversion plugins

mosra:
## Motivation and Context

This PR enables OpenEXR that's needed for @bigbike's PBR IBL and HDR / depth image export. It adds support for importing and converting OpenEXR files with half-float (for color), float (for depth) and integer (for object ID) channels, including cubemaps and multiple mip levels per file.

Apart from that, the `magnum-imageconverter` tool got extended to make working with cubemap and multi-level EXR files easier. Running `--info` will give an overview of what's there:

```sh
$ magnum-imageconverter --info cube.exr
3D image 0: PixelFormat::RGB16F Vector(256, 256, 6)
$ magnum-imageconverter --info cube-mips.exr
3D image 0:
  Level 0: PixelFormat::RGB16F Vector(256, 256, 6)
  Level 1: PixelFormat::RGB16F Vector(128, 128, 6)
  Level 2: PixelFormat::RGB16F Vector(64, 64, 6)
  Level 3: PixelFormat::RGB16F Vector(32, 32, 6)
  Level 4: PixelFormat::RGB16F Vector(16, 16, 6)
  Level 5: PixelFormat::RGB16F Vector(8, 8, 6)
  Level 6: PixelFormat::RGB16F Vector(4, 4, 6)
  Level 7: PixelFormat::RGB16F Vector(2, 2, 6)
  Level 8: PixelFormat::RGB16F Vector(1, 1, 6)
```

The following will combine 6 2D cube map faces into a single 3D EXR file:

```sh
magnum-imageconverter --layers -c envmap=cube -- \
    +x.exr -x.exr +y.exr -y.exr +z.exr -z.exr cube.exr
```

This will make a file that contains several mipmap levels:

```sh
magnum-imageconverter --levels -c envmap=cube -D3 \
    cube-256.exr cube-128.exr cube-64.exr cube-mips.exr
```

And, as the inverse of above, this will extract a single level of a particular face back into a 2D image:

```sh
magnum-imageconverter cube-mips.exr --layer 2 --level 1 +x-128.exr
```

Please check out the full documentation for more information:

-   [`OpenExrImporter`](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1OpenExrImporter.html)
-   [`OpenExrImageConverter`](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1OpenExrImageConverter.html), note the compression options there
[...]

PR #1440, opened 2021-08-16
(Maint): Improve Code Style + Package Maint + Pre-Commit Update

Skylion007:
## Motivation and Context

* This simplifies the setup.py a bit and makes it a bit more pip installable.
* Removes the MANIFEST.in which wasn't really needed and wasn't quite correct (check-manifest complained alot).
* Removed tabs throughout the repo with a new pre-commit hook.
* Updated pre-commit

## How Has This Been Tested
* Locally and with CI

## Types of changes

- [X] Docs change / refactoring / dependency upgrade
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have completed my CLA (see **CONTRIBUTING**)
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #1448, opened 2021-08-23
[PBR][IBL] PBR image based lighting class

bigbike:
## Motivation and Context
This class will compute the irradiance map, prefiltered map, BRDF lookup table necessary for the PBR IBL.

## Types of changes

## Checklist


bigbike:
@mosra : this diff is ready to be reviewed. Thank you.

Skylion007:
Really would like to get the pre-commit hooks from #1440 run to make sure there no accidental tabs in the new shaders.

src/esp/gfx/PbrImageBasedLighting.cpp
```diff
@@ -0,0 +1,292 @@
...
+      .setWrapping(Mn::GL::SamplerWrapping::ClampToEdge)
+      .setStorage(1, Mn::GL::TextureFormat::RGBA8, size);  // TODO: HDR
+
+  // TODO: HDR!!
+  // we do not use build-in function `renderToTexture` in the CubeMap class. So
+  // we will have to populate the mipmaps by ourselves in this class.
+  environmentMap_ = CubeMap(
+      environmentMapSize,
+      {CubeMap::Flag::ColorTexture | CubeMap::Flag::ManuallyBuildMipmap});
+  irradianceMap_ = CubeMap(irradianceMapSize, {CubeMap::Flag::ColorTexture});
```
mosra, line 137: Why the `{}` around the argument here? :)

src/esp/gfx/PbrImageBasedLighting.cpp
```diff
@@ -0,0 +1,292 @@
...
+namespace Cr = Corrade;
+
+namespace esp {
+namespace gfx {
+
+namespace {
+const unsigned int environmentMapSize = 1024;
+const unsigned int prefilteredMapSize = 1024;
+const unsigned int irradianceMapSize = 128;
+const unsigned int brdfLUTSize = 512;
```
mosra, line 43: constexpr these?

src/esp/gfx/PbrImageBasedLighting.cpp
```diff
@@ -0,0 +1,292 @@
...
+  Cr::PluginManager::Manager<Mn::Trade::AbstractImporter> manager;
+  std::string importerName{"AnyImageImporter"};
+  Cr::Containers::Pointer<Mn::Trade::AbstractImporter> importer =
+      manager.loadAndInstantiate(importerName);
+  CORRADE_INTERNAL_ASSERT(importer);
+
+  // TODO: HDR, No LDR in the future!
+  // temporarily using the brdflut from here:
+  // https://github.com/SaschaWillems/Vulkan-glTF-PBR/blob/master/screenshots/tex_brdflut.png
+  std::string filename = "./data/pbr/brdflut_ldr_512x512.png";
```
mosra, line 199: Hm, wouldn't it make sense to embed the LUT as a resource, the same way shader sources are? This won't work on Emscripten, for example, and generally it will also break if you don't run from the top-level directory.

src/esp/gfx/PbrImageBasedLighting.h
```diff
@@ -0,0 +1,199 @@
...
+    // BrdfLookupTable = 2, // TODO
+    EquirectangularToCubeMap = 3,
+  };
+  /**
+   * @brief get the shader based on the type
+   * @param[in] type, see @ref PbrIblShaderType
+   */
+  template <typename T>
+  Mn::Resource<Mn::GL::AbstractShaderProgram, T> getShader(
+      PbrIblShaderType type) {
```
mosra, line 147: This template function is private and only ever used from the cpp file, so its definition could be in the cpp -- no need to have it in the header.

mosra (commented):
Usual annoying question: _tests_? :P

Another one -- what would it take to make it HDR-ready already, using the EXR support from #1447? Or do you explicitly want to do keep this LDR and enable HDR in a second step?

bigbike:
>Usual annoying question: tests? :P

No, it is not annoying. Thank you for the reminder. This code has been tested perhaps a couple of hundred times in #1389.
For the unit test, I am planning to do the pixel comparisons like the existing tests for Phong shading. But it could only happen when the entire IBL functionality is put in code base.

>Another one -- what would it take to make it HDR-ready already, using the EXR support from #1447? Or do you explicitly want to do keep this LDR and enable HDR in a second step?

As we discussed before, the 1st pass is to make the IBL work using LDR, and another pass to convert it to HDR. It is rather error-prone to do it now in the middle of this project.

PR #1451, opened 2021-08-24
change frame color buffer from srgb to linear (RGBA8)

eundersander:
## Motivation and Context

We were requesting an srgb frame color buffer, which didn't affect most scenarios but does cause a visual problem for WebXR (where we end up copying the color buffer to another canvas). 

From Mosra:
>Huh. The rendering pipeline is certainly not sRGB-correct at the moment, so it definitely shouldn't render to a sRGB framebuffer... Use RGBA8 there, yes, for all platforms. It worked on desktop GL only because the Renderer::Feature::FramebufferSrgb is off by default there (as you discovered already) and so the sRGB attachment worked the same as plain RGB.

[Slack thread with more context](https://cvmlp.slack.com/archives/C01NZ9Z0VKL/p1629750596041700)

## How Has This Been Tested

Ad hoc testing in WebXR build and C++ viewer. Our graphical unit tests should also help here. (We don't have any WebXR graphical tests, so no graphical tests need to be updated with this change.)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/RenderTarget.cpp
+++ b/src/esp/gfx/RenderTarget.cpp
@@ -66,7 +66,7 @@ struct RenderTarget::Impl {
     }
 
     if (flags_ & Flag::RgbaAttachment) {
-      colorBuffer_.setStorage(Mn::GL::RenderbufferFormat::SRGB8Alpha8, size);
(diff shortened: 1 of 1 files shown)


mosra (approved):
Of course I still have to fix the whole pipeline to make it sRGB-correct, but what was done here was clearly wrong, so this is a right fix.

PR #1458, opened 2021-08-30
ArticulatedObjects: Enable support for secondary json config files.

jturner65:
## Motivation and Context
This PR introduces support of Json config files for URDF-based ArticulatedObject models in the URDFParser, which will hold default user-defined attributes (Such as bounding boxes for placements) that apply to all instances of articulated objects built from these models.   This PR also removed extraneous markers from error messages that did not match the format of our current message logging in the URDFParser.

To be compatible with this functionality, the json config file holding the user_defined attributes must meet these criteria : 
- Must be in the same directory as the AO's .urdf file
- Must have the same core filename (minus extensions) as the AO's .urdf file
- Must have the following extension :  <filename>**.ao_config.json**
- Internally, the file must have a root Json object, and within that object there must be a tag "user_defined" which references a Json object holding the user-defined values.

## How Has This Been Tested
Locally c++ and python tests.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/io/URDFParser.cpp
+++ b/src/esp/io/URDFParser.cpp
@@ -103,14 +103,13 @@ void Model::setMassScaling(float massScaling) {
 }
 
 bool Model::loadJsonAttributes(const std::string& filename) {
-  // rebuild name
   namespace CrUt = Corrade::Utility;
-  namespace CrUtDir = CrUt::Directory;
   const std::string jsonName = CrUt::formatString(
-      "{}.ao_config.json",
-      CrUtDir::splitExtension(CrUtDir::splitExtension(filename).first).first);
+      "{}.ao_config.json", CrUt::Directory::splitExtension(
+                               CrUt::Directory::splitExtension(filename).first)
+                               .first);
 
-  if (!CrUtDir::exists(jsonName)) {
+  if (!CrUt::Directory::exists(jsonName)) {
     // file does not exists, so no Json configuration defined for this Model.
     return false;
   }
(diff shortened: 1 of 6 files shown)


src/esp/physics/PhysicsManager.cpp
```diff
@@ -304,9 +304,9 @@ int PhysicsManager::addArticulatedObjectInstance(
   existingArticulatedObjects_.at(aObjID)->setSceneInstanceAttr(
       aObjInstAttributes);
 
-  // set articulated object's user-defined attributes, if any exist in scene
+  // merge articulated object's user-defined attributes, if any exist in scene
   // instance.
-  existingArticulatedObjects_.at(aObjID)->setUserAttributes(
+  existingArticulatedObjects_.at(aObjID)->mergeUserAttributes(
```
Skylion007, line 309: Nit: calling at(aObjId) is not free. The reference should really be cached between these calls, don't you think?

src/esp/io/URDFParser.cpp
```diff
@@ -185,13 +240,13 @@ bool Parser::parseURDF(std::shared_ptr<Model>& urdfModel,
         urdfModel->m_links[link->m_name] = link;
       }
     } else {
-      ESP_ERROR() << "E - failed to parse link. Aborting parse/load for"
+      ESP_ERROR() << "Failed to parse link. Aborting parse/load for"
                   << filename;
       return false;
     }
   }
   if (urdfModel->m_links.size() == 0) {
```
Skylion007, line 248:
```suggestion
  if (urdfModel->m_links.empty()) {
```

Skylion007 (approved):

src/esp/physics/PhysicsManager.cpp
jturner65: Every little bit helps, yeah

aclegg3 (approved):
Looks great, thanks for cleaning up. 👍

PR #1461, opened 2021-08-30
[PBR][IBL] Upgrade PbrDrawable and PbrShader to adopt IBL.

bigbike:
## Motivation and Context
As titled.

This is to upgrade `PbrDrawable` and `PbrShader` classes to apply the IBL.

Test result:
pbr-3

## How Has This Been Tested
Code has been fully tested in the #1389 .

## Types of changes

## Checklist


src/esp/gfx/PbrShader.cpp
```diff
@@ -238,7 +262,15 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
     setLightRanges(Cr::Containers::Array<Mn::Float>{Cr::DirectInit, lightCount_,
                                                     Mn::Constants::inf()});
   }
+
   setEmissiveColor(Magnum::Color3{0.0f});
+  PbrShader::PbrEquationScales scales;
+  if (flags_ & Flag::ImageBasedLighting) {
+    scales.IblDiffuse = 0.8;
```
Skylion007, line 269: Comments for magic numbers?

src/esp/gfx/PbrShader.cpp
```diff
@@ -318,9 +374,17 @@ PbrShader& PbrShader::setObjectId(unsigned int objectId) {
...
+                 "created with image based lighting enabled",
+                 *this);
+  setUniform(prefilteredMapMipLevelsUniform_, mipLevels);
+  return *this;
+}
+
 PbrShader& PbrShader::setBaseColor(const Mn::Color4& color) {
-  // TODO: remove this constraint after IBL is introduced.
-  if (lightCount_ != 0u) {
+  if (lightCount_ != 0u || flags_ & Flag::ImageBasedLighting) {
```
Skylion007, line 387: This snippit s repeated a lot, maybe you should factor it out into a method that returns a bool

src/esp/gfx/PbrShader.h
```diff
@@ -10,6 +10,7 @@
 #include <Corrade/Containers/ArrayView.h>
 #include <Corrade/Containers/EnumSet.h>
 #include <Magnum/GL/AbstractShaderProgram.h>
+#include <Magnum/GL/CubeMapTexture.h>
```
mosra, line 13: FYI, for reducing compile times you can do the include just in the cpp file (and then of course in files that actually use it, but it's there already anyway), here a forward declaration is enough.

mosra (commented):
I appreciate the effort you put into splitting the original work into smaller PRs, but I feel like the parts became too tiny -- I was looking for the corresponding shader code changes to see how the uniforms are actually organized but didn't find them.

Yes, I advocate for small PRs with limited scope, but the change have to be *atomic*, being able to run on its own. If I only see parts of the code though a "tunnel vision" without any relationship to the rest, I can only comment on technical details but not on the overall design or usability.

src/esp/gfx/PbrTextureUnit.h
```diff
@@ -22,6 +22,8 @@ enum TextureUnit : uint8_t {
   BrdfLUT = 6,
   PrefilteredMap = 7,
   ShadowMap0 = 8,
+  ShadowMap1 = 9,
+  ShadowMap2 = 10,  // We support at most 3 shadows (TODO)
```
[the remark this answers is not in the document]
bigbike: This has nothing to do with the IBL. But I will use it later. Sorry.

src/esp/gfx/PbrShader.h
bigbike: Yeah, good catch. I did not pay attention to it. will change.

src/esp/gfx/PbrShader.cpp
```diff
@@ -332,21 +396,32 @@ PbrShader& PbrShader::setEmissiveColor(const Magnum::Color3& color) {
...
     setUniform(metallicUniform_, metallic);
   }
   return *this;
 }
 
+PbrShader& PbrShader::setPbrEquationScales(const PbrEquationScales& scales) {
+  setUniform(scaleDirectDiffuseUniform_, scales.DirectDiffuse);
+  setUniform(scaleDirectSpecularUniform_, scales.DirectSpecular);
+  setUniform(scaleIblDiffuseUniform_, scales.IblDiffuse);
+  setUniform(scaleIblSpecularUniform_, scales.IblSpecular);
```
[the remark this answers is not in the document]
bigbike: sigh... good catch. Forgot to add the source.

bigbike:
> I appreciate the effort you put into splitting the original work into smaller PRs, but I feel like the parts became too tiny -- I was looking for the corresponding shader code changes to see how the uniforms are actually organized but didn't find them.
> 
> Yes, I advocate for small PRs with limited scope, but the change have to be _atomic_, being able to run on its own. If I only see parts of the code though a "tunnel vision" without any relationship to the rest, I can only comment on technical details but not on the overall design or usability.

My only purpose was to make your life a bit easier. Sometimes it is really hard to control. I can make it bigger.

src/esp/gfx/PbrShader.cpp
```diff
@@ -332,21 +396,32 @@ PbrShader& PbrShader::setEmissiveColor(const Magnum::Color3& color) {
...
 PbrShader& PbrShader::setMetallic(float metallic) {
-  // TODO: remove this constraint after IBL is introduced.
-  if (lightCount_ != 0u) {
+  if (lightCount_ != 0u || flags_ & Flag::ImageBasedLighting) {
     setUniform(metallicUniform_, metallic);
   }
   return *this;
 }
 
+PbrShader& PbrShader::setPbrEquationScales(const PbrEquationScales& scales) {
```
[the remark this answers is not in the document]
bigbike: In my opinion, it is less error prone and improves readability in this way. We discussed this before.
if a function is designed like this, `void foo(float, float, float, float)`, I will have to check the signature every time when I need to call it. And I will have to add comments after each parameter like this (due to habit) to make sure I would not pass wrong parameters.

```
foo(1.0, // direct diffuse
    1.0, // direct specular
         ... 
);

```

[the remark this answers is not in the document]
mosra: Ah, of course. Sorry, yes, this makes more sense then. 

Or, to solve that alternatively, have four different uniform setters, contrary to my other comment, which won't require you to have a custom struct either. Readability/usability should go first, and given that we will have to switch to UBOs for multidraw & Vulkan anyway, the performance concern doesn't matter that much.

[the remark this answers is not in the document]
bigbike: @mosra:
I combined what you suggested (using `vec4` in the shader) and the current data structure.

bigbike:
@mosra :
This diff can be reviewed again.

src/esp/assets/ResourceManager.h
```diff
@@ -1108,6 +1128,16 @@ class ResourceManager {
    * @brief See @ref setRecorder.
    */
   std::shared_ptr<esp::gfx::replay::Recorder> gfxReplayRecorder_;
+
+  /**
+   * @brief The imaged based lighting for PBR, each is a collection of
+   * an environment map, an irradiance map, a BRDF lookup table (2D texture),
+   * and a pre-fitered map
+   */
+  std::map<int, std::unique_ptr<esp::gfx::PbrImageBasedLighting>>
```
Skylion007, line 1137: Shouldn't this be an std::unordered_map? Is the order important?

bigbike: It has nothing to do with the order. Also It will not make too much difference in terms of performance. 

`unordered_map` is implemented in a linked list and thus extremely not cache-friendly. So even though the time complexity in theory is O(1) for finding and inserting an element, the overhead is actually very high. Avoiding `unordered_map` or `unordered_set` is almost a common sense. I rarely use them.

`map` has insertion time `O(logN)`. However in our usage case, perhaps we will never ever require, say 10 of such an element in the future, so you can safely assume the time cost is approximately constant.

src/esp/gfx/PbrShader.h
```diff
@@ -14,8 +14,13 @@
 
 #include "esp/core/esp.h"
 
-namespace esp {
+namespace Magnum {
+namespace GL {
+class CubeMapTexture;
+};
+};  // namespace Magnum
```
mosra, line 21: `#include <Magnum/GL/GL.h>` instead, it's a header with all forward declarations for the GL namespace. [Docs.](https://doc.magnum.graphics/magnum/compilation-speedup.html#compilation-forward-declarations)

Doing forward declarations manually is *not recommended* because it might cause build breakages in the future (the class might become a typedef and vice versa).

src/shaders/pbr.frag
```diff
@@ -221,16 +282,40 @@ void microfacetModel(vec3 specularReflectance,
...
   specularContrib = specular * tempVec;
 }
 
+#if defined(IMAGE_BASED_LIGHTING)
+// c_diff: diffuse color
+// n: normal on shading location in world space
+vec3 computeIBLDiffuse(vec3 c_diff, vec3 n) {
+  // diffuse part = c_diff * irradiance
+  // return c_diff * texture(IrradianceMap, n).rgb * Scales.iblDiffuse;
+  return c_diff * SRGBtoLINEAR(tonemap(texture(IrradianceMap, n))).rgb *
```
mosra, line 301: Wait, doing sRGB->linear conversion _after_ applying a tonemap? That's definitely not right.

Actually, could you just drop the whole `SRGBtoLINEAR` function altogether, as we talked about before? It doesn't do anything at the moment because `MANUAL_SRGB` isn't defined, here it's even used incorrectly and the operation should be done by the hardware anyway. So even once our pipeline becomes sRGB-correct, the function would have no use here, the hardware does that instead.

src/shaders/pbr.frag
```diff
@@ -328,26 +413,65 @@ void main() {
...
+  fragmentColor.rgb += iblSpecularContrib;
+#endif  // IMAGE_BASED_LIGHTING
+
 #if defined(OBJECT_ID)
   fragmentObjectId = ObjectId;
 #endif
-}
+  // PBR equation debug
+  // "none", "Diff (l,n)", "F (l,h)", "G (l,v,h)", "D (h)", "Specular"
+  if (PbrDebugDisplay > 0) {
```
mosra, line 445: I have a feeling that this should be hidden under a define -- even if not used, the extra shader instructions may make the code run slower than if it wouldn't be there at all.

src/utils/viewer/viewer.cpp
```diff
@@ -774,6 +778,11 @@ Viewer::Viewer(const Arguments& arguments)
               << simConfig_.sceneDatasetConfigFile
               << "| Loading Scene :" << simConfig_.activeSceneName;
 
+  // image based lighting (PBR)
+  if (args.isSet("ibl")) {
+    simConfig_.pbrImageBasedLighting = true;
+  }
```
mosra, line 784:
```suggestion
  simConfig_.pbrImageBasedLighting = args.isSet("ibl");
```

(sorry, had to make use of the opportunity :D)

src/esp/assets/ResourceManager.h
mosra: > is implemented in a linked list and thus extremely not cache-friendly

That's only partially correct. Neither `map` nor `unordered_map` are cache friendly, both are extremely awful _by design_ as each element has to be a separate allocation in order to satisfy the iterator invalidation requirements. The linked list only gets used when more than element falls into the same bucket, but with ~10 elements it's unlikely.

> Avoiding `unordered_map` or `unordered_set` is almost a common sense.

I'd say otherwise -- the memory-jumping overhead of balancing a tree for `map` insertion (or memory jumps for doing the binary search) is _bigger_ than basically a single lookup into a hashmap bucket. The "common sense" is to avoid any STL maps, be it a tree or a hashmap, but when you have no better option (because I didn't get to implementing better alternatives in Magnum yet), an `unordered_map` is less evil than a `map` ;)

bigbike: Good to know something new. Thank you, @mosra .

Considering the problem scale, I do not think there is too much difference between `map` and `unordered_map`. (We only have 1 environment map which may stay for quite a long time unless we would like to have different environment map for different rooms. e.g.). 

So I will just keep the `map` here.

src/esp/gfx/PbrShader.h
bigbike: >it's a header with all forward declarations
Good. Hope next time you can tell me in your first comment so I can avoid this iteration. :)

src/shaders/pbr.frag
bigbike: I understand ideally it should contain no branching in the shader. But a single `if` cause is probably OK and will not cause big difference. No?

src/esp/assets/ResourceManager.h
bigbike: You know what, I changed it to `vector` given the fact that the environment maps are static and will not be dynamically created or destroyed during the runtime (unlike the dynamic object). So I am now using the `vector` here.

bigbike:
@mosra : you can take another look when you have time. Thank you.

src/shaders/pbr.frag
mosra: It was more about the extra shader instructions, not the branching. More instructions will cause the compiled shader binary to be bigger, which may result in more instruction cache, more registers / ALUs / memory accesses being needed for each execution, which then might mean the GPU can't parallelize as much. You probably remember the hard limits on shader instruction count back "in the old days", that's no longer a problem fortunately but it can still happen that for example each core now has to use a twice as big register file / instruction cache and thus only half of the cores can be run at a time.

Yes, a GPU is mostly a black box, hard to reason about, and every hardware is different, but I think it won't hurt to have this `#ifdef`'d. Of course the driver might see the opportunity on its own and recompile the shader based on how this uniform is set so you won't see any difference.

bigbike: Yeah, it makes sense to me. Thank you.

mosra (approved):

PR #1465, opened 2021-09-02
--Use subconfigs instead of local vectors for scene instance and light attributes

jturner65:
## Motivation and Context
Use subconfig system instead of redundant member variable vectors to store stage, object and articulated object scene instances, and lighting configs.  Lights and scene instances are now supported.

## How Has This Been Tested
Locally c++ and python tests pass.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/Configuration.cpp
+++ b/src/esp/core/Configuration.cpp
@@ -297,10 +297,9 @@ int findValueInternal(const Configuration& config,
     // if found, will be greater than curLevel
     if (resLevel > curLevel) {
       return resLevel;
-    } else {
-      // remove subconfig key from breadcrumb
-      breadcrumb.pop_back();
     }
+    // remove subconfig key from breadcrumb
+    breadcrumb.pop_back();
   }
   // if not found, return lowest level having been checked
   return parentLevel;
@@ -323,11 +322,6 @@ Configuration& Configuration::operator=(const Configuration& otr) {
       configMap_[entry.first] = std::make_shared<Configuration>(*entry.second);
     }
   }
-  return *this;
-}
-Configuration& Configuration::operator=(Configuration&& otr) noexcept {
-  configMap_ = std::move(otr.configMap_);
-  valueMap_ = std::move(otr.valueMap_);
   return *this;
 }
 
--- a/src/esp/core/Configuration.cpp
+++ b/src/esp/core/Configuration.cpp
@@ -225,6 +225,15 @@ bool ConfigValue::putValueInConfigGroup(
   }  // switch
 }  // ConfigValue::putValueInConfigGroup
 
+Mn::Debug& operator<<(Mn::Debug& debug, const ConfigStoredType& value) {
+  return debug << "Type:" << getNameForStoredType(value);
+}
+
+Mn::Debug& operator<<(Mn::Debug& debug, const ConfigValue& value) {
+  return debug << "ConfigValue :(" << Mn::Debug::nospace << value.getAsString()
+               << "|" << value.getType() << Mn::Debug::nospace << ")";
+}
+
 /**
  * @brief Retrieves a shared pointer to a copy of the subConfig @ref
  * esp::core::Configuration that has the passed @p name . This will create a
@@ -262,13 +271,39 @@ void Configuration::setSubconfigPtr<Configuration>(
   configMap_[name] = std::move(configPtr);
 }  // setSubconfigPtr
 
-Mn::Debug& operator<<(Mn::Debug& debug, const ConfigStoredType& value) {
-  return debug << "Type:" << getNameForStoredType(value);
-}
+int Configuration::findValueInternal(
+    const std::string& key,
+    int parentLevel,
+    std::vector<std::string>& breadcrumb) const {
+  int curLevel = parentLevel + 1;
+  if (valueMap_.count(key) > 0) {
+    // Found at this level, access directly via key to get value
+    breadcrumb.push_back(key);
+    return curLevel;
+  }
 
-Mn::Debug& operator<<(Mn::Debug& debug, const ConfigValue& value) {
-  return debug << "ConfigValue :(" << Mn::Debug::nospace << value.getAsString()
-               << "|" << value.getType() << Mn::Debug::nospace << ")";
+  // not found by here in data maps, check subconfigs, to see what level
+  for (const auto& subConfig : configMap_) {
+    if (subConfig.first == key) {
+      // key matches name of subconfiguration
+      breadcrumb.push_back(key);
+      return curLevel;
+    }
+    // add subconfig key to breadcrumb
+    breadcrumb.push_back(subConfig.first);
+    // search this subconfiguration
+    int resLevel =
+        subConfig.second->findValueInternal(key, curLevel, breadcrumb);
+    // if found, will be greater than curLevel
+    if (resLevel > curLevel) {
+      return resLevel;
+    } else {
+      // remove subconfig key from breadcrumb
+      breadcrumb.pop_back();
+    }
+  }
+  // if not found, return lowest level having been checked
+  return parentLevel;
 }
 
 }  // namespace config
--- a/src/esp/core/Configuration.cpp
+++ b/src/esp/core/Configuration.cpp
@@ -225,6 +225,43 @@ bool ConfigValue::putValueInConfigGroup(
   }  // switch
 }  // ConfigValue::putValueInConfigGroup
 
+/**
+ * @brief Retrieves a shared pointer to a copy of the subConfig @ref
+ * esp::core::Configuration that has the passed @p name . This will create a
+ * pointer to a new sub-configuration if none exists already with that name,
+ * but will not add this configuration to this Configuration's internal
+ * storage.
+ *
+ * @param name The name of the configuration to retrieve.
+ * @return A pointer to a copy of the configuration having the requested
+ * name, or a pointer to an empty configuration.
+ */
+
+template <>
+std::shared_ptr<Configuration> Configuration::getSubconfigCopy<Configuration>(
+    const std::string& name) const {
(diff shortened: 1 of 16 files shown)


src/esp/metadata/attributes/SceneAttributes.h
```diff
@@ -450,6 +461,60 @@ class SceneAttributes : public AbstractAttributes {
...
+  }
+  res.reserve(numSubconfigs);
+  auto objInstIter = objInstConfig->getSubconfigIterator();
+  // iterate through subconfig entries, casting appropriately and adding to
+  // map if cast successful
+  for (auto objIter = objInstIter.first; objIter != objInstIter.second;
+       ++objIter) {
+    auto obj = objIter->second;
+    if (std::shared_ptr<T> objPtr = std::dynamic_pointer_cast<T>(obj)) {
+      res.push_back(objPtr);
```
Skylion007, line 507:
```suggestion
      res.push_back(std::move(objPtr));
```
or better yet
```suggestion
      res.emplace_back(std::move(objPtr));
```

jturner65: I want a copy to save in the object that it creates.

Skylion007: @jturner, it's a shared ptr though, no? So that should just more efficiently transfer the reference.

jturner65: Yeah, you're right. Thanks :)

src/esp/core/Configuration.h
```diff
@@ -218,14 +218,13 @@ class Configuration {
  public:
   // convenience typedefs
   typedef std::unordered_map<std::string, ConfigValue> ValueMapType;
-  typedef std::unordered_map<std::string, std::shared_ptr<Configuration>>
-      ConfigMapType;
+  typedef std::map<std::string, std::shared_ptr<Configuration>> ConfigMapType;
 
   Configuration() = default;
 
   Configuration(const Configuration& otr)
```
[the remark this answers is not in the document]
jturner65: Actually these should not be copy assignable - they are to follow the same behavior as the original esp::core::configuration, which had a default copy constructor but no copy assignment.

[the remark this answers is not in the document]
Skylion007: Then the copy operator should explicitly be deleted if it's not copy assignable.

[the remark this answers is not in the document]
jturner65: > 
> Actually these should not be copy assignable - they are to follow the same behavior as the original esp::core::configuration, which had a default copy constructor but no copy assignment.

This is a mistake, my apologies (what I get for responding on my phone in the car).  The original class made use of the default copy ctor/copy assignment, which was sufficient to manage the underlying Corrade configurationGroup.  I'll add the copy assignment when I get back.

src/esp/core/Configuration.cpp
```diff
@@ -234,6 +234,103 @@ Mn::Debug& operator<<(Mn::Debug& debug, const ConfigValue& value) {
...
+      configMap_[entry.first] = std::make_shared<Configuration>(*entry.second);
+    }
+  }
+  return *this;
+}
+Configuration& Configuration::operator=(Configuration&& otr) noexcept {
+  configMap_ = std::move(otr.configMap_);
+  valueMap_ = std::move(otr.valueMap_);
+  return *this;
+}
```
Skylion007, line 332:
```suggestion
Configuration& Configuration::operator=(Configuration&& otr) noexcept = default;
```

src/esp/core/Configuration.cpp
```diff
@@ -234,6 +234,103 @@ Mn::Debug& operator<<(Mn::Debug& debug, const ConfigValue& value) {
...
+    }
+    // add subconfig key to breadcrumb
+    breadcrumb.push_back(subConfig.first);
+    // search this subconfiguration
+    int resLevel =
+        findValueInternal(*subConfig.second, key, curLevel, breadcrumb);
+    // if found, will be greater than curLevel
+    if (resLevel > curLevel) {
+      return resLevel;
+    } else {
```
Skylion007, line 300: Please remove else return

Skylion007 (changes requested):

src/esp/metadata/attributes/AttributesBase.cpp
```diff
@@ -55,6 +55,36 @@ std::string getTranslationOriginName(int translationOrigin) {
...
+AbstractAttributes& AbstractAttributes::operator=(
+    const AbstractAttributes& otr) {
+  if (this != &otr) {
+    this->Configuration::operator=(otr);
+  }
+  return *this;
+}
+AbstractAttributes& AbstractAttributes::operator=(
+    AbstractAttributes&& otr) noexcept {
+  this->Configuration::operator=(std::move(otr));
```
Skylion007, line 84:
```suggestion
  this->Configuration = std::move(otr);
```
isn't this equivalent?

jturner65: Don't think it works that way.  Configuration is a base class for the AbstractAttributes.

jturner65: Although that would be really cool, if you could assign to a base class component of a derived class that way.

src/esp/metadata/attributes/LightLayoutAttributes.cpp
```diff
@@ -39,19 +39,57 @@ LightLayoutAttributes::LightLayoutAttributes(const std::string& handle)
...
+}
+LightLayoutAttributes& LightLayoutAttributes::operator=(
+    LightLayoutAttributes&& otr) noexcept {
+  // point to our own light instance config and available ids
+  availableLightIDs_ = std::move(otr.availableLightIDs_);
+  this->AbstractAttributes::operator=(
+      std::move(static_cast<AbstractAttributes>(otr)));
+  lightInstConfig_ = editSubconfig<Configuration>("light_instances");
+  return *this;
 }
```
mosra, line 79: I find it a bit problematic to have to manually implement copy/move constructors and assignments for the subclasses. Is it really essential to have the `lightInstConfig_` member manually cached here? What is in it that can't be implicitly managed by the base class already? I thought the base takes care of such hierarchic configuration on its own, and so when there's a config group named `light_instances` then a copy constructor does the right thing and copies it as well?

Same question about `SceneAttributes`.

mosra (commented):
Apologies for a potentially out-of-touch comment, I'm not very familiar with this part of the codebase, but ... I'm not sure about the direction of this set of changes. It introduces a large amount of boilerplate and maintenance overhead for something that I think should be doable in an easier way. Again, I'm not familiar with all the code, and there's a lot of it, so it might be that I'm missing some key aspect. But what I see is:

- there's a base `Configuration` class with generic (but typed) key/value store
- then there's `AbstractAttributes` that derives from it, plus derives also from some other abstract interface (which I don't understand either so I'll ignore it here)
- and then there's convenience subclasses for light info, scene info etc that provide shorthand setters & getters

Right? To me that looks something like `MaterialData` vs `PhongMaterialData` / `PbrMetallicRoughnessMaterialData` / ... in Magnum, where the base class has a generic interface, and the subclasses provide convenience accessors, queries and fallbacks to a default value in case some attribute is missing. The key design goal I had in mind when writing these was to make adding new subclasses as frictionless as possible, and I think the same approach should be picked here with `LightLayoutAttributes` & `SceneAttributes`:

0. Have the base class contain all data, in a generic way (which your PR already does, by switching to a subconfig).
1. Have the subclasses with *no* state whatsoever, because otherwise (and especially with internal pointers) you have to deal with copy/move construction manually, which is a massive error-prone annoyance. Without a state, the subclass is just the convenience getters alone, querying the base class key/value storage.
2. Have the subclasses easily createable from the base class. In case of Magnum, there's [MaterialData::as<T>()](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1MaterialData.html#a2c217a9af1f687c086392510f5a6353d), which basically returns just `static_cast<const T&>(*this)`, which is _completely fine_ if the subclass has no state on its own, although C++ purists and overmotivated Clang Tidy checks may disagree. Alternatively. if that would be a no-go, it's possible to create external "adaptors" that provide a "view" onto the `Configuration` with desired convenience APIs.
3. Then, if it's so easy to get a convenience subclass from the base, it's no longer needed to inflate the `Configuration` class with templated `conf.editSubconfig<Foo>(name)` etc (and then have to implement such possibility on every interface that returns a `Configuration`), instead one would simply do `Foo(conf.editSubconfig(name))`, for example. This also makes it possible to switch to a convenience subclass anywhere else in the code, if you already have a pre-existing `Configuration` instance.

Does that sound doable? Or did I forget some use case which makes this impossible?

src/esp/core/Configuration.cpp
```diff
@@ -234,6 +234,97 @@ Mn::Debug& operator<<(Mn::Debug& debug, const ConfigValue& value) {
...
+}
+
+template <>
+void Configuration::setSubconfigPtr<Configuration>(
+    const std::string& name,
+    std::shared_ptr<Configuration>& configPtr) {
+  configMap_[name] = std::move(configPtr);
+}  // setSubconfigPtr
+
+int findValueInternal(const Configuration& config,
```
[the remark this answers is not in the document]
jturner65: I tried putting that in an anonymous namespace but it complained.  I'll try again.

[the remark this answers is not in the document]
Skylion007: Clang-Tidy is fine with anonymous namespaces as long as they have a corresponding comment.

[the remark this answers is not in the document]
jturner65: Nothing to do with clang-tidy. Function is a friend of Configuration to access protected members. Won't work in a namespace, so I made it static.

aclegg3 (approved):
LGTM

PR #1466, opened 2021-09-02
Add cache and query for ArticulatedObject global scaling.

aclegg3:
## Motivation and Context

Articulated objects can be programmatically scaled upon instantiation in the scene. Any geometric metadata defined in the original frame should be scaled also, requiring that this global scale is cached when the object is created and queryable later.

## How Has This Been Tested

Added a simple test in python.

## Types of changes

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ matplotlib
 numba
 numpy
 numpy-quaternion
-pillow
+pillow<=8.3.1
 scipy>=1.3.0
 tqdm
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ matplotlib
 numba
 numpy
 numpy-quaternion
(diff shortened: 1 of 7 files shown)


jturner65 (approved):
LGTM.  How did I miss this? :)

requirements.txt
```diff
@@ -6,6 +6,6 @@ matplotlib
 numba
 numpy
 numpy-quaternion
-pillow
+pillow<=8.3.1
```
Skylion007, line 9: Forcing requirements because of wheel dependencies is a silly and if you HAVE to do it, then use the magic python version comments.

Skylion007 (changes requested):

requirements.txt
aclegg3: Thanks for the pip magic.

PR #1473, opened 2021-09-10
--Hide Configuration Setters in Attributes for safety

jturner65:
## Motivation and Context
The Configuration class is a robust and flexible mechanism for storing and accessing hierarchical aggregations of data, providing templated setters and getters that will save and retrieve data of various types.  

The AbstractAttributes class, which inherits from the Configuration class, provides the base for the Attributes classes that define various components of the Habitat-sim simulation and dataset ecosystem.  The purpose of these various Attributes is to provide full and safe access to all the data fields pertinent to the constructs they define.

This PR makes the Configuration base class's templated setter in AbstractAttributes protected so that it can only be accessed from the derived implementations, removing external access from any AbstractAttributes-derived class object.  This is so that a user does not set a critical value to an incorrect type accidentally, by accessing that value via the Configuration bindings instead of the provided Attributes property bindings.  This would throw an error and terminate execution should the Attributes-based accessor be used to retrieve the value. 

## How Has This Been Tested
Local C++ and python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/attributes/AttributesBase.h
+++ b/src/esp/metadata/attributes/AttributesBase.h
@@ -225,6 +225,15 @@ class AbstractAttributes : public esp::core::AbstractFileBasedManagedObject,
 
  protected:
   /**
+   * @brief Changing access to setter so that Configuration bindings cannot be
+   * used to set a resserved value to an incorrect type. The inheritors of this
+   * class provide an interface to set the values that the consumers of these
+   * Attributes depend on and this should be the only mechanism capable of doing
+   * so.
+   */
+  using Configuration::set;
+
+  /**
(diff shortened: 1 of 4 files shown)


src/esp/metadata/attributes/AttributesBase.h
```diff
@@ -224,6 +224,15 @@ class AbstractAttributes : public esp::core::AbstractFileBasedManagedObject,
   }
 
  protected:
+  /**
+   * @brief Changing access to setter so that Configuration bindings cannot be
+   * used to set a resserved value to an incorrect type. The inheritors of this
```
Skylion007, line 229: typo: reserved

aclegg3 (approved):
LGTM

PR #1480, opened 2021-09-15
[Bugfix]: make corrade plugins static

Skylion007:
## Motivation and Context

* Previously our corrade plugins were not statically linked which complicated the rpath resolution. Apparently, this was a bug and can be fixed with a simple cmake change which ensure that corrade is built statically as well.
* Now that this change is made, it vastly simplifies the rpath issues we were having before since Magnum and Habitat-Sim are both separately and statically linked.
* Fixes a bug where importing magnum would fail if habitat_sim was not imported first (due to shared lib dependencies installed in the habitat_sim folder.

## How Has This Been Tested
* Locally on OSX and on CI.

## Types of changes

## Checklist

### DIFF
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -11,9 +11,9 @@ if(NOT USE_SYSTEM_MAGNUM)
   # for slightly faster builds. If you need any of these, simply delete a line.
   set(WITH_INTERCONNECT OFF CACHE BOOL "" FORCE)
   # Ensure Corrade should be built statically if Magnum is.
-  set(BUILD_PLUGINS_STATIC ON CACHE BOOL "BUILD_PLUGINS_STATIC" FORCE)
-  set(BUILD_STATIC ON CACHE BOOL "BUILD_STATIC" FORCE)
-  set(BUILD_STATIC_PIC ON CACHE BOOL "BUILD_STATIC_PIC" FORCE)
+  set(BUILD_PLUGINS_STATIC ON CACHE BOOL "" FORCE)
+  set(BUILD_STATIC ON CACHE BOOL "" FORCE)
+  set(BUILD_STATIC_PIC ON CACHE BOOL "" FORCE)
   add_subdirectory("${DEPS_DIR}/corrade")
 endif()
 find_package(Corrade REQUIRED Utility)
@@ -172,9 +172,7 @@ endif()
 
 # Magnum. Use a system package, if preferred.
 if(NOT USE_SYSTEM_MAGNUM)
-  set(BUILD_PLUGINS_STATIC ON CACHE BOOL "BUILD_PLUGINS_STATIC" FORCE)
-  set(BUILD_STATIC ON CACHE BOOL "BUILD_STATIC" FORCE)
-  set(BUILD_STATIC_PIC ON CACHE BOOL "BUILD_STATIC_PIC" FORCE)
+  # Magnum is already set to be build statically when Corrade is above.
 
   # These are enabled by default but we don't need them right now -- disabling
   # for slightly faster builds. If you need any of these, simply delete a line.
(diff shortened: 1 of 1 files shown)


eundersander (commented):
Look good to me but I'd like to hear @mosra 's thoughts. I'm curious why the Magnum default is shared.

erikwijmans:
Didn't this get changed to shared so that Magnum's python bindings and habitat-sim could both use the same OpenGL context? i.e. to enable things like https://github.com/facebookresearch/habitat-sim/pull/727

Skylion007:
@erikwijmans I don't think so, the OpenGL context is part of Magnum not Corrade, right? Magnum is still statically built as it was before this change (see below to where Magnum), it was only Corrade that not being built statically.

Skylion007:
@erikwijmans In other words see here: https://github.com/facebookresearch/habitat-sim/blob/59b563f1ba451fd27f9d7eca9e49d2bcd69be54c/src/cmake/dependencies.cmake#L175

erikwijmans:
Yeah, on second thought this seems unrelated to that.

src/cmake/dependencies.cmake
```diff
@@ -10,6 +10,10 @@ if(NOT USE_SYSTEM_MAGNUM)
   # These are enabled by default but we don't need them right now -- disabling
   # for slightly faster builds. If you need any of these, simply delete a line.
   set(WITH_INTERCONNECT OFF CACHE BOOL "" FORCE)
+  # Ensure Corrade should be built statically if Magnum is.
+  set(BUILD_PLUGINS_STATIC ON CACHE BOOL "BUILD_PLUGINS_STATIC" FORCE)
+  set(BUILD_STATIC ON CACHE BOOL "BUILD_STATIC" FORCE)
+  set(BUILD_STATIC_PIC ON CACHE BOOL "BUILD_STATIC_PIC" FORCE)
```
mosra, line 16: Hm, can you then remove the same lines that are on line 175 below? Ideally also put a comment there that a static build is being enabled earlier.

It can also be without the docstring, it's useless anyway:

```suggestion
  set(BUILD_PLUGINS_STATIC ON CACHE BOOL "" FORCE)
  set(BUILD_STATIC ON CACHE BOOL "" FORCE)
  set(BUILD_STATIC_PIC ON CACHE BOOL "" FORCE)
```

mosra:
> Previously our corrade plugins were not statically linked [...] 

@Skylion007 This is misleading, *plugins* were always built as static. The bug here was that `BUILD_STATIC` was enabled too late and thus *on the first CMake run* Corrade was still built as dynamic, but because it's a cache variable, re-running CMake again made everything built static.

> Didn't this get changed to shared so that Magnum's python bindings and habitat-sim could both use the same OpenGL context?

@erikwijmans Yeah, it's somewhat related. With Magnum built as dynamic I wouldn't need to do all this [insane symbol deduplication gymnastics](https://github.com/mosra/magnum/blob/dbec10dbee8cc3c2b560c6e7de9166c2b6d2f002/src/Magnum/GL/Context.cpp#L616-L678), but because for this project it doesn't make sense to ship with a bazillion of `*.so`s, I had to do that and it works. Well, except for the bug that plagues #727, and except for GL entrypoints, which I didn't figure out how to deduplicate in a performant-enough way yet. On my list of things to look into when work on the Python viewer is resumed.

> I'm curious why the Magnum default is shared.

@eundersander You're asking why the default isn't static, or what's your question? Shared libraries are the reasonable default that makes sense for most packaging and installation scenarios (which is the implicit use case), less so when a project bundles all its dependencies (which is why Habitat switches Magnum, Assimp, GLFW, Bullet, OpenEXR, ... to build as static).

mosra (approved):

mathfac:
Thank you @Skylion007! Merging to fix the CI. If changes needed, let's make separate PR.

PR #1483, opened 2021-09-15
Fix Double PathFinder

erikwijmans:
## Motivation and Context

Due to legacy reasons, we create a pathfinder object in the python sim _after_ the cpp sim has been initialized. This means that the navmesh loaded via the scene dataset get's unloaded!

## How Has This Been Tested

Via tests. We have a lot of tests that require the navmesh to be loaded correctly and I do not see python saying it loaded the navmesh, so cpp must be doing it correctly!

## Types of changes

### DIFF
--- a/habitat_sim/simulator.py
+++ b/habitat_sim/simulator.py
@@ -31,7 +31,7 @@ from habitat_sim.agent.agent import Agent, AgentConfiguration, AgentState
 from habitat_sim.bindings import cuda_enabled
 from habitat_sim.logging import LoggingContext, logger
 from habitat_sim.metadata import MetadataMediator
-from habitat_sim.nav import GreedyGeodesicFollower, NavMeshSettings, PathFinder
+from habitat_sim.nav import GreedyGeodesicFollower, NavMeshSettings
 from habitat_sim.sensor import SensorSpec, SensorType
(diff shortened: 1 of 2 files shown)


jturner65 (commented):
The only time this will fail is if Sim::reconfigure is called with the same SimulatorConfiguration a it already has.  If close was called before this, the Pathfinder will have been destroyed, but t[his short circuit](https://github.com/facebookresearch/habitat-sim/blob/e51d955eae7aea0ed6406e54d2cf071913c9e2a1/src/esp/sim/Simulator.cpp#L125) will bypass the code to recreate it.

Otherwise, the Pathfinder is always [recreated as part of reconfigure](https://github.com/facebookresearch/habitat-sim/blob/e51d955eae7aea0ed6406e54d2cf071913c9e2a1/src/esp/sim/Simulator.cpp#L210)

Probably, the Pathfinder init should be moved to be before the configuration similarity check, conditional on it being nullptr, like we do with `sceneManager_`

erikwijmans:
Close also resets the config to its default value so that'll be pretty edge-casey, but yeah, I will do the same thing as `sceneManager_` to avoid a potential segfault.

jturner65 (approved):
LGTM!

Skylion007 (approved):
This also has the side effect of simplifying the python code. Nice!

PR #1488, opened 2021-09-17
[PBR][IBL] Soft shadows - Variance Shadow Mapping (VSM)

bigbike:
## Motivation and Context
This diff is to implement the soft shadows using Variance Shadow Mapping (VSM).

The result:
shadows

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/PbrDrawable.h
+++ b/src/esp/gfx/PbrDrawable.h
@@ -5,15 +5,24 @@
 #ifndef ESP_GFX_PBRDRAWABLE_H_
 #define ESP_GFX_PBRDRAWABLE_H_
 
+#include <Corrade/Containers/Optional.h>
+
 #include "esp/gfx/Drawable.h"
 #include "esp/gfx/PbrImageBasedLighting.h"
 #include "esp/gfx/PbrShader.h"
 #include "esp/gfx/ShaderManager.h"
+#include "esp/gfx/ShadowMapManager.h"
 namespace esp {
 namespace gfx {
 
 class PbrDrawable : public Drawable {
  public:
+  struct ShadowData {
+    // all the terms must be set by the user
+    ShadowMapManager* shadowMapManger = nullptr;
+    ShadowMapKeys* shadowMapKeys = nullptr;
+  };
+
   /**
    * @brief Constructor, to create a PbrDrawable for the given object using
    * shader and mesh. Adds drawable to given group and uses provided texture,
@@ -33,6 +42,13 @@ class PbrDrawable : public Drawable {
    *  @param lightSetupKey the key value for the light resource
    */
   void setLightSetup(const Magnum::ResourceKey& lightSetupKey) override;
+
+  /**
+   * @brief Set the shadow map info
+   * @param[in] shadowData, contains all the data needed in the shadow mapping
+   * @param[in] shadowFlag, can only be either ShadowsPCF or ShadowsVSM
+   */
+  void setShadowData(const ShadowData& data, PbrShader::Flag shadowFlag);
 
   static constexpr const char* SHADER_KEY_TEMPLATE = "PBR-lights={}-flags={}";
 
@@ -87,6 +103,7 @@ class PbrDrawable : public Drawable {
   Magnum::Resource<MaterialData, PbrMaterialData> materialData_;
   Magnum::Resource<LightSetup> lightSetup_;
   PbrImageBasedLighting* pbrIbl_ = nullptr;
+  Corrade::Containers::Optional<ShadowData> shadowData_;
 };
 
 }  // namespace gfx
--- a/src/esp/gfx/PbrShader.cpp
+++ b/src/esp/gfx/PbrShader.cpp
@@ -105,6 +105,7 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
       "#define OUTPUT_ATTRIBUTE_LOCATION_OBJECT_ID {}\n", ObjectIdOutput);
 
   frag.addSource(outputAttributeLocationsStream.str())
+      .addSource(flags_ & Flag::ShadowsVSM ? "#define SHADOWS_VSM\n" : "")
       .addSource(isTextured ? "#define TEXTURED\n" : "")
       .addSource(flags_ & Flag::BaseColorTexture ? "#define BASECOLOR_TEXTURE\n"
                                                  : "")
@@ -130,6 +131,8 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
                                              : "")
       .addSource(
           Cr::Utility::formatString("#define LIGHT_COUNT {}\n", lightCount_))
+      .addSource(flags_ & Flag::ShadowsVSM ? rs.get("shadowsVSM.glsl") + "\n"
+                                           : "")
       .addSource(rs.get("pbrCommon.glsl") + "\n")
       .addSource(rs.get("pbr.frag"));
 
@@ -193,6 +196,16 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
                pbrTextureUnitSpace::TextureUnit::BrdfLUT);
     setUniform(uniformLocation("PrefilteredMap"),
                pbrTextureUnitSpace::TextureUnit::PrefilteredMap);
+  }
+
+  // VSM shadows
+  if (flags_ & Flag::ShadowsVSM) {
+    setUniform(uniformLocation("ShadowMap[0]"),
+               pbrTextureUnitSpace::TextureUnit::ShadowMap0);
+    setUniform(uniformLocation("ShadowMap[1]"),
+               pbrTextureUnitSpace::TextureUnit::ShadowMap1);
+    setUniform(uniformLocation("ShadowMap[2]"),
+               pbrTextureUnitSpace::TextureUnit::ShadowMap2);
   }
 
   // cache the uniform locations
@@ -356,6 +369,33 @@ PbrShader& PbrShader::bindPrefilteredMap(Magnum::GL::CubeMapTexture& texture) {
                  "created with image based lighting enabled",
                  *this);
   texture.bind(pbrTextureUnitSpace::TextureUnit::PrefilteredMap);
+  return *this;
+}
+
+PbrShader& PbrShader::bindPointShadowMap(int index,
+                                         Magnum::GL::CubeMapTexture& texture) {
+  CORRADE_ASSERT(
+      index >= 0 && index < 3,
+      "PbrShader::bindPointShadowMap(): the texture index was illegal.", *this);
+  CORRADE_ASSERT(flags_ & Flag::ShadowsVSM,
+                 "PbrShader::bindPointShadowMap(): the shader was not "
+                 "created with shadows enabled",
+                 *this);
+  switch (index) {
+    case 0:
+      texture.bind(pbrTextureUnitSpace::TextureUnit::ShadowMap0);
+      break;
+    case 1:
+      texture.bind(pbrTextureUnitSpace::TextureUnit::ShadowMap1);
+      break;
+    case 2:
+      texture.bind(pbrTextureUnitSpace::TextureUnit::ShadowMap2);
+      break;
+
+    default:
+      CORRADE_INTERNAL_ASSERT_UNREACHABLE();
+      break;
+  }
   return *this;
 }
 
--- a/src/esp/gfx/VarianceShadowMapShader.cpp
+++ b/src/esp/gfx/VarianceShadowMapShader.cpp
@@ -1,0 +1,101 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "VarianceShadowMapShader.h"
+
+#include <Corrade/Containers/ArrayView.h>
+#include <Corrade/Containers/Reference.h>
+#include <Corrade/Utility/Debug.h>
+#include <Corrade/Utility/DebugStl.h>
+#include <Corrade/Utility/FormatStl.h>
+#include <Corrade/Utility/Resource.h>
+#include <Magnum/GL/Context.h>
+#include <Magnum/GL/Extensions.h>
+#include <Magnum/GL/Shader.h>
+#include <Magnum/GL/Version.h>
+#include <Magnum/Magnum.h>
+#include <Magnum/Math/Matrix4.h>
+
+#include <initializer_list>
+#include <sstream>
+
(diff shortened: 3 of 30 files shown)


src/esp/sim/Simulator.cpp
```diff
@@ -583,6 +592,140 @@ void Simulator::reconfigureReplayManager(bool enableGfxReplaySave) {
...
+  scene::SceneGraph& sg = getActiveSceneGraph();
+  auto& shadowManager = resourceManager_->getShadowMapManger();
+  auto& shadowMapKeys = resourceManager_->getShadowMapKeys();
+  std::vector<Mn::ResourceKey>& keys = shadowMapKeys[activeSceneID_];
+
+  // XXXXXXXXXXXX
+  // I need John's help. How to get the light positions here??!!
+  // currently the light position is hard coded for a particular replicaCAD
+  // scene
+  Mn::Vector3 lightPos[3] = {Mn::Vector3{2.77, 2.77, 5.16},
```
bigbike, line 633: @jturner65 : John, I need your help here. 

I am going to compute the shadow maps for the lights (1 map per light), and I would like to obtain the light absolute position specified in the `scene_dataset` config.

As you can see here, I hard-coded 3 light positions here to do the demo. But this is definitely wrong for a new scene.
I need a generic method to query this value (light position) within the Simulator class.

As you can see, currently this function `computeShadowMaps` is called in the viewer after the `Simulator` instance is created. So I assume `scene_dataset` config has been parsed and the light positions are already stored somewhere in some data structure. I just need a way to get the data.
Let me know how I can do it. Thanks!

dhruvbatra:
That's a beautiful rendering. Nicely done @bigbike!

src/esp/sim/Simulator.cpp
```diff
@@ -39,6 +42,12 @@ using metadata::attributes::SceneAOInstanceAttributes;
 using metadata::attributes::SceneObjectInstanceAttributes;
 using metadata::attributes::StageAttributes;
 
+namespace {
+const char* shadowMapDrawableGroupName = "static-shadow-map";
```
Skylion007, line 46: Shouldn't these be constexpr too?

src/shaders/shadowsVSM.glsl
```diff
@@ -0,0 +1,51 @@
...
+float computeShadowVSM(int idx, vec3 fragPos, vec3 lightPos) {
+  if (idx >= 3) {
+    return 1.0;  // at most 3 shadow map are supported
+  }
+  vec3 lightToFrag = fragPos - lightPos;
+  float d = length(lightToFrag);
+
+  vec2 moments = texture(ShadowMap[idx], normalize(lightToFrag)).xy;
+    // moments.x is the mean value while moments.y equals to depth * depth
+    return computeShadowUpperBound(moments, d);
```
Skylion007, line 38: Bad indentation

src/esp/sim/Simulator.cpp
jturner65: Ok, to get the LightSetup that is currently in use you can use the same mechanism by which it is created (which will is performed when Scene is instantiated).

In c++ Simulator class, config_.sceneLightSetup will always hold the key of the current scene's LightSetup
Use the following function to acquire a copy of the LightSetup for the current scene, built from the attributes  : 
```
 esp::gfx::LightSetup lightingSetup =
          metadataMediator_->getLightLayoutAttributesManager()
              ->createLightSetupFromAttributes(config_.sceneLightSetup);
```

src/esp/sim/Simulator.cpp
```diff
@@ -589,6 +600,156 @@ void Simulator::reconfigureReplayManager(bool enableGfxReplaySave) {
...
+    }
+
+    // for VSM only !!!
+    renderer_->applyGaussianFiltering(
+        *pointShadowMap, *helperShadowMap,
+        gfx::CubeMap::TextureType::VarianceShadowMap);
+  }
+}
+
+void Simulator::setShadowMapsToDrawables() {
```
jturner65, line 724: Would it be feasible and possibly appropriate to put some or all of these 3 functions that directly access the ShadowMap constructs (updateShadowMapDrawableGroup, computeShadowMaps, setShadowMapsToDrawables) in ResourceManager, so we do not have to provide public access to these constructs from outside RM?  We could pass in refs to current scenegraph and renderer (where appropriate). 

Just a thought :)

jturner65 (approved):
LGTM.  I made a few comments, but feel free to disregard :)

src/esp/sim/Simulator.cpp
bigbike: Yeah, I see your point.
My opinion: `Simulator` class is like a driver to glue the other library modules.`Shadows` are in the "resource manager". But "Drawables" belong to the `SceneGraph` which is in another module. `setShadowMapsToDrawables`, as well as the other 2 functions, will work on these 2 different modules. So I put them in the `Simulator` to avoid complicated parameter passing.

Besides `Simulator` will have to provide such APIs for the external users (e.g., `viewer`) to construct shadow maps.

src/esp/sim/Simulator.cpp
```diff
@@ -39,6 +42,14 @@ using metadata::attributes::SceneAOInstanceAttributes;
 using metadata::attributes::SceneObjectInstanceAttributes;
 using metadata::attributes::StageAttributes;
 
+namespace {
+constexpr const char* shadowMapDrawableGroupName = "static-shadow-map";
+constexpr const char* defaultRenderingGroupName = "";
+const int shadowMapSize = 1024;
+const int maxNumShadowMaps = 3;  // the max number of point shadow maps
```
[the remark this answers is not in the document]
bigbike: Yeah, I admit we did not strictly follow every item in the FB C++ guide.
For this one, adding `k` or `g` to the variables will make things complicated. My opinion is to avoid it. :)

bigbike:
@mosra : Would you mind taking a look when you are available? Thank you!

src/esp/sim/Simulator.cpp
jturner65: Fair enough.Thanks for responding :)

src/shaders/gaussianFilter.frag
```diff
@@ -0,0 +1,80 @@
...
+      // y axis
+      result +=
+          texture(SourceTexture, textureCoordinates + vec2(0.0, scale * i))
+              .rgb *
+          weight[i];
+      result +=
+          texture(SourceTexture, textureCoordinates - vec2(0.0, scale * i))
+              .rgb *
+          weight[i];
+    }
```
mosra, line 59: Instead of branching on `bool FilterHorizontally`, what about supplying `vec2 filterDirection` which is `vec2(1.0, 0.0)` for X and `vec2(0.0, 1.0)` for Y? Then the above would be just this:

```suggestion
    result +=
        texture(SourceTexture, textureCoordinates + filterDirection*scale*i)
            .rgb *
        weight[i];
    result +=
        texture(SourceTexture, textureCoordinates - filterDirection*scale*i)
            .rgb *
        weight[i];
```

src/shaders/gaussianFilter.frag
```diff
@@ -0,0 +1,80 @@
...
+layout(location = OUTPUT_ATTRIBUTE_LOCATION_COLOR) out highp vec4 fragmentColor;
+
+//------------- shader ----------------------
+//
+void main(void) {
+  ivec2 dims = textureSize(SourceTexture, 0);  // lod = 0
+  float scale = 1.0 / dims.x;
+  if (!FilterHorizontally) {
+    scale = 1.0 / dims.y;
+  }
```
mosra, line 27: Continuing from the previous suggestion, with `vec2 filterDirection` this would be 

```suggestion
  vec2 scale2 = filterDirection/dims;
  float scale = scale2.x + scale2.y; // or max(), one of those is always 0
```

mosra: Eh, actually, instead of using `texture()` and calculating `scale`, you should be using `texelFetch()` that takes integer coordinates. Since that skips filtering, it could potentially be also faster. A "level up" from there is `texelFetchOffset()`, since you have a compile-time offset here it's a great match. 

Not sure about NV in particular, probably won't affect perf much, but the reason for the `*Offset()` variant was to allow the hardware to predict which area of the texture will be accessed by a shader invocation so it can prefetch the memory. I guess it could have some non-zero impact on mobile hardware (VR, etc).

- - -

Possible further optimization idea -- when I was implementing gaussian blur about a decade ago for a silly game project, I got the best perf by [unrolling the loop by hand](https://github.com/nuclearwhales/push-the-box/blob/248540eec4b052dc210716647994e364d3672c33/src/Shaders/Blur.frag#L20-L37). Because the thing had to run on NaCl (a predecessor to WebGL 1), I couldn't use `texelFetch()` nor `texelFetchOffset()` and had to instead pass the texture coordinates from the vertex shader to avoid the HW use slow dependent texture reads. Since you can use `texelFetch()`, that part makes no sense anymore, but I still think unrolling may help ... or maybe `texelFetchOffset()` will force you to do that anyway, since it expects the offset to be a constant expression.

src/esp/gfx/PbrDrawable.h
```diff
@@ -5,15 +5,24 @@
...
 namespace esp {
 namespace gfx {
 
 class PbrDrawable : public Drawable {
  public:
+  struct ShadowData {
+    // all the terms must be set by the user
+    ShadowMapManager* shadowMapManger = nullptr;
+    ShadowMapKeys* shadowMapKeys = nullptr;
+  };
```
mosra, line 24: Not sure about the usefulness / usability of this struct -- it has pointers on which you have to assert for non-null and then you only use it in a single function, `setShadowData()`, *and then* you save it into an optional, which adds another level of potential null-ness on which the code has to branch.

Since the two members are of a different type, this is not the case where one could mix up the parameters, so why not directly have `setShadowData(ShadowMapManager&, ShadowMapKeys&, PbrShader::Flag)`? Easier for the user to supply as there's no need to fill a temporary struct. You can then save those to the class as two separate pointers, no need for a struct there either -- it's internal state of the class, so the public interface already ensures that those are either both set, or when `setShadowData()` wasn't called, then they are both null.

src/esp/gfx/PbrShader.cpp
```diff
@@ -359,6 +372,33 @@ PbrShader& PbrShader::bindPrefilteredMap(Magnum::GL::CubeMapTexture& texture) {
...
+      texture.bind(pbrTextureUnitSpace::TextureUnit::ShadowMap1);
+      break;
+    case 2:
+      texture.bind(pbrTextureUnitSpace::TextureUnit::ShadowMap2);
+      break;
+
+    default:
+      CORRADE_INTERNAL_ASSERT_UNREACHABLE();
+      break;
+  }
```
mosra, line 398:
```suggestion
  texture.bind(pbrTextureUnitSpace::TextureUnit::ShadowMap0 + index);
```

The assertion above does enough work to ensure this is always correct ;)

src/esp/gfx/VarianceShadowMapShader.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  Mn::GL::Shader vert{glVersion, Mn::GL::Shader::Type::Vertex};
+  Mn::GL::Shader frag{glVersion, Mn::GL::Shader::Type::Fragment};
+
+  std::stringstream attributeLocationsStream;
+  attributeLocationsStream << Cr::Utility::formatString(
+      "#define ATTRIBUTE_LOCATION_POSITION {}\n", Position::Location);
+
+  // Add macros
+  vert.addSource(attributeLocationsStream.str())
```
mosra, line 59: Come on, not this copypasted nonsense again :see_no_evil: 

```suggestion
  vert.addSource(Cr::Utility::formatString(
      "#define ATTRIBUTE_LOCATION_POSITION {}\n", Position::Location))
```

Same below. 

Also, can you please go over the other shaders and remove the remaining cases so it's never wrongly reused again? Somehow it always gets brought back no matter how many times we fix it :laughing:

src/esp/gfx/VarianceShadowMapShader.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  // bind attributes
+#ifndef MAGNUM_TARGET_GLES
+  if (!Mn::GL::Context::current()
+           .isExtensionSupported<
+               Mn::GL::Extensions::ARB::explicit_attrib_location>(glVersion))
+#endif
+  {
+    bindAttributeLocation(Position::Location, "vertexPosition");
+  }  // if
```
mosra, line 82: This code is also not needed at all -- delete. The shader GLSL has the attribute location hardcoded *always*.

```suggestion
```

(Same in this case, can you go over the remaining shaders in habitat and clean this up there too?)

mosra (commented):
Cool stuff you have here! :fire:

src/esp/gfx/PbrDrawable.h
bigbike: Ha, it totally makes sense. Let me try your suggestion.
You know what, the reason I had this struct here is that it is used to contain more values than its current shape. e.g., `lightNearPlane`, `lightFarPlane`. 
I removed them in this diff because I abandoned `traditional shadow cube map + Percentage Closer Filtering (PCF)` -- another way to render shadows. I use VSM here.

bigbike:
@mosra : Take a look when you have chance. Thanks!

src/esp/sim/Simulator.cpp
```diff
@@ -40,6 +43,14 @@ using metadata::attributes::SceneAOInstanceAttributes;
...
```
mosra, line 50:
```suggestion
constexpr int shadowMapSize = 1024;
constexpr int maxNumShadowMaps = 3;  // the max number of point shadow maps
```

Re naming, just FYI -- in Magnum I'm having constants with an uppercase first letter (treating them the same as enums, basically). 

Not a fan of `k` or `g` either :)

src/esp/gfx/VarianceShadowMapShader.cpp
```diff
@@ -0,0 +1,89 @@
...
+  // Add macros
+  vert
+      .addSource(Cr::Utility::formatString(
+          "#define ATTRIBUTE_LOCATION_POSITION {}\n", Position::Location))
+      .addSource(rs.get("varianceShadowMap.vert"));
+
+  std::stringstream outputAttributeLocationsStream;
+  outputAttributeLocationsStream << Cr::Utility::formatString(
+      "#define OUTPUT_ATTRIBUTE_LOCATION_COLOR {}\n", ColorOutput);
+  frag.addSource(outputAttributeLocationsStream.str())
```
mosra, line 63: I see another stringstream crime here :P

mosra (approved):
One last thing, other than that -- looks good to me :)

PR #1499, opened 2021-09-24
Migrate ci resource type in advance of deprecation

aclegg3:
## Motivation and Context

CircleCI is deprecating legacy resource types (e.g. `gpu.small`) by 2022. We should migrate to the new resource types.

## How Has This Been Tested

CI jobs running with new resource type.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 gpu: &gpu
   machine:
     image: ubuntu-1604:202104-01
-  resource_class: gpu.small
+  resource_class: gpu.nvidia.medium
   environment:
     FPS_THRESHOLD: 900
 commands:


Skylion007 (approved):

aclegg3:
Looks like gpu instances are not allocating for our ci pricing plan. Do we need to upgrade to a "scale" plan to get GPU allocation?

erikwijmans:
I don't think we have access to multi-gpu (the `.mutli`). Does `gpu.nvidia.small` work?

erikwijmans:
We also may OOM on `nvidia.gpu.small`, as that only has 8 GB vram (compared to the 16 GB on `gpu.small`). May be worth doing `nvidia.gpu.medium`.

aclegg3:
> I don't think we have access to multi-gpu (the `.mutli`). Does `gpu.nvidia.small` work?

Perhaps. Maybe this paradigm is shifting also. I'm looking at [this pricing page](https://circleci.com/pricing/) for reference.

erikwijmans:
I think multi-gpu means multiple GPUs on one VM, not multiple 1 GPU VMs (we've been using the later, not the former)

aclegg3:
> This PR is downgrading Magnum and other submodules isn't it?

Yeah, working on it. Accidentally slipped in with rebase somehow.

aclegg3:
Easier to start a new PR at this point.

PR #1507, opened 2021-09-30
Fix semantic rendering

erikwijmans:
## Motivation and Context

Semantic rendering requires an activeSemenaticSceneGraph, not a semenaticScene. The former is the Magnum SceneGraph used for rendering while the latter contains information about semantic annotations. We are currently checking for the latter while needing the former.

## How Has This Been Tested

Via the new test.

## Types of changes


aclegg3 (approved):
Looks good. Thanks for fixing this typo and adding a test.

Skylion007 (approved):

jturner65 (approved):
LGTM

PR #1509, opened 2021-09-30
[WIP]--Support config ShaderType specification when creating materials

jturner65:
## Motivation and Context
This PR supports consuming the user-specified ShaderType value from config sources to override the shaderType/material type chosen based on the loaded material.
This is still a WIP - SimTest is failing due to improvements/fixes to the default material handling (such as accurately supporting flat materials).

## Types of changes

## Checklist


src/esp/assets/ResourceManager.h
```diff
@@ -807,6 +818,54 @@ class ResourceManager {
...
+   * @param materialData The material whose type we are verifying against
+   * @param verificationType The ShaderTypeEnum we are verifying against
+   * @param mnVerificationType The @ref Mn::Trade::MaterialType bitflag that the
+   * passed material's specified type is being verified against.
+   * @return Whether or not the passed @p typeToCheck matches the passed
+   * criteria.
+   */
+  bool checkForPassedShaderType(
+      const ShaderTypeEnum typeToCheck,
+      const Cr::Containers::Optional<Mn::Trade::MaterialData>& materialData,
```
mosra, line 846: Why pass an argument through an `Optional` if you don't allow / don't expect `NullOpt` values? In that case I'd take just a `const Mn::Trade::MaterialData&`. The caller then passes `*foo` instead of `foo` (and ensures it's not a null optional, of course).

This has an equivalent semantics to a reference vs a pointer -- if the API allows `nullptr` values, a pointer is fine, otherwise it should be a reference, and the responsibility of ensuring non-nullness lies then on the caller.

I see the same being done for a lot more functions (`buildFlatShadedMaterialData()`, ...), since they were taking a const reference before I don't think the Optional has any use there either.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1648,43 +1751,257 @@ void ResourceManager::loadMaterials(Importer& importer,
...
+  const auto flags = Mn::Trade::MaterialType::Flat |
+                     Mn::Trade::MaterialType::Phong |
+                     Mn::Trade::MaterialType::PbrMetallicRoughness;
+
+  // create new material from attributes array
+  ESP_WARNING() << "before building new material";
+  Mn::Trade::MaterialData newMaterialData{flags, std::move(newAttributes)};
+  ESP_WARNING() << "after building new material";
+  // move into optional container and return
+  return Cr::Containers::Optional<Mn::Trade::MaterialData>{
```
mosra, line 1951: Same here, why wrap the return value in an optional when it's never going to be null? Designing the API this way implies that the caller should except the operation to fail and check if the return value is valid, but that never actually happens.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1809,7 +2128,84 @@ gfx::PbrMaterialData::uptr ResourceManager::buildPbrShadedMaterialData(
...
+//   finalMaterial->specularColor *= 0.15;
+
+//   // normal mapping
+//   if (material.hasAttribute(MnMatAttr::NormalTexture)) {
+//     finalMaterial->normalTexture =
+//         textures_.at(textureBaseIndex + material.normalTexture()).get();
+//   }
+
+//   return finalMaterial;
+// }  // ResourceManager::buildPhongFromPbrMetallicRoughness
```
mosra, line 2208: I don't think we need to keep unused code around, there's Git history for that ;) Same elsewhere.

src/esp/assets/ResourceManager.h
```diff
@@ -84,6 +84,8 @@ class Model;
 }
 }  // namespace io
 namespace assets {
+// used for shadertype specification
+typedef metadata::attributes::ObjectInstanceShaderType ShaderTypeEnum;
```
mosra, line 88: Do we really need the alias? it's named completely differently so it bears no resemblance to the type it aliases, there's a superfluous `Enum` suffix...

I feel we suffer from namespace overuse a bit, but introducing the same enum under a completely different name is only going to make it worse, not better in my opinion. Maybe just an `using` then?

```suggestion
using metadata::attributes::ObjectInstanceShaderType;
```

src/esp/assets/ResourceManager.cpp
```diff
@@ -1591,8 +1616,76 @@ int ResourceManager::loadNavMeshVisualization(esp::nav::PathFinder& pathFinder,
...
+ * @param material Source material for new attributes array
+ * @param newAttributes The array of attributes to be used to build a new
+ * Mn::Trade::MaterialData.
+ * @param matAttr The Mn::Trade::MaterialAttribute tag describing the attribute
+ * to add
+ * @param value A reference to the value to add.
+ */
+template <typename T>
+void safeAppend(
+    const Cr::Containers::Optional<Mn::Trade::MaterialData>& material,
```
mosra, line 1634: Again the optional here. Please don't :)

Same with `compareShaderTypeToMnMatType()`.

src/esp/assets/ResourceManager.cpp
```diff
@@ -83,6 +84,11 @@ using metadata::managers::StageAttributesManager;
 
 namespace assets {
 
+// used for shadertype specification
+typedef metadata::attributes::ObjectInstanceShaderType ShaderTypeEnum;
+// used to simplify access to magnum materialAttributes
+typedef Mn::Trade::MaterialAttribute MnMatAttr;
```
mosra, line 90: Same here, can we _please_ stop making aliases that are a *completely different name*? :pleading_face: Doing `using Mn::Trade::MaterialAttribute;` would do the job as well and won't be conflicting with anything if it stays in the scope of the cpp file (or, **ideally**, in just the one or two functions that use it, since the file is several thousand lines long).

And to be frank, except for the `Mn` / `np` / `Cr` that already went canon, I don't like the word shortening happening at various places, at all, sorry. While it probably makes sense at the point of writing the code, a month from now you'd half-forget and have to double check what it means, and when someone else comes to the code they'd have no idea at all. Most IDEs and editors can autocomplete (if I write `MatA` and hit <kbd>Ctrl</kbd> <kbd>Space</kbd> it'll fill in the rest) and given that the code is written once and then read several times, the shortenings areonly barely helping you type it out while hurting readability a lot.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1648,43 +1751,257 @@ void ResourceManager::loadMaterials(Importer& importer,
...
+    const float metalness =
+        pbrMaterial.hasMetalnessTexture() ? 0.0f : pbrMaterial.metalness();
+
+    // Heuristic to map roughness to spec power.
+    // Higher exponent makes the spec highlight larger (lower power)
+    // example calc :
+    // https://www.wolframalpha.com/input/?i=1.1+%2B+%281+-+x%29%5E4.5+*+180.0+for+x+from+0+to+1
+    // lower power for metal
+    const float maxShininess = Mn::Math::lerp(250.0f, 120.0f, metalness);
+    const float shininess = 1.1f + powf(1.0f - roughness, 4.5) * maxShininess;
```
mosra, line 1811: I know this code is mostly a copypaste of the original, but while we're at it, let's use the guaranteed-to-also-have-a-float-overload API instead:

```suggestion
    const float shininess = 1.1f + Mn::Math::pow(1.0f - roughness, 4.5f) * maxShininess;
```

Same in other cases where `powf` is used, note the `f` literal used for the power factor.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1648,43 +1751,257 @@ void ResourceManager::loadMaterials(Importer& importer,
...
+    // Experimental metalness heuristic using saturation of spec color
+    // to derive approximation of metalness
+    const Mn::Color4 specColor = phongMaterial.specularColor();
+
+    // if specColor alpha == 0 then no metalness
+    float metalness = 0.0f;
+    // otherwise, this heuristic will derive a value for metalness based on
+    // how non-grayscale the specular color is (HSV Saturation).
+    if (specColor.a() != 0.0f) {
+      metalness = specColor.toHsv().saturation;
```
mosra, line 1912:
```suggestion
      metalness = specColor.saturation();
```

avoids calculating the (to be thrown away) H and V values :)

src/esp/assets/ResourceManager.cpp
```diff
@@ -1591,8 +1616,76 @@ int ResourceManager::loadNavMeshVisualization(esp::nav::PathFinder& pathFinder,
...
+ * @tparam the type of the value to be added
+ * @param material Source material for new attributes array
+ * @param newAttributes The array of attributes to be used to build a new
+ * Mn::Trade::MaterialData.
+ * @param matAttr The Mn::Trade::MaterialAttribute tag describing the attribute
+ * to add
+ * @param value A reference to the value to add.
+ */
+template <typename T>
+void safeAppend(
```
mosra, line 1633: Instead of `safeAppend()`, which doesn't really say *what* is safe about this function, what about `appendIfNotPresent()`?

src/esp/assets/ResourceManager.cpp
```diff
@@ -1591,8 +1616,76 @@ int ResourceManager::loadNavMeshVisualization(esp::nav::PathFinder& pathFinder,
...
+ * congruient with the specified shader type.
+ */
+bool compareShaderTypeToMnMatType(
+    const ShaderTypeEnum typeToCheck,
+    const Cr::Containers::Optional<Mn::Trade::MaterialData>& materialData) {
+  switch (typeToCheck) {
+    case ShaderTypeEnum::Phong: {
+      bool compRes =
+          ((materialData->types() & Mn::Trade::MaterialType::Phong) ==
+           Mn::Trade::MaterialType::Phong);
```
mosra, line 1660: FYI the EnumSet class has a [`operator>=`](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html#aeb270daa9a8afb3a80e57ef2cb2f8314) which resembles a ⊇ set operation in mathematics, which would shorten the above to

```suggestion
          materialData->types() >= Mn::Trade::MaterialType::Phong);
```

It might be however confusing to some people because this does a different thing than it would do with pure integers. The `MaterialType` enum is just single bits so even

```suggestion
          !!(materialData->types() & Mn::Trade::MaterialType::Phong);
```

would work for it right now (the !! for forcing an explicit bool conversion), alternatively

```suggestion
          bool(materialData->types() & Mn::Trade::MaterialType::Phong);
```

if you don't like the excessive yelling :)

Same below.

mosra (commented):
Just a WIP review also, will take a closer look at the actual logic once you're finished.

Is there anything in this PR that would benefit from me implementing mutable access to material values? I didn't see anything obvious, but doesn't hurt to ask I guess :)

src/esp/assets/ResourceManager.h
jturner65: Thanks for checking things out @mosra, as always, and for pointing out these inefficiencies.  In general, when I'm refactoring existing code I try to keep my changes as minimal as possible, so I may propagate design choices that were sub-optimal.  I'll address them as I improve this WIP.

src/esp/assets/ResourceManager.cpp
jturner65: Temporary only - wanted to provide the original code to make it easy for reviewers to check for transcription errors I might have missed.

src/esp/assets/ResourceManager.h
jturner65: That's cool.

src/esp/assets/ResourceManager.cpp
jturner65: We get an optional from the importer, are you saying deref it for these calls?

src/esp/assets/ResourceManager.cpp
jturner65: wow. the enum op overload is really cool, ty.

src/esp/assets/ResourceManager.cpp
jturner65: tried to get all the literals to be floats, guess I missed a few.  thanks! :)

src/esp/assets/ResourceManager.cpp
jturner65: Fair enough -  I'm not attached to this mechanism.  

I was mainly thinking about future maintainers (I didn't think MnMatAttr was that different from "Magnum Material Attributes" but I won't argue that point :) ) being able to find things (since ResourceManager touches on -so many things-.)

src/esp/assets/ResourceManager.cpp
mosra: Yes, the intended design / workflow is basically like this:

- the importer returns an Optional in order to signal if importing failed for some reason (for example when the glTF material was referencing a texture ID that's out of bounds)
- to avoid crazy complex error handling, not all importer APIs return an Optional, for example the various `fooCount()` and `fooName()` always return a valid value -- but when an API returns an Optional you're always expected to perform error handling
- when interacting with the importer, the best way is to immediately check the returned Optional and handle the error (for example -- replacing a material with a placeholder one, for a mesh just not rendering anything, for a missing image render as if it would be just colored...)
- and then, if not null, you dereference the Optional and then pass around just the underlying type

jturner65:
> Is there anything in this PR that would benefit from me implementing mutable access to material values? I didn't see anything obvious, but doesn't hurt to ask I guess :)

Not yet.  Would be useful in the "gfx::MaterialData -> Mn::trade::MaterialData" refactor coming in PR #3 of 3 though.

PR #1515, opened 2021-10-04
[Bugfix]--address segfaults by catching nullptr attribs and exiting gracefully

jturner65:
## Motivation and Context
This PR addresses segfaults that would happen if, due to misspelling or other errors, a required or expected attributes necessary for scene building is not found (i.e. queries to MetadataMediator returned nullptr).  Now, should these queries fail, a salient message is displayed and the program ends via std::abort.

## How Has This Been Tested
C++ and python tests pass locally

## Types of changes

## Checklist

### DIFF
--- a/src/esp/sim/Simulator.cpp
+++ b/src/esp/sim/Simulator.cpp
@@ -217,6 +217,14 @@ Simulator::setSceneInstanceAttributes(const std::string& activeSceneName) {
 
   metadata::attributes::SceneAttributes::cptr curSceneInstanceAttributes =
       metadataMediator_->getSceneAttributesByName(activeSceneName);
+  // check if attributes is null - should not happen
+  CORRADE_ASSERT(
+      curSceneInstanceAttributes,
+      Cr::Utility::formatString(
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

src/esp/sim/Simulator.cpp
```diff
@@ -217,6 +217,14 @@ Simulator::setSceneInstanceAttributes(const std::string& activeSceneName) {
 
   metadata::attributes::SceneAttributes::cptr curSceneInstanceAttributes =
       metadataMediator_->getSceneAttributesByName(activeSceneName);
+  // check if attributes is null - should not happen
+  CORRADE_ASSERT(
```
eundersander, line 221: Is it possible to hit any of these by calling from python into Habitat-sim in a certain way, or by calling into Habitat-sim with certain files on the filesystem missing?

If yes, this feels more like an `ESP_CHECK`.

jturner65: yes, great point, thanks @eundersander

PR #1516, opened 2021-10-04
[CI] Fix the CI!

erikwijmans:
## Motivation and Context

For some reason our CI started having certificate issues with cmake. AFAIK this is due to Ubuntu 16.04 being EOL. This PR moves the CI to 20.04.

There was one change needed to make things run on 20.04, which was make the method we call in the Simulator constructor not virtual (as clang-tidy rightfully complains about that being a can of worms). We currently don't overload those anywhere expect in python, but python doesn't need them to be marked as virtual.

## How Has This Been Tested

With the CI


Skylion007:
The some reason is that Let's Encrypt old Root CA expired this weekend: https://portswigger.net/daily-swig/lets-encrypt-root-cert-update-catches-out-many-big-name-tech-firms

Skylion007 (approved):

erikwijmans:
Interesting. I wonder why that only seems to matter for the ubuntu 16.04 image.

tests/test_physics.py
```diff
@@ -1710,7 +1710,7 @@ def test_rigid_constraints():
             habitat_sim.physics.RigidConstraintType.PointToPoint
         )
         constraint_id_2 = sim.create_rigid_constraint(constraint_settings_2)
-        observations += simulate(sim, 5.0, produce_debug_video)
+        observations += simulate(sim, 7.0, produce_debug_video)
```
erikwijmans, line 1713: @aclegg3 Due to compiler changes the assert on L1719 was failing. Cranking up situation time seemed to fix that.

Skylion007:
Because the distro is so old, the Root CAs were never updated.

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 gpu: &gpu
   machine:
-    image: ubuntu-1604:201903-01
+    image: ubuntu-2004:202107-02
```
Skylion007, line 4:
```suggestion
    image: ubuntu-2004-cuda-11.2:202103-01
```
This has CUDA 11.2 pre-installed which should simplify the installation if that is accessible. I don't know if a Ubuntu-2004 has a cuda 11.1 preinstalled: https://circleci.com/docs/2.0/configuration-reference/

Skylion007:
LGTM. Let's merge this as it would be great if we can restore the CI. 👍

Skylion007:
@erikwijmans I found a lot less invasive fix by doing `apt-get update/ apt-get upgrade` at the beginning of the circleci file and tested in the fix_rootca_quickfix branch.

.circleci/config.yml
erikwijmans: I guess cuda-11.2 may be fine here, but in general cuda-11.2 is avoided since its a buggy slow mess.

Skylion007:
@erikwijmans Alternative quickfix: https://github.com/facebookresearch/habitat-sim/pull/1518

aclegg3:
Covered Ubuntu upgrade in #1665.

PR #1520, opened 2021-10-06
Fixing macOS linking bug

dhruvbatra:
## Motivation and Context

Build from source on macOS fails at the linking step:
"ld: library not found for -lminizip"

Searching online the problem seems to be homebrew's cmake since Mojave:
https://stackoverflow.com/questions/54068035/linking-not-working-in-homebrews-cmake-since-mojave

## How Has This Been Tested

By localing building on my mac. 

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ x] My code follows the code style of this project.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -206,3 +206,8 @@ if(BUILD_GUI_VIEWERS)
   message("Building GUI viewer")
   add_subdirectory(utils/viewer)
 endif()
+
+IF(APPLE)
+# Fix linking on 10.14+. See https://stackoverflow.com/questions/54068035
+LINK_DIRECTORIES(/usr/local/lib)
+ENDIF()


mosra:
Yep, it's Assimp :roll_eyes: The following patch should tell it to not look for `minizip` on the system but instead compile a bundled copy, which is what we want:

```patch
diff --git a/src/cmake/dependencies.cmake b/src/cmake/dependencies.cmake
index 8e5bad96..ff97b3e0 100644
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -69,6 +69,11 @@ if(BUILD_ASSIMP_SUPPORT)
     # prefixed with ASSIMP_, so better set both variants to future-proof this.
     set(INJECT_DEBUG_POSTFIX OFF CACHE BOOL "" FORCE)
     set(ASSIMP_INJECT_DEBUG_POSTFIX OFF CACHE BOOL "" FORCE)
+    # Otherwise Assimp may attempt to find minizip on the filesystem using
+    # pkgconfig, but in a poor way that expects just yelling -lminizip at the
+    # linker without any link directories would work. It won't. (The variable
+    # is not an option() so no need to CACHE it.)
+    set(ASSIMP_BUILD_MINIZIP ON)
     add_subdirectory("${DEPS_DIR}/assimp")
 
     # Help FindAssimp locate everything

```

Funnily enough it doesn't seem to actually use it for anything (or at least that's what GitHub search tells me). I don't have any easy way to reproduce the original problem, can you check if this fixes it and the `link_directories()` call is no longer needed?

dhruvbatra:
Yes, confirming that this works. Thanks @erikwijmans and @mosra.

PR #1541, opened 2021-10-20
Conda mutex

### LINKED ISSUE #1539: install doesn't currently work with mamba
This is an fyi issue that currently if one uses the `mamba` package manager, the `headless` or `withbullet` options are not respected.

I've filed an issue with the `mamba` team: https://github.com/mamba-org/mamba/issues/1223

erikwijmans:
## Motivation and Context

We should use a mutex package instead of track features to support mamba.

If this looks good ill upload the mutex packages and updated feature packages to aihabitat and aihabitat-nightly (the new feature packages are backwards compatible)

Closes #1539 

## How Has This Been Tested

On MacOS, the following all get the right package
```
mamba/conda create -n hsim-mutex-test python=3.7 habitat-sim -c conda-forge -c erikwijmans
mamba/conda create -n hsim-mutex-test python=3.7 withbullet habitat-sim -c conda-forge -c erikwijmans
mamba/conda create -n hsim-mutex-test python=3.7 headless habitat-sim -c conda-forge -c erikwijmans
mamba/conda create -n hsim-mutex-test python=3.7 headless withbullet habitat-sim -c conda-forge -c erikwijmans
```

(I only did osx and those aren't real packages, just dummies to test with, so don't try on not osx or actually try to use them)

## Types of changes

### DIFF
--- a/conda-build/matrix_builder.py
+++ b/conda-build/matrix_builder.py
@@ -52,6 +52,15 @@ def get_platform_string() -> str:
         raise RuntimeError(f"Unknown system: {platform.system()}")
 
 
+def get_headless_mode_for_test() -> bool:
+    if platform.system() == "Darwin":
+        return False
+    elif platform.system() == "Linux":  # noqa: SIM106
+        return True
+    else:
+        raise RuntimeError(f"Unknown system: {platform.system()}")
+
+
 def build_parser():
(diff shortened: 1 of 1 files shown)


Skylion007:
Just merge it as is then.

PR #1555, opened 2021-11-01
Python Viewer Navmesh Visualization via Key Press Toggle

JuanTheEngineer:
## Motivation and Context

This PR migrates the Navmesh Recompute and Toggle feature to the python viewer.

## How Has This Been Tested

Building and running `viewer.py`

## Types of changes

## Checklist


Skylion007 (approved):

examples/viewer.py
```diff
@@ -80,6 +83,9 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
         self.sim: habitat_sim.simulator.Simulator = None
         self.reconfigure_sim()
 
+        # compute NavMesh
+        self.navmesh_config_and_recompute()
```
aclegg3, line 87: As @Skylion007 suggested:
```suggestion
        # compute NavMesh if not already loaded by the scene.
        if not self.sim.pathfinder.is_loaded:
            self.navmesh_config_and_recompute()
```

examples/viewer.py
```diff
@@ -193,6 +199,7 @@ def reconfigure_sim(self) -> None:
         self.active_scene_graph = self.sim.get_active_scene_graph()
         self.default_agent = self.sim.get_agent(self.agent_id)
         self.agent_body_node = self.default_agent.scene_node
+        self.agent_body_node.translation = [2.65, 0.5, 4.5]
```
aclegg3, line 202: Let's not hardcode location in viewer.

aclegg3 (approved):
Small fixes and ready to go.

examples/viewer.py
```diff
@@ -274,6 +282,17 @@ def key_press_event(self, event: Application.KeyEvent) -> None:
             self.invert_gravity()
             logger.info("Command: gravity inverted")
 
+        elif key == pressed.N:
+            if event.modifiers == mod.SHIFT:
+                logger.info("Command: recompute navmesh")
+                self.navmesh_config_and_recompute()
+            else:
+                if self.navmesh_success:
```
Skylion007, line 290:
```suggestion
                if self.navmesh_success or self.sim.pathfinder.is_loaded:
```

examples/viewer.py
Skylion007: You also need to set `self.navmesh_success ` to `self.sim.pathfinder.is_loaded`. In fact, I would replace that variable with a direct access to `self.sim.pathfinder.is_loaded` and only use the `recompute_navmesh` return value to throw an error / exception if it fails.

aclegg3: Yeah, agreed. Let's just use `is_loaded` everywhere instead of tracking the recompute with a variable.

PR #1556, opened 2021-11-02
[HM3D Annotation] --HM3D SSD file support; Misc SSD Fixes

jturner65:
## Motivation and Context
This PR adds the following functionality for SSD support : 

- More intelligent default file name choices for stage atttributes (not only SSD but semantic mesh and navmesh as well)
- Function for setting the colormap to use for the TextureVisualizerShader, so that an external map can be used (instead of being forced to use the hard-coded Magnum turbo colormap)
- A colormap is synthesized when a file is loaded based on vertex colors, to be used for Semantic Scene visualizations.  The Magnum Turbo colormap is still supported as a fallback option.
- HM3D SSD file (house file) load and process supported based on the current annotation file format from Appen. 
- More consistent naming and access for SSD files from StageAttributes.
- Support parsing any Mn::Trade::Meshdata as a semantic scene mesh, not just ply files.

What's left is appropriate bindings and tests.

## How Has This Been Tested
Locally C++ and Python tests.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/GenericInstanceMeshData.cpp
+++ b/src/esp/assets/GenericInstanceMeshData.cpp
@@ -9,6 +9,7 @@
 #include <Corrade/Containers/ArrayViewStl.h>
 #include <Corrade/Utility/Algorithms.h>
 #include <Corrade/Utility/FormatStl.h>
+#include <Magnum/DebugTools/ColorMap.h>
 #include <Magnum/Image.h>
 #include <Magnum/ImageView.h>
 #include <Magnum/Math/Functions.h>
@@ -23,6 +24,7 @@
 #include "esp/core/Esp.h"
 #include "esp/geo/Geo.h"
 #include "esp/io/Json.h"
+#include "esp/scene/HM3DSemanticScene.h"
 
 namespace Cr = Corrade;
 namespace Mn = Magnum;
@@ -44,11 +46,20 @@ struct InstancePlyData {
    * semantic mesh for better frustum culling.
    */
   bool objIdsFromPly = false;
+  bool objPartitionsFromSSD = false;
+
+  /**
+   * @brief This is used to map every vertex to a particular partition, for
+   * culling.
+   */
+  std::vector<uint16_t> partitionIds;
 };
 
 Cr::Containers::Optional<InstancePlyData> parsePly(
     Mn::Trade::AbstractImporter& importer,
-    const std::string& plyFile) {
+    const std::string& plyFile,
+    std::vector<Magnum::Vector3ub>& colorMapToUse,
+    const std::shared_ptr<scene::SemanticScene>& semanticScene) {
   /* Open the file. On error the importer already prints a diagnostic message,
      so no need to do that here. The importer implicitly converts per-face
      attributes to per-vertex, so nothing extra needs to be done. */
@@ -94,33 +105,176 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
   Cr::Utility::copy(meshColors, Cr::Containers::arrayCast<Mn::Color3ub>(
                                     Cr::Containers::arrayView(data.cpu_cbo)));
 
-  /* Check we actually have object IDs before copying them, and that those are
-     in a range we expect them to be */
+  // Check we actually have object IDs before copying them, and that those are
+  // in a range we expect them to be
   data.objectIds.resize(meshData->vertexCount());
   Cr::Containers::Array<Mn::UnsignedInt> objectIds;
+
   if (meshData->hasAttribute(Mn::Trade::MeshAttribute::ObjectId)) {
     objectIds = meshData->objectIdsAsArray();
     data.objIdsFromPly = true;
+    data.objPartitionsFromSSD = false;
     const int maxVal = Mn::Math::max(objectIds);
     ESP_CHECK(
         maxVal <= 65535,
         Cr::Utility::formatString(
             "{}Object IDs can't be stored into 16 bits : Max ID Value : {}",
             msgPrefix, maxVal));
+    colorMapToUse.assign(Mn::DebugTools::ColorMap::turbo().begin(),
+                         Mn::DebugTools::ColorMap::turbo().end());
   } else {
-    // convert color data array to int (idx) for object ID
-    // removing duplicates returns array of unique ids for colors
-    // These ids should not be used to split the mesh with our current mesh
-    // process
+    // If object IDs are lacking, use vertex color to infer objectIDs, where
+    // each unique color corresponds to a unique objectID.
+    // These ids should probably not be used to split the mesh into submeshes
+    // for culling purposes, since there may be very many different vertex
+    // colors.
     data.objIdsFromPly = false;
-    auto out = Mn::MeshTools::removeDuplicatesInPlace(
-        meshData->mutableAttribute(Mn::Trade::MeshAttribute::Color));
-    objectIds = std::move(out.first);
+
+    // If a vertex color array and array of semantic ids are provided via a
+    // semantic scene descriptor, use these to replace the idxs found in using
+    // the removeDuplicatesInPlace process and their color mappings
+    if ((semanticScene != nullptr) && semanticScene->hasVertColorsDefined()) {
+      // built partion IDs from SSD regions, to split mesh for culling, instead
+      // of objectIDs
+      data.objPartitionsFromSSD = true;
+
+      // from SSD's objects_ array - this is the number of defined semantic
+      // descriptors corresponding to object instances.
+      // all this tapdancing because of the design of the Semantic Scene.
+      const auto& ssdObjs = semanticScene->objects();
+      std::size_t numSSDObjs = ssdObjs.size();
+
+      // map of colors as ints as key, where value is objectID and regionID from
+      // ssd
+      std::unordered_map<uint32_t, std::pair<int, int>>
+          tmpColorMapToSSDidAndRegionIndex;
+      tmpColorMapToSSDidAndRegionIndex.reserve(numSSDObjs);
+      // lambda function to convert a 3-vec representation of a color into an
+      // int
+      auto colorAsInt = [](const Mn::Vector3ub& color) -> uint32_t {
+        return (unsigned(color[0]) << 16) | (unsigned(color[1]) << 8) |
+               unsigned(color[2]);
+      };
+
+      // build maps of color ints to semantic IDs and color ints to region ids
+      // in SSD
+      // find max regions present, so we can have an "unassigned"/"unknown"
+      // region to move all verts whose region is
+
+      int maxRegion = -1;
+      for (int i = 0; i < numSSDObjs; ++i) {
+        const auto& ssdObj =
+            static_cast<scene::HM3DObjectInstance&>(*ssdObjs[i]);
+        const uint32_t colorInt = colorAsInt(ssdObj.getColor());
+        int regionIDX =
+            static_cast<scene::HM3DSemanticRegion&>(*ssdObj.region())
+                .getIndex();
+        tmpColorMapToSSDidAndRegionIndex[colorInt] = {ssdObj.getSemanticID(),
+                                                      regionIDX};
+        maxRegion = Mn::Math::max(regionIDX, maxRegion);
+      }
+
+      // increment maxRegion to use for unknown regions whose colors are not
+      // mapped
+      ++maxRegion;
+
+      // rebuild objectIDs vector (objectID per vertex) based on
+      // per vertex colors mapped
+      // 1st semantic ID for colors not found in SSD
+      std::size_t nonSSDObjID = tmpColorMapToSSDidAndRegionIndex.size();
+
+      // map regionIDs to affected verts using semantic color provided in
+      // SemanticScene, as specified in SSD file.
+      data.partitionIds.resize(meshData->vertexCount());
+      // map semantic IDs corresponding to colors from SSD file to appropriate
+      // verts (via index)
+      data.objectIds.resize(meshData->vertexCount());
+      // temporary holding structure to hold any non-SSD vert colors, so that
+      // the nonSSDObjID for new colors can be incremented appropriately
+      // not using set to avoid extra include
+      std::unordered_map<uint32_t, int> nonSSDVertColors;
+
+      // temporary map keyed by semantic ID and value being actual color for
+      // that ID. We need this since we do not know how many, if any, unique
+      // colors that do not have Semantic Mappings exist on the mesh
+      std::unordered_map<int, Mn::Vector3ub> tmpDestSSDColorMap;
+
+      // only go through all verts one time
+      // derive semantic ID and color and region/room ID for culling
+      int regionID = 0;
+      int semanticID = 0;
+      for (int vertIdx = 0; vertIdx < meshData->vertexCount(); ++vertIdx) {
+        Mn::Vector3ub meshColor = meshColors[vertIdx];
+        // Convert color to an int @ vertex
+        const uint32_t meshColorInt = colorAsInt(meshColor);
+
+        std::unordered_map<uint32_t, std::pair<int, int>>::const_iterator
+            ssdColorToIDAndRegionIter =
+                tmpColorMapToSSDidAndRegionIndex.find(meshColorInt);
+
+        if (ssdColorToIDAndRegionIter !=
+            tmpColorMapToSSDidAndRegionIndex.end()) {
+          // color is found in ssd mapping, so is legal color
+          semanticID = ssdColorToIDAndRegionIter->second.first;
+          regionID = ssdColorToIDAndRegionIter->second.second;
+        } else {
+          // color is not found in ssd mapping, so not legal color
+          // check if we've assigned a semantic ID to this color before, and if
+          // not do so
+          auto nonSSDClrRes =
+              nonSSDVertColors.insert({meshColorInt, nonSSDObjID});
+          if (nonSSDClrRes.second) {
+            // inserted, so increment nonSSDObjID
+            ++nonSSDObjID;
+          }
+          // map holds that color's nonSSDObjID
+          semanticID = nonSSDClrRes.first->second;
+          regionID = maxRegion;
+        }
+        // semantic ID for vertex
+        data.objectIds[vertIdx] = semanticID;
+        // color for
+        tmpDestSSDColorMap[semanticID] = meshColor;
+        // partition Ids for each vertex, for multi-mesh construction.
+        data.partitionIds[vertIdx] = regionID;
+      }
+
+      // color map with first nonSSDObjID-1 elements in proper order to match
+      // provided semantic IDs (so that ID is IDX of semantic color in map) and
+      // any overflow colors uniquely map 1-to-1 to an unmapped semantic ID as
+      // their index.
+      colorMapToUse.resize(tmpDestSSDColorMap.size());
+      for (const auto& elem : tmpDestSSDColorMap) {
+        colorMapToUse[elem.first] = elem.second;
+      }
+
+    } else {
+      // no remapping provided by SSD so just use what is synthesized
+      Cr::Containers::Array<Mn::Color3ub> colorsThatBecomeTheColorMap{
+          Mn::NoInit, meshColors.size()};
+      // copy the meshColors into colorsThatBecomeTheColorMap
+      Cr::Utility::copy(meshColors, colorsThatBecomeTheColorMap);
+
+      // removeDuplicatesInPlace returns array of unique ids for colors, and
+      std::pair<Cr::Containers::Array<Mn::UnsignedInt>, std::size_t> out =
+          Mn::MeshTools::removeDuplicatesInPlace(
+              Cr::Containers::arrayCast<2, char>(
+                  stridedArrayView(colorsThatBecomeTheColorMap)));
+
+      objectIds = std::move(out.first);
+      colorMapToUse.assign(colorsThatBecomeTheColorMap.begin(),
+                           colorsThatBecomeTheColorMap.end());
+
+      data.objPartitionsFromSSD = false;
+    }
   }
-  Mn::Math::castInto(Cr::Containers::arrayCast<2, Mn::UnsignedInt>(
-                         Cr::Containers::stridedArrayView(objectIds)),
-                     Cr::Containers::arrayCast<2, Mn::UnsignedShort>(
-                         Cr::Containers::stridedArrayView(data.objectIds)));
+  if (!data.objPartitionsFromSSD) {
+    // if data.objPartitionsFromSSD then data.objectIds were populated directly
+    Mn::Math::castInto(Cr::Containers::arrayCast<2, Mn::UnsignedInt>(
+                           Cr::Containers::stridedArrayView(objectIds)),
+                       Cr::Containers::arrayCast<2, Mn::UnsignedShort>(
+                           Cr::Containers::stridedArrayView(data.objectIds)));
+  }
 
   // Generic Semantic PLY meshes have -Z gravity
   const quatf T_esp_scene =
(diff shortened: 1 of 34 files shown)


src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -0,0 +1,178 @@
...
+    std::string objCategoryName = tokens[2].substr(1, tokens[2].size() - 2);
+    // room/region
+    int regionID = std::stoi(tokens[3]);
+    // build initial temp object
+    TempHM3DObject obj{0, objCategoryName, "", colorVec};
+    objInstance[instanceID] = obj;
+    // find category, build if dne
+    TempHM3DCategory tmpCat{
+        static_cast<int>(categories.size()), objCategoryName, {}};
+    auto categoryIter = categories.insert({objCategoryName, tmpCat});
```
Skylion007, line 117: Should be emplace?

src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -0,0 +1,178 @@
...
+      objItem->setObjInstanceAndName(objCatIdx++);
+    }
+    // build category item and place within temp map
+    item.second.category_ = std::make_shared<HM3DObjectCategory>(
+        HM3DObjectCategory(item.second.ID, item.second.name));
+  }
+
+  // build all categories
+  // object categories
+  scene.categories_.clear();
```
Skylion007, line 140:
```suggestion
  scene.categories_.clear();
  scene.categories_.reserve(categories.size());
```

src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -0,0 +1,178 @@
...
+  for (const auto& catItem : categories) {
+    scene.categories_.emplace_back(catItem.second.category_);
+  }
+  // build all regions
+  // regions
+  scene.regions_.clear();
```
Skylion007, line 146:
```suggestion
  scene.regions_.clear();
  scene.regions_.reserve(regions.size());
```

src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -0,0 +1,178 @@
...
+    regionPtr->index_ = regionItem.second.regionID;
+    for (TempHM3DObject* objItem : regionItem.second.objInstances) {
+      objItem->region = regionPtr;
+    }
+    scene.regions_.emplace_back(regionPtr);
+  }
+
+  // build all object instances
+  // object instances
+  scene.objects_.clear();
```
Skylion007, line 158: reserve space here too.

Skylion007 (changes requested):

src/esp/gfx/TextureVisualizerShader.h
```diff
@@ -46,12 +46,28 @@ class TextureVisualizerShader : public Magnum::GL::AbstractShaderProgram {
...
+   * colors.
+   *
+   * @param colorMap The map of colors to use
+   * @param wrapEdges Whether queries exceeding the length of @p colorMap will
+   * use the final color, or will wrap around on the texture.
+   * @return Reference to self (for method chaining)
+   */
+  TextureVisualizerShader& setColorMapTexture(
+      Corrade::Containers::ArrayView<const Magnum::Vector3ub> colorMap,
+      bool wrapEdges,
```
mosra, line 60: [Boolean function parameters are an API smell](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) in languages that don't have the concept of named parameters. I'd expose the `Mn::SamplerWrapping` type directly.

I'm also not sure about the offset/scale parameters. Why not just let the caller call also `setColorMapTransformation()`? There the parameters are documented and the responsibility/interaction is clear.

src/esp/gfx/TextureVisualizerShader.h
```diff
@@ -46,12 +46,28 @@ class TextureVisualizerShader : public Magnum::GL::AbstractShaderProgram {
...
+      float offset,
+      float scale);
+
   /**
    * @brief
    * Offset and scale applied to each input pixel value from depth texture or
    * the ObjectId texture, resulting value is then used to fetch a color from a
-   * color map bound with bindColorMapTexture(). Here we use the Magnum built-in
-   * color map: Magnum::DebugTools::ColorMap::turbo();
+   * color map bound with bindColorMapTexture(). Here we use either the Magnum
```
mosra, line 68: This whole comment is wrong :) Here's no `bindColorMapTexture()`, only `rebindColorMapTexture()`, and the shader is also not responsible for picking the Turbo color map anymore, that's supplied via `setColorMapTexture()`.

src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -0,0 +1,188 @@
...
+                                   const quatf& /*rotation*/) {
+  // convert a hex string containing 3 bytes to a vector3ub (represents a color)
+  auto getVec3ub = [](const std::string& hexValStr) -> Mn::Vector3ub {
+    size_t len = hexValStr.length();
+    std::vector<uint8_t> resVec;
+    for (size_t i = 0; i < len; i += 2) {
+      uint8_t x = strtoul(hexValStr.substr(i, 2).c_str(), nullptr, 16);
+      resVec.push_back(x);
+    }
+    return Mn::Vector3ub{resVec[0], resVec[1], resVec[2]};
```
mosra, line 83: Uhh, C++ string-to-number APIs are *bad*, but this is somehow even worse :sweat_smile: Why not parse the whole thing as a single integer and then use that?

```suggestion
    const unsigned val = std::stoul(hexValStr, nullptr, 16);
    return {uint8_t((val >> 16) & 0xff),
            uint8_t((val >>  8) & 0xff),
            uint8_t((val >>  0) & 0xff)};
```

src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto colorVec = getVec3ub(tokens[1]);
+    // object category name
+    // construct instance name by counting instances of same category
+    // get rid of quotes
+    std::string objCategoryName = tokens[2];
+    auto loc = objCategoryName.find('\"');
+    while (loc != std::string::npos) {
+      objCategoryName.erase(loc, 1);
+      loc = objCategoryName.find('\"');
+    }
```
mosra, line 114:
```suggestion
    std::string objCategoryName = Cr::Utility::String::replaceAll(tokens[2], "\"", "");
```

mosra (commented):
Didn't really go through everything because the colormap/objectID-generating code may need some further work.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -94,33 +105,184 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
+      // from SSD's objects_ array - this is the number of defined semantic
+      // descriptors corresponding to object instances.
+      // all this tapdancing because of the design of the Semantic Scene.
+      const auto& ssdObjs = semanticScene->objects();
+      std::size_t numSSDObjs = ssdObjs.size();
+
+      // map of colors as ints as key, where value is objectID from ssd
+      std::unordered_map<uint32_t, int> tmpColorMapToSSDid;
+      // map of colors as ints as key, where value is regionID from ssd
+      std::unordered_map<uint32_t, int> tmpColorMapToRegionIndex;
```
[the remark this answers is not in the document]
Skylion007: @mosra Actually, shouldn't we use Cr::Containers::Pair?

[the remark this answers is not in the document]
mosra: Doesn't matter much in this case :) The main advantage of `Containers::Pair` is trivial copyability, which doesn't really get to be used here given that every element is a separate allocation.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -94,33 +105,184 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
     objectIds = meshData->objectIdsAsArray();
     data.objIdsFromPly = true;
+    data.objPartitionsFromSSD = false;
     const int maxVal = Mn::Math::max(objectIds);
     ESP_CHECK(
         maxVal <= 65535,
         Cr::Utility::formatString(
             "{}Object IDs can't be stored into 16 bits : Max ID Value : {}",
             msgPrefix, maxVal));
+    colorMapToUse.resize(Mn::DebugTools::ColorMap::turbo().size());
```
Skylion007, line 123: Just an FYI, resize also default initializes the values in the std::vector, which seems wasteful is you are about to overwrite it. What you are looking for might be `reserve()` with `shrink_to_fit`

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -191,27 +191,51 @@ StageAttributes::ptr StageAttributesManager::initNewObjectInternal(
...
+      newAttributes->setSemanticDescriptorFilename(ssdFileName);
+
+      // Build default semantic mesh filename as root stage name ending with
+      // "_semantic.ply", for back-compat with Mp3d
+      std::string semanticMeshFilename =
+          this->findFilenameUsingCriteria(attributesHandle, {"_semantic.ply"});
+      // if present, file was found, so set value (overriding attributes
+      // default)
+      if (!semanticMeshFilename.empty()) {
+        newAttributes->setSemanticAssetHandle(semanticMeshFilename);
```
Skylion007, line 236:
```suggestion
        newAttributes->setSemanticAssetHandle(std::move(semanticMeshFilename));
```

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -191,27 +191,51 @@ StageAttributes::ptr StageAttributesManager::initNewObjectInternal(
...
+
+      // set defaults for navmesh, semantic mesh and lexicon handles
+      // start with root stage name, including path but without final extension
+      // default navmesh should have .navmesh ext
+      std::string navmeshFilename =
+          this->findFilenameUsingCriteria(attributesHandle, {".navmesh"});
+      // if present, file was found, so set value (overriding attributes
+      // default)
+      if (!navmeshFilename.empty()) {
+        newAttributes->setNavmeshAssetHandle(navmeshFilename);
```
Skylion007, line 206: Don't forget to move temp variable strings. 👍 
```suggestion
        newAttributes->setNavmeshAssetHandle(std::move(navmeshFilename));
```

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -335,16 +359,27 @@ void StageAttributesManager::setValsFromJSONDoc(
...
   if (io::readMember<std::string>(jsonConfig, "nav_asset", navmeshFName)) {
-    navmeshFName = Cr::Utility::Directory::join(stageLocFileDir, navmeshFName);
     // if "nav mesh" is specified in stage json set value (override default).
+    // navmesh filename might already be fully qualified; if not, might just be
+    // file name
+    if (!Corrade::Utility::Directory::exists(navmeshFName)) {
+      navmeshFName =
+          Cr::Utility::Directory::join(stageLocFileDir, navmeshFName);
+    }
     stageAttributes->setNavmeshAssetHandle(navmeshFName);
```
Skylion007, line 369: std::move here too

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -335,16 +359,27 @@ void StageAttributesManager::setValsFromJSONDoc(
...
+                                  semanticSceneDescriptor)) {
+    // if "semantic_descriptor_filename" is specified in stage json, set value
+    // (override default).
+    // semanticSceneDescriptor filename might already be fully qualified; if
+    // not, might just be file name
+    if (!Corrade::Utility::Directory::exists(semanticSceneDescriptor)) {
+      semanticSceneDescriptor = Cr::Utility::Directory::join(
+          stageLocFileDir, semanticSceneDescriptor);
+    }
+    stageAttributes->setSemanticDescriptorFilename(semanticSceneDescriptor);
```
Skylion007, line 382: std::move here too

Skylion007 (approved):
Looks good with @mosra 's changes.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -94,33 +105,184 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
+      // the nonSSDObjID for new colors can be incremented appropriately
+      // not using set to avoid extra include
+      std::unordered_map<uint32_t, int> nonSSDVertColors;
+
+      // only go through all verts one time
+      for (int vertIdx = 0; vertIdx < tmpMeshObjectIds.size(); ++vertIdx) {
+        // Get mesh-derived objectID @ vert
+        const int objIdAtVert = tmpMeshObjectIds[vertIdx];
+        // Get meshcolor @ vertexobjPartitionsFromSSD
+        Mn::Vector3ub meshColor = tmpMeshColorMap[objIdAtVert];
```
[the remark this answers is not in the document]
jturner65: I wanted to move this code to after this block too, but I need the number of total unique colors in the mesh in this block of code so that I could reserve the size of the `tmpDestSSDColorMap`, since I need to be able to place a particular vertex's color within `tmpDestSSDColorMap` at the index defined by the semantic ID, either from the ssd map, or from the "unassigned" region at the end of the vertex list (for colors not in the ssd).    

Instead of performing the entire index and color mapping, beforehand, though, I can just build a map of id to color first, and then copy from -that- map into the appropriate pre-allocated vector.  Still -much- cheaper than going through the verts before-hand.

src/esp/scene/HM3DSemanticScene.cpp
jturner65: That's what I get for trying to work out an SO suggestion.

src/esp/metadata/managers/StageAttributesManager.cpp
jturner65: All these string setters are const string& setters, so moving isn't really useful here, is it?

Skylion007:
Ah, whoops nvm then.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -94,33 +105,183 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
+      data.objectIds.resize(meshData->vertexCount());
+      // temporary holding structure to hold any non-SSD vert colors, so that
+      // the nonSSDObjID for new colors can be incremented appropriately
+      // not using set to avoid extra include
+      std::unordered_map<uint32_t, int> nonSSDVertColors;
+
+      // temporary map keyed by semantic ID and value being actual color for
+      // that ID. We need this since we do not know how many, if any, unique
+      // colors that do not have Semantic Mappings exist on the mesh
+      std::unordered_map<int, Mn::Vector3ub> tmpDestSSDColorMap;
```
mosra, line 199: Instead of this, why not `resize()` the `colorMapToUse` to `id + 1` that you're about to insert, and then `colorMapToUse[id] = stuff` directly? You're copying this to a linear array after anyway so this won't save anything.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -94,33 +105,183 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
+      Cr::Containers::Array<Mn::Vector3ub> tmpDestSSDColorVec{
+          Mn::NoInit, tmpDestSSDColorMap.size()};
+      for (const auto& elem : tmpDestSSDColorMap) {
+        tmpDestSSDColorVec[elem.first] = elem.second;
+      }
+
+      // colors need to be in the appropriate order
+      colorMapToUse.resize(tmpDestSSDColorVec.size());
+      Cr::Utility::copy(tmpDestSSDColorVec,
+                        Cr::Containers::arrayView(colorMapToUse));
```
mosra, line 254: See my comment above.

```suggestion
```

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -94,33 +105,183 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
+      std::pair<Cr::Containers::Array<Mn::UnsignedInt>, std::size_t> out =
+          Mn::MeshTools::removeDuplicatesInPlace(
+              Cr::Containers::arrayCast<2, char>(
+                  stridedArrayView(colorsThatBecomeTheColorMap)));
+
+      objectIds = std::move(out.first);
+      const Corrade::Containers::ArrayView<Magnum::Vector3ub>
+          clrsListFromVerts = Cr::Containers::arrayCast<Mn::Vector3ub>(
+              colorsThatBecomeTheColorMap.prefix(out.second));
+      colorMapToUse.assign(clrsListFromVerts.begin(), clrsListFromVerts.end());
```
mosra, line 273: Won't just this work? IIRC `assign()` doesn't care what input type it is as long as it's implicitly convertible.

```suggestion
      colorMapToUse.assign(colorsThatBecomeTheColorMap.begin(), colorsThatBecomeTheColorMap.begin() + out.second);
```

src/esp/gfx/Renderer.cpp
```diff
@@ -164,7 +165,9 @@ struct Renderer::Impl {
       } else if (type == sensor::SensorType::Semantic) {
         shader->bindObjectIdTexture(tgt.getObjectIdTexture());
       }
-      shader->setColorMapTransformation(colorMapOffset, colorMapScale);
+      if ((colorMapOffset >= 0) && (colorMapScale >= 0)) {
+        shader->setColorMapTransformation(colorMapOffset, colorMapScale);
```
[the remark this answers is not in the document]
jturner65: As we discussed :) the semantic visualization and the depth visualization each use different instances of TextureVisualizerShader and different compiled frag shader programs.  Because of this (that both use cases have their own ShaderProgram and compiled frag shader) they do not "cross-pollinate" their values, so it is reasonable to assume that these values will not need to be reset on subsequent draws.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -94,33 +105,183 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
+      Cr::Containers::Array<Mn::Color3ub> colorsThatBecomeTheColorMap{
+          Mn::NoInit, meshColors.size()};
+      // copy the meshColors into colorsThatBecomeTheColorMap
+      Cr::Utility::copy(meshColors, colorsThatBecomeTheColorMap);
+
+      // removeDuplicatesInPlace returns array of unique ids for colors, and
+      std::pair<Cr::Containers::Array<Mn::UnsignedInt>, std::size_t> out =
+          Mn::MeshTools::removeDuplicatesInPlace(
+              Cr::Containers::arrayCast<2, char>(
+                  stridedArrayView(colorsThatBecomeTheColorMap)));
```
mosra, line 267: I still have a feeling that populating a unique color array could be done directly in the above loop, unifying the handling... :)

src/esp/gfx/TextureVisualizerShader.h
jturner65: > I'm also not sure about the offset/scale parameters. Why not just let the caller call also `setColorMapTransformation()`? There the parameters are documented and the responsibility/interaction is clear, now it's a bit unclear how the two functions affect each other.

Originally the code was set up to fully instantiate the ShaderProgram through the constructor (including setting these values) instead of having a secondary call to setColorMapTransformation for these individual values. I tried to keep my refactoring to the existing code as minimal as possible, but if you think this should be changed, i'm all for it.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -148,27 +137,22 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
+      // map of colors as ints as key, where value is objectID and regionID from
+      // ssd
+      std::unordered_map<uint32_t, std::pair<int, int>>
+          tmpColorMapToSSDidAndRegionIndex;
+      tmpColorMapToSSDidAndRegionIndex.reserve(numSSDObjs);
       // lambda function to convert a 3-vec representation of a color into an
       // int
       auto colorAsInt = [](const Mn::Vector3ub& color) -> uint32_t {
-        return (((color[0] * 256) + color[1]) * 256) + color[2];
+        return (((color[0] << 16) + color[1]) << 16) + color[2];
```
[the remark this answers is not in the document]
jturner65: oops Thanks!

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -135,30 +296,44 @@ Cr::Containers::Optional<InstancePlyData> parsePly(
...
-  if (splitMesh && parseResult->objIdsFromPly) {
-    const InstancePlyData& data = *parseResult;
-    std::unordered_map<uint16_t, PerObjectIdMeshBuilder> objectIdToObjectData;
-    for (size_t i = 0; i < data.cpu_ibo.size(); ++i) {
-      const uint32_t globalIndex = data.cpu_ibo[i];
-      const uint16_t objectId = data.objectIds[globalIndex];
-      if (objectIdToObjectData.find(objectId) == objectIdToObjectData.end()) {
+  if (splitMesh &&
+      (parseResult->objIdsFromPly || parseResult->objPartitionsFromSSD)) {
+    std::vector<uint16_t> meshPartitionIds = parseResult->objIdsFromPly
```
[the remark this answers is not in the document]
jturner65: not supposed to :smile:

src/esp/assets/GenericInstanceMeshData.cpp
jturner65: id is different for each vert - can get smaller, so don't want to resize colorMapToUse smaller than it was for some previous vert.

src/esp/assets/GenericInstanceMeshData.cpp
jturner65: Sure, we could only conditionally build the `tmpColorMapToSSDidAndRegionIndex` if the ssd mapping exists.  If it is empty, it would make a map full of "non-ssd" colors.  

Using this code path for this case would be slower than what we are doing now, and less elegant than the solution using the magnum tools.

Skylion007:
This looks like the Simulator is segfaulting or failing some other way when loading the WebGL build.

eundersander (commented):
I left some comments.

src/esp/assets/GenericInstanceMeshData.cpp
```diff
@@ -44,11 +46,20 @@ struct InstancePlyData {
...
+  bool objPartitionsFromSSD = false;
+
+  /**
+   * @brief This is used to map every vertex to a particular partition, for
+   * culling.
+   */
+  std::vector<uint16_t> partitionIds;
 };
 
 Cr::Containers::Optional<InstancePlyData> parsePly(
```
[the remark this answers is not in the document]
jturner65: > GenericInstanceMeshData looks to be a pretty simple container for mesh data. Would it make sense to keep this class simple and move all your annotation-parsing logic to a separate class and file?

This class was always used for the semantic mesh processing - the annotation code that I added just adds flexibility for different ply formats than just the hard-coded mp3d support.

For long-term handling considerations, i see this class holding only specialized annotation information (if that - i can see a future without it too) while the general mesh data is handled by GenericMeshData, if not Mn::Trade::MeshData directly.

> Related: it would help if this class GenericInstanceMeshData had a class-level comment. Its parent class says "Base class for storing mesh asset data including geometry and topology" and I so wouldn't expect such a class to know e.g. the details of region indices in annotation data.

Again, I just expanded existing annotation functionality in this class to handle the specific needs of the Hm3D annotations - previously it used hardcoded vert ids in the ply mesh to denote submeshes for culling.  I can certainly more clearly comment the purpose of this class, but like I said, from an organizational standpoint, it would probably be better if this class only held semantic-specific functionality, expanding on existing whatever MeshData structure ended up being the backbone for this class.

A little bit of redesign/refactor would pay big dividends in this department, but I felt such engineering overhead was beyond the scope of this PR.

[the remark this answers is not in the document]
eundersander: >I felt such engineering overhead was beyond the scope of this PR

I will push back a little here on the idea "refactors are out of scope". You're adding nearly 1000 lines of code in this PR. I believe we should view such large changes as opportunities to refactor, rather than piling on more tech debt to an area of code that we already regard as needing refactor.

>it would probably be better if this class only held semantic-specific functionality
Consider renaming the class in this case.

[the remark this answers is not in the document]
jturner65: > I will push back a little here on the idea "refactors are out of scope". You're adding nearly 1000 lines of code in this PR. I believe we should view such large changes as opportunities to refactor, rather than piling on more tech debt to an area of code that we already regard as needing refactor.

I completely share your concerns, and normally would be in total agreement regarding addressing refactors and tech-debt whenever possible. 

However, the support for the Appen annotated hm3d meshes was presented to me as being urgent, and so I approached implementation as the priority. Refactors were restricted to what was necessary to support the new functionality in as generalized a way as possible while not breaking existing code (theoretically) and addressing some of the brittleness inherent in the old designs.  There's very little tech debt that's been added - the new functionality is required for hm3d annotation support, so any future reworks are going to need it on some level;  some (very small amount) of tech debt has been removed, and, unfortunately a lot of debt remains.

This affected how multiple classes performed their duties:
 - Renderer (instance shaders on construction instead of the previous method of create shader on first visualize call, expand colormap support)
 - GenericMeshLoading (support multiple vert color formats, support no baked-in ids, better multi-mesh culling support)
 - Sim/ResourceManager (create SemanticScene descriptor in resource manager to make available during Semantic Scene mesh load) 
 
 
> Consider renaming the class in this case

Good idea.

jturner65:
Most recent force-push changed how shaders were loading. Loading them in the Renderer constructor (as this PR previously introduced) was not compatible with the current Javascript calling protocols, so this mechanism was removed and the existing on-demand instancing is being used instead.

src/esp/assets/GenericInstanceMeshData.cpp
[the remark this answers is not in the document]
mosra: This is by the way why `Containers::Array` was made non-copyable -- to prevent accidents like this. And writing an explicit `Utility::copy()` is annoying enough to make you think twice if you really need to copy :)

src/esp/assets/GenericInstanceMeshData.cpp
mosra: Err, I meant this:

```cpp
if(colorMapToUse.size() <= id) colorMapToUse.resize(id + 1);
colorMapToUse[id] = stuff;
```

Using a hashmap for something that's basically contiguous indexing doesn't seem right to me :)

jturner65: Except the id does not necessarily have to be contiguous. It is provided in the semantic lexicon text document, and so long as it is unique, it can be anything.

src/utils/viewer/viewer.cpp
```diff
@@ -268,9 +269,11 @@ In LOOK mode (default):
     Click and drag to rotate the agent and look up/down.
   RIGHT:
     (With 'enable-physics') Click a surface to instance a random primitive object at that location.
+  SHIFT-LEFT:
+    Read Semantic ID and tag of clicked object (Currently only HM3D);
   SHIFT-RIGHT:
     Click a mesh to highlight it.
-  CTRL-RIGHT:
+  CTRL-RIGHT:m
```
mosra, line 276: :eyes:

src/esp/assets/GenericSemanticMeshData.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+namespace Cr = Corrade;
+namespace Mn = Magnum;
+
+namespace esp {
+namespace assets {
+
+std::vector<std::unique_ptr<GenericSemanticMeshData>>
+GenericSemanticMeshData::buildSemanticMeshData(
+    const Cr::Containers::Optional<Mn::Trade::MeshData>& srcMeshData,
```
mosra, line 30: No `Optional` here please :) The code of the function assumes the optional is non-null anyway.

mosra (commented):
Two minor things + the unresolved color map `map` issue I asked about on Slack. Other than that looks good :)

PR #1561, opened 2021-11-09
Fairmotion ShortestPath Following Character

JuanTheEngineer:
## Motivation and Context

This PR aims to move an un-posed character along a generated ShortestPath with a walk cycle created from Fairmotion data sets. The final product is able to choose two random points on the NavMesh and generates a smoothed random path. Then using the Fairmotion Habitat interface, A character is loaded and set to follow the path with a very smooth walking animation. The demo can be easily activated with a press of the 'J' key.

Here is a short video of the feature in action:

https://user-images.githubusercontent.com/59671006/142706669-947c0475-56b8-4c4e-aef4-70d9944a52d4.mov

## How Has This Been Tested

The majority of this feature lives in the Fairmotion viewer application and was tested by running that application.

## Types of changes

## Checklist


examples/fairmotion_interface.py
```diff
@@ -546,6 +550,116 @@ def belongs_to(self, obj_id: int) -> bool:
...
+        )
+        self.puck.translation = path.points[0] + mn.Vector3(0.0, 1.0, 0.0)
+
+    def update_pathfollower(self):
+        """
+        When called, this method will update the position and orientation of the
+        path follower.
+        """
+        # Skip function
+        if self.puck == None or self.path_points == None:
```
Skylion007, line 591:
```suggestion 
        if self.puck is None or self.path_points is None:
```

examples/fairmotion_interface.py
```diff
@@ -546,6 +550,116 @@ def belongs_to(self, obj_id: int) -> bool:
...
+            curr_pose,
+            curr_root_translate,
+            curr_root_rotation,
+        ) = self.convert_CMUamass_single_pose(
+            self.path_motion.poses[
+                (self.path_follow_stepper) % self.path_motion.num_frames()
+            ],
+            self.puck,
+        )
+        print(f"curr_root_translate = {curr_root_translate}")
```
Skylion007, line 605: logger

Skylion007 (changes requested):

examples/fairmotion_interface.py
```diff
@@ -546,9 +567,171 @@ def belongs_to(self, obj_id: int) -> bool:
...
+        )
+
+        # translation
+        delta_P_vector = mn.Vector3(next_root_T - curr_root_T)
+        forward_vector = mn.Vector3(Move.walk_to_walk.direction_forward)
+        up_vector = mn.Vector3.z_axis()
+
+        # 1. rotate P->  by Angle(up_v <-> up_sim) around (up_v X up_sim)
+        angle1 = mn.math.angle(up_vector.normalized(), mn.Vector3.y_axis())
+        print(f"1angle = {angle1}")
```
Skylion007, line 638: use logger

examples/fairmotion_interface.py
```diff
@@ -546,9 +566,298 @@ def belongs_to(self, obj_id: int) -> bool:
...
+
+        # First update with step_size 0 to start character
+        self.update_pathfollower_sequential(step_size=0)
+
+    def update_pathfollower_sequential(
+        self, step_size: int = 1, path_time: Optional[float] = None
+    ):
+        """
+        [FILL AFTER IMPLEMENTATION]
+        """
```
aclegg3, line 602: Missing comment

aclegg3, line 602: Missing docstring

examples/fairmotion_interface.py
```diff
@@ -546,9 +566,298 @@ def belongs_to(self, obj_id: int) -> bool:
...
+        )
+
+        # either take argument value for path_time or continue with cycle
+        self.path_time = path_time or (self.path_time + (step_size / 60.0))
+
+        # compute current frame from self.path_time w/ wrapping
+        mocap_time_length = walk.num_of_frames * (1.0 / 120.0)
+        mocap_cycles_past = math.floor(self.path_time / mocap_time_length)
+        mocap_time_curr = math.fmod(self.path_time, mocap_time_length)
+        mocap_frame = int(mocap_time_curr * 120)
```
aclegg3, line 621: Note: hardcoded motion and application framerates here. Is there a more general way to get these values?

examples/fairmotion_interface.py
```diff
@@ -546,9 +566,298 @@ def belongs_to(self, obj_id: int) -> bool:
...
+        # find distance progressed along shortest path
+        path_displacement = (
+            mocap_cycles_past * walk.map_of_total_displacement[LAST]
+            + walk.map_of_total_displacement[mocap_frame]
+        )
+
+        # handle wrapping or edgecase for dath displacement passing goal
+        if path_displacement > self.full_path_length:
+            self.path_time = 0.0
+            self.path_displacement = 0.0
```
aclegg3, line 632: Should this be an fmod or other continuous wrapper instead of a hard reset to 0?

examples/fairmotion_interface.py
```diff
@@ -546,9 +566,298 @@ def belongs_to(self, obj_id: int) -> bool:
...
+    The Move class is collection of stats that will hold the different movement motions
+    for the character to use when following a path. The character is left-footed so that
+    is our reference for with step the motions assume first.
+    """
+
+    @dataclass
+    class MotionData:
+        """
+        [FILL AFTER NEXT PR REVIEW]
+        """
```
aclegg3, line 754: Missing docstring

examples/fairmotion_interface.py
```diff
@@ -546,9 +566,298 @@ def belongs_to(self, obj_id: int) -> bool:
...
+
 
 class Preview(Enum):
     OFF = 0
     KEYFRAMES = 1
     TRAJECTORY = 2
     ALL = 3
+
+
+class Activity(Enum):
```
aclegg3, line 860: Add a small comment about these?

examples/motion_viewer.py
```diff
@@ -63,7 +64,35 @@ def __init__(
...
+        self.spline_path_traj_obj_id = -1
+
+        self.navmesh_config_and_recompute()
+
+    def debug_draw(self):
+        """
+        Additional draw commands to be called during draw_event.
+        """
+
+        def draw_frame():
```
aclegg3, line 76: This function is duplicated in fairmotion_interface.py. Maybe break it out into a common utility location to avoid duplication?

examples/motion_viewer.py
```diff
@@ -76,6 +105,7 @@ def play_motion() -> None:
             if self.fm_demo.motion is not None:
                 self.fm_demo.next_pose()
                 self.fm_demo.next_pose()
```
aclegg3, line 107: Why is this called twice? If it is to sync the 120Hz motion with the 60Hz viewer update then I suggest a slightly more general formulation (e.g. pass the timestep into next_pose and handle frame choice and tracking internally instead of assuming a hardcoded 2/1 ratio.

examples/motion_viewer.py
```diff
@@ -97,14 +127,25 @@ def key_press_event(self, event: Application.KeyEvent) -> None:
...
+                self.remove_short_path_traj_obj()
                 self.fm_demo.load_model()
 
         elif key == pressed.J:
-            if not self.sim.pathfinder.is_loaded:
+            if event.modifiers == mod.ALT:
+                self.simulating = False
+                self.fm_demo.update_pathfollower_sequential(
+                    path_time=random.uniform(0.0, 10.0)
+                )
```
aclegg3, line 138: Can this be called before `setup_pathfollower`? If not you should add some protection/warning from potential crash.

aclegg3 (commented):
Looks pretty good overall. Mostly need to address a couple of missing comments and mild edge cases.

Looks like the habitat-lab test failure is unrelated to this change and can be ignored.

examples/fairmotion_interface.py
```diff
@@ -546,9 +566,298 @@ def belongs_to(self, obj_id: int) -> bool:
...
+        look_at_path_T = mn.Matrix4.look_at(
+            char_pos, char_pos + forward_direction.normalized(), mn.Vector3.y_axis()
+        )
+
+        full_transform = walk.motion.poses[mocap_frame].get_transform(ROOT, local=True)
+        full_transform = mn.Matrix4(full_transform)
+        full_transform.translation -= walk.center_of_root_drift
+        full_transform = mn.Matrix4.from_(
+            global_neutral_correction.to_matrix(), mn.Vector3()
+        ).__matmul__(full_transform)
```
Skylion007, line 653:
```suggestion 
        ) @ full_transform
```

examples/fairmotion_interface.py
```diff
@@ -546,9 +566,298 @@ def belongs_to(self, obj_id: int) -> bool:
...
+
+        # while transform is facing -Z, remove forward displacement
+        full_transform.translation *= mn.Vector3.x_axis() + mn.Vector3.y_axis()
+        full_transform = look_at_path_T.__matmul__(full_transform)
```
Skylion007, line 657:
```suggestion
        full_transform = look_at_path_T @ full_transform
```

examples/motion_viewer.py
```diff
@@ -319,43 +363,115 @@ def remove_selector_obj(self):
...
-
-        sample1 = self.sim.pathfinder.get_random_navigable_point()
-        sample2 = self.sim.pathfinder.get_random_navigable_point()
+        self.spline_path_traj_obj_id = -1
+
+        found_path = False
+        while not found_path:
+            sample1 = None
+            sample2 = None
+            while any([sample1 is None, sample2 is None]):
```
Skylion007, line 383:
```suggestion 
            while sample1 is None or sample2 is None:
```

Skylion007 (changes requested):
Lot of nits

examples/fairmotion_interface.py
JuanTheEngineer: Yes there should be a way to

examples/fairmotion_interface.py
JuanTheEngineer: My thinking is that when the character is expected to pass the full length of the path, it should reset to the origin of the path and begin again. This should not preserve any overflow of the path_displacement because the intended task is to move the character from point a to point b and the looping is merely a demo so if shouldn't need to exhibit wrapping. What do you think?

examples/motion_viewer.py
JuanTheEngineer: Even though they both have the same code, they are different in that the motion_viewer is viewer rate called and the fairmotion one is called only during simulation. They are different but the reason I left the code to be the generic origin print is because I meant for it to be demo of what devs can do with the debug methods. It's like a head start for coding up debug lines so I though the most generic code to leave would be that. But if I move the origin code to a utility then it defeats the purpose of the head start code because devs would have to search for it in the utilities class and copy paste from there to begin adding debug draws.

examples/motion_viewer.py
JuanTheEngineer: I do with this line
```
# precondition
        if not all(
            [self.model, self.path_points, self.activity == Activity.PATH_FOLLOW_SEQ]
        ):
```

examples/fairmotion_interface.py
aclegg3: Fair enough, this use case does focus on a one-way trip.

aclegg3 (approved):
LGTM

PR #1568, opened 2021-11-16
[Part 2 of 2]--SceneInstance snapshot and save.

jturner65:
## Motivation and Context
This PR provides the functionality to build a scene instance from the current simulation state and to properly save that scene instance to a JSON file.  Saving Attributes to JSON has also been improved for all appropriate Attributes classes.

Functionality being introduced : 
- Create a SceneInstanceAttributes based on the current scene state on demand.  This will facilitate creating scenes with stable physics on load.
- Save all metadata::Attributes (including SceneInstanceAttributes) to disk as JSON based on expected format.  The existing functionality provided a generic mechanism to save Configuration values and subconfigs to JSON files, but the various metadata::Attributes classes should only save a subset of their fields as JSON (there are internal fields that have programmatic use but are not appropriate to expose).  This PR provides customized handling for each class, to only save the fields, and subconfigs, that are desired to be saved.
- Existing test has been modified to verify all save/load processes are acting as expected. 

Some extra information : 
Viewer.cpp has been modified to provide a key-stroke to record the current scene instance, saving it to disk with the same name along with an increment (copy xxxx).  Repeatedly saving the scene instance will increase the increment. An unintended, but welcome, consequence of this is that the newly saved scene instance is immediately available to cycle to via Tab/Shift-tab, where it will be loaded in and displayed.

A mechanism is also provided for saving with a new, user-provided, name, which will forcibly overwrite any existing scene instance configs with the same name.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/PhysicsManager.h
+++ b/src/esp/physics/PhysicsManager.h
@@ -967,7 +967,7 @@ class PhysicsManager : public std::enable_shared_from_this<PhysicsManager> {
   }
   /**
    * @brief This will populate the passed @p sceneInstanceAttrs with the current
-   * stage, object and articulated object insttances reflecting the currernt
+   * stage, object and articulated object instances reflecting the current
    * state of the physics world.
    * @param sceneInstanceAttrs A copy of the intialization attributes that
    * created the current scene.  The various object instance attributes will be
--- a/src/esp/physics/PhysicsManager.h
+++ b/src/esp/physics/PhysicsManager.h
@@ -974,8 +974,8 @@ class PhysicsManager : public std::enable_shared_from_this<PhysicsManager> {
    * overwritten by the current scene state data.
    */
   void buildCurrentStateSceneAttributes(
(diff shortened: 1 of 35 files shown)


Skylion007 (approved):

Skylion007 (approved):

src/esp/physics/PhysicsManager.h
```diff
@@ -965,6 +965,17 @@ class PhysicsManager : public std::enable_shared_from_this<PhysicsManager> {
               "No RigidConstraint exists with constraintId =" << constraintId);
     return rigidConstraintSettings_.at(constraintId);
   }
+  /**
+   * @brief This will populate the passed @p sceneInstanceAttrs with the current
+   * stage, object and articulated object insttances reflecting the currernt
```
aclegg3, line 970:
```suggestion
   * stage, object and articulated object instances reflecting the current
```

aclegg3 (approved):
LGTM.

PR #1573, opened 2021-11-23
[Logging]--Minimize "nospace" directives; add timestamp to log messages

jturner65:
## Motivation and Context
This PR cleans up a lot of the "::nospace" directives in the various log messages - if any message has more than 2 of these, the log macro itself is directed to use the nospace flag instead.

This PR also introduces a timestamp to the log messages, with resolution to microsecond (using std::chrono::high_resolution_clock).

## How Has This Been Tested
C++ and python tests pass locally. Logging c++ test modified to validate output properly while ignoring timestamp.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/Logging.cpp
+++ b/src/esp/core/Logging.cpp
@@ -6,6 +6,7 @@
 #include "Check.h"
 
 #include <algorithm>
+#include <chrono>
 #include <cstdlib>
 
 #include <Corrade/Containers/Array.h>
@@ -132,9 +133,22 @@ Cr::Containers::String buildMessagePrefix(Subsystem subsystem,
                                           const std::string& function,
                                           int line) {
   auto baseFileName = Cr::Utility::Directory::filename(filename);
+
+  const auto timePassed =
+      std::chrono::high_resolution_clock::now().time_since_epoch();
+  const auto timePassedSeconds =
+      std::chrono::duration_cast<std::chrono::seconds>(timePassed);
+  // for micro time
+  const auto nowMicros = std::chrono::duration_cast<std::chrono::microseconds>(
+      timePassed - timePassedSeconds);
+  std::time_t t = std::time_t(timePassedSeconds.count());
+  // format hour,min, second
+  std::tm timeNow = *std::localtime(&t);
+
   return Cr::Utility::formatString(
-      "[{}] {}({})::{} : ", subsystemNames[uint8_t(subsystem)], baseFileName,
-      line, function);
+      "[{:.02d}:{:.02d}:{:.02d}:{:.06d}]:[{}] {}({})::{} : ", timeNow.tm_hour,
+      timeNow.tm_min, timeNow.tm_sec, nowMicros.count(),
+      subsystemNames[uint8_t(subsystem)], baseFileName, line, function);
 }
 
 }  // namespace logging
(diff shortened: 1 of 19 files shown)


src/esp/core/Logging.cpp
```diff
@@ -132,9 +133,22 @@ Cr::Containers::String buildMessagePrefix(Subsystem subsystem,
...
+      timePassed - timePassedSeconds);
+  std::time_t t = std::time_t(timePassedSeconds.count());
+  // format hour,min, second
+  std::tm timeNow = *std::localtime(&t);
+
   return Cr::Utility::formatString(
-      "[{}] {}({})::{} : ", subsystemNames[uint8_t(subsystem)], baseFileName,
-      line, function);
+      "[{:.02d}:{:.02d}:{:.02d}:{:.06d}]:[{}] {}({})::{} : ", timeNow.tm_hour,
+      timeNow.tm_min, timeNow.tm_sec, nowMicros.count(),
```
mosra, line 150: Nice, wasn't aware `formatString()` is capable of such things :smile:

src/tests/LoggingTest.cpp
```diff
@@ -52,34 +52,37 @@ struct LoggingTest : Cr::TestSuite::Tester {
 
 constexpr const struct {
   const char* envString;
-  const char* expected;
+  const char* dfltDebug;
+  const char* dfltWarning;
```
mosra, line 56:
```suggestion
  const char* defaultDebug;
  const char* defaultWarning;
```

Ws knd hrd t rd otherwise :sweat_smile:

src/tests/LoggingTest.cpp
```diff
@@ -95,18 +98,33 @@ void LoggingTest::envVarTest() {
   std::ostringstream out;
   Cr::Utility::Debug debugCapture{&out};
   Cr::Utility::Warning warnCapture{&out};
+  // use contains to bypass issue with timestamp
```
mosra, line 101: Though ... the test code you have now won't catch if the logger prints more than it should :thinking: 

Since the timestamp has a fixed-width formatting, what if you'd, instead of `contains()`, compare just the suffix after?

```cpp
  CORRADE_COMPARE(Cr::Containers::StringView{out.str()}.suffix(42),
                  Cr::Containers::StringView{data.defaultDebug});
```

etc., where `42` is the actual size of the timestamp prefix? (Lazy to count, sorry :P)

mosra (commented):
A nice tidying-up PR, thanks!

src/tests/LoggingTest.cpp
jturner65: Some of them are empty though, so the suffix pukes.

jturner65: I could compare with a \n

mosra: oh right, then ... a branch?

```cpp
  Cr::Containers::StringView actual = out.str();
  CORRADE_COMPARE(actual.suffix(actual.size() < 42 : 0 : 42),
                  Cr::Containers::StringView{data.defaultDebug});
```

A less hardcoded variant but with a worse diagnostic would be using `hasSuffix()` instead of `contains()`.

(Yes, goes straight to my TODOs, this desperately needs some "String has a suffix" comparator.)

jturner65: Looks like I got it by appending newlines, which are expected.

aclegg3 (approved):
LGTM, thanks!

PR #1584, opened 2021-12-02
--Scene Dataset default attributes filename wildcard handling.

jturner65:
## Motivation and Context
This PR adds support for a wildcard/sentinel value in the name of the render or collision asset for object attributes or stage attributes (or the navmesh, Semantic Scene asset or Semantic Scene Descriptor text file for stages) in the "default_attributes" tag of the Scene Dataset configuration file.  This value, "%%CONFIG_NAME_AS_ASSET_FILENAME%%" will be replaced by the name used to synthesize the attributes upon instantiation, so a stage- or object-specific value can be requested at the Scene Dataset level, in one spot.  Any value synthesized using this tag will be verified to actually exist before it is written (otherwise the field is cleared).   

With datasets like HM3D, where the data is organized such that a scene's assets and other pertinent information is stored in a single directory specific to that scene, this enables fairly complex handling of subordinate assets without individual per-scene stage configurations.  

All c++ and python tests pass.  Verified to work with Appen POC HM3D scene annotations.
## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/attributes/AttributesBase.h
+++ b/src/esp/metadata/attributes/AttributesBase.h
@@ -16,7 +16,17 @@ namespace asset {
 enum class AssetType;
 }
 namespace metadata {
+/**
+ * @brief A tag to search for in the default_attributes section of the Scene
+ * Dataset JSON configuration files denoting that an implementation of the
+ * attributes should replace this tag with the base filename (minus all paths
+ * and extensions)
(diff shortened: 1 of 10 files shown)


Skylion007 (approved):
Looks good to me. Also good job on making all the std::functions a const ref since they aren't trivially copyable.

PR #1586, opened 2021-12-06
Src python folder

erikwijmans:
## Motivation and Context

We should put habitat-sim's python source into a separate folder to stop people from accidentally importing the source when they installed via conda.

## How Has This Been Tested

The installs all still work and nothing (new) is broken!


Skylion007 (approved):

jturner65 (approved):
LGTM

PR #1590, opened 2021-12-08
[Attributes]--Add semantic mesh orientation support

jturner65:
## Motivation and Context
This PR introduces the ability to specify a frame to use for a semantic stage asset independent of the frame specified for the stage's render asset.  This is specifically needed for MP3D support, which has semantic plys that are orthogonal to the render glbs they correspond to.  Formerly, this was handled via hardcoded transforms, but these are incompatible with HM3D and other datasets with semantic meshes.

## How Has This Been Tested
Locally c++ and python tests pass.  C++ attributes tests have been expanded to explicitly test this functionality.

## Types of changes

## Checklist

### DIFF
--- a/docs/pages/attributesJSON.rst
+++ b/docs/pages/attributesJSON.rst
@@ -69,6 +69,12 @@ Below are the handles and descriptors for various mesh assets used by a stage.
 "semantic_asset"
     - string
     - The name of the file describing the stage's semantic mesh.
+"nav_asset"
+    - string
+    - The name of the file describing the stage's nav mesh.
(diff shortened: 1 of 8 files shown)


aclegg3 (approved):
Looks good, thanks. I guess we don't need semantic origin for now?

jturner65:
> Looks good, thanks. I guess we don't need semantic origin for now?

I considered that, but separate translation support for the semantic mesh would necessitate some substantial changes beyond those in the PR so far, and I'd prefer to possibly include that in the mesh loading pipeline unification PR that is coming up soon :D .

PR #1592, opened 2021-12-08
[HM3D]--Semantic scene support/bugfixes.

jturner65:
## Motivation and Context
This PR adds various functions and refactors required to implement support for HM3D semantic scenes in the correct format.  Included are : 
- Better vertex color handling in semantic mesh load
- More rigorous string verification for sentinel string denoting Semantic Scene Descriptor file type
- Proper application of frame re-orientation to Semantic Mesh data.  This is required to enable any semantic meshes to be re-oriented in engine.
- Force all semantic meshes to be treated as Instance Meshes (temporary so that vertex colors and vertex ids are supported).

Also, some refactors to improve efficiency and debug message content in Resource Manager are included.

## How Has This Been Tested
Locally c++ and python tests

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1270,8 +1270,6 @@ bool ResourceManager::loadRenderAssetIMesh(const AssetInfo& info) {
   const std::string& filename = info.filepath;
 
   CORRADE_INTERNAL_ASSERT(resourceDict_.count(filename) == 0);
-  ESP_DEBUG() << "Frame orientation :" << info.frame.toString()
-              << "instance mesh :" << filename;
 
   /* Open the file. On error the importer already prints a diagnostic message,
      so no need to do that here. The importer implicitly converts per-face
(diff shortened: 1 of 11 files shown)


jturner65:
@aclegg The tests in install_and_test_ubuntu will not pass until the mp3d example dataset is updated to have and use the scenedataset config.

src/esp/assets/ResourceManager.cpp
```diff
@@ -123,16 +124,26 @@ ResourceManager::~ResourceManager() {
 }
 
 void ResourceManager::buildImporters() {
+  // Preferred plugins, Basis target GPU format
+  importerManager_.setPreferredPlugins("GltfImporter", {"CgltfImporter"});
+#ifdef ESP_BUILD_ASSIMP_SUPPORT
+  importerManager_.setPreferredPlugins("ObjImporter", {"AssimpImporter"});
+  Cr::PluginManager::PluginMetadata* const assimpmetadata =
+      importerManager_.metadata("AssimpImporter");
+  assimpmetadata->configuration().setValue("ImportColladaIgnoreUpDirection",
```
Skylion007, line 133: Hmm, I wish there was a way to use this if it wasn't specified in the JSON...

jturner65: Could you explain what you mean? Thanks :)

Skylion007: I am worried the fact that other importers like Blender may support Collada's up direction may cause confusion for end-users when the rotations do not match. Seems like it's harmful to ignore an up direction defined in the dae. What about other formats that have their own up direction, it feels like we should try to be consistent with the handling of the up direction across formats. (For instance, the rotation in the dataset should be relative to the up direction in the mesh file).

aclegg3 (approved):
LGTM

PR #1593, opened 2021-12-09
Upgrade to Magnum's new SceneData API

mosra:
## Motivation and Context

- Updates the scene hierarchy import code to use the new, data-oriented, batch- and ECS-friendly [Trade::SceneData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1SceneData.html) class
  - Check the docs please! :)
  - :warning: Note: the `MeshTransformNode` doesn't support multiple assignments per mesh node but the new `SceneData` does, which might become problematic for certain assets
  - There's an internal Magnum tool for converting the scene to "single-functioning objects" which could be eventually used to fix this
  - However I'd rather see the `MeshTransformNode` eventually gone, superseded by `SceneData` itself
- Updates to Basis Universal `v1_15_update2` that includes UASTC and KTX support
  - As before, the files are bundled because the repo is *huge*, so the PR diff is rather busy (sorry)
  - Note: for now I disabled Zstd support in KTX containers as it's not yet used, will add Zstd as a submodule once we start using it
- Fixes to make code compiling with Eigen 3.4 (not updating the eigen submodule tho)
- Fixes to make code work with Assimp 5.1 (not updating the assimp submodule tho)
- Updates to Assimp regarding PBR materials and other things that'll become useful for Habitat in the future

## How Has This Been Tested

- Tested scene hierarchy import locally with the `viewer` utility
- CI is passing except for hab-lab build which seems to be broken on `main` as well

## Types of changes


aclegg3 (approved):
LGTM.

Skylion007 (approved):

src/esp/assets/ResourceManager.cpp
```diff
@@ -1463,27 +1481,75 @@ bool ResourceManager::loadRenderAssetGeneral(const AssetInfo& info) {
...
-    if (!sceneData) {
-      ESP_ERROR() << "Cannot load scene, exiting";
+  // no default scene --- standalone OBJ/PLY files, for example
+  // take a wild guess and load the first mesh with the first material
+  if (fileImporter_->defaultScene() == -1) {
+    if ((fileImporter_->meshCount() != 0u) &&
+        meshes_.at(meshMetaData.meshIndex.first)) {
+      // addMeshToDrawables(metaData, *parent, drawables, 0, 0);
+      CORRADE_ASSERT_UNREACHABLE(
+          "Sorry, this is broken since August 2019 and needs to get fixed", {});
```
mosra, line 1491: I wonder if I maybe should just fix this part instead :)

src/esp/assets/ResourceManager.cpp
```diff
@@ -1463,27 +1481,75 @@ bool ResourceManager::loadRenderAssetGeneral(const AssetInfo& info) {
...
+       scene->transformations3DAsArray()) {
+    if (Cr::Containers::Optional<esp::assets::MeshTransformNode>& node =
+            nodes[transformation.first()])
+      node->transformFromLocalToParent = transformation.second();
+  }
+
+  // Add mesh indices for objects that have a mesh, again ignoring nodes that
+  // are not part of the hierarchy.
+  // TODO: There can be multiple mesh assignments for one object, unfortunately
+  // MeshTransformNode is not capable of that
```
mosra, line 1532: To avoid silent breakages, I suppose it'd be good to at least print a warning here?

Skylion007: - Absolutely, sounds good.

mosra: 988e04c87e46e8f081c3fa1d69219c5d0bf18007

src/esp/assets/ResourceManager.cpp
mosra: c68dd6005f2d46504bdeb9fd9b9277e2e1d55a8e

PR #1618, opened 2022-01-11
[Bugfix]--Fix HM3D Semantic Object ID construction

jturner65:
## Motivation and Context
This PR fixes how the Semantic Object ID is constructed for the HM3D dataset.  Previously, it was constructing the Semantic Object ID to be "<category>_<int idx within category>" (i.e. "wall_1") but this was incorrect - the current standard in engine is to have the final value of the Semantic Object ID be a unique int across all categories, corresponding to the Semantic ID in the Scene Node.  This PR fixes this.  

## How Has This Been Tested
Locally c++ and python tests pass.

## Types of changes

## Checklist


aclegg3 (approved):
Looks good to me. We should add the PoC semantic scene to the test assets download so we can verify this stuff for future proofing. We have permission to distribute that one.

PR #1621, opened 2022-01-13
--Viewer : minor bugs; refactor drawEvent

jturner65:
## Motivation and Context
This PR addresses a few minor bugs in viewer, where values that should have been mapped from constants were being mapped via magic numbers, or else not being mapped at all (such as the agent's dims if navmesh recalc is requested).

drawEvent functionality was also refactored to not recreate the Sensor Key string every frame when it only changed if sensorMode_ or visualizeMode_ changed.

## How Has This Been Tested
It's Viewer! It Always Works!

## Types of changes

## Checklist

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -979,6 +979,7 @@ void Viewer::initSimPostReconfigure() {
     esp::nav::NavMeshSettings navMeshSettings;
     navMeshSettings.agentHeight = agentConfig_.height;
     navMeshSettings.agentRadius = agentConfig_.radius;
+    navMeshSettings.agentMaxSlope = 55.0f;
     simulator_->recomputeNavMesh(*simulator_->getPathFinder().get(),
                                  navMeshSettings, true);
   } else if (!navmeshFilename_.empty()) {
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -611,7 +611,7 @@ void addSensors(esp::agent::AgentConfiguration& agentConfig, bool isOrtho) {
         sensorType == esp::sensor::SensorType::Semantic) {
       spec->channels = 1;
     }
-    spec->position = {0.0f, 1.5f, 0.0f};
+    spec->position = {0.0f, rgbSensorHeight, 0.0f};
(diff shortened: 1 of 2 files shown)


src/utils/viewer/viewer.cpp
```diff
@@ -903,6 +903,9 @@ void Viewer::initSimPostReconfigure() {
     }
   } else if (recomputeNavmesh_) {
     esp::nav::NavMeshSettings navMeshSettings;
+    navMeshSettings.agentHeight = agentConfig_.height;
+    navMeshSettings.agentRadius = agentConfig_.radius;
+    navMeshSettings.agentMaxSlope = 55.0f;
```
aclegg3, line 908: Danger here is that users will load the asset in viewer and everything looks good, but using the defaults in the episode generator for training won't produce the same results.

aclegg3 (commented):
Agreed that settings should come from any loaded config. Not sure about the hard-coded slope fix here. I understand why we should change it, but maybe it should be a change in the NavMesh default settings instead.

Also note we have the python viewer application [here](https://github.com/facebookresearch/habitat-sim/blob/08be9bd97e6b6d13ee4d50e893b51a7397570321/examples/viewer.py#L541). Should these be expected to behave the same way? Open to thoughts here.

jturner65:
> Agreed that settings should come from any loaded config. Not sure about the hard-coded slope fix here. I understand why we should change it, but maybe it should be a change in the NavMesh default settings instead.

Yeah, that would be the ideal situation, but not sure if folks are prepared for any changes that might entail.

> Also note we have the python viewer application [here](https://github.com/facebookresearch/habitat-sim/blob/08be9bd97e6b6d13ee4d50e893b51a7397570321/examples/viewer.py#L541). Should these be expected to behave the same way? Open to thoughts here.

I think they should, especially for these particular value-driven quantities (if we are recomputing the navmesh and have locally defined agent specs, we really should use those to build the navmesh, I think)

aclegg3 (approved):
LGTM, thanks.

Skylion007 (approved):

jturner65:
Got rid of a bunch of needless if checks building the string sensor key in drawEvent, and streamlined the draw/render process a bit.

PR #1632, opened 2022-01-20
--Bug fixes: Navmesh building and reconfigure in viewer.py

jturner65:
## Motivation and Context
This small PR fixes 2 bugs : 

- Viewer.py : Update the locally maintained scene name to be the unique, minimal, name of the currently loaded scene. This is necessary if the user requested a scene by some non-unique substring of a scene name (in which case the first match would be the scene rendered)
- Do not build a navmesh on a joined collision mesh devoid of vertices.  **If this is attempted, the process will assert.** Reviewers, please share if you feel this is the appropriate response.

## How Has This Been Tested
Locally all python and c++ tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/sim/Simulator.cpp
+++ b/src/esp/sim/Simulator.cpp
@@ -795,9 +795,8 @@ bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
                                  bool includeStaticObjects) {
   CORRADE_ASSERT(config_.createRenderer,
                  "::recomputeNavMesh: "
-                 "SimulatorConfiguration::createRenderer is "
-                 "false. Scene geometry is required to recompute navmesh. No "
-                 "geometry is "
+                 "SimulatorConfiguration::createRenderer is false. Scene "
+                 "geometry is required to recompute navmesh. No geometry is "
                  "loaded without renderer initialization.",
                  false);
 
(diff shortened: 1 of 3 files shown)


src/esp/sim/Simulator.cpp
```diff
@@ -888,6 +887,13 @@ bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
       }
     }
   }
+  CORRADE_ASSERT(
+      joinedMesh->vbo.size() > 0,
+      "::recomputeNavMesh: "
+      "Unable to compute a navmesh upon a non-existent mesh - "
+      "the underlying joined collision mesh has no vertices. This is "
+      "probably due to the current scene being NONE.",
+      false);
```
jturner65, line 896: Do we want to have this assert if the joinedMesh is empty? Or just fail? This will probably not occur often - would primarily happen if someone attempts to build a navmesh on the NONE scene, so just failing without the assertion would probably be fine, but I am not sure if there are possible situations where an assertion would be best (i.e. training and somehow an expected collision mesh is empty).

src/esp/sim/Simulator.cpp
```diff
@@ -888,6 +887,11 @@ bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
       }
     }
   }
+  ESP_CHECK(joinedMesh->vbo.size() > 0,
+            "::recomputeNavMesh: "
+            "Unable to compute a navmesh upon a non-existent mesh - "
+            "the underlying joined collision mesh has no vertices. This is "
+            "probably due to the current scene being NONE. Aborting");
```
jturner65, line 894: ...or is it better to have ESP_CHECK for this, which will throw in python and can be trapped in a try if necessary.

aclegg3: I like this approach, thanks for adding this check. I've seen the resulting issue a couple times and the current crash message response is non-obvious.

aclegg3 (approved):
LGTM

Issue #1641, opened 2022-01-28
Constrained look only works for a singular look action type


lucasosouza:
## Habitat-Sim version
v0.2.1

Habitat is under active development, and we advise users to restrict themselves to stable releases. Are you using the latest [release version](https://github.com/facebookresearch/habitat-sim/releases) of Habitat-Sim? Your question may already be addressed in the latest version. We may also not be able to help with problems in earlier versions because they sometimes lack the more verbose logging needed for debugging.

Main branch contains 'bleeding edge' code and should be used at your own risk.

## Docs and Tutorials
Did you read the docs? https://aihabitat.org/docs/habitat-sim/

Did you check out the tutorials? https://aihabitat.org/tutorial/2020/

Perhaps your question is answered there. If not, carry on!

## ❓ Questions and Help

In `_rotate_local` there is a check to only allow constraints to be defined for a "singular look action type":

```
    if constraint is not None:
        rotation = scene_node.rotation

        if abs(float(rotation.angle())) > 0:
            ref_vector = mn.Vector3()
            ref_vector[axis] = 1

            if mn.math.angle(ref_vector, rotation.axis().normalized()) > mn.Rad(1e-3):
                raise RuntimeError(
                    "Constrained look only works for a singular look action type"
                )
```

What is the rationale behind this if block? We are trying to implement a camera that can only rotate 90deg in either direction. There is a check in that same function that ensures whatever constraint defined in the actuation spec is observed, so the look_angle never goes beyond -90 or +90. But if we implement constraints for both look_left and look_right actions, look_left will work, while look_right will raise a runtime error once the camera's rotation angle is > 0 due to the exception defined above. 

We just started working with habitat_sim (awesome library), so apologies in advance if it has a simple explanation.

erikwijmans:
Looks like I missed an `abs` when writing that check. That check is there because the code doesn't handle constraining rotations with rotations along multiple axes because that gets tricky due to chained rotations.

lucasosouza:
Thanks Erik. That makes sense. Where would the missing `abs` go? It seems to me it should be applied to the rotation axis used in the assert. This works but I am wondering if there is a simpler version: `mn.Vector3(np.abs(rotation.axis().normalized()))`

erikwijmans:
I was thinking `abs(float(mn.math.angle(ref_vector, rotation.axis().normalized())))` but that works too

PR #1645, opened 2022-02-01
Fixes bug that only allow constraints to be set in one direction of a singular axis

lucasosouza:
## Motivation and Context

Fixes bug that only allow constraints to be set in one direction of a singular axis

https://github.com/facebookresearch/habitat-sim/issues/1641

## How Has This Been Tested

It is a bug fix, discussed in the issue linked above. We have applied it already to our local implementation of habitat_sim. 

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ x] I have read the **CONTRIBUTING** document.
- [ x ] I have completed my CLA (see **CONTRIBUTING**)
- [ x ] All new and existing tests passed.

### DIFF
--- a/src_python/habitat_sim/agent/controls/default_controls.py
+++ b/src_python/habitat_sim/agent/controls/default_controls.py
@@ -42,8 +42,9 @@ def _rotate_local(
         if abs(float(rotation.angle())) > 0:
             ref_vector = mn.Vector3()
             ref_vector[axis] = 1
+            abs_rotation_axis = mn.Vector3(np.abs(rotation.axis().normalized()))
 
-            if mn.math.angle(ref_vector, rotation.axis().normalized()) > mn.Rad(1e-3):
+            if mn.math.angle(ref_vector, abs_rotation_axis) > mn.Rad(1e-3):
                 raise RuntimeError(
                     "Constrained look only works for a singular look action type"
                 )
--- a/src_python/habitat_sim/agent/controls/default_controls.py
+++ b/src_python/habitat_sim/agent/controls/default_controls.py
@@ -39,11 +39,13 @@ def _rotate_local(
     if constraint is not None:
         rotation = scene_node.rotation
 
-        if (abs(float(rotation.angle())) > 0
-            and 1.0 - abs(rotation.axis().normalized()[axis]) > 1e-3):
-                raise RuntimeError(
-                    "Constrained look only works for a singular look action type"
-                )
+        if (
+            abs(float(rotation.angle())) > 0
+            and 1.0 - abs(rotation.axis().normalized()[axis]) > 1e-3
+        ):
+            raise RuntimeError(
+                "Constrained look only works for a singular look action type"
+            )
 
         look_vector = rotation.transform_vector(FRONT)
         if axis == 0:
--- a/src_python/habitat_sim/agent/controls/default_controls.py
+++ b/src_python/habitat_sim/agent/controls/default_controls.py
@@ -40,11 +40,7 @@ def _rotate_local(
         rotation = scene_node.rotation
 
         if abs(float(rotation.angle())) > 0:
-            ref_vector = mn.Vector3()
-            ref_vector[axis] = 1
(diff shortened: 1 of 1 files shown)


src_python/habitat_sim/agent/controls/default_controls.py
```diff
@@ -42,8 +42,9 @@ def _rotate_local(
         if abs(float(rotation.angle())) > 0:
             ref_vector = mn.Vector3()
             ref_vector[axis] = 1
+            abs_rotation_axis = mn.Vector3(np.abs(rotation.axis().normalized()))
 
-            if mn.math.angle(ref_vector, rotation.axis().normalized()) > mn.Rad(1e-3):
```
Skylion007, line 46:
```suggestion
            if mn.math.angle(ref_vector, mn.math.abs(abs_rotation_axis)) > mn.Rad(1e-3):
```

Skylion007: Actually never mind, this function isn't exposed on the magum-bindings. Ping @mosra

Skylion007 (approved):

src_python/habitat_sim/agent/controls/default_controls.py
lucasosouza: It looks cleaner, but I don't see mn.math.abs in my habitat_sim environment. I might have the wrong version: 
```
>>> dir(mn.math)
['__doc__', '__loader__', '__name__', '__package__', '__spec__', 'acos', 'angle', 'asin', 'atan', 'cos', 'cross', 'dot', 'e', 'inf', 'intersect', 'intersects', 'join', 'lerp', 'lerp_shortest_path', 'nan', 'pi', 'pi_half', 'pi_quarter', 'sin', 'sincos', 'slerp', 'slerp_shortest_path', 'sqrt2', 'sqrt3', 'sqrt_half', 'tan', 'tau']
```

Skylion007: You are right, those functions aren't include in the Python binding. Does @erikwijmans `abs(float(mn.math.angle(ref_vector, rotation.axis().normalized())))` diff work? Or is there some subtle missing there? That would be marginally more efficient (since we don't have to reconstruct a mn.Vector3 from a numpy array).

Skylion007: Err, wait that won't work nicely either since you need to back into a mn.Rad. I think the original way you did it might be the best until the Magnum bindings are updated.

mosra: Quite an essential omission, sorry -- I'll add this.

Skylion007: @lucasosouza ^

mosra: As of mosra/magnum-bindings@0120b3f76895efb71f38043638be2f78f31c4ade, the python bindings have [math.abs()](https://doc.magnum.graphics/python/magnum/math/#abs-3776c) alongside other missing scalar/vector functions. Got delayed a bit because the whole world somehow decided to drop Python 3.6 support this week.

src_python/habitat_sim/agent/controls/default_controls.py
```diff
@@ -39,14 +39,13 @@ def _rotate_local(
...
-            ref_vector = mn.Vector3()
-            ref_vector[axis] = 1
-
-            if mn.math.angle(ref_vector, rotation.axis().normalized()) > mn.Rad(1e-3):
-                raise RuntimeError(
-                    "Constrained look only works for a singular look action type"
-                )
+        if (
+            abs(float(rotation.angle())) > 0
+            and 1.0 - abs(rotation.axis().normalized()[axis]) > 1e-3
```
Skylion007, line 44: Might want to add a comment that this is in Radians

Skylion007 (approved):

Skylion007 (approved):

PR #1648, opened 2022-02-02
Fix performance issues with calc MOBB

Skylion007:
## Motivation and Context

This fixes 2 minor performance issues. 
1.  is that all std::vectors in this algorithm are of known sizes at runtime and should reserve the appropiate amount of heap space for their calculation
1. There is an std::vector that was being treated more as a tuple. I replaced it with an std::array since the size is known at compile time.

## How Has This Been Tested
* PyTest and CI

## Types of changes

## Checklist

### DIFF
--- a/src/esp/geo/OBB.cpp
+++ b/src/esp/geo/OBB.cpp
@@ -4,6 +4,7 @@
 
 #include "OBB.h"
 
+#include <array>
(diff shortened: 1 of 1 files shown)


jturner65 (approved):
LGTM

PR #1653, opened 2022-02-03
Pyupgrade to modern typing syntax and enable pyupgrade pre-commit hook

Skylion007:
## Motivation and Context
* This upgrades our type annotation to modern Python syntax. It's a lot more succinct and removes a lot of typing imports which may potentially allow faster startups. This may break some old runtime type checkers, but we don't use TypeGuard at all anymore and mypy has full support. 
* For instance '|' can now replace Union[] and Optional etc.
* This enables a future flag which will be the default in 3.11 anyhow so might as well get the code written
* This also adds a pre-commit hook called pyupgrade which ensures we are using the latest syntax available for 3.7 and other best practices. 
* I also added an isort config which adds the __future__ import for annotations to all python files which do at least one import.

## How Has This Been Tested
* locally with pre-commit and pytest

## Types of changes

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,6 +60,13 @@ repos:
     -   id: black
         exclude: ^examples/tutorials/(nb_python|colabs)
 
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v2.31.0
+    hooks:
+    -   id: pyupgrade
+        args: ["--py37-plus"]
+        exclude: "^setup.py$"
+
 -   repo: https://github.com/myint/autoflake
     rev: v1.4
     hooks:
@@ -92,6 +99,8 @@ repos:
         pass_filenames: false
         additional_dependencies:
         - attrs
+        - numpy
+        - numpy-quaternion
 
 -   repo: https://github.com/seddonym/import-linter
     rev: v1.2.6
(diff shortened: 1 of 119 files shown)


Skylion007:
@mosra This change seems blocked by m.css support?

mosra:
Nope, m.css just passes through a typing exception.

The issue is on your side, [PEP604](https://www.python.org/dev/peps/pep-0604/) (`|` instead of `Union`) is only since Python 3.10.

Skylion007:
@mosra Okay so interesting, it's not 'SUPPORTED' per say, but mypy and stuff can parse all those fields out fine as long as delayed annotations is enabled (since they don't bother the interpreter until runtime)'.

Skylion007:
Ah
```PEP 585 normally requires Python 3.9 and PEP 604 would require 3.10. The use in versions prior to that, the current default is 3.8, is only possible with postponed evaluation of type annotations, [PEP 563](https://www.python.org/dev/peps/pep-0563/). This can be enabled starting in 3.7 with from __future__ import annotations. However this workaround also means that any dynamic evaluation will throw an error. Some common examples are the use of typing.get_type_hints or eval.```

mosra:
Yes, but *I'm* relying on the interpreter being able to understand the types, so I can link them etc. So if you run m.css on 3.7, it'll fail. You'd have to run m.css on 3.10.

I'd say let's just postpone changes that require 3.10 until 3.10 is our minimum version.

jturner65 (approved):
LGTM. Very cool that the type annotation support is so easily accessed now

aclegg3 (approved):
LGTM

PR #1656, opened 2022-02-06
Pathfinder sample near point

Skylion007:
## Motivation and Context

* This allow the efficient sampling of points of the navmesh within a specifeid radius of an existing point on the navmesh. This is useful as it allows the user to control the region the points are sampled from within the Navmesh.
* I also applied some clang-tidy fixits to make this file happy and a few other miscellaneous style fixes.

## How Has This Been Tested
* Locally with pytest. All existing tests pass.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/ShortestPathBindings.cpp
+++ b/src/esp/bindings/ShortestPathBindings.cpp
@@ -82,7 +82,7 @@ void initShortestPathBindings(py::module& m) {
            "max_tries"_a = 10)
       .def("get_random_navigable_point_near",
            &PathFinder::getRandomNavigablePointAroundSphere, "circle_center"_a,
-           "radius"_a, "max_tries"_a = 10)
+           "radius"_a, "max_tries"_a = 100)
       .def("find_path", py::overload_cast<ShortestPath&>(&PathFinder::findPath),
            "path"_a)
       .def("find_path",
--- a/src/esp/bindings/ShortestPathBindings.cpp
+++ b/src/esp/bindings/ShortestPathBindings.cpp
@@ -80,6 +80,8 @@ void initShortestPathBindings(py::module& m) {
            "meters_per_pixel"_a, "height"_a)
       .def("get_random_navigable_point", &PathFinder::getRandomNavigablePoint,
            "max_tries"_a = 10)
+      .def("get_random_navigable_point_near",
+           &PathFinder::getRandomNavigablePointAroundSphere)
       .def("find_path", py::overload_cast<ShortestPath&>(&PathFinder::findPath),
            "path"_a)
       .def("find_path",
--- a/src/esp/bindings/ShortestPathBindings.cpp
+++ b/src/esp/bindings/ShortestPathBindings.cpp
@@ -81,7 +81,8 @@ void initShortestPathBindings(py::module& m) {
       .def("get_random_navigable_point", &PathFinder::getRandomNavigablePoint,
            "max_tries"_a = 10)
       .def("get_random_navigable_point_near",
-           &PathFinder::getRandomNavigablePointAroundSphere)
+           &PathFinder::getRandomNavigablePointAroundSphere, "circle_center"_a,
+           "radius"_a, "max_tries"_a = 10)
       .def("find_path", py::overload_cast<ShortestPath&>(&PathFinder::findPath),
            "path"_a)
       .def("find_path",
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -186,7 +186,7 @@ class IslandSystem {
 
       for (int iVert = 0; iVert < poly->vertCount; ++iVert) {
         islandVerts.emplace_back(Eigen::Map<vec3f>(
-            &tile->verts[static_cast<ptrdiff_t>(poly->verts[iVert]*) 3]));
+            &tile->verts[static_cast<ptrdiff_t>(poly->verts[iVert]) * 3]));
       }
 
       // Iterate over all neighbours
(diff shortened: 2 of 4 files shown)


src/esp/nav/PathFinder.cpp
```diff
@@ -946,9 +953,46 @@ vec3f PathFinder::Impl::getRandomNavigablePoint(const int maxTries /*= 10*/) {
...
+    const float radius,
+    const int maxTries) {
+  if (getNavigableArea() <= 0.0)
+    throw std::runtime_error(
+        "NavMesh has no navigable area, this indicates an issue with the "
+        "NavMesh");
+
+  vec3f pt = vec3f::Constant(Mn::Constants::nan());
+  dtPolyRef start_ref = 0;  // ID to start our search
+  dtStatus status = navQuery_->findNearestPoly(
```
Skylion007, line 971: This does have some edge cases with spheres that span multiple floors, but I am not sure the best way to resolve those. (It looks like it does a random selection from a search of all points discovered that fit into the circle from the starting ref).

src/esp/nav/PathFinder.cpp
```diff
@@ -946,9 +953,46 @@ vec3f PathFinder::Impl::getRandomNavigablePoint(const int maxTries /*= 10*/) {
...
+        start_ref, circleCenter.data(), radius, filter_.get(), frand, &rand_ref,
+        pt.data());
+    if (dtStatusSucceed(status) && (pt - circleCenter).norm() <= radius) {
+      break;
+    }
+  }
+  if (i == maxTries) {
+    ESP_ERROR() << "Failed to getRandomNavigablePoint.  Try increasing max "
+                   "tries if the navmesh is fine but just hard to sample from";
+    return vec3f::Constant(Mn::Constants::nan());
```
Skylion007, line 993: Also wasn't sure if I should just return the nearest point here if the "random" selection of neighboring points failed for some reason. I figured to return a NaN just to be consistent even if it  does find a valid point (AKA the nearest point only).

erikwijmans (commented):
Some comments on code.

One idea for an alternative implementation of the function, albeit more code: You could do similar sampling as `findRandomPoint` but only select tiles and points that have point within the sphere. You may have to do some rejection sampling tho and it may be slower, but it wouldn't have the edge case you described.

Skylion007:
@erikwijmans How would I select the tiles / points within the sphere? Calling randomPoint() over and over again is not performant as @srama2512 has noted, this is an alternative that would be. If I could select a point within the sphere, than I could also solve this by choosing a random startref instead of using findNearestPolygon.

erikwijmans:
I'd iterate over all polys and then check if that poly has any vert within the sphere. If it does, then that poly is valid to be sampled (you can do reservoir sampling to avoid constructing that list explicitly). Then given your sampled poly, you can sample a random point on that poly (you can see how Detour does that here: https://github.com/erikwijmans/recastnavigation/blob/8eef00be6003a6bf9c49cb4c126662e05e58d508/Detour/Source/DetourNavMeshQuery.cpp#L286-L310)

This only gives you a high chance of sampling a point within the sphere, not a guarantee, so you will have to run the above procedure a few times. If the sphere is pretty big, it'll get something quickly. If it's small it may take some tries.

One thing to note: `dtNavMeshQuery::findRandomPointAroundCircle` is not guaranteed to sample a point within the circle, as the point may be slightly outside the circle.

Skylion007:
> One thing to note: dtNavMeshQuery::findRandomPointAroundCircle is not guaranteed to sample a point within the circle, as the point may be slightly outside the circle, yeah we reject those points in the num_tries loop.

src/esp/bindings/ShortestPathBindings.cpp
```diff
@@ -80,6 +80,9 @@ void initShortestPathBindings(py::module& m) {
            "meters_per_pixel"_a, "height"_a)
       .def("get_random_navigable_point", &PathFinder::getRandomNavigablePoint,
            "max_tries"_a = 10)
+      .def("get_random_navigable_point_near",
+           &PathFinder::getRandomNavigablePointAroundSphere, "circle_center"_a,
+           "radius"_a, "max_tries"_a = 10)
```
jturner65, line 85: Maybe a helptext describing what's going on here? :)

jturner65:
> I'd iterate over all polys and then check if that poly has any vert within the sphere. If it does, then that poly is valid to be sampled (you can do reservoir sampling to avoid constructing the list of valid polys explicitly). Then given your sampled poly, you can sample a random point on that poly.

- Would there not be a concern for over-representing points in small polys? Or will the reservoir sampling process weight the polys considered based on their area?
- Is there no corner-table-type representation of the mesh available to provide adjacency info? If not, would adding this be worthwhile?

erikwijmans:
Yeah, you can use the poly area to do weighted reservoir sampling.

If you wanted to do better than just searching over all polys, you could use `dtNavMeshQuery::queryPolygons` to get everything with the sphere (in a similar fashion to`dtNavMeshQuery::findNearestPoly`). That may be quicker.

Skylion007:
Seems like the default of 10 tries does sometimes fail on the test_scene navmeshes, I guess we should raise the defaults. 100 seems to work fine.

src/esp/bindings/ShortestPathBindings.cpp
Skylion007: I don't want this to be that public of a function, yet, but I can add a docstring.

jturner65 (approved):
LGTM. Very cool functionality.

Skylion007:
@erikwijmans @aclegg3 going to merge this if there aren't any objections.

PR #1658, opened 2022-02-08
--Semantic texture support

jturner65:
## Motivation and Context
This PR provides support for semantically annotated textures on meshes. Bounding boxes are still calculated using vertex annotations, but the actual semantic annotations are derived from textures.

To enable texture-based semantics both of the following settings must be set to true
 - The dataset used must support texture semantic annotations (currently only HM3D) - the scene dataset config or stage config must have the field ` "has_semantic_textures": true`
 - SimulatorConfiguration /sim_config has a field "use_semantic_textures" (in python) that needs to be set to true.
 
 Loading of texture-based semantics is currently approx. 15-20% slower than loading the same scene for vertex annotation.  This primarily caused by the online conversion from semantic color textures to integer semantic id textures. These will eventually be precomputed.

TODO : test case to verify integrity of annotation.

## How Has This Been Tested
Locally all c++ and python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/GenericSemanticMeshData.cpp
+++ b/src/esp/assets/GenericSemanticMeshData.cpp
@@ -141,37 +141,22 @@ GenericSemanticMeshData::buildSemanticMeshData(
 
       // map of colors as ints as key, where value is objectID and regionID from
       // ssd
-      std::unordered_map<uint32_t, std::pair<int, int>>
-          tmpColorMapToSSDidAndRegionIndex;
-      tmpColorMapToSSDidAndRegionIndex.reserve(numSSDObjs);
-      // lambda function to convert a 3-vec representation of a color into an
-      // int
-      auto colorAsInt = [](const Mn::Color3ub& color) -> uint32_t {
-        return (unsigned(color[0]) << 16) | (unsigned(color[1]) << 8) |
-               unsigned(color[2]);
-      };
+      const std::unordered_map<uint32_t, std::pair<int, int>>
+          tmpColorMapToSSDidAndRegionIndex =
+              semanticScene->getSemanticColorToIdAndRegionMap();
 
-      // build maps of color ints to semantic IDs and color ints to
-      // region ids in SSD find max regions present, so we can have an
-      // "unassigned"/"unknown" region to move all verts whose region is
+      // find max regions present, so we can have a specific region to move all
+      // verts whose region/color is "unassigned"/"unknown"
       int maxRegion = -1;
       for (int i = 0; i < numSSDObjs; ++i) {
-        const auto& ssdObj = ssdObjs[i];
-        const uint32_t colorInt = colorAsInt(ssdObj->getColor());
-        int semanticID = ssdObj->semanticID();
-        int regionIDX = ssdObj->region()->getIndex();
-        tmpColorMapToSSDidAndRegionIndex[colorInt] = {semanticID, regionIDX};
-        maxRegion = Mn::Math::max(regionIDX, maxRegion);
+        maxRegion = Mn::Math::max(ssdObjs[i]->region()->getIndex(), maxRegion);
       }
-      // largest possible known semanticID will be size of colorMapToUse at this
-      // point -1 due to idx 0 not being a valid map.  If unknown/unmapped
-      // colors are present in mesh, colorMapToUse will grow to hold them in
-      // subsequent step.
-      // 1st semantic ID for colors not found in SSD objects ==> 1 more than the
-      // current highest id in the colormap
+      // If unknown/unmapped colors are present in mesh, colorMapToUse will grow
+      // to hold them in subsequent step. 1st semantic ID for colors not found
+      // in SSD objects ==> 1 more than the current highest id in the colormap
       std::size_t nonSSDObjID = colorMapToUse.size();
 
-      // increment maxRegion to use largest value as unknown region for objects
+      // increment maxRegion to use as value for unknown region for objects
       // whose colors are not mapped in given ssd data.
       ++maxRegion;
 
@@ -199,7 +184,9 @@ GenericSemanticMeshData::buildSemanticMeshData(
       for (int vertIdx = 0; vertIdx < numVerts; ++vertIdx) {
         Mn::Color3ub meshColor = meshColors[vertIdx];
         // Convert color to an int @ vertex
-        const uint32_t meshColorInt = colorAsInt(meshColor);
+        const uint32_t meshColorInt = (unsigned(meshColor[0]) << 16) |
+                                      (unsigned(meshColor[1]) << 8) |
+                                      unsigned(meshColor[2]);
 
         std::unordered_map<uint32_t, std::pair<int, int>>::const_iterator
             ssdColorToIDAndRegionIter =
--- a/src/esp/assets/GenericSemanticMeshData.cpp
+++ b/src/esp/assets/GenericSemanticMeshData.cpp
@@ -154,24 +154,13 @@ GenericSemanticMeshData::buildSemanticMeshData(
       // build maps of color ints to semantic IDs and color ints to
       // region ids in SSD find max regions present, so we can have an
       // "unassigned"/"unknown" region to move all verts whose region is
-
       int maxRegion = -1;
-      // build the color map with first maxSemanticID elements in proper order
-      // to match provided semantic IDs (so that ID is IDX of semantic color in
-      // map) and any overflow colors uniquely map 1-to-1 to an unmapped
-      // semantic ID as their index.
-      colorMapToUse.resize(numSSDObjs);
       for (int i = 0; i < numSSDObjs; ++i) {
         const auto& ssdObj = ssdObjs[i];
-        const Mn::Vector3ub ssdColor = ssdObj->getColor();
-        const uint32_t colorInt = colorAsInt(ssdColor);
+        const uint32_t colorInt = colorAsInt(ssdObj->getColor());
         int semanticID = ssdObj->semanticID();
         int regionIDX = ssdObj->region()->getIndex();
         tmpColorMapToSSDidAndRegionIndex[colorInt] = {semanticID, regionIDX};
-        if (colorMapToUse.size() <= semanticID) {
-          colorMapToUse.resize(semanticID + 1);
-        }
-        colorMapToUse[semanticID] = ssdColor;
         maxRegion = Mn::Math::max(regionIDX, maxRegion);
       }
       // largest possible known semanticID will be size of colorMapToUse at this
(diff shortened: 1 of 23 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -659,6 +712,23 @@ scene::SceneNode* ResourceManager::loadAndCreateRenderAssetInstance(
...
+  std::string modifiedAssetName = info.filepath;
+
+  // check materialId
+  if (info.overridePhongMaterial != Cr::Containers::NullOpt) {
+    if (materialId.empty()) {
+      // if passed value is empty, synthesize new color and new materialId
+      // based on this color and values specified in info
+      materialId = createColorMaterial(*info.overridePhongMaterial);
+    }
+    modifiedAssetName += "?" + materialId;
```
Skylion007, line 726:
```suggestion
    modifiedAssetName += '?' + materialId;
```

src/esp/assets/GenericSemanticMeshData.cpp
```diff
@@ -301,64 +269,68 @@ GenericSemanticMeshData::buildSemanticMeshData(
...
+                                        semanticScene->objects(), dbgMsgPrefix);
   }
 
+  ////
+  return semanticMeshData;
+}
+
+std::vector<std::unique_ptr<GenericSemanticMeshData>>
+GenericSemanticMeshData::partitionSemanticMeshData(
+    std::unique_ptr<GenericSemanticMeshData>& semanticMeshData) {
```
mosra, line 303: Why does this take the `semanticMeshData` as a mutable argument? I don't see the function modifying it inside, so it could be `const&` instead.

Moreover -- similarly to my usual complaints about `Optional` overuse -- why the pointer wrapper? Just `const GenericSemanticMeshData&` would be enough as well, that way you also don't need to handle the case when a null pointer would get passed. Similarly, I think on the output it could also be just `std::vector<GenericSemanticMeshData>` -- unless there's something really specific done in the class, it should be relatively cheap to move.

src/esp/assets/GenericSemanticMeshData.cpp
```diff
@@ -25,11 +25,10 @@ namespace Mn = Magnum;
 namespace esp {
 namespace assets {
 
-std::vector<std::unique_ptr<GenericSemanticMeshData>>
+std::unique_ptr<GenericSemanticMeshData>
```
mosra, line 28: Yes, this is a pre-existing condition, but since you're modifying this part of the code anyway -- here I think the `unique_ptr` wouldn't need to be either.

src/esp/assets/ResourceManager.cpp
```diff
@@ -284,6 +284,40 @@ bool ResourceManager::loadSemanticSceneDescriptor(
...
+  semanticColorAsInt_.clear();
+  semanticColorAsInt_.reserve(semanticColorMapBeingUsed_.size());
+  // build listing of colors as ints, with idx being semantic ID
+  std::transform(semanticColorMapBeingUsed_.cbegin(),
+                 semanticColorMapBeingUsed_.cend(),
+                 std::back_inserter(semanticColorAsInt_),
+                 [](const Mn::Color3ub& color) -> uint32_t {
+                   return (unsigned(color[0]) << 16) |
+                          (unsigned(color[1]) << 8) | unsigned(color[2]);
+                 });
```
mosra, line 318: Recurring TODO for me: finally provide a builtin utility to make such nightmare code unnecessary :scream:

src/esp/assets/ResourceManager.cpp
```diff
@@ -384,6 +422,12 @@ bool ResourceManager::loadStage(
   if (!isSeparateSemanticScene) {
     flags |= RenderAssetInstanceCreationInfo::Flag::IsSemantic;
   }
+  // if texture-based semantic mesh specify in creation
+  // TODO: Currently do not support single mesh texture-based semantic and
+  // render stages
+  // if (renderInfo.hasSemanticTextures) {
+  //   flags |= RenderAssetInstanceCreationInfo::Flag::IsTextureBasedSemantic;
+  // }
```
mosra, line 430: Just so I know, what's blocking this? As far as my understanding of the code here goes, this should be possible...

src/esp/assets/ResourceManager.cpp
```diff
@@ -2159,87 +2312,185 @@ void ResourceManager::loadMeshes(Importer& importer,
...
+
+  Cr::Containers::StridedArrayView2D<const Mn::Color3ub> input =
+      srcImage.pixels<Mn::Color3ub>();
+  Cr::Containers::StridedArrayView2D<Mn::UnsignedShort> output =
+      resImage.pixels<Mn::UnsignedShort>();
+  for (std::size_t y = 0; y != size.y(); ++y) {
+    Cr::Containers::StridedArrayView1D<const Mn::Color3ub> inputRow = input[y];
+    Cr::Containers::StridedArrayView1D<Mn::UnsignedShort> outputRow = output[y];
+    for (std::size_t x = 0; x != size.x(); ++x) {
+      /* Fugly. Sorry. Needs better API on Magnum side. */
```
mosra, line 2336: The comment belongs to the colorInt creation below ;)

src/esp/assets/ResourceManager.cpp
```diff
@@ -2159,87 +2312,185 @@ void ResourceManager::loadMeshes(Importer& importer,
...
+      const Mn::Color3ub color = inputRow[x];
+
+      const Mn::UnsignedInt colorInt =
+          color.r() << 16 | color.g() << 8 | color.b();
+      outputRow[x] = clrToSemanticId[colorInt];
+    }
+  }
+  // upload the `integer` image to an integer Texture2D and then use as
+  // a semantic texture in the shader
```
mosra, line 2345: :eyes: ... feels like these `//` comments should have been removed from this function, it makes no sense here

Same below, heh.

src/esp/bindings/SimBindings.cpp
```diff
@@ -52,6 +52,11 @@ void initSimBindings(py::module& m) {
           R"(See tutorials/async_rendering.py)")
       .def_readwrite("frustum_culling", &SimulatorConfiguration::frustumCulling)
       .def_readwrite("enable_physics", &SimulatorConfiguration::enablePhysics)
+      .def_readwrite(
+          "use_semantic_textures",
+          &SimulatorConfiguration::useSemanticTexturesIfFound,
+          R"(If the loaded scene/dataset supports semantically annotated textures, use these for
+      semantic rendering.)")
```
mosra, line 59: Same as in the viewer here, I think the default would make more sense to be inverted.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -151,6 +156,10 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
   if (flags_ & Mn::Shaders::PhongGL::Flag::NormalTexture) {
     shader_->bindNormalTexture(*(materialData_->normalTexture));
   }
+  if ((flags_ & Mn::Shaders::PhongGL::Flag::ObjectIdTexture) ==
+      Mn::Shaders::PhongGL::Flag::ObjectIdTexture) {
```
mosra, line 160: FYI: [Containers::EnumSet::operator>=()](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html#aeb270daa9a8afb3a80e57ef2cb2f8314)

```suggestion 
  if (flags >= Mn::Shaders::PhongGL::Flag::ObjectIdTexture) {
```

src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -195,6 +193,30 @@ bool SemanticScene::buildHM3DHouse(std::ifstream& ifs,
...
+  for (const auto& objPtr : scene.objects_) {
+    int idx = objPtr->semanticID();
+    if (scene.semanticColorMapBeingUsed_.size() <= idx) {
+      // both are same size, so resize both at same time
+      scene.semanticColorMapBeingUsed_.resize(idx + 1);
+      scene.semanticColorToIdAndRegion_.reserve(idx + 1);
+    }
+    scene.semanticColorMapBeingUsed_[idx] = objPtr->getColor();
+    scene.semanticColorToIdAndRegion_.insert(
+        std::pair<uint32_t, std::pair<int, int>>(
```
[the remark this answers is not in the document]
jturner65: > Should be using std::make_pair here, right?
clang-tidy complained when I had it like that.

> /home/circleci/project/habitat-sim/src/esp/scene/HM3DSemanticScene.cpp:214:9: error: for C++11-compatibility, use pair directly [google-build-explicit-make-pair,-warnings-as-errors]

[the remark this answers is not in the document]
Skylion007: Hmm, we use C++14 though? Seems like something might not be set right.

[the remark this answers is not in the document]
Skylion007: @jturner65 This should be detected by the C++ compiler flags. If it's not something, is going wrong and the C++ compatibility level isn't being set properly.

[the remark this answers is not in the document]
Skylion007: Wow, they don't seem ot check if it's in C++11 mode if the check is specified. I'll remove this check as we aren't targeting any GCC versions that old, are we?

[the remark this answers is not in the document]
Skylion007: https://github.com/facebookresearch/habitat-sim/pull/1661 @jturner65 tada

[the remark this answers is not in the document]
jturner65: thanks, that seemed bogus to me too.

src/esp/assets/ResourceManager.cpp
jturner65: It is possible, but, as currently implemented in habitat, if the renderAsset is somehow configured to also have semantic textures, the current process would not load the textures as needed to render them, but would rather attempt to convert all loaded textures to integers for semantic rendering.  Since, for now, this won't happen (we don't have render assets that also have semantic textures) I put this here mainly as a note for the future, when I refactor the loading pipeline.

src/esp/assets/GenericSemanticMeshData.cpp
jturner65: These functions return a smart pointer to the object because these are getting moved into the meshes_ map, which holds smart pointers of objects.

aclegg3 (approved):
LGTM. We'll revisit with follow-up PRs to address remaining TODOs.

PR #1663, opened 2022-02-14
Fix flaky pytest-cov

Skylion007:
## Motivation and Context
* Noticed we still had pytest-cov pinned from when were still on 3.6. Now that we aren't, let's free up the requirement. I noticed it's been flaky lately. Hopefully this may fix it.

## How Has This Been Tested
* CI

## Types of changes


jturner65 (approved):
LGTM

PR #1682, opened 2022-03-03
[BUGFIX]--fix replicaCAD quickstart to not default to showing the NONE scene

jturner65:
## Motivation and Context
This addresses ReplicaCAD_quickstart.py only showing the NONE scene when run (the code would default to using the first scene in the list of available scenes, which would be the NONE scene since they are alphabetically sorted).  Now the NONE scene is returned only if, for some reason, the list of scenes is empty.  This is only an issue when running the code in python, since the drop-down scene select is not available there.

## How Has This Been Tested
Locally tested both python and colab.

## Types of changes

## Checklist


aclegg3 (approved):
LGTM

dhruvbatra:
Yup, I get a video now. It looks like the agent was spawned below the stairs or in a corner because all I saw was gray, but a scene was loaded. 

Let's land this and I can test again after rebasing.

PR #1684, opened 2022-03-04
Fix noise model

### LINKED ISSUE #1455: Question about Redwood Depth Noise Implementation
## Habitat-Sim version
v0.2.1

Habitat-lab commit: `bc85d09`
Habitat-sim commit: `fc7fb11`

## Docs and Tutorials
Did you read the docs? https://aihabitat.org/docs/habitat-sim/ **Yes**

Did you check out the tutorials? https://aihabitat.org/tutorial/2020/ **Yes**

Perhaps your question is answered there. If not, carry on!

## ❓ Questions and Help

I have a question about Redwood Depth Noise as a follow-up to the #987. Specifically, I am wondering whether the original GPU version is correct while the CPU version is off. Two major reasons:

### From Literature

From the original paper ["Unsupervised Intrinsic Calibration of Depth Sensors via SLAM"](http://www.roboticsproceedings.org/rss09/p27.html)'s Sec. IV, the author states

> ...  Finally, we discretize the depth multiplier images into bins of 8 × 6 pixels. At 640×480 resolution and a maximum range of 10m, this results in 32,000 parameters.

Essentially, this means the noise bins are of shape `[80, 80, 5]`. From the commit `d3d150c62f7d47c4350dd64d798017b2f47e66a9`, which is before #987:

1. GPU version gives `[80, 80, 5]`
https://github.com/facebookresearch/habitat-sim/blob/d3d150c62f7d47c4350dd64d798017b2f47e66a9/src/esp/sensor/RedwoodNoiseModel.cu#L13-L14

2. CPU version gives `[80, 100, 4]`
https://github.com/facebookresearch/habitat-sim/blob/d3d150c62f7d47c4350dd64d798017b2f47e66a9/habitat_sim/sensors/noise_models/redwood_depth_noise_model.py#L105-L106

IMHO, I think GPU version is correct.

### From Running 

Logically, I think the following line of function `_undistort` will cause out-of-bounds (OOB) error if we have noise bins of shape `[80, 100, 4]`:
https://github.com/facebookresearch/habitat-sim/blob/fc7fb11ccec407753a73ab810d1dbb5f57d0f9b9/habitat_sim/sensors/noise_models/redwood_depth_noise_model.py#L39

The reason is that since our noise bins are of shape `[80, 100, 4]`, the `min(max(i1, 0), 4)` and `min(i2, 4)` may output value of `4`, which will go out of the boundary.

However, we never encounter such OOB issue. After investigation, I find it is because `numba` does not check OOB error by default. See [this documentation](https://numba.pydata.org/numba-doc/dev/reference/pysemantics.html#bounds-checking):

> By default, instead of causing an IndexError, accessing an out-of-bound index of an array in a Numba-compiled function will return invalid values or lead to an access violation error (it’s reading from invalid memory locations).

I have verified it by the following:
```python
@numba.jit(nopython=True, fastmath=True)
def _undistort(x, y, z, model):
    i2 = int((z + 1) / 2)
    i1 = int(i2 - 1)
    a = (z - (i1 * 2.0 + 1.0)) / 2.0
    x = x // 8
    y = y // 6

    d = 4

    f = (1 - a) * model[y, x, d] + a * model[y, x, d]

    if f < 1e-5:
        return 0
    else:
        return z / f
``` 

- When `numba` is **disabled**, I encountered OOB error.
- When `numba` is **enabled**, no error will be thrown.

Any suggestions on this? Thanks a lot.

erikwijmans:
## Motivation and Context

When matching the GPU to CPU implementation of the noise model, I copied over a bug from the CPU side.

This PR also fixes some other issues in the GPU implementation, like not writing the curand states back to global memory and gives each thread it's own state (that should have been done earlier, the CUDA docs are confusing on if they a thread safe or not but they aren't).

Closes #1455 

## How Has This Been Tested

Via the test.

## Types of changes

### DIFF
--- a/src/esp/sensor/RedwoodNoiseModel.cu
+++ b/src/esp/sensor/RedwoodNoiseModel.cu
@@ -113,7 +113,6 @@ namespace sensor {
 namespace impl {
 
 struct CurandStates {
-  CurandStates() : devStates(0), nStates_(0) {}
   void alloc(const int nStates, const int maxThreadsPerBlock) {
     if (nStates > nStates_) {
       release();
@@ -136,10 +135,10 @@ struct CurandStates {
 
   ~CurandStates() { release(); }
 
-  curandState_t* devStates;
+  curandState_t* devStates = 0;
 
  private:
-  int nStates_;
+  int nStates_ = 0;
 };
 
 CurandStates* getCurandStates() {
--- a/src/esp/sensor/RedwoodNoiseModel.cu
+++ b/src/esp/sensor/RedwoodNoiseModel.cu
@@ -151,6 +151,7 @@ void freeCurandStates(CurandStates* curandStates) {
 }
 
 void simulateFromGPU(const int maxThreadsPerBlock,
+                     const int warpSize,
                      const float* __restrict__ devDepth,
                      const int H,
                      const int W,
@@ -160,7 +161,7 @@ void simulateFromGPU(const int maxThreadsPerBlock,
                      float* __restrict__ devNoisyDepth) {
   const int totalConcurrency = std::ceil(static_cast<float>(H * W) / 4.0f);
   const int nThreads =
-      std::min(std::max(roundToNearestMultiple(totalConcurrency, 32), 1),
+      std::min(std::max(roundToNearestMultiple(totalConcurrency, warpSize), 1),
                maxThreadsPerBlock);
   const int nBlocks =
       std::ceil(static_cast<float>(totalConcurrency) / nThreads);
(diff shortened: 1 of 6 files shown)


src/esp/sensor/RedwoodNoiseModel.cu
```diff
@@ -134,7 +139,7 @@ struct CurandStates {
   curandState_t* devStates;
 
  private:
-  int n_blocks_;
+  int nStates_;
```
Skylion007, line 142:
```suggestion
  int nStates_ = 0;
```

Skylion007 (approved):

src/esp/sensor/RedwoodNoiseModel.cu
erikwijmans: Makes sense as that's more consistent with my usage elsewhere.

PR #1688, opened 2022-03-08
Properly use texture transformation matrix for flat materials as well.

mosra:
## Motivation and Context

Fixes texture transformation with flat materials. Was done for Phong but not for flat.

Once we port away from these custom structures and use Magnum's `MaterialData` everywhere, this kind of bugs will become *much* less likely.

## How Has This Been Tested

Trust Me :tm:

## Types of changes

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -2224,6 +2224,14 @@ gfx::PhongMaterialData::uptr ResourceManager::buildFlatShadedMaterialData(
   // flat materials
   auto finalMaterial = gfx::PhongMaterialData::create_unique();
 
+  // texture transform, if the material has a texture; if there's none the
+  // matrix is an identity
+  // TODO this check shouldn't be needed, remove when this no longer asserts
+  // in magnum for untextured materials
+  if (material.hasTexture()) {
+    finalMaterial->textureMatrix = material.textureMatrix();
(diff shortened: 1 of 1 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -2224,6 +2224,13 @@ gfx::PhongMaterialData::uptr ResourceManager::buildFlatShadedMaterialData(
   // flat materials
   auto finalMaterial = gfx::PhongMaterialData::create_unique();
 
+  // texture transform, if the material has a texture; if there's none the
+  // matrix is an identity
+  // TODO this check shouldn't be needed, remove when this no longer asserts
+  // in magnum for untextured materials
+  if (material.hasTexture())
+    finalMaterial->textureMatrix = material.textureMatrix();
+
```
jturner65, line 2233: braces please, as per Meta C++ style guide :)

jturner65 (commented):
A nit

src/esp/assets/ResourceManager.cpp
mosra: Clang Tidy was not bullying me about this, so I thought I'm fine here. Always pushing the limits!

jturner65 (approved):

aclegg3 (approved):
Thanks for the fix!

PR #1694, opened 2022-03-17
ReplicaCAD v1.4

aclegg3:
## Motivation and Context

Updates the default downloader version for ReplicaCAD to v1.4.

Changes with v1.4:
- NavMeshes recomputed for agent radius 0.3m (from default 0.1m) corresponding to Fetch robot base as used in Habitat 2.0 (default navmeshes also included in a separate folder for optional use)
- Minor furniture layout modifications to the following scenes in order to better accommodate robot access to the full set of receptacles for rearrange tasks:
  - sc1_staging_02
  - sc1_staging_05
  - sc1_staging_12
  - sc1_staging_03
  - sc1_staging_06
  - sc1_staging_09
  - sc1_staging_10
  - sc1_staging_13
  - sc1_staging_16
  - sc2_staging_00
  - sc2_staging_07
  - sc2_staging_09
  - sc3_staging_03
  - sc3_staging_05
  - sc3_staging_11
  - sc4_staging_08
- Fixed layout error with Fridge intersecting tv_stand in sc2_staging_20
- Minor furniture layout modifications to the following scenes in order to move furniture blocking kitchen drawers from fully opening:
  - sc1_staging_03
  - sc1_staging_05
  - sc1_staging_09
  - sc1_staging_10
  - sc1_staging_14
  - sc1_staging_15
  - sc2_staging_03
  - sc2_staging_06
  - sc2_staging_09
  - sc2_staging_13
- Added visual geometry to `frl_apartment_shoebox_01.glb` which didn't have a bottom.

## How Has This Been Tested

Locally

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1700, opened 2022-03-20
Mobile Manipulator Bug Fix

ASzot:
## Motivation and Context

This PR fixes two issues:
1. The mobile manipulator joint creation was incorrect. The arm joint parameters were being applied to all joints rather than just the arm joints. The default motor parameters should be applied to the non-arm and non-base parameters. Fixing this bug results in much more stable arm movement.
2. The head camera offset was too small, so that when the robot moved around, the head camera would be occluded by the head geometry.

## Types of changes

## Checklist

### DIFF
--- a/src_python/habitat_sim/utils/datasets_download.py
+++ b/src_python/habitat_sim/utils/datasets_download.py
@@ -109,16 +109,16 @@ def initialize_test_data_sources(data_path):
             "version": "1.0",
         },
         "replica_cad_dataset": {
-            "source": "https://dl.fbaipublicfiles.com/habitat/ReplicaCAD/ReplicaCAD_dataset_v1.4.zip",
-            "package_name": "ReplicaCAD_dataset_v1.4.zip",
+            "source": "https://dl.fbaipublicfiles.com/habitat/ReplicaCAD/ReplicaCAD_dataset_v1.3.zip",
+            "package_name": "ReplicaCAD_dataset_v1.3.zip",
             "link": data_path + "replica_cad",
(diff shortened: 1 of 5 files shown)


aclegg3 (commented):
These changes make sense. 

Looks like the gripper open test is failing, maybe because it used to have arm motor settings and now has default settings?

src_python/habitat_sim/utils/datasets_download.py
```diff
@@ -109,16 +109,16 @@ def initialize_test_data_sources(data_path):
...
+            "version": "1.4",
         },
         "replica_cad_baked_lighting": {
-            "source": "https://dl.fbaipublicfiles.com/habitat/ReplicaCAD/ReplicaCAD_baked_lighting_v1.3.zip",
-            "package_name": "ReplicaCAD_baked_lighting_v1.3.zip",
+            "source": "https://dl.fbaipublicfiles.com/habitat/ReplicaCAD/ReplicaCAD_baked_lighting_v1.4.zip",
+            "package_name": "ReplicaCAD_baked_lighting_v1.4.zip",
             "link": data_path + "replica_cad_baked_lighting",
-            "version": "1.3",
+            "version": "1.4",
```
aclegg3, line 121: Let's keep these changes isolated to #1694, so remove before merge.

ASzot: Done

aclegg3 (approved):
LGTM. When tests pass, let's merge.

PR #1710, opened 2022-03-29
Python Sensor debug_draw

aclegg3:
## Motivation and Context

It can be useful to define custom debug drawing methods with a derived Simulator class (e.g. in Habitat-lab). This PR adds a hook to a `debug_draw()` function stub which is called in `simulator.get_sensor_observations()` and can be overridden for application specific debugging visualizations.

For example, 

## How Has This Been Tested

Overrode this in a Habitat-lab task and added [DebugLineRender](https://aihabitat.org/docs/habitat-sim/habitat_sim.gfx.DebugLineRender.html) commands for an evaluation camera sensor.

## Types of changes

## Checklist

### DIFF
--- a/src_python/habitat_sim/simulator.py
+++ b/src_python/habitat_sim/simulator.py
@@ -425,6 +425,7 @@ class Simulator(SimulatorBackend):
             agent_sensorsuite = self.__sensors[agent_id]
             for _sensor_uuid, sensor in agent_sensorsuite.items():
                 sensor.draw_observation()
+                self.debug_draw()
 
         # As backport. All Dicts are ordered in Python >= 3.7
         observations: Dict[int, ObservationDict] = OrderedDict()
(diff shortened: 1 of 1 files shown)


src_python/habitat_sim/simulator.py
```diff
@@ -437,6 +438,13 @@ def get_sensor_observations(
             return next(iter(observations.values()))
         return observations
 
+    def debug_draw(self, sensor_uuid: Optional[str] = None) -> None:
```
Skylion007, line 441: Shouldn't this be attached to the sensor object instead?

aclegg3: I thought about this, but it adds extra complexity to the system overall. Users would need to override each sensor's debug_draw, possibly deriving the sensor type also (e.g. to only draw in one of multiple RGB sensors) instead of defining all the debug visualizations in one, centralized place.

Skylion007: Yeah, but one of the longterm goals is to remove the sensor dict from the Simulator overall and this just complicates this down the road.

aclegg3: fair point, I'll think on this a bit

PR #1720, opened 2022-04-06
orthonormalize the rotationShear() part of transforms

erikwijmans:
## Motivation and Context

When setting scene graph transformations, we need to explicitly normalize the rotation component to be orthonormal as sometimes floating point errors accumulate.

I also preemptively added an orthornomalize call when updating the camera positions for a robot as that seems like it'll be prone to this issue.

## How Has This Been Tested

This was discovered in https://github.com/facebookresearch/habitat-lab/issues/697, so I used the repro script there.

## Types of changes

### DIFF
--- a/src_python/habitat_sim/utils/common.py
+++ b/src_python/habitat_sim/utils/common.py
@@ -14,6 +14,10 @@ import magnum as mn
 import numpy as np
 import quaternion as qt
 
+from habitat_sim._ext.habitat_sim_bindings.core import (  # noqa: F401
+    orthonormalize_rotation_shear,
+)
+
 
 def quat_from_coeffs(coeffs: Union[Sequence[float], np.ndarray]) -> qt.quaternion:
     r"""Creates a quaternion from the coeffs returned by the simulator backend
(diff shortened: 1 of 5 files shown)


aclegg3 (approved):

Skylion007 (approved):

src_python/habitat_sim/utils/common.py
```diff
@@ -14,6 +14,10 @@
 import numpy as np
 import quaternion as qt
 
+from habitat_sim._ext.habitat_sim_bindings.core import (  # noqa: F401
```
Skylion007, line 17: @erikwijmans Actually, just realized that this should have been imported via `bindings/__init__.py` this should have been caught by the import-linter, but I suppose moving the python to src_python may have broken it.

Skylion007 (commented):
Noticed one minor thing.

src_python/habitat_sim/utils/common.py
erikwijmans: Examples and tests shouldn't, but I think it's fine to important from the bindings directly with src_python. I didn't import this in `bindings/__init__.py` as I think this function's canonical home should be here as this is where we have other rigid transform utils.

PR #1728, opened 2022-04-21
Update Magnum to latest and adapt to API deprecation

mosra:
## Motivation and Context

Updates Corrade & Magnum to latest and adapts to breaking changes that happened over the last four (or more?) months:

- `Utility::Directory` is deprecated in favor of [Utility::Path](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1Path.html). It's shorter to type, but the main reason for a name change was that it would be impossible to preserve backwards compatibility due to all changes that had to be made:
    - `read()`, `map()`, `list()` and others now return `Optional<Array>` instead of `Array` to clearly distinguish a failure from an empty file or directory
    - all functions calling into OS APIs now properly report failures including the OS-specific error code / message
    - `read()` is now up to 2x faster thanks to the output memory not being unnecessarily zero-initialized anymore
    - `readString()` is up to 3x faster thanks to the output memory no longer being zero-initialized and *then copied* to a `std::string`, the memory ownership is now instead *transferred* from an `Array` to a `String` without a copy
    - All functions now take string views instead of allocated strings, meaning a lot less temporary string copies and allocations needs to be made. Functions operating on paths such as `split()` and `splitExtension()` now also *return* string views, meaning you can perform complex path decomposing operations with no temporary allocations. (But watch out for dangling views, of course.)
    - APIs with silly Unix-like names such as `rm()` or `mkpath()` dating back to 2008 were changed to clearer names
- Array, array view, string and string view slicing APIs were cleaned up for more consistency, deprecating functions with ambiguous naming
- `TestSuite`, `PluginManager`, `Utility::Resource` and `Trade` APIs were ported to use string views, removing many cases of unnecessary temporary std::string allocations
- All function taking or returning strings and string views now also work with C++17 `std::string_view` (but Corrade's string views [are better, of course](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BasicStringView.html) :trollface:)
- New [TestSuite::Compare::StringContains](https://doc.magnum.graphics/corrade/classCorrade_1_1TestSuite_1_1Compare_1_1StringContains.html) and other comparators for advanced string comparisons with detailed failure diagnostics
- As a side-product of the glTF exporter that I'm working on, Corrade now has a streaming [Utility::JsonWriter](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1JsonWriter.html) as well as a minimalistic read-only [Utility::Json](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Json.html) tokenizer & parser. Because, after suffering through various limitations and annoyances with 3rd party glTF parsers and writers, this turned out to be the most straightforward option. The parser took about two days to write and has speed and memory use comparable to rapidjson (i.e., there's still a lot of room for improvement).
- A new set of [CORRADE_*_DEBUG_ASSERT()](https://doc.magnum.graphics/corrade/DebugAssert_8h.html) macros that get compiled out on release builds, for use in tight loops or element accessors where a regular `CORRADE_ASSERT()` would make things too slow. These aren't used in any Magnum code yet, but I'll be gradually adding them to various `operator[]` accessors that didn't have any bounds checks exactly because it would have too much overhead in release builds.
- A new [MeshTools::boundingSphereBouncingBubble()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a4345353fda33aa694181c42067a7ecd7) algorithm for calculating a tight bounding sphere of an arbitrary mesh. Compared to AABB it's rotation invariant and collisions are cheaper to calculate when doing frustum culling. [Sphere/frustum collision algorithms are already there](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Math_1_1Intersection.html#aa7c4b8af19c89fa58368e6e1bb04da5a) for quite some time. (@erikwijmans, FYI)
- Building blocks for batch-friendly glTF assets:
    - Custom property and extension import for glTF materials in [CgltfImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1CgltfImporter.html#Trade-CgltfImporter-behavior-materials). Custom property import for glTF scenes is being worked on, and together these features will be used to describe mesh ranges and texture layers to submit to a multi-draw call.
    - 2D array image conversion in [BasisImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1BasisImageConverter.html), together with related plumbing in `AnyImageConverter` and `magnum-imageconverter`

The thousands of changed files in the PR diff are only because of copyright years being updated, the real changes don't amount to *that* much ;)

## How Has This Been Tested

I demand the CI to pass already :green_circle:

### DIFF
--- a/src/esp/assets/BaseMesh.cpp
+++ b/src/esp/assets/BaseMesh.cpp
@@ -42,7 +42,7 @@ void BaseMesh::convertMeshColors(
   } else {
     Mn::Math::packInto(
         Cr::Containers::arrayCast<2, float>(stridedArrayView(colors))
-            .except({0, 1}),
+            .exceptSuffix({0, 1}),
         Cr::Containers::arrayCast<2, Mn::UnsignedByte>(
             stridedArrayView(meshColors)));
   }
--- a/src/esp/assets/PTexMeshData.cpp
+++ b/src/esp/assets/PTexMeshData.cpp
@@ -16,7 +16,7 @@
 #include <Corrade/Utility/Assert.h>
 #include <Corrade/Utility/Debug.h>
 #include <Corrade/Utility/DebugStl.h>
-#include <Corrade/Utility/Directory.h>
+#include <Corrade/Utility/Path.h>
 #include <Magnum/GL/BufferTextureFormat.h>
 #include <Magnum/GL/TextureFormat.h>
 #include <Magnum/ImageView.h>
@@ -37,18 +37,18 @@ namespace assets {
 
 void PTexMeshData::load(const std::string& meshFile,
                         const std::string& atlasFolder) {
-  if (!Cr::Utility::Directory::exists(meshFile)) {
+  if (!Cr::Utility::Path::exists(meshFile)) {
     Cr::Utility::Fatal{-1} << "PTexMeshData::load: Mesh file" << meshFile
                            << "does not exist.";
   }
-  if (!Cr::Utility::Directory::exists(atlasFolder)) {
+  if (!Cr::Utility::Path::exists(atlasFolder)) {
     Cr::Utility::Fatal{-1} << "PTexMeshData::load: The atlasFolder"
                            << atlasFolder << "does not exist.";
   }
 
   // Parse parameters
   const auto& paramsFile = atlasFolder + "/parameters.json";
-  if (!Cr::Utility::Directory::exists(paramsFile)) {
+  if (!Cr::Utility::Path::exists(paramsFile)) {
     Cr::Utility::Fatal{-1} << "PTexMeshData::load: The parameter file"
                            << paramsFile << "does not exist.";
   }
@@ -490,7 +490,7 @@ void PTexMeshData::loadMeshData(const std::string& meshFile) {
     // splitMesh(...)
 
     // See detailed comments in front of the splitMesh(...)
-    std::string subMeshesFilename = Corrade::Utility::Directory::join(
+    std::string subMeshesFilename = Corrade::Utility::Path::join(
         atlasFolder_, "../habitat/sorted_faces.bin");
     submeshes_ = loadSubMeshes(originalMesh, subMeshesFilename);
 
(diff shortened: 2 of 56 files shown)


Skylion007:
@mosra Going from `empty()` to `isEmpty()` is an interesting choice considering the former was more compatible with STL types and best practices.

aclegg3 (approved):
LGTM

mosra:
@Skylion007 It's just that STL was first to make such convention, and that it was first does not necessarily make it a "best practice" :) STL also has a convention where `int bar:foo()` is a getter and `void bar::foo(int)` is a setter, and I don't know about many projects that would follow that. Counterpoint -- all other bool getters I have are already named `isSomething()`, which is considered a best practice in other codebases [such as Qt](https://doc.qt.io/qt-5/qlist.html#isEmpty), and `empty()` was *the only* weird outlier. Moreover `empty()` started to feel a lot more like "make a container empty" than "is a container empty", `isEmpty()` doesn't have that ambiguity.

I'm not strictly following STL design language anyway, so I don't think this breaks anything either -- and if people want, they can keep using native STL types like `std::span` or `std::string_view` instead, with implicit conversions happening automatically when interfacing with Magnum.

Skylion007 (approved):

jturner65 (approved):
Wow, looks great.

mosra:
This might break potential other long-running PRs that still rely on `<Corrade/Utility/Directory.h>` being included, so I'll leave the merge up to you ;)

jturner65:
> This might break potential other long-running PRs that still rely on `<Corrade/Utility/Directory.h>` being included, so I'll leave the merge up to you ;)

FWIW I think it's for the PR author to make sure they retain compatibility with main.

Skylion007:
>> This might break potential other long-running PRs that still rely on <Corrade/Utility/Directory.h> being included, so I'll leave the merge up to you ;)

@mosra Will it break or just issue deprecation warnings?

mosra:
I removed all existing includes (because they *also* issue deprecation warnings), so there's nothing providing the Directory namespace anymore.

As a middle-ground safe-to-merge solution the changes that replace `Directory.h` with `Path.h` could be reverted and then submitted as a follow-up PR "once the dust settles".

aclegg3:
> > This might break potential other long-running PRs that still rely on `<Corrade/Utility/Directory.h>` being included, so I'll leave the merge up to you ;)
> 
> FWIW I think it's for the PR author to make sure they retain compatibility with main.

I agree. I'll take a quick look through active PRs, but I think we can move on past the longer standing ones and assume they'll need refactoring anyway if/when they are picked back up.

mosra:
Okay then! Don't blame me when something breaks :P

PR #1734, opened 2022-04-25
[viewer.py] unloaded stage cycling

aclegg3:
## Motivation and Context

Add feature to viewer.py for cycling through unloaded stages in a SceneDataset via `SHIFT-TAB` keypress. 

Currently, scenes will not be created for stage files without separate configs unless explicitly set as the `scene_id`. viewer.py scene cycling with `TAB` cycles registered scene instances, so unregistered stages will never be loaded. This feature adds a way to force new stages to be loaded with cycling to make investigation of SceneDatasets such as gibson and MP3D a bit easier.

## How Has This Been Tested

Locally w/ gibson

## Types of changes

## Checklist

### DIFF
--- a/examples/viewer.py
+++ b/examples/viewer.py
@@ -328,6 +328,7 @@ class HabitatSimInteractiveViewer(Application):
 
         elif key == pressed.TAB:
             # NOTE: (+ALT) - reconfigure without cycling scenes
+            # NOTE: (+SHIFT) - cycle through stages without corresponding scenes
             if not alt_pressed:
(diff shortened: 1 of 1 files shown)


aclegg3:
Note: considering refactor to use same approach as Viewer.cpp: [construct a list of scene/stage handles in advance and then cycle through that list](https://github.com/facebookresearch/habitat-sim/blob/d506d13c1c37bc461d72b7e1fd691ab0506e4ff8/src/utils/viewer/viewer.cpp#L933).

PR #1740, opened 2022-05-02
Rigid object scaling

aclegg3:
## Motivation and Context

Restricting object scaling to pre-initialization via templates is very limiting. 
This PR adds support for programmatic scaling of instances.

Details:
Because the top level RigidObject SceneNode transform is a Bullet compatible MotionState and therefore does not support scaling, we apply scaling directly to the Bullet collision object and render scale is stored in a child SceneNode of the visualNode_.

**TODO:**
This solution does not play nice with the RenderReplay functionality which only tracks RigidObject top level SceneNode and therefore misses the scale updates not applied during initialization. Potentially replay functionality should track scale explicitly from the new `scalingNode_` for `RigidObjects`.

## How Has This Been Tested

Tested locally and added unit test.

https://user-images.githubusercontent.com/1445143/166329291-3b1bc0ee-8f91-407e-b37f-5fe355f94599.mov

## Types of changes

## Checklist


src/esp/physics/bullet/BulletRigidObject.cpp
```diff
@@ -389,23 +389,28 @@ void BulletRigidObject::constructAndAddRigidBody(MotionType mt) {
...
+            CollisionGroupHelper::getMaskForGroup(configuredCollisionGroup_)));
   } else {
-    bWorld_->addRigidBody(bObjectRigidBody_.get(), int(CollisionGroup::Dynamic),
-                          uint32_t(CollisionGroupHelper::getMaskForGroup(
-                              CollisionGroup::Dynamic)));
+    configuredCollisionGroup_ = CollisionGroup::Dynamic;
+    bWorld_->addRigidBody(
+        bObjectRigidBody_.get(), int(configuredCollisionGroup_),
+        uint32_t(
+            CollisionGroupHelper::getMaskForGroup(configuredCollisionGroup_)));
```
jturner65, line 413: Perhaps move the addRigidBody to be outside the if block? Also, how about making the C-style casts into static_casts ?

src/esp/physics/bullet/BulletRigidObject.h
```diff
@@ -395,6 +395,42 @@ class BulletRigidObject : public BulletBase,
...
+      // Automatically recompute the intertia to match the collision object
+      // scale.
+      btVector3 bInertia = {0, 0, 0};
+      bObjectShape_->calculateLocalInertia(getMass(), bInertia);
+      bObjectRigidBody_->setMassProps(getMass(), bInertia);
+      // maintain the current collisionGroup
+      bWorld_->addRigidBody(bObjectRigidBody_.get(),
+                            int(configuredCollisionGroup_),
+                            uint32_t(CollisionGroupHelper::getMaskForGroup(
+                                configuredCollisionGroup_)));
```
jturner65, line 428: C-style casts again.

examples/tutorials/colabs/replay_tutorial.ipynb
```diff
@@ -334,6 +334,10 @@
     "chefcan_3 = rigid_obj_mgr.add_object_by_template_handle(chefcan_template_handle)\n",
     "chefcan_3.translation = [2.4, -0.64, -0.28]\n",
     "\n",
+    "# scale some objects\n",
+    "# TODO: this doesn't work because replay is only tracking the MotionState, not the scaling SceneNode further in the tree. Needs a refactor.\n",
```
eundersander, line 338: Gfx-replay support for dynamic scaling would be a small change to `Recorder::getInstanceState` and `RenderAssetInstanceState`. If we should decide to move forward with merging this, can you make this change as part of this PR? (So that we avoid incurring this tech debt.)

See also `GfxReplayTest::testSimulatorIntegration` for C++ unit-testing.

examples/viewer.py
```diff
@@ -782,6 +845,9 @@ def print_help_text(self) -> None:
     't':        Load URDF from filepath
                 (+SHIFT) quick re-load the previously specified URDF
                 (+ALT) load the URDF with fixed base
+    'o':        Add a random object to the scene in front of the agent.
```
eundersander, line 848: For future reference: it would be nice if the python viewer's UI didn't become an explosion of hotkeys as happened to the C++ viewer. It seems to me the root of the problem is that our K different viewer use cases all have to fight over the same set of physical keys.

src/esp/bindings/PhysicsObjectBindings.cpp
```diff
@@ -209,8 +209,9 @@ void declareRigidBaseWrapper(py::module& m,
 
       /* --- Geometry & Transformations --- */
 
-      .def_property_readonly("scale", &RigidBaseWrapper::getScale,
-                             ("Get the scale of the " + objType).c_str())
+      .def_property("scale", &RigidBaseWrapper::getScale,
+                    &RigidBaseWrapper::setScale,
```
eundersander, line 213: `setScale` is an expensive operation, right? To me, property getter/setters imply negligible runtime cost. You might consider an explicit python `set_scale` to discourage python users from doing lazy things like:
```
obs.scale = 1.0
if larger:
  obs.scale = 2.0
```

src/esp/physics/PhysicsManager.cpp
```diff
@@ -224,6 +224,8 @@ int PhysicsManager::addObject(
     resourceManager_.addObjectToDrawables(obj->getInitializationAttributes(),
                                           obj->visualNode_, drawables,
                                           obj->visualNodes_, lightSetup);
+    // Set the scalingNode_ to the first child of the visualNode_
+    obj->scalingNode_ = obj->visualNodes_[1];
```
eundersander, line 228: Can you add some kind of assert that `visualNodes_[1]` is actually what you're expecting here? E.g. in case upstream code changes.

src/esp/physics/RigidBase.h
```diff
@@ -334,11 +334,27 @@ class RigidBase : public esp::physics::PhysicsObjectBase {
...
+    } else {
+      return initializationAttributes_->getScale();
+    }
+  }
+
+  /** @brief Set the non-uniform scaling of the object.
+   * @param scale The desired non-uniform scaling of the object.
+   */
+  virtual void setScale(const Magnum::Vector3& scale) {
+    if (objectMotionType_ != MotionType::STATIC) {
```
eundersander, line 352: Seems confusing to silently fail for STATIC objects and objects without `scalingNode_`.

eundersander (commented):
I left some comments. We should also finish the discussion about whether we truly want to support dynamic scaling. https://cvmlp.slack.com/archives/G0131KVLBLL/p1651526528000309

aclegg3:
Decided against this approach.

PR #1753, opened 2022-05-13
--Add ConfigValue support for Mn::Matrix3 (3x3 matrices of floats)

jturner65:
## Motivation and Context
This PR adds support for 3x3 float Magnum Matrices to the ConfigValue/Configuration subsystem.  This includes writing/reading json values, as well internally storing the data within a ConfigValue structure.

IOTest has also been streamlined somewhat, to make adding tests to validate json read/write for new types easier.

## How Has This Been Tested
Locally c++ and python tests pass.  CoreTest and IOTest have been modified to test appropriate functionality.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/Configuration.cpp
+++ b/src/esp/core/Configuration.cpp
@@ -191,7 +191,7 @@ std::string ConfigValue::getAsString() const {
           Cr::Utility::formatString("[[{} {} {}]", v.x(), v.y(), v.z());
       for (int i = 1; i < m.Size; ++i) {
         v = m.row(i);
-        Cr::Utility::formatInto(res, res.length(), ", [{} {} {}]", v.x(), v.y(),
+        Cr::Utility::formatInto(res, res.length(), "[{} {} {}]", v.x(), v.y(),
                                 v.z());
       }
       Cr::Utility::formatInto(res, res.length(), "]");
--- a/src/esp/core/Configuration.cpp
+++ b/src/esp/core/Configuration.cpp
@@ -307,7 +307,7 @@ int Configuration::loadFromJson(const io::JsonGenericValue& jsonObj) {
     } else if (obj.IsBool()) {
       set(key, obj.GetBool());
     } else if (obj.IsArray() && obj.Size() > 0) {
-      // non-empty array
+      // non-empty array of values
       if (obj[0].IsNumber()) {
         // numeric vector or quaternion
         if (obj.Size() == 3) {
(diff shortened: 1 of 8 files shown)


src/esp/core/Configuration.cpp
```diff
@@ -287,25 +306,53 @@ int Configuration::loadFromJson(const io::JsonGenericValue& jsonObj) {
...
+              << "Config cell in JSON document contains key" << key
+              << "referencing an unsupported numeric array of length :"
+              << obj.Size() << "so skipping.";
         }
-      } else if (obj.Size() == 4) {
-        // assume is quaternion
-        Magnum::Quaternion val{};
-        if (io::fromJsonValue(obj, val)) {
-          set(key, val);
+      } else if (obj[0].IsArray() && obj[0][0].IsNumber()) {
```
mosra, line 333: I suppose this writes a JSON in a form of

```json
[[1, 0, 0],
 [0, 1, 0],
 [0, 0, 1]] 
```

right? I'd prefer an one-dimensional array instead, because:

- this is how glTF does it (in case you'd eventually want JSON compatibility to some degree)
- it's faster to parse, faster to read, faster to write, faster to validate (one just dumps array of 9 values, or checks that the numeric array has 9 values; which is especially convenient with the new [Utility::JsonToken::asFloatArray()](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1JsonToken.html#ab0aeb26b5fe66652c95f6ab465718651) and [Utility::JsonWriter::writeArray()](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1JsonWriter.html#a7f783366d73062d35e5a8059e8f01a0b) APIs, and the *also new* [Matrix::data()](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1RectangularMatrix.html#a18852de01a97e7f65398097520364288) API that returns a sized array, thus the copy from/to JSON can be a single `Utility::copy(json.asFloatArray(), matrix.data())` / `json.writeFloatArray(matrix.data())` call)
- there doesn't need to be any extra logic to flip from a column-major representation to a row-major representation, it's just a data copy

src/esp/core/Configuration.cpp
```diff
@@ -183,6 +184,19 @@ std::string ConfigValue::getAsString() const {
...
+      std::string res =
+          Cr::Utility::formatString("[[{} {} {}]", v.x(), v.y(), v.z());
+      for (int i = 1; i < m.Size; ++i) {
+        v = m.row(i);
+        Cr::Utility::formatInto(res, res.length(), ", [{} {} {}]", v.x(), v.y(),
+                                v.z());
+      }
+      Cr::Utility::formatInto(res, res.length(), "]");
+      return res;
+    }
```
mosra, line 199: Ah wait, why is this class reinventing Magnum's own builtin debug output for the vectors, quaternions and radians? It's as simple as this, for *any* Magnum type:

```cpp
std::ostringstream out;
out << get<T>();
return out.str();
```

Besides minor formatting differences, I don't see why you would have to write all that code again.

tests/test_configs.py
```diff
@@ -41,6 +41,11 @@ def test_core_configuration():
     config.set("vec3", my_vec3)
     assert config.get("vec3") == my_vec3
 
+    # Magnum::Matrix3 (3x3)
+    my_mat3x3 = mn.Matrix3((1.1, 2.2, -3.3), (4.4, 5.0, -6.6), (7.7, 8.0, -9.9))
+    config.set("mat3x3", my_mat3x3)
+    assert config.get("mat3x3")
```
mosra, line 47:
```suggestion
    assert config.get("mat3x3") == my_mat3x3
```

? :eyes:

aclegg3 (approved):
LGTM

src/esp/core/Configuration.cpp
jturner65: The to-strings were done this way originally (with vec3 and quats) to remove the commas between the elements (this function was being used to provide data in a csv-friendly format).  I figured that if folks wanted the magnum-based stream version, they could get it easily enough in the manner you described.

Skylion007: @mosra Is there a flag for Magnum::Debug::Flag for NoComma?

jturner65: I was looking for that, but it looks like the stream functions [specify the comma explicitly ](https://github.com/mosra/magnum/blob/5f5cfb10946a344b9dab704cbf6922fc3ee5bbb0/src/Magnum/Math/RectangularMatrix.h#L682)

mosra: Yeah the formatting is static because the intention is for it to be human-readable, not machine-consumable. Though in your case it's not CSV-friendly either due to the `[ ]`... :) 

I have a different idea for a generic machine-readable output -- all Math classes have a `.data()` accessor, which (in latest Magnum) provides a sized array reference. So then you could have a single function, such as `printCsvFriendly(Cr::Containers::ArrayView<const T>)`, which converts a numeric array into a space-delimited string, and then use it for any Math type.

src/esp/core/Configuration.cpp
jturner65: My original thinking was that I wanted to make something that was a bit more human-readable, but I did notice the gltf standard (Which I am shooting for strict adherence to) so doing things this way makes sense to me.

PR #1763, opened 2022-05-25
[bugfix] - moveUp moveDown

aclegg3:
## Motivation and Context

`moveUp` and `moveDown` actions (in C++) are used exclusively in the viewer application to raise and lower the agent camera. Current implementation uses camera local "up" translation. When the camera is pitched this results in moving the camera node off-axis, changing the axis of rotation. 

This PR changes moveUp and moveDown action to apply parent coordinate system +Y translation instead of local "up" so the camera always stays constrained to the agent's "up" axis.

If moving the camera off-axis is desirable it should be implemented as a separate action.

## How Has This Been Tested

Local in viewer.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/scene/ObjectControls.cpp
+++ b/src/esp/scene/ObjectControls.cpp
@@ -28,7 +28,9 @@ SceneNode& moveLeft(SceneNode& object, float distance) {
 
 SceneNode& moveUp(SceneNode& object, float distance) {
   // TODO: this assumes no scale is applied
-  object.translateLocal(object.transformation().up() * distance);
+  // Note: this is not a body action and is applied to the sensor rather than
+  // the Agent, so it will move the sensor in Agent's +Y (up) direction
+  object.translate(Magnum::Vector3(0, 1, 0) * distance);
   return object;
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

jturner65 (approved):

src/esp/scene/ObjectControls.cpp
```diff
@@ -28,7 +28,9 @@ SceneNode& moveLeft(SceneNode& object, float distance) {
 
 SceneNode& moveUp(SceneNode& object, float distance) {
   // TODO: this assumes no scale is applied
-  object.translateLocal(object.transformation().up() * distance);
+  // Note: this is not a body action and is applied to the sensor rather than
+  // the Agent, so it will move the sensor in Agent's +Y (up) direction
+  object.translate(Magnum::Vector3(0, 1, 0) * distance);
```
Skylion007, line 33: @aclegg3 This is a magic number (or rather vector). May want to make it a global constant somewhere.

PR #1766, opened 2022-05-26
--AttributesManagers : Move abstract method to obj mgrs; simplify Object/Stage Manager creation;

jturner65:
## Motivation and Context
This small PR simplifes the AttributesManagers architecture somewhat, moving an abstract method out of the base class template into the derived class template where it is used and appropriate.  Previously inappropriate AttributeManagers had access to this

The instantiation of the Object and Stage attributes managers has also been simplified and no longer requires a separate post-constructor init call.

## How Has This Been Tested
Locally c++ and python tests all pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/metadata/managers/StageAttributesManager.cpp
+++ b/src/esp/metadata/managers/StageAttributesManager.cpp
@@ -25,24 +25,31 @@ using attributes::StageAttributes;
 namespace managers {
 
 StageAttributesManager::StageAttributesManager(
-    ObjectAttributesManager::ptr objectAttributesMgr,
-    PhysicsAttributesManager::ptr physicsAttributesManager)
+    AssetAttributesManager::cptr assetAttributesMgr,
+    PhysicsAttributesManager::cptr physicsAttributesManager)
     : AbstractObjectAttributesManager<StageAttributes,
                                       ManagedObjectAccess::Copy>::
-          AbstractObjectAttributesManager("Stage", "stage_config.json"),
-      objectAttributesMgr_(std::move(objectAttributesMgr)),
+          AbstractObjectAttributesManager("Stage",
+                                          "stage_config.json",
+                                          assetAttributesMgr),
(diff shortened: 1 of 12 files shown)


Skylion007:
@jturner65 Pulling from main should fix most of these tests.

jturner65:
> @jturner65 Pulling from main should fix most of these tests.

Some of them are regarding virtual dispatch - I was experimenting with a configuration that will probably not work.

Skylion007:
Ah, yeah I noticed those a well. Seems tricky to fix properly.

jturner65:
> Ah, yeah I noticed those a well. Seems tricky to fix properly.

Probably going to have to undo what I added, was more of an experiment to get some initialization done in the constructor instead of having a separate init function.

src/esp/metadata/managers/StageAttributesManager.cpp
```diff
@@ -25,24 +25,31 @@ using attributes::StageAttributes;
...
+
+}  // StageAttributesManager::ctor
+
+void StageAttributesManager::createDefaultPrimTemplatesForObjType() {
   // create none-type stage attributes and set as undeletable
   // based on default
   auto tmplt = this->postCreateRegister(
       StageAttributesManager::initNewObjectInternal("NONE", false), true);
   std::string tmpltHandle = tmplt->getHandle();
   this->undeletableObjectNames_.insert(tmpltHandle);
```
Skylion007, line 51:
```suggestion
  this->undeletableObjectNames_.insert(std::move(tmpltHandle));
```

Skylion007 (commented):
Nit

src/esp/metadata/managers/StageAttributesManager.cpp
Skylion007: Actually nvm, went and seperated out these fixes into there own PR.

PR #1819, opened 2022-07-20
Update Magnum to latest & fix various minor buildsystem issues

mosra:
## Motivation and Context

Changes that aren't really related to the upcoming second batch renderer PR, thus submitting them separately.

Regular update of Magnum submodules:

- Adapting to a minor API deprecation (`BoolVector` -> `BitVector`, for consistency with new `BitArray` containers and because it's one character less to type)
- Using now-builtin `std::hash` specialization for `Containers::String`
- No other changes worth mentioning, the major piece of work done lately isn't published yet

And some minor fixes. For faster iterations I'm having a second build directory that uses a system-installed Magnum and has various features disabled, and various code rot prevented that from working:

- Various CMake updates that are needed if Magnum is external (and are no-op if it's a subproject)
- Making a build without Bullet work again
- Making a build with dynamic Magnum work again

And finally, certain bindings code used a private header from `magnum-bindings`. It's private so I don't guarantee any forward compatibility, I don't want anybody to use it :stuck_out_tongue_closed_eyes:, and it doesn't even get installed which causes a build failure if Magnum is external. I made a copy of that utility directly inside Habiat and the bindings code now uses that instead.

## How Has This Been Tested

Builds locally. 

## Types of changes

### DIFF
--- a/src/esp/bindings/EnumOperators.h
+++ b/src/esp/bindings/EnumOperators.h
@@ -11,15 +11,29 @@ namespace esp {
 
 /* Pybind's py::arithmetic() doesn't work on strongly-typed enums with custom
    operators */
-template<class T> void pybindEnumOperators(pybind11::enum_<T>& e) {
-    e
-        .def("__or__", [](const T& a, const T& b) { return T(typename std::underlying_type<T>::type(a | b)); })
-        .def("__and__", [](const T& a, const T& b) { return T(typename std::underlying_type<T>::type(a & b)); })
-        .def("__xor__", [](const T& a, const T& b) { return T(typename std::underlying_type<T>::type(a ^ b)); })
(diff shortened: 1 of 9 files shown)


Skylion007:
> Using now-builtin std::hash specialization for Containers::String

Nice.

Skylion007:
Hmm... "$SRC_DIR/src/esp/bindings/GfxBindings.cpp:15:10: fatal error: 'esp/bindings/EnumOperators.h' file not found"

mosra:
Lol, sorry. Fixed.

Skylion007:
Now clang-format is complaining a bit.

mosra:
Fixed.

(Note to self: don't open PRs at 8 PM.)

Skylion007 (approved):

PR #1834, opened 2022-08-18
Update to Magnum with initial support for glTF export and CPU feature dispatch

mosra:
## Motivation and Context

Just a regular boring dependency update ... with 20k+ lines of changes.

- [Interfaces for generic scene export](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1AbstractSceneConverter.html), including meshes, images, scene hierarchy, materials, lights etc. Not hooked up into the command-line tools *yet* so things like `magnum-sceneconverter gripper_link.dae gripper_link.gltf` aren't possible at the moment. But it's getting closer!

- Initial implementation of a [glTF export plugin](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1GltfSceneConverter.html), which is a prerequisite for the whole preprocessing pipeline in the batch renderer. Currently just (multi-)mesh export. Material, image and scene hierarchy export is in a branch and undergoing a final cleanup.

- [Compile-time and runtime detection of CPU features](https://blog.magnum.graphics/backstage/cpu-feature-detection-dispatch/) (such as AVX or ARM NEON) and dispatch. To be used to select among optimized code paths for batch transformation processing, broadphase visibility culling and other hot code paths in the batch renderer. Earlier benchmarks I did showed a possibility of getting a 2x to 9x speedup in transformation processing by using AVX intrinsics, depending on the compiler.

    image

- All other changes and fixes that were too small and insignificant in comparison.

Updates to the batch renderer are depending on these, I'll submit them separately -- wanted to submit the heavy but uncontroversial stuff first to iron out potential build issues.

## How Has This Been Tested

My CIs are passing. I'm hopeful it'll eventually be the case here as well :]

## Types of changes

### DIFF
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -216,13 +216,10 @@ if(NOT USE_SYSTEM_MAGNUM)
   set(MAGNUM_WITH_GLFWAPPLICATION OFF CACHE BOOL "" FORCE)
   set(MAGNUM_WITH_EIGEN ON CACHE BOOL "" FORCE) # Eigen integration
   set(MAGNUM_WITH_IMGUI ON CACHE BOOL "" FORCE) # ImGui integration
-  # GltfSceneConverter and KtxImageConverter needed by BatchRendererTest,
-  # GltfSceneConverter is not in magnum-plugins master yet
+  # GltfSceneConverter and KtxImageConverter are needed only by
+  # BatchRendererTest and are optional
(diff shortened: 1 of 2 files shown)


aclegg3 (approved):
Thanks, merge when tests pass. :+1:

Skylion007 (approved):
Looks good to me. There seems to be a few build errors that have popped up though.

PR #1838, opened 2022-08-25
viewer.py: Object QA Application

jrreyna:
## Motivation and Context

Changed branch name so making new draft pull request
Allows us to load in objects from datasets (like a banana, bowl, cup, etc) to simple_room.glb and see how they interact with the table geometry. The table mesh is simple and flat, so it allows us to identify any glaring problems with the geometry or physics of an object from a dataset. You can iterate through each object in the dataset, they will be displayed and rotated in front of you, then you can turn on physics and apply impulses, forces, and torques on them to see how they move. You can also make recordings of your tests by hitting the 'L' key to start and stop recording.

This is the command I use to run it:
python examples/viewer.py --scene ./data/test_assets/scenes/simple_room.glb --dataset ./data/objects/ycb/ycb.scene_dataset_config.json

After "--scene," put the .glb you want to use, and after "--dataset," put the scene_dataset_config.json you want to use. The positions of the agent and objects are hardcoded for simple_room.glb as of now

You can also just run:
python examples/viewer.py and it will just use "./data/test_assets/scenes/simple_room.glb" and "./data/objects/ycb/ycb.scene_dataset_config.json" as defaults

Notable key bindings I added were:

'P': iterate forward through the list of dataset objects and display it
'I': iterate backward through the list of dataset objects and display it
'O': snap the existing object down to the table while turning on physics
'L': used to start and stop recording the screen
'1': apply an impulse to the object (hardcoded to be upwards and a little off its center of mass)
'2': apply a force to the object (hardcoded to be upwards and a little off its center of mass)
'3': apply an impluse torque to the object (hardcoded to be counter-clockwise)
'4': apply a torque to the object (hardcoded to be counter-clockwise)

You can see the other key bindings when you run it, as help text is printed in the terminal

## How Has This Been Tested

Nothing strenuous as of now, just gradual and periodic testing on my (John Reyna) machine as I was writing and debugging

## Types of changes

## Checklist


Skylion007:
The submodules are regressed from master actually.

johnrreyna:
> The submodules are regressed from master actually.

Ugh I don't know how that happened. I must have updated the submodules from a different branch at some point since my last commit. Thanks

PR #1848, opened 2022-08-30
[HM3D] - update hm3d importer to expect another entry after region id

aclegg3:
## Motivation and Context

New semantic txt files will include a region string after the id. This violates an assumption that region id will be the last token in each line.

TODO:
- we should parse the region name

## How Has This Been Tested

locally with new semantic txt file

## Types of changes

## Checklist

### DIFF
--- a/src/esp/scene/HM3DSemanticScene.cpp
+++ b/src/esp/scene/HM3DSemanticScene.cpp
@@ -134,9 +134,9 @@ bool SemanticScene::buildHM3DHouse(std::ifstream& ifs,
                               uint8_t(colorInt & 0xff)};
     // object category will possibly have commas
     const std::string objCategoryName = tokens[1];
-    // room/region is always last token - get rid of first comma
-    int regionID =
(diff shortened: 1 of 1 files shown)


src/esp/scene/HM3DSemanticScene.cpp
```diff
@@ -134,9 +134,9 @@ bool SemanticScene::buildHM3DHouse(std::ifstream& ifs,
                               uint8_t(colorInt & 0xff)};
     // object category will possibly have commas
     const std::string objCategoryName = tokens[1];
-    // room/region is always last token - get rid of first comma
-    int regionID =
-        std::stoi(Cr::Utility::String::trim(tokens[tokens.size() - 1], " ,"));
+
+    // region id comes after instace category raw text
+    int regionID = std::stoi(Cr::Utility::String::trim(tokens[2], " ,"));
```
jturner65, line 139: One thing to note here - the split that creates the `tokens` array is splitting on empty quotes.  If a text string region name, encapsulated within quotes, is added, that will also be an element in this tokens array at idx 3.

Issue #1842, opened 2022-08-29
GUI Text for viewer.py


aclegg3:
## 🚀 Feature
Add a Text GUI interface to viewer.py.

## Motivation

Currently, we use `Magnum::ImGuiIntegration` to render text to the window for viewer.cpp (C++ viewer application). We'd like to have a similar interface for adding text to frames in viewer.py (Python viewer application)

## Additional context

At some point we (@mosra) discussed this and determined there may be some blockers implementing this feature in Python. If so, let's record them here for tracking.

mosra:
I don't remember any blockers for implementing this, really. Except that I'd have to expose a few more Magnum APIs to Python, which is nothing complex on its own. I could do that.

Back then we used ImGui because it allowed to have a tiny FPS counter done with a single line of code ... but at the cost of a massive new dependency, of which we use not even 1%. Because ImGui is a large library on its own, exposing all of it to Python is not simple. There's [PyImGui](https://pyimgui.readthedocs.io/en/latest/) but for that I'd first need to make it work with the Magnum backend, and that's currently an unknown amount of work. Instead, I'd really like to use the far more lightweight `Magnum::Text` library, for both C++ and Python.

aclegg3:
> I don't remember any blockers for implementing this, really. Except that I'd have to expose a few more Magnum APIs to Python, which is nothing complex on its own. I could do that.

That would be great. If the investment is small, I'd like to get this working.

> I'd really like to use the far more lightweight Magnum::Text library, for both C++ and Python.

That sounds fine to me. Remove a dependency and have a better mirror between the C++ and python implementations. Is that ready to use for C++? Looks like we'd need bindings to support python.

mosra:
Yes, that's what I meant -- the `Text` library is ready to be used from C++, it just needs to have its essential functionality exposed to Python. I'll reach out to you on Slack to discuss priorities.

PR #1853, opened 2022-09-02
Use Magnum's builtin text rendering instead of ImGui; expose it in Python

mosra:
## Motivation and Context

As we discussed in #1842, this does the following:

- Drops ImGui in favor of Magnum's own text rendering. All ImGui was ever used for was a single text HUD, so not even 1% of its functionality. IIRC, originally it was chosen because Magnum's own text rendering "seemed hard", but I don't think that's the case -- the amount of code is basically the same for both. But what we gain compared to ImGui is that:
  - The viewer doesn't *need* to constantly redraw to function anymore, which may be especially useful for resource-efficient web / mobile use cases. ImGui needed that in order to handle events and do other things that we didn't make use of.
  - We can easily do text rendering in Python now. With ImGui it'd be about having to rewrite PyImGui with a Magnum backend, or using a completely different solution altogether.
- To keep at least some of the warm fuzzy feeling, I reused `ProggyClean.ttf` that ImGui embeds.
- Besides that, **the [Magnum::Text](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Text.html) library is also now exposed to Python**:
  - [Here is the documentation](https://doc.magnum.graphics/python/magnum/text/)
  - [Here is the Text example, but in Python](https://doc.magnum.graphics/python/examples/#text)
- *Besides that*, this also contains the complete [GltfSceneConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1GltfSceneConverter.html) plugin, which is now capable of saving most scenes. But since it doesn't contain the [related `magnum-sceneconverter` changes](https://tmp.magnum.graphics/tmp/presentation-asset-pipeline/#cover) yet, it's not really useful to anybody, actually. And for the batch renderer use case it also still needs custom scene field support and sampler property export to be finished.

## How Has This Been Tested

Here's how it looks now. Probably not pixel-perfect to what was there before, but I'd say pretty close. Coincidentally, in this particular case both ImGui and Magnum use the exact same font renderer. (I'd personally make the text *significantly* smaller, it's obnoxiously big, let me know if you want that.)

image

If anybody else should be added as a reviewer, please add them ;)


aclegg3 (commented):
Running the viewer after installing on Fedora 36 I get:
```
PluginManager::Manager::Manager(): none of the plugin search paths in {../lib/magnum/fonts, magnum/fonts} exists and pluginDirectory was not set, skipping plugin discovery
PluginManager::Manager::load(): plugin TrueTypeFont is not static and was not found in 
Cannot open font file

```

mosra:
Fixed, sorry! Forgot that you all use a build with bundled Magnum :sweat_smile:

aclegg3 (approved):
Works for me! Thanks for slicing out another dependency.
The text is rather larger now, let's shrink it down a bit.

Skylion007 (approved):

mosra:
@aclegg3 is it better now? I realized too late that the original setting was DPI-dependent and what looked alright on my 4K display with 1.6875x scaling was probably way too huge on non-HiDPI screens.

- - -

Also I added a commit that links the font plugin for Python bindings, but unfortunately due to the *dreaded* problem with native Python modules being isolated from each other, it will only work if the following is done:

```py
import ctypes, sys
flags = sys.getdlopenflags()
sys.setdlopenflags(flags | ctypes.RTLD_GLOBAL)

import magnum
# Now there's no StbTrueTypeFont yet

import habitat_sim
# Now there is, but only thanks to RTLD_GLOBAL
```

An alternative solution would be to force CMake 3.13+ to be used and link the font plugin to the magnum bindings module *from outside*, similarly to how Erik Wijmans did it for the Basis compressor [here](https://github.com/facebookresearch/habitat-sim/blob/2662204daa6d6324ced92b7e9e0ae0365cef9e41/src/utils/imageconverter/CMakeLists.txt#L6-L15). Then the plugin would be available right after `import magnum`.

This is yet another case of the exact same problem stemming from static libraries getting duplicated across multiple shared modules. I'll try to find a manual workaround at least -- something that needs an explicit piece of code but doesn't require fiddling with the CMake build or with dlopen flags.

aclegg3:
> @aclegg3 is it better now? 

Thanks, much better now!

> Also I added a commit that links the font plugin for Python bindings, but unfortunately due to the _dreaded_ problem with native Python modules being isolated from each other, it will only work if the following is done:
> ...

A bit messy, but functional. We can add this chunk to the viewer python with a comment to this effect.

> An alternative solution would be to force CMake 3.13+ to be used ...

We already require CMake 3.14.0, so this doesn't sound like a blocker to me.

mosra:
3.14? Oh, haha, didn't know that. Good then, will change it to link the plugin directly to the magnum module.

Skylion007:
@mosra I'd much rather upgrade it to 3.14. It's a pretty old-CMake and cmake is pretty easy to update use pip install.

Skylion007:
@mosra Happy to bump the cmake version for 3.14. I just checked and the latest OpenEXR is specifically cmake `3.12 OR NEWER`, so let's just bump the cmake version. There are a couple places we should do this, mainly the CricleCI and pyproject.toml.

I'm even fine bumping up to 3.20 as we already require that for our emsdk build.

Skylion007:
`               # static Bullet libs, and Bullet static libs required CMake 3.13+.`
from the CI comments.

PR #1857, opened 2022-09-07
Serialize lights into gfx-replay json format.

0mdc:
## Motivation and Context

This changeset adds light serialization to the gfx-replay format.

Lights are serialized incrementally - they are only written to the json file when lighting changes occur. We assume that the keyframes are always read sequentially.

<details><summary>`GfxReplayTest::testLightIntegration` Replay File</summary>

```json
{
    "keyframes": [
        {
            "lightsChanged": true,
            "lights": [
                {
                    "vector": [
                        1.5,
                        2.0,
                        2.5,
                        1.0
                    ],
                    "color": [
                        1.5,
                        2.0,
                        5.0
                    ],
                    "model": "global"
                },
                {
                    "vector": [
                        -10.0,
                        4.25,
                        10.0,
                        1.0
                    ],
                    "color": [
                        5.0,
                        5.0,
                        0.0
                    ],
                    "model": "camera"
                }
            ]
        },
        {
            "lightsChanged": true,
            "lights": [
                {
                    "vector": [
                        0.0,
                        1.2,
                        -4.0,
                        1.0
                    ],
                    "color": [
                        4.0,
                        4.0,
                        4.0
                    ],
                    "model": "object"
                }
            ]
        },
        {},
        {
            "lightsChanged": true,
            "lights": [
                {
                    "vector": [
                        0.1,
                        0.2,
                        -0.3,
                        0.0
                    ],
                    "color": [
                        0.0,
                        0.0,
                        1.0
                    ],
                    "model": "global"
                }
            ]
        },
        {
            "lightsChanged": true
        }
    ]
}
```
</details>

## How Has This Been Tested

- New unit test (`GfxReplayTest::testLightIntegration`).

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/replay/Keyframe.h
+++ b/src/esp/gfx/replay/Keyframe.h
@@ -55,7 +55,7 @@ struct Keyframe {
   std::vector<std::pair<RenderAssetInstanceKey, RenderAssetInstanceState>>
       stateUpdates;
   std::unordered_map<std::string, Transform> userTransforms;
-  std::vector<LightInfo> lights;
+  Cr::Containers::Optional<std::vector<LightInfo>> lights;
 
   ESP_SMART_POINTERS(Keyframe)
 };
--- a/src/esp/gfx/replay/Keyframe.h
+++ b/src/esp/gfx/replay/Keyframe.h
@@ -55,9 +55,8 @@ struct Keyframe {
   std::vector<std::pair<RenderAssetInstanceKey, RenderAssetInstanceState>>
       stateUpdates;
   std::unordered_map<std::string, Transform> userTransforms;
-  // Contains the full set of lights in the keyframe.
-  // When NullOpt, lights have not changed from the previous keyframe.
-  Cr::Containers::Optional<std::vector<LightInfo>> lights;
+  std::vector<LightInfo> lights;
+  bool lightsChanged = false;
 
   ESP_SMART_POINTERS(Keyframe)
 };
--- a/src/esp/gfx/replay/Keyframe.h
+++ b/src/esp/gfx/replay/Keyframe.h
@@ -55,6 +55,8 @@ struct Keyframe {
   std::vector<std::pair<RenderAssetInstanceKey, RenderAssetInstanceState>>
       stateUpdates;
   std::unordered_map<std::string, Transform> userTransforms;
+  // Contains the full set of lights in the keyframe.
+  // When NullOpt, lights have not changed from the previous keyframe.
   Cr::Containers::Optional<std::vector<LightInfo>> lights;
 
   ESP_SMART_POINTERS(Keyframe)
--- a/src/esp/gfx/replay/Keyframe.h
+++ b/src/esp/gfx/replay/Keyframe.h
@@ -55,6 +55,7 @@ struct Keyframe {
   std::vector<std::pair<RenderAssetInstanceKey, RenderAssetInstanceState>>
       stateUpdates;
   std::unordered_map<std::string, Transform> userTransforms;
+  std::vector<LightInfo> lights;
 
   ESP_SMART_POINTERS(Keyframe)
 };
--- a/src/esp/io/JsonEspTypes.cpp
+++ b/src/esp/io/JsonEspTypes.cpp
@@ -11,9 +11,7 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
                              JsonAllocator& allocator) {
   JsonGenericValue obj(rapidjson::kObjectType);
 
-  if (!keyframe.loads.empty()) {
-    io::addMember(obj, "loads", keyframe.loads, allocator);
-  }
+  io::addMember(obj, "loads", keyframe.loads, allocator);
 
   if (!keyframe.creations.empty()) {
     JsonGenericValue creationsArray(rapidjson::kArrayType);
@@ -27,9 +25,7 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
     io::addMember(obj, "creations", creationsArray, allocator);
   }
 
-  if (!keyframe.deletions.empty()) {
-    io::addMember(obj, "deletions", keyframe.deletions, allocator);
-  }
+  io::addMember(obj, "deletions", keyframe.deletions, allocator);
 
   if (!keyframe.stateUpdates.empty()) {
     JsonGenericValue stateUpdatesArray(rapidjson::kArrayType);
@@ -54,6 +50,7 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
   }
 
   if (keyframe.lightsChanged) {
+    io::addMember(obj, "lightsChanged", true, allocator);
     io::addMember(obj, "lights", keyframe.lights, allocator);
   }
 
@@ -106,9 +103,8 @@ bool fromJsonValue(const JsonGenericValue& obj,
     }
   }
 
-  itr = obj.FindMember("lights");
-  if (itr != obj.MemberEnd()) {
-    keyframe.lightsChanged = true;
+  io::readMember(obj, "lightsChanged", keyframe.lightsChanged);
+  if (keyframe.lightsChanged) {
     io::readMember(obj, "lights", keyframe.lights);
   }
 
--- a/src/esp/io/JsonEspTypes.cpp
+++ b/src/esp/io/JsonEspTypes.cpp
@@ -49,6 +49,16 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
     io::addMember(obj, "userTransforms", userTransformsArray, allocator);
   }
 
+  if (!keyframe.lights.empty()) {
+    JsonGenericValue lightsArray(rapidjson::kArrayType);
+    for (const auto& light : keyframe.lights) {
+      JsonGenericValue lightObj(rapidjson::kObjectType);
+      io::addMember(lightObj, "light", light, allocator);
+      lightsArray.PushBack(lightObj, allocator);
+    }
+    io::addMember(obj, "lights", lightsArray, allocator);
+  }
+
   return obj;
 }
 
@@ -95,6 +105,16 @@ bool fromJsonValue(const JsonGenericValue& obj,
       io::readMember(userTransformObj, "name", name);
       io::readMember(userTransformObj, "transform", transform);
       keyframe.userTransforms[name] = transform;
+    }
+  }
+
+  itr = obj.FindMember("lights");
+  if (itr != obj.MemberEnd()) {
+    const JsonGenericValue& lightsArray = itr->value;
+    for (const auto& lightObj : lightsArray.GetArray()) {
+      gfx::LightInfo light;
+      io::readMember(lightObj, "light", light);
+      keyframe.lights.emplace_back(std::move(light));
     }
   }
 
--- a/src/esp/io/JsonEspTypes.cpp
+++ b/src/esp/io/JsonEspTypes.cpp
@@ -11,7 +11,9 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
                              JsonAllocator& allocator) {
   JsonGenericValue obj(rapidjson::kObjectType);
 
-  io::addMember(obj, "loads", keyframe.loads, allocator);
+  if (!keyframe.loads.empty()) {
+    io::addMember(obj, "loads", keyframe.loads, allocator);
+  }
 
   if (!keyframe.creations.empty()) {
     JsonGenericValue creationsArray(rapidjson::kArrayType);
@@ -25,7 +27,9 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
     io::addMember(obj, "creations", creationsArray, allocator);
   }
 
-  io::addMember(obj, "deletions", keyframe.deletions, allocator);
+  if (!keyframe.deletions.empty()) {
+    io::addMember(obj, "deletions", keyframe.deletions, allocator);
+  }
 
   if (!keyframe.stateUpdates.empty()) {
     JsonGenericValue stateUpdatesArray(rapidjson::kArrayType);
@@ -50,7 +54,6 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
   }
 
   if (keyframe.lightsChanged) {
-    io::addMember(obj, "lightsChanged", true, allocator);
     io::addMember(obj, "lights", keyframe.lights, allocator);
   }
 
@@ -103,8 +106,9 @@ bool fromJsonValue(const JsonGenericValue& obj,
     }
   }
 
-  io::readMember(obj, "lightsChanged", keyframe.lightsChanged);
-  if (keyframe.lightsChanged) {
+  itr = obj.FindMember("lights");
+  if (itr != obj.MemberEnd()) {
+    keyframe.lightsChanged = true;
     io::readMember(obj, "lights", keyframe.lights);
   }
 
--- a/src/esp/io/JsonEspTypes.cpp
+++ b/src/esp/io/JsonEspTypes.cpp
@@ -49,9 +49,9 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
     io::addMember(obj, "userTransforms", userTransformsArray, allocator);
   }
 
-  if (!keyframe.lights.empty()) {
+  if (keyframe.lights && !keyframe.lights->empty()) {
     JsonGenericValue lightsArray(rapidjson::kArrayType);
-    for (const auto& light : keyframe.lights) {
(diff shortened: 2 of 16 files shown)


src/tests/GfxReplayTest.cpp
```diff
@@ -449,6 +456,79 @@ void GfxReplayTest::testSimulatorIntegration() {
...
+  // record a playback file
+  {
+    SimulatorConfiguration simConfig{};
+    simConfig.enableGfxReplaySave = true;
+    simConfig.createRenderer = false;
+    simConfig.enablePhysics = false;
+    auto sim = Simulator::create_unique(simConfig);
+    CORRADE_VERIFY(sim);
+
+    sim->setLightSetup(lightSetup, lightSetupKey);
```
eundersander, line 481: How about multiple calls to `setLightSetup` (with differing setups) and `saveKeyframe`? And then we can verify that lights are getting dynamically added and removed properly during playback.

0mdc: I updated the test accordingly.

src/tests/GfxReplayTest.cpp
```diff
@@ -449,6 +456,101 @@ void GfxReplayTest::testSimulatorIntegration() {
...
+  const LightInfo pointLight2{Magnum::Vector4{0.0f, 1.2f, -4.0f, 1.0f},
+                              {4.0, 4.0, 4.0},
+                              LightPositionModel::Global};
+  const LightInfo pointLight3{Magnum::Vector4{0.1f, 0.2f, -0.3f, 1.0f},
+                              {0.0, 0.0, 1.0},
+                              LightPositionModel::Global};
+  const LightSetup lightSetup0{pointLight0, pointLight1};
+  const LightSetup lightSetup1{pointLight2};
+  const LightSetup lightSetup2{pointLight3};
+  const std::string lightSetup0Key = "custom_lighting_0";
```
eundersander, line 492: Ah, another thing, let's only use the default light setup key. I believe the key is `""`, or omit the key argument altogether in the API calls. I discussed this in the design doc.

The problem with using three different light setup keys is that we never actually change/delete lights unless we repeatedly call `setLightSetup` for the same key.

0mdc:
Added light serialization. Two issues remain:

- Light position model enum is not safely serialized.
- Absolute light states are saved rather than deltas. If lights are unchanged for the duration of the replay, they will show up for every keyframe.

src/esp/io/JsonMagnumTypes.h
```diff
@@ -65,6 +65,68 @@ inline bool fromJsonValue(const JsonGenericValue& obj, Magnum::Vector3& val) {
...
+/**
+ * @brief Specialization to handle Magnum::Color3 values. Populate passed @p
+ * val with value. Returns whether successfully populated, or not. Logs an error
+ * if inappropriate type.
+ *
+ * @param obj json value to parse
+ * @param val destination value to be populated
+ * @return whether successful or not
+ */
+inline bool fromJsonValue(const JsonGenericValue& obj, Magnum::Color3& val) {
```
mosra, line 112: To avoid code repetition you should be able to do something like this here, the `Color3`  is a direct subclass of `Vector3`:

```cpp
inline bool fromJsonValue(const JsonGenericValue& obj, Magnum::Color3& val) {
    return fromJsonValue(obj, static_cast<Magnum::Vector3&>(val));
}
```

Actually, it feels to me like this overload shouldn't be needed at all, the compiler could pick the `Vector3` variant on its own, I hope?

Same would be with the `Color4` variant below, which could now be taken care of with the `Vector4` variant, and I think the `toJsonValue()` for `Color3` and `Color4` wouldn't need to exist either?

(Personal feeling and [not a shameless self-promo at all](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Json.html#Utility-Json-usage-direct-array-access): wow, rapidjson is *verbose*...)

src/tests/GfxReplayTest.cpp
```diff
@@ -465,6 +476,107 @@ void GfxReplayTest::testSimulatorIntegration() {
...
+      CORRADE_COMPARE(a[i].vector, b[i].vector);
+      CORRADE_COMPARE(static_cast<int>(a[i].model),
+                      static_cast<int>(b[i].model));
+      CORRADE_COMPARE(a[i].color, b[i].color);
+    }
+  };
+
+  const auto testFilepath =
+      Corrade::Utility::Path::join(DATA_DIR, "./gfx_replay_test.json");
+  const LightInfo pointLight0{Magnum::Vector4{1.5f, 2.0f, 2.5f, 1.0f},
```
mosra, line 493:
```suggestion
  const LightInfo pointLight0{{1.5f, 2.0f, 2.5f, 1.0f},
```

should work too I think, same below

src/esp/io/JsonEspTypes.cpp
```diff
@@ -49,6 +49,16 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
     io::addMember(obj, "userTransforms", userTransformsArray, allocator);
   }
 
+  if (keyframe.lights && !keyframe.lights->empty()) {
```
eundersander, line 52: How does this logic differentiate an empty vector (a light setup with no lights) from a `nullopt` (no light setup)?

src/esp/io/JsonEspTypes.cpp
```diff
@@ -49,6 +49,16 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
...
+    JsonGenericValue lightsArray(rapidjson::kArrayType);
```
eundersander, line 53: We shouldn't have to write all this tedious serialization. `io::addMember` should already "just work" for Corrade::Optional and std::vector. Can this be simplified to `io::addMember(obj, keyframe.lights)`?

src/esp/io/JsonEspTypes.cpp
```diff
@@ -98,6 +108,17 @@ bool fromJsonValue(const JsonGenericValue& obj,
     }
   }
 
+  itr = obj.FindMember("lights");
```
eundersander, line 111: Likewise here, `io::readMember(keyframe.lights)` should "just work".

src/esp/io/JsonEspTypes.cpp
```diff
@@ -146,16 +167,43 @@ bool fromJsonValue(const JsonGenericValue& obj,
...
   return shaderTypeSucceess;
 }
 
+JsonGenericValue toJsonValue(const esp::gfx::LightPositionModel& x,
+                             JsonAllocator& allocator) {
+  return toJsonValue(metadata::attributes::getLightPositionModelName(x),
+                     allocator);
+}
+
+bool fromJsonValue(const JsonGenericValue& obj,
```
eundersander, line 187: Ug, serializing enums as strings like this is *so* much boilerplate code. We should find a better long-term solution.

src/esp/gfx/replay/Keyframe.h
```diff
@@ -55,6 +55,7 @@ struct Keyframe {
   std::vector<std::pair<RenderAssetInstanceKey, RenderAssetInstanceState>>
       stateUpdates;
   std::unordered_map<std::string, Transform> userTransforms;
+  Cr::Containers::Optional<std::vector<LightInfo>> lights;
```
eundersander, line 58: I'm a little torn on serializing our entire quirky `LightInfo` struct (including LightPositionModel). I talked in the design doc about wanting gfx-replay to be agnostic to the renderer, and most renderers IMO won't offer "object-local" or "camera-local" lights. I think we only need to support lights specified in global space. Also, I don't think anybody is using object-local or camera-local lights in Habitat. I don't think we'll bother implementing them in the new batch renderer.

See also this discussion here: https://cvmlp.slack.com/archives/G0131KVLBLL/p1660649432149979 .

It is probably fine to leave this as-is, but let's beware introducing unneeded complexity into the gfx-replay format in general.

eundersander (commented):
Looking good so far. I left a few comments.

src/esp/gfx/replay/Keyframe.h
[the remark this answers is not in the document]
0mdc: That's correct.

I picked this format because it's simple, and because we rarely change lights during a simulation. Therefore, eliminating duplicates from consecutive light set changes won't optimize the replay file size in most cases.

I'll add a comment to clarify how we use the data here.

src/esp/io/JsonMagnumTypes.h
0mdc: > Actually, it feels to me like this overload shouldn't be needed at all, the compiler could pick the Vector3 variant on its own, I hope?

The compiler picks `Color3` 🙁.

> (Personal feeling and [not a shameless self-promo at all](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Json.html#Utility-Json-usage-direct-array-access): wow, rapidjson is verbose...)

We definitely could look into using Corrade for this. It might ease up future extensibility, especially as we expect to expand these functionalities. It would also take out one dependency.

mosra: > The compiler picks Color3 :slightly_frowning_face:.

Ah well, so if you don't have the `fromJsonValue(..., Color3&)` overload at all, it'll fail to compile? Strange, I thought it would just go with the base class. Or maybe my understanding of how overload resolution works for mutable references is wrong.

In any case, the explicit `static_cast` would reduce the code duplication at least partially.

src/esp/io/JsonEspTypes.cpp
mosra: Unhelpful suggestion: store it as numbers? glTF does the same for various data types, sampler modes etc, and that simplifies parsing and writing tremendously. A downside is that you have to ensure that the numbering doesn't change in the future, and the JSON is not that much human-readable after.

src/esp/io/JsonEspTypes.cpp
```diff
@@ -49,6 +49,16 @@ JsonGenericValue toJsonValue(const gfx::replay::Keyframe& keyframe,
     io::addMember(obj, "userTransforms", userTransformsArray, allocator);
   }
 
+  if (keyframe.lights && !keyframe.lights->empty()) {
+    JsonGenericValue lightsArray(rapidjson::kArrayType);
+    for (const auto& light : *keyframe.lights) {
+      JsonGenericValue lightObj(rapidjson::kObjectType);
+      io::addMember(lightObj, "light", light, allocator);
+      lightsArray.PushBack(lightObj, allocator);
```
Skylion007, line 57:
```suggestion
      lightsArray.PushBack(std::move(lightObj), allocator);
```
I am not sure if it actually defines a move ctor, but if clang-tidy doesn't complain this should be what you want to do.

src/esp/io/JsonMagnumTypes.h
```diff
@@ -65,6 +65,68 @@ inline bool fromJsonValue(const JsonGenericValue& obj, Magnum::Vector3& val) {
...
+  if (obj.IsArray() && obj.Size() == 3) {
+    Magnum::Vector3 vec3;
+    for (rapidjson::SizeType i = 0; i < 3; ++i) {
+      if (obj[i].IsNumber()) {
+        vec3[i] = obj[i].GetDouble();
```
Skylion007, line 117: Vector3 stores floats right? Why do we keep calling getDouble? Are we also serializing it as a double? These storing floats as doubles add up.

mosra: JSON's only numeric type is doubles, yes. Not sure if rapidjson even has 32-bit-float APIs, usually there isn't any significant difference in parsing as a float or a double.

src/esp/io/JsonEspTypes.cpp
0mdc: The `Optional` `bool()` operator [is overridden to check whether the container has a value](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1Optional.html#a389cd0c17eac32b9194f922fccc91934), much like [the implementation of `std::optional`](https://en.cppreference.com/w/cpp/utility/optional/operator_bool).

src/esp/io/JsonMagnumTypes.h
Skylion007: Oh darn, didn't know that about JSON. Yet another reason to avoid using it for our dataset.jsons in Habitat-Lab. 0-0 You can set the max decimal places in RapidJSON globally it appears, but other than that no. :(

src/esp/io/JsonEspTypes.cpp
0mdc: `clang-tidy` complained, so I removed the `std::move` here.

src/esp/gfx/replay/Keyframe.h
0mdc: I completely agree.

I feel like it doesn't hurt to leave the `LightPositionModel` field as-is for now. We should look into specifying a standard light data layout as we integrate them with the batch renderer.

I am still unfamiliar with the batch renderer, but if we are following a pure ECS approach, we might want to separate the transform component from the light data.

>I think we only need to support lights specified in global space.

Object-local lights may be occasionally useful (e.g. if the simulated agent has a light). However, I don't think that we are doing anything like that at the moment.
However, I don't know how camera-local lights would be useful.

src/esp/io/JsonEspTypes.cpp
0mdc: I think that this would make sense. It would yield smaller json files and would simplify the code. We can write unit tests to catch issues with backwards compatibility.

src/esp/io/JsonEspTypes.cpp
Skylion007: Ah, so it is trivially copyable. NVM

src/esp/io/JsonEspTypes.cpp
jturner65: Hi!  FWIW all the values backed by string->enum maps in the metadata were so chosen to facilitate user field modification in the JSON itself, which was expected to occur in the various config files these enum-backed string fields were supported in.  

I'm not sure if that is still the intent for these fields, but understand that changing the json values of any of these from strings to ints, while addressing some cumbersome parsing code, may also invalidate many extant config files, both in datasets and in use by folks.

0mdc: In this case, it's probably better to keep things as-is for now.

In the future, we could include a version to the json files so that we can run them through a backwards compatibility layer.

jturner65: Until then, it would also be very useful, i think, to put these the String->enum mapping function right next to the enum->string versions in metadata::attributes namespace (AttributesEnumMaps.h/cpp file).

src/esp/io/JsonMagnumTypes.h
eundersander: Related: see `usePrettyWriter` and `maxDecimalPlaces` in `writeJsonToFile`. I added this in a half-hearted attempt to minimize the filesize of gfx-replay files.

src/esp/io/JsonEspTypes.cpp
0mdc: I changed the approach to include a dirty flag, `lightsChanged`, instead of using `Corrade::Optional` to wrap the vector.

src/esp/io/JsonEspTypes.cpp
```diff
@@ -103,8 +106,9 @@ bool fromJsonValue(const JsonGenericValue& obj,
     }
   }
 
-  io::readMember(obj, "lightsChanged", keyframe.lightsChanged);
-  if (keyframe.lightsChanged) {
+  itr = obj.FindMember("lights");
```
eundersander, line 109: Hmmm, I'm not thrilled about this. I have a dream of one day implementing a new backend for `io::addMember` and `io::readMember` (e.g. serialize to binary, or some other format). So I generally don't like to see us using JSON-specific APIs like `FindMember` inside these functions.

I realize I'm being kinda picky. Let's discuss over VC soon and figure out the best way forward.

src/esp/io/JsonStlTypes.h
```diff
@@ -53,10 +53,8 @@ void addMember(JsonGenericValue& value,
                rapidjson::GenericStringRef<char> name,
                const std::vector<T>& vec,
                JsonAllocator& allocator) {
-  if (!vec.empty()) {
-    addMember(value, name, toJsonArrayHelper(vec.data(), vec.size(), allocator),
-              allocator);
-  }
+  addMember(value, name, toJsonArrayHelper(vec.data(), vec.size(), allocator),
```
eundersander, line 56: I know I kinda suggested this earlier, but now I'm having second thoughts. See this discussion here: https://stackoverflow.com/questions/9619852/what-is-the-convention-in-json-for-empty-vs-null

Let's discuss more over VC soon.

0mdc: As discussed, I reverted the last commit so that a `lightsChanged` flag is inserted when changes occur. This is simpler and safer.

eundersander (approved):
LGTM!

PR #1849, opened 2022-08-30
[WIP] Sensor update and refactor

jrreyna:
## Motivation and Context

Refactoring sensor code in python to mirror what the C++ code does. Continuing work from pull request #1151. 

## How Has This Been Tested

Not tested yet, will use "python tests in tests/" and C++ tests in "src/tests/"

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1858, opened 2022-09-08
Chore: Remove unnecessary casts from SensorSpec bindings

Skylion007:
## Motivation and Context

While reviewing #1849 I noticed that we are doing some very inefficent things for modifying the attributes on the Python object. Mainly we are casting to a C++ object and then back again. Since we know the first argument will be an injected "self" argument, that it will be non-null, and it will manage it's own ownership we can just use a lightweight handle pointing to the underlying py::object. This means we don't have to do any casting and can just call the relevant C++ python API directly. 
## How Has This Been Tested

Locally by running the test_sensors tests.

## Types of changes

## Checklist


aclegg3 (approved):
Cool.

PR #1832, opened 2022-08-17
[WIP] Move settings.py to habitat_sim.utils

jaraujo98:
## Motivation and Context

One task I have to do frequently is to start a “default” simulator (mostly for writing custom viewers). The only option I am aware of is to use the code in `habitat-sim/examples/settings.py` to get a default configuration. Having to rely on this file has been a pain point. I have to either run my scripts from the examples folder or the sim folder (this leads to inconsistent imports: on some files the statement is `from settings import ...`, on others it is `from examples.settings import ...`). I can also play around with `sys.path` or `PYTHONPATH`, but this not a viable solution since it can clash with other packages that also have folders named `examples` or files named `settings`.

This PR moves the `settings.py` file into `habitat_sim.utils`, thus making its functionality unambiguously available everywhere. I kept a file named `settings.py` inside `examples` that simply imports the original settings file to make sure existing scripts that rely on it don't break.

## How Has This Been Tested

Tests (some of which rely on this file) pass. The existing examples also continue working.

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1862, opened 2022-09-13
Viewer py add magnum text

jrreyna:
## Motivation and Context

Using the magnum text code merged in PR #1853 to add text to the viewer.py app. I had to update the settings.py code. For some reason, it was drastically different from what I originally used for my the code in my #1838 PR so it wasn't working. It couldn't find "habitat_sim.utils.settings" so I had to change it to "examples.settings." I think the former refers to the settings file used for viewer.cpp

## How Has This Been Tested

Locally, and with the CI tests. Here are two screenshots that show the mouse mode changing. We don't have python bindings yet to expose frame profiler functionality

Screenshot from 2022-09-26 15-01-21
Screenshot from 2022-09-26 15-03-12

## Types of changes

## Checklist


Skylion007:
Seems like you accidentally reverting some submodules? Like pybind11?

jrreyna:
> Seems like you accidentally reverting some submodules? Like pybind11?

Yes I have no idea what is happening. I made a branch from origin/main, and it says I am up to date, but the magnum-bindings submodule is also not correct. I am missing several files and changes that I see were pushed in the #1853 PR. I also tried the command  `git submodule update src/deps/magnum-bindings` but it is not doing anything

Skylion007:
> > Seems like you accidentally reverting some submodules? Like pybind11?
> 
> Yes I have no idea what is happening. I made a branch from origin/main, and it says I am up to date, but the magnum-bindings submodule is also not correct. I am missing several files and changes that I see were pushed in the #1853 PR. I also tried the command `git submodule update src/deps/magnum-bindings` but it is not doing anything

You need to cd into those folders, git checkout the proper commits (and then git add the submodule folder) (or git checkout the .submodule file directly and then do the git submodule update).

examples/viewer.py
```diff
@@ -937,14 +939,29 @@ def next_frame() -> None:
...
 
     # Setting up sim_settings
     sim_settings: Dict[str, Any] = default_sim_settings
-    sim_settings["scene"] = args.scene
-    sim_settings["scene_dataset_config_file"] = args.dataset
+
+    # this interactive viewer was developed with these arguments,
+    # so make them the default if none provided
+    if args.scene == "NONE":
+        scene_name = "./data/test_assets/scenes/simple_room.glb"
```
ykarmesh, line 951: Why dont you use `./data/test_assets/scenes/simple_room.glb` as the value of `default` value for `--scene` in the argument parser?

examples/viewer.py
```diff
@@ -937,14 +939,29 @@ def next_frame() -> None:
...
+    else:
+        scene_name = args.scene
+
+    if args.dataset == "default":
+        dataset_name = "./data/objects/ycb/ycb.scene_dataset_config.json"
```
ykarmesh, line 956: Same as the scene recommendation

examples/viewer.py
```diff
@@ -15,13 +15,15 @@
 
 import magnum as mn
 import numpy as np
+from magnum import *
```
ykarmesh, line 18: Avoid using wildcard import

examples/settings.py
```diff
@@ -3,11 +3,31 @@
...
-example_settings = {
+default_sim_settings = {
+    # settings shared by example.py and benchmark.py
     "max_frames": 1000,
+    "width": 640,
+    "height": 480,
+    "default_agent": 0,
+    "sensor_height": 1.5,
+    "hfov": 90,
+    "color_sensor": True,  # RGB sensor (default: ON)
```
ykarmesh, line 16: Comments are a little redundant, can be removed imo

examples/settings.py
```diff
@@ -3,11 +3,31 @@
...
-from habitat_sim.utils.settings import make_cfg as _make_cfg
+import habitat_sim.agent
 
-example_settings = {
+default_sim_settings = {
+    # settings shared by example.py and benchmark.py
     "max_frames": 1000,
+    "width": 640,
+    "height": 480,
+    "default_agent": 0,
```
ykarmesh, line 13: What is this config for?

examples/viewer.py
jrreyna: ah, I didn't see that was an option here until now 😅:
parser.add_argument(
        "--scene",
        default="NONE",
        type=str,
        help='scene/stage file to load (default: "NONE")',
    )

examples/settings.py
jrreyna: `default_agent` is used in `viewer.py` and `width`/`height` are used in other scripts in the `examples` directory

examples/viewer.py
```diff
@@ -733,6 +783,51 @@ def exit_event(self, event: Application.ExitEvent):
...
+            f"""
+{self.fps} FPS
+{placeholder} drawables, {placeholder} culled
+{sensor_subtype_string}
+Last {self.num_frames_to_track} frames:
+    Frame time: {placeholder} ms
+    CPU duration: {placeholder} ms
+    GPU duration: {placeholder} µs
+    Vertex fetch ratio: {placeholder}
+    Primitives clipped: {placeholder} %
```
mosra, line 811: Ah yes, I should expose the FrameProfiler to Python as well. Adding a TODO for myself.

examples/viewer.py
```diff
@@ -73,6 +76,50 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
...
+            os.path.join(os.path.dirname(__file__), "Cousine-Regular.ttf"), 180.0
+        )
+
+        # Glyphs we need to render everything
+        self._cache = text.DistanceFieldGlyphCache(
+            mn.Vector2i(2048), mn.Vector2i(512), 22
+        )
+        self._font.fill_glyph_cache(
+            self._cache,
+            "abcdefghijklmnopqrstuvwxyz"
```
Skylion007, line 91:
```suggestion
            string.ascii_lowercase,
            string.ascii_uppercase
```
You will need to `import string`

examples/viewer.py
```diff
@@ -733,6 +783,51 @@ def exit_event(self, event: Application.ExitEvent):
...
+    GPU duration: {placeholder} µs
+    Vertex fetch ratio: {placeholder}
+    Primitives clipped: {placeholder} %
+Mouse Interaction Mode: {mouse_mode_string}
+            """
+        )
+        self._shader.draw(self._window_text.mesh)
+
+    def get_sensor_subtype_string(self, sensor_subtype):
+        if sensor_subtype == SensorSubType.PINHOLE:
```
Skylion007, line 818: It's not quite as pretty, but can't we just figure out a way to stringify the name of the enum? Makes it easier to extend. Something like
```python
    return str(sensor_subtype.value)
```

examples/viewer.py
```diff
@@ -73,6 +76,50 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
...
+        )
+
+        # Glyphs we need to render everything
+        self._cache = text.DistanceFieldGlyphCache(
+            mn.Vector2i(2048), mn.Vector2i(512), 22
+        )
+        self._font.fill_glyph_cache(
+            self._cache,
+            "abcdefghijklmnopqrstuvwxyz"
+            "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
Skylion007, line 92:
```suggestion
            string.ascii_lowercase,
            string.ascii_uppercase,
```
You will need to `import string`

examples/viewer.py
```diff
@@ -73,6 +76,50 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
...
+            "0123456789:-_+,.! %µ",
```
Skylion007, line 93: Wouldn't this be sufficient for the glyph cache?
```suggestion
            string.printable,
```

examples/viewer.py
jrreyna: Ah thank you, I was looking into the gl module in Python to see if there were any profiling options, but just decided to bring it up later with the team

examples/viewer.py
jrreyna: Yeah probably, I can check it out. I was just using the code from the magnum example page, so there are probably improvements that can be made for our specific purposes

examples/viewer.py
jrreyna: If that is possible in Python, that would be sweet. I'll look into it. I just remember trying to do that in C++ one time and it was tough

examples/viewer.py
mosra: There's (currently) an issue where if you put too many glyphs in, it silently fails to populate the glyph cache. I'll fix that eventually but for now I'd say safest is to have the characters strictly enumerated like this. Once I have that fixed, I can clean this code up.

jaraujo98:
`habitat_sim.utils.settings` came from PR #1832. Since you have the viewer code from that PR, my guess is that you installed `habitat_sim` using `python setup.py install` (which copies the contents of `src_python` to the Python `site-packages` folder and then ignores all changes that you `git pull`) instead of `python setup.py develop` (which creates a link to the `src_python` code in your local clone of the repo, which can be kept up to date).

jrreyna:
@jaraujo98 

> `habitat_sim.utils.settings` came from PR #1832. Since you have the viewer code from that PR, my guess is that you installed `habitat_sim` using `python setup.py install` (which copies the contents of `src_python` to the Python `site-packages` folder and then ignores all changes that you `git pull`) instead of `python setup.py develop` (which creates a link to the `src_python` code in your local clone of the repo, which can be kept up to date).

Oh thanks for pointing that out. Yeah, the code didn't recognize habitat_sim.utils.settings so I was confused. I knew there was a settings.py file in the example directory that I was using beforehand, so I just used that one. So what do you recommend? I imagine changing the `examples.settings` import line to a `habitat_sim.utils.settings` import. Should I rebuild it using `python setup.py develop` instead of install? Do I run `python setup.py develop --bullet`?

jaraujo98:
> Do I run python setup.py develop --bullet?

Yes, that's what I use. You can `git checkout main examples/settings.py` to restore `settings.py` to how it was before, and then use `from habitat_sim.utils.settings import default_sim_settings, make_cfg` in `viewer.py`. Sorry for the confusion... XD

jrreyna:
> > Do I run python setup.py develop --bullet?
> 
> Yes, that's what I use. You can `git checkout main examples/settings.py` to restore `settings.py` to how it was before, and then use `from habitat_sim.utils.settings import default_sim_settings, make_cfg` in `viewer.py`. Sorry for the confusion... XD

No worries that makes sense. So the settings.py file that `habitat_sim.utils.settings` refers to is the file `/home/johnreyna/dev/habitat-sim/src_python/habitat_sim/utils/settings.py`, right?

jaraujo98:
Yes, that's correct. That settings file has the bare minimum settings you need to start a viewer. Everything else that is added in `examples.settings` is for the demo runner and the benchmark scripts, so you don't really need it.

jrreyna:
> Yes, that's correct. That settings file has the bare minimum settings you need to start a viewer. Everything else that is added in `examples.settings` is for the demo runner and the benchmark scripts, so you don't really need it.

It seems that `./examples/settings.py` has the bare minimum and `./src_python/habitat_sim/utils/settings.py` is the one with a lot more code. Is that what you meant?

aclegg3:
Where are these extra fonts coming from? Are they included on purpose?

johnrreyna:
> Where are these extra fonts coming from? Are they included on purpose?

Yeah, I just added them to this directory in case the font directory changes and the default font is no longer in the right place. I want a default font loaded so that the user doesn't always have to provide one in the command line, and right now it just loads it from the current directory. I can remove them and just provide the relative path of the font to the function that loads it in _init_

examples/viewer.py
Skylion007: Can still be simplified using
```suggestion
string.ascii_lowercase + string.ascii_uppercase +  string.digits + ":-_+,.! %µ"
```

jrreyna: Fo sho. I will make that change

src_python/habitat_sim/utils/settings.py
```diff
@@ -13,12 +13,12 @@
     "default_agent": 0,
     "sensor_height": 1.5,
     "hfov": 90,
-    "color_sensor": True,  # RGB sensor (default: ON)
-    "semantic_sensor": False,  # semantic sensor (default: OFF)
-    "depth_sensor": False,  # depth sensor (default: OFF)
-    "ortho_rgba_sensor": False,  # Orthographic RGB sensor (default: OFF)
-    "ortho_depth_sensor": False,  # Orthographic depth sensor (default: OFF)
-    "ortho_semantic_sensor": False,  # Orthographic semantic sensor (default: OFF)
+    "color_sensor": True,
```
Skylion007, line 16: Comments should be repaired / replaced?

Skylion007:
Why can't we use the same proggytff file that C++ user is using btw?

src_python/habitat_sim/utils/settings.py
johnrreyna: Karmesh suggested I remove them bc they are kind of redundant.  I can replace them though if anyone finds them useful

johnrreyna:
> Why can't we use the same proggytff file that C++ user is using btw?

That proggy ttf file was copied directly into the same directory as viewer.cpp, so I did something similar and copied the file into the same directory as viewer.py. Though I think the best solution is to have a fonts directory and specify the relative path to it.

src/utils/viewer/resources.conf
```diff
@@ -1,4 +1,4 @@
 group=fonts
 
 [file]
-filename=ProggyClean.ttf
+filename=../../../data/fonts/ProggyClean.ttf
```
mosra, line 4: To avoid having to hardcode the ugly relative path in `viewer.cpp`, add an alias here ([docs](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Resource.html#Utility-Resource-compilation)):

```suggestion
filename=../../../data/fonts/ProggyClean.ttf
alias=ProggyClean.ttf
```

src/utils/viewer/viewer.cpp
```diff
@@ -899,7 +899,8 @@ Viewer::Viewer(const Arguments& arguments)
 
     Cr::Utility::Resource rs{"fonts"};
     font_ = fontManager_.loadAndInstantiate("TrueTypeFont");
-    if (!font_ || !font_->openData(rs.getRaw("ProggyClean.ttf"), fontSize))
+    std::string font_relative_path = "../../../data/fonts/ProggyClean.ttf";
+    if (!font_ || !font_->openData(rs.getRaw(font_relative_path), fontSize))
```
mosra, line 903: ... and this change then isn't needed.

src/utils/viewer/resources.conf
jrreyna: oo that's slick. Thanks, I wasn't exactly sure how the resources.conf file worked

src/utils/viewer/viewer.cpp
jrreyna: Do you know if I can do something similar in viewer.py to avoid hardcoding the ugly relative path?

ykarmesh (approved):
@aclegg3 @johnrreyna I am approving this PR.

src/utils/viewer/viewer.cpp
mosra: Is the `viewer.py` meant to be packaged / installed in some way (`setup.py`, pip etc.)? I know very little about Python packaging, but if it is, I'd put the font file into some "well-known package data location" (if there's such a thing) and reference it from there. If not, then I'm afraid the path has to be hardcoded that way. But I'd make it relative to `__file__`, at least, so it doesn't depend on CWD.

Another idea I had was that we embed the font directly inside the Habitat libraries instead of the C++ viewer, and expose that array to Python. Such as 

```cpp
Containers::ArrayView<const char> fontData() {
    Cr::Utility::Resource rs{"fonts"};
    return rs.getRaw("ProggyClean.ttf");
}
```

and then, if the above would be exposed as `habitat.font_data()` to Python, you'd do 

```py
self.display_font.open_data(habitat.font_data(), size)
```

instead of `self.display_font.open_file("that/ugly/path.ttf", size)`. A downside is that the binary now embeds a font file that's only needed if using a GUI viewer, but as long as we stay with that tiny 40 kB font file, I think the convenience wins over memory use :)

examples/viewer.py
```diff
@@ -73,6 +94,59 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
             key.Z: "move_up",
         }
 
+        # Load a TrueTypeFont plugin and open the font file
+        self.display_font = text.FontManager().load_and_instantiate("TrueTypeFont")
+        relative_path_to_font = "../data/fonts/ProggyClean.ttf"
+        self.display_font.open_file(
+            os.path.join(os.path.dirname(__file__), relative_path_to_font),
+            180,
```
mosra, line 102:
```suggestion
            13,
```

examples/viewer.py
```diff
@@ -73,6 +94,59 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
...
+        )
+
+        # Glyphs we need to render everything
+        self.glyph_cache = text.GlyphCache(mn.Vector2i(2048))
```
mosra, line 106:
```suggestion
        self.glyph_cache = text.GlyphCache(mn.Vector2i(256))
```

... we discussed this on Slack some weeks ago, so just a reminder :) It'll also avoid unnecessary GPU memory use, and since the font is a bitmap one, rendering it at 10x the size will not make it any prettier.

examples/viewer.py
```diff
@@ -73,6 +94,59 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
...
+            mn.Vector2(self.viewport_size)
+        ) @ mn.Matrix3.translation(
+            mn.Vector2(
+                self.viewport_size[0]
+                * -HabitatSimInteractiveViewer.TEXT_DELTA_FROM_CENTER,
+                self.viewport_size[1]
+                * HabitatSimInteractiveViewer.TEXT_DELTA_FROM_CENTER,
+            )
+        )
+        self.shader = shaders.DistanceFieldVectorGL2D()
```
mosra, line 135:
```suggestion
        self.shader = shaders.VectorGL2D()
```

Since you aren't really using any outline anyway, let's use a simpler shader. (The same is used in the `viewer.cpp`.)

examples/viewer.py
```diff
@@ -733,6 +810,29 @@ def exit_event(self, event: Application.ExitEvent):
         event.accepted = True
         exit(0)
 
+    def draw_text(self, sensor_spec):
+        self.shader.bind_vector_texture(self.glyph_cache.texture)
+        self.shader.transformation_projection_matrix = self.window_text_transform
+        self.shader.color = [1.0, 1.0, 1.0]
+        self.shader.outline_color = [1.0, 1.0, 1.0]
+        self.shader.outline_range = (0.45, 0.40)
```
mosra, line 818:
```suggestion
```

Those aren't defined in the simpler non-distance-field shader.

data/fonts/README.txt
```diff
@@ -0,0 +1,364 @@
+----------------------------------------------------------------------
```
mosra, line 1: I'd remove everything from this directory except for the `ProggyClean.ttf`. Unnecessary and irrelevant now that the font file is hardcoded anyway. Tho it might be good to credit 

> `ProggyClean.ttf` by Tristan Grimmer

somewhere, although I don't know if there's even such a place in the docs or README.

Skylion007:
Do we need the other font files besides proggy?

jrreyna:
> Do we need the other font files besides proggy?

I'm not sure, I just used the same file that was in the viewer.cpp file. Those fonts were in a directory in the imgui submodule, so I don't even think those font files are on main, right?

johnrreyna:
> Do we need the other font files besides proggy?

Just talked to mosra, probably not. Will delete them for now

Skylion007:
@johnrreyna Please squash your commits next time when merging.

PR #1866, opened 2022-09-16
Add islands to NavMesh API

aclegg3:
## Motivation and Context

Though the underlying C++ nav module has been computing islands on the back-end, we hadn't yet exposed island filtered queries for NavMesh operations.

This PR introduces new island operations:
- `get_island(point)` - get the island index closest to a point
- `island_area(index)` - get the area of a specific island by index
- `island_radius(point)` and `island_radius(index)` - get the radius heuristic for island size
- `num_islands` - property for number of islands on a NavMesh

and adds optional island specification to other operations:
- `get_random_navigable_point()` and `get_random_navigable_point_near()` - now can query a specific island by index
- `snap_point()` - now can snap to a specified island
- `build_navmesh_vertices()` - can get vertex buffer for a specific island
- `build_navmesh_vertex_indices()` - can get index buffer for a specific island

Also adds and updates docstrings for the full Nav module in both C++ and Python.

## How Has This Been Tested

Added new pytests for island functionality.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -108,6 +108,8 @@ std::tuple<dtStatus, dtPolyRef, vec3f> projectToPoly(
 }  // namespace
 
 namespace impl {
+// some systems lack this typedef (e.g. emscripten build)
+typedef unsigned short int ushort;
 
 //! (flag & flag) operator wrapper for function pointers
 inline ushort andFlag(ushort curFlags, ushort flag) {
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -924,6 +924,7 @@ float polyArea(const dtPoly* poly, const dtMeshTile* tile) {
 // Also compute the NavMesh areas for later query.
 void impl::IslandSystem::removeZeroAreaPolys(dtNavMesh* navMesh) {
   islandsToArea_ = std::unordered_map<uint32_t, float>();
+  islandsToArea_.reserve(islandsToPolys_.size());
   // initialize the area cache.
   for (auto& itr : islandsToPolys_) {
     islandsToArea_[itr.first] = 0.0;
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -241,6 +241,7 @@ class IslandSystem {
 
     if (islandIndex == ID_UNDEFINED) {
       // all islands
+      islands.reserve(islandsToPolys_.size());
       for (auto& itr : islandsToPolys_) {
         islands.push_back(itr.first);
       }
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -145,9 +145,6 @@ class IslandSystem {
         if (navMesh->isValidPolyRef(startRef) &&
             (polyToIsland_.find(startRef) == polyToIsland_.end())) {
           uint32_t newIslandId = islandRadius_.size();
-          islandsToPolys_.emplace(std::piecewise_construct,
-                                  std::forward_as_tuple(newIslandId),
-                                  std::tuple<>());
           expandFrom(navMesh, filter, newIslandId, startRef, islandVerts);
 
           // The radius is calculated as the max deviation from the mean for all
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -246,7 +246,8 @@ class IslandSystem {
         islands.push_back(itr.first);
       }
     } else if (invert) {
-      // all but a  single island
+      // all but a single island
+      islands.reserve(islandsToPolys_.size());
       for (auto& itr : islandsToPolys_) {
         if (itr.first != islandIndex) {
           islands.push_back(itr.first);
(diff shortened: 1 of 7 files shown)


src/esp/nav/PathFinder.cpp
```diff
@@ -169,15 +181,114 @@ class IslandSystem {
...
+                                   ushort flag,
+                                   int islandIndex = ID_UNDEFINED,
+                                   bool setFlag = true,
+                                   bool invert = false) {
+    assertValidIsland(islandIndex);
+    CORRADE_ASSERT(navMesh != nullptr, "invalid navMesh pointer", );
+    std::vector<int> islands;
+
+    if (islandIndex == ID_UNDEFINED) {
+      // all islands
```
Skylion007, line 243:
```suggestion
      // all islands
      islands.reserve(islandsToPolys_.size());
```

src/esp/nav/PathFinder.cpp
```diff
@@ -169,15 +181,114 @@ class IslandSystem {
...
+      for (auto& itr : islandsToPolys_) {
+        islands.push_back(itr.first);
+      }
+    } else if (invert) {
+      // all but a  single island
```
Skylion007, line 248:
```suggestion
      // all but a single island
      islands.reserve(islandsToPolys_.size());
```

src/esp/nav/PathFinder.cpp
```diff
@@ -134,6 +145,7 @@ class IslandSystem {
         if (navMesh->isValidPolyRef(startRef) &&
             (polyToIsland_.find(startRef) == polyToIsland_.end())) {
           uint32_t newIslandId = islandRadius_.size();
+          islandsToPolys_[newIslandId] = std::vector<dtPolyRef>();
```
Skylion007, line 148: This is inefficient. Since we don't support C++17, unfortunately, the proper way to do this is a bit ugly:
```suggestion
          islandsToPolys.emplace(std::piecewise_construct,
                                                 std::forward_as_tuple(newIslandId),
                                                 std::tuple<>());
```

src/esp/nav/PathFinder.cpp
```diff
@@ -793,35 +919,50 @@ float polyArea(const dtPoly* poly, const dtMeshTile* tile) {
 // Some polygons have zero area for some reason.  When we navigate into a zero
 // area polygon, things crash.  So we find all zero area polygons and mark
 // them as disabled/not navigable.
-// Also compute the total NavMesh area for later query.
-void PathFinder::Impl::removeZeroAreaPolys() {
-  navMeshArea_ = 0;
+// Also compute the NavMesh areas for later query.
+void impl::IslandSystem::removeZeroAreaPolys(dtNavMesh* navMesh) {
+  islandsToArea_ = std::unordered_map<uint32_t, float>();
```
Skylion007, line 924:
```suggestion
  islandsToArea_ = std::unordered_map<uint32_t, float>();
  islandsToArea_.reserve(islandsToPolys_.size());
```

src/esp/nav/PathFinder.cpp
```diff
@@ -1371,11 +1599,13 @@ PathFinder::Impl::getTopDownView(const float metersPerPixel,
...
 
-assets::MeshData::ptr PathFinder::Impl::getNavMeshData() {
-  if (meshData_ == nullptr && isLoaded()) {
-    meshData_ = assets::MeshData::create();
-    std::vector<esp::vec3f>& vbo = meshData_->vbo;
-    std::vector<uint32_t>& ibo = meshData_->ibo;
+assets::MeshData::ptr PathFinder::Impl::getNavMeshData(
+    int islandIndex /*= ID_UNDEFINED*/) {
+  islandSystem_->assertValidIsland(islandIndex);
+  if (islandMeshData_[islandIndex] == nullptr && isLoaded()) {
```
Skylion007, line 1605: This will actually initialize with the default constructrible value if it doesn't exist (which I believe is nullptr?) You may want to do a proper key check here then and do another emplace.

As an added bonus, a reference to the key:value std::pair created by emplace will be returned so you can just return that instead of looking up the key in the unordered_map again for the return value.

Skylion007 (changes requested):
I added some code suggestions to address some performance concerns with std::unordered_map (which is admittingly complex to use properly pre-C++17 with try_emplace).

Mainly:
1. Calling reserve() to prevent tons of reallocations / rehashing of the hashmap as the container grows
2. Emplacing the std::vector directly into the map during construction
3. Emplacing the value directly if it makes sense (non-primitive type).

src/esp/nav/PathFinder.cpp
```diff
@@ -134,6 +145,9 @@ class IslandSystem {
         if (navMesh->isValidPolyRef(startRef) &&
             (polyToIsland_.find(startRef) == polyToIsland_.end())) {
           uint32_t newIslandId = islandRadius_.size();
+          islandsToPolys_.emplace(std::piecewise_construct,
+                                  std::forward_as_tuple(newIslandId),
+                                  std::tuple<>());
```
Skylion007, line 150:
```suggestion
```
Looking closer at the code, I don't think we actually need this part at all. the '[]' operator will default initialize a value if it doesn't exist and luckily the std::vector is default constructible. So as long you don't access it with a lower level call like .at(), this is unnecessary.

Skylion007 (approved):

PR #1868, opened 2022-09-19
[WIP] Handle zero scaling mesh transforms

aclegg3:
## Motivation and Context

Some mesh transforms contain zeros in the scaling component of the transform. Likely this is erroneous, but currently this will crash the simulator during import. This patch isolates the issue and adds a check with an error message and corrective measure applied automatically to prevent crashing.

We should think about the correct handling here.

## How Has This Been Tested

Locally with a crashing asset. 

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1783,10 +1783,21 @@ bool ResourceManager::loadRenderAssetGeneral(const AssetInfo& info) {
   // Set transformations. Objects that are not part of the hierarchy are
   // ignored, nodes that have no transformation entry retain an identity
   // transformation.
-  for (const Cr::Containers::Pair<unsigned, Mn::Matrix4>& transformation :
+  for (Cr::Containers::Pair<unsigned, Mn::Matrix4>& transformation :
        scene->transformations3DAsArray()) {
     if (Cr::Containers::Optional<esp::assets::MeshTransformNode>& node =
             nodes[transformation.first()]) {
+      if (transformation.second().scaling().product() == 0) {
+        ESP_ERROR() << "0 value in scaling vector for mesh transform. "
+                       "Indicates an issue with source asset. Overriding to "
+                       "small nonzero and continuing.";
+        ESP_ERROR() << transformation.second();
+        for (auto ix = 0; ix < 3; ++ix) {
+          if (transformation.second()[ix][ix] == 0) {
+            transformation.second()[ix][ix] = 1e-7;
+          }
+        }
+      }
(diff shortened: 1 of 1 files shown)


src/esp/assets/ResourceManager.cpp
```diff
@@ -1783,10 +1783,21 @@ bool ResourceManager::loadRenderAssetGeneral(const AssetInfo& info) {
...
        scene->transformations3DAsArray()) {
     if (Cr::Containers::Optional<esp::assets::MeshTransformNode>& node =
             nodes[transformation.first()]) {
+      if (transformation.second().scaling().product() == 0) {
+        ESP_ERROR() << "0 value in scaling vector for mesh transform. "
+                       "Indicates an issue with source asset. Overriding to "
+                       "small nonzero and continuing.";
+        ESP_ERROR() << transformation.second();
+        for (auto ix = 0; ix < 3; ++ix) {
+          if (transformation.second()[ix][ix] == 0) {
```
mosra, line 1796:
```suggestion
          if (transformation.second()[ix].xyz().isZero()) {
```

could properly handle also cases with rotations? i.e., a matrix that's rotating by 90° on Y has zeros on diagonal, but is still valid:

```py
>>> from magnum import *
>>> Matrix4.rotation_y(Deg(90))
Matrix(0, 0, 1, 0,
       0, 1, 0, 0,
       -1, 0, 0, 0,
       0, 0, 0, 1)

```

src/esp/assets/ResourceManager.cpp
```diff
@@ -1783,10 +1783,21 @@ bool ResourceManager::loadRenderAssetGeneral(const AssetInfo& info) {
...
+            transformation.second()[ix][ix] = 1e-7;
```
mosra, line 1797: My linear algebra is too rusty to know a better solution than what you have here. Ideally it would add the value to where it should be given a particular rotation, but I have no idea what is the math leading to that.

An alternative option would be checking if the decomposed TRS values are present, via [translationRotationScaling3DFor()](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1SceneData.html#ab401e58557037318e44a8bf1b94f33af), and if they are (which is the case with some but not all glTF files, as we talked about earlier on Slack), then looking just at zero scale components, adjusting them, and composing a matrix out of it again. That way I think it would be robust enough to not even need to print an error message, the error would be printed only if there is just the matrix.

mosra (commented):
Still on the fence on whether zero scale should be something we accept as a valid scenario (similarly as we accept negative scale) or something we fail hard on.

PR #1870, opened 2022-09-22
(bugfix): Prevent unnecessary copies with more std::move and emplace

Skylion007:
## Motivation and Context

* When I was reviewing @jturner65's PR, I noticed a series of strings that were being copied into their std::set. This explicitly moves them when inserting them into the set instead which should prevent an unnecessary copy.

## How Has This Been Tested
Locally

## Types of changes

## Checklist


jturner65 (approved):
Looks good to me.  Thanks!

PR #1876, opened 2022-09-29
Properly handle non-indexed collision meshes

mosra:
## Motivation and Context

To avoid this silly assert:

```
Trade::MeshData::indexType(): the mesh is not indexed
```

Can happen quite a lot with STL files, which are also quite likely to be referenced from URDFs. See also https://cvmlp.slack.com/archives/CFN5TAUSD/p1663172543970699 for another report of the same problem -- and apologies in case it's already being fixed elsewhere, I couldn't find any related PR.

Also contains an unrelated change cleaning up a no-longer-used define (I suppose that was for the long-gone `io` functionality that was replaced with Corrade's `Utility::Path`).

## How Has This Been Tested

Hit this myself when loading a random old version of the Hab Fetch robot. After the fix it no longer blows up.


Skylion007 (approved):

aclegg3 (approved):
Thanks, LGTM.

PR #1878, opened 2022-09-29
Update Magnum submodules with support for querying MeshData size from Python

mosra:
## Motivation and Context

I was actually false-advertising on the call that it's possible to query [`trade.MeshData.index_data`](https://doc.magnum.graphics/python/magnum/trade/MeshData/#index_data) and [`trade.MeshData.vertex_data`](https://doc.magnum.graphics/python/magnum/trade/MeshData/#vertex_data) size from Python. So I added those.

For images, [`trade.ImageData2D.data`](https://doc.magnum.graphics/python/magnum/trade/ImageData2D/#data) was a thing before already. Those are byte array views, so `len()` is the byte count.

Also contains all other submodule updates that are a part of #1874 as well. Nothing controversial or attention-worthy there, since all those changes are mostly just related to glTF export that only the batch renderer uses.

## How Has This Been Tested

My CIs pass.


Skylion007:
Nit: on the magnum-bindings. You cast from handle to C++ back to handle with `py::cast(self)`. You could just have self start as py::handle cast to C++ and save a conversion.

Skylion007 (approved):

mosra:
Hm, but there I need to access `self.indexData()` for example. So I would need to (manually) cast to C++ anyway, no? Not sure what you mean exactly.

aclegg3 (approved):

Skylion007:
> Hm, but there I need to access self.indexData() for example. So I would need to (manually) cast to C++ anyway, no? Not sure what you mean exactly.

Ah, I guess in this case it doesn't make quite as much sense. It's very useful for when you are using only python APIs. Also, it would save you a python->C++->python roundtrip, but probably not worth the added complexity here.

mosra:
Yeah, here it's not that much of a perf bottleneck, so I prefer to have it done the least error-prone way possible.

Anyway, thanks for the hints, as always. Much appreciated :)

PR #1882, opened 2022-09-30
Dataset object processor

jrreyna:
## Motivation and Context

See issue #1886. We need to be able to assess millions of assets to determine any that can be problematic when simulating, e.g. use too much memory, the meshes are too complex or messy, or they behave non-realistically during physics simulations.

## How Has This Been Tested

Locally

## Types of changes

## Checklist


jrreyna:
You see the commits from a previous PR because I accidentally created this branch off another branch instead of origin/main. But I pulled and merged from main so that the files changed in that old branch were overridden

tools/dataset_object_processor.py
```diff
@@ -51,7 +52,62 @@ def print_in_color(print_string="", color=PrintColors.WHITE) -> None:
...
+    manager = trade.ImporterManager()
+    importer = manager.load_and_instantiate("AnySceneImporter")
+    importer.open_file(asset_path)
+    print_in_color(f"Plugin dir: {manager.plugin_directory}", PrintColors.PURPLE)
+    print_in_color("-" * 72, PrintColors.PURPLE)
+
+    # Get mesh data
+    print_in_color("\nMesh Data", PrintColors.GREEN)
+    print_in_color("-" * 72, PrintColors.GREEN)
+    print_in_color(f"Scene importer class: {type(importer)}", PrintColors.GREEN)
```
mosra, line 70: Just FYI: I'm not exposing the particular importer plugin subclasses (like `GltfImporter`, `AssimpImporter`...) to Python (and I never plan to), so `type(importer)` would show the same (`magnum.trade.AbstractImporter`) for all plugins. The way to query which plugin class/module is actually used would be through [plugin()](https://doc.magnum.graphics/corrade/classCorrade_1_1PluginManager_1_1AbstractPlugin.html#a15b2a4e5e20275e08d4e5ad34b387e8b) (which isn't exposed to Python at the moment). In case of the `Any*Importer` it's further complicated because it delegates to a concrete plugin implementation internally. The [ImporterFlag::Verbose](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Trade.html#a137d832c7ce6e6b654022321941d70f3ad4a9fa383ab700c5bdd6f31cf7df0faf) (which... isn't exposed to Python at the moment either) makes it print the chosen plugin name to the output. TL;DR: there isn't much point in showing this particular piece of information.

To get a better idea of how the plugins get picked and behave I think it might be helpful for you to get the prebuilt tools binaries I linked on Slack and play with `magnum-sceneconverter --info <some-file.gltf>`. The `-I` option allows you to choose a particular importer plugin (`AnySceneImporter` is the default, which delegates to `GltfImporter` for glTF files, or you can specify `GltfImporter` directly, you can also pick `AssimpImporter` because Assimp sort-of knows glTF as well), `--verbose` will print additional information like what plugins got used to open a particular image etc. The `--importer-info` will then print info about a particular plugin specified with `-I`. For a full picture, try with `*.dae`, `*.ply` and `*.stl` files as well if you find them.

If you don't want to spend time reinventing things, you might possibly also want to steal from / get inspired from how the `magnum-sceneconverter --info` output looks like, what it contains and how is it formatted. Not everything is exposed to Python at the moment, but the essential stuff should be -- and I'm always happy to extend the bindings where needed.

*(Apologies if this is overwhelming, I'm just trying to share the info. If you feel it's too much and you don't need that many details, let me know ;))*

jrreyna: No worries that was all good information. Yeah once I saw that the output is always `magnum.trade.AbstractImporter` I realized more how the Python side works. I assumed it was doing some delegating behind the scenes to determine the correct C++ importer class to use.

mosra: All that delegating magic is inside the `Any*` plugins themselves -- if you use them instead of a concrete [plugin name or alias](https://doc.magnum.graphics/magnum/file-formats.html), they will attempt to detect based on file extension and delegate to *some* implementation for given format: https://github.com/mosra/magnum/blob/master/src/MagnumPlugins/AnySceneImporter/AnySceneImporter.cpp#L85-L144

In case of images, there's also detection based on file signatures, similarly to what the Unix `file` command does: https://github.com/mosra/magnum/blob/17638a2ad50094ece8eeb7ad7f9b06033d05d843/src/MagnumPlugins/AnyImageImporter/AnyImageImporter.cpp#L202-L270

tools/dataset_object_processor.py
```diff
@@ -21,6 +22,21 @@
 ROBOT_PATH = ""  # TODO, which dataset is this? Robot fetch?
 
 
+class MemoryUnitConverter:
+    """
+    class to convert computer memory value units, e.g.
+    1,000,000 bytes to 1 megabyte
```
mosra, line 28: 1048576? :eyes:

jrreyna: Embarrassingly, I just learned that it is not 1000 bytes in a KB, but actually 1024.... apparently some people like to use the term "kibibyte" to refer to 1024 bytes bc KB is so commonly associated with 1000. I now know why my 500 GB external hard drive only has 466 GB :(

mosra: I don't understand why the units for storage devices are messed up like that, but everywhere else it was always powers of 1024 :)

tools/dataset_object_processor.py
mosra: Yes, although there's no practical difference compared to using it through the AnySceneImporter.

jrreyna: Is there any case where there WOULD be a practical difference? I also like my code, style-wise, to be very readable. Sometimes, if the filetype is very well known, I like to be explicit of what kind of variable/class an object is

tools/dataset_object_processor.py
jrreyna: Yeah I was having a conversation with my tech-savvy friend about it, and his theory is that most people would be confused by computer memory being measured in powers of 2 to be very confusing, so they just round it to the nearest thousand, million, etc to avoid explaining why a GB is seemingly arbitrarily decided to be 1,048,576 bytes 😆

tools/dataset_object_processor.py
```diff
@@ -0,0 +1,339 @@
...
+    rigid_object_manager = sim.get_rigid_object_manager()
+    print_in_color(rigid_object_manager.get_objects_CSV_info(), PrintColors.CYAN)
+
+    # Get object attribute manager for objects from dataset, load the dataset,
+    # store all the object template handles in a List, then process each asset
+    object_attributes_manager = sim.get_object_template_manager()
+    object_attributes_manager.load_configs(dataset_path)
+    object_template_handles = object_attributes_manager.get_file_template_handles("")
+    print_in_color(
+        f"\nnumber of ojects in dataset: {len(object_template_handles)}",
```
Skylion007, line 205:
```suggestion
        f"\nnumber of objects in dataset: {len(object_template_handles)}",

File: tools/dataset_object_processor.py (hunk 1 above)
Comment (mosra) [reply tools/dataset_object_processor.py]: Okay, so the differences are:

- (the obvious one) if you have a *very minimal* build/installation that has only `GltfImporter` but no `AnySceneImporter` (or equivalently for other file types), you have to use `GltfImporter` directly
- the `AnySceneImporter` has some overhead compared to using `GltfImporter` directly because of the file format detection and delegation to another plugin (but nothing that would be in any way significant, so it's not a useful optimization strategy)
- if you would want to read the plugin-specific configuration for whatever reason, `GltfImporter` gives you back [all its options](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1GltfImporter.html#Trade-GltfImporter-configuration), but `AnySceneImporter` will give you back an empty configuration, because on its own it doesn't have any option. Unlikely to affect practical use in code, but matters when you'd want to check the config options in [`magnum-sceneconverter --info-importer`](https://doc.magnum.graphics/magnum/magnum-sceneconverter.html#magnum-sceneconverter-example), for example.
- using `GltfImporter` instead of `AnySceneImporter` makes it clearer what you're trying to do, yes, you're right about that .. but OTOH it can be seen as hardcoding a specific file format, so when someone would want to use the same code for loading COLLADA or OBJ/PLY/STL... files in the future, they would have to change the code first

File: tools/dataset_object_processor.py (hunk 2 above)
Comment (mosra) [reply tools/dataset_object_processor.py]: Ah well, to solve this I'd need to expose all the plugin configuration APIs to Python. Which is a rather big chunk of work, and I kinda hoped that it wouldn't be needed so soon :sweat_smile: I got an idea for a potential workaround, but that didn't work in the end, so you'll have to suffer the warnings for a while until I get to exposing the configuration APIs. It'll happen eventually as a lot other dataset processing tasks need them as well, but not immediately. Sorry.

To explain: some datasets contain images encoded with Basis Universal, which is an image format that's able to transcode itself to GPU-specific block-compression texture formats such as [BCn](https://learn.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression#bc1) or [ASTC](https://en.wikipedia.org/wiki/Adaptive_scalable_texture_compression). Those are usually about encoding 4x4 blocks of pixels into 64 or 128 bits, thus eight or four times less memory than RGBA with 8 bits per channel. But unfortunately there's no single format all GPUs would support, so one has to explicitly say which format to transcode to, and if it's not done, it decompresses back to RGBA8 that's guaranteed to be supported everywhere. Which however also means that for the Basis images you get four (or eight) times as much data, skewing the stats.

> The script takes a while because it has to print it out for each dataset object

It's not due to the warning, but rather due to the decoding process, which takes time. Regular PNG images also take some time to decode, but Basis images usually contain mip levels so there's much more of them, taking more time, and the RGBA8 decoding code path is probably not well optimized by the library.

Comment (jrreyna) [reply tools/dataset_object_processor.py]: Ah I see, no problem then, take your time

File: tools/dataset_object_processor.py [hunk 4]
```diff
@@ -0,0 +1,374 @@
...
+    BYTES = 0
+    KILOBYTES = 1
+    MEGABYTES = 2
+    GIGABYTES = 3
+
+    UNIT_STRS = ["bytes", "KB", "MB", "GB"]
+    UNIT_CONVERSIONS = [1, 1 << 10, 1 << 20, 1 << 30]
+
+
+class PrintColors:
```
Skylion007, line 103:
```suggestion
from enum import Enum
class PrintColors(Enum):
```

tools/dataset_object_processor.py
```diff
@@ -0,0 +1,374 @@
...
+    ENDC = "\033[0m"
+    BOLD = "\033[1m"
+    UNDERLINE = "\033[4m"
+
+
+def print_in_color(print_string="", color=PrintColors.WHITE) -> None:
+    """
+    Allows us to print to console in different colors
+    """
+    print(color + print_string + PrintColors.ENDC)
```
Skylion007, line 131: Why aren't we just using Colorama for this?

Skylion007: Tons of other ones like Colored etc: https://pypi.org/project/colored/

PR #1646, opened 2022-02-02
Sanchitgarg/audio sensor

sanchitgarg:
## Motivation and Context

With this change, we give our awesome virtual agents the ability to hear. This enables new research avenues like Audio-Visual navigation. 

We are adding a new git-submodule : `rlr-audio-propagation`. 
This git repo hosts the header and binaries for Meta RL Research's audio propagation engine. A new AudioSensor C++ class is added that calls into this new lib. The C++ AudioSensor class is exposed to python via the existing pybindings.

Further, we are adding a new build flag : `--audio`. 
This is an optional flag. Any user who wants to use the audio sensor will have to build with this flag.

Please see the **examples/tutorials/audio_agent.py** script for a tutorial on how to use it. 
Also, please refer to **docs/AUDIO.md** which goes into more details about the feature.

## How Has This Been Tested

The code was tested locally using the C++ viewer. A new audio_agent.py script has been added as well to test the change. Further, we tested the integration with the SoundSpaces team. They were able to generate impulse response in they navigation system.

## Types of changes

- [x] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1892, opened 2022-10-06
Temporarily work around lost code coverage reports

mosra:
## Motivation and Context

Builds on `main` are getting a lot of these errors due to the coverage being collected after the Audio build (originating from https://github.com/facebookresearch/habitat-sim/pull/1646), losing the coverage collected earlier:

> libgcov profiling error:/home/circleci/project/habitat-sim/build/esp/bindings/CMakeFiles/habitat_sim_bindings.dir/Bindings.cpp.gcda:overwriting an existing profile data with a different timestamp

With this change the coverage is back at 60% (from 40%), and still includes also everything from the Audio build.

The better long-term solution, however, would be like something in https://github.com/facebookresearch/habitat-sim/pull/1860 -- this unnecessarily duplicates a lot of the CI setup code. (@Skylion007 or anybody else who knows YAML or CircleCI, do you know if there is a possibility to reuse / deduplicate the markup somehow?)

## How Has This Been Tested

Comparing these two reports gives a ~20% increase in code coverage:

- https://app.codecov.io/gh/facebookresearch/habitat-sim/tree/main/src/esp
- https://app.codecov.io/gh/facebookresearch/habitat-sim/tree/codecov-gimme-back-those-20-percent/src/esp

As far as I checked, the Audio build coverage (such as `src/esp/sensor/AudioSensor.cpp`) seems to stay the same, so the second coverage upload gets merged with the first one. That's only from a cursory look though, if you see anything suspicious in the new reports, let me know.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -583,24 +583,7 @@ jobs:
               . ~/emsdk/emsdk_env.sh
               # Generate JS CodeConv
               npm run test_with_coverage
-      -  run:
-          name: Build, install habitat-sim with audio and run audio_agent script
-          no_output_timeout: 10m
-          command: |
-              export PATH=$HOME/miniconda/bin:$PATH
-              . activate habitat;
-              cd habitat-sim
-              pip install -r requirements.txt --progress-bar off
-              pip install imageio imageio-ffmpeg
-              git submodule update --init --recursive --jobs 8
-              python -u setup.py install --build-type "Release" --lto --headless --audio
-              python examples/tutorials/audio_agent.py
-      - save_cache:
-          key: ccache-{{ arch }}-{{ .Branch }}-{{ .BuildNum }}
-          background: true
-          paths:
-            - /home/circleci/.ccache
-      - run:
+      - run: &upload_coverage
           name: Upload test coverage
           command: |
               export PATH=$HOME/miniconda/bin:/usr/local/cuda/bin:$PATH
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -580,6 +580,29 @@ jobs:
...
+
+              # Uploading test coverage for JS code
+              ./codecov coverage_js/coverage-final.json -cF JavaScript
+
+              #Uploading test coverage for C++ code
+              lcov --directory . --capture --output-file coverage.info
+              # Replaces -1 linecount with zero to prevent lcov from crashing:
+              # https://github.com/psycofdj/coverxygen/issues/6
+              sed -i -e 's/,-1$/,0/g' coverage.info
+              #lcov --remove coverage.info "*/deps/*" --output-file coverage.info > /dev/null
```
Skylion007, line 602: Quick question, we hy are these commented out? Don't we still not want dependencies to count towards coverage? At the very lest we should add these globs to .coveragerc then.

Skylion007 (approved):
Other than that, looks good.

Skylion007:
> (@Skylion007 or anybody else who knows YAML or CircleCI, do you know if there is a possibility to reuse / deduplicate the markup somehow?) You absolute can deduplicate it by either using yaml anchors or writing a custom. For the anchor see how install_cuda is deduplicated: https://github.com/facebookresearch/habitat-sim/blob/82fdc0bc110e93b3e350f24b055e0f26a9777707/.circleci/config.yml#L111

.circleci/config.yml
mosra: I just copied what was there from the other part :) I don't want to touch anything here, only want to fix what's broken.

Will use the yaml anchor, thanks.

Skylion007 (approved):

PR #1812, opened 2022-07-15
Update theme.css

mrestrepo1:
Revised CSS to match Meta brand.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1893, opened 2022-10-06
Better colors for "info" and "success" notes in the documentation and website

mosra:
## Motivation and Context

Sorry for opening too many PRs lately :D I was in the process of writing high-level docs for the batch renderer in #1874, submitted #1889 for doc updates already, but then wanted to put some note blocks into the doc copy, realized they look kinda awful, and had to fix those too. Before:

After:

The "warning" note was good enough already, but apparently those two were never used so far so they still had the dark theme colors. I didn't spend time updating the remaining color styles for "primary" and "danger" as I don't need them at the moment, and this was enough of a detour already :sweat_smile: Same for #1812, originally I wanted to merge those changes with mine, but the font and spacing changes felt like a lot more time investment would be needed to get things look consistent again, so I didn't either. The two PRs don't conflict with each other, at least.

Also updated the m.css submodule to latest, because why not, and added a note into the documentation README about updating CSS colors. Reverting a change from #1889 that added a few new Doxygen aliases, as those were present in `Doxyfile-mcss` already; added a new alias for relative references that look shorter in the output, and added a predefined `DOXYGEN_GENERATING_OUTPUT` macro for hiding internal and nasty-looking code from Doxygen.

Once merged, I'll transfer the CSS changes to the lab and website repos.

## How Has This Been Tested

See screenshots above.

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -261,11 +261,7 @@ TAB_SIZE               = 4
 # commands \{ and \} for these it is advised to use the version @{ and @} or use
 # a double escape (\\{ and \\})
 
-ALIASES                = \
-    "cb{1}=@code{\1}" \
-    "cpp=@code{.cpp}" \
-    "cmake=@code{.cmake}" \
-    "ce=@endcode"
+ALIASES                =
 
 # Set the OPTIMIZE_OUTPUT_FOR_C tag to YES if your project consists of C sources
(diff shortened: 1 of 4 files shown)


0mdc (approved):
Looks good!

Skylion007 (approved):

PR #1903, opened 2022-10-19
[Nav] - Islands topdown map

aclegg3:
## Motivation and Context
It can be very useful to get a topdown map with island annotations and now after #1866 this is pretty easy. 
In addition to binary maps of navigability (`PathFinder.get_topdown_view`), the API now provides `PathFinder.get_topdown_island_view` which returns an integer array containing island indices or `-1` for non-navigable cells.

## How Has This Been Tested

New pytests.

See the examples below for a scene with multiple islands due to low doorways:
image

6AGcGQf2wof_topdown_islands npy

## Types of changes

## Checklist


src/esp/nav/PathFinder.cpp
```diff
@@ -1625,7 +1628,44 @@ PathFinder::Impl::getTopDownView(const float metersPerPixel,
...
+}
+
+typedef Eigen::Matrix<int, Eigen::Dynamic, Eigen::Dynamic> MatrixXi;
+
+MatrixXi PathFinder::Impl::getTopDownIslandView(const float metersPerPixel,
+                                                const float height,
+                                                const float eps) const {
+  std::pair<vec3f, vec3f> mapBounds = bounds();
+  vec3f bound1 = mapBounds.first;
+  vec3f bound2 = mapBounds.second;
```
Skylion007, line 1648:
```suggestion
  vec3f bound1 = std::move(mapBounds.first);
  vec3f bound2 = std::move(mapBounds.second);
```

Not sure if they are trivially copyable (so double check with clang-tidy), but I seem to remember them not being non-trivially movable.

src/esp/nav/PathFinder.cpp
```diff
@@ -1625,7 +1628,44 @@ PathFinder::Impl::getTopDownView(const float metersPerPixel,
...
+
+  float xspan = std::abs(bound1[0] - bound2[0]);
+  float zspan = std::abs(bound1[2] - bound2[2]);
+  int xResolution = xspan / metersPerPixel;
+  int zResolution = zspan / metersPerPixel;
```
Skylion007, line 1653: Do you want to truncate instead of rounding? 
```suggestion
  int xResolution = (xspan + 0.5) / metersPerPixel;
  int zResolution = (zspan + 0.5) / metersPerPixel;
```

src_python/habitat_sim/utils/viz_utils.py
```diff
@@ -287,3 +287,27 @@ def semantic_to_rgb(semantic_image: np.ndarray) -> np.ndarray:
...
+
+    :param island_top_down_map_data: The island index map data from Pathfinder.get_topdown_island_view()
+
+    :return: rgb Image of islands at the desired slice.
+    """
+
+    island_map = Image.new("RGB", island_top_down_map_data.shape)
+    pixels = island_map.load()
+    white = int("0xffffff", base=16)
+    for x in range(island_top_down_map_data.shape[0]):
```
Skylion007, line 304: setting all the pixels to white is easily vectorizable and maybe faster to do before the loop?

jturner65: **Image.new** has a 3 argument overload where it takes (mode, size, color), where... 

> The colour is given as a single value for single-band images, and a tuple for multi-band images (with one value for each band). If the colour argument is omitted, the image is filled with black. If the colour is None, the image is not initialised.

That'd probably be better, no?

Skylion007: Easier AND faster.

src/esp/nav/PathFinder.h
```diff
@@ -606,14 +606,31 @@ class PathFinder {
...
    */
-  Eigen::Matrix<bool, Eigen::Dynamic, Eigen::Dynamic> getTopDownView(
-      float metersPerPixel,
-      float height);
+  Eigen::Matrix<bool, Eigen::Dynamic, Eigen::Dynamic>
+  getTopDownView(float metersPerPixel, float height, float eps = 0.5);
+
+  /**
+   * @brief Get a 2D grid marking island index for navigable cells and -1 for
+   * non-navigable cells at a specified hieght and resolution.
```
0mdc, line 620: `hieght`

src/esp/bindings/ShortestPathBindings.cpp
```diff
@@ -141,9 +141,14 @@ void initShortestPathBindings(py::module& m) {
...
-      .def("get_topdown_view", &PathFinder::getTopDownView,
-           R"(Returns the topdown view of the PathFinder's navmesh.)",
-           "meters_per_pixel"_a, "height"_a)
+      .def(
+          "get_topdown_view", &PathFinder::getTopDownView,
+          R"(Returns the topdown view of the PathFinder's navmesh at a given vertical slice with eps slack.)",
+          "meters_per_pixel"_a, "height"_a, "eps"_a = 0.5)
+      .def(
+          "get_topdown_island_view", &PathFinder::getTopDownIslandView,
+          R"(Returns the topdown view of the PathFinder's navmesh with island indicies at each point or -1 for non-navigable cells for a given vertical slice with eps slack.)",
```
0mdc, line 150: `indices`

tests/test_navmesh.py
```diff
@@ -316,3 +316,46 @@ def test_navmesh_islands(test_scene):
...
+        # check the computed data against the cache
+        assert np.array_equal(binary_topdown_map_cached, binary_top_down_map)
+        assert np.array_equal(islands_topdown_map_cached, island_top_down_map)
+
+        # example/test of simple image generation from island maps
+        from habitat_sim.utils.viz_utils import get_island_colored_map
+
+        island_colored_map_image = get_island_colored_map(island_top_down_map)
+        island_colored_map_image.save(filename + ".png")
+        # island_colored_map_image.show()
```
0mdc, line 361: Do we intend to compare `island_colored_map_image` against a baseline image?

If not, I would comment out this block to avoid generating unnecessary files.

src/esp/nav/PathFinder.cpp
aclegg3: I don't think this is the behavior we'd want. This would add 0.5 meters to the navmesh bounds before dividing it into `metersPerPixel` sized cells. If we want to round, probably should use `math.ceil(xspan / metersPerPixel)` instead. In either case, this is the difference of 1 pixel. :shrug:

tests/test_navmesh.py
```diff
@@ -330,8 +333,9 @@ def test_topdown_map(test_scene):
     hab_cfg = habitat_sim.utils.settings.make_cfg(cfg_settings)
 
     with habitat_sim.Simulator(hab_cfg) as sim:
-        scene_bb = sim.get_active_scene_graph().get_root_node().cumulative_bb
-        height = scene_bb.y().min
+        # use the lowest navmesh vert as the slice height
+        navmesh_verts = sim.pathfinder.build_navmesh_vertices(-1)
+        height = min([x[1] for x in navmesh_verts])
```
Skylion007, line 338:
```suggestion
        height = min(x[1] for x in navmesh_verts)
```

0mdc (approved):

src/esp/nav/PathFinder.cpp
```diff
@@ -1625,7 +1628,44 @@ PathFinder::Impl::getTopDownView(const float metersPerPixel,
...
+      curx = curx + metersPerPixel;
+    }
+    curz = curz + metersPerPixel;
+    curx = startx;
+  }
+
+  return topdownMap;
+}
+
+typedef Eigen::Matrix<int, Eigen::Dynamic, Eigen::Dynamic> MatrixXi;
```
Skylion007, line 1641: Should be use using, not typedef.

PR #1745, opened 2022-05-04
MVP Batch renderer WIP

eundersander:
See corresponding Hab-lab PR for details: https://github.com/facebookresearch/habitat-lab/pull/863

## How Has This Been Tested

See `src/tests/ReplayBatchRendererTest.cpp`.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1904, opened 2022-10-19
Replay Batch Renderer

0mdc:
## Motivation and Context

This changeset is a continuation of the work started in #1745 by @eundersander  (see https://github.com/facebookresearch/habitat-lab/pull/863 for full context).

## How Has This Been Tested

`ReplayBatchRendererTest` is successful.

## Types of changes

## Checklist


eundersander (approved):
LGTM!

mosra:
Just making sure -- is this meant to eventually also abstract away from using a `SceneNode` or that's to be done in a later PR?

On my side I have some *really dirty* code that attempts to make that abstraction, by storing an ID inside a `SceneNode*`. If you would find it useful, I can share the patch.

Skylion007 (approved):

src/tests/ReplayBatchRendererTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  ReplayBatchRendererConfiguration batchRendererConfig;
+  batchRendererConfig.sensorSpecifications = std::move(sensorSpecifications);
+  batchRendererConfig.numEnvironments = numEnvs;
+  ReplayBatchRenderer batchRenderer(batchRendererConfig);
+  auto renderer = batchRenderer.getRenderer();
+
+  std::vector<std::vector<char>> buffers(numEnvs);
+  std::vector<Mn::MutableImageView2D> imageViews;
```
mosra, line 143: FYI you could use Magnum's [Image2D](https://doc.magnum.graphics/magnum/classMagnum_1_1Image.html) instead of these two. It's an owning container so there's no need to maintain a separate buffer for it, and it's implicitly convertible to a `[mutable]ImageView2D` as well.

0mdc: I'll move on with this changeset and improve the test case when the real batch renderer will be integrated here.

PR #1920, opened 2022-10-27
Sim settings update

NakuraMino:
## Motivation and Context

The current `habitat_sim.utils.settings.make_cfg()` fn is limited to single first person sensors with the uuid `color_sensor`, `semantic_sensor` etc. 

e.g. 

```
if settings["color_sensor"]:
        color_sensor_spec = create_camera_spec(
            uuid="color_sensor",
            hfov=settings["hfov"],
            sensor_type=habitat_sim.SensorType.COLOR,
            sensor_subtype=habitat_sim.SensorSubType.PINHOLE,
        )
```
where `camera_sensor_spec.position = [0, settings["sensor_height"], 0]`.

Since many of the examples in `examples/colabs/` use a 1st person sensor AND a 3rd person sensor (e.g. https://github.com/facebookresearch/habitat-sim/blob/main/examples/tutorials/nb_python/ECCV_2020_Interactivity.py#L178), it would be best if you could flexibly add multiple cameras through `sim_settings`. 

## How Has This Been Tested

This passes CI/CD, and these changes are also tested through `pytest tests/*.py`. They all pass.

## Types of changes

## Checklist


dhruvbatra:
Thanks for thinking through these quality of usage issues.

src_python/habitat_sim/utils/settings.py
```diff
@@ -8,27 +8,26 @@
...
+            "width": 640,
+            "height": 480,
+            "sensor_height": 1.5,
+            "hfov": 90,
+            "sensor_subtype": habitat_sim.SensorSubType.PINHOLE,
+            "sensor_type": habitat_sim.SensorType.COLOR,
+            "position": [0, 1.5, 0],
+            "channels": 3,
+            "camera_type": habitat_sim.CameraSensorSpec,
+        },
```
aclegg3, line 26: I suggest we pull this out as a default_sensor_settings dict which is copied into the default settings dict.

Also, let's simplify a bit and remove "sensor_subtype", "sensor_type", "channels", "camera_type", since those are determined by the sensor type and users shouldn't need to both with these.

"sensor_height" is redundant if we have "position", so let's remove that one.

NakuraMino: sensor settings have been pulled out in its own dictionary. Sensor type, sensor subtypes etc are inferred from uuids, similar to how the original make_cfg fn did so, so fn won't work perfectly unless uuid is well-labeled, but I think that's a reasonable tradeoff.

src_python/habitat_sim/utils/settings.py
```diff
@@ -132,62 +70,71 @@ def create_fisheye_spec(**kw_args):
...
-    if settings["equirect_depth_sensor"]:
-        equirect_depth_sensor_spec = create_equirect_spec(
-            uuid="equirect_depth_sensor",
-            sensor_type=habitat_sim.SensorType.DEPTH,
-            channels=1,
+    for sensor_cfg in settings["sensors"]:
+        sensor_cfg = {**default_sensor_settings, **sensor_cfg}
+        uuid = sensor_cfg["uuid"]
+
+        sensor_type = (
```
aclegg3, line 90: Instead of pivoting off sensor name substrings, maybe we include "sensor_type" and "sensor_subtype" as fields in the sensor settings dict defaulting to color pinhole.

src_python/habitat_sim/utils/settings.py
```diff
@@ -11,24 +11,22 @@
...
-    "ortho_semantic_sensor": False,
-    "fisheye_rgba_sensor": False,
-    "fisheye_depth_sensor": False,
-    "fisheye_semantic_sensor": False,
-    "equirect_rgba_sensor": False,
-    "equirect_depth_sensor": False,
-    "equirect_semantic_sensor": False,
     "seed": 1,
     "physics_config_file": "data/default.physics_config.json",
+    "sensors": [
```
aclegg3, line 16: Instead of a list of dicts, maybe we use a dict of dicts? The key could be the uuid which would prevent duplication and allow dict merging overrides.

aclegg3 (commented):
A couple of design thoughts.

src_python/habitat_sim/utils/settings.py
NakuraMino: sure, I think it might be okay to pivot off sensor name for the camera type (i.e. `fisheye`, `equirect`) unless we also pass in the camera spec type (i.e. `habitat_sim.FisheyeSensorDoubleSphereSpec`, `habitat_sim.EquirectangularSensorSpec`) in the sensor_settings_dict. But sensor_type and sensor_subtype will be updated to use the sensor settings dict

jrreyna:
Fails lab_build_habitat on the build machine, but passes it on my local machine

NakuraMino:
@jrreyna for what its worth, #1924 updates all the `tutorials/` file, but its an incomplete PR at this point. Saw you were updating some tutorial files so just thought this might be helpful if you want to check it out. Good luck with this, and thanks for taking ownership :)

jrreyna:
> @jrreyna for what its worth, #1924 updates all the `tutorials/` file, but its an incomplete PR at this point. Saw you were updating some tutorial files so just thought this might be helpful if you want to check it out. Good luck with this, and thanks for taking ownership :)

Yes I saw, thank you! I was using that as a reference, I just wanted to put it all in one PR for simplicity

PR #1901, opened 2022-10-12
Dataset object physics processor

jrreyna:
## Motivation and Context

See PR #1882 and issue number #1900 to get the details. This PR is specifically for the physics processing aspect of dataset objects

## How Has This Been Tested

Locally and with CI tests

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1934, opened 2022-11-04
Fix an assertion caused by static plugins being duplicated in multiple dynamic libraries

mosra:
## Motivation and Context

A double-fix for the following assertion, which was encountered very often by @ykarmesh (original Slack thread for reference: [1](https://cvmlp.slack.com/archives/CFN5TAUSD/p1666301835495499), [2](https://cvmlp.slack.com/archives/G0131KVLBLL/p1666752514758909?thread_ts=1666633016.509929&cid=G0131KVLBLL)):

```
Assertion inserted.second failed at .../src/Corrade/PluginManager/AbstractManager.cpp:276
```

1. As of mosra/corrade@ba9a173c9310cb82fdf0bc83622b4e6828ac67d9, this assertion is gone. However because duplicated static plugins likely point to something being wrong with the way the project is built, it's still a warning.
2. Then, to avoid the warnings, I changed the build system to only link the static plugins to the leaf executables / dynamic modules, not to every static library, which is how it should be. Linking them to static libraries led to them being compiled several times over, only for the extra copies getting discarded by the final link step, increasing compile times for no reason. And ultimately, linking the static plugins to both the Magnum and the Habitat Python module (starting in #1901) led to them being present in two (or even more?) copies when both modules get `import`ed, which was the root cause for the original assertion.

@jrreyna & @ykarmesh -- if you want to have a quick fix for #1901 without merging all this, it should be enough to just `git pull` latest `master` in `src/deps/corrade`. You'll get warnings printed that way, but no assertion.

## How Has This Been Tested

Even though I originally thought this is an *impossible bug* (sorry), I managed to reproduce the error in an isolated scenario, *on all platforms*. So it's a bit strange that @jrreyna wasn't hitting this on Mac.

Nevertheless, the removed assert makes it a non-fatal problem. *Ideally* the Python module should also not print any warnings related to duplicate plugins, please let me know if it still does.


Skylion007 (approved):

mosra:
I'm now at a state where it's

- either building and passing tests on the CI, but linking the static plugins to both Python modules, and thus probably causing PluginManager warnings on systems that are not the CI,
- or not causing any warnings on systems that are not the CI, but failing on the CI because the Habitat Python module doesn't see the plugins linked to the Magnum Python module, dying on an error related to `ObjImporter`.

I'm going to leave this PR as-is for now, until I understand what's going on. The essential fix in Corrade is going to get included as part of #1901 instead.

PR #1947, opened 2022-11-23
Decouple gfx-replay player from scene graph; integrate the batch renderer into it

0mdc:
## Motivation and Context

Decouples `gfx::replay::Player` from `SceneNode` in order to allow the batch renderer to be integrated, and integrates the batch renderer. Gradually taken over by @mosra.

- [x] Introducing an opaque `NodeHandle` type, which can be cast into either `SceneNode*` or an index, depending on the implementation.
- [x] Moving `gfx::replay::Player` implementation into a pure virtual interface
- [x] Creating a base `AbstractReplayRenderer` class, moving the original `ReplayBatchRenderer` to just a `ReplayRenderer`
- [x] Creating a brand new `ReplayBatchRenderer` class
- [x] Adapting the test to run with both, expecting a ~matching output
- [x] Adding a GUI `replayer` utility, to which you can pass _a set of_ JSON replay files, which then get played back
  - Uses the batch renderer by default, pass `--classic` to it to switch to the classic one
  - Use `--profile` to print frame profiling info to the output
- [x] Ability to preload composite files in `AbstractReplayRenderer` (which would be a no-op for the classic one), hooking that up into the `replayer`
- [x] ~~Showing batch renderer frame stats in the profiling output~~ postponed, would require a change in Magnum's `FrameProfiler`
- [x] Ability to repeat the replay files or not
- [x] "Resolve" the Emscripten build failure -- attempting to use the classic renderer in the gfx-replay pipeline under Emscripten will assert. I'm not sure what exactly to do there, and I'm not sure if such workflow is even something we want to support moving forward.
- [x] Resolving remaining essential TODOs, like the commented-out cleanup on Player destruction

The following shows running 

```sh
./build/src/utils/replayer/replayer \
    ./data/RE_policy.replay.json \
    ./data/R_human_table_cabclosed_simplified.replay.fixed.json \
    ./data/v3_sc0_staging_00.replay.json \
    --preload ./data/composite-replicacad.gltf \
    --preload ./data/composite-ycb.gltf \
    --preload ./data/composite-fetch.gltf 
```
    
with replay files taken from [here](https://github.com/eundersander/habitat-sim/tree/batch_renderer_demo/data) (with the `.fixed.json` being patched to reference a real file instead of a `cubeSolid` placeholder) and composite files taken from [this](https://cvmlp.slack.com/archives/G0131KVLBLL/p1670629210548899) and [this](https://cvmlp.slack.com/archives/G0131KVLBLL/p1670624833220139) Slack thread:

replayer-2x

## How Has This Been Tested

~~CI is not passing because the PR is done from a forked repo.~~ 

~~I found a hack which makes it :green_circle:~~

No idea what's up with the CI. It used to pass, now it fails on `AttributeError: module 'torch._C' has no attribute '_cuda_setDevice'` again.


src/esp/gfx/replay/PlayerCallbacks.h
```diff
@@ -0,0 +1,71 @@
...
+  SetNodeTransform setNodeTransform_;
+  SetNodeSemanticId setNodeSemanticId_;
+  ChangeLightSetup changeLightSetup_;
+};
+
+/**
+ * @brief Factory method that instantiates PlayerCallbacks boilerplate to handle
+ * scene graph nodes.
+ */
+static inline PlayerCallbacks createSceneGraphPlayerCallbacks() {
```
mosra, line 43: The `static` isn't doing what you probably think it's doing -- it's the same as wrapping the function in a `namespace {`, makes a local copy of the function inside every file that includes it.

I'd actually make it deinlined in `PlayerCallbacks.cpp` instead, that way the header doesn't need to include basically anything, just a bunch of forward declarations for stuff from `gfx::` and `esp::assets::` and `Magnum/Magnum.h` for Magnum forward declarations.

src/esp/gfx/replay/PlayerCallbacks.h
```diff
@@ -0,0 +1,71 @@
...
+  LoadAndCreateRenderAssetInstance loadAndCreateRenderInstance_;
+  DeleteAssetInstance deleteAssetInstance_;
+  SetNodeTransform setNodeTransform_;
+  SetNodeSemanticId setNodeSemanticId_;
+  ChangeLightSetup changeLightSetup_;
+};
+
+/**
+ * @brief Factory method that instantiates PlayerCallbacks boilerplate to handle
+ * scene graph nodes.
```
mosra, line 41: Might be worth mentioning that only a subset of the callbacks is set up, not all of them.

src/esp/scene/SceneNode.h
```diff
@@ -335,6 +334,29 @@ void preOrderFeatureTraversalWithCallback(SceneNode& node, Callable&& cb) {
...
+    return;
+  }
+
+  // See also RigidBase setSemanticId. That function uses a prepared container
+  // of visual nodes, whereas this function traverses the subtree to touch all
+  // nodes (including visual nodes). The results should be the same.
+  auto cb = [&](esp::scene::SceneNode& node) {
+    node.setSemanticId(semanticId);
+  };
+  esp::scene::preOrderTraversalWithCallback(*node, cb);
```
mosra, line 357:
```suggestion
  preOrderTraversalWithCallback(*node, cb);
```

src/esp/scene/SceneNode.h
```diff
@@ -10,7 +10,6 @@
 #include <Corrade/Containers/Containers.h>
 #include <Corrade/Containers/Optional.h>
 #include <Magnum/Math/Range.h>
-#include <Magnum/SceneGraph/Object.h>
```
mosra, line 13: Why? `SceneNode` depends on `SceneGraph::Object`, it's just typedef'd to `MagnumObject`.

src/esp/scene/SceneNode.h
```diff
@@ -335,6 +334,29 @@ void preOrderFeatureTraversalWithCallback(SceneNode& node, Callable&& cb) {
       const_cast<const SceneNode&>(node), constFeatureCb);
 }
 
+/**
+ * @brief Set the semantic ID of a scene graph subtree.
+ *
+ * @param node Root node of the subtree.
+ * @param semanticId Semantic ID to apply to the subtree.
+ */
+static inline void setSemanticIdForSubtree(esp::scene::SceneNode* node,
```
mosra, line 343: No `static` for free functions here either, thanks ;)

And I'd put it into the `*.cpp` as well, it doesn't need to be in the header and you save a bit on compile times that way.

src/tests/GfxReplayTest.cpp
```diff
@@ -344,13 +348,15 @@ void GfxReplayTest::testPlayer() {
...
+  auto callbacks = esp::gfx::replay::createSceneGraphPlayerCallbacks();
+  callbacks.loadAndCreateRenderInstance_ =
       [&](const esp::assets::AssetInfo& assetInfo,
           const esp::assets::RenderAssetInstanceCreationInfo& creation) {
         return nullptr;
       };
-  auto dummyLightsCallback = [&](const esp::gfx::LightSetup& lights) -> void {};
-  esp::gfx::replay::Player player(dummyAssetsCallback, dummyLightsCallback);
+  callbacks.changeLightSetup_ =
+      [&](const esp::gfx::LightSetup& lights) -> void {};
```
mosra, line 358:
```suggestion
      [](const esp::gfx::LightSetup& lights) -> void {};
```

Maybe just my personal issue, but I don't like using capturing lambdas by default. The `std::function` overhead is bad enough as is, no need to make it even worse :) Same elsewhere.

mosra (commented):
Overall, this is exactly how I wanted this to be. A bunch of minor comments.

src/esp/gfx/replay/PlayerCallbacks.h
```diff
@@ -37,32 +44,11 @@ struct PlayerCallbacks {
 };
 
 /**
- * @brief Factory method that instantiates PlayerCallbacks boilerplate to handle
- * scene graph nodes.
+ * @brief Factory method that instantiates PlayerCallbacks boilerplate for
+ * callbacks that handle scene graph nodes. Note that this does not provide an
+ * implementation for all callbacks.
```
mosra, line 49: :D Well, that ... still means the user has to go to the source to see which callbacks aren't set. My intent with the documentation is to always provide enough info so the user *doesn't* need to browse the source. I'd say something like this:

``` 
 * [...] Note that you still need to provide
 * @ref PlayerCallbacks::loadAndCreateRendererInstance_ and 
 * @ref PlayerCallbacks::changeLightSetup_ yourself.
```

I assume I won't be the only one integrating a new renderer (Eric wanted this to work with arbitrary 3rd party renderers), so it'll benefit others as well.

src/esp/sim/ReplayBatchRenderer.cpp
```diff
@@ -48,12 +51,12 @@ ReplayBatchRenderer::ReplayBatchRenderer(
...
         parentNode, cfg.sensorSpecifications);
 
-    envs_.emplace_back(EnvironmentRecord{
-        .player_ = gfx::replay::Player(assetCallback, lightCallback),
-        .sceneID_ = sceneID,
-        .semanticSceneID_ = semanticSceneID,
-        .sensorParentNode_ = &parentNode,
-        .sensorMap_ = std::move(sensorMap)});
+    envs_.emplace_back(
+        EnvironmentRecord{.player_ = gfx::replay::Player(callbacks),
```
Skylion007, line 55:
```suggestion
        EnvironmentRecord{.player_ = gfx::replay::Player(std::move(callbacks)),
```

src/esp/gfx/replay/ReplayManager.h
```diff
@@ -37,13 +38,8 @@ class ReplayManager {
...
    */
-  void setPlayerCallbacks(
-      const Player::LoadAndCreateRenderAssetInstanceCallback&
-          loadAndCreateRenderAssetInstanceCallback,
-      const Player::ChangeLightSetupCallback& changeLightSetupCallback) {
-    loadAndCreateRenderAssetInstanceCallback_ =
-        loadAndCreateRenderAssetInstanceCallback;
-    changeLightSetupCallback_ = changeLightSetupCallback;
+  void setPlayerCallbacks(const PlayerCallbacks& playerCallbacks) {
+    playerCallbacks_ = playerCallbacks;
```
Skylion007, line 42:
```suggestion
  void setPlayerCallbacks(PlayerCallbacks playerCallbacks) {
    playerCallbacks_ = std::move(playerCallbacks);
```
Do not copy this if you don't need to (ie if it's an rvalue).

src/esp/gfx/replay/Player.h
```diff
@@ -40,20 +33,11 @@ namespace replay {
...
-      std::function<void(const gfx::LightSetup& lights)>;
-
   /**
    * @brief Construct a Player.
    * @param callback A function to load and create a render asset instance.
    */
-  explicit Player(const LoadAndCreateRenderAssetInstanceCallback&
-                      loadAndCreateRenderAssetInstanceCallback,
-                  const ChangeLightSetupCallback& lightSetupCallback);
+  explicit Player(const PlayerCallbacks& callbacks);
```
Skylion007, line 40:
```suggestion
  explicit Player(PlayerCallbacks callbacks);
```
To match the other comment I made.

src/esp/gfx/replay/Player.cpp
```diff
@@ -30,12 +31,7 @@ Keyframe Player::keyframeFromString(const std::string& keyframe) {
   return res;
 }
 
-Player::Player(const LoadAndCreateRenderAssetInstanceCallback&
-                   loadAndCreateRenderAssetInstanceCallback,
-               const ChangeLightSetupCallback& changeLightSetupCallback)
-    : loadAndCreateRenderAssetInstanceCallback(
-          loadAndCreateRenderAssetInstanceCallback),
-      changeLightSetupCallback(changeLightSetupCallback) {}
+Player::Player(const PlayerCallbacks& callbacks) : callbacks_(callbacks) {}
```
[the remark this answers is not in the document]
mosra: Ah, yes, good point -- I completely forgot that `std::function` is this nasty complex type.

src/esp/sim/Simulator.cpp
```diff
@@ -687,15 +688,17 @@ void Simulator::reconfigureReplayManager(bool enableGfxReplaySave) {
...
-      [this](const gfx::LightSetup& lights) -> void {
-        this->setLightSetup(lights);
-      });
+      -> gfx::replay::GfxReplayNode* {
+    return loadAndCreateRenderAssetInstance(assetInfo, creation);
+  };
+  callbacks.changeLightSetup_ = [this](const gfx::LightSetup& lights) -> void {
+    setLightSetup(lights);
+  };
+  gfxReplayMgr_->setPlayerCallbacks(callbacks);
```
Skylion007, line 701:
```suggestion
  gfxReplayMgr_->setPlayerCallbacks(std::move(callbacks));
```

Skylion007 (approved):

src/esp/sim/Simulator.h
```diff
@@ -919,7 +920,7 @@ class Simulator {
    * @param assetInfo the asset to load
    * @param creation how to create the instance
    */
-  scene::SceneNode* loadAndCreateRenderAssetInstance(
+  gfx::replay::GfxReplayNode* loadAndCreateRenderAssetInstance(
```
eundersander, line 923: It feels weird to create this dependency on gfx-replay here in `Simulator.h`. I'm thinking about hypothetical other users of this function besides gfx-replay. It also seems more intuitive (and more useful) that `loadAndCreateRenderAssetInstance` would directly return a usable `SceneNode*`. We aren't trying to abstract away the concept of scene nodes in `Simulator.h`, right?

(In a similar vein, we can imagine that the new Magnum batch renderer's `addMeshHierarchy` will still return something like an `int` , not a `gfx::replay::GfxReplayNode*`.)

src/esp/gfx/replay/PlayerCallbacks.h
```diff
@@ -0,0 +1,57 @@
...
+namespace replay {
+
+// Represents either a scene::SceneNode* or batch renderer entity index.
+class GfxReplayNode;
+
+/**
+ * @brief Callbacks that contain the implementation for gfx::replay::Player.
+ */
+struct PlayerCallbacks {
+  using LoadAndCreateRenderAssetInstance = std::function<GfxReplayNode*(
```
eundersander, line 26: Why not an interface class with virtual member functions? Why all these function callbacks?

Member functions have a few benefits over callbacks that I see:
* a more common paradigm compared to passing around a pack of callbacks
* possibly lower runtime overhead (virtual function call versus `std::function`)
* elegant way to either provide default implementation or leave it undefined (pure virtual, triggering compiler error)

eundersander (commented):
I left some comments.

src/esp/gfx/replay/PlayerCallbacks.h
0mdc: An interface would make a lot more sense, I completely agree.

The idea is that, in this state, it will be easier for @mosra to integrate his changes into the replay batch renderer.

I would refactor this into an interface after the usage patterns are all present.

src/esp/sim/ReplayBatchRenderer.cpp
```diff
@@ -29,15 +29,18 @@ ReplayBatchRenderer::ReplayBatchRenderer(
   sceneManager_ = scene::SceneManager::create_unique();
 
   for (int envIdx = 0; envIdx < config_.numEnvironments; ++envIdx) {
-    auto assetCallback =
+    auto callbacks = gfx::replay::createSceneGraphPlayerCallbacks();
```
Skylion007, line 32: @eundersander Not sure if virtual methods would make sense here. Some of the callbacks require capturing local scope which would be difficult to do with virtual methods. I would venture that this is the reason they were callbacks before.

src/esp/gfx/replay/PlayerCallbacks.h
Skylion007: As long as there is an elegant way for the callbacks to capture local variables like the resource manager instance and "this" pointer, that would work, but I am not sure there is a good way to do that. :/

Skylion007:
FYI: saw some minor perf issues while looking at this PR (unrelated to this change) and fixed them in #1954

src/esp/sim/Simulator.h
mosra: This is reverted in ae16bbcca2925ec564a28ed7c9198a6028155844.

src/esp/gfx/replay/PlayerCallbacks.h
mosra: This is implemented in e2ddaaa3bae55092e7a6dcd2a2b489212ca26473, see the commit message for details.

A lambda is just a convenience syntax sugar for a stateful class, which is what I turned the implementation into. Was pretty frictionless in the end.

src/esp/sim/ReplayBatchRenderer.h
```diff
@@ -23,75 +22,211 @@ class Renderer;
...
+  Corrade::Containers::Array<EnvironmentRecord> envs_;
+
+  /* If standalone_ is true, renderer_ contains a RendererStandalone */
+  bool standalone_;
+  Corrade::Containers::Pointer<esp::gfx_batch::Renderer> renderer_;
 
-  ESP_SMART_POINTERS(ReplayBatchRenderer)
+  Corrade::Containers::String theOnlySensorName_;
+  Mn::Matrix4 theOnlySensorProjection_;
 };
```
0mdc, line 230: `ESP_SMART_POINTERS(ReplayBatchRenderer)` is missing here.

mosra: Yeah, I was wondering what do we actually need it for, and tried not adding it here :) Especially given this class would be used only through the base interface.

mosra: Added.

src/esp/gfx/replay/Player.cpp
```diff
@@ -25,17 +55,22 @@ void Player::readKeyframesFromJsonDocument(const rapidjson::Document& d) {
...
-          loadAndCreateRenderAssetInstanceCallback),
-      changeLightSetupCallback(changeLightSetupCallback) {}
+Keyframe Player::keyframeFromStringUnwrapped(
+    const Cr::Containers::StringView keyframe) {
+  Keyframe res;
+  rapidjson::Document d;
+  d.Parse<0>(keyframe.data(), keyframe.size());
+  esp::io::fromJsonValue(d, res);
+  return res;
+}
```
mosra, line 70: @eundersander I went ahead and added an optimized variant of the keyframe loading, see https://github.com/facebookresearch/habitat-sim/pull/1947/commits/136211fa611887516f4c0a49738c380b2ea54b24 for the full change. In particular:

- it takes a string view, because otherwise it'd have to allocate a `std::string` for every keyframe sliced out of the input file
- it expects it to contain directly the keyframe object, i.e. containing `"loads"` and `"creations"`, not a `"keyframe"
- and it passes an explicit size because the keyframe data can get rather huge and running `strlen()` on it again feels unnecessary

I kep the original API intact to avoid breaking any existing use. Is that fine or should we switch all code to use the new API?

src/esp/gfx/replay/Player.cpp
```diff
@@ -101,10 +136,7 @@ void Player::close() {
...
 void Player::clearFrame() {
   for (const auto& pair : createdInstances_) {
-    // TODO: use NodeDeletionHelper to safely delete nodes owned by the Player.
-    // the deletion here is unsafe because a Player may persist beyond the
-    // lifetime of these nodes.
-    delete pair.second;
+    implementation_->deleteAssetInstance(pair.second);
   }
   createdInstances_.clear();
   assetInfos_.clear();
```
mosra, line 142: @eundersander continuing with our conversation, I feel like I should add a `AbstractPlayerImplementation::clear()` interface, which either `delete`s in the SceneGraph or `clear()`s in the batch renderer; and call that from here instead. Does that sound alright?

mosra: A sketch for this is in e96df3ba9c372473d7e5013564451a3d0f7c2d06.

src/esp/gfx/replay/Player.cpp
```diff
@@ -25,17 +62,22 @@ void Player::readKeyframesFromJsonDocument(const rapidjson::Document& d) {
...
+    const Cr::Containers::StringView keyframe) {
+  Keyframe res;
+  rapidjson::Document d;
+  d.Parse<0>(keyframe.data(), keyframe.size());
+  esp::io::fromJsonValue(d, res);
+  return res;
+}
+
+Player::Player(AbstractPlayerImplementation& implementation)
+    : implementation_{implementation} {}
```
Skylion007, line 80: Why not? 
```suggestion
Player::Player(AbstractPlayerImplementation implementation)
    : implementation_{std::move(implementation)} {}
```

src/esp/gfx/replay/Player.h
```diff
@@ -40,20 +151,19 @@ namespace replay {
...
    * @brief Construct a Player.
-   * @param callback A function to load and create a render asset instance.
+   *
+   * The @p implementation is assumed to be owned by the caller for the whole
+   * lifetime of the @ref Player instance.
    */
-  explicit Player(const LoadAndCreateRenderAssetInstanceCallback&
-                      loadAndCreateRenderAssetInstanceCallback,
-                  const ChangeLightSetupCallback& lightSetupCallback);
+  explicit Player(AbstractPlayerImplementation& implementation);
```
Skylion007, line 160:
```suggestion
explicit Player(AbstractPlayerImplementation implementation);
```
same as above

src/esp/gfx/replay/ReplayManager.h
```diff
@@ -60,9 +56,7 @@ class ReplayManager {
 
  private:
   std::shared_ptr<Recorder> recorder_;
-  Player::LoadAndCreateRenderAssetInstanceCallback
-      loadAndCreateRenderAssetInstanceCallback_;
-  Player::ChangeLightSetupCallback changeLightSetupCallback_;
+  std::unique_ptr<AbstractPlayerImplementation> playerImplementation_;
```
Skylion007, line 59: Nice!

src/esp/gfx/replay/Player.cpp
Skylion007: Fair enough.

Skylion007: wait am confused, won't this still create duplicate instance? The AbstractPlayer caller and `_implementation` or am I misunderstanding something?

mosra: `implementation_` is a reference as well :)

src/esp/sim/ReplayBatchRenderer.cpp
```diff
@@ -79,9 +171,21 @@ ReplayBatchRenderer::ReplayBatchRenderer(
...
+      auto& visualSensor =
+          static_cast<esp::sensor::VisualSensor&>(kv.second.get());
+      renderer_->bindRenderTarget(visualSensor);
+    }
+  }
 }
 
-ReplayBatchRenderer::~ReplayBatchRenderer() {
+ReplayRenderer::~ReplayRenderer() {
   ESP_DEBUG() << "Deconstructing ReplayBatchRenderer";
```
0mdc, line 189: `"Deconstructing Replay~~Batch~~Renderer"`

src/esp/sim/ReplayBatchRenderer.cpp
```diff
@@ -93,15 +197,46 @@ ReplayBatchRenderer::~ReplayBatchRenderer() {
...
+  return envs_[envIndex].player_;
+}
+
+void ReplayRenderer::doSetSensorTransform(unsigned envIndex,
+                                          const std::string& sensorName,
+                                          const Mn::Matrix4& transform) {
+  auto& env = envs_[envIndex];
+
+  ESP_CHECK(env.sensorMap_.count(sensorName),
+            "ReplayBatchRenderer::setSensorTransform: sensor "
```
0mdc, line 224: "Replay~~Batch~~Renderer::..."

src/esp/sim/ReplayBatchRenderer.cpp
mosra: More like I should remove that useless debug print...

src/esp/sim/ReplayBatchRenderer.h
```diff
@@ -23,29 +22,123 @@ class Renderer;
...
 
-class ReplayBatchRenderer {
+class ReplayRenderer : public AbstractReplayRenderer {
  public:
-  class EnvironmentRecord {
-   public:
+  struct EnvironmentRecord {
+    explicit EnvironmentRecord(
+        std::unique_ptr<gfx::replay::AbstractSceneGraphPlayerImplementation>
+            playerImplementation)
```
Skylion007, line 137:
```suggestion
        std::unique_ptr<gfx::replay::AbstractSceneGraphPlayerImplementation>&&
            playerImplementation)
```

mosra: Does this actually matter? It's not like the type would be copyable, one has to `std::move()` it always.

src/esp/gfx/replay/Player.cpp
mosra: Actually I went with a shared_ptr in the end, see https://github.com/facebookresearch/habitat-sim/pull/1947/commits/e85af83fb02f843a4b9a7687db62ea1db22c861c for details.

.clang-tidy
```diff
@@ -110,3 +112,18 @@ HeaderFilterRegex: "src/esp/"
 #   really improve readability in any way. Keeping this as a warning tho, as it
 #   *might* actually be useful in certain cases, just not treating it as an
 #   error.
+# performance-no-int-to-ptr
+#   "error: integer to pointer cast pessimizes optimization opportunities"
+#   triggered by gfx::replay::NodeHandle, which can be either a pointer (for
+#   the SceneGraph backend) or an arbitrary integer value (for the new
+#   data-oriented backend). Clang Tidy, tell me, how else am I supposed to
+#   implement type-safe handles?!
```
Skylion007, line 120: With `//NOLINTNEXTLINE`? I'd rather keep this check if possible as it's otherwise useful (mostly for catching types that should ptr_diff.

src/esp/sim/ReplayBatchRenderer.cpp
```diff
@@ -168,8 +362,179 @@ scene::SceneNode* ReplayBatchRenderer::loadAndCreateRenderAssetInstance(
...
+                 "ReplayBatchRenderer: expecting exactly one sensor", );
+  const auto& sensor = static_cast<esp::sensor::CameraSensorSpec&>(
+      *cfg.sensorSpecifications.front());
+
+  gfx_batch::RendererConfiguration configuration;
+  configuration.setTileSizeCount(Mn::Vector2i{sensor.resolution}.flipped(),
+                                 environmentGridSize(cfg.numEnvironments));
+  if ((standalone_ = cfg.standalone))
+    renderer_.emplace<gfx_batch::RendererStandalone>(
+        configuration, gfx_batch::RendererStandaloneConfiguration{});
```
0mdc, line 382: We can carry the GPU ID here:
```
renderer_.emplace<gfx_batch::RendererStandalone>(
        configuration, gfx_batch::RendererStandaloneConfiguration{}
          .setCudaDevice(cfg.gpuDeviceId));
```

mosra: Good point, I'll add this -- seems important.

.clang-tidy
mosra: I thought about this, and I don't see a scenario where *explicitly reinterpreting* pointer as an integer would be harming any optimization opportunities. In other words, given that we ban C-style casts with `-Wold-style-cast` anyway, this is hard to do by accident, and if there's a `reinterpret_cast` from an integer to a pointer or back, it's probably intentional -- i.e., to implement a handle type.

src/esp/sim/ReplayBatchRenderer.cpp
mosra: Ah well, the default `ReplayRendererConfiguration::gpuDeviceId` is `0`, but `0` means the first CUDA device -- i.e., if you have no NV cards (or the NV card disconnected as in my case), it would fail to create the context.

The default should be `-1` I think, i.e. picking whatever the primary GPU is, so it's not restricted to work only on NVidias. Can I change the default to `-1`?

0mdc: Oh that's right. `-1` by default makes sense then.

mosra: Not sure if I want to go through the whole CI process again -- we want to merge ASAP -- could you add this in the followup PR? :)

0mdc: Sounds good!

eundersander: The default `ReplayRendererConfiguration::gpuDeviceId` is `0` because the default `SimulatorConfiguration::gpuDeviceId` is `0`. @0mdc , if you're making a change, consider making it in both places.

src/esp/sim/ReplayBatchRenderer.h
```diff
@@ -23,29 +22,123 @@ class Renderer;
...
+
+  ESP_SMART_POINTERS(AbstractReplayRenderer)
 };
 
-class ReplayBatchRenderer {
+class ReplayRenderer : public AbstractReplayRenderer {
  public:
-  class EnvironmentRecord {
-   public:
+  struct EnvironmentRecord {
```
eundersander, line 134: Since this struct is only used by one backend, couldn't it declare the actual PlayerImplementation member here?

src/esp/sim/ReplayBatchRenderer.h
```diff
@@ -54,42 +147,103 @@ class ReplayBatchRenderer {
...
+  void doRender(Magnum::GL::AbstractFramebuffer& framebuffer) override;
+
+  /* If standalone_ is true, renderer_ contains a RendererStandalone. Has to be
+     before the EnvironmentRecord array because Player calls
+     gfx_batch::Renderer::clear() on destruction. */
+  bool standalone_;
+  Corrade::Containers::Pointer<esp::gfx_batch::Renderer> renderer_;
+
+  // TODO pimpl all this?
+  struct EnvironmentRecord {
```
eundersander, line 238: This use of a virtual interface per env doesn't strike me as very data-oriented. I was hoping to only see virtual-call overhead at the renderer level, not per environment (remember we may eventually have thousands of envs in the batch use case).

src/utils/replayer/replayer.cpp
```diff
@@ -0,0 +1,226 @@
...
+#include <Corrade/Utility/Json.h>
+#include <Corrade/Utility/Path.h>
+#include <Magnum/DebugTools/FrameProfiler.h>
+#include <Magnum/GL/DefaultFramebuffer.h>
+#include <Magnum/Platform/GlfwApplication.h>
+
+#include "esp/sensor/CameraSensor.h"
+#include "esp/sim/ReplayBatchRenderer.h"
+
+namespace {
```
eundersander, line 13: How about some file-level/class-level documentation about this Replayer app? Even just a sentence describing what it is.

eundersander (commented):
I left some comments.

src/esp/sim/ReplayBatchRenderer.h
```diff
@@ -23,29 +22,123 @@ class Renderer;
...
+      Corrade::Containers::ArrayView<const Magnum::MutableImageView2D>
+          imageViews) = 0;
+
+  virtual void doRender(Magnum::GL::AbstractFramebuffer& framebuffer) = 0;
+
+  ESP_SMART_POINTERS(AbstractReplayRenderer)
 };
 
-class ReplayBatchRenderer {
+class ReplayRenderer : public AbstractReplayRenderer {
```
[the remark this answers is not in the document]
mosra: `ClassicReplayRenderer`?

src/esp/sim/ReplayBatchRenderer.h
mosra: Hmm. Are you more concerned about the size of the `EnvironmentRecord` struct (due to the vtables and other overhead), or about the one-virtual-call-per-transformation-update overhead?

The former could be solved relatively easily by having the Player remember `envId`, because that's the only single thing that differs between the instances, and I did it this way because it was the most direct change from the `std::function`s that were there originally (where *each* `std::function` was a virtual interface on its own). This change isn't hard, won't affect the public API in any way, and I propose doing it in the next iteration, as this PR is big enough already :)

The latter, one-virtual-call-per-transformation-update would have to be solved in the Player itself, where it would have to put the (parsed) transforms into some memory location and then submit them all at once with a single (virtual) call. The batch renderer is already prepared for such workflow, as it [exposes the transformation array directly](https://aihabitat.org/docs/habitat-sim/classesp_1_1gfx__batch_1_1Renderer.html#acfb2439f0fe5e5e7397eaf3d517a49ab), for the classic renderer it'd be about looping over the passed array and updating. That's a more substantial change, but still doable, and I think it should go into a standalone PR on its own.

src/utils/replayer/replayer.cpp
mosra: Oh yeah, definitely, kinda forgot to do this when the code transformed from a rough test to an actual utility :D

src/esp/sim/ReplayBatchRenderer.h
mosra: Yes, it could, but the problem is that the Player takes it via a `shared_ptr` because of shitty reasons explained in a382d41d0a3a24f9926e0b4b094124d010d125d4 (pybind11 ownership issues).

I can try reverting this commit and try to figure out a better way to deal with pybind11.

src/esp/sim/ReplayBatchRenderer.h
```diff
@@ -23,29 +22,123 @@ class Renderer;
...
  public:
```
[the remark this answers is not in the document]
mosra: Sure, that makes sense.

src/esp/sim/ReplayBatchRenderer.h
eundersander: >the one-virtual-call-per-transformation-update overhead?
This is my concern (not the struct size).

>That's a more substantial change, but still doable, and I think it should go into a standalone PR on its own.
Yep, agree it's a substantial change. We should do this in the future.

>it would have to put the (parsed) transforms into some memory location and then submit them all at once with a single (virtual) call.
Consider that the Python caller will have gathered all keyframes before passing them on to the batch renderer. So we should change the Python API to pass all keyframes at once (thus no need to use additional memory).

src/esp/sim/ReplayBatchRenderer.h
eundersander: Ok, I don't really care about this small inefficiency for the legacy/stub batch renderer. You can just leave it as-is.

src/esp/sim/ReplayBatchRenderer.h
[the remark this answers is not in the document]
eundersander: Sounds good to me!

src/esp/sim/ReplayBatchRenderer.h
mosra: Just to clarify, I meant having something like

```cpp
virtual void setNodeTransforms(
    StridedArrayView1D<const NodeHandle> nodes,
    StridedArrayView1D<const Magnum::Vector3> translations,
    StridedArrayView1D<const Magnum::Quaternion> rotations);
```

in the `AbstractPlayerImplementation`, thus having exactly one virtual call per frame. Or whichever other variant where the data get passed in as a reference to some regularly shaped memory.

eundersander: Ultimately, when we go to optimize the batch replay renderer, I suspect we'll just avoid using the Player base class. A keyframe is a straightforward data structure to map to a sequence of new instance and light transforms (and the other ops like instance creation); we should just write that code directly. I don't care that much about sharing this Keyframe-parsing logic between the two backends. I'm okay with some code duplication. Sorry I didn't bring this up before; I do think our solution here in this PR (a shared Player base class) is fine for the near term.

src/utils/replayer/replayer.cpp
mosra: Done in https://github.com/facebookresearch/habitat-sim/pull/1947/commits/14a81f71f2a90b2d85b72b4a9edd75ce83a2a026 -- it's a `--help` text, similarly to the `viewer` app.

src/esp/sim/ReplayBatchRenderer.h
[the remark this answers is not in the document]
mosra: Done in https://github.com/facebookresearch/habitat-sim/pull/1947/commits/098e914c46584048073588beeaa4c862eb8a125e.

src/esp/sim/ReplayBatchRenderer.h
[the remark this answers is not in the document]
mosra: Done in https://github.com/facebookresearch/habitat-sim/pull/1947/commits/cadfa0f472f24b348d183c3f69c8fd61c2ae0c60.

.clang-tidy
```diff
@@ -64,6 +64,8 @@ readability-uniqueptr-delete-release,
 -modernize-use-trailing-return-type,
 -bugprone-easily-swappable-parameters,
 -bugprone-forward-declaration-namespace,
+-performance-no-int-to-ptr,
+-misc-misplaced-const,
```
Skylion007, line 68: This is a specific optin. Just delte it from clang-tidy from above.

src/esp/sim/ReplayBatchRenderer.h
mosra: Okay, so -- anything left to do here regarding this point? :) As I see it, it's all for subsequent PRs.

eundersander: Nope, all for subsequent PRs.

PR #1951, opened 2022-11-28
[bugfix] - skip non-triangle mesh primitives in mesh join for navmesh recomputation

aclegg3:
## Motivation and Context

@mosra fixed a bug with non-triangle primitive support in #1888. This worked for physics, but was not being applied to navmesh recomputation resulting in artifacts for affected scenes. See [comments ](https://github.com/facebookresearch/habitat-sim/pull/1888#issuecomment-1312803701)for details. 

## How Has This Been Tested

Locally with FP scene 102816627.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -2971,12 +2971,17 @@ void ResourceManager::joinHierarchy(
         meshes_.at(node.meshIDLocal + metaData.meshIndex.first)
             ->getCollisionMeshData();
     int lastIndex = mesh.vbo.size();
-    for (const auto& pos : meshData.positions) {
-      mesh.vbo.push_back(Mn::EigenIntegration::cast<vec3f>(
-          transformFromLocalToWorld.transformPoint(pos)));
-    }
-    for (const auto& index : meshData.indices) {
-      mesh.ibo.push_back(index + lastIndex);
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

src/esp/assets/ResourceManager.cpp
```diff
@@ -2971,12 +2971,17 @@ void ResourceManager::joinHierarchy(
...
+      ESP_WARNING() << "Unsupported mesh primitive in join: "
+                    << meshData.primitive << Mn::Debug::nospace << ", skipping";
+    } else {
+      for (const auto& pos : meshData.positions) {
+        mesh.vbo.push_back(Mn::EigenIntegration::cast<vec3f>(
+            transformFromLocalToWorld.transformPoint(pos)));
+      }
+      for (const auto& index : meshData.indices) {
+        mesh.ibo.push_back(index + lastIndex);
+      }
```
Skylion007, line 2984: Actually, @aclegg3 refactor the if statement to an early return instead.

aclegg3: We need to skip instead because we still have a recursive step on the children of this node.

mukulkhanna:
Thanks Alex! Yes, this seems to be working. Here's a before (without the above commit) and after (with the above commit) of a sample scene's topdown map:

PR #1961, opened 2022-12-06
Assorted updates to the batch renderer

mosra:
## Motivation and Context

Contains various minor updates and fixes that *don't* depend on changes from #1947, thus hopefully possible to be reviewed & merged in isolation:

- Ability to bake-in an initial transformation for added hierarchy (i.e., without it being overwritten next time the hierarchy transform is updated). Will be used for specifying initial scale and coordinate frame correction and should fix 99.7% of nasty transformation bugs I had in the initial gfx-replay demo screenshots in September.
- Opt-in flag to generate texture mipmaps. This is something that Vulkan will not give us for free in the driver (and so the assets will need to contain full mip pyramids themselves), so it's not enabled by default, but it's currently used to have a pixel-perfect match with the classic renderer.
- Adds a more efficient way to retrieve rendered images -- instead of putting them into a newly-allocated memory, it reads them to a pre-allocated view. 
- **Extracted projection matrix calculation out of `CameraSensor`** to a utility function on the `CameraSensorSpec` so I can use it even in scenegraph-less contexts. I felt like redoing all that calculation from scratch on my side would be too error prone, so I did it like this, and simplified the code slightly in the process. I hope it doesn't conflict with the upcoming/ongoing sensor refactor. The new API entrypoint isn't used anywhere at the moment, but it'll be in the `ReplayBatchRenderer` that depends on #1947.
- Adds `gfx::RenderTarget::blitRgbaTo()` which is like `blitRgbaToDefault()` but with an ability to choose an arbitrary rectangle of an arbitrary framebuffer, which I need for feature parity in the batch renderer replay GUI. The `blitRgbaToDefault()` is now built on top of this new API.

## How Has This Been Tested

The first two parts have a corresponding update in `GfxBatchRendererTest`. 

The projection matrix is I assume tested somewhere already. I tried to exactly match the original behavior, including taking the `hfov` override from `CameraSensor` instead of `CameraSensorSpec`, please let me know if something feels off.

### DIFF
--- a/src/esp/gfx_batch/RendererStandalone.cpp
+++ b/src/esp/gfx_batch/RendererStandalone.cpp
@@ -11,6 +11,7 @@
 #include <Magnum/GL/Renderbuffer.h>
 #include <Magnum/GL/RenderbufferFormat.h>
 #include <Magnum/Image.h>
+#include <Magnum/ImageView.h>
 #include <Magnum/PixelFormat.h>
 
 #ifdef MAGNUM_TARGET_EGL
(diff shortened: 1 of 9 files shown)


0mdc:
Looks good!

0mdc (approved):

src/esp/gfx_batch/RendererStandalone.cpp
```diff
@@ -164,13 +165,42 @@ Mn::Image2D RendererStandalone::colorImage() {
                                   colorFramebufferFormat());
 }
 
+void RendererStandalone::colorImageInto(const Magnum::Range2Di& rectangle,
+                                        const Mn::MutableImageView2D& image) {
```
0mdc, line 169: Nitpick: I would rename `colorImageInto` and `depthImageInto` to something a bit more explicit.

0mdc (approved):

PR #1964, opened 2022-12-08
Batch renderer integration into Python viewer.

0mdc:
## Motivation and Context

This changeset integrates the batch renderer into the Python viewer.

New command line args:
`--enable-batch-renderer` enables batch rendering.
`--num-environments` specifies the number of concurrent environments to render.
`--width`, `--height` specify the window resolution.
`--composite-files` allows to pass composite gltf files to Magnum.

Example usage:
`python examples/viewer.py --enable-batch-renderer --num-environments 6 --width 1024 --height 896 --scene "data/scene_datasets/habitat-test-scenes/skokloster-castle.glb"`

batch_renderer_6env

`python examples/viewer.py --enable-batch-renderer --num-environments 64 --width 1024 --height 1024`

batch_renderer_64env

## How Has This Been Tested

This was tested locally, and in automated tests here: #1947.
No regression found when testing the viewer with the classical renderer.

## Types of changes

## Checklist


examples/viewer.py
```diff
@@ -46,24 +46,43 @@ class HabitatSimInteractiveViewer(Application):
...
+        self.enable_batch_renderer: bool = self.sim_settings["enable_batch_renderer"]
+        self.num_env: int = (
+            self.sim_settings["num_environments"] if self.enable_batch_renderer else 1
+        )
+
+        # Compute environment camera resolution based on the number of environments to render in the window.
+        surface_size: tuple(int, int) = (
+            self.sim_settings["window_width"],
+            self.sim_settings["window_height"],
+        )
```
mosra, line 60: In the `replayer` I have the camera resolution as the configuration parameter, with the window size being inferred from that. Not sure which of the two is the better approach, but we should probably use the same in both :)

examples/viewer.py
```diff
@@ -1036,15 +1122,49 @@ def next_frame() -> None:
         action="store_true",
         help="Override configured lighting to use synthetic lighting for the stage.",
     )
+    parser.add_argument(
+        "--enable_batch_renderer",
```
mosra, line 1126:
```suggestion
        "--enable-batch-renderer",
```

I think this is more consistent with option naming everywhere else. Same below.

examples/viewer.py
```diff
@@ -263,15 +286,30 @@ def draw_event(
...
+                # Render
+                self.replay_renderer.render(mn.gl.default_framebuffer)
+        else:
+            self.sim[0]._Simulator__sensors[keys[0]][keys[1]].draw_observation()
+            agent = self.sim[0].get_agent(keys[0])
+            self.render_camera[0] = agent.scene_node.node_sensor_suite.get(keys[1])
+            self.debug_draw(0)
+            self.render_camera[0].render_target.blit_rgba_to_default()
+            # draw CPU/GPU usage data and other info to the app window
+            # TODO: Doesn't work with batch renderer.
```
mosra, line 310: Should be fixed in https://github.com/facebookresearch/habitat-sim/pull/1947/commits/d025013ea6f00b9bb54e299de48d69e2a227bded.

src/esp/sim/ReplayBatchRenderer.cpp
```diff
@@ -353,6 +353,7 @@ gfx::replay::NodeHandle ReplayRenderer::loadAndCreateRenderAssetInstance(
 
 ReplayBatchRenderer::ReplayBatchRenderer(
     const ReplayRendererConfiguration& cfg) {
+  flextGLInit(Magnum::GL::Context::current()); // TODO: Avoid globals duplications across SOs.
```
0mdc, line 356: @mosra : FYI

examples/viewer.py
0mdc: For user-driven visualization purposes, I think it makes more sense to define a window size, then subdivide the surface based on the environment count. It saves a napkin calculation.

Eventually, this viewer will become an interface exposed to the lab. It will be up to the implementer to figure out what makes sense to them.

src/esp/sim/Simulator.cpp
```diff
@@ -195,11 +195,6 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
     flextGLInit(Magnum::GL::Context::current());
 #endif
     renderer_->acquireGlContext();
-  } else {
-    CORRADE_ASSERT(
-        !Magnum::GL::Context::hasCurrent(),
-        "Simulator::reconfigure() : Unexpected existing context when "
-        "createRenderer==false", );
```
0mdc, line 202: I removed this assert as it is obsolete.
In this case, a context exists (GLFW) but the simulators are headless.

src/esp/sim/Simulator.cpp
```diff
@@ -890,12 +885,6 @@ double Simulator::getPhysicsTimeStep() {
 bool Simulator::recomputeNavMesh(nav::PathFinder& pathfinder,
                                  const nav::NavMeshSettings& navMeshSettings,
                                  const bool includeStaticObjects) {
-  ESP_CHECK(config_.createRenderer,
-            "::recomputeNavMesh: "
-            "SimulatorConfiguration::createRenderer is false. Scene "
-            "geometry is required to recompute navmesh. No geometry is "
-            "loaded without renderer initialization.");
-
```
0mdc, line 898: I removed this assert as it is obsolete.
`ResourceManager` already imports scene geometry to process physics whether or not the renderer is present.

mosra (approved):

examples/viewer.py
```diff
@@ -162,65 +179,70 @@ def __init__(self, sim_settings: Dict[str, Any]) -> None:
...
+            ):
+                self.navmesh_config_and_recompute(i)
 
         self.time_since_last_simulation = 0.0
         LoggingContext.reinitialize_from_env()
         logger.setLevel("INFO")
         self.print_help_text()
 
-    def draw_contact_debug(self):
+    def draw_contact_debug(self, env_index):
```
eundersander, line 200: Is there any way to avoid introducing `env_index` everywhere in viewer.py? E.g. I also see it above for navmesh stuff. I can almost guarantee that nobody will do contact debugging *while* using multiple envs. Using the python viewer with multiple envs is super-niche (basically only Mikael will use this) and so I hate to see `env_index` complicating all the code throughout this file.

As background, when I originally thought about integrating the batch renderer into viewer.py, I was really just thinking about using the batch renderer to render the scene, e.g. so we could troubleshoot visual differences between the renderers. I suspect this tiled viewer use-case is super-niche and something 95% of viewer.py users will never care about.

examples/viewer.py
```diff
@@ -263,15 +286,28 @@ def draw_event(
...
-        self.sim._Simulator__sensors[keys[0]][keys[1]].draw_observation()
-        agent = self.sim.get_agent(keys[0])
-        self.render_camera = agent.scene_node.node_sensor_suite.get(keys[1])
-        self.debug_draw()
-        self.render_camera.render_target.blit_rgba_to_default()
-        mn.gl.default_framebuffer.bind()
+        if self.enable_batch_renderer:
+            for i in range(self.num_env):
+                # Apply keyframe
+                keyframe = self.sim[i].gfx_replay_manager.extract_keyframe()
```
eundersander, line 292: This logic here with sensors/set_sensor_transform/etc could be encapsulated in a separate function to keep this draw function a bit more readable.

examples/viewer.py
```diff
@@ -333,26 +369,56 @@ def reconfigure_sim(self) -> None:
...
+        self.render_camera = []
+        for i in range(self.num_env):
+            self.default_agent.append(self.sim[i].get_agent(self.agent_id))
+            self.render_camera.append(
+                self.default_agent[i].scene_node.node_sensor_suite.get("color_sensor")
+            )
+
         # set sim_settings scene name as actual loaded scene
-        self.sim_settings["scene"] = self.sim.curr_scene_name
+        self.sim_settings["scene"] = self.sim[0].curr_scene_name
```
eundersander, line 402: In general I'd like to see more code in viewer.py like this that can just assume one sim. Maybe a helper, `self.get_sim()` or `self.default_sim`. Or maybe just keep `self.sim` and add `self.tiled_sims` for the niche tiled use-case.

Again, my high-level point is that 95% of viewer.py users don't need or care about multiple envs, and I'd like to keep most of the code in this file simpler.

examples/viewer.py
```diff
@@ -412,12 +478,18 @@ def key_press_event(self, event: Application.KeyEvent) -> None:
         elif key == pressed.H:
             self.print_help_text()
 
+        elif key == pressed.ONE:
```
eundersander, line 481: A bit of background, in `viewer.cpp`, each individual Habitat RE gradually added custom debugging functionality over time, to the point where that tool has a baffling huge set of hotkeys and is really not user-friendly. I would love to see us not make that mistake again.

Maybe my concrete suggestion here is to not add this hotkey. I suspect @0mdc is the only person that will ever use this functionality (you could add this code locally or in a branch).

examples/viewer.py
```diff
@@ -444,15 +516,15 @@ def key_press_event(self, event: Application.KeyEvent) -> None:
             )
 
         elif key == pressed.SPACE:
-            if not self.sim.config.sim_cfg.enable_physics:
+            if not self.sim[self.active_env_idx].config.sim_cfg.enable_physics:
```
eundersander, line 519: Again, let's aim to replace all this with a helper like `self.get_sim()` or `self.default_sim` or otherwise.

eundersander (commented):
Good progress here. I left some comments.

examples/viewer.py
0mdc: Makes sense. I'll clean that up.

src/esp/sim/SimulatorConfiguration.h
```diff
@@ -36,7 +36,13 @@ struct SimulatorConfiguration {
    */
   bool enablePhysics = false;
   /**
-   * @brief todo
+   * @brief Toggles batch rendering, which accelerates rendering of multiple
+   * environments in parallel.
+   */
+  bool enableBatchRenderer = false;
```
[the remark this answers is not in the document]
0mdc: Right now, it's only used by `simulator.py` [here](https://github.com/facebookresearch/habitat-sim/blob/ac814edda6b30511fe565a10863735519d57ad12/src_python/habitat_sim/simulator.py#L93). The `habitat-lab` default for `create_renderer` is `False`, so the `simulator.py` `create_renderer` override logic is expected by current simulator users.

I agree that batch rendering is not a property of the simulator. We could rename this flag `enableBatchRenderer` to something more explicit like `headless`.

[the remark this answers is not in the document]
eundersander: If it's only used in `simulator.py`, then declare it there or somewhere in Python. C++ `SimulatorConfiguration` is specifically for configuring the C++ `Simulator` class and this field creates confusion there.

PR #1970, opened 2022-12-20
Lighting support in the batch renderer

mosra:
## Motivation and Context

Things left to do:

- [x] Base implementation in the batch renderer
- [x] Distinguish between flat-shaded and shaded materials and render them appropriately
- [x] ~~Make it possible to draw shaded and flat meshes in a single draw call (currently the batch count is 2 instead of 1 in the test and it binds some additional texture, *huh?*)~~ it's fine, the second draw call is for the third untextured sphere
- [x] Tests for the base implementation
- [x] Tests for `clearLights()`
- [x] Properly implement the point vs directional light distinction
- [x] Check code coverage -- [all good](https://app.codecov.io/gh/facebookresearch/habitat-sim/blob/let-there-be-light/src/esp/gfx_batch/Renderer.cpp)
- [x] Enable and test per-scene light culling (i.e., the shader using not always the maximum light count, but the exact count each scene uses)
  - Note that this PR *won't* include per-draw light culling. Apart from this being out of scope of this initial PR, Habitat first needs to include a "range" value for all lights, which is essential as it contributes to the falloff. Without it the falloff is infinite.
- [x] Hook up with the gfx-replay pipeline
- [x] Hook up with `replayer`, enabled with an explicit `--max-light-count` option.
- [x] Adapt #1969 to include normals and properly mark flat-shaded vs PBR materials

How to enable this within the Python viewer, for @0mdc:

1. Pass an appropriate `maxLightCount` to the renderer configuration, as the default is 0, which flat-shades everything.
2. There's no step 2. Just ensure you're using correct composites and that an appropriate light setup is passed through.

## How Has This Been Tested

Look at those pretty pictures! :red_circle: :yellow_circle: 

yaaay

### DIFF
--- a/src/esp/gfx_batch/Renderer.cpp
+++ b/src/esp/gfx_batch/Renderer.cpp
@@ -58,6 +58,7 @@ struct RendererConfiguration::State {
   Mn::Vector2i tileSize{128, 128};
   Mn::Vector2i tileCount{1, 1};
   Mn::UnsignedInt maxLightCount{0};
+  Mn::Float ambientFactor{0.1f};
 };
 
 RendererConfiguration::RendererConfiguration() : state{Cr::InPlaceInit} {}
@@ -79,6 +80,12 @@ RendererConfiguration& RendererConfiguration::setTileSizeCount(
 RendererConfiguration& RendererConfiguration::setMaxLightCount(
     Mn::UnsignedInt count) {
   state->maxLightCount = count;
+  return *this;
+}
+
+RendererConfiguration& RendererConfiguration::setAmbientFactor(
+    Mn::Float factor) {
+  state->ambientFactor = factor;
   return *this;
 }
 
@@ -206,6 +213,7 @@ struct Renderer::State {
   RendererFlags flags;
   Mn::Vector2i tileSize, tileCount;
   Mn::UnsignedInt maxLightCount;
+  Mn::Float ambientFactor;
   /* Indexed with Mn::Shaders::PhongGL::Flag, but I don't want to bother with
      writing a hash function for EnumSet */
   // TODO have a dedicated shader for flat materials
@@ -266,6 +274,7 @@ void Renderer::create(const RendererConfiguration& configurationWrapper) {
   state_->tileSize = configuration.tileSize;
   state_->tileCount = configuration.tileCount;
   state_->maxLightCount = configuration.maxLightCount;
+  state_->ambientFactor = configuration.ambientFactor;
   const std::size_t sceneCount = configuration.tileCount.product();
   state_->projections = Cr::Containers::Array<ProjectionPadded>{sceneCount};
   state_->scenes = Cr::Containers::Array<Scene>{sceneCount};
@@ -576,7 +585,8 @@ bool Renderer::addFile(const Cr::Containers::StringView filename,
         const auto& phongMaterial =
             material->as<Mn::Trade::PhongMaterialData>();
         materials[i] = Mn::Shaders::PhongMaterialUniform{}
-                           .setAmbientColor(phongMaterial.diffuseColor() * 0.1f)
+                           .setAmbientColor(phongMaterial.diffuseColor() *
+                                            state_->ambientFactor)
                            .setDiffuseColor(phongMaterial.diffuseColor())
             /* Specular not used (and shader compiled with NoSpecular). Much
                plastic. Very fugly. Thus we also don't need shininess for
--- a/src/esp/gfx_batch/Renderer.h
+++ b/src/esp/gfx_batch/Renderer.h
@@ -91,6 +91,21 @@ struct RendererConfiguration {
    */
   RendererConfiguration& setMaxLightCount(Magnum::UnsignedInt count);
 
+  /**
+   * @brief Set ambient factor
+   *
+   * How much of the diffuse color / texture is used for ambient light in
+   * Phong-shaded materials. Default value is @cpp 0.1f @ce, setting it to
+   * @cpp 0.0f @ce will make the render completely black in areas without light
+   * contribution. In comparison, flat-shaded materials (i.e., imported with
+   * @ref Magnum::Trade::MaterialType::Flat) have the ambient factor always
+   * @cpp 1.0 @ce and are unaffected by lights.
+   *
+   * This value can currently only set upfront due to how materials are
+   * implemented internally.
+   */
+  RendererConfiguration& setAmbientFactor(Magnum::Float factor);
+
  private:
   friend Renderer;
   struct State;
--- a/src/esp/sim/BatchReplayRenderer.cpp
+++ b/src/esp/sim/BatchReplayRenderer.cpp
@@ -154,6 +154,10 @@ BatchReplayRenderer::BatchReplayRenderer(
         }
 
         renderer_.lightColors(sceneId_)[lightId] = light.color;
+        // TODO use gfx::getAmbientLightColor(lights) once it's not hardcoded
+        //  to an arbitrary value and once it's possible to change the ambient
+        //  factor in the renderer at runtime (and not just in
+        //  RendererConfiguration::setAmbientFactor())
         // TODO range, once Habitat has that
       }
     }


0mdc:
Good progress so far!

mosra:
This PR is now ready for a review. What's left is implementing the distinction between point vs directional lights, and updating the compositor scripts in #1947 to include normals and mark only the acually baked materials as flat.

A *horrible* screenshot for ~~your viewing pleasure~~ -- the overlapping objects are an issue with the input data, and the extremely contrasted lighting as well. I'll dive into this and try to create some more visually pleasing setup.

image

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -106,6 +108,30 @@ BatchReplayRenderer::BatchReplayRenderer(
           Mn::Matrix4::from(rotation.toMatrix(), translation);
     }
 
+    void changeLightSetup(const esp::gfx::LightSetup& lights) override {
+      !Mn::Debug{} << "Yay" << lights.size();
```
Skylion007, line 112: debug comment?

mosra: Yay! :D

mosra:
#1947 is now updated with lighting-aware composite creation and this PR is thus considered finished from my side. Output with the new composites for reference (above was using the loose glTF files) -- the main difference is that the Replica room model is now correctly using a flat-shaded material and thus isn't affected by lighting, making it look weird.

image

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -106,6 +108,49 @@ BatchReplayRenderer::BatchReplayRenderer(
...
+             about the Z axis direction, which is always the "target" vector
+             normalized. */
+          renderer_.transformations(sceneId_)[nodeId] =
+              Mn::Matrix4::lookAt({}, light.vector.xyz(), Mn::Vector3::yAxis());
+          lightId = renderer_.addLight(
+              sceneId_, nodeId, gfx_batch::RendererLightType::Directional);
+        }
+
+        renderer_.lightColors(sceneId_)[lightId] = light.color;
+        // TODO range, once Habitat has that
```
0mdc, line 150: 👍

0mdc (approved):
Looks good to me!

Skylion007 (approved):

src/esp/gfx_batch/Renderer.h
```diff
@@ -81,6 +81,14 @@ struct RendererConfiguration {
   RendererConfiguration& setTileSizeCount(const Magnum::Vector2i& tileSize,
                                           const Magnum::Vector2i& tileCount);
 
+  /**
+   * @brief Set max light count per draw
+   *
+   * By default no lights are used, i.e. flat-shaded rendering.
+   * @see @ref Renderer::maxLightCount()
+   */
+  RendererConfiguration& setMaxLightCount(Magnum::UnsignedInt count);
```
eundersander, line 90: How about a comment with some guidance for how to pick this number? E.g. Assuming a single draw can possibly be the entire scene, then the max count should probably be similar to the expected max light count across all scenes (is that even true?).

src/esp/gfx_batch/Renderer.cpp
```diff
@@ -522,17 +553,47 @@ bool Renderer::addFile(const Cr::Containers::StringView filename,
...
+          materialTextureTransformations[i] = {0, 0, {}};
+        else
+          materialTextureTransformations[i] = {
+              flatMaterial.texture() + textureOffset,
+              flatMaterial.textureLayer(), flatMaterial.textureMatrix()};
+      } else {
+        const auto& phongMaterial =
+            material->as<Mn::Trade::PhongMaterialData>();
+        materials[i] = Mn::Shaders::PhongMaterialUniform{}
+                           .setAmbientColor(phongMaterial.diffuseColor() * 0.1f)
```
eundersander, line 579: Ug, what is this 0.1? Are we hard-coding the amount of ambient light?

src/esp/gfx_batch/Renderer.cpp
```diff
@@ -522,17 +553,47 @@ bool Renderer::addFile(const Cr::Containers::StringView filename,
...
+                           .setDiffuseColor(phongMaterial.diffuseColor())
+            /* Specular not used (and shader compiled with NoSpecular). Much
```
eundersander, line 581: Can you explain this better? Is the problem that our existing assets have too much spec on their materials? Or is the problem that our point lighting doesn't produce nice-looking highlights?

src/esp/gfx_batch/Renderer.cpp
```diff
@@ -748,11 +809,17 @@ bool Renderer::addFile(const Cr::Containers::StringView filename,
...
+        Mn::Shaders::PhongGL::Flag::UniformBuffers |
+        Mn::Shaders::PhongGL::Flag::NoSpecular |
+        Mn::Shaders::PhongGL::Flag::LightCulling;
+    if (!(state_->flags >= RendererFlag::NoTextures)) {
       shaderFlags |= Mn::Shaders::PhongGL::Flag::AmbientTexture |
                      Mn::Shaders::PhongGL::Flag::TextureArrays |
                      Mn::Shaders::PhongGL::Flag::TextureTransformation;
+      /* Enable also a diffuse texture if we're rendering with lights */
+      if (state_->maxLightCount)
+        shaderFlags |= Mn::Shaders::PhongGL::Flag::DiffuseTexture;
```
eundersander, line 821: To sanity-check: a PBR material's baseColor texture would get loaded once and be set as both the ambient texture and diffuse texture for the corresponding Phong material, yes?

src/esp/gfx_batch/Renderer.cpp
```diff
@@ -1030,6 +1186,49 @@ void Renderer::draw(Mn::GL::AbstractFramebuffer& framebuffer) {
...
+
+    /* Finish transformation-dependent per-draw info, upload it */
+    for (std::size_t i = 0; i != scene.transformationIds.size(); ++i) {
+      scene
+          .drawsSorted[i]
+          /* Extract normal matrix */
+          .setNormalMatrix(state_->absoluteTransformationsSorted[i]
+                               .transformationMatrix.normalMatrix())
+          // TODO light culling should happen here
+          .setLightOffsetCount(0, scene.lights.size());
```
eundersander, line 1198: Where is maxLightCount enforced?

src/esp/gfx_batch/Renderer.h
```diff
@@ -538,26 +564,64 @@ class Renderer {
...
                                Corrade::Containers::StringView name);
 #endif
 
+  std::size_t addEmptyNode(Magnum::UnsignedInt sceneId);
+
+  /**
+   * @brief Add a light
+   * @param sceneId         Scene ID, expected to be less than
+   *    @ref sceneCount()
+   * @param nodeId          Node ID to inherit transformation from, returned
```
eundersander, line 587: Do we support the use case of a gltf mesh hierarchy with a light attached to some subnode? E.g. a lamp asset with a light attached, where the subnode offsets from the base of the lamp to the location of the bulb.

Er I'm guessing we do and it doesn't require an explicit `addLight`. I'm guessing adding that mesh hierarchy also adds the child light. Which is a whole can of worms we haven't really discussed (adding lights embedded in a gltf: the legacy renderer doesn't support this, right?).

src/esp/gfx_batch/Renderer.h
```diff
@@ -573,13 +637,47 @@ class Renderer {
...
   Corrade::Containers::StridedArrayView1D<Magnum::Matrix4> transformations(
       Magnum::UnsignedInt sceneId);
 
+  /**
+   * @brief Colors of all lights in the scene
+   * @param sceneId   Scene ID, expected to be less than @ref sceneCount()
+   *
+   * Returns a view on colors of all lights in given scene. Desired usage is to
+   * update only colors at indices returned by @ref addLight(), updating colors
+   * at other indices is possible but could have unintended consequences.
```
eundersander, line 656: Just curious, how do we end up with lights in here other than via `addLight`?

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -106,6 +108,49 @@ BatchReplayRenderer::BatchReplayRenderer(
...
+      }
+
+      renderer_.clearLights(sceneId_);
+      for (std::size_t i = 0; i != lights.size(); ++i) {
+        const gfx::LightInfo& light = lights[i];
+        CORRADE_INTERNAL_ASSERT(light.model == gfx::LightPositionModel::Global);
+
+        const std::size_t nodeId = renderer_.addEmptyNode(sceneId_);
+
+        /* Clang Tidy, you're stupid, why do you say that "lightId" is not
```
eundersander, line 126: >an accidentally zero-initialized variable is *also* a bug, you know?

wut

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -106,6 +108,49 @@ BatchReplayRenderer::BatchReplayRenderer(
...
+           range asserts in all functions so your "suggestions" are completely
+           unhelpful. */
+        std::size_t lightId;  // NOLINT
+        if (light.vector.z()) {
+          renderer_.transformations(sceneId_)[nodeId] =
+              Mn::Matrix4::translation(light.vector.xyz());
+          lightId = renderer_.addLight(sceneId_, nodeId,
+                                       gfx_batch::RendererLightType::Point);
+        } else {
+          /* The matrix will be partially NaNs if the light vector is in the
```
eundersander, line 139: Well that sounds pretty shady. You're really okay with NaNs floating around here? What if one day we want to assert that all transformations are orthonormal?

In past codebases, I used to have a "getArbitraryOrthogonalVector" and/or "getArbitraryLookatTransform" for this kinda thing.

eundersander (commented):
Good progress. My comments are mostly just aimed at gaining understanding.

src/esp/gfx_batch/Renderer.h
mosra: It's the max number of lights applied per draw.

The semantic will change slightly once light culling is in (for which it's essential that the lights contain range info, which they do in glTF files but not yet in Habitat-supplied lights). With light culling, this will be an upper bound of lights used per draw (picking the brightest from the set, ignoring the rest) and the number of lights in total will be handled internally, without any need for user-specified settings. 

Currently without the culling it's both the max number of lights applied per draw and the max number in total, because that's the same thing. I'll update the docs once

src/esp/gfx_batch/Renderer.cpp
mosra: Yes -- see [this Slack post for details](https://cvmlp.slack.com/archives/G0131KVLBLL/p1671575565719199?thread_ts=1671497345.493199&cid=G0131KVLBLL).

I wanted to achieve an acceptable look until there's a proper PBR pipeline where one doesn't need to hand-tune this kind of thing anymore. Same for specular highlights. The materials are all encoded as PBR, which means there's no info about "classic" specular highlights or ambient anymore.

src/esp/gfx_batch/Renderer.cpp
mosra: Same as [in case of the ambient](https://github.com/facebookresearch/habitat-sim/pull/1970#discussion_r1063620089), the PBR materials in glTF don't encode any info about "classic" specular so the renderer has to guesstimate and it usually guesses wrong. From my experience the specular "white spots" were what usually people complained about when seeing the renderer output (such as [the recent Floorplanner discussion on Slack](https://cvmlp.slack.com/archives/CFN5TAUSD/p1671556128596599)).

src/esp/gfx_batch/Renderer.cpp
mosra: Yep.

src/esp/gfx_batch/Renderer.cpp
mosra: On line 831, in `setLightCount()` passed to [Shaders::PhongGL::Configuration](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1PhongGL_1_1Configuration.html) -- it's a shader-compile-time setting, not a uniform input.

[As I said above](https://github.com/facebookresearch/habitat-sim/pull/1970#discussion_r1063616162), this will change slightly when light culling is done -- the compile-time light count will be handled internally, and the max per-draw light count will be limiting the count of lights used per draw, picking just the ones with most influence.

src/esp/sim/BatchReplayRenderer.cpp
Skylion007: @mosra Just add NOLINTNEXTLINE comment then. :)

src/esp/gfx_batch/Renderer.h
mosra: > Do we support the use case of a gltf mesh hierarchy with a light attached to some subnode?

Yes, although not everything is in place yet, that's what I designed for, to make it easier to author scenes with uniform lighting, as it's significantly easier to tune light intensities/ranges in Blender and export a self-contained glTF than supplying them externally via a JSON config, or having to explicitly add also a light every time you add a lamp to the scene.

I want to treat those as being a part of the asset (and thus not something to be explicitly supplied by the simulator apart from a high-level use like "hold a torch" or "add a lamp there"), same as e.g. emissive textures are. And eventually the datasets will contain also lightprobe locations and other light-related data needed by PBR, which would otherwise have to be proxied through Habitat's data pipeline.

> adding lights embedded in a gltf: the legacy renderer doesn't support this, right?

I was pushing for loading lights from the glTF several times (because, again, it's *significantly* easier to author the assets that way), all the scaffolding is in place, but probably nobody did the work to hook it up within habitat. ¯\\\_(ツ)_/¯

src/esp/gfx_batch/Renderer.h
mosra: From lights being implicitly added via `addHierarchy()`, [as discussed above](https://github.com/facebookresearch/habitat-sim/pull/1970#discussion_r1063643063).

src/esp/sim/BatchReplayRenderer.cpp
mosra: Because if the variable was meant to be initialized to something, but by accident wasn't, the zero-initialization would hide that bug, causing the program to behave seemingly fine but incorrect. While not having it initialized at all (kept as a random memory) could possibly lead to a random OOB assertion down the line, or cause the program to behave randomly, making it easier to discover the bug.

At least in my experience it was always like this, bugs caused by random memory were easier to discover than bugs where the variables were zero-initialized by accident.

src/esp/sim/BatchReplayRenderer.cpp
mosra: I'm okay with that, yes. I view it as the NaNs signifying that the X/Y basis is unimportant and only Z matters, and it works as expected even in presence of matrix transformations -- NaNs stay where they should stay:

```py
>>> Matrix4.rotation_x(Deg(30.0))@\
    Matrix4.look_at(Vector3(), Vector3.y_axis(), Vector3.y_axis())
Matrix(-nan, -nan, 0, 0,
       -nan, -nan, -0.866025, 0,
       -nan, -nan, -0.5, 0,
       -nan, -nan, 0, 1)
```

If it eventually causes asserts for whatever reason, I can fix those, but even orthonormal transforms are sometimes impossible [such as in this case of meshes flattened along a certain axis](https://github.com/facebookresearch/habitat-sim/pull/1868). And inventing an always-orthogonal transformation just to avoid NaNs would only mean having extra code to test and it'd be all discarded in the end anyway.

src/esp/gfx_batch/Renderer.cpp
eundersander: How about re-using `LightSetup.h getAmbientLightColor`? If we matched that value, would it be easier to get visual parity with the classic renderer?

src/esp/gfx_batch/Renderer.h
eundersander: To be clear: is this fully hooked up? If we add a hierarchy that includes a light, will the light get added?

src/esp/sim/BatchReplayRenderer.cpp
eundersander: >cause the program to behave randomly, making it easier to discover the bug

Not my experience. Nondeterministic bugs are scary. E.g. by leaving it uninitialized, it gets initialized to zero in debug builds but something random and nonzero in release builds, so the entire EA Sports Nascar 2005 eng team spends the entire afternoon tracking down why only the release build is broken. :)

src/esp/sim/BatchReplayRenderer.cpp
eundersander: Some bold ideas from Mosra today, haha. Another counterpoint: what if someone tries to debug-render all the node coordinate frames as red/green/blue axes or whatever? Now you have NaNs in vertices you're submitting to the driver.

src/esp/gfx_batch/Renderer.cpp
mosra: I could look into that, yup -- I remember you spent a considerable amount of time on the PBR -> Phong conversion, so it would be a shame to ignore it.

How should I deal with the original requirement that `gfx_batch` stays completely isolated from the rest of Habitat? (I was considering putting some PBR -> Phong conversion directly into Magnum, but that time would probably be better spent on actually making PBR work.)

But, re visual parity -- the recent Floorplanner discussion made me think I should maybe try something from scratch instead of trying to match how it looks in the classic renderer :sweat_smile:

src/esp/gfx_batch/Renderer.h
mosra: Not yet -- I didn't find time to actually equip datasets with lights to have something to test with and didn't want to add completely uncovered code to this PR, so it's not here here. The "compositor" in #1969 and the glTF exporter also don't handle those for a similar reason, but both are relatively trivial pieces of code.

Once I have data to test with, I will be definitely bugging everyone on Slack with this feature ;)

src/esp/gfx_batch/Renderer.h
mosra: > lights come from the JSON metadata. The new batch renderer should work the same way

It does, it consumes lights from the JSON metadata, via the gfx-replay :) But I see what you mean -- when I get to hooking this up, I'll add a flag to skip lights coming from the input file, similarly to how there's a flag to skip texture.

src/esp/gfx_batch/Renderer.cpp
eundersander: >the original requirement that gfx_batch stays completely isolated from the rest of Habitat?

Yep. So this would be some caller like replayer.cpp passing `LightSetup.h getAmbientLightColor` to the batch renderer, maybe via a new public method or the Configuration struct.

>the recent Floorplanner discussion made me think I should maybe try something from scratch instead of trying to match how it looks in the classic renderer
I don't have the full context here, but I think it'll be valuable going forward to be able to swap between the classic and batch renderer, with Mikael working to reach parity.

src/esp/sim/BatchReplayRenderer.cpp
mosra: Counterpoint to a counterpoint: GCC 12 has [`-ftrivial-auto-var-init=pattern`](https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Optimize-Options.html#index-ftrivial-auto-var-init) which initializes local variables in debug builds to some bit pattern. MemorySanitizer can do something similar for the heap as well, and maybe for local variables too.

So if the variable is accidentally left uninitialized, these options can be used to make the bug behave deterministically. But if it gets zero-initialized because Clang Tidy demanded to, and then accidentally not set to the value it should have, it's harder to discover, as it would behave as a "regular" value and won't trigger any assertions or sanitizer reports. In my opinion. But my general opinion is fail fast & fail hard, which is probably something completely undesirable in game development :)

src/esp/sim/BatchReplayRenderer.cpp
mosra: That's also fine, it'll show the blue axis only (and yes, I'm doing exactly that regularly) ;)

src/esp/gfx_batch/Renderer.cpp
mosra: Oh, wait, this is about the *light* ambient color, not the per-material ambient color, I missed that.

So, just to clarify we're on the same page -- the aspect of diffuse / base color used partially for ambient is fine in this case, and you're only concerned about the factor being hardcoded, right? Not about the fact that the same texture is used for both.

Having it configurable from the light makes sense, but it'll be a global factor (i.e., there's no way to have different lights with different ambient factors affecting different meshes, the ambient factors will get all summed up and used as a single value). But that should be fine, I think?

src/esp/sim/BatchReplayRenderer.cpp
mosra: Hmm, that's right -- a mesh transformed by this matrix would be all NaNs. A visualizer using the base axes directly wouldn't tho, it would show just the blue axis, which is maybe what is meant to be visualized?

I'm taking it too far here, am I :D

src/esp/sim/BatchReplayRenderer.cpp
eundersander: If this were my code, I'd still prefer to initialize to an aggressively-invalid value like -1 or something, but I don't feel too strongly and you've clearly given this some thought. Fine to leave this as-is!

src/esp/sim/BatchReplayRenderer.cpp
eundersander: >it would show just the blue axis, which is maybe what is meant to be visualized?

Ohhhh, that's going to be a mystery rabbit hole for whatever poor junior SWE attempts this visualization.

Speaking of rabbit holes, I don't like dragging out these little code review nits, and I don't want to get into a mode of demanding changes, so I'll ask you to make the final decision here!

src/esp/gfx_batch/Renderer.cpp
eundersander: >you're only concerned about the factor being hardcoded, right?
Yes.

>Not about the fact that the same texture is used for both.
No problem here. I expect this.

Somewhat related, I know we've disagreed in this past a bit on this nuance of the Phong lighting model. My preference is for a material's ambient color to generally match its diffuse color, and for that material ambient color to get multiplied with an ambient light color at runtime (every frame), such that you could hypothetically vary the color/amount of ambient light in the scene every frame. Whereas I know you've spoke in the past of preferring to pre-multiply these two and basically bake them into the material's ambient color. I'm too lazy to track down that Slack thread from a year ago, but more specifically, you said that many 3D assets in the wild already do this pre-baking by convention. Anyway, we don't have to debate this now or change the current code in this respect.

eundersander: >Having it configurable from the light makes sense, but it'll be a global factor 
Exactly, a global factor, not per light. I think of ambient light color as sorta like the cosmic background radiation. It's a constant amount of light shining in all directions, everywhere in the scene. A lazy proxy for e.g. env-map-based light.

eundersander (approved):
Not sure if you have more changes in progress related to my recent comments, but I feel like our threads are resolved at least philosophically, so I'm giving the green check!

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -106,6 +108,49 @@ BatchReplayRenderer::BatchReplayRenderer(
...
+        const std::size_t nodeId = renderer_.addEmptyNode(sceneId_);
+
+        /* Clang Tidy, you're stupid, why do you say that "lightId" is not
+           initialized?! I'm initializing it right in the branches below, I
+           won't zero-init it just to "prevent bugs" because an accidentally
+           zero-initialized variable is *also* a bug, you know? Plus I have
+           range asserts in all functions so your "suggestions" are completely
+           unhelpful. */
+        std::size_t lightId;  // NOLINT
+        if (light.vector.z()) {
```
0mdc, line 133: We want to look at the `w` component of the light vector to determine its type.

Skylion007: @0mdc I am not following, is this flagging a bug as it is currently written?

mosra: Oh, whoops -- yep, that's a bug. Fixing...

src/esp/sim/BatchReplayRenderer.cpp
mosra: I added a TODO -- once I have a convenience utility for this in Magnum, I change to it instead of `lookAt()`.

src/esp/sim/BatchReplayRenderer.cpp
mosra: Fixed in 3906aaad2199121ba8d2928ab025828c0d6b3170.

PR #1975, opened 2022-12-22
gfx-replay: Record scene graph deletions.

0mdc:
## Motivation and Context

When the simulator changes scene, the previous scene graph is not deleted. Because `Recorder` relies on scene node deletion to issue deletion entries to gfx-replay, the previous scene is not cleared from the replay renderer.

This changeset adds a method to delete all gfx-replay nodes within a scene graph so that the replay renderer can correctly render process scene changes.

## How Has This Been Tested

Tested through Python viewer.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -1428,7 +1428,7 @@ scene::SceneNode* ResourceManager::createSemanticRenderAssetInstance(
     scene::SceneNode* parent,
     DrawableGroup* drawables) {
   if (creation.isTextureBasedSemantic()) {
-    // Treat texture-based semantic meshes as General/Primitves.
+    // Treat texture-based semantic meshes as General/Primitives.
     return createRenderAssetInstanceGeneralPrimitive(creation, parent,
                                                      drawables, nullptr);
   }
@@ -1820,7 +1820,7 @@ bool ResourceManager::loadRenderAssetGeneral(const AssetInfo& info) {
 
     // If meshIDLocal != -1 then we have multiple meshes assigned to the same
     // MeshTransformNode.  We make subsequent meshes children of the first mesh
-    // we've seen, and give them identity trasnforms.
+    // we've seen, and give them identity transforms.
     // TODO: either drop MeshTransformNode in favor of SceneData or use
     // Mn::SceneTools::convertToSingleFunctionObjects() when it's exposed.
     esp::assets::MeshTransformNode* tmpNode = &*node;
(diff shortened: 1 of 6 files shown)


src/esp/gfx/replay/Player.cpp
```diff
@@ -161,6 +161,18 @@ void Player::applyKeyframe(const Keyframe& keyframe) {
     assetInfos_[assetInfo.filepath] = assetInfo;
   }
 
+  for (const auto& deletionInstanceKey : keyframe.deletions) {
```
eundersander, line 164: What if we create an instance and then delete it in the same frame? Wouldn't this "leak" the instance during replay? I suspect this scenario was the reasoning behind my initial ordering of these operations inside `applyKeyframe`.

src/esp/gfx/replay/Recorder.h
```diff
@@ -57,6 +58,15 @@ class Recorder {
...
   void onLoadRenderAsset(const esp::assets::AssetInfo& assetInfo);
 
+  /**
+   * @brief Record deletion of all render instances in a scene graph.
+   * Because scene graphs are currently leaked when the active scene changes, we
+   * cannot rely on node deletion to issue gfx-replay deletion entries. This
+   * function allows to circumvent this issue.
+   * @param sceneGraph The scene graph to delete.
+   */
+  void onDeleteSceneGraph(const esp::scene::SceneGraph& sceneGraph);
```
eundersander, line 68: This callback name is confusing because we're not actually deleting a scene graph in the caller code (right? If we were, we wouldn't need this). I suggest a rename to `onHideSceneGraph`, `onDisableSceneGraph`, or even `onLeakSceneGraph` to capture what's really happening in Simulator.cpp. You should also revise the comment here to point people to the specific file/line in Simulator.cpp where the scene graph is leaking. Also, the description `The scene graph to delete` should be updated to something like `the scene graph being hidden/disabled/leaked`.

eundersander (commented):
Left some comments.

src/esp/gfx/replay/Player.cpp
0mdc: You're right, this scenario would cause an issue.

src/esp/assets/ResourceManager.cpp
```diff
@@ -1428,7 +1428,7 @@ scene::SceneNode* ResourceManager::createSemanticRenderAssetInstance(
     scene::SceneNode* parent,
     DrawableGroup* drawables) {
   if (creation.isTextureBasedSemantic()) {
-    // Treat texture-based semantic meshes as General/Primitves.
+    // Treat texture-based semantic meshes as General/Primitives.
```
eundersander, line 1431: Nit: I would argue the grammar was better before. General/Primitive is a category.

src/esp/scene/SceneManager.h
```diff
@@ -30,6 +27,9 @@ class SceneManager {
   SceneGraph& getSceneGraph(int sceneID);
   const SceneGraph& getSceneGraph(int sceneID) const;
 
+  // Returns the count of registered scene graphs
+  inline int getSceneGraphCount() { return sceneGraphs_.size(); }
```
Skylion007, line 31: Wow, implicit loss of percision here:
```suggestion
  inline size_t getSceneGraphCount() { return sceneGraphs_.size(); }
```

PR #1976, opened 2022-12-23
Clear replay renderer when all instances are deleted.

0mdc:
## Motivation and Context

The batch renderer currently cannot delete individual instances. Because of this, it quickly reaches internal limits when repeatedly changing scenes. The batch renderer can however delete all instances at once.

This changeset clears the frame when the player detects that the applied keyframe is deleting all instances.

This implementation is preferred to an explicit `clearFrame` keyframe flag as it doesn't force the implementer (blender script, etc) to delete its instances in a specific way. However, it cannot detect leaking instances from a misused recorder.

## How Has This Been Tested

Tested in python viewer by repeatedly changing scenes.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/replay/Player.cpp
+++ b/src/esp/gfx/replay/Player.cpp
@@ -154,11 +154,21 @@ void Player::clearFrame() {
 
 void Player::applyKeyframe(const Keyframe& keyframe) {
   for (const auto& assetInfo : keyframe.loads) {
-    CORRADE_INTERNAL_ASSERT(assetInfos_.count(assetInfo.filepath) == 0);
     if (failedFilepaths_.count(assetInfo.filepath) != 0u) {
       continue;
     }
+    // TODO: This overwrites the previous AssetInfo. This is not ideal. Consider
+    // including AssetInfo in creations rather than using keyframe loads.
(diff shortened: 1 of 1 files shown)


src/esp/gfx/replay/Player.cpp
```diff
@@ -188,16 +198,18 @@ void Player::applyKeyframe(const Keyframe& keyframe) {
...
   }
 
-  for (const auto& deletionInstanceKey : keyframe.deletions) {
-    const auto& it = createdInstances_.find(deletionInstanceKey);
-    if (it == createdInstances_.end()) {
-      // missing instance for this key, probably due to a failed instance
-      // creation
-      continue;
-    }
+  if (!frameCleared) {
```
eundersander, line 201: It might be more readable to clear `deletions` above and then leave this part of the function untouched.

eundersander (approved):
Left a minor comment but otherwise LGTM!

src/esp/gfx/replay/Player.cpp
0mdc: The keyframe is a const ref, and should remain so because of the function usage. Given this, I think that it is clearer to use a condition.

Skylion007 (approved):

PR #1984, opened 2023-01-14
Updated semantic annotations in habitat-lab

rpartsey:
## Motivation and Context

I'm not sure if loading semantic annotations in habitat-lab should be part of habitat-sim README.md, but if so, then it has to be updated to Hydra style.

## Types of changes

- [x ] Docs change / refactoring / dependency upgrade

### DIFF
--- a/DATASETS.md
+++ b/DATASETS.md
@@ -112,18 +112,39 @@ habitat-viewer --dataset '<PATH TO HM3D>/hm3d_annotated_basis.scene_dataset_conf
 
 To load semantic annotations in habitat-lab:
 
-* Add the semantic sensor to the list of agent sensors:
+* Update your config by adding the semantic sensor to the list of agent sensors:
 
-    ```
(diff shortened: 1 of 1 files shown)


DATASETS.md
```diff
@@ -112,18 +112,39 @@ habitat-viewer --dataset '<PATH TO HM3D>/hm3d_annotated_basis.scene_dataset_conf
...
+      # add semantic sensor:
+      - /habitat/simulator/sim_sensors@habitat.simulator.agents.<agent_name>.sim_sensors.semantic_sensor: semantic_sensor
+      - /habitat/task/rearrange: <task>
+      - /habitat/dataset/rearrangement: <dataset>
+      - _self_
+    
+    habitat:
+      simulator:
+        agents:
+          <agent_name>:
```
Skylion007, line 132: NIT: it's not quite clear that this is an variable when listed as key like this. Is there a sensible defualt we can provide them. Or at least add comments to those things which need to be changed.

rpartsey: Thank you @Skylion007!
How about now?

PR #1990, opened 2023-01-26
[WIP] - Add viewer.py to Habitat-sim module

aclegg3:
## Motivation and Context

Currently, the interactive sim viewer (`example/viewer.py`) lives in the `examples/` directory is therefore not exposed within the `habitat_sim` python module. As such, it cannot be used with Habitat-sim conda build or imported and extended in Habitat-lab.

This PR copies the viewer code into the `src_python` directory and makes minor edits for code style and flaking requirements.

See [Habitat-lab PR 1099](https://github.com/facebookresearch/habitat-lab/pull/1099) for example use as external module to extend. 

TODO:
Once this is ready for merge we'll update documentation and possibly deprecate the examples/viewer.py.

## How Has This Been Tested

With Habitat-sim module installed,

1.  run the new viewer from its python file locally

```
python src_python/habitat_sim/viewer/viewer.py --dataset data/scene_datasets/mp3d_example/mp3d.scene_dataset_config.json --scene 17DRP5sb8fy
```

2.  run the new viewer from the module:
 ```
python -m habitat_sim.viewer.viewer --dataset data/scene_datasets/mp3d_example/mp3d.scene_dataset_config.json --scene 17DRP5sb8fy
```

-  Issue: module run is currently failing with:
```
Platform::WindowlessEglContext: cannot make context current: EGL_BAD_ACCESS
Platform::WindowlessEglApplication::makeCurrent(): cannot make context current: EGL_BAD_ACCESS
Failed to make platform current

```
Likely related: we require the following to avoid this issue with the python file run:
```
flags = sys.getdlopenflags()
sys.setdlopenflags(flags | ctypes.RTLD_GLOBAL)
```
possibly the module run is picking things up differently.

## Types of changes

## Checklist

### DIFF
--- a/src_python/habitat_sim/viewer/viewer.py
+++ b/src_python/habitat_sim/viewer/viewer.py
@@ -1,0 +1,1170 @@
+# Copyright (c) Meta Platforms, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import ctypes
+import math
+import os
+import string
+import sys
+import time
+from enum import Enum
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+import git
+
+flags = sys.getdlopenflags()
+sys.setdlopenflags(flags | ctypes.RTLD_GLOBAL)
+
+import magnum as mn
+import numpy as np
+from magnum import shaders, text
+from magnum.platform.glfw import Application
+
+import habitat_sim
+from habitat_sim import ReplayRenderer, ReplayRendererConfiguration, physics
+from habitat_sim.logging import LoggingContext, logger
+from habitat_sim.utils.common import quat_from_angle_axis
+from habitat_sim.utils.settings import default_sim_settings, make_cfg
+
+repo = git.Repo(".", search_parent_directories=True)
+dir_path = repo.working_tree_dir
+# %cd $dir_path
+data_path = os.path.join(dir_path, "data")
+
+
+class HabitatSimInteractiveViewer(Application):
+
+    # the maximum number of chars displayable in the app window
+    # using the magnum text module. These chars are used to
+    # display the CPU/GPU usage data
+    MAX_DISPLAY_TEXT_CHARS = 256
+
+    # how much to displace window text relative to the center of the
+    # app window (e.g if you want the display text in the top left of
+    # the app window, you will displace the text
+    # window width * -TEXT_DELTA_FROM_CENTER in the x axis and
+    # window height * TEXT_DELTA_FROM_CENTER in the y axis, as the text
+    # position defaults to the middle of the app window)
+    TEXT_DELTA_FROM_CENTER = 0.49
+
+    # font size of the magnum in-window display text that displays
+    # CPU and GPU usage info
+    DISPLAY_FONT_SIZE = 16.0
+
+    def __init__(self, sim_settings: Dict[str, Any]) -> None:
+        self.sim_settings: Dict[str, Any] = sim_settings
+
+        self.enable_batch_renderer: bool = self.sim_settings["enable_batch_renderer"]
+        self.num_env: int = (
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by aclegg3** [comment]
@mosra thoughts on the EGL issue with the running from the module?

File: src_python/habitat_sim/viewer/viewer.py [hunk 1]
```diff
@@ -0,0 +1,1170 @@
...
+            key.X: "move_down",
+            key.Z: "move_up",
+        }
+
+        # Load a TrueTypeFont plugin and open the font file
+        self.display_font = text.FontManager().load_and_instantiate("TrueTypeFont")
+        self.display_font.open_file(
+            os.path.join(os.path.dirname(__file__), self.sim_settings["font_path"]),
+            13,
+        )
```
aclegg3, line 124: @mosra had to replace the relative filepath with a configurable one which can be updated in consumer code. Generally we can't expect a `data` folder to be available from conda build, so we should consider how to handle this. 

Maybe we should include the font file with the python module source code instead?

jaraujo98:
I'm a strong supporter of this change! A couple of thoughts (from when I was thinking about this a few months ago):

- Several of the current keybindings make sense for an example viewer, but might not be interesting for every viewer application, and it is helpful to have those keys free for different functionality. From that perspective, I think it could make sense to keep keybindings to a minimum in the base viewer class and add the remaining ones in an `ExampleViewer` subclass implemented in the `examples/viewer.py` file.
- The same thing goes for the command line arguments. This makes sense for the example viewer, but maybe not for a base class definition inside `src_python`.

aclegg3:
> keep keybindings to a minimum in the base viewer class and add the remaining ones in an `ExampleViewer` subclass implemented in the `examples/viewer.py` file... The same thing goes for the command line arguments...

Thanks, I agree. I'll start a design doc to help plan which UI features should be core vs. example. :+1:

src_python/habitat_sim/viewer/viewer.py
0mdc: > Maybe we should include the font file with the python module source code instead?

I think that this would make sense.

mosra: Yep, that makes sense -- when John did that originally we weren't sure what the use case would be, and hardcoding the relative path to `data/` was the most straightforward way.

mosra:
> `sys.setdlopenflags(flags | ctypes.RTLD_GLOBAL)`

A bit of background, just to have the full picture here -- by default, Python loads every module in its own namespace (`RTLD_LOCAL`), making it impossible for a different module to see its globally exported symbols. This makes sense for security reasons. `RTLD_GLOBAL` disables that.

The other half of the problem that Habitat uses Magnum compiled statically, and while I eliminated most mutable globals, Magnum has still three left -- for the plugin manager, for logging redirection and for the GL context. Those, with Magnum being compiled statically, get one copy in the `habitat_sim.so` Python module and one in the `magnum.so` Python module, and then I have to do cursed things to make both modules refer to the same globals so operation done in one module is visible by the other. And those cursed things then only work with `RTLD_GLOBAL`. Without that, being each loaded in its own namespace, the modules would need to share a common address of the globals in some manual way, possibly by going through Python, and so far I didn't figure out a nice way to do this that wouldn't become a new problem on its own.

A *real* solution to all these problems would be to have Magnum compiled dynamically, but that was historically not an option as far as I remember. Maybe we could reconsider, it'd definitely fix 98 of 100 recurring nightmares I have :D

The EGL error is most probably related to the above -- `EGL_BAD_MATCH` is emitted [if the context is already active in another thread](https://registry.khronos.org/EGL/sdk/docs/man/html/eglMakeCurrent.xhtml), so it's possible that due to the global GL context not being shared properly between the two modules it was thinking the context wasn't active in another thread (while it actually was, for example in the background renderer if that's enabled in the build) and tried to use it.

PR #2001, opened 2023-02-08
Better Camera unproject

aclegg3:
## Motivation and Context

Previously camera unproject function would return a normalized ray which was useful for distance calculations and raycasting, but could not be used directly with depth image values (on view plane) for depth unprojection into point clouds.

This PR adds two features:

1. Added `normalized` variable to unproject (default true)
2. When `normalized==False`, unproject now returns a correctly scaled ray direction in world units to get a precise 3D location from a depth pixel value.

## How Has This Been Tested

Modified unit test to validate `normalized` variable.
New unit tests demonstrate depth unprojection and leverage Bullet collision detection to validate the feature.

Example images from optional visual test output showing a sphere object for a random depth unprojected pixel:
tmpabvkkg6z
tmpftysrtyg

## Types of changes

## Checklist


src/esp/gfx/RenderCamera.cpp
```diff
@@ -212,8 +213,11 @@ esp::geo::Ray RenderCamera::unproject(const Mn::Vector2i& viewportPosition) {
   ray.direction =
       ((object().absoluteTransformationMatrix() * projectionMatrix().inverted())
            .transformPoint(normalizedPos) -
-       ray.origin)
-          .normalized();
+       ray.origin) /
+      1000.0;
```
jturner65, line 217: Why 1000?  Should this be a defined constant somewhere?

jturner65: Considering the intent of this change, perhaps some documentation regarding rescaling the normalized camera ray to fit the desired use would be more appropriate?

aclegg3: Great question, this is the far plane distance.
I updated the code and docs such that the scaling is on the user side.
Also added a test to validate this.

Skylion007 (approved):

src/esp/gfx/RenderCamera.cpp
jturner65: how about we get the far distance by doing this right there? @mosra does this look appropriate?

```
auto zFar = Frustum::fromMatrix(projectionMatrix()*cameraMatrix()).far();
auto dist = zFar(3)/zFar.xyz().length();

```

jturner65 (approved):
LGTM, except for pesky pylint :) .

src/esp/gfx/RenderCamera.cpp
mosra: The `Frustum` class is unnecessary overhead -- it calculates also all other planes and you need just one of them, and only its distance.

The "most obvious" way to calculate the distance of the far plane is by doing an inverse transform of a point lying on the far plane (i.e., calculating what point gets projected to `(0, 0, 1)`). Something like this:

```cpp
projectionMatrix().inverted().transformPoint(Vector3::zAxis(1.0f)).z()
```

But that's imprecise and not fast either due to the inversion. After looking at the [Matrix4::perspectiveProjection()](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Matrix4.html#a6475b3ef155c9142b890c8133504ae9b) docs and a bit of math suffering I think the equation is like this:

```cpp
near = projection[3][2]/(projection[2][2] - 1.0f);
far = projection[3][2]/(1.0f + projection[2][2]);
```

I could add those directly on the Matrix4 class I think, just need to do a bit of testing and correctly handle infinite far plane.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -209,11 +210,20 @@ esp::geo::Ray RenderCamera::unproject(const Mn::Vector2i& viewportPosition) {
           Magnum::Vector2{1.0f},
       1.0};
 
+  // compute the far plane
+  auto zFar =
+      Mn::Frustum::fromMatrix(projectionMatrix() * cameraMatrix()).far();
+  float farDistance = zFar[3] / zFar.xyz().length();
+
   ray.direction =
       ((object().absoluteTransformationMatrix() * projectionMatrix().inverted())
```
mosra, line 219: (Ignore if this utility isn't meant to be used in performance-sensitive scenarios.)

The inverted projection matrix could maybe get cached, otherwise it'll be the most time consuming calculation of the whole function.

src/esp/gfx/RenderCamera.cpp
```diff
@@ -209,11 +210,20 @@ esp::geo::Ray RenderCamera::unproject(const Mn::Vector2i& viewportPosition) {
...
            .transformPoint(normalizedPos) -
-       ray.origin)
-          .normalized();
+       ray.origin) /
+      farDistance;
+
+  if (normalized) {
+    ray.direction = ray.direction.normalized();
+  }
```
mosra, line 226: (Again ignore if this utility isn't meant to be used in performance-sensitive scenarios.)

A better way would be to have this split into two functions, one `unproject()` that doesn't normalize, and an `unprojectNormalized()` that calls `unproject()`, normalizes its result and returns it. That way there's no branch. (Though that way it'll be a breaking change, because `unproject()` will not normalize anymore, so you might want to invent different naming of the two overloads.)

src/esp/gfx/RenderCamera.cpp
jturner65: Thanks @mosra ! Do we not need the product of the projection & camera transforms for the far plane calculation?

mosra: Yes -- the equations I have give you the distance to where far Z plane is relative to the camera. So then if you transform the camera somewhere, you don't need to do the complicated extraction of the far plane again, only transforming that "far vector" with the camera matrix. I think. My math is rusty.

```cpp
// Negative Z is forward
Vector3 cameraRelativeZIsHere = cameraMatrix().transformPoint(Vector3::zAxis(-far));
```

src/esp/gfx/RenderCamera.cpp
aclegg3: Good point, caching makes sense.

src/esp/gfx/RenderCamera.cpp
aclegg3: Yeah, thanks. I thought about this also, but didn't want a breaking change or to add more bulk to the API. I think this one can stay for now.

src_python/habitat_sim/utils/settings.py
```diff
@@ -30,6 +31,7 @@
     "equirect_semantic_sensor": False,
     "seed": 1,
     "physics_config_file": "data/default.physics_config.json",
+    "enable_physics": True,
```
0mdc, line 34: What are the consequences of enabling physics by default here? Do we want to do this?

tests/test_gfx.py
```diff
@@ -37,22 +47,101 @@ def test_unproject():
...
+                mn.Rad(random.random()), mn.Vector3(1, 0, 0)
+            )
+            depth_camera.node.rotation = render_camera.node.rotation
+
+            # do the unprojection from depth image
+            view_point = mn.Vector2i(
+                random.randint(0, render_camera.viewport[0] - 1),
+                random.randint(0, render_camera.viewport[1] - 1),
+            )
+            # NOTE: use un-normlized rays scaled to unit z distance for this application
```
0mdc, line 119: Typo: _normlized_

0mdc (approved):
LGTM! Left some minor remarks.

src_python/habitat_sim/utils/settings.py
aclegg3: Fair point. This means any Simulator initialized from the default settings util will have access to Bullet API (if installed).
Given that tests are passing, I don't think this is problem. 

Some justification:
1. If user installed bullet compatible Habitat build, they should get Bullet API by default.
2. This incurs very little overhead for static scenes (builds collision meshes, but doesn't do any heavy compute), so it shouldn't reduce performance for non-physics use cases most of the time (e.g. PointNav with HM3D).
3. This is a fairly new utility Dict for quick-start and is not currently used downstream by Habitat-lab (outside of one test for the rearrange task which is physics based), so it is not likely a breaking change.

We could add the breaking change tag to the PR and mention this in the description.
If you think this warrants more discussion I could move this change to its own PR as it is not necessary here.

0mdc: Sounds good. I think we can proceed with it here.

jturner65: @aclegg3 @0mdc What's the verdict on this change? Should we add the breaking change tag?  Otherwise, this PR looks good to merge.

0mdc: I think that it would be preferable to add this change to another PR for traceability. However, I don't feel strongly either way.

PR #2014, opened 2023-02-22
MultiGoalShortestPath improvements

aclegg3:
## Motivation and Context

Currently, if any of the requested_end points don't snap to the navmesh, `MultiGoalShortestPath` will early abort and return `inf` distance.
Expected behavior: if *any* of the points in `requested_ends` are valid, the shortest path should be computed accurately.

Additional feature: adds `closest_end_point_index` to `MultiGoalShortestPath` struct API. Makes it easier to know which point form the input set was actually selected as closets. Corresponds to the end of the `path`.

## How Has This Been Tested

TODO:
May need to update tests to check these edge cases:
- all points off navmesh
- some points off navmesh
- all points on other island
- all point son other island or off navmesh

## Types of changes

## Checklist

- [x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/src/esp/bindings/ShortestPathBindings.cpp
+++ b/src/esp/bindings/ShortestPathBindings.cpp
@@ -70,7 +70,7 @@ void initShortestPathBindings(py::module& m) {
       .def_readwrite(
           "closest_end_point_index",
           &MultiGoalShortestPath::closestEndPointIndex,
-          R"(he index of the closest end point corresponding to end of the shortest path. Will be -1 if no path exists.)");
+          R"(The index of the closest end point corresponding to end of the shortest path. Will be -1 if no path exists.)");
 
   py::class_<NavMeshSettings, NavMeshSettings::ptr>(
       m, "NavMeshSettings",
--- a/src/esp/bindings/ShortestPathBindings.cpp
+++ b/src/esp/bindings/ShortestPathBindings.cpp
@@ -66,7 +66,11 @@ void initShortestPathBindings(py::module& m) {
           R"(A list of points that specify the shortest path on the navigation mesh between requestedStart and the closest (by geodesic distance) point in requestedEnds. Will be empty if no path exists.)")
       .def_readwrite(
           "geodesic_distance", &MultiGoalShortestPath::geodesicDistance,
-          R"(The total geodesic distance of the path. Will be inf if no path exists.)");
+          R"(The total geodesic distance of the path. Will be inf if no path exists.)")
+      .def_readwrite(
+          "closest_end_point_index",
+          &MultiGoalShortestPath::closestEndPointIndex,
+          R"(he index of the closest end point corresponding to end of the shortest path. Will be -1 if no path exists.)");
 
   py::class_<NavMeshSettings, NavMeshSettings::ptr>(
       m, "NavMeshSettings",
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -1335,6 +1335,7 @@ bool PathFinder::Impl::findPathSetup(MultiGoalShortestPath& path,
                                      dtPolyRef& startRef,
                                      vec3f& pathStart) {
   path.geodesicDistance = std::numeric_limits<float>::infinity();
+  path.closestEndPointIndex = -1;
   path.points.clear();
 
   // find nearest polys and path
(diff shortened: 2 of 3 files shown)


src/esp/bindings/ShortestPathBindings.cpp
```diff
@@ -66,7 +66,11 @@ void initShortestPathBindings(py::module& m) {
           R"(A list of points that specify the shortest path on the navigation mesh between requestedStart and the closest (by geodesic distance) point in requestedEnds. Will be empty if no path exists.)")
       .def_readwrite(
           "geodesic_distance", &MultiGoalShortestPath::geodesicDistance,
-          R"(The total geodesic distance of the path. Will be inf if no path exists.)");
+          R"(The total geodesic distance of the path. Will be inf if no path exists.)")
+      .def_readwrite(
+          "closest_end_point_index",
+          &MultiGoalShortestPath::closestEndPointIndex,
+          R"(he index of the closest end point corresponding to end of the shortest path. Will be -1 if no path exists.)");
```
jturner65, line 73: Slight spelling issue
`R"(he index of the closest end point corresponding to end of the shortest path. Will be -1 if no path exists.)");`

should be 
`R"(The index of the closest end point corresponding to end of the shortest path. Will be -1 if no path exists.)");`

src/esp/nav/PathFinder.cpp
```diff
@@ -1349,18 +1350,26 @@ bool PathFinder::Impl::findPathSetup(MultiGoalShortestPath& path,
...
         projectToPoly(rqEnd, navQuery_.get(), filter_.get());
 
     if (status != DT_SUCCESS || endRef == 0) {
-      return false;
+      ESP_DEBUG() << "Can't project end-point to navmesh, skipping: " << rqEnd;
+    } else {
+      numValidPoints++;
     }
 
     path.pimpl_->endRefs.emplace_back(endRef);
```
jturner65, line 1366: Do you really want to potentially add an endRef == 0 to the store of endRefs?

jturner65, line 1366: Do you really want to potentially add an endRef == 0 to the store of endRefs?

src/esp/nav/PathFinder.cpp
```diff
@@ -1349,18 +1350,26 @@ bool PathFinder::Impl::findPathSetup(MultiGoalShortestPath& path,
...
     path.pimpl_->pathEnds.emplace_back(pathEnd);
+
+    if (numValidPoints == 0) {
```
Skylion007, line 1369: Hold on, won't this still abort if the FIRST validPoint is doesn't snap even if subsequent ones are valid? Are the paths sorted?

src/esp/nav/PathFinder.cpp
```diff
@@ -1349,18 +1350,26 @@ bool PathFinder::Impl::findPathSetup(MultiGoalShortestPath& path,
...
+  int numValidPoints = 0;
   for (const auto& rqEnd : path.getRequestedEnds()) {
     dtPolyRef endRef = 0;
     vec3f pathEnd;
     std::tie(status, endRef, pathEnd) =
         projectToPoly(rqEnd, navQuery_.get(), filter_.get());
 
     if (status != DT_SUCCESS || endRef == 0) {
-      return false;
+      ESP_DEBUG() << "Can't project end-point to navmesh, skipping: " << rqEnd;
```
Skylion007, line 1361: If it's really skipping, shouldn't it have a continue logic in the loop?

src/esp/nav/PathFinder.cpp
Skylion007: If it is sorted, add a comment.

aclegg3: Lol, thank you for catching this. Typo, the conditional should be outside the loop.

src/esp/nav/PathFinder.cpp
aclegg3: Because the original end-points will be sorted in the next step, we still need an entry for invalid points so the indices will match. This warning message is primarily intended to let the user know that they are providing invalid points in case that is an unexpected bug in their algorithm. The skip here is implicit.

aclegg3: Added a way to track this and skip more explicitly.

jturner65 (approved):
LGTM!

jturner65 (approved):

jturner65 (approved):

PR #2011, opened 2023-02-16
Add replay renderer bindings to get CUDA device pointers.

0mdc:
## Motivation and Context

This adds replay renderer bindings to get CUDA device pointers from Python.

## How Has This Been Tested

Device pointers are tested in `BatchReplayRendererTest`.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2018, opened 2023-02-27
Add batch renderer API for rendering standalone applications.

0mdc:
## Motivation and Context

This changeset adds a general `render()` method to replay renderer to render into default framebuffer.
This is used by standalone applications, i.e. applications that don't create their own context, like Habitat-lab.

## How Has This Been Tested

Tested locally on Habitat-lab integration prototype.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx_batch/Renderer.h
+++ b/src/esp/gfx_batch/Renderer.h
@@ -696,6 +696,11 @@ class Renderer {
       Magnum::UnsignedInt sceneId);
 
   /**
+   * @brief Draw all scenes in the default framebuffer.
+   */
+  virtual void draw();
+
+  /**
    * @brief Draw all scenes into provided framebuffer
    *
    * The @p framebuffer is expected to have a size at least as larger as the
--- a/src/esp/gfx_batch/Renderer.h
+++ b/src/esp/gfx_batch/Renderer.h
@@ -485,7 +485,7 @@ class Renderer {
   /**
    * @brief Max light count
    *
-   * By default there's zero lights, i.e. flat-shaded renderering.
+   * By default there's zero lights, i.e. flat-shaded rendering.
    * @see @ref RendererConfiguration::setMaxLightCount()
    */
   Magnum::UnsignedInt maxLightCount() const;
--- a/src/esp/sim/AbstractReplayRenderer.cpp
+++ b/src/esp/sim/AbstractReplayRenderer.cpp
@@ -90,6 +90,10 @@ void AbstractReplayRenderer::setSensorTransformsFromKeyframe(
   return doSetSensorTransformsFromKeyframe(envIndex, prefix);
 }
 
+void AbstractReplayRenderer::render() {
+  doRender();
+}
+
 void AbstractReplayRenderer::render(
     Cr::Containers::ArrayView<const Mn::MutableImageView2D> imageViews) {
   CORRADE_ASSERT(imageViews.size() == doEnvironmentCount(),
--- a/src/esp/sim/AbstractReplayRenderer.h
+++ b/src/esp/sim/AbstractReplayRenderer.h
@@ -95,11 +95,15 @@ class AbstractReplayRenderer {
   void setSensorTransformsFromKeyframe(unsigned envIndex,
                                        const std::string& prefix);
 
-  // Renders and waits for the render to finish
+  // Render into the default framebuffer. Assumes the framebuffer color & depth
+  // is cleared.
+  void render();
+
+  // Renders and waits for the render to finish.
   void render(Corrade::Containers::ArrayView<const Magnum::MutableImageView2D>
                   imageViews);
 
-  // Assumes the framebuffer color & depth is cleared
+  // Assumes the framebuffer color & depth is cleared.
   void render(Magnum::GL::AbstractFramebuffer& framebuffer);
 
   // Retrieve the color buffer as a CUDA device pointer. */
@@ -136,6 +140,8 @@ class AbstractReplayRenderer {
   /* envIndex is guaranteed to be in bounds */
   virtual void doSetSensorTransformsFromKeyframe(unsigned envIndex,
                                                  const std::string& prefix) = 0;
+
+  virtual void doRender() = 0;
 
   /* imageViews.size() is guaranteed to be same as doEnvironmentCount() */
   virtual void doRender(
--- a/src/esp/sim/AbstractReplayRenderer.h
+++ b/src/esp/sim/AbstractReplayRenderer.h
@@ -103,7 +103,8 @@ class AbstractReplayRenderer {
   void render(Corrade::Containers::ArrayView<const Magnum::MutableImageView2D>
                   imageViews);
 
-  // Assumes the framebuffer color & depth is cleared.
+  // Renders into an external framebuffer. Assumes the framebuffer color & depth
+  // is cleared.
   void render(Magnum::GL::AbstractFramebuffer& framebuffer);
 
   // Retrieve the color buffer as a CUDA device pointer. */
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -418,10 +418,15 @@ void initSimBindings(py::module& m) {
       .def("clear_environment", &AbstractReplayRenderer::clearEnvironment,
            "Clear all instances and resets memory of an environment.")
       .def("render",
-           static_cast<void (AbstractReplayRenderer::*)(
-               Magnum::GL::AbstractFramebuffer&)>(
+           static_cast<void (AbstractReplayRenderer::*)()>(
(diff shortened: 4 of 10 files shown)


src/esp/gfx_batch/Renderer.h
```diff
@@ -695,6 +695,11 @@ class Renderer {
   Corrade::Containers::StridedArrayView1D<Magnum::Float> lightRanges(
       Magnum::UnsignedInt sceneId);
 
+  /**
+   * @brief Draw all scenes in the default framebuffer.
+   */
+  virtual void draw();
```
mosra, line 701: Hmm, I don't like the `virtual` here, plus the implicit reliance on the default framebuffer. The low-level API should be as explicit as possible, and this function goes against that.

Why can't the `BatchReplayRenderer` just call `draw(GL::defaultFramebuffer)` on its own? Then it wouldn't need a new `draw()` overload at all.

src/esp/sim/AbstractReplayRenderer.cpp
```diff
@@ -90,6 +90,10 @@ void AbstractReplayRenderer::setSensorTransformsFromKeyframe(
   return doSetSensorTransformsFromKeyframe(envIndex, prefix);
 }
 
+void AbstractReplayRenderer::render() {
+  doRender();
+}
```
mosra, line 95: This could just call `doRender(Mn::GL::defaultFramebuffer)` instead, avoiding the need for a new `doRender()` overload altogether. 

Just FYI, a design goal I have with the private virtual `do*()` APIs is to have as little of them as possible, and all convenience handled by the public non-virtual APIs instead. A new `do*()` API means touching *seven* places in the code, while adding a convenience public API that reuses an existing `do*()` function is just one declaration and one definition.

src/esp/sim/AbstractReplayRenderer.h
```diff
@@ -95,11 +95,16 @@ class AbstractReplayRenderer {
   void setSensorTransformsFromKeyframe(unsigned envIndex,
                                        const std::string& prefix);
 
-  // Renders and waits for the render to finish
+  // Render into the default framebuffer. Assumes the framebuffer color & depth
+  // is cleared.
+  void render();
```
mosra, line 100: Hmm, thinking further about all this, what was wrong with calling `render(Mn::GL::defaultFramebuffer)` from the application code in the first place?

I see that it could possibly trip up [on an assert](https://github.com/facebookresearch/habitat-sim/blob/b6186aa8aaf9783f791ec4ffea00b091a766a96f/src/esp/sim/BatchReplayRenderer.cpp#L240-L242) which is there to prevent accidents like rendering to some arbitrary location but then trying to retrieve a CUDA image which would be inevitably stale or completely empty. But then the solution is to fix the root cause that causes the assert (such as not using a standalone renderer if you're rendering to a window) or rethink the use case for the assertion. Adding a completely new code path that sidesteps the assertion isn't a solution.

> Assumes the framebuffer color & depth is cleared.

Because this I see as a problem, there's a new API that *looks* easy, but it has implicit requirements that might not be obvious from reading the application code alone, and is a reason I didn't add such overload myself. Consider this, from the [replayer app](https://github.com/facebookresearch/habitat-sim/blob/b6186aa8aaf9783f791ec4ffea00b091a766a96f/src/utils/replayer/replayer.cpp#L197-L233):

```cpp
  Mn::GL::defaultFramebuffer.clear(Mn::GL::FramebufferClear::Color |
                                   Mn::GL::FramebufferClear::Depth);
                                   
  ...

  // Explicitly passing the framebuffer. The code implies it's rendering
  // to the framebuffer the application prepared above, and nothing else.
  replayRenderer_->render(Mn::GL::defaultFramebuffer);
```

Compared to 

```cpp                 
  // Renders where exactly? Is the application supposed to bind some 
  // concrete framebuffer before? Is it supposed to clear it? Or is it 
  // cleared automatically?
  replayRenderer_->render();
```

It makes things even worse / more confusing given there's a parameter-less [draw()](https://aihabitat.org/docs/habitat-sim/classesp_1_1gfx__batch_1_1RendererStandalone.html#a3e2fe82df57fbef78034b1bc7fc2a2e7) on the standalone batch renderer, which does exactly the opposite of the new `render()` here -- in the standalone batch renderer it draws into an *internal* framebuffer, clearing it before.

mosra (changes requested):
Hmm, I don't know, I don't think this actually solves anything that wouldn't be achievable with the original API already.

The comments reflect my thought process while reading the code, probably enough to act upon just the last one.

src/esp/sim/AbstractReplayRenderer.h
0mdc: I agree with those remarks. I like the explicit approach better too.

>what was wrong with calling render(Mn::GL::defaultFramebuffer) from the application code in the first place?

When writing this code, I had these two concerns:
* The assert trips-up when used from Habitat-Lab, which is a standalone renderer. Here, I am rendering to the main framebuffer, offscreen, and fetching the results using CUDA pointers.
* Magnum depedency is required by the application to pass the parameter. *This is not a valid concern anymore as we are moving forward with adding Magnum as a dependency to Habitat-Lab.*

Given this, I would roll-back the API changes, but allow for using the render function from a standalone renderer.

mosra: > Magnum depedency is required by the application to pass the parameter.

~~If that would be a valid concern, I'd propose having an explicitly-named `renderToDefaultFramebuffer()` API on the `AbstractReplayRenderer` class instead, which calls `render(GL::defaultFramebuffer)` internally, and no other change.~~ But that still isn't the right solution I think -- see below.

> Here, I am rendering to the main framebuffer, offscreen, and fetching the results using CUDA pointers.

Okay, let's talk about how to solve this best -- does this mean you call `render(list of images)` as well as `render(default framebuffer)`? I.e., it's possible that both get called in the same frame? Um, actually, what do you call to render to just CUDA images and not fetch them via image views? I don't see any relevant API on the current `AbstractReplayRenderer` class, huh, sounds like a strange omission :thinking: 

In any case, instead of rendering twice (which also has to do all draw preparation, culling and whatnot twice) I think it'd be better to render just once the internal framebuffer and then blit it to the default one. So, adding APIs like `blitColorInto(destination, sourceRectangle, destinationRectangle)` / `blitDepthInto(destination, sourceRectangle, destinationRectangle)` to the `BatchRendererStandalone` class, and then calling those after.

0mdc: >Okay, let's talk about how to solve this best -- does this mean you call render(list of images) as well as render(default framebuffer)?

Not exactly. I'm calling *either* one or the other:

For the CUDA code path, I am using `render(default framebuffer)`, then `getCudaColorBufferDevicePointer()` to get the results. I'm using `torch::from_blob()` ([ref](https://pytorch.org/cppdocs/api/function_namespacetorch_1ad7fb2a7759ef8c9443b489ddde494787.html)) to build a tensor that references that GPU memory.

For the fallback non-CUDA code path, I am using `render(list of images)`.

>Um, actually, what do you call to render to just CUDA images and not fetch them via image views?

`getCudaColorBufferDevicePointer()` calls `RendererStandalone::colorCudaBufferDevicePointer()`, which copies the default framebuffer into a linear CUDA buffer.

mosra: Huh...

> copies the default framebuffer

... are you sure this works and doesn't copy something stale? Because this code is always reading from the internal framebuffer, not the one that was passed to `render()`: https://github.com/facebookresearch/habitat-sim/blob/107013b63136648af00b52af189b99821a4924e9/src/esp/gfx_batch/RendererStandalone.cpp#L214-L215

I just can't see how this could do the right thing :sweat_smile: The whole `ReplayRenderer` API feels strange to me, and the CUDA pointer retrievals added in #2011 just can't work if I see correctly -- there's simply no way to render directly to the internal framebuffer and then retrieve the images as CUDA pointers. Sorry for not catching that earlier. 

I think splitting `render(list of images)` into `render()` and `colorImagesInto(list of images)` / `depthImagesInto(list of images)` could fix it, then you'll have a possibility of rendering into the internal framebuffer and then either copying them via CUDA or to the CPU. And for rendering to the default framebuffer the `blit*()` APIs I proposed above.

0mdc: I see.

Visually, the results were fine which led me to believe that this was the original intent.

>I think splitting render(list of images) into render() and colorImagesInto(list of images) / depthImagesInto(list of images) could fix it, then you'll have a possibility of rendering into the internal framebuffer and then either copying them via CUDA or to the CPU.

This makes more sense! I'll try that out.

0mdc:
Closing as obsolete.

After discussion, we decided to move forward with a different API for rendering with CUDA from standalone applications:
`CudaPointers render(RenderTargets)`

PR #2019, opened 2023-02-27
Allow the batch renderer to draw empty scenes.

0mdc:
## Motivation and Context

This replaces an assertion that prevents drawing empty scenes by an early-out.

## How Has This Been Tested

Tested locally and via unit test.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx_batch/Renderer.cpp
+++ b/src/esp/gfx_batch/Renderer.cpp
@@ -1076,6 +1076,10 @@ Cr::Containers::StridedArrayView1D<Mn::Float> Renderer::lightRanges(
 }
 
 void Renderer::draw(Mn::GL::AbstractFramebuffer& framebuffer) {
+  if (state_->shaders.empty()) {
+    return;
+  }
+
   /* Process scenes that are marked as dirty */
   // TODO this could be a separate step to allow the user to control when it
   //  runs
--- a/src/esp/gfx_batch/Renderer.cpp
+++ b/src/esp/gfx_batch/Renderer.cpp
@@ -1076,10 +1076,6 @@ Cr::Containers::StridedArrayView1D<Mn::Float> Renderer::lightRanges(
 }
 
 void Renderer::draw(Mn::GL::AbstractFramebuffer& framebuffer) {
-  // TODO allow this (currently addFile() sets up shader limits)
-  CORRADE_ASSERT(!state_->meshes.isEmpty(),
-                 "Renderer::draw(): no file was added", );
-
   /* Process scenes that are marked as dirty */
   // TODO this could be a separate step to allow the user to control when it
   //  runs
--- a/src/esp/gfx_batch/Renderer.cpp
+++ b/src/esp/gfx_batch/Renderer.cpp
@@ -1076,7 +1076,10 @@ Cr::Containers::StridedArrayView1D<Mn::Float> Renderer::lightRanges(
 }
 
 void Renderer::draw(Mn::GL::AbstractFramebuffer& framebuffer) {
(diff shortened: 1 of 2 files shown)


src/esp/gfx_batch/Renderer.cpp
```diff
@@ -1076,10 +1076,6 @@ Cr::Containers::StridedArrayView1D<Mn::Float> Renderer::lightRanges(
 }
 
 void Renderer::draw(Mn::GL::AbstractFramebuffer& framebuffer) {
-  // TODO allow this (currently addFile() sets up shader limits)
-  CORRADE_ASSERT(!state_->meshes.isEmpty(),
-                 "Renderer::draw(): no file was added", );
-
```
mosra, line 1082: The reason there's an assertion and a TODO is that the shaders aren't initialized if `addFile()` wasn't called, and IIRC that led to a GL error somewhere in the code below.

To pin down the behavior, can you add a dedicated test case for this into `GfxBatchRendererTest.cpp` and verify that it doesn't produce a GL error (with `MAGNUM_VERIFY_NO_GL_ERROR()`)? If it does, then it probably needs an early out in case there's nothing to draw from.

0mdc: Sounds good!

src/esp/gfx_batch/Renderer.cpp
```diff
@@ -1076,9 +1076,9 @@ Cr::Containers::StridedArrayView1D<Mn::Float> Renderer::lightRanges(
 }
 
 void Renderer::draw(Mn::GL::AbstractFramebuffer& framebuffer) {
-  // TODO allow this (currently addFile() sets up shader limits)
-  CORRADE_ASSERT(!state_->meshes.isEmpty(),
-                 "Renderer::draw(): no file was added", );
+  if (state_->shaders.empty()) {
```
mosra, line 1079: Tightening & explaining this a bit:

```suggestion
  /* If addFile() was not called we don't have the shaders set up yet. In that case
       there should be no meshes to render from either, so nothing to do. */
  if (state_->shaders.empty()) {
    CORRADE_INTERNAL_ASSERT(state_->meshes.isEmpty());
```

src/tests/GfxBatchRendererTest.cpp
```diff
@@ -1710,6 +1713,27 @@ void GfxBatchRendererTest::multipleMeshes() {
...
+          .setFlags(esp::gfx_batch::RendererStandaloneFlag::QuietLog)
+  };
+  // clang-format on
+
+  for (int i = 0; i < 4; ++i) {
+    esp::gfx_batch::SceneStats stats = renderer.sceneStats(i);
+    CORRADE_COMPARE(stats.nodeCount, 0);
+    CORRADE_COMPARE(stats.drawCount, 0);
+    CORRADE_COMPARE(stats.drawBatchCount, 0);
+  }
```
mosra, line 1731: No need to test this here, checking stats before any file is added is already done elsewhere.

src/tests/GfxBatchRendererTest.cpp
```diff
@@ -1710,6 +1713,27 @@ void GfxBatchRendererTest::multipleMeshes() {
   CORRADE_COMPARE(color.pixels<Mn::Color4ub>()[24][88], 0xcc00cc_rgb);
 }
 
+void GfxBatchRendererTest::emptyScene() {
```
mosra, line 1716:
```suggestion
void GfxBatchRendererTest::renderNoFileAdded() {
```

mosra (approved):

PR #2024, opened 2023-03-01
[bugfix] - color overflow in get_island_colored_map

aclegg3:
## Motivation and Context

If there are more islands than pre-set colors we should invent some new ones to avoid overflow.

Also fixes a bug with the images being reflected because PIL corrdinate system is +Y downwards with origin in the top left.

## How Has This Been Tested

Locally with large disjoint navmeshes.

## Types of changes

## Checklist


jturner65 (approved):

Skylion007:
@aclegg3 We shouldn't pick a random color, we should try to pick colors that are maximally far away from each other in the appropiate colorspace like HSV. There is a nice algorithm to do this which is used for colormaps, but it's a bit tricky here since we have a manual colormap defined first and then we have to exhaust the remaining colors.

PR #2028, opened 2023-03-09
--Configuration support for Magnum Vector2, Vector4, Color3 and Color4, UserDefinedAttr support of str Lists

jturner65:
## Motivation and Context
This PR adds Configuration support for Magnum Vector4.  With regard to the user_defined attributes, the type for equivalently sized fields in JSON is determined by substring tags found in the numeric fields' labels -  'orient', 'quat', or 'rotat' will denote a quaternion for a size 4 numeric field, and otherwise the size 4 fields will be Magnum vector4 's.

These are some features that I worked on some before my last contract ended and felt would be useful to add now.

NOTE : I removed Explicit Color3 and Color4 since these were superfluous, but I retrained accessors that would treat Color3 and Color4 getter requests the same as Vector3 and Vector4 respectively.

NOTE : I also added support of user-defined attributes containing a list of variable length of strings, which was already specified as supported in the documentation :) .

## How Has This Been Tested
All Python and C++ tests pass. C++ tests were added to exhaustively test label-based type inference.

## Types of changes

## Checklist

### DIFF
--- a/docs/pages/attributesJSON.rst
+++ b/docs/pages/attributesJSON.rst
@@ -52,17 +52,17 @@ Configuring Child Attributes
 Stages, Objects, Lights, and Scenes can be linked from .json, created directly from assets (e.g. .glb) or configured directly in this file's configuration nodes via the syntax below:
 
 "stages" \| "objects" \| "articulated_objects" \| "light_setups" \| "scene_instances"
-    - json object
+    - JSON object
     - Configuration pertaining to the specified attribute type.
 
     - "default_attributes"
-        - json object
-        - Set default attributes for all like objects in the dataset. Individual objects can override these values. See StageAttributes, ObjectAttributes, etc... below.
+        - JSON object
+        - Set default attributes for all like objects in the dataset. Individual JSON instances can override these values. See StageAttributes, ObjectAttributes, etc... below.
     - "paths"
-        - json object
+        - JSON object
         - keyed by file extension to search for. Value is a list of paths relative to this file to search for the designated filetype.
     - "configs"
-        - list of json objects
+        - list of JSON objects
         - Define modified copies or completely new Attributes objects directly in the SceneDatasetAttributes. See StageAttributes, ObjectAttributes, etc... below.
             - "original_file"
                 - string
@@ -71,11 +71,11 @@ Stages, Objects, Lights, and Scenes can be linked from .json, created directly f
                 - string
                 - Handle/name for the new *Attributes*. Used to reference it withing its *AttributesManager*.
             -  "attributes"
-                - json object
+                - JSON object
                 - Attribute configuration to override defaults or copied values.
 
 "semantic_scene_descriptor_instances" \| "navmesh_instances"
-    - json object
+    - JSON object
     - key\|value pairs link handles to relative filepaths for all Semantic Scene Descriptor (SSD) and NavMesh files. These can then be referenced by name within `SceneInstanceAttributes`.
 
 `SceneInstanceAttributes`_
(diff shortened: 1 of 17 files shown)


src/esp/bindings/ConfigBindings.cpp
```diff
@@ -24,8 +24,14 @@ py::object getObjectForConfigValue(const ConfigValue& value) {
...
     case ConfigStoredType::String:
       return py::cast(value.get<std::string>());
+    case ConfigStoredType::MagnumColor3:
+      return py::cast(value.get<Mn::Color3>());
     case ConfigStoredType::MagnumVec3:
       return py::cast(value.get<Mn::Vector3>());
+    case ConfigStoredType::MagnumVec4:
+      return py::cast(value.get<Mn::Vector4>());
+    case ConfigStoredType::MagnumColor4:
+      return py::cast(value.get<Mn::Color4>());
```
mosra, line 34: What I do in most APIs (SceneData, MeshData, MaterialData) is treating `Vector4` and `Color4` as the same type, and same for the three-component type. There's no reason not to, it's three/four numbers in both cases, and it reduces the API surface area quite significantly.

mosra: That should be possible with the Vector type as well, the rgb(a) accessors are defined there already, next to xyz(w). What Color has on top is mainly colorspace conversion (sRGB, HSV, XYZ).

But nevertheless, what I meant was that you *can* still use `Color4` to both set and query a value, just that it'd be internally treated the same as `Vector4`. Which also implies that you can set/query `Vector4` values as `Color4` and vice versa, and it doesn't cause any error.

jturner65: Oh, yeah, I remember that.  Ok, so me saving this as a color3 and color4 is needlessly redundant. I'll remove those two. Thanks!

jturner65: ...but keep the accessors, so folks can get Vector3 and Vector4 elements as Color3 and Color4 elements, if they wish.

0mdc (approved):
Looks good to me!

We might want to consider supporting `Mn::Vector2` as well.

docs/pages/attributesJSON.rst
```diff
@@ -502,12 +509,37 @@ For example, :ref:`ObjectAttributes.get_user_config` returns the configuration o
...
+[0,1]                   Magnum Vector2 (float)
+[0,1,2]                 Magnum Vector3 (float)
+[0,1,3,4]               Magnum Vector4 (float)      1
+[0,1,2,3]               Magnum Quaternion (float)   1
+[1,2,3,4,5,6,7,8,9]     Magnum Matrix3 (float)      2
+
+======================= =========================== ===================
+
+1 - If a length-4 numeric vector's tag contains as a substring 'quat', 'orient' or 'rotat', case-insensitive, the object will be loaded and processed as a Magnum::Quaternion (w,x,y,z) by the parser. Otherwise, it will be loaded as a Magnum::Vector4.
+2 - A length-9 numeric vector will be mapped into a Magnum Matrix3 in column-major order.
```
aclegg3, line 541: I think you need an extra newline here to break the paragraph.

aclegg3 (approved):
Thanks for fixing and adding things here. Also appreciate the typo fixes throughout. :)

docs/pages/attributesJSON.rst
jturner65: Wasn't sure if the doc parser would do that for us.

aclegg3: Yeah, I used the github rst file viewer to noticed this. Maybe different parsers will do things differently, but I think generally a newline is required in rst to break paragraphs.

PR #2031, opened 2023-03-14
Disable sim sensors when using the batch renderer.

0mdc:
## Motivation and Context

This change prevents the simulators to allocate sensor memory and render them when using the batch renderer.

## How Has This Been Tested

Tested locally from habitat-lab.

## Types of changes

## Checklist


src_python/habitat_sim/simulator.py
```diff
@@ -431,18 +437,23 @@ def get_sensor_observations(
...
-
-        # As backport. All Dicts are ordered in Python >= 3.7
+        # As backport. All Dicts are ordered in Python >= 3.7.
         observations: Dict[int, ObservationDict] = OrderedDict()
+
+        # Draw observations for classic non-batched renderer.
+        if not self.config.enable_batch_renderer:
+            for agent_id in agent_ids:
+                agent_sensorsuite = self.__sensors[agent_id]
+                for _sensor_uuid, sensor in agent_sensorsuite.items():
```
Skylion007, line 447:
```suggestion
                for sensor in agent_sensorsuite.values():
```

src_python/habitat_sim/simulator.py
```diff
@@ -431,18 +437,23 @@ def get_sensor_observations(
...
+                    sensor.draw_observation()
+
+        # Get observations.
         for agent_id in agent_ids:
```
eundersander, line 451: Can you explain what's going on here? How can we safely call `get_observation` if we haven't rendered here? Is this relying on some external code to have already triggered rendering?

src_python/habitat_sim/simulator.py
```diff
@@ -576,6 +587,10 @@ def __init__(self, sim: Simulator, agent: Agent, sensor_id: str) -> None:
 
         self._spec = self._sensor_object.specification()
 
+        # When using the batch renderer, no memory is allocated here.
+        if self._sim.config.enable_batch_renderer:
```
eundersander, line 591: Feels a bit dangerous to just return out of the function early. How about wrap all the renderer code below in a helper function and only call that for `not enable_batch_renderer`.

Also, is it safe/correct to skip the noise model piece below for the batch renderer case?

src_python/habitat_sim/simulator.py
```diff
@@ -431,18 +437,23 @@ def get_sensor_observations(
...
-            agent_sensorsuite = self.__sensors[agent_id]
-            for _sensor_uuid, sensor in agent_sensorsuite.items():
-                sensor.draw_observation()
-
-        # As backport. All Dicts are ordered in Python >= 3.7
+        # As backport. All Dicts are ordered in Python >= 3.7.
         observations: Dict[int, ObservationDict] = OrderedDict()
+
+        # Draw observations for classic non-batched renderer.
+        if not self.config.enable_batch_renderer:
```
eundersander, line 444: As a readability thing, it might be nice to see an else block here with a `pass` and a comment about the batch renderer being a work in progress or not implemented yet or whatever. This would help answer my question/confusion below ("how can we safely call...").

eundersander (approved):
I left some comments. Have a look and try to address, but basically LGTM.

src_python/habitat_sim/simulator.py
0mdc: > How about wrap all the renderer code below in a helper function and only call that for not enable_batch_renderer

Sounds good!

> Also, is it safe/correct to skip the noise model piece below for the batch renderer case?

The noise modeling has to be done by the batch renderer. I think that it could make sense to preserve the noise model function loading here and call it from Habitat-Lab after rendering.

src_python/habitat_sim/simulator.py
0mdc: When batch rendering, `get_observation` returns `None`, creating an empty "placeholder" for this sensor uuid. The batch renderer replaces this value by the rendered data in `post_step()`.

This is a bit of a hack as Habitat-Lab has checks to make sure that the sensor uuid is present between `reset()`/`step()` and `post_step()`.

PR #2035, opened 2023-03-16
Gfx-replay polish: workaround for material-overrides; new keyframe getter

eundersander:
## Motivation and Context

Some polish needed to get the batch renderer working in a new Python application.

## How Has This Been Tested

Ran it locally.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/replay/Recorder.h
+++ b/src/esp/gfx/replay/Recorder.h
@@ -133,15 +133,13 @@ class Recorder {
   /**
    * @brief write saved keyframes as individual strings ['{"keyframe": ...}',
    * '{"keyframe": ...}', ...]
-   * @param incremental whether future saved keyframes will be incremental
    *
-   * Set incremental to true if you are using keyframes incrementally, e.g.
-   * repeated calls to this function and feeding them to a renderer. Set to
-   * false if you want the next call to writeSavedKeyframes* to be "standalone",
-   * i.e. the next frame will properly initialize all scene state and not rely
-   * on previous keyframes.
+   * Use this function if you are using keyframes incrementally, e.g.
+   * repeated calls to this function and feeding them to a renderer. Contrast
+   * with writeSavedKeyframesToFile, which "consolidates" before discarding old
+   * keyframes to avoid losing state information.
    */
-  std::vector<std::string> writeSavedKeyframesToStringArray(bool incremental);
+  std::vector<std::string> writeIncrementalSavedKeyframesToStringArray();
 
   /**
    * @brief returns JSONized version of given keyframe.
--- a/src/esp/gfx/replay/Recorder.h
+++ b/src/esp/gfx/replay/Recorder.h
@@ -126,9 +126,22 @@ class Recorder {
                                  bool usePrettyWriter = false);
 
   /**
-   * @brief write saved keyframes to string.
+   * @brief write saved keyframes to string. '{"keyframes": [{...},{...},...]}'
    */
   std::string writeSavedKeyframesToString();
+
+  /**
+   * @brief write saved keyframes as individual strings ['{"keyframe": ...}',
+   * '{"keyframe": ...}', ...]
+   * @param incremental whether future saved keyframes will be incremental
+   *
+   * Set incremental to true if you are using keyframes incrementally, e.g.
+   * repeated calls to this function and feeding them to a renderer. Set to
+   * false if you want the next call to writeSavedKeyframes* to be "standalone",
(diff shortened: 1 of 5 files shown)


0mdc (approved):
LGTM!

src/esp/gfx/replay/Recorder.h
```diff
@@ -126,10 +126,23 @@ class Recorder {
...
+   * '{"keyframe": ...}', ...]
+   * @param incremental whether future saved keyframes will be incremental
+   *
+   * Set incremental to true if you are using keyframes incrementally, e.g.
+   * repeated calls to this function and feeding them to a renderer. Set to
+   * false if you want the next call to writeSavedKeyframes* to be "standalone",
+   * i.e. the next frame will properly initialize all scene state and not rely
+   * on previous keyframes.
+   */
+  std::vector<std::string> writeSavedKeyframesToStringArray(bool incremental);
```
mosra, line 144: Boolean arguments are an API smell -- `writeSavedKeyframesToStringArray(false)` could imply "not write", "not saved", "not array" or any other meaning until you actually check the API documentation.

For consistency with `writeSavedKeyframesToString()` i think the easiest would be adding a new `writeIncrementalSavedKeyframesToStringArray()` variant that doesn't call the `consolidate*()` inside.

src/esp/gfx/replay/Recorder.h
```diff
@@ -126,10 +126,23 @@ class Recorder {
...
 
   /**
-   * @brief write saved keyframes to string.
+   * @brief write saved keyframes to string. '{"keyframes": [{...},{...},...]}'
    */
   std::string writeSavedKeyframesToString();
 
+  /**
+   * @brief write saved keyframes as individual strings ['{"keyframe": ...}',
+   * '{"keyframe": ...}', ...]
```
mosra, line 135: Hmm, this just piles on more and more workarounds that were initially caused by the parser expecting a single keyframe to not be a `{...}` but rather `{"keyframe": {...}}`, which then means one then has to process `[{...}, {...}, {...}]` into `[{"keyframe": {...}}, {"keyframe": {...}}, {"keyframe": {...}}]` and it's all a lot more code and work and runtime overhead than if the API would accept just the `{...}` directly.

I remember already partially working around that for the `replayer` utility with adding a [new variant that doesn't require the `"keyframe"` around](https://github.com/facebookresearch/habitat-sim/blob/e6e3cfd1eb7f5822c1b2e95de64059ea32c2902d/src/esp/gfx/replay/Player.cpp#L70-L77) -- could we switch to using that API always everywhere and reduce the two `writeSavedKeyframesToString*()` to be just one, returning either a `[{...}, {...}]` or a list of `{...}`, `{...}`? The former would be a lot less allocations, so preferrable from my side. 

I assume the main reason for this vector-returning API to exist is to avoid a JSON parsing-splitting-serizalization roundtrip on the Python side, right? Separating a `[{...}, {...}]` to individual keyframes as string sub-ranges is doable in an efficient way with no string copies or number parsing for example like in the `replayer` utility below, so the Python bindings could be doing that instead.

https://github.com/facebookresearch/habitat-sim/blob/e6e3cfd1eb7f5822c1b2e95de64059ea32c2902d/src/utils/replayer/replayer.cpp#L112-L126

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -99,11 +99,12 @@ BatchReplayRenderer::BatchReplayRenderer(
...
             << "not found in any composite file, loading from the filesystem";
-        // TODO asserts might be TOO BRUTAL?
-        CORRADE_INTERNAL_ASSERT_OUTPUT(
+
+        ESP_CHECK(
             renderer_.addFile(creation.filepath,
                               gfx_batch::RendererFileFlag::Whole |
-                                  gfx_batch::RendererFileFlag::GenerateMipmap));
+                                  gfx_batch::RendererFileFlag::GenerateMipmap),
+            "addFile failed for " << creation.filepath);
```
mosra, line 107:
```suggestion
            "addFile failed for" << creation.filepath);
```

src/esp/gfx/replay/Recorder.h
eundersander: I agree with your suggestions but I'm just being lazy here. I think of all these keyframe APIs as temporary until we do a perf optimization pass and refactor them. We've talked about moving away from JSON altogether.

mosra: Right, then it doesn't matter that much. I wasn't aware that this was all meant to be temporary.

PR #2044, opened 2023-03-22
GPU-to-CPU batch renderer API to render color and depth sensors into images.

0mdc:
## Motivation and Context

This changes the replay renderer API to support rendering of both color and depth sensors into images.
Semantic image views are expected to be added to the API at a later time.

Accompanying lab PR: https://github.com/facebookresearch/habitat-lab/pull/1408

Prior to this change, this API could only produce color images.

Before:
```
replay_renderer.render(color)
```

After:
```
replay_renderer.render(color_images=color, depth_images=depth)
```

## How Has This Been Tested

On Habitat-Lab and CI.

## Types of changes

## Checklist


src/esp/sim/AbstractReplayRenderer.cpp
```diff
@@ -91,12 +91,23 @@ void AbstractReplayRenderer::setSensorTransformsFromKeyframe(
...
+    CORRADE_ASSERT(colorImageViews.size() == doEnvironmentCount(),
+                   "ReplayRenderer::render(): expected"
+                       << doEnvironmentCount() << "color image views but got"
+                       << colorImageViews.size(), );
+  }
+  if (depthImageViews != nullptr) {
+    CORRADE_ASSERT(depthImageViews.size() == doEnvironmentCount(),
+                   "ReplayRenderer::render(): expected"
+                       << doEnvironmentCount() << "depth image views but got"
+                       << depthImageViews.size(), );
```
mosra, line 108: Might want to use ESP_CHECK in both instead, as this is possible to be triggered from Python, and ESP_CHECK will do a proper Python exception instead of an abort.

Hmm, and maybe also check for formats / sizes? Not sure how far do we want to go with the checks tho :)

src/esp/sim/AbstractReplayRenderer.cpp
```diff
@@ -91,12 +91,23 @@ void AbstractReplayRenderer::setSensorTransformsFromKeyframe(
 }
 
 void AbstractReplayRenderer::render(
-    Cr::Containers::ArrayView<const Mn::MutableImageView2D> imageViews) {
-  CORRADE_ASSERT(imageViews.size() == doEnvironmentCount(),
-                 "ReplayRenderer::render(): expected" << doEnvironmentCount()
-                                                      << "image views but got"
-                                                      << imageViews.size(), );
-  return doRender(imageViews);
+    Corrade::Containers::ArrayView<const Magnum::MutableImageView2D>
```
mosra, line 94:
```suggestion
    Cr::Containers::ArrayView<const Mn::MutableImageView2D>
```

so it's not so excessively long. Same elsewhere and in other `*.cpp` files -- but not headers, that breaks cross-linking / confuses the generated documentation.

Given it was using the shorthands before, I suspect an IDE is to blame for expanding this? :D I don't have a solution for IDEs picking the shorter variant, unfortunately.

src/tests/BatchReplayRendererTest.cpp
```diff
@@ -182,7 +182,7 @@ void BatchReplayRendererTest::testIntegration() {
       renderer->setSensorTransformsFromKeyframe(envIndex, userPrefix);
     }
 
-    renderer->render(imageViews);
+    renderer->render(imageViews, nullptr);
```
mosra, line 185: Hmm, probably would be good to test all combinations here (color only, depth only, color + depth) so there's no accidentally unhandled / crashing case.

For verifying the depth, in the batch renderer tests themselves I'm only checking if the values make sense (being 0.999something, not 0 and not 1 in a few arbitrarily chosen pixels). Should be enough I think, no need to do any image comparison.

src/esp/bindings/SimBindings.cpp
```diff
@@ -425,12 +425,14 @@ void initSimBindings(py::module& m) {
       .def(
           "render",
           [](AbstractReplayRenderer& self,
-             std::vector<Mn::MutableImageView2D> images) {
-            self.render(images);
+             std::vector<Mn::MutableImageView2D> colorImageViews,
+             std::vector<Mn::MutableImageView2D> depthImageViews) {
+            self.render(colorImageViews, depthImageViews);
```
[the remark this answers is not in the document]
0mdc: I changed the PR description accordingly.

src/esp/bindings/SimBindings.cpp
```diff
@@ -435,12 +435,14 @@ void initSimBindings(py::module& m) {
       .def(
           "render",
           [](AbstractReplayRenderer& self,
-             std::vector<Mn::MutableImageView2D> images) {
-            self.render(images);
+             std::vector<Mn::MutableImageView2D> colorImageViews,
+             std::vector<Mn::MutableImageView2D> depthImageViews) {
```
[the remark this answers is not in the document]
0mdc: Great idea! I'll incorporate this to this changeset.

src/esp/sim/ClassicReplayRenderer.cpp
```diff
@@ -211,11 +213,28 @@ void ClassicReplayRenderer::doRender(
...
+          imageViews = nullptr;
+          break;
+      }
+
 #ifdef ESP_BUILD_WITH_BACKGROUND_RENDERER
-      renderer_->enqueueAsyncDrawJob(
-          visualSensor, sceneGraph, imageViews[envIndex],
-          esp::gfx::RenderCamera::Flags{
-              gfx::RenderCamera::Flag::FrustumCulling});
+      // todo: investigate flags (frustum culling?)
```
eundersander, line 231: Is this a real todo?

src/tests/BatchReplayRendererTest.cpp
```diff
@@ -203,53 +268,76 @@ void BatchReplayRendererTest::testIntegration() {
...
+            Cr::Utility::Path::join(screenshotDir, groundTruthImageFile),
+            (Mn::DebugTools::CompareImageToFile{maxThreshold, meanThreshold}));
+      }
+      // Test depth output
+      if (data.testFlags & TestFlag::Depth) {
+        const auto depth = depthImageViews[envIndex];
+        float pixelA = depth.pixels<Mn::Float>()[32][32];
+        float pixelB = depth.pixels<Mn::Float>()[64][64];
+        CORRADE_VERIFY(pixelA > 0.0f);
+        CORRADE_VERIFY(pixelA != pixelB);
```
eundersander, line 339: How about a comment explaining this line.

src/esp/sim/AbstractReplayRenderer.cpp
```diff
@@ -91,12 +91,21 @@ void AbstractReplayRenderer::setSensorTransformsFromKeyframe(
...
 void AbstractReplayRenderer::render(
-    Cr::Containers::ArrayView<const Mn::MutableImageView2D> imageViews) {
-  CORRADE_ASSERT(imageViews.size() == doEnvironmentCount(),
-                 "ReplayRenderer::render(): expected" << doEnvironmentCount()
-                                                      << "image views but got"
-                                                      << imageViews.size(), );
-  return doRender(imageViews);
+    Cr::Containers::ArrayView<const Mn::MutableImageView2D> colorImageViews,
+    Cr::Containers::ArrayView<const Mn::MutableImageView2D> depthImageViews) {
+  if (colorImageViews != nullptr) {
```
eundersander, line 96: I'm curious about the `nullptr` check here and then the use of default `std::vector<Mn::MutableImageView2D>{}` above. How does this work? If I omit `color_images` from my python call, do I get a zero-length `colorImageViews` here that fails this `ESP_CHECK`?

eundersander (commented):
I left some comments. I'm not super-familiar with this code. Maybe @mosra can take a final look and give a green check.

jturner65 (approved):
I may have missed something (I'm not that familiar with the batch renderer), but to me this looks good to go.

src/esp/sim/AbstractReplayRenderer.cpp
0mdc: Comparing `ArrayView` to `nullptr` checks for an empty array. I will standardize those avoid ambiguity.

eundersander: >Comparing ArrayView to nullptr checks for an empty array.
ಠ_ಠ

PR #2057, opened 2023-04-03
Replay renderer: add line-rendering and unproject()

eundersander:
## Motivation and Context

Adding some features (and hacks) to our experimental replay renderers, needed for a Python application for SIRo.

* integrate DebugLineRender into replay renderers
* add ClassicReplayRenderer.unproject
* stub implementation for BatchReplayRenderer.unproject
* lighting hack in ClassicReplayRenderer to get ReplicaCAD stages to render correctly

## How Has This Been Tested

Local testing.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -454,7 +454,11 @@ void initSimBindings(py::module& m) {
           [](AbstractReplayRenderer& self) {
             return py::capsule(self.getCudaColorBufferDevicePointer());
           },
-          R"(Retrieve the depth buffer as a CUDA device pointer.)");
+          R"(Retrieve the depth buffer as a CUDA device pointer.)")
+      .def("debug_line_render", &AbstractReplayRenderer::getDebugLineRender,
+           R"(Get visualization helper for rendering lines.)")
+      .def("unproject", &AbstractReplayRenderer::unproject,
+           R"(Unproject a screen-space point to a world-space ray.)");
 }
 
 }  // namespace sim
--- a/src/esp/sim/AbstractReplayRenderer.h
+++ b/src/esp/sim/AbstractReplayRenderer.h
@@ -116,6 +116,14 @@ class AbstractReplayRenderer {
   std::shared_ptr<esp::gfx::DebugLineRender> getDebugLineRender(
       unsigned envIndex);
 
+  /**
+   * @brief Unproject a 2D viewport point to a 3D ray with origin at camera
+   * position. Ray direction is normalized.
+   *
+   * @param envIndex
+   * @param viewportPosition The 2D point on the viewport to unproject
+   * ([0,width], [0,height]).
+   */
   esp::geo::Ray unproject(unsigned envIndex,
                           const Magnum::Vector2i& viewportPosition);
 
--- a/src/esp/sim/AbstractReplayRenderer.h
+++ b/src/esp/sim/AbstractReplayRenderer.h
@@ -8,7 +8,12 @@
 #include <Magnum/GL/GL.h>
 #include <Magnum/Magnum.h>
 
+#include "esp/core/Check.h"
 #include "esp/core/Esp.h"
+#include "esp/geo/Geo.h"
+#include "esp/gfx/DebugLineRender.h"
+
+#include <memory>
 
 namespace esp {
 
@@ -108,6 +113,17 @@ class AbstractReplayRenderer {
   // Retrieve the depth buffer as a CUDA device pointer. */
   virtual const void* getCudaDepthBufferDevicePointer();
 
+  std::shared_ptr<esp::gfx::DebugLineRender> getDebugLineRender(
+      unsigned envIndex);
+
+  esp::geo::Ray unproject(unsigned envIndex,
+                          const Magnum::Vector2i& viewportPosition);
+
+ protected:
+  void checkEnvIndex(unsigned envIndex);
+
+  std::shared_ptr<esp::gfx::DebugLineRender> debugLineRender_;
+
  private:
   /* Implementation of all public API is in the private do*() functions,
      similarly to how e.g. Magnum plugin interfaces work. The public API does
@@ -143,6 +159,9 @@ class AbstractReplayRenderer {
           imageViews) = 0;
 
   virtual void doRender(Magnum::GL::AbstractFramebuffer& framebuffer) = 0;
+
+  virtual esp::geo::Ray doUnproject(unsigned envIndex,
+                                    const Mn::Vector2i& viewportPosition) = 0;
 
   ESP_SMART_POINTERS(AbstractReplayRenderer)
 };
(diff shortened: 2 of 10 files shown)


src/esp/bindings/SimBindings.cpp
```diff
@@ -454,7 +454,11 @@ void initSimBindings(py::module& m) {
           [](AbstractReplayRenderer& self) {
             return py::capsule(self.getCudaColorBufferDevicePointer());
           },
-          R"(Retrieve the depth buffer as a CUDA device pointer.)");
+          R"(Retrieve the depth buffer as a CUDA device pointer.)")
+      .def("debug_line_render", &AbstractReplayRenderer::getDebugLineRender,
+           R"(Get visualization helper for rendering lines.)")
+      .def("unproject", &AbstractReplayRenderer::unproject,
```
eundersander, line 460: Some context: I'm writing a SIRo Python application that uses this. That app sets up a division between the renderer and the sim, paving the way for these pieces to live in separate processses or separate machines across a network (client/server architecture). Eventually, we need a way for sim code to unproject from screen space to world space, e.g. given a mouse cursor position (or arbitrary screen position), compute a world ray. I'd eventually like a replay renderer to provide an `unprojectHelper` object that encapsulates the camera matrix (including projection/viewport params). This object can be serialized and passed back to the sim, where the sim can do as many arbitrary `unproject`s as it wants. For now, as a bandaid, I'm having the renderer unproject the mouse cursor position, call this helper, and return the world ray to the sim.

See the unproject usage here (WIP):
https://github.com/eundersander/habitat-lab/blob/eundersander/magnum_glfw_interactive_play/habitat-lab/habitat/utils/gui_app_wrapper.py#L189

TLDR: this API should eventually go away in favor of a `getUnprojectHelper` API.

src/esp/bindings/SimBindings.cpp
```diff
@@ -454,7 +454,11 @@ void initSimBindings(py::module& m) {
           [](AbstractReplayRenderer& self) {
             return py::capsule(self.getCudaColorBufferDevicePointer());
           },
-          R"(Retrieve the depth buffer as a CUDA device pointer.)");
+          R"(Retrieve the depth buffer as a CUDA device pointer.)")
+      .def("debug_line_render", &AbstractReplayRenderer::getDebugLineRender,
```
eundersander, line 458: Context: see my earlier comment about separate sim and renderer. We eventually need a way for the sim to convey lines to the renderer. For now in my SIRo Python app, since both pieces live in the same thread, I'm simply giving the sim direct access to the `debug_line_render` object where it can directly add lines. See here:
https://github.com/eundersander/habitat-lab/blob/eundersander/magnum_glfw_interactive_play/habitat-lab/habitat/utils/gui_app_wrapper.py#L219

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -274,6 +278,24 @@ void BatchReplayRenderer::doRender(
...
+    debugLineRender_->flushLines(projCamMatrix, renderer_->tileSize());
+  }
+}
+
+esp::geo::Ray BatchReplayRenderer::doUnproject(
+    unsigned envIndex,
+    const Mn::Vector2i& viewportPosition) {
+  // temp stub implementation: produce a placeholder ray that varies with
+  // viewportPosition
+  return esp::geo::Ray(
```
eundersander, line 295: Another shortcut here. We can implement this properly later. As is, this will be useful for helping @0mdc integrate the batch replay renderer into the SIRo Python app.

src/esp/sim/ClassicReplayRenderer.cpp
```diff
@@ -31,6 +31,10 @@ ClassicReplayRenderer::ClassicReplayRenderer(
   resourceManager_ =
       std::make_unique<assets::ResourceManager>(metadataMediator, flags);
 
+  // hack to get ReplicCAD non-baked stages to render correctly
+  resourceManager_->getShaderManager().setFallback(
```
eundersander, line 35: Long-term, I hope @0mdc can figure out why the lighting isn't working correctly for the classic replay renderer.

mosra: I remember I was doing some material patching for the ReplicaCAD composites, in particular [adding or removing the flat/unlit material bit](https://github.com/facebookresearch/habitat-sim/blob/07d0b3069068446c886cccb9fae8cffaf0783f2f/src/utils/compositor/compositor-replicacad.cpp#L230-L239) (not sure which was the case exactly), because otherwise both the prebaked and non-baked arch files were rendered with the same shader, looking wrong.

So if this works for you for the batch renderer & my composites but not the classic one, that might be the reason.

src/esp/sim/BatchReplayRenderer.cpp
```diff
@@ -256,6 +257,9 @@ void BatchReplayRenderer::doRender(
                  "with a standalone renderer", );
   static_cast<gfx_batch::RendererStandalone&>(*renderer_).draw();
 
+  // todo: integrate debugLineRender_->flushLines
+  CORRADE_INTERNAL_ASSERT(!debugLineRender_);
```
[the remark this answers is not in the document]
mosra: On my "eventually nice to have" TODO list I have a batch-friendly line generator for the new [line shader](https://doc.magnum.graphics/magnum/classMagnum_1_1Shaders_1_1LineGL.html) (right now it's only possible to generate & draw a line mesh as a whole, not gradually add to it). 

Depending on how soon you'd like to have it working for the batch renderer, I can prioritize it.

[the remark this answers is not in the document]
eundersander: Related FYI, I can't think of a use case where we'll need line-rendering for a multi-tile (multi-env) setup. I.e. thick lines are *only* for showing to humans, not AI model training, because they are fundamentally a UX thing and not photorealistic. They're only for single-tile usage of the batch renderer.

src/esp/sim/AbstractReplayRenderer.h
0mdc, line 1: It would be great to include a screenshot-based test for line rendering and unit test for unproject.
I don't mind doing this in future changesets as this code is not fully developed.

src/esp/sim/AbstractReplayRenderer.h
```diff
@@ -108,6 +113,17 @@ class AbstractReplayRenderer {
   // Retrieve the depth buffer as a CUDA device pointer. */
   virtual const void* getCudaDepthBufferDevicePointer();
 
+  std::shared_ptr<esp::gfx::DebugLineRender> getDebugLineRender(
+      unsigned envIndex);
+
+  esp::geo::Ray unproject(unsigned envIndex,
```
0mdc, line 119: It would be helpful to comment that this is a viewport point to world-space ray function.

0mdc (approved):
LGTM!

src/esp/sim/AbstractReplayRenderer.h
eundersander: I added a unit test for `unproject`. An image test for line-rendering will probably be super-flaky, because thick-line support is flaky in OpenGL. @mosra has indicated that "real" thick-line rendering will be coming to Magnum eventually, so we could do an image test at that time.

PR #2058, opened 2023-04-03
--Have pre-commit use node version compatible with Ubuntu 18.04

jturner65:
## Motivation and Context
Apparently pre-commit 's default behavior is to install node as part of the prep process.  This PR tells pre-commit to use the system version of node if it exists.  (This was causing a problem on Ubuntu 18.04 since the installed version of node was requiring glibc 2.28 and 18.04 uses 2.27.)

## How Has This Been Tested
Pre-commit now works properly on 18.04

## Types of changes

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,6 @@ exclude: '^(build|data/datasets|data/scene_datasets|node_modules/|src/deps|src/o
 
 default_language_version:
     python: python3
-    node: system
 
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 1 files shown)


jturner65:
Note : An alternative mechanism for this would be to restrict the version of node to use/install so that we do not have the incompatibilities with Ubuntu 18.04.

0mdc (approved):
LGTM.

It may be preferable to isolate this from the system libs, but AFAIK we're only requiring node for pre-commit checks.

jturner65:
Apparently if node isn't found on the system it won't then try to download it, but will puke.  -sigh-. So I guess I have to figure out the minimum acceptable version of node to use.

jturner65:
Specifying an older, official release version of node to use for the eslint hook seems to behave properly, including in Ubuntu 18.04.  We might consider the feasibility of removing the eslint hook, if we are moving away from javascript support/development in the future.

0mdc (approved):

PR #2062, opened 2023-04-05
--(Bugfix) Fix normal transformation calc and address backface culling in case of negative scaling/reflections

jturner65:
## Motivation and Context
It is common for an artist to use negative scaling to reflect an asset, using the same mesh for two different configurations. Habitat-sim was using the incorrect matrix to transform normals (the cofactor matrix vs. the inverse-transform of the rotationScale matrix), which became apparent when negative scaling was used in a SceneInstance configuration to instantiate a reflected asset in floorplanner dataset.  

Reflection also changes winding order, which is used for backface culling. This PR addresses this as well, by changing backface culling winding order if reflection is caused due to a negative scale.

Below you can see this fix in action, the top image is what the scene looked like before the fix, and the bottom shows with the fix in place.

An Image comparison-based test was also added to SimTest to verify this process.

TODO (outside the scope of this PR): 
 - Normal Matrix calc needs to be addressed in batch renderer (via Magnum change)
 - Backface culling needs to be addressed in batch renderer.  

## How Has This Been Tested
Existing tests passed C++ and Python

## Types of changes

## Checklist

beforeNegScaleFix
afterNegScaleFix


src/esp/gfx_batch/Renderer.cpp
```diff
@@ -1206,7 +1206,9 @@ void Renderer::draw(Mn::GL::AbstractFramebuffer& framebuffer) {
           .drawsSorted[i]
           /* Extract normal matrix */
           .setNormalMatrix(state_->absoluteTransformationsSorted[i]
-                               .transformationMatrix.normalMatrix())
+                               .transformationMatrix.rotationScaling()
+                               .inverted()
+                               .transposed())
```
jturner65, line 1211: This addresses the erroneous normalMatrix() but does not yet address flipping the backFace culling winding value.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -143,7 +154,7 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
               : node_.getSemanticId())
       .setTransformationMatrix(transformationMatrix)
       .setProjectionMatrix(camera.projectionMatrix())
-      .setNormalMatrix(transformationMatrix.normalMatrix());
+      .setNormalMatrix(invNormalMat.transposed());
```
mosra, line 157: You're transposing twice, I don't think that is necessary. Just don't transpose at all.

Same for the PBR drawable.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -132,6 +133,16 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
 
   updateShaderLightingParameters(transformationMatrix, camera);
 
+  // Find determinant to calculate backface culling winding dir
+  auto rotScale = transformationMatrix.rotationScaling();
+  const float normalDet = rotScale.determinant();
+  auto invNormalMat = rotScale.comatrix().transposed() / normalDet;
```
mosra, line 139: A comment explaining the relation between a comatrix and inverse wouldn't hurt, I think. Something like this (which also includes the transpositions cancelling each other out).

```cpp
  // Normal matrix is calculated as `m.inverted().transposed()`, and
  // `m.inverted()` is the same as `m.comatrix().transposed()/m.determinant()`. 
  // We need the determinant to figure out the winding direction as well, thus 
  // we calculate it separately and then do 
  // `(m.comatrix().transposed()/determinant).transposed()`, which is the same 
  // as `m.comatrix()/determinant`.
```

Same for the PBR drawable.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -143,7 +160,7 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
               : node_.getSemanticId())
       .setTransformationMatrix(transformationMatrix)
       .setProjectionMatrix(camera.projectionMatrix())
-      .setNormalMatrix(transformationMatrix.normalMatrix());
+      .setNormalMatrix(invTransNormalMat);
```
mosra, line 163:
```suggestion
      .setNormalMatrix(normalMatrix);
```

That's what the matrix ultimately is, so I think it's clearest to name it like that. Same for the other drawable.

src/esp/gfx_batch/Renderer.cpp
jturner65: Fixing this for batch rendering is outside the scope of this PR.

aclegg3 (approved):
Solution looks good to me, thanks for fixing this!

I do wonder if there is any measure-able performance impact of additional maths in the draw() function. Can we check this before merging, maybe you and @mosra already thought about this.

Also, a nit: we're currently duplicating the solution and comments in PBRDrawable and GenericDrawable. Maybe this can be factored out into a function in base Drawable class for re-use?

jturner65:
> I do wonder if there is any measure-able performance impact of additional maths in the draw() function. Can we check this before merging, maybe you and @mosra already thought about this.

Well, there really isn't anything additional than what was there.  I'm just taking the calc apart so that I can query the determinant for face winding dir.  (The inverse.transpose == cofactor/det )
 
> Also, a nit: we're currently duplicating the solution and comments in PBRDrawable and GenericDrawable. Maybe this can be factored out into a function in base Drawable class for re-use?

I was going to do that but I needed two results - the boolean based on the determinant sign and the inverseTranspose normal matrix. I guess I could pass a ref and populate the results.

0mdc (approved):
LGTM!

It's probably impractical to do this here, but it would be helpful to provide a problematic asset so that we can quickly create a screenshot-based test. This would help addressing the batch rendering scenario.

jturner65:
> LGTM!
> 
> It's probably impractical to do this here, but it would be helpful to provide a problematic asset so that we can quickly create a screenshot-based test. This would help addressing the batch rendering scenario.

Great Idea! I'll try to put that together and get it in by Monday.

src/esp/gfx_batch/Renderer.cpp
mosra, line 1211: Actually, can you revert this part in the batch renderer? 

I decided to keep the `normalMatrix()` in (and just fix it) because it's a lot more convenient than having to remember and type out `foo.rotationScaling().inverted().transposed()`. So, once I get to doing the rest of the fix, I would be changing this back to `normalMatrix()` anyway.

The remaining changes are extracting the determinant calculation so they can't use `normalMatrix()` anymore.

mosra:
> it would be helpful to provide a problematic asset so that we can quickly create a screenshot-based test

It'd be good to have for the classic renderer in any case, but in the batch renderer it'll need a bit more depending on whether I'll go with a separate draw call or a specialized secondary index buffer. I'll be adding a test when I get to writing the full fix anyway, no need to do it twice.

src/esp/gfx_batch/Renderer.cpp
jturner65: Ok, I'll take care of that on Monday :)

jturner65:
The tests include image against file and image against image.  This second test type required a copy of the observation buffer to be made for the ground truth, non-negative-scaled observation, since the observation buffer is shared (Otherwise subsequent observations would overwrite previous observations and they would always match).

PR #2065, opened 2023-04-10
--[BugFix]Testing disabled renderer in simtest should only happen 1 time

jturner65:
## Motivation and Context
Currently the tests in simtest are instanced in such a way that they are each executed 2 times, once with a direct simulator creation, and once with a simulator created with a pre-existing metadata mediator, using Magnum's instanced testing protocols.  [This PR (#1308)](https://github.com/facebookresearch/habitat-sim/pull/1308) added support and testing for creating simulator without a renderer. 

The test that was added (createMagnumRenderingOff) did not follow the existing protocol for creating sims (calling the appropriate array of function pointers), but instead directly built it within the test's body. Due to this, the same test of the same functionality is being executed twice.

This PR makes the createMagnumRenderingOff test a regular test, and not an instanced one, so that it is executed only 1 time.  An alternative solution to this would be to test rendering off with both a directly-built simulator and one built using a prebuilt MM, but this would require further redesign of the testing protocol/layout to work properly.

## How Has This Been Tested
Current c++ tests pass, and createMagnumRenderingOff test is only executed once.

## Types of changes

## Checklist

### DIFF
--- a/src/tests/SimTest.cpp
+++ b/src/tests/SimTest.cpp
@@ -171,8 +171,8 @@ SimTest::SimTest() {
             &SimTest::loadingObjectTemplates,
             &SimTest::buildingPrimAssetObjectTemplates,
             &SimTest::addObjectByHandle,
-            &SimTest::addSensorToObject,
-            &SimTest::createMagnumRenderingOff}, Cr::Containers::arraySize(SimulatorBuilder) );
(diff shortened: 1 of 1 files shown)


eundersander (approved):
>should this functionality be tested both ways
I would say no, not necessary. This is a test of `createRenderer=false` and that feels unrelated to MetadataMediator stuff.

PR #2070, opened 2023-04-11
Add runtime perf stats for troubleshooting perf problems

eundersander:
## Motivation and Context

`Simulator.get_runtime_perf_stat_names` and `get_runtime_perf_stat_values` together are a flexible way to report a large number of arbitrary named scalars to Python. For now, I focus on some physics counts and Drawable counts.

I opted for separate `get...names` and `get...values` so that we can just call `get...values` on every env step and avoid the cost of creating/passing strings every step.

## How Has This Been Tested

Unit test added to SimTest.cpp

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -2910,6 +2910,12 @@ void ResourceManager::createDrawable(Mn::GL::Mesh* mesh,
                              : nullptr);  // pbr image based lighting
       break;
   }
+
+  drawableCountAndNumFaces_.first += 1;
+  if (mesh) {
+    drawableCountAndNumFaces_.second += mesh->count() / 3;
+  }
+
 }  // ResourceManager::createDrawable
 
 void ResourceManager::initDefaultLightSetups() {
--- a/src/esp/assets/ResourceManager.h
+++ b/src/esp/assets/ResourceManager.h
@@ -727,6 +727,19 @@ class ResourceManager {
       const std::shared_ptr<metadata::attributes::StageAttributes>&
           stageAttributes);
 
+  /**
+   * @brief Get the count of Drawables and the total face count across all
+   * Drawables in the scene. This is helpful for troubleshooting runtime perf.
+   * See also resetDrawableCountAndNumFaces.
+   */
+  auto getDrawableCountAndNumFaces() { return drawableCountAndNumFaces_; }
+
+  /**
+   * @brief Reset this count and this number. See also
+   * getDrawableCountAndNumFaces.
+   */
+  void resetDrawableCountAndNumFaces() { drawableCountAndNumFaces_ = {0, 0}; }
+
  private:
   /**
    * @brief Load the requested mesh info into @ref meshInfo corresponding to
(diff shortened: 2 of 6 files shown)


src/esp/sim/Simulator.cpp
```diff
@@ -1339,5 +1343,32 @@ int Simulator::getAgentObservationSpaces(
   }
   return spaces.size();
 }
+
+std::vector<std::string> Simulator::getRuntimePerfStatNames() {
+  return {"num rigid",           "num active rigid",    "num artic",
+          "num active overlaps", "num active contacts", "draw count",
+          "draw faces"};
```
0mdc, line 1350: Nit: I would use `drawable` instead of the shortened `draw` to ensure that this doesn't get misinterpreted.

0mdc (approved):
LGTM!

src/esp/assets/ResourceManager.cpp
```diff
@@ -2910,6 +2910,12 @@ void ResourceManager::createDrawable(Mn::GL::Mesh* mesh,
                              : nullptr);  // pbr image based lighting
       break;
   }
+
+  drawableCountAndNumFaces_.first += 1;
+  if (mesh) {
+    drawableCountAndNumFaces_.second += mesh->count() / 3;
```
mosra, line 2916: Best would be to query this on the `Trade::MeshData` instance before a `GL::Mesh` is made out of it (it's available at the point `createDrawable()` is called, as far as I see), there you get both index and vertex count, total data size etc.

For robust calculation of primitive count that does the right thing for triangle strips/fans as well there's [MeshTools::primitiveCount()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a746d76c4b365565d5f796d2ca74dc8bc) (meant to be called with `meshData.primitive()`, not the GL mesh primitive).

src/tests/SimTest.cpp
```diff
@@ -173,6 +175,7 @@ SimTest::SimTest() {
             &SimTest::addObjectByHandle,
             &SimTest::addSensorToObject}, Cr::Containers::arraySize(SimulatorBuilder) );
   addTests({&SimTest::createMagnumRenderingOff});
+  addTests({&SimTest::getRuntimePerfStats});
```
jturner65, line 178: Probably want to add this test to the list with createMagnumRenderingOff.

src/esp/assets/ResourceManager.h
```diff
@@ -1196,6 +1209,7 @@ class ResourceManager {
    * @brief The mesh data for loaded assets.
    */
   std::map<int, std::shared_ptr<BaseMesh>> meshes_;
+  std::pair<int, int> drawableCountAndNumFaces_{0, 0};
```
Skylion007, line 1212: Should we use Corrade Pair here?

mosra: StringView for that list of strings might make sense too, if they will always be static.

But neither seems too perf-critical to me, so I don't mind either way.

PR #2076, opened 2023-04-18
Articulated object skinning

0mdc:
## Context

To enable training of collaborative agents, we need the ability to render humans in simulation.

This changeset introduces skinned mesh support to Habitat-Sim by enabling articulated objects to render a skinned mesh that follows along its articulations.

[skinning.webm](https://user-images.githubusercontent.com/110583667/233442043-646172e6-3363-4e8c-b52e-3b08da9a8b5d.webm)

## How it works

Articulated objects are defined by URDF files, while skinned meshes are creating using glTF or FBX files.

Upon creating an articulated object, Habitat looks for a `<asset name>.ao_config.json` file, which may contain a `render_asset` field containing the skinned mesh path.

If a skinned mesh is found, the system instantiates both objects and string-matches glTF bone names with URDF link names. During this process, new child transform nodes are created for each articulated object link so that the skinned mesh can be driven.

For the system to work as desired, both URDF and glTF definitions need to share the **same bone names, dimensions and unit conventions**. Blender scripts that generates URDF files from rigged meshes will be made available for enabling quick creation of models that can satisfy these constraints.

## ao_config.json

The following parameters are added in `ao_config.json`:

- `render_asset`: Specifies the path to the GLTF model. Activates skinning and disables primitive rendering.
- `debug_render_primitivies`: Flag that turns on primitive rendering on top of skinned mesh for debugging.

ao_skinning_params
**Left:** default, **Middle:** `render_asset`, **Right:** `render_asset` + `debug_render_primitivies`

## Current limitations

- PBR shading is not implemented.
    - This is a small change that will be done after the architecture is pinned down, i.e. after this PR is merged.
- Gfx-replay not yet supported.
- Batch rendering not yet supported.
- Scaling of URDF file not yet supported.

## How has this been tested

Tested locally on `viewer.cpp` and on Habitat-Lab prototype human interactive viewer.

## Types of changes

## Checklist


src/esp/assets/RenderAssetInstanceCreationInfo.h
```diff
@@ -44,6 +49,7 @@ struct RenderAssetInstanceCreationInfo {
   Corrade::Containers::Optional<Magnum::Vector3> scale;
   Flags flags;
   std::string lightSetupKey;
+  std::shared_ptr<physics::ArticulatedObject> rig;
```
eundersander, line 52: RenderAssetInstanceCreationInfo is meant to be a lightweight object that can be easily serialized (JsonEspTypes.h), so I'm a little confused to see this member here. How does it get serialized?

eundersander (commented):
I'd love to see a unit test soon with sample data. I'm curious e.g. are you supposed to strip the per-link visual elements out of the URDF if you're also attaching a skinned model.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -183,6 +187,39 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
     shader_->bindObjectIdTexture(*(materialData_->objectIdTexture));
   }
 
+  if (skinData_) {
+    // Gather joint transformations
+    auto& skin = skinData_->skinData->skin;
+    auto& jointIdToArticulatedObjectNodes =
+        skinData_->jointIdToArticulatedObjectNode;
+    auto& transformNodes = skinData_->jointIdToTransformNode;
+    Cr::Containers::Array<Mn::Matrix4> jointTransformations{
```
eundersander, line 196: This looks like dynamic allocation. Can we just do this once when creating the GenericDrawable?

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -183,6 +187,39 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
...
+      auto jointNodeIt = transformNodes.find(skin->joints()[i]);
+      if (jointNodeIt != transformNodes.end()) {
+        jointTransformations[i] =
+            invRootTransform *
+            jointNodeIt->second->absoluteTransformationMatrix() *
+            skin->inverseBindMatrices()[i];
+        lastTransform = jointTransformations[i];
+      } else {
+        // Joint not found - use last transform.
+        jointTransformations[i] = lastTransform;
```
eundersander, line 217: Why does this failed-find happen in practice?

eundersander (commented):
I left a few more minor comments. I like that articulated objects generally don't know anything about skinned models, and skinned models don't know *that much* about articulated objects (just pointers to SceneNodes). Nice to keep these as decoupled as possible.

src/esp/gfx/GenericDrawable.cpp
0mdc: This happens if the URDF doesn't perfectly match the skinned model.
For example, it's somewhat common for skinned models to have extra leaf bones.

Our rig-to-URDF tool scripts should prevent this from happening.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2849,10 +2902,54 @@ void ResourceManager::addComponent(
...
                  computeAbsoluteAABBs,  // compute absolute aabbs
-                 staticDrawableInfo);   // a vector of static drawable info
+                 staticDrawableInfo,    // a vector of static drawable info
+                 skinData);             // instance skinning data
   }
 }  // addComponent
 
+void ResourceManager::mapSkinnedModelToArticulatedObject(
+    const MeshTransformNode& meshTransformNode,
+    const RenderAssetInstanceCreationInfo& creationInfo,
```
jturner65, line 2912: You only seem to be using the rig here, maybe just pass that pointer, instead of the entire creation info.

src/esp/assets/RenderAssetInstanceCreationInfo.h
0mdc: I agree that this is not ideal.

I put the rig there because it's a straightforward way to pass it down the resource manager with minimal changes. It does not get serialized like other fields.

I'm not sold on a solution. In the spirit of making this gfx-replayable, perhaps this could be a rig ID that would be independently registered in the resource manager, _and eventually_ match a rig representation in the batch renderer.

What do you think?

eundersander: Yeah. Let's just assume this will get revisited when you integrate skinning into gfx-replay. I'm happy for this to merge as-is.

src/esp/gfx/GenericDrawable.cpp
eundersander: In that case, can we do a loud error on startup instead of this silent failure?

src/esp/gfx/SkinData.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
```
jturner65, line 1: Instead of pragma the Facebook design guide says we should use header guards. 

Also needs the copyright

jturner65, line 1: Instead of pragma the Facebook design guide says we should use header guards. 

Also needs the copyright

jturner65 (approved):
LGTM except for the minor things with the SkinData header.  This is really cool!

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -193,7 +233,16 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
...
 void GenericDrawable::updateShader() {
-  Mn::UnsignedInt lightCount = lightSetup_->size();
+  const Mn::UnsignedInt lightCount = lightSetup_->size();
+  const Mn::UnsignedInt jointCount =
+      skinData_ ? skinData_->skinData->skin->joints().size() : 0;
+  const Mn::UnsignedInt perVertexJointCount =
+      skinData_ ? skinData_->skinData->perVertexJointCount : 0;
+
+  if (skinData_) {
+    jointTransformations_ = Cr::Containers::Array<Mn::Matrix4>{
```
eundersander, line 243: Isn't this still a dynamic allocation? Can we just resize `jointTransformations_` here instead of creating and assigning a new array? The resize will be a no-op assuming the array was already the correct size from the previous call to `updateShader`.

src/esp/gfx/SkinData.h
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @brief Stores skinning data for an instance.
+ * Contains association information of graphics bones and articulated object
+ * links.
+ */
+struct InstanceSkinData {
+  /** @brief Pointer to loaded skin data for the instance. */
+  const std::shared_ptr<SkinData>& skinData;
+  /** @brief Map between skin joint IDs and articulated object nodes. */
+  std::unordered_map<int, const scene::SceneNode*>
```
eundersander, line 33: If I'm seeing correctly, you're only using this in one place, to get `invRootTransform`. Constructing a whole map seems like overkill for that.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2849,10 +2901,53 @@ void ResourceManager::addComponent(
...
+    const std::shared_ptr<gfx::InstanceSkinData>& skinData) {
+  // Find skin joint ID that matches the node
+  const auto& gfxBoneName = meshTransformNode.name;
+  const auto& boneNameJointIdMap = skinData->skinData->boneNameJointIdMap;
+  const auto jointIt = boneNameJointIdMap.find(gfxBoneName);
+  if (jointIt != boneNameJointIdMap.end()) {
+    int jointId = jointIt->second;
+
+    // Find articulated object link ID that matches the node
+    const auto& linkIds = rig->getLinkIdsWithBase();
```
eundersander, line 2921: Doesn't `getLinkIdsWithBase` return a vector by value? Are you making a reference to a temporary?!?

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -183,6 +189,40 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
...
+      const auto jointNodeIt = transformNodes.find(skin->joints()[i]);
+      if (jointNodeIt != transformNodes.end()) {
+        jointTransformations_[i] =
+            invRootTransform *
+            jointNodeIt->second->absoluteTransformationMatrix() *
+            skin->inverseBindMatrices()[i];
+        lastTransform = jointTransformations_[i];
+      } else {
+        // Joint not found - use last transform.
+        jointTransformations_[i] = lastTransform;
```
eundersander, line 220: Ok, I understand now why this find failure isn't an error. As you say, some of the bones in the skinned model don't need to be driven from an articulated link. But why is `lastTransform` the correct thing to set it to?

eundersander (commented):
I left some more comments.

Issue #2087, opened 2023-04-29
Semantic Sensor Issue in the latest main branch #2803


Jianghanxiao:
## Habitat-Sim version
main

## 🐛 Bug

The semantic sensor for FP dataset doesn't work. This is caused by the default value of flags_ in PbrDrawable. The [judging](https://github.com/facebookresearch/habitat-sim/commit/7d04cd50c693baca3ae3779fa4ee23c8e30da034#diff-7defdaa1f6c4ee65898cbc0d7e736247f90fb928ba30735ebcd9980f7ff64c23R204) here will always return True because the flags_ is initialized with that PbrShader::Flag::ObjectId. Here we should call this one to fix "materialData_->attribute<bool>("hasPerVertexObjectId")".

## Steps to Reproduce
Any scene in FP will have the problem

jturner65:
This PR should address this issue
https://github.com/facebookresearch/habitat-sim/pull/2090

PR #2088, opened 2023-04-29
Fix the semantic sensor that are incorrect for some dataset

Jianghanxiao:
## Motivation and Context

Fix the issue in #2087 

## How Has This Been Tested

Test on FP dataset

## Types of changes

The Flags_ in PbrDrawable is initialized with ObjectId, which makes useless to judge in the shader setting

## Checklist


Jianghanxiao:
Ohhh, maybe the best way is actually to initialize the FLAG_ with a 0. What's your opinions?

Jianghanxiao:
This is better fixed in #2090

PR #2083, opened 2023-04-25
--Convert materials to use magnum materials.

jturner65:
## Motivation and Context
This PR replaces the habitat gfx::MaterialData with Magnum's material implementations. This will expand support and potential capabilities, particularly for multi-layer materials and extensions, and make improvements to the PBR rendering in particular be much more easily realized.

## How Has This Been Tested

Existing tests pass. 

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2091, opened 2023-05-01
--[PBR part 3 of 3]Map KHR Extensions to PBRDrawables and Implement in Shader

jturner65:
## Motivation and Context
This PR ~~will introduce some limited support in our existing~~ rewrites our PBR shader and extends it to support Khronos-specified materials layers (ior, specular textures, clearcoat, anisotropy)  by 
- Mapping the ClearCoat and KHR extension fields that exist in a particular material layer to the PBRDrawable cache
- Using those cached values to be written into the shader
- Using those fields in the shader for the fragment color calculation.

Tangent-space TBN frame derivation is also supported when precomputed tangents are not provided, which enables the use of normal textures.

Extensions supported so far : 
1. KHR_materials_ior : instead of using a hardcoded DielectricSpecular of 0.04 (which assumes a material IOR of 1.5), we are using material-provided IOR (if exists) to derive the DielectricSpecular.  This mainly served as proof of concept and pipeline test.
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_ior

2. KHR_materials_specular : The values in this extension consist of float and texture strength and color and texture specular colors.  These are used to modify the fresnel specular normal incidence and glancing incidence terms.
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_specular

3. KHR_materials_clearcoat : The values in this extension consist of a float and texture strength, a 3-vector and texture color and a texture-based normal.  Currently the normal support is incomplete (it uses base normal), but otherwise clearcoat is supported.
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_clearcoat

4. Generating tangents for normal textures, clearcoat normal textures and anisotropy when precomputed tangents are not provided.  Using a partial derivative calculation in the fragment shader on both position and texture coordinates, the ability to generate tangent-space TBN frame was added.  This follows the concept given here https://jcgt.org/published/0009/03/04/paper.pdf Section 3.3.

5. KHR_materials_anisotropy : The values in this extension consist of a strength value, a direction in tangent-space, and potentially a texture. These are used to give a directional bias to the specular calculation. 
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_anisotropy

## Comparison Before and After videos of PBR changes using a Floorplanner scene
A few things to note : 
 - There is no Image Based Lighting in either example.  These are intended only to showcase the differences in the direct lighting PBR calculations.
 - Both videos have the same lighting configuration.
 
Below is the current state of the PBR shader:

https://github.com/facebookresearch/habitat-sim/assets/1093953/3db68d7d-2bbf-492c-8f76-0c6e7baa62aa

And here is an example of the same scene with this PR's updated PBR shader computations

https://github.com/facebookresearch/habitat-sim/assets/1093953/8aa71ed9-9f94-483b-aa4b-afe47f79fa88

A few improvements to notice : 
1. The floor in the first example is washed out and hard to see. Compare this to the floor in the updated shader.
2. Note the shadows in interior of the the washer/dryer between the old and new shader. In the new shader they follow a more realistic interior geometry, which becomes even more apparent with IBL (see below). This improved "shadow"/normal behavior is due to the addition of **generated tangents**.
3. The picture frame by the dinner table has appropriate specularity in the new shader. This was lacking due to a small calculation bug in the original PBR shader.
4.  There are a cluster of 3 white capsule-like statues at approximately 1/3 of the way through each video.  In the current PBR shader they are flat, whereas in the new PBR shader they have a nice shine and a dark edge to them - that's the **clearcoat** layer in action.
5. In the old shader many objects are flat and dull, including the plants, furnishings in the kitchen, the ceramic cactus in the bedroom and the thermos-looking statues by the television.  Compare these with the new shader.  Many of these owe their shine to a **clearcoat** layer, and some have **specular textures.**

## New PBR with IBL
Here's that same scene lit with both the current default lights on main and the current IBL setup on main, weighted equally 50/50.

https://github.com/facebookresearch/habitat-sim/assets/1093953/734b8996-2071-43e9-8b26-7b792e81834b

## How Has This Been Tested
Existing c++ and python tests pass, with new test images where appropriate.

## Types of changes

## Checklist


jturner65:
I fully expect Simtest to fail, since I am changing the PBR shader and the core lighting. The images for a few tests will probably have to be retaken.

src/esp/gfx/PbrDrawable.h
```diff
@@ -42,8 +44,107 @@ class PbrDrawable : public Drawable {
     Mn::GL::Texture2D* roughnessTexture = nullptr;
     Mn::GL::Texture2D* metallicTexture = nullptr;
```
mosra, line 46: Er, wait, I thought the shader was using a combine roughness + metallic texture? Because that's what glTF uses, and that's why roughness is in the G channel and metalness in B.

Same is for clearcoat below (factor in R, roughness in G), glTF for some reason uses separate texture properties, however you can enforce / check for this packing with [hasLayerFactorRoughnessTexture()](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PbrClearCoatMaterialData.html#aaf4e07e8fd6af18831e271a8d0e67fbd).

src/esp/gfx/PbrShader.h
```diff
@@ -185,28 +190,91 @@ class PbrShader : public Magnum::GL::AbstractShaderProgram {
...
+    /**
+     * Has Roughness Texture in ClearCoat layer
+     */
+    CCLayer_RoughnessTexture = (1 << 15) | ClearCoatLayer,
+    /**
+     * Has Normal Texture in ClearCoat layer
+     */
+    CCLayer_NormalTexture = (1 << 16) | ClearCoatLayer,
+
+    CCLayer_NormalTextureScale = (1 << 17) | ClearCoatLayer,
```
mosra, line 213: The `NormalTextureScale` flag is useless, that feature is so cheap that it makes sense to have it enabled implicitly together with `NormalTexture` (I do exactly that in the Phong shader). Thus same here for the clear coat layer.

src/esp/gfx/PbrShader.h
```diff
@@ -185,28 +190,91 @@ class PbrShader : public Magnum::GL::AbstractShaderProgram {
...
+    // TODO : possibly break out each layer's flags to a separate flags state
+    // machine?
+    /**
+     * Has ClearCoat Texture in ClearCoat layer
+     */
+    CCLayer_CCTexture = (1 << 14) | ClearCoatLayer,
+    /**
+     * Has Roughness Texture in ClearCoat layer
+     */
+    CCLayer_RoughnessTexture = (1 << 15) | ClearCoatLayer,
```
mosra, line 207: These could be merged into one, `ClearCoatFactorRoughnessTexture`, as per my suggestion with the swizzles above.

Similarly for the code that was already there, it doesn't make sense to have separate `MetallicTexture` and `RoughnessTexture` as there's already `NoneRoughnessMetallicTexture`. I doubt these variants were ever sufficiently tested given there's basically no assets that wouldn't have a combined R/M texture, so just drop them.

src/esp/gfx/PbrShader.h
```diff
@@ -185,28 +190,91 @@ class PbrShader : public Magnum::GL::AbstractShaderProgram {
...
+
+    /**
+     * Has Specular Texture in KHR_materials_specular layer
+     */
+    SpecLayer_SpecTexture = (1 << 19) | SpecularLayer,
+
+    /**
+     * Has Specular Color Texture in KHR_materials_specular layer
+     */
+    SpecLayer_SpecColorTexture = (1 << 20) | SpecularLayer,
```
mosra, line 230: Again, as [the extension states](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md), these two are commonly packed into a single texture, so you could treat them as one.

Unfortunately, as there's no builtin material support in Magnum for `KHR_materials_specular`, there's no easy check if the material indeed has such packing, so you have to do that by hand for now.

src/esp/gfx/PbrShader.h
```diff
@@ -185,28 +190,91 @@ class PbrShader : public Magnum::GL::AbstractShaderProgram {
      * will use the node's semantic ID
      */
     InstancedObjectId = (1 << 12) | ObjectId,
+
+    /**
+     * Has ClearCoat layer.
+     */
+    ClearCoatLayer = 1 << 13,
+    // TODO : possibly break out each layer's flags to a separate flags state
+    // machine?
```
mosra, line 199: Just name them nicer, and no shorthands ;) With the suggested variant reduction I think it's not too bad.

Also I wouldn't add everyting under the sun, just what _actually_ gets implemented in the shader in the end. E.g., for `KHR_materials_transmission` there's not even proper support for correct depth-ordered transparency yet, and additionally `KHR_materials_volume` would need to be implemented with some kind of raytracing. It'll make sense to implement those once Habitat gets _some_ transparency support at least.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2088,20 +2088,23 @@ namespace {
...
+                                  std::string& debugStr) {
   switch (typeToCheck) {
     case ObjectInstanceShaderType::Phong: {
       bool compRes =
           bool(materialData.types() & Mn::Trade::MaterialType::Phong);
-      ESP_DEBUG() << "Forcing to Phong | Material currently"
-                  << (compRes ? "supports" : "does not support") << "Phong";
+      Cr::Utility::formatInto(debugStr, debugStr.size(),
+                              "Forcing to Phong | Material currently {} Phong",
+                              (compRes ? "supports" : "does not support"));
```
mosra, line 2099: I hope you're aware of this, but just in case -- if debug output is disabled, the whole `ESP_DEBUG()` expression (and similar) is a no-op, so as cheap as if you wouldn't print anything at all. We designed them like that with Erik Wijmans to make sure they're not adding extra overhead when the output isn't wanted.

However, what you do here (and what seems to be a common pattern in your code) is _unconditionally_ appending a formatted string to somewhere, which will be done always, allocating memory that's then possibly thrown away at the end. A whole different performance characteristic, and possibly a significant contribution to the overall slowness given how extremely verbose Habitat's logging became over the years.

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -142,6 +144,158 @@ void PbrDrawable::setMaterialValuesInternal(
...
+    if (materialData_->hasAttribute("#KHR_materials_ior", "ior")) {
+      matCache.ior_Index =
+          materialData_->attribute<float>("#KHR_materials_ior", "ior");
+    }
+
+  }  // has KHR_materials_ior layer
+
+  ////////////////
+  // KHR_materials_specular layer
+  if (materialData_->hasLayer("#KHR_materials_specular")) {
```
mosra, line 200: While material attribute lookup by name is logarithmic, layer lookup by name is linear, so with each of those you have a `O(m + log n)`. I'll document that better, it doesn't seem to be stated clearly enough. There's [findLayerId()](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1MaterialData.html#a1845070c3a94b31a749e318cb16bc887), use it like this:

```cpp
if(const Containers::Optional<UnsignedInt> specularLayerId = 
    materialData_->findLayerId("#KHR_materials_specular")) 
{
    materialData->attribute(specularLayerId, "...");
    ...
}    
```

And `hasAttribute()` and `attribute()` is also an unnecessary double lookup, so you can use the same pattern with `findAttributeId()` `findAttribute()`, `attributeOr()` etc. -- read the docs. It's also *significantly* less typing that way!

```cpp
if(const Containers::Optional<Float> specularFactor = 
    materialData_->findAttribute<Float>(specularLayerId, "specularFactor")) 
{
    matCache.spec_SpecularFactor = specularFactor;
}
```

This might apply also to the new code in https://github.com/facebookresearch/habitat-sim/pull/2083, I unfortunately didn't have the bandwidth to check everything there as well.

src/esp/assets/ResourceManager.cpp
jturner65: Yes, I am aware of this, and I do understand the cost overhead of unconditionally constructing the string. However, I felt the overhead was worth it, at least now while in active development, in order to minimize the number of lines of debug spam.  Would be great if we could make the debug streams able to be appended without newlines.

mosra: [Flag::NoNewlineAtTheEnd](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html#aa346e932b5141006c49777ddacfaf600af29e1bcd00663e0b5d5a74d31baa9e96) passed to `ESP_DEBUG()`? All flags that Corrade's `Debug` constructor accepts the macro can as well.

jturner65: So as long as all the debug messages in a sequence are semantically related, I can use that flag to create a coherent informational line from multiple sequential sources? I like that, thanks!

jturner65: Unfortunately that also prepends the informational header to each separate stream call (timestamp, subsystem, file, etc).  Looks like excluding that redundant information would require some extra logging development.  I'll put that on my plate to investigate (proper support of no newlines in logging).

src/esp/gfx/LightSetup.cpp
```diff
@@ -76,12 +76,34 @@ LightSetup getLightsAtBoxCorners(const Magnum::Range3D& box,
...
+       {0.6, 0.6, 0.6},
+       LightPositionModel::Global},  // downward
+
+      // camera-relative
+      {{0.0, -0.1, -1.0, 0.0},
+       {0.3, 0.3, 0.3},
+       LightPositionModel::Camera},  // forward
+      {{0.0, -0.3, 1.0, 0.0},
+       {0.5, 0.5, 0.5},
+       LightPositionModel::Camera},  // backward (for glancing speculars)
```
0mdc, line 104: I would avoid emitting light from the camera by default.

src/shaders/pbrStructs.glsl
```diff
@@ -0,0 +1,240 @@
...
+                   // anisotropyRotation provides additional rotation to
+                   // the vectors in the texture.
+};
+
+#if (LIGHT_COUNT > 0)
+
+//////////////////////////
+// Structure holding per-light anisotropy-related lighting info to faciliate
+// passing as function args
+struct AnistropyDirectLight {
```
0mdc, line 73: Typo: Anistropy

src/shaders/pbrStructs.glsl
```diff
@@ -0,0 +1,240 @@
...
+  // diffuse color used by calculations
+  vec3 diffuseColor;
+  // Achromatic dielectric material f0 : fresnel reflectance at normal
+  // incidence based on given IOR. Specular layer data modfies this
+  vec3 dielectric_f0;
+  // dielextric/metallic f0 : fresnel reflectance color at normal
+  // incidence based on given IOR. Specular layer data modfies this
+  vec3 specularColor_f0;
+  // glancing incidence dielectric/metallic reflectance. Specular layer data
+  // modfies this
```
0mdc, line 153: Nit: mod`i`fies

src/shaders/pbrUniforms.glsl
```diff
@@ -0,0 +1,152 @@
...
+#endif
+
+// -------------- uniforms ----------------
+#if defined(OBJECT_ID)
+uniform highp uint uObjectId;
+#endif
+
+#if defined(NORMAL_TEXTURE)
+uniform mediump float uNormalTextureScale
+#ifndef GL_ES
```
0mdc, line 27: I'm unsure here, but I think that we can omit these `#ifndef GL_ES`.

src/esp/gfx/PbrShader.cpp
```diff
@@ -529,7 +774,8 @@ PbrShader& PbrShader::setLightColor(unsigned int lightIndex,
       lightIndex < lightCount_,
       "PbrShader::setLightColor: lightIndex" << lightIndex << "is illegal.",
       *this);
-  Mn::Vector3 finalColor = intensity * PBR_LIGHT_SCALE * color;
+  Mn::Vector3 finalColor = intensity * color;
+  // finalColor *= PBR_LIGHT_SCALE;
```
0mdc, line 778: We probably want to remove this here too.

0mdc (commented):
Nice work! I had a first look and left some minor comments.

src/esp/gfx/PbrTextureUnit.h
```diff
@@ -9,21 +9,30 @@
...
   Emissive = 3,
-  EnvironmentMap = 4,
-  IrradianceMap = 5,
-  BrdfLUT = 6,
-  PrefilteredMap = 7,
-  ShadowMap0 = 8,
-  ShadowMap1 = 9,
-  ShadowMap2 = 10,  // We support at most 3 shadows currently
+  ClearCoatFactor = 4,
+  ClearCoatRoughenss = 5,
```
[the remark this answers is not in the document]
jturner65: Not sure what you mean.  Clearcoat Roughness is available as a texture as well as a material-wide float value.

src/shaders/pbrBSDF.glsl
```diff
@@ -0,0 +1,221 @@
+// Copyright (c) Meta Platforms, Inc. and its affiliates.
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
```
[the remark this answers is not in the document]
jturner65: why would we not do that?

src/esp/gfx/PbrTextureUnit.h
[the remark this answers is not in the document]
0mdc: I meant: Rough`e`nss

src/shaders/pbrMaterials.glsl
```diff
@@ -145,7 +145,9 @@ PBRData buildPBRData() {
   pbrInfo.view = normalize(uCameraWorldPos - position);
 
   // cos angle between view and normal
-  pbrInfo.n_dot_v = abs(dot(pbrInfo.n, pbrInfo.view));
+  // clamp to ensure range adherence, abs instead of epsilon to avoid errors at
+  // clancing angles
+  pbrInfo.n_dot_v = clamp(dot(pbrInfo.n, pbrInfo.view), 0.0, 1.0);
```
jturner65, line 150: I was experimenting with using the absolute value of the normal dot view instead of the clamp, and forgot to revert it. This in general looked just as good as the clamp, except in certain situations where it seems negative normal dot view results were indicated from a texture to give the object some kind of depth. In this case, the abs(n_dot_v) ended up looking weird.

First, an example of negative n_dot_v - frags with negative n_dot_v are the brightly colored teal->yellow, with the more yellow having the higher abs magnitude


ndotvNegative

Here's what this looks like if we just leave the negative n_dot_v  : 

ndotvNegNoGreen

Here's what the abs(n_dot_v) yielded - more character/variation in the area with the negative n_dot_v, but doesn't look quite right, spatially.

absNdotV

And here's the clamped negative n_dot_v to [0,1]  - the inner area is a bit ambiguous, but it at least looks appropriate, geometrically.

clampedNdotV

I don't clamp the normal to epsilon because that introduced artifacts around the edges. There are guards against zero n_dot_v where they would be needed.

jturner65: I thought it would be appropriate to chronicle one of the fundamental design decisions here (how to precondition n_dot_v) .

src/esp/gfx/PbrTextureUnit.h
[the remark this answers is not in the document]
jturner65: Ooops! thanks! Didn't even see it when you pointed it out. -sigh-

src/shaders/pbrBSDF.glsl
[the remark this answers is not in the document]
jturner65: Everywhere that I was influenced by existing code I left links for attribution.  There really isn't any copy-pasting of anything although certain implementations are very similar. Most of the functionality was developed sufficiently long ago (2012-2013) that it's kind of ubiquitous, in my limited experience, among these kinds of shaders.

The one exception to this is the partial-derivative-driven derivation of TBN.  This was from a paper in 2020.  I give reference to the original code, and to a newer, simpler implementation.

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -419,6 +424,51 @@ void PbrDrawable::draw(const Mn::Matrix4& transformationMatrix,
     shader_->setTextureMatrix(matCache.textureMatrix);
   }
 
+  // clearcoat data
+  if (flags_ & PbrShader::Flag::ClearCoatLayer) {
```
mosra, line 428: Maybe just change all flag `&`s to `>=`? Unnecessary inconsistency that makes the reader look twice, and prone to accidents if some of the flags would be multiple bits.

src/esp/gfx/PbrShader.h
```diff
@@ -496,7 +582,7 @@ class PbrShader : public Magnum::GL::AbstractShaderProgram {
    */
   PbrShader& setLightColor(unsigned int lightIndex,
                            const Magnum::Vector3& color,
-                           float intensity = 1.0);
+                           float intensity = 1);
```
mosra, line 585: :eyes: 

`1.0f` maybe?

src/esp/gfx/PbrTextureUnit.h
```diff
@@ -9,21 +9,30 @@
 #include <Corrade/Containers/EnumSet.h>
 
 namespace esp {
+typedef unsigned char uint8_t;
```
mosra, line 12: :scream: why? 

This is a standard type, `#include <cstdint>`.

src/esp/gfx/PbrShader.cpp
```diff
@@ -129,6 +170,11 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
                      ? rs.getString("shadowsVSM.glsl") + "\n"
                      : "")
       .addSource(rs.getString("pbrCommon.glsl") + "\n")
```
mosra, line 172: The `\n` shouldn't be necessary for the files anywhere, the files themselves should have a trailing newline. Git itself warns if they don't and I think clang-format even enforces that.

Also, the append of a `"\n"` will cause an unnecessary string allocation & copy, while without it'd be a zero-copy operation as the source string view is constant global memory.

src/esp/gfx/PbrShader.cpp
```diff
@@ -113,14 +126,42 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
...
+      .addSource(flags_ >= Flag::AnisotropyLayerTexture
+                     ? "#define ANISOTROPY_LAYER_TEXTURE\n"
+                     : "")
+
       .addSource(flags_ & Flag::PrecomputedTangent
                      ? "#define PRECOMPUTED_TANGENT\n"
                      : "")
       .addSource(flags_ & Flag::ImageBasedLighting
-                     ? "#define IMAGE_BASED_LIGHTING\n"
+                     ? "#define IMAGE_BASED_LIGHTING\n#define TONE_MAP\n"
```
mosra, line 162:
```suggestion
                     ? "#define IMAGE_BASED_LIGHTING\n"
                       "#define TONE_MAP\n"
```

feels more readable to me. Question is whether clang-format has the same opinion tho.

src/esp/gfx/PbrShader.cpp
```diff
@@ -55,6 +51,26 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
...
+              Flag::NormalTexture | Flag::EmissiveTexture))) ||
+       // clear coat
+       ((flags_ >= Flag::ClearCoatTexture) ||
+        (flags_ >= Flag::ClearCoatRoughnessTexture) ||
+        (flags_ >= Flag::ClearCoatNormalTexture)) ||
+       // specular layer
+       ((flags_ >= Flag::SpecularLayerTexture) ||
+        (flags_ >= Flag::SpecularLayerColorTexture)) ||
+       // anisotropy - always needs texCoords to get tangentspace map
+       (bool(flags_ & Flag::AnisotropyLayer)) ||
```
mosra, line 68: Same here, just use `>=` unconditionally, looks more consistent.

src/esp/gfx/Drawable.h
```diff
@@ -130,12 +130,13 @@ class Drawable : public Magnum::SceneGraph::Drawable3D {
...
+   * @param reset whether to reset underlying material or to write over it
    */
-  virtual void setMaterialValues(
-      CORRADE_UNUSED const Magnum::Resource<Magnum::Trade::MaterialData,
-                                            Magnum::Trade::MaterialData>&
-          material) {}
+  virtual void setMaterialValues(CORRADE_UNUSED const Magnum::Resource<
+                                     Magnum::Trade::MaterialData,
+                                     Magnum::Trade::MaterialData>& material,
+                                 CORRADE_UNUSED bool reset) {}
```
mosra, line 139: I don't like boolean argument in public APIs, it's unclear what is it supposed to mean unless you go and look at the definiton. E.g., is it "not setting the material values"? Is it a "dry run"? Etc.. 

Can this be done in some other way? As far as I see, in both the drawable implementations it's delegating to a private `setMaterialValuesInternal()` (for which a bool would be fine, as it's a private API). So having something like this -- for example -- would make sense I think:

```cpp
public:
    void setMaterialValues(material) {
        setMaterialValueInternal(Resource material, false);
    }
    void resetMaterialValues(material) {
        setMaterialValueInternal(Resource material, true);
    }
   
private:
    virtual void setMaterialValuesInternal(Resource material, bool reset);
```

Related useful reading: http://www.gotw.ca/publications/mill18.htm

src/esp/assets/ResourceManager.cpp
```diff
@@ -2445,70 +2441,76 @@ void ResourceManager::loadMaterials(Importer& importer,
...
+
         materialData = createUniversalMaterial(*materialData);
       }
 
       // This material data has any per-shader as well as global custom
       // user-defined attributes set excluding texture pointer mappings
       Corrade::Containers::Optional<Magnum::Trade::MaterialData>
           custMaterialData;
-
+      std::string shaderBeingUsed;
```
mosra, line 2469: Can be a StringView as it's always set to a literal.

mosra (commented):
Didn't go through the actual shader algorithms, but the GLSL code itself looks alright.

src/esp/gfx/PbrTextureUnit.h
jturner65: I was getting a warning and I thought this was better than an include :D

src/esp/gfx/PbrShader.cpp
jturner65: I'll find out :D

src/esp/gfx/Drawable.h
jturner65: Sure, I like that

src/esp/gfx/PbrTextureUnit.h
mosra: I don't think aliasing `uint8_t` / `std::uint8_t` with `esp::uint8_t` is a good idea.

Skylion007 (approved):

aclegg3 (approved):
LGTM, great step toward better rendering. :+1:

PR #2094, opened 2023-05-03
--[BUGFIX] Reverse directional lights before sending to uniform;

jturner65:
A long standing issue/bug with Habitat is that our directional lights have always been backwards - pointing away from the direction they are intended to be pointing toward.  This PR fixes this issue, without removing the performance benefit the backwards direction gave (in the phong and pbr shaders).  

The existing light setups using directional lights have been corrected in this PR, but any datasets that provide directional lighting will need to have those lights reversed, so this should probably be considered a breaking change.

## How Has This Been Tested
All existing c++ and python tests pass.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/GenericDrawable.cpp
+++ b/src/esp/gfx/GenericDrawable.cpp
@@ -144,7 +144,7 @@ void GenericDrawable::updateShaderLightingParameters(
     // flip directional lights to faciliate faster, non-forking calc in
     // shader.  Leave non-directional lights unchanged
     pos *= (pos[3] * 2) - 1;
-    lightPositions.emplace_back(std::move(pos));
+    lightPositions.emplace_back(pos);
 
     const auto& lightColor = (*lightSetup_)[i].color;
     lightColors.emplace_back(lightColor);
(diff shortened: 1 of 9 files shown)


Skylion007 (approved):

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -139,8 +139,12 @@ void GenericDrawable::updateShaderLightingParameters(
...
   for (Mn::UnsignedInt i = 0; i < lightSetup_->size(); ++i) {
     const auto& lightInfo = (*lightSetup_)[i];
-    lightPositions.emplace_back(getLightPositionRelativeToCamera(
-        lightInfo, transformationMatrix, cameraMatrix));
+    Mn::Vector4 pos = getLightPositionRelativeToCamera(
+        lightInfo, transformationMatrix, cameraMatrix);
+    // flip directional lights to faciliate faster, non-forking calc in
+    // shader.  Leave non-directional lights unchanged
+    pos *= (pos[3] * 2) - 1;
+    lightPositions.emplace_back(pos);
```
Skylion007, line 147:
```suggestion
    lightPositions.emplace_back(std::move(pos));
```

Skylion007: Oh whoops, I think it actually has a trivial copy ctor. It is some  ofthe Eigen classes are movable.

PR #2097, opened 2023-05-04
enable recompute_navmesh when creating sim with create_renderer==false

eundersander:
## Motivation and Context

The navmesh is needed for blind agents, batch rendering, and other use cases where the sim is created with `create_renderer==false`.

Note a potential source of confusion: when `create_renderer==false`, we don't load textures, upload mesh vertices to the GPU, or create "GL meshes". However, we still load mesh data from disk, so we can still generate a navmesh on the CPU.

## How Has This Been Tested

Local testing on 2019 Macbook Pro

## Types of changes

## Checklist


0mdc (approved):

eundersander:
Merging this now as SIRo has an urgent need, but @aclegg3 I'd still like your thoughts when you get a chance!

Issue #2095, opened 2023-05-03
Physics simulation is unstable and different from that of pybullet


rishiagarwal2000:
## Habitat-Sim version
main

## 🐛 Bug

I tried simulating a urdf from [here](https://github.com/Stanford-TML/circle_dataset/tree/main/src/viewer#:~:text=scene%20and%20the%20URDF%20skeletons) in habitat-sim and pybullet. The outputs for the simulation are attached. Apparently, the simulations differ significantly which seems unexpected to me because habitat-sim uses the same pybullet simulation internally. Also, the simulation is highly unstable. I am not sure if this is a bug, or if I am using it incorrectly. It would be great if you could help me with this.

## Steps to Reproduce

Pybullet snippet
```python
import pybullet as p
import time
import pybullet_data
from scipy.spatial.transform import Rotation as R
import sys
import imageio

physicsClient = p.connect(p.GUI)#or p.DIRECT for non-graphical version
p.setAdditionalSearchPath(pybullet_data.getDataPath()) #optionally
p.setGravity(0,0,-9.8)
planeId = p.loadURDF("plane.urdf")
startPos = [0,0,1.1]
startOrientation = p.getQuaternionFromEuler([0,0,0])

humanId = p.loadURDF("CIRCLE_assets/subjects/2.urdf", startPos, startOrientation) #R.from_euler('xyz', [0, 0, 0], degrees=True).as_quat())
jointIndices = [i for i in range(p.getNumJoints(humanId)) if p.getJointInfo(humanId, i)[2] == p.JOINT_SPHERICAL]
print([p.changeDynamics(humanId, i, linearDamping=0, angularDamping=0, jointDamping=0.0) for i in jointIndices])

print("starting sim loop")

for i in range (100):
    p.stepSimulation()
    time.sleep(1./240.)
    
p.disconnect()
```

Habitat-sim snippet
```python
import habitat_sim
from habitat_sim.logging import logger
from settings import default_sim_settings, make_cfg

import habitat_sim.physics as phy
from habitat_sim.utils.common import quat_from_magnum, quat_to_magnum
from scipy.spatial.transform import Rotation as R

import magnum as mn
import numpy as np
from typing import Any, Callable, Dict, List, Optional, Tuple
import pathlib
import json

import os
import imageio

def global_correction_quat(up_v: mn.Vector3, forward_v: mn.Vector3) -> mn.Quaternion:
    """
[...]

PR #2098, opened 2023-05-04
Add intertiaFromURDF option to URDF loading API

aclegg3:
## Motivation and Context

As discussed in issue #2095 some URDF files behave poorly when inertia is computed automatically from collision shapes. This PR adds the API option to parse inertia matrices directly from the URDF file instead.

Also the viewer URDF loader now creates JointMotors by default.

## How Has This Been Tested

In the viewer app (as shown in the linked issue)

## Types of changes

## Checklist

### DIFF
--- a/examples/viewer.py
+++ b/examples/viewer.py
@@ -560,13 +560,31 @@ class HabitatSimInteractiveViewer(Application):
                 self.cached_urdf = urdf_file_path
                 aom = self.sim.get_articulated_object_manager()
                 ao = aom.add_articulated_object_from_urdf(
-                    urdf_file_path, fixed_base, 1.0, 1.0, True
+                    urdf_file_path,
+                    fixed_base,
+                    1.0,
+                    1.0,
+                    True,
+                    maintain_link_order=False,
+                    intertia_from_urdf=False,
                 )
                 ao.translation = (
                     self.default_agent.scene_node.transformation.transform_point(
                         [0.0, 1.0, -1.5]
                     )
                 )
+                # check removal and auto-creation
+                joint_motor_settings = habitat_sim.physics.JointMotorSettings(
+                    position_target=0.0,
+                    position_gain=1.0,
+                    velocity_target=0.0,
+                    velocity_gain=1.0,
+                    max_impulse=1000.0,
+                )
+                existing_motor_ids = ao.existing_joint_motor_ids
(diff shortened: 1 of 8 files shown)


Skylion007 (approved):

jturner65 (approved):
LGTM! One thing, any python calls to "add_articulated_object_from_urdf" that include a lighting key as a positional argument will have to be refactored after this hits to use keyword mapping.

src/esp/physics/objectManagers/ArticulatedObjectManager.h
```diff
@@ -121,6 +129,7 @@ class ArticulatedObjectManager
       float massScale = 1.0,
       bool forceReload = false,
       bool maintainLinkOrder = false,
+      bool intertiaFromURDF = false,
```
0mdc, line 132: Would it be valuable to specify this flag in the content (via attributes) instead of the parameter?

examples/viewer.py
```diff
@@ -560,13 +560,31 @@ def key_press_event(self, event: Application.KeyEvent) -> None:
...
+                    position_target=0.0,
+                    position_gain=1.0,
+                    velocity_target=0.0,
+                    velocity_gain=1.0,
+                    max_impulse=1000.0,
+                )
+                existing_motor_ids = ao.existing_joint_motor_ids
+                for motor_id in existing_motor_ids:
+                    ao.remove_joint_motor(motor_id)
+                ao.create_all_motors(joint_motor_settings)
```
0mdc, line 587: It would be worthwhile to create a test that asserts the URDF asset pipeline.

0mdc (approved):
LGTM! I added a few general comments.

examples/viewer.py
aclegg3: Check out test_physics from [this line](https://github.com/facebookresearch/habitat-sim/blob/f9ff5407038c923e6a89e004d9681d744bda31ce/tests/test_physics.py#L794) on. There could be more, but this does all the basics I think.

aclegg3:
> LGTM! One thing, I am pretty sure any python calls to "add_articulated_object_from_urdf" that include a lighting key as a positional argument will have to be refactored after this hits to use keyword mapping.

True, I'll mark this as a breaking change for that reason.
Doesn't look like Habitat-lab `main` ever calls this with a light setup (which should be rare).

PR #2102, opened 2023-05-10
--[BUGFIX] Single channel texture support

jturner65:
## Motivation and Context
We were not handling single channel textures, causing them to display red.  This PR addresses this - any single channel textures get their swizzle modified to pull RGB from the R channel.  This currently is not supported in magnum for WebGL builds, so a warning is displayed instead.

## How Has This Been Tested
C++ and python Tests pass locally

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -2961,15 +2961,27 @@ void ResourceManager::loadTextures(Importer& importer,
         if (image->isCompressed()) {
           format = Mn::GL::textureFormat(image->compressedFormat());
         } else {
-          format = Mn::GL::textureFormat(image->format());
+          const auto pixelFormat = image->format();
+          format = Mn::GL::textureFormat(pixelFormat);
           // Modify swizzle for single channel textures so that they are
           // greyscale
-          if (pixelFormatSize(image->format()) == 1) {
+          if (pixelFormatChannelCount(pixelFormat) == 1) {
 #ifdef MAGNUM_TARGET_WEBGL
-            ESP_WARNING() << "Single Channel Texture ID" << iTexture
-                          << "Loaded as RGB";
+            ESP_WARNING() << "Single Channel Greyscale Texture : ID" << iTexture
+                          << "incorrecting displays as red instead of "
+                             "greyscale due to greyscale expansion"
+                             "not yet supported in WebGL.";
 #else
             currentTexture->setSwizzle<'r', 'r', 'r', '1'>();
+#endif
+          } else if (pixelFormatChannelCount(pixelFormat) == 2) {
+#ifdef MAGNUM_TARGET_WEBGL
+            ESP_WARNING() << "Two Channel Greyscale + Alpha Texture : ID"
+                          << iTexture
+                          << "incorrecting displays due to greyscale expansion"
+                             "not yet supported in WebGL.";
+#else
+            currentTexture->setSwizzle<'r', 'r', 'r', 'g'>();
 #endif
           }
         }
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -2962,6 +2962,16 @@ void ResourceManager::loadTextures(Importer& importer,
           format = Mn::GL::textureFormat(image->compressedFormat());
         } else {
           format = Mn::GL::textureFormat(image->format());
+          // Modify swizzle for single channel textures so that they are
+          // greyscale
+          if (pixelFormatSize(image->format()) == 1) {
+#ifdef MAGNUM_TARGET_WEBGL
+            ESP_WARNING() << "Single Channel Texture ID" << iTexture
+                          << "Loaded as RGB";
+#else
+            currentTexture->setSwizzle<'r', 'r', 'r', '1'>();
+#endif
+          }
         }
 
         // For the very first level, allocate the texture
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -2965,21 +2965,22 @@ void ResourceManager::loadTextures(Importer& importer,
           format = Mn::GL::textureFormat(pixelFormat);
           // Modify swizzle for single channel textures so that they are
           // greyscale
-          if (pixelFormatChannelCount(pixelFormat) == 1) {
+          Mn::UnsignedInt pixelCount = pixelFormatChannelCount(pixelFormat);
+          if (pixelCount == 1) {
 #ifdef MAGNUM_TARGET_WEBGL
             ESP_WARNING() << "Single Channel Greyscale Texture : ID" << iTexture
-                          << "incorrecting displays as red instead of "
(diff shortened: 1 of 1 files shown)


0mdc (approved):
LGTM!

src/esp/assets/ResourceManager.cpp
```diff
@@ -2962,6 +2962,16 @@ void ResourceManager::loadTextures(Importer& importer,
           format = Mn::GL::textureFormat(image->compressedFormat());
         } else {
           format = Mn::GL::textureFormat(image->format());
+          // Modify swizzle for single channel textures so that they are
+          // greyscale
+          if (pixelFormatSize(image->format()) == 1) {
```
mosra, line 2967: Not the right API, there's also 16-bit formats, half-floats, ... which can all be single-channel too. You're looking for [pixelFormatChannelCount()](https://doc.magnum.graphics/magnum/namespaceMagnum.html#ab6e2600dba325d29743fad354e60b1e2) instead.

mosra (changes requested):

src/esp/assets/ResourceManager.cpp
```diff
@@ -2962,6 +2962,16 @@ void ResourceManager::loadTextures(Importer& importer,
...
+#ifdef MAGNUM_TARGET_WEBGL
+            ESP_WARNING() << "Single Channel Texture ID" << iTexture
+                          << "Loaded as RGB";
```
mosra, line 2970: This isn't really saying what it will result in. It should mention something about the texture being wrongly red because greyscale expansion isn't implemented here.

As I said on Slack, on WebGL this is doable (you'd do the channel expansion CPU side) but needs a tiny helper to  invent the corresponding expanded pixel format enum value. I'm in the process of adding that.

src/esp/assets/ResourceManager.cpp
mosra: Oh, and you may want to handle two-channel images as well, which usually represent greyscale + alpha, and thus should be expanded to RRRG.

jturner65: Thanks!

src/esp/assets/ResourceManager.cpp
```diff
@@ -2961,15 +2961,27 @@ void ResourceManager::loadTextures(Importer& importer,
         if (image->isCompressed()) {
           format = Mn::GL::textureFormat(image->compressedFormat());
         } else {
-          format = Mn::GL::textureFormat(image->format());
+          const auto pixelFormat = image->format();
+          format = Mn::GL::textureFormat(pixelFormat);
           // Modify swizzle for single channel textures so that they are
           // greyscale
-          if (pixelFormatSize(image->format()) == 1) {
+          if (pixelFormatChannelCount(pixelFormat) == 1) {
```
mosra, line 2968: The pixel format property queries are relatively large switches internally, so you may want to cache the result instead of querying twice.

src/esp/assets/ResourceManager.cpp
```diff
@@ -2961,15 +2961,27 @@ void ResourceManager::loadTextures(Importer& importer,
...
 #ifdef MAGNUM_TARGET_WEBGL
-            ESP_WARNING() << "Single Channel Texture ID" << iTexture
-                          << "Loaded as RGB";
+            ESP_WARNING() << "Single Channel Greyscale Texture : ID" << iTexture
+                          << "incorrecting displays as red instead of "
```
mosra, line 2971:
```suggestion
                          << "incorrectly displays as red instead of "
```

same below

src/esp/assets/ResourceManager.cpp
```diff
@@ -2961,15 +2961,27 @@ void ResourceManager::loadTextures(Importer& importer,
...
+                             "greyscale due to greyscale expansion"
+                             "not yet supported in WebGL.";
```
mosra, line 2973:
```suggestion
                             "not yet implemented in WebGL.";
```

mosra, line 2973:
```suggestion
                             "not yet implemented in WebGL.";
```

src/esp/assets/ResourceManager.cpp
jturner65: "incorrecting" wow. I was tired. heh.

Skylion007:
Odd we are getting a physics flake, does the tolerances need to be adjusted?

jturner65:
> Odd we are getting a physics flake, does the tolerances need to be adjusted?

Yeah, i was wondering if that's just one of the borky things going on with ci this morning of if its actually something else.

Skylion007 (approved):

src/esp/assets/ResourceManager.cpp
```diff
@@ -2961,7 +2961,30 @@ void ResourceManager::loadTextures(Importer& importer,
...
+          // greyscale
+          Mn::UnsignedInt channelCount = pixelFormatChannelCount(pixelFormat);
+          if (channelCount == 1) {
+#ifdef MAGNUM_TARGET_WEBGL
+            ESP_WARNING() << "Single Channel Greyscale Texture : ID" << iTexture
+                          << "incorrectly displays as red instead of "
+                             "greyscale due to greyscale expansion"
+                             "not yet implemented in WebGL.";
+#else
+            currentTexture->setSwizzle<'r', 'r', 'r', '1'>();
```
eundersander, line 2976: Seems reasonable, but also odd that we need to do this ourselves in ResourceManager. If this is universally correct, it would be nice if Magnum did it automatically @mosra . Related: will we need a similar fixup for the batch renderer?

mosra: Magnum doesn't do this automatically as it's a choice between:

1. having full control over the input data:
    - either offline preprocessing the JPEG/PNG images to never be single-channel,
    - or for example using the WebM format that is always RGB(A)
    - or by converting the images to a RGB(A) block-compressed format (in other words, Basis-compressed assets won't have this problem)
2. configuring a particular image import plugin to [expand single-channel images to RGB](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StbImageImporter.html#Trade-StbImageImporter-configuration)
3. performing a CPU-side expansion on load time (such as broadcasting the red channel to RRR using [Utility::copy()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#a6cd0c04124a0d7c4879d4cf6da3556a2))
4. using the `ARB_texture_swizzle` extension where supported (as done here)

As a middleware it provides the above four options, but it's on the application developer to pick the one that suits given use case best. On my task list there's some `TextureTools::expandGrayscale()` utility that would make option 3 more convenient, but it's doable with the lower-level APIs as well.

For the batch renderer this fixup is [done in the compositor scripts already](https://github.com/facebookresearch/habitat-sim/blob/07d0b3069068446c886cccb9fae8cffaf0783f2f/src/utils/compositor/CompositorState.cpp#L63). At runtime all images are stored in a (block-compressed) RGBA texture atlas, so this problem isn't there.

mosra (approved):

PR #2105, opened 2023-05-15
additional magnum importer plugins in python build dependencies

aclegg3:
## Motivation and Context

Python importer used in Habitat-lab requires linked Magnum plugins from Habitat=sim.

## How Has This Been Tested

NA

## Types of changes

## Checklist

### DIFF
--- a/src/cmake/dependencies.cmake
+++ b/src/cmake/dependencies.cmake
@@ -350,14 +350,33 @@ if(NOT USE_SYSTEM_MAGNUM)
     # for Python as well; and reset that back to strange build procedures that
     # turn some features off again later can still work.
     if(BUILD_GUI_VIEWERS)
-      set(MAGNUM_PYTHON_BINDINGS_STATIC_PLUGINS
-          MagnumPlugins::StbTrueTypeFont Magnum::AnySceneImporter
-          MagnumPlugins::AssimpImporter CACHE STRING "" FORCE
+      set(
+        MAGNUM_PYTHON_BINDINGS_STATIC_PLUGINS
+        MagnumPlugins::StbTrueTypeFont
+        Magnum::AnySceneImporter
+        Magnum::AnyImageImporter
+        Magnum::AnyImageConverter
+        MagnumPlugins::StbImageImporter
+        MagnumPlugins::StbImageConverter
+        MagnumPlugins::BasisImporter
+        MagnumPlugins::StanfordImporter
+        MagnumPlugins::GltfImporter
+        MagnumPlugins::AssimpImporter
+        CACHE STRING "" FORCE
       )
     else()
-      set(MAGNUM_PYTHON_BINDINGS_STATIC_PLUGINS Magnum::AnySceneImporter
-                                                MagnumPlugins::AssimpImporter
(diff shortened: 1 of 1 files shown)


src/cmake/dependencies.cmake
```diff
@@ -350,14 +350,33 @@ if(NOT USE_SYSTEM_MAGNUM)
...
+        Magnum::AnySceneImporter
+        MagnumPlugins::AssimpImporter
+        Magnum::AnyImageImporter
+        Magnum::AnyImageConverter
+        MagnumPlugins::StbImageImporter
+        MagnumPlugins::StbImageConverter
+        MagnumPlugins::BasisImporter
+        MagnumPlugins::StanfordImporter
+        MagnumPlugins::GltfImporter
+        CACHE STRING "" FORCE
```
mosra, line 379: Hmm, now that I see it in full, it might make sense to put these in a common variable so future additions don't need to happen to two places:

```cmake
set(common_plugins
  Magnum::AnyImageConverter
  Magnum::AnyImageImporter
  Magnum::AnySceneImporter
  MagnumPlugins::AssimpImporter
  MagnumPlugins::BasisImporter
  MagnumPlugins::GltfImporter
  MagnumPlugins::StanfordImporter
  MagnumPlugins::StbImageImporter)
if(BUILD_GUI_VIEWERS)
  set(MAGNUM_PYTHON_BINDINGS_STATIC_PLUGINS 
    ${common_plugins}
    MagnumPlugins::StbTrueTypeFont
    CACHE STRING "" FORCE)
else()
  set(MAGNUM_PYTHON_BINDINGS_STATIC_PLUGINS 
    ${common_plugins}
    CACHE STRING "" FORCE)
endif()
```

aclegg3: Yep, makes sense. Thanks for the code.

mosra (approved):

Skylion007 (approved):

src/cmake/dependencies.cmake
```diff
@@ -357,7 +357,6 @@ if(NOT USE_SYSTEM_MAGNUM)
       MagnumPlugins::AssimpImporter
       MagnumPlugins::BasisImporter
       MagnumPlugins::GltfImporter
-      MagnumPlugins::StanfordImporter
```
mosra: > remove StanfordImporter as it doesn't support ASCII ply.

Ah, that's one of my TODOs. I'm not interested in implementing parsing of ASCII PLY files, but it could delegate to Assimp for those.

Where do these come from, btw? Is there some Habitat code that writes them? Might be a candidate for refactor / replacement with builtin Magnum PLY export functionality.

What `StanfordImporter` can do and Assimp can't is import of files with per-vertex semantic IDs. I remember that used to be used quite often, not sure if that's still relevant for current dataasets.

aclegg3: Thanks for the thoughts on this topic.
I don't think semantic ids in .ply are necessary for python import, but ASCII ply is an easy format for hand-edited or human-readable mesh data. Agreed we generally don't try to use these in practice.

PR #2111, opened 2023-05-17
[breaking] - NavMeshSettings refactor

aclegg3:
## Motivation and Context

Several recent needs have exposed the brittleness of current NavMesh defaults for Simulator reconfiguration. This PR aims to provide much more flexibility in configuration and reduce unnecessary compute while remaining simple for default use-cases.

**This is a breaking change.**

Currently, when a Simulator is initialized, the default agent radius and height are used to trigger a re-compute of the NavMesh if any other settings are modified. This cannot be turned off and other settings cannot be modified without explicitly recomputing again after instantiation which can be costly in aggregate and occasionally block use-ability altogether (e.g. if some property of the stage causes navmesh computation to fail).

After this change:
1. An optional [NavMeshSettings](https://aihabitat.org/docs/habitat-sim/habitat_sim.nav.NavMeshSettings.html) object is added to [SimulatorConfiguration](https://aihabitat.org/docs/habitat-sim/habitat_sim.sim.SimulatorConfiguration.html). This can be configured by user code before the Simulator is instantiated and will trigger NavMesh re-computation as necessary. If omitted, the NavMesh will not be constructed (however, the scene may still load one from file).
2. The option to include [RigidObjects](https://aihabitat.org/docs/habitat-sim/habitat_sim.physics.ManagedBulletRigidObject.html) with [MotionType](https://aihabitat.org/docs/habitat-sim/habitat_sim.physics.html#MotionType) STATIC has been migrated into [NavMeshSettings](https://aihabitat.org/docs/habitat-sim/habitat_sim.nav.NavMeshSettings.html) to unify the API and eliminate the need for variable chains to configure this feature.
3. Backwards compatible default behavior (agent specific default NavMesh re-computation) is now realized through and configurable in [utils.settings.py](https://aihabitat.org/docs/habitat-sim/habitat_sim.utils.settings.html). (NOTE this API documentation is not updated until the PR is merged, see the file changes for details).
4. (breaking deprication) Any automated search for navmeshes matching a specific raw asset has been removed from Simulator.py. This is outdated and should be done in configuration files or the stage loader rather than shunted into the initialization step.

**TODO: Habitat-lab will need to be updated to use the new API. I'll write, link, and test that PR before merging this.**

Should cover the needs of #2104.

## How Has This Been Tested

Updated tests should cover this.
I did additional local testing with python viewer to double check.

## Types of changes

## Checklist


src/esp/sim/SimulatorConfiguration.h
```diff
@@ -5,11 +5,15 @@
 #ifndef ESP_SIM_SIMULATORCONFIGURATION_H_
 #define ESP_SIM_SIMULATORCONFIGURATION_H_
 
+#include <Corrade/Containers/Optional.h>
```
0mdc, line 8: This include is unused.

src/esp/sim/SimulatorConfiguration.h
```diff
@@ -5,11 +5,15 @@
...
 #include <string>
 
 #include "esp/core/Esp.h"
+#include "esp/nav/PathFinder.h"
```
0mdc, line 12: This should probably be forward-declared as this class is widely used.

0mdc, line 12: This should probably be forward-declared as this class is widely used.

src/esp/sim/SimulatorConfiguration.h
0mdc, line 8: This include is unused.

0mdc (approved):
LGTM! I left some minor remarks.

src/esp/sim/SimulatorConfiguration.h
aclegg3: Yeah, I thought so too. However, the compiler does not like the forward declaration here, so I needed the full the include.

PR #2112, opened 2023-05-18
--[PBR Update 1 of 3] Map materials' extension/layer data to Drawables cache

jturner65:
## Motivation and Context
This PR is the first of 3 (or possibly 4) PRs to implement Habitat's new PBR materials extension support, [drawn from the original WIP found here](https://github.com/facebookresearch/habitat-sim/pull/2091).

In this PR the gltf extension data present in the Magnum material are loaded into each drawable's existing cache structure, to be more easily accessed each draw.  No changes to the shader are present in this PR.

## How Has This Been Tested
Existing C++ and python tests pass locally.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/GenericDrawable.cpp
+++ b/src/esp/gfx/GenericDrawable.cpp
@@ -198,10 +198,8 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
       // fragment shader
       .setObjectId(
           static_cast<RenderCamera&>(camera).useDrawableIds() ? drawableId_
-          : ((((flags_ & Mn::Shaders::PhongGL::Flag::InstancedObjectId) ==
-               Mn::Shaders::PhongGL::Flag::InstancedObjectId) ||
-              ((flags_ & Mn::Shaders::PhongGL::Flag::ObjectIdTexture)) ==
-                  Mn::Shaders::PhongGL::Flag::ObjectIdTexture))
+          : (((flags_ >= Mn::Shaders::PhongGL::Flag::InstancedObjectId) ||
+              (flags_ >= Mn::Shaders::PhongGL::Flag::ObjectIdTexture)))
               ? 0
               : node_.getSemanticId())
       .setTransformationMatrix(transformationMatrix)
--- a/src/esp/gfx/PbrDrawable.cpp
+++ b/src/esp/gfx/PbrDrawable.cpp
@@ -375,8 +375,7 @@ void PbrDrawable::draw(const Mn::Matrix4& transformationMatrix,
       // the fragment shader
       .setObjectId(static_cast<RenderCamera&>(camera).useDrawableIds()
                        ? drawableId_
-                       : ((flags_ & PbrShader::Flag::InstancedObjectId) ==
-                                  PbrShader::Flag::InstancedObjectId
+                       : (flags_ >= PbrShader::Flag::InstancedObjectId
                               ? 0
                               : node_.getSemanticId()))
       .setProjectionMatrix(camera.projectionMatrix())
--- a/src/esp/gfx/PbrDrawable.cpp
+++ b/src/esp/gfx/PbrDrawable.cpp
@@ -7,6 +7,7 @@
 #include <Corrade/Containers/ArrayViewStl.h>
 #include <Corrade/Utility/FormatStl.h>
 #include <Magnum/Trade/MaterialData.h>
+#include <Magnum/Trade/PbrClearCoatMaterialData.h>
 #include <Magnum/Trade/PbrMetallicRoughnessMaterialData.h>
 
 #include <Magnum/GL/Renderer.h>
@@ -126,6 +127,7 @@ void PbrDrawable::setMaterialValuesInternal(
     }
     if (tmpMaterialData.normalTextureScale() != 1.0f) {
       flags_ |= PbrShader::Flag::NormalTextureScale;
+      matCache.normalTextureScale = tmpMaterialData.normalTextureScale();
       CORRADE_ASSERT(tmpMaterialData.normalTextureScale() > 0.0f,
                      "PbrDrawable::PbrDrawable(): the normal texture scale "
                      "must be positive.", );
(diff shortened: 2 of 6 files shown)


src/esp/gfx/PbrShader.cpp
```diff
@@ -119,10 +119,8 @@ PbrShader::PbrShader(Flags originalFlags, unsigned int lightCount)
                      ? "#define NONE_ROUGHNESS_METALLIC_TEXTURE\n"
                      : "")
       .addSource(flags_ & Flag::NormalTexture ? "#define NORMAL_TEXTURE\n" : "")
-      .addSource(flags_ & Flag::NormalTextureScale
-                     ? "#define NORMAL_TEXTURE_SCALE\n"
-                     : "")
       .addSource(flags_ & Flag::ObjectId ? "#define OBJECT_ID\n" : "")
+      .addSource(flags_ & Flag::ClearCoatLayer ? "#define CLEAR_COAT\n" : "")
```
jturner65, line 123: Oops one flag got past me as I cherry picked this. If this antagonizes anyone :grin: I'll get rid of it, but it's going to be added in the next PR anyway.

src/esp/gfx/GenericDrawable.cpp
```diff
@@ -198,10 +198,8 @@ void GenericDrawable::draw(const Mn::Matrix4& transformationMatrix,
       // fragment shader
       .setObjectId(
           static_cast<RenderCamera&>(camera).useDrawableIds() ? drawableId_
-          : ((((flags_ & Mn::Shaders::PhongGL::Flag::InstancedObjectId) ==
-               Mn::Shaders::PhongGL::Flag::InstancedObjectId) ||
-              ((flags_ & Mn::Shaders::PhongGL::Flag::ObjectIdTexture)) ==
-                  Mn::Shaders::PhongGL::Flag::ObjectIdTexture))
+          : (((flags_ >= Mn::Shaders::PhongGL::Flag::InstancedObjectId) ||
```
eundersander, line 201: Why do we compare flags with >=? Isn't flags basically a bitfield?

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -56,13 +58,17 @@ void PbrDrawable::setMaterialValuesInternal(
   matCache.metalness = tmpMaterialData.metalness();
   matCache.emissiveColor = tmpMaterialData.emissiveColor();
 
-  if (materialData_->hasAttribute("metallicTexturePointer") &&
-      materialData_->hasAttribute("roughnessTexturePointer")) {
+  const Cr::Containers::Optional<Mn::GL::Texture2D*> metallicTexturePtr =
```
eundersander, line 61: This is pretty verbose. Could we use `auto` for the type of `metallicTexturePtr`? And/or a local, reusable lambda function?

src/esp/gfx/PbrDrawable.cpp
```diff
@@ -117,31 +123,265 @@ void PbrDrawable::setMaterialValuesInternal(
...
     }
-    if (tmpMaterialData.normalTextureScale() != 1.0f) {
-      flags_ |= PbrShader::Flag::NormalTextureScale;
-      CORRADE_ASSERT(tmpMaterialData.normalTextureScale() > 0.0f,
-                     "PbrDrawable::PbrDrawable(): the normal texture scale "
-                     "must be positive.", );
-    }
+    // normal texture scale
+    matCache.normalTextureScale = tmpMaterialData.normalTextureScale();
+    CORRADE_ASSERT(tmpMaterialData.normalTextureScale() > 0.0f,
```
eundersander, line 136: I realize you didn't add this assert, but anyway it would be nice to convert it to an ESP_CHECK. Doesn't feel like an assert.

src/esp/gfx/PbrShader.h
```diff
@@ -172,41 +164,102 @@ class PbrShader : public Magnum::GL::AbstractShaderProgram {
...
+    InstancedObjectId = (1 << 11) | ObjectId,
+
+    /**
+     * Has ClearCoat layer.
+     */
+    ClearCoatLayer = 1 << 12,
+    /**
+     * Has ClearCoat Texture in ClearCoat layer
+     */
+    ClearCoatTexture = (1 << 13) | ClearCoatLayer,
```
eundersander, line 188: I haven't seen this convention before so maybe a comment explaining. I guess you are basically grouping similar enum values to have a shared bit.

src/esp/gfx/PbrShader.h
```diff
@@ -131,18 +131,10 @@ class PbrShader : public Magnum::GL::AbstractShaderProgram {
      */
     NormalTexture = 1 << 6,
 
-    /**
```
eundersander, line 134: Just curious, why was this removed? (My guess: your new version of the shader will always use a scale, so you didn't need this flag anymore.)

src/esp/gfx/PbrShader.h
eundersander: Also, I see we're using these with `Corrade::EnumSet` and from [here](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html) I see "The only requirement for the enum type is that all the values must be binary exclusive." @mosra thoughts?

eundersander (commented):
I left some comments! I'll let others weigh in on the enum usage concerns before approving.

src/esp/gfx/GenericDrawable.cpp
jturner65: that's how magnum compares multi-bit flags

src/esp/gfx/PbrShader.h
jturner65: Yep. Talked to mosra about it and it's sufficiently cheap to always enable.

src/esp/gfx/PbrDrawable.cpp
mosra: Won't negative scales simply invert the normal texture, i.e. a weird but valid use case? I don't think this need an assertion at all.

src/esp/gfx/PbrShader.h
mosra: I'll remove that line from the docs. Having a particular flag a superset of another is a very common use case, so the "requirement" makes no sense nowadays.

src/esp/gfx/GenericDrawable.cpp
mosra: [Docs for reference](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html#aeb270daa9a8afb3a80e57ef2cb2f8314), `>=` is chosen because it looks close to a ⊇, a symbol for a superset.

src/esp/gfx/PbrDrawable.cpp
mosra: I don't like `auto` but I think it's fine to use it in this case -- the type is almost clear from the findAttribute call below.

src/esp/gfx/PbrDrawable.cpp
jturner65: how about abs(matCache.normalTextureScale) > 0?

jturner65: oh, it won't bug out when I try to query it in the shader?  cool

Skylion007 (approved):

src/esp/gfx/PbrDrawable.cpp
mosra: Why would it? You simply get a zero value, so as long as you don't divide anything with it, you'll get a valid result.

src/esp/gfx/GenericDrawable.cpp
eundersander: Ok so, what the heck do these lines mean? :) Is it testing whether `flags_` includes `InstanceObjectId`?

mosra: > Is it testing whether `flags_` includes `InstanceObjectId`?

Yes. It's a shorthand for `(a & other) == other`, as the docs say. It does exactly what it did before -- checking if one set of bits is a superset of the other set -- just in a more concise way.

PR #2100, opened 2023-05-09
Update Magnum with bit array bindings, application cursor bindings and more

mosra:
## Motivation and Context

Random assortment of Magnum-side updates. Opened mainly in order to add support for cursor changing in python applications, a larger batch of changes with new scene and material tools that includes also some breaking changes is still pending. 

Adding more people as reviewers mainly so you get the changelog. (Let me know if anybody else would be interested in these so I can include them too.) What changed in your areas of interest:

- New [Utility::copyMasked()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility.html#aa0f3e747d2eba025af0c123cfb7c4227) algorithm for copying an array to another with a bit mask. ECS, DOD, etc. No LZCNT-like optimizations yet due to lack of time.
- Direct element access in multi-dimensional strided (bit) views. You can now access e.g. a 3D view element by `[{a, b, c}]` instead of `[a][b][c]`. These two get optimized down to an equivalent code in Release builds, but generate less overhead in Debug builds.
- Added [Trade::ImporterFlag::Quiet](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Trade.html#a137d832c7ce6e6b654022321941d70f3a098753f8980036f4b936e3d4b6997111), [Trade::ImageConverterFlag::Quiet](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Trade.html#a9db1b8d09ba3b3d39bca8c40597cbea0a098753f8980036f4b936e3d4b6997111) and [Trade::SceneConverterFlag::Quiet](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Trade.html#ac3f3c2001df6c6e4761ed6e4e70d700da098753f8980036f4b936e3d4b6997111) for suppressing warning output from importer and converter plugins, with corresponding Python bindings. All plugins respect these now. Not hooked up as a `-q` option in the command-line tools yet. A fun task for someone could be to hook this flag up with the `HABITAT_LOG` env var, just please don't enable the flags always globally. They're there for a reason and can point to potential data loss issues. (Requested by @aclegg3, sorry that the PR took a while)
[...]

PR #2115, opened 2023-05-19
[logging] - set Magnum importer and converter Quiet flags 

aclegg3:
## Motivation and Context

Leverage new Magnum APIs from #2100 to set Quiet flags when corresponding sim subsystem logging level is below debug.

## How Has This Been Tested

Locally with HSSD.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/CubeMap.cpp
+++ b/src/esp/gfx/CubeMap.cpp
@@ -403,6 +403,10 @@ bool CubeMap::saveTexture(TextureType type,
   if (!(converter = manager.loadAndInstantiate("AnyImageConverter"))) {
     return false;
   }
+  // set image converter flags if gfx logging is quieted
+  if (!isLevelEnabled(logging::Subsystem::gfx, logging::LoggingLevel::Debug)) {
+    converter->addFlags(Magnum::Trade::ImageConverterFlag::Quiet);
+  }
 
   const char* coordStrings[6] = {"+X", "-X", "+Y", "-Y", "+Z", "-Z"};
   for (int iFace = 0; iFace < 6; ++iFace) {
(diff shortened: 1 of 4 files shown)


src/esp/gfx/CubeMap.cpp
```diff
@@ -403,6 +403,10 @@ bool CubeMap::saveTexture(TextureType type,
   if (!(converter = manager.loadAndInstantiate("AnyImageConverter"))) {
     return false;
   }
+  // set image converter flags if gfx logging is quieted
+  if (!isLevelEnabled(logging::Subsystem::gfx, logging::LoggingLevel::Debug)) {
```
mosra, line 407: Hmm, doesn't this mean that if I set "warnings only", i.e. [LoggingLevel::Warning](https://aihabitat.org/docs/habitat-sim/namespaceesp_1_1logging.html#af964da7d81bdebf67382b6f9dc78c4d1a0eaadb4fcb48a0a0ed7bc9868be9fbaa), warnings from the importer will be suppressed?

As in, shouldn't that be checking the level between `Debug` and `Quiet` instead? Apologies if I misunderstood how the levels work.

```suggestion
  if (!isLevelEnabled(logging::Subsystem::gfx, logging::LoggingLevel::Warning)) {
```

src/esp/gfx/CubeMap.cpp
```diff
@@ -403,6 +403,10 @@ bool CubeMap::saveTexture(TextureType type,
...
+    converter->addFlags(Magnum::Trade::ImageConverterFlag::Quiet);
+  }
```
mosra, line 409: While you're at it there's also a `*Flag::Verbose`, could you hook it up in a similar way? That would match the `LoggingLevel::VeryVerbose` level I think.

src/esp/gfx/CubeMap.cpp
aclegg3: Makes sense. Moved this to Warning level.

PR #2117, opened 2023-05-24
Decrease minimum required GL version for most shaders

### LINKED ISSUE #2089: Pass 'Combine redundant instructions' is not initialized.
python examples/viewer.py --dataset data/replica_cad/replicaCAD.scene_dataset_config.json --scene apt_1
I did this .
```

[15:13:58:654358]:[Assets] ResourceManager.cpp(984)::loadStageInternal : Attempting to load stage data/replica_cad/configs/stages/../../stages/frl_apartment_stage.glb
[15:13:58:655350]:[Assets] ResourceManager.cpp(866)::loadRenderAsset : Loading general asset named: data/replica_cad/configs/stages/../../stages/frl_apartment_stage.glb
[15:13:58:655680]:[Assets] ResourceManager.cpp(1655)::configureImporterManagerGLExtensions : Importing Basis files as ASTC 4x4.
Pass 'Combine redundant instructions' is not initialized.
Verify if there is a pass dependency cycle.
Required Passes:
Segmentation fault (core dumped)
```
I got this error . How I do solve this error?

Skylion007:
## Motivation and Context
* Some of our more rarely used shaders use GL versions that some GPUs don't fully support despite the fact we aren't using any features that necessitate these new versions. Downgrade the GL version where appropriate to improve shader compatibility.  Fixes #2089

## How Has This Been Tested
* Locally with pytest test suite

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/DoubleSphereCameraShader.cpp
+++ b/src/esp/gfx/DoubleSphereCameraShader.cpp
@@ -35,7 +35,7 @@ DoubleSphereCameraShader::DoubleSphereCameraShader(
 #ifdef MAGNUM_TARGET_WEBGL
   Mn::GL::Version glVersion = Mn::GL::Version::GLES300;
 #else
-  Mn::GL::Version glVersion = Mn::GL::Version::GL410;
+  Mn::GL::Version glVersion = Mn::GL::Version::GL330;
 #endif
 
   // this is not the file name, but the group name in the config file
--- a/src/esp/gfx/EquirectangularShader.cpp
+++ b/src/esp/gfx/EquirectangularShader.cpp
@@ -40,7 +40,7 @@ EquirectangularShader::EquirectangularShader(Flags flags)
 #ifdef MAGNUM_TARGET_WEBGL
   Mn::GL::Version glVersion = Mn::GL::Version::GLES300;
 #else
-  Mn::GL::Version glVersion = Mn::GL::Version::GL410;
+  Mn::GL::Version glVersion = Mn::GL::Version::GL330;
 #endif
 
   // this is not the file name, but the group name in the config file
--- a/src/esp/gfx/PbrEquiRectangularToCubeMapShader.cpp
+++ b/src/esp/gfx/PbrEquiRectangularToCubeMapShader.cpp
@@ -40,7 +40,7 @@ PbrEquiRectangularToCubeMapShader::PbrEquiRectangularToCubeMapShader() {
 #ifdef MAGNUM_TARGET_WEBGL
   Mn::GL::Version glVersion = Mn::GL::Version::GLES300;
 #else
-  Mn::GL::Version glVersion = Mn::GL::Version::GL410;
+  Mn::GL::Version glVersion = Mn::GL::Version::GL330;
 #endif
 
   // this is not the file name, but the group name in the config file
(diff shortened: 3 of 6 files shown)


Skylion007:
> > Also it's not an issue since these shaders don't use any 4+ GL features (except maybe the PTex Shader).
> 
> Actually, I'm getting a cast error in my PBR work if I use 3.30 instead of 4.1

Ah, interesting... Could you post the full cast error?

jturner65:
It does seem like folks are able to use 4.1 on some macs.  Could this be a configuration/setup issue for the user?

mosra:
Going with GLSL 3.30 is a good thing because it gets you closer to the syntax WebGL 2 expects, so this is a good change. Anything from newer versions can be opted in via extensions (you can look at code of Magnum's builtin shaders if you're interested, they're also mostly just on GLSL 3.30 or even earlier).

So if you fix that one casting issue that popped up, I think this PR is good to go. It might be good to have the shader code validated for WebGL 2 as well. I'll see if I can provide you the tools to do that without having to compile & run the whole Emscripten build. But even without that, the PR already fixes _something_, so it could get merged even without the shaders being valid for WebGL 2.

> It does seem like folks are able to use 4.1 on some macs.

Some older Macs have only GL 3.3 IIRC, and those Macs don't even have Metal, just GL. Another possible case is older GPUs, GL 4.0 requires certain features such as 64-bit integers and floats in shaders that older architectures don't have any support for.

jturner65:
@Skylion007 Probably want to add PTex shader too, no?

mosra (changes requested):
> PTex shader 

I don't think it's worth modifying that one. Its test coverage is very low (if any at all), no current dataset uses PTex, and so if you accidentally break it you may not even know. Leave it at 4.1.

This still needs to fix that cast error though you mentioned, so no approval yet -- John, can you say what to change and where?

jturner65:
> I don't think it's worth modifying that one. Its test coverage is very low (if any at all), no current dataset uses PTex, and so if you accidentally break it you may not even know. Leave it at 4.1.
Should we remove that support then?  There's so much custom ptex stuff, if we're not going to support it anymore then we should consider removing it, I think.

> This still needs to fix that cast error though you mentioned, so no approval yet -- John, can you say what to change and where?

I made these changes, and fixed the issues that resulted, in [the most recent commit in my PBR/IBL WIP Pr ](https://github.com/facebookresearch/habitat-sim/pull/2091/commits/790cc136a042a31f363211b672147c32965e8711) .

Skylion007:
@jturner65 Whoops, accidentally removed your approval.

jturner65:
One thing that needs to be tested that I did not get to is the various [texture queries for ID](https://github.com/facebookresearch/habitat-sim/blob/dfdc577aa8b69d84d250808def667713bf0fe926/src/shaders/doubleSphereCamera.frag#L70) - not sure if that will cause the cast complaint.  I'll check it tomorrow if nobody else has by then :).

mosra:
`texture()` returns a vector type matching the sampler type ([docs](https://registry.khronos.org/OpenGL-Refpages/gl4/html/texture.xhtml), the `g` stands for a "generic type prefix"). So in this case, `sampler2D` will return a `vec4`, `usampler2D` an `uvec4`. No cast needed.

mosra (approved):

PR #2119, opened 2023-05-29
--[PBR Update 2 of 3] PBR rewrite-motivated default lighting and IBL fix/reworking

jturner65:
## Motivation and Context
As part of [the ongoing PBR shader rewrite/functional expansion effort,](https://github.com/facebookresearch/habitat-sim/pull/2091) this PR contains changes that were made to the default lighting configurations, as well as providing a few fixes.  Specifically : 

1. A heuristically-derived constant intensity scaling factor was added to the PBR shader code, instead of having a large scaling/intensity factor hardcoded into the default lighting values themselves. This way the lights will work acceptably for both phong and PBR shaded objects. NOTE : This hardcoded value is obviated and removed in the PBR fragment shader PR.
2. The default lights themselves were tweaked to give reasonable performance, both intensity and visuals, for both PBR and Phong shaded scenes.  NOTE : PBR performance is in reference to the new shader functionality, not the existing shader, although the existing shader also looks better.
3. The IBL implementation was fixed so that it will work with the python viewer. This required a rework of how it was instantiated (i.e. taking it out of the ResourceManager constructor).
4. Shader direct/indirect diffuse & specular lighting balance in PBR was adjusted so that IBL lighting is not overpowered/washed out.  NOTE : Although this balancing was made with the new PBR shader in mind, the existing PBR output looks better as well.
5. New test images were added to reflect the new lighting. NOTE : These new test images will need to be replaced again when the new PBR shader work is merged.

NOTE : Many of these changes were driven by the new PBR shader, and so may not appear ideal with the current shader.

## How Has This Been Tested
Existing c++ and python tests pass, with the addition of new test images to account for lighting changes.

## Types of changes

## Checklist


jturner65:
One thing to note : any existing dataset that has lighting configs defined and is intended for PBR shading will probably need those lights adjusted.  AFAIK ReplicaCAD is the only dataset currently that would need this rebalancing, but I'm not sure. @aclegg3 ?

src/esp/gfx/LightSetup.cpp
```diff
@@ -76,12 +76,34 @@ LightSetup getLightsAtBoxCorners(const Magnum::Range3D& box,
...
+       {0.6, 0.6, 0.6},
+       LightPositionModel::Global},  // downward
+
+      // camera-relative
+      {{0.0, -0.1, -1.0, 0.0},
+       {0.3, 0.3, 0.3},
+       LightPositionModel::Camera},  // forward
+      {{0.0, -0.3, 1.0, 0.0},
+       {0.5, 0.5, 0.5},
+       LightPositionModel::Camera},  // backward (for glancing speculars)
```
0mdc, line 104: I would avoid emitting light from the camera by default here.

src/esp/assets/ResourceManager.h
```diff
@@ -744,6 +726,22 @@ class ResourceManager {
...
+   * NOTE!!! Such an image MUST be SPECIFIED in the
+   * ~/habitat-sim/data/pbr/PbrImages.conf
+   * and be put in that folder.
+   * example image:
+   * ~/habitat-sim/data/pbr/lythwood_room_4k.png
+   *
+   * TODO: this will be replaced by config-driven IBL asset loading and
+   * registration
+   */
+  void initPbrImageBasedLighting(const std::string& hdriImageFilename);
```
0mdc, line 743: What is the current workflow to generate these cubemaps?

src/esp/gfx/LightSetup.cpp
jturner65: This is done specifically so that surface details such as normal or clearcoat texture/specular can be readily seen in the new PBR shader regardless of camera orientation (i.e. demo screen captures).

src/esp/sim/Simulator.cpp
```diff
@@ -200,6 +196,11 @@ void Simulator::reconfigure(const SimulatorConfiguration& cfg) {
 #endif
     renderer_->acquireGlContext();
   }
+  // load IBL assets if appropriate and not loaded already
+  // TODO : So many things.  Needs to be config driven, for one.
```
mosra, line 200: Hm, given you're replacing the "ground truth" test images here, I'm thinking it could make sense to load an actual HDR image here, not a LDR. Which again changes the intensity / balance a bit, but closer to the ideal I think.

As far as I remember, the JPEG was a temporary thing Yili did because at that point EXR loading wasn't done from my side yet. But now it is and -- as far as I'm aware -- is also built & enabled inside Habitat, and loading it here instead of the JPEG should hopefully "just work". The `*.exr` can be downloaded from here: https://polyhaven.com/a/lythwood_room

src/esp/gfx/LightSetup.cpp
```diff
@@ -76,12 +76,34 @@ LightSetup getLightsAtBoxCorners(const Magnum::Range3D& box,
...
+
+  };
```
mosra, line 106: Just so everyone is aware -- this is 7 lights instead of 5, which makes the fragment shader roughly as much as 40% more computationally expensive. 

Personally I don't think that many lights is needed, even a three-point lighting was often enough to get a balanced output.

src/esp/sim/Simulator.cpp
jturner65: Unfortunately, the exr file is 17 mb, too big without LFS.  Do we want to force LFS on folks to get anything IBL working?  

As far as HDR images are concerned, I already have a few that I've been playing with, I just wanted to keep the changes minimal with this PR.  There's going to be a follow-up (after the PBR shader work gets merged in, ideally) that will include a more rigorous redesign of the IBL subsystem. (i.e. config-driven for one)

src/esp/gfx/LightSetup.cpp
jturner65: I was aware of the potential impact of these two additional camera-related lights, but their intention is specifically to give easily visualized specular/textural response for those materials that have texture-based surface normals, clearcoat texture and cc normal texture and anisotropy (so far).  

If folks think this is too expensive, I can remove the two camera facing lights, and instead provide a lighting config that would define the lights in this manner that folks making demos could consume.

src/esp/gfx/LightSetup.cpp
jturner65: Like I said below, I can remove these in favor of a custom lighting config added to /data that folks could just load up to take screen shots and such.

src/esp/sim/Simulator.cpp
jturner65: To be honest, the jpg seems to me to be working fine, at least for demonstration purposes.

mosra: You can make it smaller with `magnum-imageconverter`. I don't think a 4K image is needed for our use case, the below shrinks the file to 5.7 MB. 

```sh
magnum-imageconverter lythwood_room_4k.exr -C StbResizeImageConverter -c size="1024 512" out.exr
```

It could be also converted from 32-bit floats to half-floats to drop ~half of the data but you'd have to write a Python script for that -- the EXR plugins don't have any convenience toggles for data conversion right now.

> I just wanted to keep the changes minimal with this PR

I only mentioned this because I saw the changes in test images, and the LDR JPEG is really not a good option so for me this felt like a good opportunity to switch.

src/esp/gfx/LightSetup.cpp
jturner65: It is also pretty easy to just author a new light config with fewer lights if the default is considered too computationally expensive.

src/esp/sim/Simulator.cpp
jturner65: Cool, thanks!  I'll do this.

src/esp/sim/ClassicReplayRenderer.cpp
```diff
@@ -27,9 +27,8 @@ ClassicReplayRenderer::ClassicReplayRenderer(
   SimulatorConfiguration simConfig;
   simConfig.createRenderer = true;
   auto metadataMediator = metadata::MetadataMediator::create(simConfig);
-  assets::ResourceManager::Flags flags{};
   resourceManager_ =
-      std::make_unique<assets::ResourceManager>(metadataMediator, flags);
+      std::make_unique<assets::ResourceManager>(metadataMediator);
```
Skylion007, line 31:
```suggestion
      std::make_unique<assets::ResourceManager>(std::move(metadataMediator));
```

src_python/habitat_sim/utils/settings.py
```diff
@@ -32,6 +32,7 @@
     "seed": 1,
     "physics_config_file": "data/default.physics_config.json",
     "enable_physics": True,
+    "pbr_image_based_lighting": False,
```
Skylion007, line 35: Will there be other types of lighting supported in the future, should we make this an enum to future proof the API?

jturner65: Not exclusionary, to the best of my knowledge.  We will also add an HBAO flag eventually.

src/esp/sim/Simulator.cpp
mosra: For anyone else tracking this, https://github.com/mosra/magnum-plugins/commit/b8a1f670b21d2347515ccf707f5d42cb79efd810 and https://github.com/mosra/magnum-ci/actions/runs/5115547986 has a new option to import 32-bit EXRs as 16-bit, which is usable for significant size reduction.

src_python/habitat_sim/utils/settings.py
jturner65: Actually I did set up the fragment shader to support picking which extensions are supported (to isolate potential bugs and also to possibly speed up rendering), but that's in the most rudimentary stage at the moment, and I'm not sure how the functionality will be exposed to the users.  A bit early to add anything related to that in this PR.

src/esp/gfx/LightSetup.cpp
0mdc: I think that one directional light and an ambient light would suffice as a default.

jturner65: A single directional light will not look good at all for PBR scenes (I've tried it), and ambient light is neither used nor useful in PBR, since it lacks a light vector.  Of course it probably will be fine with Phong lighting, so perhaps we should set up a different default for Phong? 

At the -very least-, I would suggest the 5 base global directional lights I have in the current default setup for PBR.

jturner65:
Ok, I've modified the default lights - now there are only 4, and they seem to give reasonable illumination in both Phong and PBR scenes (using the old PBR shader).  See images 

Phong shader : 
phongDefaultLights

Current (before refactor/rewrite) PBR shader :
currPBRDefaultLights

jturner65:
Here are these default lights using the new PBR shader - very similar (some of the surfaces are a bit darker due to clearcoat effect).

newPBRDefaultLights

aclegg3 (approved):
Looks good. We'll merge once ReplicaCAD update is ready.

PR #2120, opened 2023-05-31
Add skinning support for gfx-replay and legacy replay rendering.

0mdc:
## Motivation and Context

This changeset adds support for:

- Skinned mesh gfx-replay support
- Classic replay rendering of skinned meshes

Capture from the SIRo sandbox tool, which renders using the classic replay renderer:

https://github.com/facebookresearch/habitat-sim/assets/110583667/271a88e4-3b3c-4433-8cd8-efa63fcf935b

### Overview

This changeset introduces a new structure, `Rig`, which contains a node for each bone of a skinned model. During asset instantiation, the `Rig` is associated to the model instance so that it can be posed during rendering.

Therefore, the skinned model is animated by modifying transforms of the `Rig` bone nodes. This is currently done in either of those ways:
* via articulated objects: Bone nodes are parented to their matching articulated object links.
* via replay rendering: Bone node world transforms are written from the `rigUpdate.pose` in gfx-replay keyframes.

### Gfx-replay

Two new keyframe fields are introduced:
* rigCreations: Registers a new rig, along with its bone names.
* rigUpdates: Lists the new transforms for each bone (pose).

Additionally, RenderAssetCreationInfo now has a rigId field.

## How Has This Been Tested

* Unit test
* Tested in Python viewer, Blender and SIRo Sandbox Tool

## Types of changes

## Checklist


src/esp/gfx/replay/Recorder.cpp
```diff
@@ -216,6 +237,33 @@ void Recorder::updateInstanceStates() {
       instanceRecord.recentState = state;
     }
   }
+
+  // Update rig states
+  for (const auto& rigItr : rigNodes_) {
```
eundersander, line 242: How about don't write an update if the rig state hasn't changed? Similar to above logic.

src/esp/gfx/replay/Recorder.cpp
```diff
@@ -208,6 +228,7 @@ RenderAssetInstanceState Recorder::getInstanceState(
 }
 
 void Recorder::updateInstanceStates() {
```
eundersander, line 230: Can you update this function name? Since you are updating instance states and also ... "rig instances" (?)

src/esp/gfx/replay/Recorder.cpp
```diff
@@ -216,6 +237,33 @@ void Recorder::updateInstanceStates() {
...
+      }
+
+      Transform absTransform{absTransformMat.translation(),
+                             Magnum::Quaternion::fromMatrix(rotationShear)};
+      boneUpdate.absTransform = absTransform;
+
+      boneUpdate.rigId = rigId;
+      boneUpdate.boneId = boneIdx;
+
+      currKeyframe_.boneUpdates.emplace_back(boneUpdate);
```
eundersander, line 264: What is the reason for doing K bone updates instead of a single "rig update" (that updates all K bones)? It looks like your `boneIdx` just goes from 0..numBones, so you could entirely avoid writing those bone indices if you did a single update for all of them, yes?

src/esp/io/JsonEspTypes.cpp
```diff
@@ -150,9 +158,9 @@ bool fromJsonValue(const JsonGenericValue& obj,
                    metadata::attributes::ObjectInstanceShaderType& x) {
   std::string shaderTypeToUseString;
   // read as string
-  bool shaderTypeSucceess = fromJsonValue(obj, shaderTypeToUseString);
+  bool shaderTypeSuccess = fromJsonValue(obj, shaderTypeToUseString);
```
eundersander, line 161: Thanks for fixing all the spelling typos!

src/esp/gfx/replay/Keyframe.h
```diff
@@ -39,18 +39,43 @@ struct RenderAssetInstanceState {
...
+  int rigId = ID_UNDEFINED;
+  int boneId = ID_UNDEFINED;
+  std::string boneName;
+};
+
+/**
+ * @brief The dynamic state of a rigged object bone that is tracked by the
+ * replay system every frame.
+ */
+// TODO: Serializer doesn't support collections in nested types. Therefore,
```
eundersander, line 57: This isn't my understanding. I thought our serialization code could handle arbitrary nesting of all serialization-supported types including e.g. vectors of vectors or vectors of objects containing vectors. What nesting did you want to do and what error did you hit?

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -945,5 +938,40 @@ void BulletPhysicsManager::removeRigidConstraint(int constraintId) {
   }
 }
 
+void BulletPhysicsManager::instantiateSkinnedModel(
```
eundersander, line 941: @aclegg3 maybe you wanna take a look at changes to BulletPhysicsManager.

src/esp/gfx/replay/Recorder.h
```diff
@@ -186,6 +196,7 @@ class Recorder {
   Keyframe currKeyframe_;
   std::vector<Keyframe> savedKeyframes_;
   RenderAssetInstanceKey nextInstanceKey_ = 0;
+  std::unordered_map<int, std::vector<scene::SceneNode*>> rigNodes_ = {};
```
eundersander, line 199: Nit: the `= {}` is unnecessary for `unordered_map` AFAIK.

src/esp/assets/ResourceManager.cpp
```diff
@@ -3522,6 +3522,41 @@ void ResourceManager::setLightSetup(gfx::LightSetup setup,
                      Mn::ResourcePolicy::Manual);
 }
 
+int ResourceManager::registerRigInstance(gfx::Rig&& rig) {
```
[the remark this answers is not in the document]
0mdc: Yes, I think that this would make sense. `ResourceManager` is already bloated enough and the lifetime of these instance would be better controlled.

>Possibly related: BatchReplayRenderer doesn't (and never should) have a ResourceManager.
Agreed. I'm on the fence whether we should do that refactor here, or along with the skinned mesh integration with the batch renderer.

src/esp/gfx/replay/Keyframe.h
0mdc: I wanted to serialize `RigCreations` and `RigUpdates` structs rather than individual bones to simplify the code and avoid writing one `rigId` per bone entry.

Both of these would be structs, with collections of bones and rig-specific data if needed.

My understanding is that the json serialization templating breaks down when a top-level object has a collection of structs, which themselves have collections. This applies to vectors, optionals, and others collections.

I considered supporting that outside of the scope of this changeset. I would focus on making the feature work and revisit serialization altogether when we look into optimization.

@jturner65 can you chime in here?

src/esp/gfx/replay/Recorder.cpp
0mdc: See: https://github.com/facebookresearch/habitat-sim/pull/2120#discussion_r1228735908

src/esp/assets/RigManager.h
```diff
@@ -0,0 +1,73 @@
...
+ private:
+  /**
+   * @brief The rigs for instantiated skinned assets.
+   */
+  std::unordered_map<int, gfx::Rig> _rigInstances;
+
+  /**
+   * @brief The next available unique ID for instantiated rigs.
+   */
+  int _nextRigInstanceID = 0;
```
jturner65, line 67: In the future we could revisit this so that [it managed IDs similar to how PhysicsManager does](https://github.com/facebookresearch/habitat-sim/blob/c5f38b3382627a5aa742b4528202f394642b5de4/src/esp/physics/PhysicsManager.h#L1249C20-L1249C38), reusing freed IDs instead of just incrementing, on the off chance that the RigManager may persist through billions of training cycles.

jturner65 (approved):
LGTM. Always nice to see some things move out of ResourceManager that can.

0mdc:
FYI - Tests were passing prior to rebasing to main due to our CI image losing support.

`Job was rejected because resource class gpu.nvidia.medium, image ubuntu-2004-cuda-11.4:202110-01 is not a valid resource class`

The [last run](https://github.com/facebookresearch/habitat-sim/pull/2120/commits/477379d0e868585e713d8a8dfc97f82dc2df02f8) was green.

eundersander (approved):
LGTM!

PR #2136, opened 2023-06-25
Replay renderer partial deletion

0mdc:
## Motivation and Context

The batch renderer can only delete full scenes - it cannot delete single objects.

To circumvent this, deletions are processed as such:

1. Latest transforms are cached.
2. All instances are deleted.
3. Remaining instances are re-created.
4. Latest transforms are re-applied.

## How Has This Been Tested

Tested locally.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/gfx/replay/Player.cpp
+++ b/src/esp/gfx/replay/Player.cpp
@@ -239,54 +239,8 @@ void Player::applyKeyframe(const Keyframe& keyframe) {
       implementation_->deleteAssetInstance(it->second);
       createdInstances_.erase(deletionInstanceKey);
     }
-    // The batch renderer can only clear the scene entirely; it cannot delete
-    // individual objects. To process deletions, all instances are deleted,
-    // remaining instances are re-created and latest transform updates are
-    // re-applied.
-  } else if (keyframe.deletions.size() > 0) {
-    // Cache latest transforms
-    std::unordered_map<RenderAssetInstanceKey, Mn::Matrix4> latestTransforms{};
-    for (const auto& pair : this->createdInstances_) {
-      const RenderAssetInstanceKey key = pair.first;
-      latestTransforms[key] = implementation_->getNodeTransform(pair.second);
-    }
-
-    // Delete all instances
-    implementation_->deleteAssetInstances(createdInstances_);
-
-    // Remove deleted instances from records
-    for (const auto& deletion : keyframe.deletions) {
-      const auto& createInstanceIt = createdInstances_.find(deletion);
-      if (createInstanceIt == createdInstances_.end()) {
-        // Missing instance for this key due to a failed instance creation
-        continue;
-      }
-      createdInstances_.erase(createInstanceIt);
-      const auto& creationInfoIt = creationInfos_.find(deletion);
-      if (creationInfoIt == creationInfos_.end()) {
-        // Missing instance for this key due to a failed instance creation
-        continue;
-      }
-      creationInfos_.erase(creationInfoIt);
-    }
-
-    for (const auto& pair : createdInstances_) {
-      const RenderAssetInstanceKey key = pair.first;
-      const auto& creationInfoIt = creationInfos_.find(key);
-      if (creationInfoIt == creationInfos_.end()) {
-        // Missing instance for this key due to a failed instance creation
-        continue;
-      }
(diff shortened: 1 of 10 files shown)


src/esp/gfx/replay/Player.cpp
```diff
@@ -223,27 +221,78 @@ void Player::applyKeyframe(const Keyframe& keyframe) {
     const auto& instanceKey = pair.first;
     CORRADE_INTERNAL_ASSERT(createdInstances_.count(instanceKey) == 0);
     createdInstances_[instanceKey] = node;
+    creationInfos_[instanceKey] = adjustedCreation;
   }
 
-  if (!frameCleared) {
+  // TODO: Move the code below into renderer-specific implementations.
```
0mdc, line 227: @eundersander 

The implementation suggested here is done in the player itself rather than the renderer-specific implementations, therefore goes against the current design.

I did it here because the `RenderAssetInstanceKey` bookkeeping is done here. Otherwise, it would have to be refactored into `AbstractSceneGraphPlayerImplementation` and `BatchPlayerImplementation`.

As previously discussed, we would like to have each renderer to have their own player, I figured that this dynamic cast is an acceptable trade-off.

Let me know if you think that this should be refactored immediately.

src/esp/gfx/replay/Player.cpp
```diff
@@ -223,27 +221,78 @@ void Player::applyKeyframe(const Keyframe& keyframe) {
...
       implementation_->deleteAssetInstance(it->second);
       createdInstances_.erase(deletionInstanceKey);
     }
+    // The batch renderer can only clear the scene entirely; it cannot delete
+    // individual objects. To process deletions, all instances are deleted,
+    // remaining instances are re-created and latest transform updates are
+    // re-applied.
+  } else if (keyframe.deletions.size() > 0) {
+    // Cache latest transforms
+    std::unordered_map<RenderAssetInstanceKey, Mn::Matrix4> latestTransforms{};
```
eundersander, line 248: Let's get out of the habit of per-frame dynamic allocations. Make this a class member and then clear and re-use it here.

src/esp/gfx/replay/Player.h
```diff
@@ -85,15 +85,32 @@ class AbstractPlayerImplementation {
...
+  virtual void setNodeTransform(const NodeHandle node,
+                                const Mn::Matrix4& transform) = 0;
+
+  /**
+   * @brief Get node transform.
+   *
+   * The @p handle is expected to be returned from an earlier call to
+   * @ref loadAndCreateRenderAssetInstance() on the same instance.
+   */
+  virtual const Mn::Matrix4 getNodeTransform(const NodeHandle node) const = 0;
```
eundersander, line 112: Maybe rename this `hackGetNodeTransform` to convey that this accessor is part of our hack?

eundersander (commented):
I left some comments. Let me know when I should take another look!

src/esp/gfx/replay/Player.cpp
```diff
@@ -239,16 +264,12 @@ void Player::applyKeyframe(const Keyframe& keyframe) {
...
-    // individual objects. To process deletions, all instances are deleted,
-    // remaining instances are re-created and latest transform updates are
-    // re-applied.
   } else if (keyframe.deletions.size() > 0) {
     // Cache latest transforms
-    std::unordered_map<RenderAssetInstanceKey, Mn::Matrix4> latestTransforms{};
     for (const auto& pair : this->createdInstances_) {
       const RenderAssetInstanceKey key = pair.first;
-      latestTransforms[key] = implementation_->getNodeTransform(pair.second);
+      latestTransformCache_[key] =
```
eundersander, line 271: You should clear this cache before every use so it doesn't grow and grow over time.

eundersander (commented):
Can you add a unit test that will trigger this code? BatchReplayRendererTest::testIntegration is close but it doesn't do any deletions.

src/esp/sim/BatchPlayerImplementation.cpp
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) Meta Platforms, Inc. and its affiliates.
```
eundersander, line 1: I assume this code was moved from `BatchReplayRenderer.cpp` without changes so I'm not looking closely at this file.

PR #2140, opened 2023-06-27
--[BE Week] Minimize logging spam

jturner65:
## Motivation and Context

The default logging behavior in habitat-sim has been overwhelming for some time now.  So much debug information is displayed to the console that any important warnings or errors get lost in the noise.  This PR is intended to address this issue.

Habitat-sim supports 4 logging levels : **VERY_VERBOSE**,  **DEBUG**, **WARNING** and **ERROR**. The intended purpose of these logging level messages is as follows :

- **ERROR** means some action or process has failed that the user needs to be aware of.
- **WARNING** means that although a process didn't outright fail, it may be yielding an unexpected/inappropriate result that the user should take note of.
- **DEBUG** means "here's the result of a process"
- **VERY_VERBOSE** means "here's the result of a process with in-depth explanation".

This PR is geared toward making adjustments to the logging behavior of habitat-sim in order to make console output less overwhelming and more useful.  The main changes are as follows : 

1. Currently Habitat-Sim sets the logging level by default to be **DEBUG**.  This level is inappropriate for general use of the simulator, and is the primary culprit for information overload.  This PR sets the default logging level to  **WARNING**  instead.  
2. Many messages in the assets, physics and metadata subsystems got demoted from warning to debug.

NOTE : Users can still modify their logging level to get more or fewer messages from the simulator by following [the guidelines on this page.](https://aihabitat.org/docs/habitat-sim/logging.html)

To give an idea of the changes in this PR, before this PR's changes, loading a Floorplanner scene would yield over 45k messages to the console.  With the changes in this PR, loading the same scene now prints 4 messasges.  Note : all of the information -can- still be printed to the console if desired (if hunting down a bug, for instance) but the user would have to specify a [more verbose logging level as in the instructions here.
](https://aihabitat.org/docs/habitat-sim/logging.html)

## How Has This Been Tested
C++ and python tests pass. The logging test needed to be slightly modified.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/Logging.h
+++ b/src/esp/core/Logging.h
@@ -122,7 +122,7 @@ namespace logging {
  * greater than or equal to the logging level within that subsystem, the log
  * will be printed.
  *
- * The convience levels, @ref Verbose, @ref Default and @ref Quiet are the primary user facing
+ * The convenience levels, @ref Verbose, @ref Default and @ref Quiet are the primary user facing
  * options.  @ref Quiet turns off logging for all but errors, @ref Default is the default and
  * shows warnings and errors, while @ref Verbose gives debug messages along with warnings and
  * errors. These names exist such that logging levels can be added and removed
--- a/src/esp/core/Logging.h
+++ b/src/esp/core/Logging.h
@@ -338,7 +338,7 @@ class LogMessageVoidify {
       espLoggingSubsystem(), esp::logging::LoggingLevel::VeryVerbose,    \
       (Corrade::Utility::Debug{Corrade::Utility::Debug::defaultOutput(), \
                                __VA_ARGS__}),                            \
-      "Message")
+      "Verbose")
 /**
  * @brief Debug level logging macro.
  */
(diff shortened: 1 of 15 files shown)


0mdc:
Awesome! Thanks for tackling this!

src/esp/core/Logging.h
```diff
@@ -122,27 +122,43 @@ namespace logging {
  * greater than or equal to the logging level within that subsystem, the log
  * will be printed.
  *
- * The convience levels, @ref Verbose and @ref Quiet are the primary user facing
- * options.  @ref Quiet turns off logging and @ref Verbose is the default and
- * shows things that are generally useful.  These names exist such that logging
- * levels can be added and removed without users needed to change what they set
- * logging to.
+ * The convience levels, @ref Verbose, @ref Default and @ref Quiet are the primary user facing
```
eundersander, line 125: *convenience

eundersander (approved):
LGTM!

0mdc (approved):
LGTM!

PR #2142, opened 2023-06-29
--[WIP][BE Week]Rework ESP_CHECK to not abort; Add ESP_FATAL to provide core dumps

jturner65:
## Motivation and Context
Currently our ESP_CHECK macro was exiting the program via std::abort(), providing a core dump. However, it is most often used to verify data or check the results of file IO, for which a core dump is undesirable.  This PR changes ESP_CHECK to std::exit(1) and instead adds a new macro ESP_FATAL that will provide a core dump on fail.

## How Has This Been Tested
Current C++ and Python tests pass

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/Bindings.cpp
+++ b/src/esp/bindings/Bindings.cpp
@@ -47,8 +47,15 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
 
   /* This function pointer is used by ESP_CHECK(). If it's null, it
      std::abort()s, if not, it calls it to cause a Python AssertionError */
-  esp::core::throwInPython = [](const char* const message) {
+  esp::core::throwAssertInPython = [](const char* const message) {
     PyErr_SetString(PyExc_AssertionError, message);
+    throw pybind11::error_already_set{};
+  };
+
+  /* This function pointer is used by ESP_FATAL(). If it's null, it
+     std::exit()s, if not, it calls it to cause a Python RuntimeError */
+  esp::core::throwRuntimeInPython = [](const char* const message) {
+    PyErr_SetString(PyExc_RuntimeError, message);
     throw pybind11::error_already_set{};
   };
 
--- a/src/esp/core/Check.h
+++ b/src/esp/core/Check.h
@@ -12,9 +12,9 @@
 /** @file
   @brief ESP_CHECK macro, for use with fatal runtime errors.
 
-  Below is an overview of asserts, ESP_CHECK, exception-throwing, and warnings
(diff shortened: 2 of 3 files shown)


src/esp/core/Check.h
```diff
@@ -51,26 +51,48 @@
...
 
+// For use in ESP_CHECK
+[[noreturn]] void throwIfInPythonOtherwiseExit(const char* message);
+
 }  // namespace core
 }  // namespace esp
 
 /* A runtime check that must pass, otherwise we consider this a fatal runtime
-error. The program terminates with the supplied error message. */
+error due to programmer error. The program terminates with the supplied error
```
eundersander, line 72: Instead of adding these comments at the definitions of ESP_FATAL and ESP_CHECK, can you rewrite the comment at the top of the file? We have to talk about all of these in one place so someone can look in one place to decide which error-handling code is appropriate.

Spoiler: I'm skeptical that there's any real difference between ESP_FATAL and assert, so I expect we'll need some discussion to arrive at a consensus here.

src/esp/bindings/Bindings.cpp
```diff
@@ -47,11 +47,18 @@ PYBIND11_MODULE(habitat_sim_bindings, m) {
...
      std::abort()s, if not, it calls it to cause a Python AssertionError */
-  esp::core::throwInPython = [](const char* const message) {
+  esp::core::throwAssertInPython = [](const char* const message) {
     PyErr_SetString(PyExc_AssertionError, message);
     throw pybind11::error_already_set{};
   };
 
+  /* This function pointer is used by ESP_FATAL(). If it's null, it
+     std::exit()s, if not, it calls it to cause a Python RuntimeError */
+  esp::core::throwRuntimeInPython = [](const char* const message) {
```
eundersander, line 57: I think the comments are backwards here. Looking at your code below, ESP_CHECK throws RuntimeError and ESP_FATAL throws AssertionError.

eundersander (commented):
I left some comments. I suggest splitting this PR into the ESP_CHECK fix (which we all agree on) and the new ESP_FATAL (I'm not convinced yet).

src/esp/bindings/Bindings.cpp
jturner65: Yep, sorry, from a previous incarnation.

src/esp/core/Check.h
jturner65: I was thinking the benefit of ESP_FATAL was that it would provide python message capability.

eundersander: Let's discuss on Slack or in an eng meeting.

PR #2177, opened 2023-08-07
--Attributes/Managers Maintenance : Separate Monolithic ObjectAttributes; Remove unnecessary abstract method; Build JSON From String

jturner65:
## Motivation and Context
This PR takes care of 4 longstanding issues with the Attributes/Managers subsystem : 
- It separates the AbstractObject, Object and Stage attributes into separate source files. These were originally combined when the attributes/managers subsystem was in a more flat layout. Breaking them apart so they are easier to manage and understand and match their respective managers' structure.
- It removes an unnecessary abstract function path that was unused by most managers. Only Stage Attributes and Object Attributes should ever need to query whether an asset handle is a valid Primitive asset handle, so this removes the query from all managers other than those.
- It adds a function to ManagedFileBasedContainer to build a Managed object/Attributes directly from an appropriately configured string without needed to load a file from disk. 
- Renames a file to match the class template it describes. -sigh-

## How Has This Been Tested
All C++/Python tests pass

## Types of changes

## Checklist


src/esp/core/managedContainers/ManagedFileBasedContainer.h
```diff
@@ -321,15 +350,15 @@ class ManagedFileBasedContainer : public ManagedContainer<T, Access> {
   }
 
   /**
-   * @brief Verify passd @p filename is legal document of type T. Returns
+   * @brief Verify passd @p filename is legal document of type U. Returns
    * loaded document in passed argument if successful. This requires
-   * appropriate specialization for each type name, so if this method is
-   * executed it means no appropriate specialization exists for passed type of
-   * document.
+   * appropriate specialization for each type name, so if this speciric method
```
aclegg3, line 355:
```suggestion
   * appropriate specialization for each type name, so if this specific method
```

aclegg3 (approved):
LGTM

PR #2181, opened 2023-08-09
--[BugFix] GibsonSemanticScene JSON access.

jturner65:
## Motivation and Context
Reading a JSON element should be check exists then read, not attempt to read then check if null, which is what the GibsonSemanticScene attempts to do.  This was failing an assertion in RapidJSON that was only visible if compiled in Debug mode; in Release RapidJSON silenced the assertion and attempted to work around the problem.

## How Has This Been Tested
All C++ and python tests (in both release and debug)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/scene/GibsonSemanticScene.cpp
+++ b/src/esp/scene/GibsonSemanticScene.cpp
@@ -69,12 +69,17 @@ bool SemanticScene::buildGibsonHouse(const io::JsonDocument& jsonDoc,
       object->category_ = std::move(category);
     }
 
-    if (jsonObject.HasMember("location")) {
-      const auto& jsonCenter = jsonObject["location"];
+    io::JsonGenericValue::ConstMemberIterator jsonLocIter =
+        jsonObject.FindMember("location");
+    if (jsonLocIter != jsonObject.MemberEnd()) {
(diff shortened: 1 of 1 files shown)


aclegg3 (approved):
Thanks for identifying the issue and fixing it despite. :rocket:

0mdc (approved):
LGTM!

src/esp/scene/GibsonSemanticScene.cpp
```diff
@@ -69,12 +69,12 @@ bool SemanticScene::buildGibsonHouse(const io::JsonDocument& jsonDoc,
       object->category_ = std::move(category);
     }
 
-    const auto& jsonCenter = jsonObject["location"];
-    if (!jsonCenter.IsNull()) {
+    if (jsonObject.HasMember("location")) {
+      const auto& jsonCenter = jsonObject["location"];
```
mosra, line 73: Hm... but `HasMember()` forces a double lookup, which isn't great. `FindMember()` might be better in both cases?

jturner65: Cool! Made the changes here for this PR. Have another PR pending making the changes in the rest of the codebase.

PR #2187, opened 2023-08-16
[BE] - Error messages for physics specific API calls without implementation

aclegg3:
## Motivation and Context

Physics API calls made without Bullet installed or active return default values. This can cause confusion for users when the code runs, but returns unexpected values. For example, it should be clear that there are no ray cast hits because physics is not available rather than because the ray did not hit anything.

There could be an argument to make these warnings instead of errors. In either case, the code executes rather than crashing. 

## How Has This Been Tested

Locally in viewer application. CI test without physics active.

## Types of changes

## Checklist


jturner65 (approved):
LGTM

PR #2189, opened 2023-08-18
--[BE/Bugfix] Early Bullet/Physics enabled check; C++ test fixes for no-bullet; Python enable_physics default change if no_bullet

jturner65:
## Motivation and Context
This PR contains a bugfix and a few updates/improvements relating to whether Bullet dynamics library is present or not.
- An ESP_CHECK was added early in Simulator::reconfigure() that would fail if the user has set their Simulator Configuration to  have physics enabled but did not compile Habitat-Sim with Bullet enabled. If this is the state of things, we want it discovered as early as possible so that the user can address it before a long training regimen is initiated.
- SimTest was updated to only set enablePhysics to true if built with Bullet, and also to test -all- tests using both SimConfig and MM-based simulator creation; furthermore, the tests that expressly rely on Bullet physics - createMagnumRenderingOff and testArticulatedObjectSkinned - are only enabled if Bullet is present and enabled. 
- PhysicsTest and ReplicaSceneTest also got similar treatment, where tests that were dependent on Bullet only were executed if built with Bullet.
- The default python sim_cfg value for enable_physics was changed from True to "habitat_sim.bindings.built_with_bullet", which will still be true for folks who have bullet but will not automatically fail for folks who do not unless they explicitly change it.
- Some python tests and examples were updated to only require Bullet/enable physics if they actually used dynamics, so that the test would still execute if Bullet was not present. Those examples that did require dynamics are now skipped properly if Bullet is not present.

## How Has This Been Tested
Locally C++ and python tests pass. Some of the python tests required modifications or skipping if bullet was not present).

## Types of changes

## Checklist


tests/test_examples.py
```diff
@@ -63,10 +83,14 @@ def powerset(iterable):
         ("examples/tutorials/async_rendering.py",),
     ],
 )
-def test_example_modules(args):
+def test_example_modules_no_bullet(args):
```
aclegg3, line 86: Naming nit. Some of these (e.g. `managed_rigid_object_tutorial`) are optional physics rather than no-physics.
```suggestion
def test_example_modules_optional_bullet(args):
```

tests/test_examples.py
```diff
@@ -63,10 +83,14 @@ def powerset(iterable):
...
     run_main_subproc(args)
 
 
+@pytest.mark.skipif(
+    not habitat_sim.bindings.built_with_bullet,
+    reason="Bullet physics required for ReplicaCAD Artciulated Objects.",
```
aclegg3, line 92:
```suggestion
    reason="Bullet physics required for ReplicaCAD Articulated Objects.",
```

aclegg3 (approved):
A few minor nits ans suggestion.
Overall, looks good. I like your idea to fill the enable_physics default from the `built_with_bullet` arg. :+1:

PR #2190, opened 2023-08-21
--[BE/Bugfix]Modify creation of Articulated Object from URDF file name.

jturner65:
All AO creation is driven by ArticulatedObjectAttributes now, with the caveat that to be used, any ArticulatedObjectAttributes must reference an existing and legitimate URDF file by relative path name from which to actually build the AO. ArticulatedObjectAttributes can be created using the name of their corresponding URDF file, in which case they will reference -that- URDF, and there is no need to complain if there is no existing ArticulatedObjectAttributes matching that name, since a legal one will be created .  This PR removes the erroneous and potentially misleading complaint, and proceeds silently to make a new ArticulatedObjectAttributes referencing the passed URDF file if one does not already exist.

This was peeled off [this PR, to minimize topical divergence.](https://github.com/facebookresearch/habitat-sim/pull/2189)

## How Has This Been Tested
All C++ and python tests pass (it's just a message change)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/PhysicsManager.cpp
+++ b/src/esp/physics/PhysicsManager.cpp
@@ -486,13 +486,17 @@ int PhysicsManager::addArticulatedObjectFromURDF(
     const std::string& lightSetup) {
   // Retrieve or create the appropriate ArticulatedObjectAttributes to create
   // this AO.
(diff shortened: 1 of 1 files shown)


aclegg3 (approved):
Thanks!

PR #2192, opened 2023-08-22
--Implementation of Nvidia's HBAO effect (soft shadows in corners and crevices) via Magnum

jturner65:
## Motivation and Context
This PR implements [Nvidia's HBAO as per the shaders found here](https://github.com/nvpro-samples/gl_ssao) but using Magnum's graphics framework.  This work was built from, and compared against, a reference implementation of the Nvidia sample code [using Nvidia's graphics framework that was integrated into Habitat-Sim in this WIP PR](https://github.com/facebookresearch/habitat-sim/pull/1997).

Both Classic and Cache-Aware implementations are available, as well as Perpective and Orthographic sensor support (driven by the projection matrix configuration of the owning visual sensor).

TODO for This PR : 
- ~~Update test images to match the working effect.~~
- ~~Gate test so that it does not attempt to run in WebGL builds~~ not necessary

TODO Next PR/Future Work : 
- Configuration support to allow easy customization of effect
- Add MSAA support
- Eventually retool and cleanup to be more generic (i.e. dispose of separate "classic" and "cache-aware" paths)
- Validate effect when accessed via Habitat-Lab

Demonstration of with/without effect via CacheAware algorithm, both Perspective and Orthographic projection.

https://github.com/facebookresearch/habitat-sim/assets/1093953/b5908c40-fc34-41aa-9ef5-073f09c44a83

## How Has This Been Tested
New C++ tests to verify integrity of effect; existing c++ and python tests pass

## Types of changes

## Checklist


0mdc:
Pretty cool! Nice job!

jturner65:
Note : currently the test images are out of date but the test is still passing, so this needs to be investigated.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,352 @@
...
+};
+
+GfxBatchHbaoTest::GfxBatchHbaoTest() {
+  addTests({&GfxBatchHbaoTest::generateTestData});
+
+  addInstancedTests({&GfxBatchHbaoTest::test},
+                    Cr::Containers::arraySize(TestData));
+}
+
+constexpr Mn::Vector2i Size{256, 256};
```
mosra, line 168: Got an idea for easy verification that the aspect ratio handling is implemented correctly: change this to something non-square (that still reasonably covers the scene, of course) and add a second test variant (`testRotated()`?) that uses `Size.flipped()` instead of `Size` and loads the images rotated by 90°:

```cpp
/* This rotates the depth image by 90° */
Cr::Containers::Array<char> rotatedDepth{Cr::NoInit, depth->data.size()};
Cr::Utility::copy(depth->pixels().flipped<1>().transposed<0, 1>(), 
    Cr::Containers::StridedArrayView2D<char>{rotatedDepth, {std::size_t(Size.y()), std::size_t(Size.x())}});

Mn::GL::Texture2D inputDepthTexture;
inputDepthTexture
    .setStorage(1, Mn::GL::TextureFormat::DepthComponent32F, Size.flipped())
    .setSubImage(0, {}, Mn::ImageView2D{depth->format(), Size.flipped(), rotatedDepth});
```

Similarly for the color image. The final comparison would be to the original image, but with the pixels rotated back:

```cpp
    CORRADE_COMPARE_WITH(
            output.read({{}, Size}, {Mn::PixelFormat::RGBA8Unorm})
                .pixels<Color4ub>().transposed<1, 0>().flipped<1>(),
            Cr::Utility::Path::join(TEST_ASSETS, data.filename),
            (Mn::DebugTools::CompareImageToFile{data.maxThreshold, data.meanThreshold}));
```

If everything is correct in the implementation, the test should pass without any difference compared to the non-rotated case.

(Details about strided array views and their dimension flipping and transposition are in [this blog post](https://blog.magnum.graphics/backstage/multidimensional-strided-array-views/).)

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,352 @@
...
+        .setTransformationMatrix(
+            Mn::Matrix4::translation({0.0f, -1.0f, -7.5f}) *
+            Mn::Matrix4::rotationX(-80.0_degf) *
+            Mn::Matrix4::rotationZ(-90.0_degf))
+        .setLightPositions({
+            {10.0f, 10.0f, 10.0f, 0.0f},
+            {-5.0f, -5.0f, 10.0f, 0.0f},
+            {0.0f, 5.0f, -10.0f, 0.0f},
+        })
+        .setLightColors({0xffffff_rgbf, 0xffdddd_rgbf, 0xddddff_rgbf})
```
mosra, line 256: Given all the code duplication here, this test would maybe be better instanced, similarly as the other one? With this as input:

```cpp
const struct {
    const char* name;
    const char* filenameSuffix;
    Mn::Matrix4 projection;
} GenerateTestDataData[]{
    {"perspective", "", 
        Mn::Matrix4::perspectiveProjection(45.0_degf, 1.0f, 0.1f, 100.0f)},
    {"orthographic", "-ortho",
        Mn::Matrix4::orthographicProjection(size, 0.1f, 100.0f)},
};
```

src/tests/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 find_package(Magnum REQUIRED DebugTools AnyImageConverter AnySceneImporter)
 find_package(
   MagnumPlugins REQUIRED GltfImporter StbImageImporter KtxImporter
-  OPTIONAL_COMPONENTS GltfSceneConverter KtxImageConverter
+  OPTIONAL_COMPONENTS GltfSceneConverter KtxImageConverter StanfordImporter
```
mosra, line 8: Can you add `OpenExrImporter`, `OpenExrImageConverter` here, and `AnyImageImporter` to the `find_package(Magnum)` above? While it has no effect for the default build with Magnum bundled, the build with external Magnum would break without it.

src/shaders/gfx_batch/hbao/bilateralblur.frag
```diff
@@ -0,0 +1,71 @@
+/*
```
mosra, line 1: Just so we don't forget: run all these through

```sh
magnum-shaderconverter --validate --input-version "300 es" --output-version opengl <shader-file>
```

and fix the issues it complains about, so the initial version is without WebGL shader compiler errors. May also need to add various `-D` definitions depending on the shader, such as `-DAO_LAYERED=1 -DAO_INTERLEAVED=1` etc. for `hbao.frag`.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,352 @@
...
+void GfxBatchHbaoTest::generateTestData() {
+  Cr::PluginManager::Manager<Mn::Trade::AbstractImporter> importerManager;
+  Cr::Containers::Pointer<Mn::Trade::AbstractImporter> importer =
+      importerManager.loadAndInstantiate("AnySceneImporter");
+  CORRADE_VERIFY(importer);
+
+  /* magnum-sceneconverter van-gogh-room.glb --concatenate-meshes
+   * van-gogh-room.mesh.ply */
+  CORRADE_VERIFY(importer->openFile(Cr::Utility::Path::join(
+      TEST_ASSETS, "hbao_tests/van-gogh-room.mesh.ply")));
```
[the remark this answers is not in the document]
jturner65: eh, 500 kb isn't that bad, is it? :)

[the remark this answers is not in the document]
mosra: It's more about the principle / general workflow. The original file wasn't commited, so I didn't put this one here either. And (again), the file isn't needed for verifying the HBAO workflow, it's only needed once, to generate the input data locally. In other words, the test-data-generation case could even be a completely separate executable / python script that isn't being run on the CI at all, but this way it felt to me that it's less prone to code rot and getting out of sync with the actual images.

With the "isn't that bad" mindset the `van-gogh-room.glb` could be in the repo already instead of being downloaded separately, allowing the single-joined-mesh variant of to be generated by the test on-the-fly. Which would be the cleanest way. Commiting just the derived file and not the original doesn't make a lot of sense to me.

src/esp/bindings/SimBindings.cpp
```diff
@@ -96,6 +96,9 @@ void initSimBindings(py::module& m) {
       .def_readwrite(
           "requires_textures", &SimulatorConfiguration::requiresTextures,
           R"(Whether or not to load textures for the meshes. This MUST be true for RGB rendering.)")
+      .def_readwrite("hbao_visual_effects",
+                     &SimulatorConfiguration::horizonBasedAmbientOcclusion,
+                     R"(Whether or not to enable HBAO visual effects.)")
```
0mdc, line 101: Nit: I would rename this `enable_hbao`.

src/esp/gfx_batch/Hbao.h
```diff
@@ -0,0 +1,180 @@
...
+  Hbao(const Hbao&) = delete;
+  Hbao(Hbao&&) noexcept;
+
+  ~Hbao();
+
+  Hbao& operator=(const Hbao&) = delete;
+  Hbao& operator=(Hbao&&) noexcept;
+
+  /**
+   * @brief Set the configurable quantites of the HBAO alogirthm based on user
```
0mdc, line 139: Typo: `quantites` and `alogirthm`

src/esp/gfx_batch/Hbao.h
```diff
@@ -0,0 +1,180 @@
...
+  // least until we support MSAA
+
+  // HbaoConfiguration& setUseBlur(bool state) {
+  //   // special blur is dependent on blur being enabled, so both should be
+  //   // disabled if blur is.
+  //   flags_ = (state ? flags_ | HbaoFlag::Blur
+  //                   : flags_ & ~(HbaoFlag::UseAoSpecialBlur |
+  //                   HbaoFlag::Blur));
+  //   return *this;
+  // }
```
0mdc, line 55: Should this be uncommented?

src/esp/gfx_batch/Hbao.cpp
```diff
@@ -0,0 +1,1069 @@
...
+    }
+
+  uniform.setSubData(0, {&uniformData, 1});
+}
+
+Mn::Vector4 buildClipInfo(const Mn::Matrix4& projectionMatrix,
+                          bool orthographic) {
+  // TODO this still looks like there's some better way this should be
+  // extracted.  Definitely change to use near and far calcs upstream when
+  // available!!
```
0mdc, line 829: 👍

src/esp/gfx_batch/Hbao.cpp
```diff
@@ -0,0 +1,1069 @@
...
+
+    // TODO use our own
+    Mn::GL::Shader vert{GlslVersion, Mn::GL::Shader::Type::Vertex};
+    vert.addSource(rs.getString("hbao/fullscreenquad.vert"));
+
+#ifndef MAGNUM_TARGET_WEBGL
+    const bool passthroughSupported =
+/* Needs
+ * https://github.com/mosra/magnum/commit/5f287df332fee7fcc2df6e8055853b86d1db3ed2
+ */
```
0mdc, line 275: Does our current Magnum version support this?

0mdc (commented):
Did you have a change to measure rendering performance with/without HBAO?

src/esp/gfx_batch/Hbao.cpp
mosra: No, and that's why I put that comment here in the first place :) I need to make a Magnum update PR first, but I'm delaying that to once I'm back in the project as the set of changes since last update is rather nontrivial.

Same case is for the near/far plane extraction.

src/esp/gfx_batch/Hbao.cpp
mosra: Might want to reference this commit in the TODO: https://github.com/mosra/magnum/commit/0d31f7461b31698ea5bf92ec66ff5056a6ad7360

src/esp/gfx_batch/Hbao.h
mosra: > Blur should always be used, otherwise there's nasty grid-like artifacts, at least until we support MSAA

@jturner65 that's, as far as I understood, how it should be (the blur is what smoothens out the grid-like noise), and MSAA won't fix it. Ability to disable the blur is there for debugging purposes.

src/esp/gfx_batch/Hbao.h
```diff
@@ -0,0 +1,180 @@
...
+                    : flags_ & ~HbaoFlag::LayeredImageLoadStore);
+    return *this;
+  }
+  HbaoConfiguration& setUseLayeredGeometryShader(bool state) {
+    // LayeredImageLoadStore and LayeredGeometryShader are mutually exclusive
+    flags_ = (state ? (flags_ | HbaoFlag::LayeredGeometryShader) &
+                          ~HbaoFlag::LayeredImageLoadStore
+                    : flags_ & ~HbaoFlag::LayeredGeometryShader);
+    return *this;
+  }
```
mosra, line 78: I don't know, wouldn't it be easier to just make the flags supersets of one another? I didn't do that originally because it required extra brain cycles, but

```cpp
enum class HbaoFlag {
  Blur = 1 << 0,
  UseAoSpecialBlur = (1 << 1)|Blur,
  LayeredImageLoadStore = 1 << 2,
  LayeredGeometryShader = (1 << 3)|LayeredImageLoadStore
};
```

should cover pretty much all these cases without needing extra setters. I.e., if you enable `UseAoSpecialBlur`, it'll add `Blur` as well, if you enable `LayeredImageLoadStore|LayeredGeometryShader` it'll pick the GS as "higher priority" (assuming the passthrough is faster than image load/store). So it's then just a matter of documenting the relations.

With this you of course need to test for the flags with `>=` instead of `&` in the code, but that's a simple change.

src/esp/gfx_batch/Hbao.cpp
```diff
@@ -0,0 +1,1069 @@
...
+  setConfiguration(configuration);
+}
+
+Hbao::Hbao(Hbao&&) noexcept = default;
+
+Hbao::~Hbao() = default;
+
+Hbao& Hbao::operator=(Hbao&&) noexcept = default;
+
+void Hbao::setConfiguration(const HbaoConfiguration& configuration) {
```
mosra, line 603: Wait a moment, I thought we agreed that you'd just replace the instance when configuration changes? This setter is only adding extra potential for bugs as it may not reininitialize all internal state properly.

In other words, this is what I mean:

```cpp
Hbao hbao;

...

// when settings change
hbao = Hbao{newHbaoConfiguration};
```

That way the instance is, when constructed, always correct, and there's no chance some state may be stale because `setConfiguration()` forgot to overwrite something that was only populated in a constructor.

For changing actual runtime parameters (as we discussed), there should be direct setters instead, separate from the `HbaoConfiguration` class which should have only the initial setup.

src/esp/gfx_batch/Hbao.cpp
```diff
@@ -0,0 +1,1069 @@
...
+  HbaoUniformData hbaoUniformData;
+
+  Mn::GL::Mesh triangle, triangleLayered;
+
+  HbaoConfiguration configuration;
+
+  Mn::Vector4 random[HbaoRandomNumElements * MaxSamples];
+};
+
+Hbao::Hbao(Mn::NoCreateT) noexcept {}
```
mosra, line 586: Drop this, it's no longer used as the Renderer has an Optional instead, and it's potentially dangerous.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+  Mn::Matrix4 projMatrices[2] = {
+      Mn::Matrix4::perspectiveProjection(45.0_degf, 1.0f, 0.1f, 100.0f),
+      Mn::Matrix4::orthographicProjection(Magnum::Vector2(4.0f, 4.0f),
+                                          0.1f,
+                                          100.0f)};
+  std::string srcRGBFileNames[2] = {"van-gogh-room.color.png",
+                                    "van-gogh-room.color-ortho.png"};
+
+  std::string srcDepthFileNames[2] = {"van-gogh-room.depth.exr",
```
mosra, line 48: `std::string` :scream: 

Also, these could all be a file-local struct, and the instanced data could then reference them:

```cpp
struct Projection {
    Mn::Matrix4 projection;
    const char* sourceColorFilename;
    const char* sourceDepthFilename;
};

const Projection PerspectiveData{
    ...
};

const Projection OrthographicData{
    ...
};
```

...

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,317 @@
...
+                                      "van-gogh-room.depth-ortho.exr"};
+};
+
+const struct {
+  const char* name;
+  const char* filename;
+  bool classic;
+  int projIDX;
```
mosra, line 56: ... and then you'd have 

```suggestion
  const Projection& projection;
```

here, and below it'll be either `PerspectiveData` or `ProjectionData` instead of `0` or `1`, and similar elsewhere.

mosra (commented):
Accidentally opened a GitHub notification so had to go through the code and write comments :)

jturner65:
> Did you have a change to measure rendering performance with/without HBAO?

Not yet, I am going to investigate that today.

src/tests/GfxBatchHbaoTest.cpp
jturner65: Yeah, thanks, that's much better

src/shaders/gfx_batch/hbao/bilateralblur.frag
jturner65: This has been completed, and with the exception of the geometry shader, they've all been made compatible.

src/esp/gfx_batch/Hbao.h
jturner65: I'm going to change these as we discussed, with a "no blur" flag that turns off all blur, and the UseAOSpecialBlur flag to use the special blur, with the non-special blur being the default behavior.

src/esp/gfx_batch/Hbao.cpp
jturner65: I'll rework this when I work on the user-driven attributes-based customization functionality.

src/tests/GfxBatchHbaoTest.cpp
jturner65: Put these tests in and they are exposing a bias in the algorithm that seems like it exists in the shader, so I have marked them Expect to Fail and will investigate this further. The effect is present and visually pleasing, however, so I don't think this should hold up the PR.

src/tests/GfxBatchHbaoTest.cpp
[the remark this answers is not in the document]
jturner65: Ply has been removed.

aclegg3 (approved):
Looks good for initial merge, thanks @jturner65 and @mosra!
We can follow-up with additional polish.

src/esp/gfx_batch/Hbao.cpp
```diff
@@ -0,0 +1,1088 @@
...
+
+  Mn::Vector2 projScale;
+  Mn::Int projOrtho;
+  Mn::Int _pad1;
+
+  Mn::Vector4 float2Offsets[AoRandomTextureSize * AoRandomTextureSize];
+  Mn::Vector4 jitters[AoRandomTextureSize * AoRandomTextureSize];
+};
+
+static_assert(sizeof(HbaoUniformData) % 4 == 0, "Not a nice uniform struct");
```
mosra, line 545: Dang, this should have been 16

src/esp/gfx_batch/Hbao.cpp
```diff
@@ -0,0 +1,1088 @@
...
+    state_->viewNormal.attachTexture(Mn::GL::Framebuffer::ColorAttachment{0},
+                                     state_->sceneViewNormal, 0);
+
+    const Mn::GL::TextureFormat aoFormat =
+        configuration.flags() & HbaoFlag::UseAoSpecialBlur
+            ? Mn::GL::TextureFormat::RG16F
+            : Mn::GL::TextureFormat::R8;
+    state_
+        ->hbaoResult
+        // TODO filter?!
```
mosra, line 692: Obsolete TODOs, below also. I had marked these as I expected some issues, but there weren't any in the end.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,413 @@
...
+  esp::gfx_batch::HbaoFlags flags;
+  Mn::Float intensity, radius, blurSharpness;
+  Mn::Float maxThreshold, meanThreshold;
+} TestData[]{
+    // Perspective projection
+    {"classic, defaults", "hbao-classic", true,
+     esp::gfx_batch::HbaoConfiguration{}.flags(),
+     esp::gfx_batch::HbaoConfiguration{}.intensity(),
+     esp::gfx_batch::HbaoConfiguration{}.radius(),
+     esp::gfx_batch::HbaoConfiguration{}.blurSharpness(), 0.0f, 0.0f},
```
mosra, line 109: Heh, these intensity / radius / shaprness copypastes are kinda annoying. What about just storing a "multiplier" here, have it `1.0f` most of the time, and then multiply in the test directly?

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,413 @@
...
+    {"cache-aware, layered with geometry shader, AO special blur",
+     "hbao-cache-geom-sblur", false,
+     esp::gfx_batch::HbaoConfiguration{}.flags() |
+         (esp::gfx_batch::HbaoFlag::LayeredGeometryShader |
+          esp::gfx_batch::HbaoFlag::UseAoSpecialBlur),
+     esp::gfx_batch::HbaoConfiguration{}.intensity(),
+     esp::gfx_batch::HbaoConfiguration{}.radius(),
+     esp::gfx_batch::HbaoConfiguration{}.blurSharpness(), 1.0f, 0.1f},
```
mosra, line 169: As we discussed previously, might make sense to add the GS + image load/store variants again but compare them to the default `"hbao-cache"` images with a higher threshold, to be sure they don't diverge too much.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+namespace {
+
+namespace Cr = Corrade;
+namespace Mn = Magnum;
+using namespace Mn::Math::Literals;
+
+constexpr const char* baseTestFilename = "van-gogh-room";
+
+const std::string SourceTestPlyDir =
```
mosra, line 40: Just FYI, `Utility::Path::join()`, `Utility::format()` all return `Containers::String`, and the APIs taking strings use `StringView`s.

Not that I'd want that changed away from `std::string` across all of Habitat, but existing code in `gfx_batch` is using the Corrade ones exclusively so it'd be great if it was kept consistent. (I'm probably just annoying at this point, sorry.)

src/esp/gfx_batch/Hbao.cpp
jturner65: was going to ask about that.

src/tests/GfxBatchHbaoTest.cpp
jturner65: Not at all. I wonder about these things, but I don't want to overly annoy you with questions about these little things. :)

src/esp/gfx_batch/Hbao.h
```diff
@@ -0,0 +1,178 @@
...
```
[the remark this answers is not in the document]
jturner65: Actually was going to get rid of setFlags(), since we  have dependent/mutually exclusive states that need to be properly tracked. LayeredImage and LayeredGeometry are mutually exclusive, so I don't think they can be subsets of each other.

0mdc (approved):
The current state looks good to me!

src/tests/GfxBatchHbaoTest.cpp
jturner65: Yeah, rearranged that stuff now to be a bit less wordy.

src/tests/GfxBatchHbaoTest.cpp
jturner65: I'll revisit this in HBAO version 1.1 :)

mosra (approved):

PR #2199, opened 2023-09-07
Add option to toggle frustum culling in replay renderer.

0mdc:
## Motivation and Context

This changeset adds a config option to disable culling for replay renderers.

The classic replay renderer currently has an issue related to frustum culling. This allows to work-around the issue for immediate needs.

## How Has This Been Tested

Tested locally in Python viewer.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -379,6 +379,9 @@ void initSimBindings(py::module& m) {
           R"(List of sensor specifications for one simulator. For batch rendering, all simulators must have the same specification.)")
       .def_readwrite("gpu_device_id", &ReplayRendererConfiguration::gpuDeviceId,
                      R"(The system GPU device to use for rendering)")
+      .def_readwrite("enable_frustum_culling",
+                     &ReplayRendererConfiguration::enableFrustumCulling,
+                     R"(Controls whether frustum culling is enabled.)")
       .def_readwrite(
           "force_separate_semantic_scene_graph",
           &ReplayRendererConfiguration::forceSeparateSemanticSceneGraph,
--- a/src/esp/sim/AbstractReplayRenderer.h
+++ b/src/esp/sim/AbstractReplayRenderer.h
@@ -55,6 +55,8 @@ class ReplayRendererConfiguration {
    */
   bool leaveContextWithBackgroundRenderer = false;
 
+  bool enableFrustumCulling = true;
+
   std::vector<std::shared_ptr<sensor::SensorSpec>> sensorSpecifications;
 
   ESP_SMART_POINTERS(ReplayRendererConfiguration)
--- a/src/esp/sim/ClassicReplayRenderer.cpp
+++ b/src/esp/sim/ClassicReplayRenderer.cpp
@@ -236,11 +236,14 @@ void ClassicReplayRenderer::doRender(
       }
 
 #ifdef ESP_BUILD_WITH_BACKGROUND_RENDERER
+      esp::gfx::RenderCamera::Flags flags{};
+      if (config_.enableFrustumCulling) {
+        flags |= gfx::RenderCamera::Flag::FrustumCulling;
+      }
+
       if (imageViews.size() > 0) {
-        renderer_->enqueueAsyncDrawJob(
-            visualSensor, sceneGraph, imageViews[envIndex],
-            esp::gfx::RenderCamera::Flags{
-                gfx::RenderCamera::Flag::FrustumCulling});
+        renderer_->enqueueAsyncDrawJob(visualSensor, sceneGraph,
+                                       imageViews[envIndex], flags);
       }
 #else
       // TODO what am I supposed to do here?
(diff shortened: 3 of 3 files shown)


src/esp/bindings/SimBindings.cpp
```diff
@@ -379,6 +379,9 @@ void initSimBindings(py::module& m) {
           R"(List of sensor specifications for one simulator. For batch rendering, all simulators must have the same specification.)")
       .def_readwrite("gpu_device_id", &ReplayRendererConfiguration::gpuDeviceId,
                      R"(The system GPU device to use for rendering)")
+      .def_readwrite("enable_frustum_culling",
```
jturner65, line 382: Perhaps we can rename this to something like 'force_culling_off' or 'force_disable_culling' so there's no ambiguity in its purpose.

src/esp/sim/AbstractReplayRenderer.h
```diff
@@ -55,6 +55,8 @@ class ReplayRendererConfiguration {
    */
   bool leaveContextWithBackgroundRenderer = false;
 
+  bool enableFrustumCulling = true;
```
jturner65, line 58: disableFrustumCulling ?

src/esp/sim/ClassicReplayRenderer.cpp
```diff
@@ -236,11 +236,14 @@ void ClassicReplayRenderer::doRender(
       }
 
 #ifdef ESP_BUILD_WITH_BACKGROUND_RENDERER
+      esp::gfx::RenderCamera::Flags flags{};
+      if (config_.enableFrustumCulling) {
+        flags |= gfx::RenderCamera::Flag::FrustumCulling;
+      }
```
jturner65, line 243: Probably want to change this so that if disableFrustumCulling is true then it overrides whatever the config specifies.

src/esp/sim/ClassicReplayRenderer.cpp
```diff
@@ -258,6 +261,10 @@ void ClassicReplayRenderer::doRender(
 void ClassicReplayRenderer::doRender(
     Magnum::GL::AbstractFramebuffer& framebuffer) {
   const Mn::Vector2i gridSize = environmentGridSize(config_.numEnvironments);
+  esp::gfx::RenderCamera::Flags flags{};
+  if (config_.enableFrustumCulling) {
+    flags |= gfx::RenderCamera::Flag::FrustumCulling;
+  }
```
jturner65, line 268: Force it off here too.

src/esp/bindings/SimBindings.cpp
eundersander: My two cents: a flag named "enable..." is more intuitive than "disable...". I prefer Mikael's name here. Of course it should default to true for this feature.

src/esp/sim/ClassicReplayRenderer.cpp
eundersander: I don't understand this comment. There's `config_.ourNewCullingFlag` (name TBD). Is there also another config that is relevant here?

eundersander (approved):
LGTM but Mikael and John should reach consensus before merging.

src/esp/bindings/SimBindings.cpp
jturner65: The purpose is to force it off if it isn't working, though, right?

src/esp/sim/ClassicReplayRenderer.cpp
jturner65: From what I understand this PR is providing a RenderReplay-specific flag to disable culling if necessary, regardless of SimulatorConfiguration setting, but perhaps my understanding is wrong.

jturner65:
> LGTM but Mikael and John should reach consensus before merging.

If this works how it is needed, then I'm fine with it. It just seems kind of confusing to me.  

If the purpose of this new flag to **enable** frustum culling in replay regardless of simConfig setting then I think the naming and structure is fine
If the purpose of the new flag is to **disable** culling in replay regardless of simConfig setting then I would rename it.
If the purpose is to provide independent Replay control of culling, then the name is also fine but then SimConfig's setting should be ignored.

src/esp/bindings/SimBindings.cpp
0mdc: The `SimulatorConfiguration` semantic is the same than the one suggested here.

https://github.com/facebookresearch/habitat-sim/blob/main/src/esp/sim/SimulatorConfiguration.h#L36
```
//! Enable or disable the frustum culling optimisation
bool frustumCulling = true;
```

src/esp/sim/ClassicReplayRenderer.cpp
jturner65: After discussing with Mikael, I see that SimConfig's role in this process with ReplayRenderer is not as relevant as I had initially thought (not relevant at all, in fact :) ), so ignore my suggestions.

jturner65:
After discussing with Mikael, I see that SimConfig's role in this process with ReplayRenderer is not as relevant as I had initially thought (not relevant at all, in fact :) ), so many of my suggestions are equally irrelevant :)

jturner65 (approved):

PR #2212, opened 2023-09-25
--Add benchmarking for HBAO effect.

jturner65:
## Motivation and Context
This PR adds benchmarking for the HBAO effect for various permutations of the various options available.

Here's how various configurations are performing (thanks @mosra for the great graphic)

HBAOPerformance

## How Has This Been Tested
C++ tests pass.

## Types of changes

## Checklist

### DIFF
--- a/src/tests/GfxBatchHbaoTest.cpp
+++ b/src/tests/GfxBatchHbaoTest.cpp
@@ -56,7 +56,7 @@ const Cr::Containers::String testHBAOImageDir =
 constexpr Mn::Vector2i Size{320, 240};
 
 // Aspect ratio of given size
-constexpr const float AspectRatio = float(Size.x()) / Size.y();
+const float AspectRatio = Mn::Vector2{Size}.aspectRatio();
 
 struct Projection;
 struct TestDataType;
@@ -64,8 +64,6 @@ struct TestDataType;
 struct GfxBatchHbaoTest : Mn::GL::OpenGLTester {
   explicit GfxBatchHbaoTest();
   void generateTestData();
-
-  Cr::Containers::Optional<Mn::GL::Texture2D> loadImageIntoTexture();
 
   /// @brief Validation Tests ///
 
@@ -76,7 +74,7 @@ struct GfxBatchHbaoTest : Mn::GL::OpenGLTester {
    * @param proj The projection matrix this test consumes
    */
   void testHBAOData(const TestDataType& data,
-                    const Cr::Containers::String& filename,
+                    Cr::Containers::StringView filename,
                     const Projection& proj);
 
   /**
@@ -87,7 +85,7 @@ struct GfxBatchHbaoTest : Mn::GL::OpenGLTester {
    * @param proj The projection matrix this test consumes
    */
   void testFlippedHBAOData(const TestDataType& data,
-                           const Cr::Containers::String& filename,
+                           Cr::Containers::StringView filename,
                            const Projection& proj);
 
   /**
@@ -334,7 +332,7 @@ void GfxBatchHbaoTest::generateTestData() {
 }  // GfxBatchHbaoTest::generateTestData()
 
 void GfxBatchHbaoTest::testHBAOData(const TestDataType& data,
-                                    const Cr::Containers::String& filename,
+                                    Cr::Containers::StringView filename,
                                     const Projection& projData) {
   if ((data.config.flags() & esp::gfx_batch::HbaoFlag::LayeredImageLoadStore) &&
       !(
@@ -389,8 +387,8 @@ void GfxBatchHbaoTest::testHBAOData(const TestDataType& data,
 
   MAGNUM_VERIFY_NO_GL_ERROR();
 
-  auto config{data.config};
-  esp::gfx_batch::Hbao hbao{config.setSize(Size)};
+  esp::gfx_batch::Hbao hbao{
+      esp::gfx_batch::HbaoConfiguration{data.config}.setSize(Size)};
   MAGNUM_VERIFY_NO_GL_ERROR();
   // test projection drawing for both classic and cache-aware algorithms
 
@@ -406,10 +404,9 @@ void GfxBatchHbaoTest::testHBAOData(const TestDataType& data,
 
 }  // GfxBatchHbaoTest::testHBAOData
 
-void GfxBatchHbaoTest::testFlippedHBAOData(
-    const TestDataType& data,
-    const Cr::Containers::String& filename,
-    const Projection& projData) {
+void GfxBatchHbaoTest::testFlippedHBAOData(const TestDataType& data,
+                                           Cr::Containers::StringView filename,
+                                           const Projection& projData) {
   if ((data.config.flags() & esp::gfx_batch::HbaoFlag::LayeredImageLoadStore) &&
       !(
 #ifdef MAGNUM_TARGET_GLES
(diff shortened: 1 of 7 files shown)


src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -438,53 +503,130 @@ void GfxBatchHbaoTest::testFlippedHBAOData(const TestDataType& data,
...
+  // Scale image size to use for benchmark uniformly so projection matrix aspect
+  // ration not affected.
+  Mn::Vector2i BenchImageSize = Size * 4;
+  const size_t arraySize = BenchImageSize.x() * BenchImageSize.y() * 8;
+
+  // For benchmarks source color and depth are just empty white images
+
+  Cr::Containers::Array<char> depthData{Cr::DirectInit, arraySize, '\xff'};
+  Mn::Trade::ImageData2D depth{Mn::PixelFormat::Depth32F, BenchImageSize,
+                               std::move(depthData)};
```
mosra, line 565: If you use a view, you don't need the move:

```suggestion
  Mn::Trade::ImageDataView2D depth{Mn::PixelFormat::Depth32F, BenchImageSize,
                               depthData};
```

I often even make the image view temporary in the `setSubImage()` call, not sure what clang-format makes out of that tho.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -438,53 +503,130 @@ void GfxBatchHbaoTest::testFlippedHBAOData(const TestDataType& data,
...
+              .isExtensionSupported<
+                  Mn::GL::Extensions::ARB::shader_image_load_store>()
+#endif
+              ))
+    CORRADE_SKIP("Image load/store not supported");
+
+  // Scale image size to use for benchmark uniformly so projection matrix aspect
+  // ration not affected.
+  Mn::Vector2i BenchImageSize = Size * 4;
+  const size_t arraySize = BenchImageSize.x() * BenchImageSize.y() * 8;
```
mosra, line 559: FYI, there's this too:

```suggestion
  const size_t arraySize = BenchImageSize.product() * 8;
```

Hm, and I think you only need `4`, both the depth and color image has a four-pixel component type.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -39,70 +39,130 @@ namespace Cr = Corrade;
...
+  Cr::Containers::Optional<Mn::GL::Texture2D> loadImageIntoTexture();
+
+  /// @brief Validation Tests ///
 
+  /**
+   * @brief Test standard orientation HBAO effect
+   * @param data The pertinent test data for the test
+   * @param filename The name of the validation file to test against
+   * @param proj The projection matrix this test consumes
+   */
```
mosra, line 77: I wouldn't bother with Doxygen commands here. The tool doesn't see them anyway :)

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -438,53 +503,130 @@ void GfxBatchHbaoTest::testFlippedHBAOData(const TestDataType& data,
...
+  Mn::GL::Framebuffer output{{{}, BenchImageSize}};
+  output.attachTexture(Mn::GL::Framebuffer::ColorAttachment{0},
+                       outputColorTexture, 0);
+
+  /* No clear, that would kill the base image */
+
+  MAGNUM_VERIFY_NO_GL_ERROR();
+
+  auto config{data.config};
+  esp::gfx_batch::Hbao hbao{config.setSize(BenchImageSize)};
```
mosra, line 592:
```suggestion
  esp::gfx_batch::Hbao hbao{esp::gfx_batch::HbaoConfiguration{config}.setSize(BenchImageSize)};
```

Is a terser way to do the same, if you feel that's still readable.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -438,53 +503,130 @@ void GfxBatchHbaoTest::testFlippedHBAOData(const TestDataType& data,
...
+      output.read({{}, BenchImageSize}, {Mn::PixelFormat::RGBA8Unorm}),
+      resultImage,
+      (Mn::DebugTools::CompareImage{data.maxThreshold, data.meanThreshold}));
+
+}  // GfxBatchHbaoTest::testHBAOData
+
+void GfxBatchHbaoTest::benchmarkPerspective() {
+  auto&& data = BenchData[testCaseInstanceId()];
+  setTestCaseDescription(
+      Cr::Utility::format("{}, perspective benchmark.", data.name));
```
mosra, line 614:
```suggestion
      Cr::Utility::format("{}, perspective", data.name));
```

Redundant, same below.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -39,70 +39,130 @@ namespace Cr = Corrade;
...
 
+  /**
+   * @brief Test standard orientation HBAO effect
+   * @param data The pertinent test data for the test
+   * @param filename The name of the validation file to test against
+   * @param proj The projection matrix this test consumes
+   */
   void testHBAOData(const TestDataType& data,
-                    const std::string& filename,
+                    const Cr::Containers::String& filename,
```
mosra, line 79: Since you're making the changes already, function arguments should always be just string views (as opposed to return values, which usually need to be full owned strings). And it's a small view (just two pointers) so it can be passed as a value:

```suggestion
                    Cr::Containers::StringView filename,
```

Same below.

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -39,70 +39,130 @@ namespace Cr = Corrade;
...
+struct TestDataType;
 
 struct GfxBatchHbaoTest : Mn::GL::OpenGLTester {
   explicit GfxBatchHbaoTest();
   void generateTestData();
-  void testPerspective();
 
-  void testPerspectiveFlipped();
-  void testOrthographic();
+  Cr::Containers::Optional<Mn::GL::Texture2D> loadImageIntoTexture();
```
mosra, line 68: Unused?

src/tests/GfxBatchHbaoTest.cpp
```diff
@@ -39,70 +39,130 @@ namespace Cr = Corrade;
...
+// TODO : consider moving these images to habitat_test_secenes repository.
+const Cr::Containers::String testHBAOImageDir =
     Cr::Utility::Path::join(TEST_ASSETS, "hbao_tests");
 
+// Size of images to test for validation
 constexpr Mn::Vector2i Size{320, 240};
 
-constexpr const float aspectRatio = float(Size.x()) / Size.y();
+// Aspect ratio of given size
+constexpr const float AspectRatio = float(Size.x()) / Size.y();
```
mosra, line 59: Floating-point `Vector2` has [aspectRatio()](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Vector2.html#a4f097e00b03c17ffe212da2c8371d614). Though it isn't `constexpr` for some reason, but that doesn't really matter in a test.

```suggestion
const float AspectRatio = Mn::Vector2{Size}.aspectRatio();
```

src/tests/GfxBatchHbaoTest.cpp
jturner65: Just looking out for my future self :D

mosra: I mean, comments yes, for the human reader, but `@brief` and `@param` I wouldn't bother :)

mosra (approved):
Thanks!

PR #2218, opened 2023-09-27
--[BUGFIX] -Ellipses in paths

jturner65:
## Motivation and Context
If an ellipsis is specified in a path and this ellipsis attempts to back-step across an OS link, functions that attempt to read this path will get lost. On such function we use is **glob()**.  Such path configurations may be present in our scene dataset configurations, where the user might wish to access assets from another installed dataset that requires navigating to their directory outside the current dataset directory.

This PR adds a function that recursively removes ellipses if it can remove their parent directory (so a path with leading ellipses would be ignored). Tests are also added to verify functionality, and the path data being used for glob calls is now being appropriately filtered, albeit only in the case where the path within the scene dataset config path list starts with leading ellipses (in order to access assets or configs in another dataset, for instance). 

For a future TODO I want to normalize all the metadata paths in the system, but that is outside the scope of what I need to do for Habitat 3.0 right now.

## How Has This Been Tested
C++ and python tests pass locally (c++ test cases have been added to test the functionality)

## Types of changes

## Checklist

### DIFF
--- a/src/esp/io/Io.cpp
+++ b/src/esp/io/Io.cpp
@@ -23,8 +23,33 @@ std::string changeExtension(const std::string& filename,
 
 }  // changeExtension
 
+std::string filterPath(const std::string& srcPath) {
+  std::size_t ellipsisLoc = srcPath.find("/../");
+  if ((ellipsisLoc == std::string::npos) || (ellipsisLoc == 0)) {
+    // not found
+    return srcPath;
+  }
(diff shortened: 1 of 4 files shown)


0mdc:
Nice! This was also causing caching issues in replay rendering, as the same file would sometimes be instantiated from different relative paths.

aclegg3 (approved):

0mdc (approved):
LGTM!

PR #2221, opened 2023-09-29
More accurate get_random_navigable_point_near

aclegg3:
## Motivation and Context

**TL;DR:** This PR adds a new implementation for sampling random points within a circle.

### Context:
Currently, we use a Detour function `findRandomPointAroundCircle` for these queries. This function is implemented with a breadth-first search over navmesh polygons, excluding polygons outside the circle, starting from the closest snap point. 
**The problem with this function is that it excludes regions of the navmesh which are inside the circle, but separated from the region containing the start point.**

For example, this image shows points resulting from the query within the circle centered on the red object. Note that only bed points are considered because the "ramp" off the bed is outside the circle, cutting it off from the rest of that room and the other rooms.

image

To fix this behavior and accurately sample all points within the circle, two options are considered:
1. Change Detour to allow the breadth-first search to continue outside the circle by removing [this check](https://github.com/recastnavigation/recastnavigation/blob/cd898904b72a300011fbb24d578620bafa08ef2c/Detour/Source/DetourNavMeshQuery.cpp#L429).
2. (proposed) Implement our own filter function to remove polygons outside the circle and then use a standard sample query with rejection for distance. 

This PR implements option 2 for better efficiency. See performance notes below for details.

### Performance Notes:
The previous method was very fast because it incorrectly only touches polygons connected to the start point and within the circle. Any fix will be inherently slower because it must consider the entire navmesh (or island) to correctly filter candidate polygons.

Option (1) above is empirically about **100x slower** than the original approach and often fails to find a point within the iteration limit. :slightly_frowning_face: Likely due to the overhead of a search based sampling and high rejection rate.

Option (2) is about **10x slower** than the original approach and does not appear to fail, offering better stability. There is overhead in the initial filtering phase, but then sampling is very fast. **Future advantage:** multi-point sampling could be easily batched for additional efficiency as the filter would only need to be set once for the batch.

Profile for the pictured scenario on my machine:
- current: 5.588531494140625e-06
- option 1: 2.3991942405700683e-04
- option 2: 7.155656814575195e-05

### Other Notes:

This is breaking change because the behavior of this function will change for downstream use. It will now more accurately sample points in a circle which could be considered a bug fix.

## How Has This Been Tested

Results with new implementation:
image

Navmesh with path between rooms visualized for reference:

nav_path

No CI tests presently. Could invent a contrived example to ensure this works.

## Types of changes

- [x Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -365,7 +365,7 @@ class IslandSystem {
         bool inRange = false;
         for (int iVert = 0; iVert < poly->vertCount; ++iVert) {
           int nVert = (iVert + 1) % 3;
-          float tseg;
+          float tseg = 0;
           float distSqr = dtDistancePtSegSqr2D(
               circleCenter.data(),
               &tile->verts[static_cast<size_t>(poly->verts[iVert]) * 3],
--- a/src/esp/nav/PathFinder.cpp
+++ b/src/esp/nav/PathFinder.cpp
@@ -479,8 +479,8 @@ struct PathFinder::Impl {
                                             int maxTries,
                                             int islandIndex /*= ID_UNDEFINED*/);
   vec3f getRandomNavigablePointInCircle(const vec3f& circleCenter,
-                                        float radius,
-                                        int maxTries,
+                                        const float radius,
+                                        const int maxTries,
                                         int islandIndex /*= ID_UNDEFINED*/);
 
   bool findPath(ShortestPath& path);
@@ -1313,7 +1313,7 @@ vec3f PathFinder::Impl::getRandomNavigablePointInCircle(
     }
   }
 
-  // reset the poly flag identifing polys off the target island
+  // reset the poly flag identifying polys off the target island
   islandSystem_->setPolyFlagForIsland(
       navMesh_.get(), PolyFlags::POLYFLAGS_OFF_ISLAND, ID_UNDEFINED,
       /*setFlag=*/false, /*invert=*/true);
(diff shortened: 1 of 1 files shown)


jimmytyyang (commented):
Hi Alex, thank you so much for this!

This new function allows us to sample a wide range of points homogeneously, but we will still sample a point that is on the other room.  So that robot still has a problem picking up the object. Do we consider that we fix that in the habitat-lab side or habitat-sim side? If it is the former, then we need to change the corresponding function there.

jimmytyyang (approved):
LGTM! Thanks for working on this! this is great!

src/esp/nav/PathFinder.cpp
```diff
@@ -1199,6 +1277,57 @@ vec3f PathFinder::Impl::getRandomNavigablePoint(
...
+      float xd = circleCenter[0] - pt[0];
+      float yd = circleCenter[2] - pt[2];
+      float d2 = xd * xd + yd * yd;
+      if (d2 < radSqr) {
+        break;
+      }
+    }
+  }
+
+  // reset the poly flag identifing polys off the target island
```
0mdc, line 1316: Nit: `identifing`

src/esp/nav/PathFinder.cpp
```diff
@@ -404,6 +478,10 @@ struct PathFinder::Impl {
                                             float radius,
                                             int maxTries,
                                             int islandIndex /*= ID_UNDEFINED*/);
+  vec3f getRandomNavigablePointInCircle(const vec3f& circleCenter,
+                                        float radius,
+                                        int maxTries,
+                                        int islandIndex /*= ID_UNDEFINED*/);
```
0mdc, line 484: Nit: Implementation has a signature with `const float radius` and `const int maxTries`.

0mdc (approved):
LGTM! Thanks for the detailed description.

I left some very minor comments.

Skylion007 (approved):

PR #2242, opened 2023-10-17
[P0 - Do not merge] Lighting tweaks for P0.

0mdc:
* Increase ambient intensity to brighten spot and humanoids, which are phong-shaded instead of PBR-shaded.
* Increase PBR light intensity for replay rendering.

### DIFF
--- a/src/esp/gfx/LightSetup.cpp
+++ b/src/esp/gfx/LightSetup.cpp
@@ -104,7 +104,7 @@ Magnum::Color3 getAmbientLightColor(const LightSetup& lightSetup) {
   } else {
     // todo: add up ambient terms from all lights in lightSetup
     // temp: hard-coded ambient light tuned for ReplicaCAD
-    float ambientIntensity = 0.60;
+    float ambientIntensity = 0.9;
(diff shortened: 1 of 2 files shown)


src/esp/gfx/LightSetup.cpp
```diff
@@ -104,7 +104,7 @@ Magnum::Color3 getAmbientLightColor(const LightSetup& lightSetup) {
   } else {
     // todo: add up ambient terms from all lights in lightSetup
     // temp: hard-coded ambient light tuned for ReplicaCAD
-    float ambientIntensity = 0.4;
+    float ambientIntensity = 0.9;
```
jturner65, line 107: Another option instead of increasing the ambient intensity here would be just to remove the ambientIntensity scaling completely [here](https://github.com/facebookresearch/habitat-sim/blob/97489d1ba71a65e3b3c0101fc9d8560ad20baa7c/src/esp/gfx/GenericDrawable.cpp#L139C20-L139C38). This would be equivalent to having this intensity value set to 1.

PR #2243, opened 2023-10-18
Update Magnum to latest

mosra:
## Motivation and Context

Update of the Magnum submodules containing changes from the last 4+ months.

- The Basis plugins now use a Zstd submodule instead of a heavily outdated build bundled with Basis itself. It should be significantly faster and more secure also. (Actually, it seems that Zstd wasn't actually present in the repo before, which means UASTC files probably never worked? Now they will.)
- Debug performance improvements for [EnumSet](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1EnumSet.html), [StridedArrayView](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1StridedArrayView.html), [BitArray](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BitArray.html) / [BitArrayView](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BasicBitArrayView.html) and [String](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1String.html) containers.
- It's now possible to call [StringView::join()](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1BasicStringView.html#a852ca05678c06ef86d7d5e5d79a7099f) with a `std::vector<std::string>` argument :scream: 
- Reduced dependency on STL headers, in particular `<utility>` and `<tuple>` in more places
- Various fixes for compatibility with newer Emscripten versions
- The [Pointer](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1Pointer.html) class now checks at compile times that a type is defined on deletion, preventing resource leaks due to the destructor not being called. It also disallows conversion to non-trivial/non-virtual bases, where a resource leak would happen due to the derived class destructor not being called.
- `Math::angle()` for quaternions was deprecated as it's actually returning a half-angle. [Math::halfAngle()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Math.html#aedc37065c5dad676c5dd5f10cc54f82c) should be used now instead as it's named appropriately (and `math.half_angle()` in case of Python), an `angle()` function returning a full angle will be reintroduced after enough time passes to make sure it doesn't change behavior of existing code.
- [Trade::MeshData](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1MeshData.html) and the [GltfImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1GltfImporter.html) now support meshes with morph targets.
- Working around a bug in Qualcomm Adreno drivers (Oculus Quest, e.g.) causing shaders to be silently miscompiled, not displaying anything
- Adapted detection of certain WebGL extensions to changes in their names
- Worked around some more cases where containers using `T{}` didn't work with classes that had a base with an explicit constructor.
- Recognizing the `NV_geometry_shader_passthrough` extension
- [Near/far plane distance queries in Matrix4](https://doc.magnum.graphics/magnum/classMagnum_1_1Math_1_1Matrix4.html#aaa4ce31fd0d85bc2c40fd877b00b0910)
- [Ability to draw a sub-range of an indexed GL::Mesh](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Mesh.html#a1691e32b1b771faf96de56d6cbcdd475) without having to create a temporary `GL::MeshView`
- [Platform::GlfwApplication](https://doc.magnum.graphics/magnum/classMagnum_1_1Platform_1_1GlfwApplication.html) now emits a viewport event when moving the windows across macOS displays with different DPI. Note that this *isn't* a complete support for DPI change events across all platforms, that's still quite far ahead.
- [GltfSceneConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1GltfSceneConverter.html) now preserves PBR materials even if they're empty, instead of removing them
- [BasisImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1BasisImporter.html) and [KtxImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1KtxImporter.html) now warn if the file is missing orientation metadata, to aid with debugging asset errors
- Improved test coverage for [JpegImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1JpegImageConverter.html) and [PngImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PngImageConverter.html)
- Support for more legacy formats in [DdsImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1DdsImporter.html), especially to be found in various FBX files
- Emscripten support in [OpenExrImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1OpenExrImporter.html) and [OpenExrImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1OpenExrImageConverter.html), in particular correct exception handling when a file fails to be parsed
- Other changes, unrelated to Habitat.

## How Has This Been Tested

:green_book:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -46,3 +46,6 @@
 [submodule "src/deps/rlr-audio-propagation"]
 	path = src/deps/rlr-audio-propagation
 	url = https://github.com/facebookresearch/rlr-audio-propagation
+[submodule "src/deps/zstd"]
+	path = src/deps/zstd
+	url = https://github.com/facebook/zstd
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -59,7 +59,10 @@ option(USE_SYSTEM_BULLET "Use system Bullet instead of a bundled submodule" OFF)
 # Zstd is used only by the BasisImporter plugin, if external Magnum is used it
 # doesn't need to be built at all.
 include(CMakeDependentOption)
-cmake_dependent_option(USE_SYSTEM_ZSTD "Use system zstd instead of a bundled submodule" OFF "NOT USE_SYSTEM_MAGNUM" ON)
+cmake_dependent_option(
+  USE_SYSTEM_ZSTD "Use system zstd instead of a bundled submodule" OFF
+  "NOT USE_SYSTEM_MAGNUM" ON
+)
 option(
   BUILD_BASIS_COMPRESSOR
   "Wether or not to build the basis compressor.  Loading basis compressed meshes does NOT require this."
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -56,6 +56,10 @@ option(USE_SYSTEM_MAGNUM "Use system Magnum instead of a bundled submodule" OFF)
 option(USE_SYSTEM_PYBIND11 "Use system Pybind11 instead of a bundled submodule" OFF)
 option(USE_SYSTEM_RAPIDJSON "Use system RapidJSON instead of a bundled submodule" OFF)
 option(USE_SYSTEM_BULLET "Use system Bullet instead of a bundled submodule" OFF)
+# Zstd is used only by the BasisImporter plugin, if external Magnum is used it
+# doesn't need to be built at all.
+include(CMakeDependentOption)
+cmake_dependent_option(USE_SYSTEM_ZSTD "Use system zstd instead of a bundled submodule" OFF "NOT USE_SYSTEM_MAGNUM" ON)
 option(
   BUILD_BASIS_COMPRESSOR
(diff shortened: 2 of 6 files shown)


jturner65 (approved):
Glad to see the Near/Far plane calcs make it in to Habitat finally.

0mdc (approved):

Skylion007 (approved):

aclegg3 (approved):

PR #2235, opened 2023-10-10
--PBR Primitive support and gl context bugfix

jturner65:
## Motivation and Context
This PR introduces PBR material support and generation for primitives and generated meshes such as trajectories, which were formerly restricted to just phong materials, which looked out of place in scenes that were predominantly rendered using PBR with image-based lighting.  This also adds vertex color support for PBR materials.  Lastly, this also fixes a bug that was present if trying to build a PBR containing scene without a context (which was being done in a few of the tests).

This functionality was originally going to be part of the PBRTesting PR but I chose to go a different direction with that test due to the difficulty in assigning custom materials to primitive assets on creation.

## How Has This Been Tested
C++ and python tests pass locally.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2262, opened 2023-11-10
Set the default fallback material as Phong.

0mdc:
## Motivation and Context

The default materials were changed to PBR [here](https://github.com/facebookresearch/habitat-sim/pull/2235), including the default fallback material. When rendering with depth cameras, no texture is loaded from materials. The fallback material is instead applied to objects.

Because skinning only works with Phong, it breaks when the only sensors present in the simulation are depth.

This changes the default fallback material to Phong. We can revert this change when skinning support is added to PBR.

## How Has This Been Tested

Tested locally on the SIRo HITL tool.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -2181,6 +2181,17 @@ Mn::Trade::MaterialData ResourceManager::buildDefaultMaterial() {
   return materialData;
 }  // ResourceManager::buildDefaultMaterial
 
+// Specifically for building materials that relied on old defaults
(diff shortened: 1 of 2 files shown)


jturner65 (approved):
LGTM. I can prune this if necessary once I get the PBR skinning to work.

eundersander (approved):
As I understand, this is reverting a portion of the recent https://github.com/facebookresearch/habitat-sim/pull/2235 , so I imagine it's fairly safe. I'm not too familiar with the current state of materials or what the fallbackMaterial is used for. I'll defer to John and Mikael.

Issue #2213, opened 2023-09-26
"TypeError: cannot pickle '_magnum.Color4' object" when running benchmark.py in v0.2.5


sanjayharesh:
Hi there,

I'm trying to run the benchmarking script [here](https://github.com/facebookresearch/habitat-sim/blob/v0.2.5/examples/benchmark.py) in v0.2.5 and I'm getting the following error. 

`TypeError: cannot pickle '_magnum.Color4' object`

Unfortunately, don't get more detail in the error message thought it seems to be because of multiprocessing and removing the multiprocessing here: https://github.com/facebookresearch/habitat-sim/blob/c8887c8cf421b6df2d77489d42f4e20488e184eb/examples/demo_runner.py#L391

seems to make it work. 

For completeness, here's my setup
habitat-sim v0.2.5 installed from source. 
Ubuntu 18.04
Python 3.9 and Cuda version 11.7.

aclegg3:
Thanks for raising this, @sanjayharesh,

I can reproduce this issue locally.

Looks like this particular code isn't run our CI, so possible it has become stale.

mosra:
FYI, this should be implemented with https://github.com/mosra/magnum-bindings/commit/f561337caa9094c4a99411fac402342343f4ecb4, however I don't have an easy way to verify that it indeed works in the full context of Habitat. 

Assuming #2243 is included in your working copy, it should be enough to just pull the submodule in `src/deps/magnum-bindings` to this commit to get the change and check.

ORionTMT:
I also ran into the same problem with v0.3.0, have you managed to solve it yet?

PR #2278, opened 2023-12-07
Update Magnum to latest

mosra:
## Motivation and Context

Assorted ~10kLOC update containing mainly various prerequisites for the batch renderer, and additions to Python bindings.

- More efficient packing of `GL::Buffer` and `GL::Mesh` internals -- it used to take 48 bytes per mesh attribute, now it's just 24. It also no longer does nasty voodoo with in-place construction of a `std::vector`.
- New [MaterialTools::removeDuplicates()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MaterialTools.html#a4bf4918fab8d9162d96298d8dcfe423c) utility for reducing the amount of redundant materials in a (batch composite) file. There's a corresponding `magnum-sceneconverter --remove-duplicate-materials` option as well, and it's in the [materialtools](https://doc.magnum.graphics/python/magnum/materialtools/) Python bindings too. In practice it will need a material canonicalization utility (such as removing empty layers, redundant default attributes and such) to truly remove all duplicates, that's coming next.
- New [SceneTools::mapIndexField()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1SceneTools.html#ae2cf6f6d3737ea1b5855c7e693f3ddc0) family of tools for remapping mesh, material, camera, ... references in a scene, such as for example after duplicate material removal.
- Python bindings:
    - Implemented pickling support for all Math types. Could help with https://github.com/facebookresearch/habitat-sim/issues/2213, unfortunately I didn't get a confirmation from anybody that the proposed changes actually work, so it's just a "best intentions" fix, I can't claim it actually resolves the problem.
    - Exposed the [trade.MaterialData](https://doc.magnum.graphics/python/magnum/trade/MaterialData/) class, the [materialtools](https://doc.magnum.graphics/python/magnum/materialtools/) library and corresponding APIs in importers and scene converters, which was one of the last missing pieces for full Python-side support for arbitrary asset processing and conversion.
    - The [math](https://doc.magnum.graphics/python/magnum/math/) module now raises exceptions in all cases where the C++ code fires an assertion. Those used to be assertions that were left even in release builds (and thus causing the Python interpreter to abort, which is bad), but got changed to debug-only asserts in https://github.com/facebookresearch/habitat-sim/pull/2100 (thus causing the Python bindings to silently give wrong results, which is perhaps even worse). Invalid usage now raises Python exceptions as it should. Reported by @xavierpuigf.
- Other changes.™

Oh, and I made [Utility::String::lowercase()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1String.html#a8f89429f64be0b47f198ef3332f29be1) >10x faster while being idle on the weekly call, and then 2x more using AVX2, so **~25x faster**. See mosra/corrade@808520106e93fcfe158375b7658e0121c439ea27 and https://github.com/mosra/corrade/compare/8010c5cc39d5b17c868c180e34c146c38eac6105...d40ff9dde28bc72a398b1b402bb5ca6844549c2d for details, the "naive" case below is the previous implementation.

image

## How Has This Been Tested

:green_heart:


0mdc (approved):
Thank you!

PR #2299, opened 2024-01-10
-- [Part 1 of 2] Semantic Attributes Metadata : Support for Semantic Regions - Add attributes/config support

jturner65:
## Motivation and Context
This PR is part 1 of a 2 part changeset that adds support for semantic regions in a dataset-agnostic way through the use of the existing metadata subsystem tooling. 

This component adds manager, attributes and JSON config support for semantic scene descriptions that include region annotations for poly-loop-based extrusion region definitions, integrated into the existing metadata subsystem, managed by the existing scene dataset handling and accessed during scene creation. This PR does not introduce any breaking changes to the existing system, but does expand the possible metadata loading/handling in the scene dataset configs for the "semantic_scene_descriptor_instances" tag. 

Part 2 of this PR will use these configs to instantiate Semantic Regions with the appropriate geometry and location that will support point containment checks and debug rendering.

## How Has This Been Tested
Existing c++ and python tests pass.

## Types of changes

## Checklist


src/esp/bindings/AttributesManagersBindings.cpp
```diff
@@ -237,6 +239,15 @@ void declareBaseAttributesManager(py::module& m,
             "if it does not.")
                .c_str(),
            "handle"_a)
+      .def("get_first_matching_template_by_handle",
```
aclegg3, line 242: Why introduce this instead of getting all matches and selecting the first like we do already? Doing that puts responsibility on user code to handle the case of multiple matches and offers the opportunity for an assertion.

aclegg3 (approved):
Overall looks good. Once the tests pass, let's merge it.

src/esp/bindings/AttributesManagersBindings.cpp
jturner65: I was doing this to provide a shortcut for folks who know they are going to use the first handle they get back, so they didn't have to deal with querying the handles and then querying for the actual attributes.  Seems like the first index tends to be the goto for this kind of process, and I thought I'd abstract it away. The other mechanism is still present.

PR #2314, opened 2024-02-01
Activate objects when moved

aclegg3:
## Motivation and Context

Context: I noticed that edge cases existed where discrete collision detection results were not accurate. Investigation revealed that when a sleeping object was moved kinematically into contact with another sleeping object, the contact would be ignored by discrete collision detection routine.

After investigation, I concluded that the most sensible solution would be to force wake any sleeping object which is moved kinematically. This handles the current bug in discrete collision checking as well as making more sense (teleporting and object near a sleeping island could cause reactions and so the island should be awoken).

BREAKING CHANGE: any downstream code taking advantage of this bug will no longer work as it did previously. This fix will create some minimal additional overhead as kinematic motion of a sleeping object may wake other nearby islands during simulation. 

Note that unmoved sleeping objects will not wake from discrete collision checking with kinematic shapes unless step_physics is called while the objects are in contact. This should continue to efficiently support use cases such as checking for valid contact-free states between simulation steps.

## How Has This Been Tested

Added new test scenarios to the python physics test.

## Types of changes

## Checklist


jturner65 (approved):
LGTM.

aclegg3:
Looks like joint constraint pytest is reporting different torques after this change. I didn't see any change in behavior from the videos other than new random numbers. Perhaps this test was faulty to begin with. Removed the assertion.

jturner65:
> Looks like joint constraint pytest is reporting different torques after this change. I didn't see any change in behavior from the videos other than new random numbers. Perhaps this test was faulty to begin with. Removed the assertion.

I'll take a look at this in the morning, just to be sure.

0mdc (approved):
LGTM!

PR #2321, opened 2024-02-12
Ability to render the viewer content at quarter resolution on very-high-DPI screens

mosra:
## Motivation and Context

Changes needed in the native viewer application and the RenderTarget to allow for rendering at a resolution that's smaller than the actual framebuffer. For simplicity halving the framebuffer width and height, thus quarter of the size, however the limit at which this switch is done is configurable, so it's also achievable on non-HiDPI displays if you pass `--quarter-resolution-scaling-limit 1` to the app.

It also:

- Removes use of 4x MSAA which achieved nothing as the content was rendered to a non-multisampled framebuffer in the first place.
- Cleans up various interesting ways of doing vector math.
- Removes an assertion in the RenderTarget that required the source and destination size to match, and switches the filter to linear to not look *that* blocky.
- Removes an errorneous multiplication by `dpiScaling()` when getting a framebuffer-relative mouse position. The DPI scaling value is already baked into the framebuffer size at that point, thus multiplying again gives an incorrect value.

See also https://cvmlp.slack.com/archives/CFN5TAUSD/p1697296903351319 for original context.

## How Has This Been Tested

On my Linux box with a 4K display, 1.6875x DPI scaling and `--quarter-resolution-scaling-limit 1.6`. The way the limit calculation is done it should work on a Mac as well.

## Types of changes

### DIFF
--- a/src/utils/viewer/viewer.cpp
+++ b/src/utils/viewer/viewer.cpp
@@ -667,9 +667,7 @@ Key Commands:
 #endif  // ESP_BUILD_WITH_AUDIO
 };      // class viewer declaration
 
-void addSensors(esp::agent::AgentConfiguration& agentConfig, bool isOrtho) {
-  const auto viewportSize = Mn::GL::defaultFramebuffer.viewport().size();
-
+void addSensors(esp::agent::AgentConfiguration& agentConfig, bool isOrtho, const Mn::Vector2i& framebufferSize) {
   auto addCameraSensor = [&](const std::string& uuid,
                              esp::sensor::SensorType sensorType) {
     agentConfig.sensorSpecifications.emplace_back(
@@ -718,8 +716,7 @@ void addSensors(esp::agent::AgentConfiguration& agentConfig, bool isOrtho) {
     spec->fisheyeModelType = modelType;
     spec->resolution = viewportSize.flipped();
     // default viewport size: 1600 x 1200
-    spec->principalPointOffset =
-        Mn::Vector2(viewportSize[0] / 2, viewportSize[1] / 2);
+    spec->principalPointOffset = Mn::Vector2{framebufferSize}/2.0f;
     if (modelType == esp::sensor::FisheyeSensorModelType::DoubleSphere) {
       // in this demo, we choose "GoPro":
       spec->focalLength = {364.84f, 364.86f};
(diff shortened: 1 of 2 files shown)


jturner65 (approved):
Thanks mosra!

src/utils/viewer/viewer.cpp
```diff
@@ -863,11 +864,25 @@ Viewer::Viewer(const Arguments& arguments)
       .addOption("agent-transform-filepath")
       .setHelp("agent-transform-filepath",
                "Specify path to load camera transform from.")
+      .addOption("quarter-resolution-scaling-limit", "2.0")
```
eundersander, line 867: I guess a more intuitive interface to expose to the command line would be a simple bool, something like "--reduce-resolution-for-high-dpi-displays". Which should probably set the scaling limit to something like 1.6 or 1.9, right?

eundersander (approved):
One minor request, otherwise LGTM. This is mainly reference code for our other GUI apps floating around (like the HITL apps) where we'll reimplement this logic. Thanks for doing this!

src/utils/viewer/viewer.cpp
eundersander: Alternately, at least a better description for this arg so people know how to use it.

mosra: That's the thing, I didn't expect people to use this option :) The default behavior makes sense to me. The option is there just as a fallback, to tune when the size reduction kicks in, in rare cases such as when the desktop scaling is configured weird, or when the GPU is not powerful enough to render at full resolution, or when making screenshots and so you _do_ want a full resolution even on a HiDPI screen.

As I understood the request, the main objective was to reduce the resolution _by default_ on very-high-DPI displays, which is for me with 200% display scale and higher. On lower density displays with 1.5x or 1.6x scaling it looks visibly uglier (as it did on my 4K 23" screen), so there it isn't enabled, while for >=2x scaled displays you'll see the difference only if you really focus on it (as it was the case on my 4K 16" display).

eundersander: Ok, I think I understand now. `quarter-resolution-scaling-limit==2.0` means "reduce resolution for very-high-DPI displays" (by which we mean Retina and other 200%+ display scales). `quarter-resolution-scaling-limit==9999` would mean "don't reduce resolution, regardless of DPI". `quarter-resolution-scaling-limit==1.0` would mean "always reduce resolution, regardless of DPI".

Anyway I don't think we particularly need this feature for the C++ viewer, which isn't used much any more. This is a helpful reference. At some point, @0mdc or I can port this to the HITL framework's replay-renderer, where Retina has been bothering us.

0mdc (approved):
LGTM!

PR #2322, opened 2024-02-14
Update Magnum to latest

mosra:
## Motivation and Context

Quick update containing mainly these two fixes:

- A refactor of `GL::Mesh` internals in November resulted in a regression where the enclosing VAO wasn't properly deleted, causing GPU memory leaks on certain drivers when meshes were destructed during application lifetime. Sorry about that, the leak is plugged now.
- Added a workaround for DPI scaling detection where on KDE the `Xft.dpi` property is set only if desktop scaling is set to something else than 100%. Discovered by @0mdc.

There's Other Things™ as well, but none of these seem too relevant to Habitat. Feel free to look at the commit log for these.

## How Has This Been Tested

:green_circle:


0mdc (approved):
Thank you!

jturner65 (approved):

PR #2316, opened 2024-02-05
--[WIP] Add Panoptic/Instance ID support to Semantic sensor

jturner65:
## Motivation and Context
This PR will add the ability to have the semantic sensor produce images containing object ID of the objects rendered per-pixel in the same way that it currently produces semantic ID observations.

Still a WIP.

## Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2325, opened 2024-02-22
--Define constant and bindings for stage ID

jturner65:
## Motivation and Context
The stage ID is referenced in python and from lab in numerous places, usually as a mark of where added objects' ids begin, or else as a final target for collision detection. Unfortunately, this is often referenced or checked by using magic number references to the stage (i.e. -1).

This PR introduces a global constant to denote the stage ID (`esp::RIGID_STAGE_ID`), and bindings to that constant to expose it to python (`habitat_sim.stage_id`), which should be used to refer to the stage ID.  A matching Habitat-Lab PR is pending for this support as well, once this PR is merged.  Ultimately this is to facilitate t[he instanced -based semantic sensor support PR here 
](https://github.com/facebookresearch/habitat-sim/pull/2316)

## How Has This Been Tested
All c++ and python tests pass locally

## Types of changes

## Checklist

### DIFF
--- a/examples/viewer.py
+++ b/examples/viewer.py
@@ -723,7 +723,7 @@ class HabitatSimInteractiveViewer(Application):
                 hit_object, ao_link = -1, -1
                 hit_info = raycast_results.hits[0]
 
-                if hit_info.object_id >= 0:
+                if hit_info.object_id > habitat_sim.stage_id:
                     # we hit an non-staged collision object
                     ro_mngr = self.sim.get_rigid_object_manager()
                     ao_mngr = self.sim.get_articulated_object_manager()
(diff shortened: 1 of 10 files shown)


examples/viewer.py
```diff
@@ -723,7 +723,7 @@ def mouse_press_event(self, event: Application.MouseEvent) -> None:
                 hit_object, ao_link = -1, -1
                 hit_info = raycast_results.hits[0]
 
-                if hit_info.object_id >= 0:
+                if hit_info.object_id > habitat_sim.stage_id:
```
aclegg3, line 726: Using "greater than" here still assumes that stage_id will always be less than all object ids. While this is the case currently does it defeat the purpose of having a variable?

aclegg3 (approved):
This step seems fine. My comments are mostly forward thinking about when this id will change in the future. 
For example, replacing "==0" with "==stage_id+1" seems like a strong assumption about the value.
Next step may be to update the "allocateObjectID" system to handle this value. :thinking:

PR #2334, opened 2024-03-05
--[CI Fix] Fix miniconda versions

jturner65:
## Motivation and Context
Using the latest miniconda versions is referencing python 3.12 which is undesirable, and causes osx builds to break (pkgutil.ImpImporter class was deprecated in python 3.3 and removed in 3.12.)

This PR will freeze our miniconda version to most recent python 3.10 version for both linux and mac.

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,7 +115,7 @@ commands:
           command: |
               if [ ! -d ~/miniconda ]
               then
-                curl -o ~/miniconda.sh -O  https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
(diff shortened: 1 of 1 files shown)


0mdc (approved):
LGTM!

We'll probably want to extend support to Python 3.12 in the foreseeable future.

jturner65:
> LGTM!
> 
> We'll probably want to extend support to Python 3.12 in the foreseeable future.

Yeah, definitely (supposed to be markedly faster with strings, for instance), but it will certainly require a some investigation and reworking since there are many modules that have been deprecated for a while that were removed with 3.11 and 3.12.

aclegg3 (approved):

aclegg3:
Thanks for investigating and fixing this. :+1:

PR #2336, opened 2024-03-12
--Region query tools

jturner65:
## Motivation and Context
This PR adds functions to test for point-region membership and point collection-region membership with a weighting factor, to account for possibly nested regions. If a point is only found in a single region, its weighting in that region is 1. 

Otherwise, each region is assigned a weighting that is inversely proportional to the area of a particular region containing the point.  The weighting is calculated like so : 
 ` 

         1 - (region_area/total_region_area)

`
where `region_area` is the area of a specific region the point is found in and total_region_area is the total area of all regions that point is found in. 

For collections of points that are assumed to belong to the same construct, the region membership check has been modified to give ~~a weighting value only to the innermost (i.e. smallest area) region in nested regions. In other words, in the case of nested regions, a point is only considered found in the innermost region.~~ the same weighting value to all regions containing a particular point. In the case of nested regions, this may result in an object being fully contained within the containing region, while only partially contained within the nested internal region.

## How Has This Been Tested
Locally c++ and python tests. Python test modified to verify that with non-nested regions it returns the same result as the non-weighted versions

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/SceneBindings.cpp
+++ b/src/esp/bindings/SceneBindings.cpp
@@ -256,19 +256,10 @@ void initSceneBindings(py::module& m) {
       .def("get_regions_for_points", &SemanticScene::getRegionsForPoints,
            R"("Compute SemanticRegion containment for a set of points. Return a
               sorted list of tuple pairs with each containing region index and
-              the percentage of points contained by that region.)",
-           "points"_a)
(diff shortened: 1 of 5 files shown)


aclegg3 (approved):
LGTM, thanks

PR #2355, opened 2024-04-02
Update Magnum for ASCII PLY import and artifact-free text rendering

mosra:
Another month+ worth of Magnum updates, this time containing two important usability features:

- The [StanfordImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StanfordImporter.html) and [StlImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StlImporter.html) plugins now delegate to Assimp for ASCII files instead of failing the import. Which brings the best of the two worlds -- I don't need to suffer through implementing an ASCII PLY or STL parser myself, and you don't need to suffer with slow import of binary PLY and STL files by defaulting to Assimp instead. In particular, https://github.com/facebookresearch/habitat-lab/pull/1869 can be reverted after this PR is merged, and everything should continue to work.
- Text rendering artifacts that started happening after an update to the new and more efficient [TextureTools::AtlasLandfill](https://doc.magnum.graphics/magnum/classMagnum_1_1TextureTools_1_1AtlasLandfill.html) atlas packer are now gone again. In particular, the one-pixel glyph padding that was enabled for example in https://github.com/facebookresearch/habitat-sim/pull/2279 is done by default now. [Explanation of the behavior and rationale for the new default is here.](https://doc.magnum.graphics/magnum/classMagnum_1_1Text_1_1AbstractGlyphCache.html#Text-AbstractGlyphCache-padding)

Besides that, the following changes are worth mentioning:

- A new [Debug::hex](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html#a7f6dc96772c55a35fb5d72e1ef27cfa5) modifier to conveniently print integer values as hexadecimal. As a drive-by optimization, instances of the `Debug` class now occupy 16.67% less memory.
- The [SceneGraph](https://doc.magnum.graphics/magnum/scenegraph.html) library is now capable of rendering more than 65k objects at once. The original limitation was due to a 16-bit type used internally, however there's enough free padding bytes to make it 32-bit without enlarging size of scene graph objects in memory.
- [Containers::StaticArray](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1StaticArray.html) is now trivially copyable and also usable in `constexpr` contexts. In practice that means better optimization opportunities for the compiler, IIRC Habitat isn't making any extensive use of this type anywhere.
- The [GL::Mesh](https://doc.magnum.graphics/magnum/classMagnum_1_1GL_1_1Mesh.html) could draw from vertex buffers larger than 4 GB before, now it can draw from index buffers larger than 4 GB as well. I've already hit the 4 GB vertex buffer limit with batch renderer before, it's just a matter of time when it happens with indices too, so better be prepared.
- The Magnum CMake `Find` modules were updated to properly support the case where each project is installed in a different prefix, unblocking one Habitat use case: https://github.com/mosra/magnum-integration/issues/105
- The [StbImageImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1StbImageImporter.html) now uses NEON on ARM machines, such as the new Macs, which may cause image import to be slightly faster.
- In addition to a [WebPImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1WebPImporter.html) there's now a [WebPImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1WebPImageConverter.html) as well, together with support for the WebP format in both the glTF importer and exporter. Mentioning just in case you'd have use for a format that's both lossless and lossy and can achieve a better compression ratio than PNG or JPEG. The prebuilt tool binaries have this plugin enabled now as well.

(Sorry, I deleted the PR template by accident. Tests are :green_circle: on my end, that `mypy` failure wasn't me.)

Also mentioning @aclegg3 -- can't add you as a reviewer for some reason.

### DIFF
--- a/src/cmake/FindAssimp.cmake
+++ b/src/cmake/FindAssimp.cmake
@@ -17,7 +17,7 @@
 #   This file is part of Magnum.
 #
 #   Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
-#               2020, 2021, 2022 Vladimír Vondruš <mosra@centrum.cz>
+#               2020, 2021, 2022, 2023 Vladimír Vondruš <mosra@centrum.cz>
 #   Copyright © 2017 Jonathan Hale <squareys@googlemail.com>
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a
@@ -137,6 +137,11 @@ if((TARGET assimp OR TARGET assimp::assimp) AND NOT _ASSIMP_HAS_USELESS_CONFIG)
     # doesn't define any usable INTERFACE_INCLUDE_DIRECTORIES for some reason
     # (the $<BUILD_INTERFACE:> in there doesn't get expanded), so let's extract
     # that from the SOURCE_DIR property instead.
+    #
+    # TODO this could be probably fixable by using target_link_libraries()
+    # instead of set_target_properties() because it evaluates generator
+    # expressions, but that needs CMake 3.10+, before that
+    # target_link_libraries() can't be called on INTERFACE targets.
     if(_ASSIMP_INTERFACE_INCLUDE_DIRECTORIES MATCHES "<BUILD_INTERFACE:")
         get_target_property(_ASSIMP_INTERFACE_INCLUDE_DIRECTORIES assimp SOURCE_DIR)
         get_filename_component(_ASSIMP_INTERFACE_INCLUDE_DIRECTORIES ${_ASSIMP_INTERFACE_INCLUDE_DIRECTORIES} DIRECTORY)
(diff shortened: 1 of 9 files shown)


aclegg3:
Thanks @mosra, I'm working on the CI issue here: #2356

aclegg3 (approved):
Tests pass, so assuming all the build changes are good. :)
Thanks, mosra!
Also, CI lint issue should be resolved on `main`.

jturner65 (approved):

0mdc (approved):
Thank you!

PR #2363, opened 2024-04-17
--ConfigValue Updates - Matrix4 support, pointer-backed data store.

jturner65:
## Motivation and Context
This PR changes how esp::core::config::Configurations store data by having the underlying ConfigValue class store a pointer to the data in question instead of directly storing the data (i.e. the class member _data is now a pointer to a pointer to the data instead of a pointer to the data) if the data type is too large to be stored within the _data buffer.  This PR also introduces, as a way to demonstrate this new feature, support for a new type in the configs, **Magnum::Matrix4.**

Currently on main branch, every ConfigValue includes a buffer that holds the actual data being consumed;  due to the size of this buffer, every ConfigValue takes up 72 bytes of memory. This PR decreases this to 16 bytes + whatever heap allocation is necessary to hold the larger-than-8 byte types (the config value consists of a data array, now 8 bytes, and an enum denoting the type, 4 bytes, along with 4 bytes padding.)

Here's the memory footprint for each ConfigValue, along with any heap allocation if necessary (for larger objects), in this PR.
```
**Type        New**
ConfigValue buffer stores value : 
bool          16
int           16
double        16
Vector2       16
Mn::Rad       16
ConfigValue buffer stores ptr to value : 
Vector3      16 + 12
Vector4      16 + 16   (same for quaternion)
Matrix 3     16 + 36
String       16 + 32
Matrix4 (incoming in this PR) 16 + 64
```

## How Has This Been Tested
Local c++ and python tests pass

## Types of changes

## Checklist


aclegg3 (approved):
Cool, LGTM

PR #2396, opened 2024-05-16
--Add Configuration State Flags - Part 1

jturner65:
## Motivation and Context
This is the first of 2 PRs to handle this functionality. The 2nd PR will handle making sure configuration filepaths do not get saved to JSON with inappropriate path information.  

This PR is intended to expand the capabilities of the Configurations without incurring much cost by providing status flags for each ConfigValue that reside within the 4 byte gap the current _type field causes due to 8 byte alignment.

- Modifies the ConfigValType (int) _type field to be u_int64_t _typeAndFlags, where the lower 32 bits correspond to the previous _type value and the higher 32 bits are used to record state of the ConfigValue, along with necessary plumbing to consume the type value in the same way as before.

- Adds ConfigValStatus enum (u_int64_t) to hold bit flags representing state of the ConfigValue. Currently 3 flags are supported, ConfigValStatus::isDefault , ConfigValStatus::isHidden and ConfigValStatus::isTranslated.
        -- **isDefault** : represents programmatic initialization of a value, as opposed to it being set intentionally from a file or user input.
        -- **isHidden** : represents a value being used internally and not part of the actual Configuration data hierarchy. These values we never want to write to file or expose to a user unless for debugging purposes.
         -- **isTranslated** : represents a value that is stored as a string but is translated, currently to enum values, upon consumption.

 - Adds "init" setters that set the ConfigValStatus::isDefault bit of the ConfigValue's _typeAndFlags variable to 1, while the legacy "set" setters set it to 0.
    
  - Adds "hidden" setters that will set the ConfigValStatus::isHidden bit. The Attributes that consume this bit have as their keys double-underscore + Hungarian notation names, to make them obvious compared to the standard camel-case format of non-hidden values.

  - All existing Attributes constructors have their default value setting changed from using variants of "set" to using the new "init"  and "setHidden" setters where appropriate.

 - A check on the "isDefault" and "isHidden" flags is added when before writing to JSON.

Not only will these features prevent the unnecessary writing to disk of default fields, but they can also be used to prevent configurations being written in a potentially damaging or indeterminant state due to modifications done internally as part of the program's flow.

Also adds a repathing tool in the io namespace : 
```
io::getPathRelativeToAbsPath(const std::string& toRelPath,
                                     const std::string& absPath)
```

which converts the absolute path `toRelPath` to be relative to the given target `absPath`

## How Has This Been Tested
Locally c++ and python tests pass

## Types of changes

## Checklist


src/esp/core/Configuration.h
```diff
@@ -1206,14 +1598,29 @@ class Configuration {
...
+    // Configuration
     if (result.second) {
       result.first->second = std::make_shared<T>();
     }
     return result;
   }
 
  private:
+  /**
+   * @brief Write the passed @p configValue to a json objbect tagged with
```
aclegg3, line 1610:
```suggestion
   * @brief Write the passed @p configValue to a json object tagged with
```

src/esp/metadata/attributes/ArticulatedObjectAttributes.cpp
```diff
@@ -11,25 +11,32 @@ namespace attributes {
...
-  setBaseType(getAOBaseTypeName(ArticulatedObjectBaseType::Free));
-  // Set the default source for the interia calculation to be computed
-  setInertiaSource(
+
+  init("render_asset", "");
+  init("semantic_id", 0);
+  // Initialize the default base type to be free joint
+  initTranslated("base_type",
+                 getAOBaseTypeName(ArticulatedObjectBaseType::Free));
+  // Initialize the default source for the interia calculation to be computed
```
aclegg3, line 22:
```suggestion
  // Initialize the default source for the inertia calculation to be computed
```

src/esp/metadata/attributes/ObjectAttributes.cpp
```diff
@@ -10,44 +10,41 @@ namespace attributes {
...
-  // or instance config
-  setShaderType(getShaderTypeName(ObjectInstanceShaderType::Material));
-  // TODO remove this once ShaderType support is complete
-  setForceFlatShading(false);
-  setIsVisible(true);
-  setSemanticId(0);
+  init("use_bounding_box_for_collision", false);
+  init("join_collision_meshes", false);
+
+  init("semantic_id", 0);
```
aclegg3, line 24: Is this "unknown" or could it be a random class?
Should probably reserve the default for unknown.
Doesn't have to be this PR, just thought I'd mention it.

aclegg3 (approved):
Looks good, let's see downstream tests pass before merging to prevent issues with nightly.

I like that the per-property setters were replaced with a generic API.
Should also make searching for these easier.

src/esp/bindings/AttributesBindings.cpp
```diff
@@ -234,6 +234,18 @@ void initAttributesBindings(py::module& m) {
       .value("TREE_TRAVERSAL",
              metadata::attributes::ArticulatedObjectLinkOrder::TreeTraversal);
 
+  //
+  // ==== AssetType ====
+  // Describes the type of asset used for rendering, collsions, or semantics
+  py::enum_<metadata::attributes::AssetType>(m, "AssetType")
```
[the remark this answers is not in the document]
jturner65: The getters provide this value as an enum. This does remind me, though , that this needs to be read only now, since we don't want manual manipulation of this as it could break things in arcane ways. The whole asset-type component system needs to be removed, since it has been obviated by various config settings.

src/esp/metadata/attributes/ObjectAttributes.cpp
jturner65: This semantic ID being set to 0 corresponds to the stage.

PR #2412, opened 2024-06-21
Update README.md in docs

danieltmeta:
This PR is to update the installation instructions for the doc folder for html documentation reference.

## Types of changes

## Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,20 +1,27 @@
 ## Building Docs Locally
 ### First Time Setup:
-Install required packages:
-```
-conda install -c conda-forge doxygen==1.8.16
-conda install jinja2 pygments docutils
+Create a new conda environment:
+
+```bash
+# We require python>=3.9 and cmake>=3.10
+conda create -n habitat python=3.9 cmake=3.14.0
+conda activate habitat
 ```
 
-To build the docs, you will also need to install the following
-packages on Linux :
+Install required packages:
 ```
-sudo apt install --allow-change-held-packages \
-                  texlive-base \
-                  texlive-latex-extra \
-                  texlive-fonts-extra \
-                  texlive-fonts-recommended
+cd habitat-sim
+pip install -r requirements.txt --progress-bar off
+pip install imageio imageio-ffmpeg
+conda install -y -c conda-forge doxygen=1.9.5
+conda install -y  jinja2 pygments docutils
 ```
+
+To build the docs, you will also need to install textlive on Fedora Linux :
+```
+sudo dnf install texlive-scheme-full
+```
+
 or on OSX :
 ```
 brew install --cask mactex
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,20 +1,27 @@
 ## Building Docs Locally
 ### First Time Setup:
+Create a new conda environment:
```
aclegg3, line 3:
```suggestion
If a conda environment for habitat does not already exist, create a new conda environment:
```

docs/README.md
```diff
@@ -1,20 +1,27 @@
...
+To build the docs, you will also need to install textlive on Fedora Linux :
 ```
-sudo apt install --allow-change-held-packages \
-                  texlive-base \
-                  texlive-latex-extra \
-                  texlive-fonts-extra \
-                  texlive-fonts-recommended
+sudo dnf install texlive-scheme-full
 ```
+
```
aclegg3, line 24: Let's leave both options. The `apt install` option is for Debian systems like Ubuntu.

docs/README.md
```diff
@@ -27,10 +34,14 @@ export PATH=$PATH:/Library/TeX/Distributions/.DefaultTeX/Contents/Programs/texbi
 
 ### Building Docs:
 ```bash
-python setup.py build_ext --inplace
```
aclegg3, line 30: did this not work for you? I think this worked for me last time.

docs/README.md
```diff
@@ -27,10 +34,14 @@ export PATH=$PATH:/Library/TeX/Distributions/.DefaultTeX/Contents/Programs/texbi
...
 cd docs
 git submodule update --init
-./build.sh # or ./build-public.sh when deploying to aihabitat.org
+
+cd habitat-sim
+./build.sh
```
aclegg3, line 41: You may need `--with-bullet` to compile the physics libraries in order for their docstrings to be available after building the docs

aclegg3 (changes requested):
Thanks for updating with your experience. Some thoughts/comments before we change the docs.

docs/README.md
danieltmeta: I agree, let's keep both

docs/README.md
danieltmeta: It didn't work for me, for some reason I had to go to the root and run ./build.sh, then go back down to the docs folder.

docs/README.md
danieltmeta: That might be it. My method somehow make the "magnum" part load, maybe that was the physics libraries.

docs/README.md
aclegg3: Interesting. I agree you can do the source build this way to get a similar result, but the inplace build should do it and worked for me in the past.

docs/README.md
aclegg3: Magnum will load either way, but some parts of the code are compiled out when bullet isn't included.

0mdc (commented):
Added some comments!

PR #2420, opened 2024-07-01
--[Bugfix] - Compute AO cumulative bbox on creation unconditionally.

jturner65:
## Motivation and Context
We've been computing AO bboxes during creation only if they were specified to use visuals based on link shapes.  This PR changes this to always compute cumulative bboxes for loaded AOs.  This was not a problem before due to simulator::reconfigure calling reset() which would recompute the cumulative bbox on the scene root node, which would hit the AOs as well, but any AOs that were loaded after reconfigure would be missed.

## How Has This Been Tested
Local C++ and Python tests pass.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/physics/ArticulatedObject.h
+++ b/src/esp/physics/ArticulatedObject.h
@@ -997,6 +997,16 @@ class ArticulatedObject : public esp::physics::PhysicsObjectBase {
         metadata::attributes::ArticulatedObjectAttributes>();
   }
 
+  /**
+   * @brief Compute the cumulative bbox for this AO
+   */
+  void computeAOCumulativeBB() {
+    baseLink_->node().computeCumulativeBB();
+    for (const auto& link : links_) {
+      link.second->node().computeCumulativeBB();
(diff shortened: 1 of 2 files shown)


aclegg3 (commented):
This is because the links are not children of the root node on the SceneGraph. When no visual nodes are present, there isn't any  content for the link AABB to include. I'm already WIP on writing the correct way to compute these AABBs for the full shape.

jturner65:
> This is because the links are not children of the root node on the SceneGraph. When no visual nodes are present, there isn't any content for the link AABB to include. I'm already WIP on writing the correct way to compute these AABBs for the full shape.

The additional code in this PR fixed the problem where newly-added AOs did not have their bboxes computed.

aclegg3 (approved):
Thanks

aclegg3:
> I'm already WIP on writing the correct way to compute these AABBs for the full shape.

This is an additional change. I'll build on top of this improvement.

PR #2421, opened 2024-07-02
ManagedObject.getAabb()

aclegg3:
## Motivation and Context

Adds a standard API function for querying the AABB of ManagedObjects.

In C++:
```
ManagedObject.getAabb()
```
In python (property):
```
obj.aabb
```

For `ManagedArticulatedObject`s, this function computes the cumulative bounding box from all the child links in the current state. A dirty flag is cached to ensure that re-computation effort is only applied when querying the property after a state change.

## How Has This Been Tested

New tests are added to check that the function works for rigids with golden number and for AOs that all links are within the computed cumulative bounding box.

## Types of changes

## Checklist


src_python/habitat_sim/simulator.py
```diff
@@ -439,6 +439,13 @@ def _sensors(self) -> Dict[str, "Sensor"]:
         # TODO Deprecate and remove
         return self.__sensors[self._default_agent_id]
 
+    @property
+    def scene_bb(self) -> mn.Range3D:
+        """
+        The bounding box of the currently instantiated scene.
+        """
+        return self.get_active_scene_graph().get_root_node().cumulative_bb
+
```
jturner65, line 448: Instead of doing this in simulator.py, how about we put it in a sim binding? Or would that not work

jturner65 (approved):
Just one question/suggestion

src_python/habitat_sim/simulator.py
aclegg3: Yeah, we can do that. Did this for quick implementation initially.

PR #2435, opened 2024-07-25
--[BUGFIX] - Revert markerset subconfig removals; Don't write empty configs to file

jturner65:
Removing the markerset subconfigs when they were empty caused memory corruption in certain rare cases (python variables retaining possession of a removed variable), so [the change introduced in this PR](https://github.com/facebookresearch/habitat-sim/pull/2434) is being reverted; 

Instead, a subconfig with no values in its entire subconfig hierarchy (all of its own subconfigs and so on, instead of just in its own primary map) will not be written to JSON.

## How Has This Been Tested
Locally c++ and python tests. Also verified with python app that fix works.

## Types of changes

## Checklist


0mdc (approved):
LGTM!

PR #2466, opened 2024-09-09
[bugfix] - raycast buffer distance

aclegg3:
## Motivation and Context

This PR fixes a long-standing raycasting bug related to the behavior of Bullet physics raycasting within the collision margin of a convex shape.

**TL;DR: we add a hidden "buffer distance" to the raycast to avoid missing an object from within its margin.**

**Context:** 
- Bullet physics raycasts do not register hits from within convex shapes, neither entering, nor exiting.
- Each collision shape has a configurable margin distance which is used to pad the collision detection algorithm to better catch upcoming impacts between dynamic shapes and avoid discrete collision issues near boundaries.
- (a) Raycasts against convex shapes hit the shape, not the margin.
- (b) Raycast origins within the margin are not considered.
- Taken together, (a) and (b) result in a narrow region of space near the surface of an object where a raycast will not register the object it is about to hit, though expecting a positive hit distance from the API. This "null" zone causes a number of bugs as we attempt to snap objects onto each other or detect which objects are resting upon others.

**Solution:**
- The simple solution is to back-pad the raycast origins and then re-scale the resulting distances such that rays cast within the margin still hit the shape as expected while any hits between the buffer and origin are discarded.

## How Has This Been Tested

CI raycast tests still pass as before with expected global hit points and distances, despite the buffer.
Added new unit tests to specifically target this bug and demonstrate the solution.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/bindings/SimBindings.cpp
+++ b/src/esp/bindings/SimBindings.cpp
@@ -296,6 +296,7 @@ void initSimBindings(py::module& m) {
           R"(Perform discrete collision detection for the scene. Physics must be enabled. Warning: may break simulation determinism.)")
       .def(
           "cast_ray", &Simulator::castRay, "ray"_a, "max_distance"_a = 100.0,
+          "buffer_distance"_a = 0.08,
           R"(Cast a ray into the collidable scene and return hit results. Physics must be enabled. max_distance in units of ray length.)")
       .def("set_object_bb_draw", &Simulator::setObjectBBDraw, "draw_bb"_a,
            "object_id"_a,
--- a/src/esp/physics/bullet/BulletPhysicsManager.cpp
+++ b/src/esp/physics/bullet/BulletPhysicsManager.cpp
@@ -484,7 +484,8 @@ void BulletPhysicsManager::debugDraw(const Magnum::Matrix4& projTrans) const {
 }
 
 RaycastResults BulletPhysicsManager::castRay(const esp::geo::Ray& ray,
-                                             double maxDistance) {
+                                             double maxDistance,
+                                             double bufferDistance) {
   RaycastResults results;
   results.ray = ray;
   double rayLength = static_cast<double>(ray.direction.length());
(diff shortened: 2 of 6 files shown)


src/esp/bindings/SimBindings.cpp
```diff
@@ -296,6 +296,7 @@ void initSimBindings(py::module& m) {
           R"(Perform discrete collision detection for the scene. Physics must be enabled. Warning: may break simulation determinism.)")
       .def(
           "cast_ray", &Simulator::castRay, "ray"_a, "max_distance"_a = 100.0,
+          "buffer_distance"_a = 0.08,
```
0mdc, line 299: Nit: 8cm is a pretty big distance for a buffer. Have you explored setting a lower value? (<=0.01 would be good).

src/esp/physics/bullet/BulletPhysicsManager.cpp
```diff
@@ -505,7 +508,13 @@ RaycastResults BulletPhysicsManager::castRay(const esp::geo::Ray& ray,
     hit.normal = Magnum::Vector3{allResults.m_hitNormalWorld[i]};
     hit.point = Magnum::Vector3{allResults.m_hitPointWorld[i]};
     hit.rayDistance =
-        (static_cast<double>(allResults.m_hitFractions[i]) * maxDistance);
+        (static_cast<double>((allResults.m_hitFractions[i])) - scaledBuffer) *
+        totalLength / rayLength;
```
jturner65, line 512: maybe cache totalLength / rayLength before the loop?

0mdc (approved):
LGTM overall. I would consider lowering the buffer distance.

jturner65 (approved):
Looks good!

src/esp/bindings/SimBindings.cpp
aclegg3: This value must be greater than the maximum expected collision margin (default is 4cm). I think 0.08 (8 cm) therefore is fine as we are ditching any hits within that distance anyway. FYI, the minimum I'd want to use is 4cm, so I doubled that.

src/esp/physics/bullet/BulletPhysicsManager.cpp
aclegg3: Sure, could do. I feel like the math could be more elegant here, honestly, but I didn't come up with a better version so moving on for now. :confounded:

PR #2415, opened 2024-06-25
Add stubs to python bindings to provide typing information.

0mdc:
## Motivation and Context

This changeset adds stubs to python bindings to provide typing information.
Stub generation depends on `pybind11-stubgen`.

https://github.com/facebookresearch/habitat-sim/assets/110583667/86593700-7896-4dc5-9ca7-a6d077beaad9

**Note that this is WIP. Stub contains some errors, but mostly work.**

## How Has This Been Tested

Tested by importing `habitat_sim` in an external Python file.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have completed my CLA (see **CONTRIBUTING**)
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2477, opened 2024-09-28
Generate Python stubs with m.css for IDE autocompletion and type checking

mosra:
Based off #2415, but this time with m.css instead of pybind11-stubgen. Which ensures that the output is consistent with online docs, includes various special casing for Habitat, module renaming, etc.

[Work-in-progress documentation for the stub generator.](https://mcss.mosra.cz/documentation/python/#stub-generation)

I'm marking it as a draft because even though it works quite well on my own Magnum bindings, there's of course an infinite amount of possibilities where it could break on other code, after all it's just about two weeks of work. Important differences to the previous approach:

- The bindings are generated for _the whole_ Habitat project, instead of just the bindings. My thinking is that it'll make more sense this way, because using the `__all__` properties you can control what gets shown in IDE autocompletion hints and thus better drive users to the desirable way of use, _matching the actual documentation_.
- As far as my local testing goes, it should be capable of deciphering _more_ pybind11 annotations than what pybind11-stubgen did, and for those it fails to decipher, it'll fail gracefully. It also exposes enums as actual enums, without having to do something extra.
- **There are no docstrings at the moment.** Right now my focus was on producing code that actually makes sense and can be interpreted by Python. Putting docstrings into the output is relatively trivial in comparison, I'll do that later.
  - Also, because it's made by the same thing as the online docs, there are a few possibilities. It can be just the one-line summaries, it can be _all of it_, but it can also be a summary + a link to online docs. Your choice, let me know.
- You'll probably want to have stubs for Magnum bindings as well, I guess? For those I still have a bunch of local patches so it won't work for you (it asserts otherwise), but here's an archive of them pregenerated: [corrade+magnum-stubs.zip](https://github.com/user-attachments/files/17172138/corrade%2Bmagnum-stubs.zip)
- Eventually, the tool will be capable of producing just the stubs, without also generating the HTML output (and, hopefully, no longer relying on `docutils` etc.). Right now it does both, so it's slower than it could theoretically be.

Looking forward for your feedback. Let me know once you stumble upon something that feels off.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -264,6 +264,12 @@ class CMakeBuild(build_ext):
         for ext in self.extensions:
             self.build_extension(ext)
 
+        # Build stubs for all of 'habitat_sim' including bindings to provide
+        # type information.
+        subprocess.run(
+            subprocess.run(["docs/m.css/documentation/python.py", "docs/conf-stubs.py"])
+        )
+
     def build_extension(self, ext):
         extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -450,21 +456,28 @@ def load(filename):
...
         version=habitat_sim.__version__,
-        author="FAIR A-STAR",
+        author="Meta",
         description="A high performance simulator for training embodied agents",
         long_description="",
         packages=find_packages(where="src_python"),
         package_dir={"": "src_python"},
         install_requires=requirements,
         tests_require=["hypothesis", "pytest-benchmark", "pytest"],
+        setup_requires=["pybind11-stubgen"],
```
mosra, line 466: Hmm, this probably shouldn't be here :thinking:

PR #2478, opened 2024-09-30
--[Bugfix] - Compare Configuration scalar doubles properly

jturner65:
## Motivation and Context
This PR adds, and tests, proper fuzzy comparison between scalar ConfigValues that require it (i.e. doubles) using the same Magnum mechanism that supported Magnum floating point vector-based types use.  This comparison uses a scaled epsilon value that acts as a threshold between two values being equal or not.  See [here for comparison details](https://github.com/mosra/magnum/blob/ccb96f73c7154326589594e6c4260be9e796ecfb/src/Magnum/Math/TypeTraits.h#L485).

Support for fuzzy comparison must be implemented (following the pattern here for doubles) for any numeric types added to the ConfigValue supported list only if they do not use a similar comparison mechanism already (i.e. this does not need to be included for any magnum floating-point types)

## How Has This Been Tested
Locally c++ and python. Tests have been added to validate the process for both below-threshold and above-threshold values.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/core/Configuration.cpp
+++ b/src/esp/core/Configuration.cpp
@@ -884,7 +884,7 @@ Mn::Debug& operator<<(Mn::Debug& debug, const Configuration& cfg) {
 
 bool operator==(const Configuration& a, const Configuration& b) {
   if ((a.getNumSubconfigs() != b.getNumSubconfigs()) ||
-      (a.getNumValues() != b.getNumValues())) {
+      (a.getNumVisibleValues() != b.getNumVisibleValues())) {
     return false;
   }
   for (const auto& entry : a.configMap_) {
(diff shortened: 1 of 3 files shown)


aclegg3 (approved):
LGTM, double makes sense for python APIs particularly. Any other types you predict will need to be explicitly marked (which we may ever use)?

jturner65:
> LGTM, double makes sense for python APIs particularly. Any other types you predict will need to be explicitly marked (which we may ever use)?

We would probably need extend the support if we ever want to differentiate (type-wise) between different size floating points (i.e. add explicit 32bit float support) or if we wanted to support non-magnum floating-point-based numeric constructs.

PR #2497, opened 2024-11-11
[BE] - Remove C++ Agent and Controls

aclegg3:
## Motivation and Context

Concepts of Agents and actions are not part of the Simulation foundation domain. These concepts were moved here to support JS build and internal testing but are completely disjoint from python implementation making them a distraction and tech dept more than useful.

This PR removes those classes from C++ and refactors the necessary functionality in tests and viewer.

Philosophy: C++ Simulator should function on `SensorsSuites` attached to `SceneNodes` on the `SceneGraph`. These attachments should be made by user code downstream (e.g. to simulate Agents) or directly created with ManagedObjects and the `Simulator.addSensorToObject()`. Instead of expecting some arcane API (i.e. ObjectControls) to manage the motion of these SceneNodes, it should be driven by object kinematics, downstream action APIs, or direct user control.

## How Has This Been Tested

C++ and python tests.

## Types of changes

## Checklist


src/esp/sim/Simulator.cpp
```diff
@@ -1126,132 +1054,114 @@ esp::sensor::Sensor& Simulator::addSensorToObject(
...
+      return static_cast<sensor::VisualSensor&>(sensor->second.get())
+          .drawObservation(*this);
     }
   }
   return false;
 }
 
-bool Simulator::visualizeObservation(int agentId, const std::string& sensorId) {
-  agent::Agent::ptr ag = getAgent(agentId);
+bool Simulator::visualizeObservation(const std::string& sensorUuid) {
```
jturner65, line 1108: Minor thing, but there's a naming inconsistency with the observation-based functions.  If we plan on ever supporting non-sensor-based observations, then Simulator::drawObservation and Simulator::visualizeObservation should probably be renamed to drawSensorObservation and visualizeSensorObservation, respectively.

If we do not ever plan on having non-sensor-driven observations, then getSensorObservation should probably just be getObservation

jturner65 (approved):
Minor thing related to naming, but otherwise LGTM

src/esp/sim/Simulator.cpp
aclegg3: Good point for a follow-up. I'll be touching these again with the next phase of refactor. :+1:

PR #2502, opened 2024-11-14
--[BE] - SensorAttributes/Managers refactor Part 1

jturner65:
## Motivation and Context
This PR is the first part of the Sensor refactor effort, responsible for enabling the creation of Sensor-type-specific Attributes from provided SensorSpecs, such that the attributes themselves will be used to create the Sensors (which will be implemented in Part 2).
Specifically, this PR provides the following : 
1. An Attributes hierarchy mirroring the SensorSpec hierarchy
2. A SensorAttributesManager that will manage these SensorAttributes and provide access to them, including a method for creating the appropriate Attributes from a given SensorSpec.
3. Tests to verify the appropriate creation and initialization of a particular Attributes from its equivalent SensorSpec.

## How Has This Been Tested
Locally c++ tests pass. New tests were added to test the valid creation and data content of Attributes from SensorSpecs.

## Types of changes

## Checklist


src/esp/metadata/managers/SensorAttributesManager.cpp
```diff
@@ -0,0 +1,259 @@
...
+                     << static_cast<int32_t>(sensorSpec->sensorSubType),
+                 nullptr);
+  // Build and initialize appropriate class of Attributes based on class name
+  auto sensorAttributes =
+      this->initNewObjectInternal(sensorAttrClassName, false);
+  // Populate attributes from sensorSpec
+  sensorAttributes->populateWithSensorSpec(sensorSpec);
+  // TODO : Rename attributes to appropriate name and register
+  // Build name using more than just spec uuid?
+  std::string newAttrHandle = sensorSpec->uuid;
```
aclegg3, line 89: Probably need some de-duplication system for unique naming. Spec will go away, so this doesn't need to be tied to that field. Naming could be a part of registering if it doesn't come from a filename for example.

src/esp/metadata/managers/SensorAttributesManager.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+  // This should never fail
+  this->setValsFromJSONDoc(sensorAttributes, jsonConfig);
+  return sensorAttributes;
+}  // SensorAttributesManager::buildObjectFromJSONDoc
+
+void SensorAttributesManager::setValsFromJSONDoc(
+    AbstractSensorAttributes::ptr attribs,
+    const io::JsonGenericValue& jsonConfig) {
+  // TODO support loading values from JSON docs for each type of
```
aclegg3, line 164: Would be good to have 1 default for each sensor type, maybe in test_assets? Or programmatically created and added like prims.

aclegg3 (approved):
Looks like a good step toward the planned design. :+1: 
Added some thoughts about next steps mostly.

src/esp/metadata/attributes/CubeMapSensorAttributes.cpp
```diff
@@ -0,0 +1,153 @@
...
+// Fisheye CubeMap
+
+FisheyeSensorAttributes::FisheyeSensorAttributes(const std::string& handle)
+    : AbstractCubeMapSensorAttributes("FisheyeSensorAttributes", handle) {
+  // init double-sphere model-specific values
+  init("ds_alpha", 0.59f);
+  init("ds_xi", -0.18f);
+  // TODO not a legal assignment - focal length needs to be positive in both
+  // fields
+  init("focal_length", Magnum::Vector2(0.0, 0.0));
```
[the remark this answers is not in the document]
jturner65: Not that I've seen in the code - the spec is initialized to 0,0 and then the 'sanity check' makes sure that neither value  is 0, so it is dependent on user input. 

I'll investigate further in the python code (and the lit) to see if there's some good default to use.

src/esp/metadata/managers/SensorAttributesManager.cpp
jturner65: Yeah, I planned on doing that in the 2nd PR.

src/esp/metadata/managers/SensorAttributesManager.cpp
jturner65: Yeah, postCreateRegister below does this (makes sure the name is unique - uuid becomes uuid_XXXX where XXXX is a unique number.

src/esp/metadata/attributes/CubeMapSensorAttributes.cpp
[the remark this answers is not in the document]
jturner65: Seems like 10 mm would be a reasonable choice.

PR #2511, opened 2024-12-02
--[BE] Merge most habitat libraries into a single library to get rid of circular deps.

jturner65:
## Motivation and Context
Habitat has long suffered from circular dependencies since every separate namespace was built as a separate library.  Although having these individual libraries sped up compilation time for developers, it forced link multiplicity of up to 4 for certain libraries to link correctly. 

This PR merges most of these separate libraries (except for bindings and gfx_batch) into a single library, **habsim**, that gets rid of the issues with circular dependencies. Developer compilation time will increase a little since the entire library will have to be rebuilt for any changes to any non-bindings/non-gfx_batch subsystem, but the benefits of no longer having circular dependencies.

This PR also removes all Plugins  (except for PrimitiveImporter) from the main library and instead has shared libraries (i.e. bindings), applications and tests link to the ones they use.

Lastly, this PR introduces support for using Magnum::UfbxImporter for Obj and Fbx files, in an effort to diminish our reliance on Assimp. I think the only support we require from Assimp right now is ascii ply files (StanfordImporter handles binary plys).

## Types of changes
Locally c++ tests pass

## Checklist


aclegg3 (approved):
Thanks @jturner65 , given that most 2nd order users are installing from conda these days and full install isn't that much of a drag, this simplifying refactor seems like a good move.
We can break things back up later if we feel there is value.

mosra:
> Developer compilation time will increase a little since the entire library will have to be rebuilt for any changes to any non-bindings/non-gfx_batch subsystem

Just to clarify, there's no increase, as this happened before as well. In other words, there's no negative effect of this PR whatsoever, only :sparkles: positives :sparkles:.

Before, if you modified something in the, say, `assets` library, everything depending on the `assets` library got relinked, thus also the massive LTO'ed bindings, and ultimately (most of) all unit tests due to the circular dependencies of everything depending on everything else. Now, if you modify something in the `assets/` directory, executables and libraries depending on the `habsim` library gets relinked, thus basically the same set as before, i.e. everything.

Incremental building of just a single unit test is doable the same way as before, by running `ninja SimTest` for example, and the set of incrementally built files shouldn't change compared to before either.

As Alex says, if a certain part of the codebase gets iterated on a lot in isolation, the library can be broken up again. Which is already the case for `gfx_batch`, which is why it's kept separate.

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
+  assets/MeshMetaData.h
+  assets/RenderAssetInstanceCreationInfo.cpp
+  assets/RenderAssetInstanceCreationInfo.h
+  assets/ResourceManager.cpp
+  assets/ResourceManager.h
+  assets/RigManager.cpp
+  assets/RigManager.h
+)
+
+list(APPEND assets_SOURCES ${PbrIBlImageResources})
```
mosra, line 287:
```suggestion
  assets/RigManager.h
  ${PbrIBlImageResources}
)
```

feels cleaner to me, same with other resources.

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
+  ## Enable physics profiling
+  #add_compile_definitions(BT_ENABLE_PROFILE=0)
+  #add_definitions(-DBT_ENABLE_PROFILE)
+
+endif() #BUILD_WITH_BULLET
+
+#################
+## nav
+set(nav_SOURCES nav/GreedyFollower.cpp nav/GreedyFollower.h nav/PathFinder.cpp
+                nav/PathFinder.h
```
mosra, line 453: Stupid autoformatter, can it be convinced to not do this? :sweat_smile: Same elsewhere, putting more files on the same line makes it too easy to miss some when looking at the list.

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
-add_subdirectory(nav)
-add_subdirectory(sensor)
-add_subdirectory(sim)
+##############################
+
+set(habsim_SOURCES ${core_SOURCES})
+list(
+  APPEND
+  habsim_SOURCES
+  ${geo_SOURCES}
```
mosra, line 515: Why not just this? Whitespace is treated as a list item separator; variables, unless put into quotes, get expanded as individual items also.

```suggestion
set(habsim_SOURCES 
  ${core_SOURCES}
  ${geo_SOURCES}
```

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
+         Magnum::SceneTools
+         Magnum::Shaders
+         Magnum::Trade
+         Magnum::Primitives
+         MagnumPlugins::BasisImporter
+         MagnumPlugins::GltfImporter
+         MagnumPlugins::PrimitiveImporter
+         MagnumPlugins::StanfordImporter
+         MagnumPlugins::StbImageImporter
+         MagnumPlugins::StbImageConverter
```
mosra, line 576: While you're doing this, could you also make the change to link the plugins (those, and all other `Importer`, `Converter` targets, also the ones linked conditionally below) only to shared libraries (i.e., the bindings module) and tests, and only where needed? E.g. the `IoTest` won't need any plugins.

Doing that is a significant speedup of the total build / link time.

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
+         MagnumIntegration::Eigen
+         Corrade::Utility
+         Magnum::AnyImageConverter
+  PRIVATE tinyxml2 Detour Recast
+)
+
+# Link windowed application library if needed
+if(BUILD_GUI_VIEWERS)
+  target_link_libraries(habsim PUBLIC Magnum::GlfwApplication)
+endif()
```
mosra, line 586: I thought only the `viewer` executable needs this one, not the library? Last time I checked it wasn't used in the libraries, and if it now is, it's a bad code smell.

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
+
+if(OpenMP_CXX_FOUND)
+  target_link_libraries(habsim PUBLIC OpenMP::OpenMP_CXX)
+endif()
+
+target_link_libraries(
+  habsim
+  PUBLIC Magnum::WindowlessApplication
+)
```
mosra, line 605: On the other hand, this one is fine, the windowless application _is_ used inside the library.

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
+if(BUILD_WITH_BULLET)
+  target_link_libraries(habsim PUBLIC MagnumIntegration::Bullet Bullet::Dynamics)
+endif() #BUILD_WITH_BULLET
+
+# enable when finished? Slows compilation substantially
+# set_directory_properties(PROPERTIES CORRADE_USE_PEDANTIC_FLAGS ON)
```
mosra, line 611: That fix we talked about is in https://github.com/mosra/corrade/commit/280e6b6874e0902086e321d733ce0e90e0c52920. But again, there's _quite a few_ breaking changes in Magnum since the last update, so I'm not sure when it would be the best time to update here.

src/esp/CMakeLists.txt
```diff
@@ -17,15 +17,595 @@ if(BUILD_WARNINGS_AS_ERRORS)
...
+  target_include_directories(habsim PRIVATE ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
+endif()
+
+if(BUILD_WITH_AUDIO)
+  include_directories(${DEPS_DIR}/rlr-audio-propagation/RLRAudioPropagationPkg/headers)
+  target_link_libraries(habsim PUBLIC ${RLRAudioPropagation_LIBRARY})
+endif()
+
+target_link_libraries(
+  habsim
```
mosra, line 557: Hmm, what about naming this `habitat_sim`? Like the repo, like the project itself, consistently with how `habitat_sim_bindings` are named. To me the shorthand feels like an unnecessary introduction of a new name that's not used anywhere else.

src/esp/CMakeLists.txt
jturner65: An artifact of my incremental development. I'll change it.

src/esp/CMakeLists.txt
jturner65: I'll ask you for specifics in slack if I need them :)

src/esp/CMakeLists.txt
jturner65: We can discuss the breaking changes in slack. TBH I would think it worthwhile to investigate updating Habitat to the most recent Magnum so any future Magnum growth will be more easily incorporated should it be useful, but of course that would depend on the scope of the changes.

src/esp/CMakeLists.txt
jturner65: Oh wait, I think we need that for the python viewers too.

src/esp/CMakeLists.txt
mosra: What I wanted to say is that I unfortunately don't currently have the bandwidth to adapt to the changes. In C++ the changed APIs are just emitting deprecation warnings, so that's fine (you can upgrade gradually as necessary, for example), but for Python there's no such deprecation workflow in place, and so all the code needs to be adapted immediately. 

For example there are no application mouse events anymore, it's a [PointerEvent](https://doc.magnum.graphics/python/magnum/platform/glfw/Application/PointerEvent/) now, to have a single abstraction for a mouse, pen and touch. Again, in C++ the old [MouseEvent](https://doc.magnum.graphics/magnum/classMagnum_1_1Platform_1_1GlfwApplication_1_1MouseEvent.html) interface is deprecated, but in Python it's just gone.

If you'd have time to do these changes, however, that's a way... :)

0mdc (approved):
Nice! Thanks for doing this!

src/esp/CMakeLists.txt
jturner65: # cmake-format: off & # cmake-format: on work for this

PR #2521, opened 2024-12-17
Update Magnum to latest

mosra:
## Motivation and Context

My 34th PR updating Magnum? It's been going on for a while it seems. 

The changes go back to July, so here's just a condensed list of what you might be interested in:

- Updated CMake scripts to not warn about version deprecation with CMake 3.31+.
- Magnum now bundles CMake Find modules for 3rd party libraries on its own, meaning Habitat no longer needs to carry them around. Keeping the top-level `FindMagnum` etc bundled would provide a slightly better error message when a system-wide dependency is not found, but given that Habitat defaults to bundled Magnum I don't think it's worth the pain to continue to have them here.
- Various mutable getters, especially in `Containers`, are now marked as `constexpr` on C++14. Let me know if you need any others.
- New [MeshTools::generateTrivialIndices()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a38b601022df7936ed5484a0aea7ba201) utility as a counterpart to all other index generation APIs. The [MeshTools::generateIndices()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#a8b0d9114105a36fbd2147ee129e2017d) is in turn a lot less restrictive, allowing you to throw basically any mesh at it, and it gives back an indexed one no matter what.
- `Platform::Application::Mouse*Event` APIs got deprecated in favor of `Pointer*Event`, which is an abstraction over a mouse as well as a touch and pen input. [High-level intro to pointer events is here.](https://doc.magnum.graphics/magnum/platform.html#platform-windowed-pointer-events) Unfortunately (fortunately?) GLFW doesn't support touch input at all, so for you it's just a matter of renaming and not having to care about anything else, which I did in this PR. (Not for Python though, but @jturner65 promised he'll be on that right after.)
- The `MeshTools` library finally got nice [high-level intro docs](https://doc.magnum.graphics/magnum/meshtools.html), and as a side effect some APIs are now slightly easier to use. (Basically it was easier to fix some APIs to behave better than to have to explain why they're shitty like that.)
- Lots of changes and new features in the [Text](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Text.html) library, but nothing that would really affect you when using the venerable `Text::Renderer2D` APIs. _Yet._
- The [BasisImageConverter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1BasisImageConverter.html) plugin got updated to latest Basis Universal, supporting OpenCL for faster conversion, and HDR formats.
- Various warning fixes, especially related to ARM systems and macOS.

## How Has This Been Tested

:green_apple:


aclegg3:
Thanks @mosra, I'll work on the viewer.py PointerEvent conversion soon. Doesn't look like it should be much of a change. :)

aclegg3:
Ok, that should do it. Feel free to peek.

aclegg3 (approved):
LGTM, thanks!

PR #2539, opened 2025-01-29
Add RenderInstanceHelper for doing visual-only (non-physics) scenes

eundersander:
## Motivation and Context

RenderInstanceHelper lets user python code add render asset instances to the Habitat visual scene and pose them with an efficient numpy batch interface (`RenderInstanceHelper.set_world_poses`).

The overall interface is intentionally minimal, so as to simplify porting the backend to other renderers like the experimental batch renderer.

In terms of specifying the render asset instance, we currently only offer 1. asset filepath (e.g. a GLB file) and 2. a semantic ID. We  can expand this in the future, e.g. scale and material overrides. I'm avoiding using `AssetInfo` and `RenderAssetInstanceCreationInfo` at the python level because I don't think all the options should be exposed to python users.

## How Has This Been Tested

Added a test to SimTest.cpp. Also tested locally in an [example HITL app](https://github.com/facebookresearch/habitat-lab/blob/eundersander/isaac_vr/habitat-lab/habitat/isaac_sim/usd_visualizer.py) that lives outside Habitat-sim.

## Types of changes

## Checklist


src/esp/bindings/SimBindings.cpp
```diff
@@ -138,6 +170,54 @@ void initSimConfigBindings(py::module& m) {
...
+            if (orientations.ndim() != 2 ||
+                orientations.shape(0) != num_instances ||
+                orientations.shape(1) != 4) {
+              throw std::runtime_error(
+                  "orientations must be a 2D array of shape (num_instances, "
+                  "4).");
+            }
+
+            if (!is_contiguous(positions) || !is_contiguous(orientations)) {
+              throw std::runtime_error(
```
0mdc, line 207: Is this error passed to Python?

src/esp/sim/RenderInstanceHelper.cpp
```diff
@@ -0,0 +1,77 @@
...
+RenderInstanceHelper::RenderInstanceHelper(Simulator& sim,
+                                           bool useXYZWOrientations) {
+  sim_ = &sim;
+  isXYZW_ = useXYZWOrientations;
+}
+
+int RenderInstanceHelper::AddInstance(const std::string& assetFilepath,
+                                      int semanticId) {
+  esp::assets::AssetInfo assetInfo;
+  assetInfo.filepath = assetFilepath;
```
0mdc, line 24: Are we resolving the path downstream? Just making sure that `a/b/c` is equivalent to `a/b/../b/c`.

src/esp/sim/RenderInstanceHelper.h
```diff
@@ -0,0 +1,48 @@
...
+  void ClearAllInstances();
+  int GetNumInstances();
+  void SetWorldPoses(float* positions,
+                     size_t positionsSize,
+                     float* orientations,
+                     size_t orientationsSize);
+
+ private:
+  Simulator* sim_ = nullptr;
+  std::vector<scene::SceneNode*> instances_;
```
0mdc, line 39: We should consider creating a `GetInstanceNode(int index)` function to enable handling bounding boxes and the like. This can be done later as needed.

src/esp/sim/RenderInstanceHelper.h
```diff
@@ -0,0 +1,48 @@
...
+class Simulator;
+}
+}  // namespace esp
+
+namespace esp {
+namespace sim {
+
+class RenderInstanceHelper {
+ public:
+  RenderInstanceHelper(Simulator& sim, bool useXYZWOrientations = true);
```
0mdc, line 27: We should consider a name that self-documents the purpose of this class from an API user's perspective. Perhaps something along the lines of `MeshRenderer`?

0mdc (approved):
LGTM! This looks very useful.

I left some high-level comments.

src/esp/bindings/SimBindings.cpp
eundersander: yes, std::runtime_error becomes Python RuntimeError exception

src/esp/sim/RenderInstanceHelper.cpp
eundersander: I don't know, but this kind of path fixup makes more sense to me to do either (a) where the path is constructed or (b) where it is used for OS operations. Doesn't make sense to attempt the fixup here.

src/esp/sim/RenderInstanceHelper.h
eundersander: On the contrary, I don't intend to expose Habitat SceneNodes! One design goal here is to make it easy to port RenderInstanceHelper to a different renderer backend, e.g. the Magnum batch renderer or the Isaac Sim renderer.

src/esp/sim/RenderInstanceHelper.h
eundersander: I dunno, this was the best name I could come up with to capture its very narrow scope. This class itself doesn't actually do any rendering; the user still has to call `habitat_sim.render()` or whatever.

PR #2542, opened 2025-02-04
--[Bugfix] - Fix attributes save

jturner65:
This PR addresses when an attributes filename has to be constructed from its handle. One of the paths to do this looked for the attributes' original file directory in its handle, and then removed that directory +1 from the handle to account for the once-missing trailing slash.  

However,  PR  #2541 modified the fileDirectory that is stored in the attributes to include the trailing slash, which required the modifications in this PR.

This also fixes a minor attributes issue where a boolean default value is not set as an init value (i.e. so it won't be written out to json if it is not changed).

## How Has This Been Tested
Locally c++ and python

## Types of changes

## Checklist


0mdc (approved):
LGTM! Thank you!

PR #2543, opened 2025-02-05
Add stage scaling from configured scale values.

aclegg3:
## Motivation and Context

TL;DR: allows scaling set in the stage configs to propagate through to render and collision assets.

Context:
- We have allowed non-uniform (3D) scale vectors to be configured for object and stage configs ([doc page](https://aihabitat.org/docs/habitat-sim/attributesJSON.html#stageattributes)) for some time.
- However, the stage loading pathway has not used these values during construction of the render and collision objects.
- This PR simply queries existing configured scale when constructing the render instance and collision shapes for the stage asset.

NOTE: this is not tested for split semantic assets. It may require some additional implementation for split semantic scene graphs like HM3D, scannet, MP3D, etc... For context: this is because we load an additional asset into an additional SceneGraph for such assets. The scale should be pulled and set there as well.

## How Has This Been Tested

Locally in viewer. Should be CI tested.

## Types of changes

## Checklist

### DIFF
--- a/src/esp/assets/ResourceManager.cpp
+++ b/src/esp/assets/ResourceManager.cpp
@@ -496,6 +496,9 @@ bool ResourceManager::loadStage(
       renderInfo.filepath, Cr::Containers::NullOpt, flags, renderLightSetupKey);
   ESP_DEBUG() << "Start load render asset" << renderInfo.filepath << ".";
 
+  // Set the stage scale from the config
+  renderCreation.scale = stageAttributes->getScale();
+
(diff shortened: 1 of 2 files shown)


jturner65:
We revisited this 2 years ago, and were going to investigate why it had not been implemented already. I thought at the time that there was a specific reason that we had discussed back when I was implementing the scene instantiation even earlier, on the old slack, but I wasn't able to access the old slack at the time.  I'm going to investigate this a bit to see if I can find what the reason is (I have all our old conversations from that slack, so I'll look real quick)

jturner65 (approved):

PR #2551, opened 2025-03-05
Update to latest Magnum with more efficient glTF mesh import and various UX improvements

mosra:
## Motivation and Context

Regular, 35th update of Magnum dependencies.

This update mainly fixes an old TODO for a more efficient vertex data import in `GltfImporter`, where certain files such as [the Buggy from the glTF sample models repository](https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/Buggy) expanded to over 800 MB of mesh data, while the file alone is just ~7 MB.

With this fix I removed a [MeshTools::interleave()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1MeshTools.html#aa9114dc837636b2f5870289c05aa909b) call inside `GenericMeshData` which was primarily put there to counter the above issue. In certain cases, such as depth-only rendering, it may be beneficial to have the model files processed for example with positions *not* interleaved with the other attributes, and this call would undo that optimization.

Other potentially useful changes besides that:

- Continued cleanup of STL include bloat from all headers, in particular `<utility>` is no longer used anywhere
- [Debug](https://doc.magnum.graphics/corrade/classCorrade_1_1Utility_1_1Debug.html) output can now be redirected to [Containers::String](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1String.html) in addition to `std::ostream`
- Added a long-time-missing [arrayClear()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Containers.html#a81af700f2a5cc53378175ade17bf7d26) API for growable arrays. I searched through the Habitat code and replaced all cases of `arrayResize(..., 0)` with this function.
- [Path::path()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1Path.html#a637fd9afa2270f964273d594852a13b6) and [Path::filename()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1Path.html#a641be10ee95f3176a31e41fccef17efa) shorthands for [Path::split()](https://doc.magnum.graphics/corrade/namespaceCorrade_1_1Utility_1_1Path.html#afa00c1be4062e72ca953c6fec8360ac0). This I also replaced throughout the Habitat code for more readability.
- [StridedArrayView::slice()](https://doc.magnum.graphics/corrade/classCorrade_1_1Containers_1_1StridedArrayView.html#a3d573361f3fbe0c9758a8fa86ae4fac3)-to-a-member-function now works consistently for all possible function overloads
- Fixed [Trade::PbrMetallicRoughnessMaterialData::hasOcclusionRoughnessMetallicTexture()](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1PbrMetallicRoughnessMaterialData.html#a8dfe2a853a28a15da2cc5da114b5e18e) to correctly recognize the case with a packed metallic/roughness + separate occlusion texture
- [GltfImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1GltfImporter.html) no longer uses `std::stable_sort()` for deduplicating mesh and material attributes, which avoids uncontrollable allocations inside STL algorithms and makes the process faster
- [OpenExrImporter](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1OpenExrImporter.html) now works around bugs in OpenEXR 3.3 that prevent files from being correctly loaded

## How Has This Been Tested

:green_circle:

### DIFF
--- a/src/cmake/FindBullet.cmake
+++ b/src/cmake/FindBullet.cmake
@@ -27,7 +27,8 @@
 #   This file is part of Magnum.
 #
 #   Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
-#               2020, 2021, 2022, 2023 Vladimír Vondruš <mosra@centrum.cz>
+#               2020, 2021, 2022, 2023, 2024, 2025
+#             Vladimír Vondruš <mosra@centrum.cz>
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a
 #   copy of this software and associated documentation files (the "Software"),
--- a/src/cmake/FindCorrade.cmake
+++ b/src/cmake/FindCorrade.cmake
@@ -277,7 +277,7 @@
 #   This file is part of Corrade.
 #
 #   Copyright © 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016,
-#               2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024
+#               2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024, 2025
 #             Vladimír Vondruš <mosra@centrum.cz>
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a
--- a/src/cmake/FindMagnum.cmake
+++ b/src/cmake/FindMagnum.cmake
@@ -208,7 +208,7 @@
 #   This file is part of Magnum.
 #
 #   Copyright © 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019,
-#               2020, 2021, 2022, 2023, 2024
+#               2020, 2021, 2022, 2023, 2024, 2025
 #             Vladimír Vondruš <mosra@centrum.cz>
 #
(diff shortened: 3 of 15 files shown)


mosra:
This version finally passes the CI. In https://github.com/facebookresearch/habitat-sim/pull/2551/commits/209da220fb202e7a2f440defdad354ab17ada6a4 I had to add an extra check because the code below is using `mutableIndices()` for some reason, and the mesh might not always have those, for example with builtin [Primitives::cubeSolid()](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Primitives.html#a3293ada161344543b8341f486b3f62c6) that takes the data out of constant memory.

It however feels to me like it shouldn't be necessary to have mutable access to collision mesh positions and indices, so an alternative fix would be to make the [CollisionMeshData](https://github.com/facebookresearch/habitat-sim/blob/35cf4c7c5e095053eed373f6c28d4194c4620e49/src/esp/assets/CollisionMeshData.h#L41-L51) views `const`. Let me know if you want me to do that instead.

0mdc (approved):
LGTM! Thank you!

PR #2555, opened 2025-03-17
Opt-in support for multisampled color rendering

mosra:
## Motivation and Context

As requested by @eundersander on Slack. By enabling `simConfig.multisample` (exposed to both C++ and Python) all sensors that are just color-only will produce a 8x multisampled output to avoid aliasing.

## How Has This Been Tested

There's a new test verifying the output with this flag enabled.


mosra:
Note: the failing test is due to the lighting setup being _somehow_ different in my working copy, resulting in an image that's uniformly brighter in the parts affected by lighting. I suspect it's caused by something that isn't tracked by Git (some config file different? some scene/stage/whatever file missing, and defaults being picked instead?), and I just couldn't figure out what it was.

To fix that, run the failing test locally with `-XS$(pwd)` and replace the file it generated.

jturner65:
> Note: the failing test is due to the lighting setup being _somehow_ different in my working copy, resulting in an image that's uniformly brighter in the parts affected by lighting. I suspect it's caused by something that isn't tracked by Git (some config file different? some scene/stage/whatever file missing, and defaults being picked instead?), and I just couldn't figure out what it was.
> 
> To fix that, run the failing test locally with `-XS$(pwd)` and replace the file it generated.

I'll investigate this when I can. If need be I can regenerate the test image.

PR #2633, opened 2026-05-07
[CI] Fix CUDA 13 build and mypy type annotations

aclegg3:
## Summary

Fixes two CI failures on `main`:

### 1. Source build (Linux GPU CI) — CUDA 13 incompatibility

The Linux GPU CI workflow auto-detects the CUDA version from `nvidia-smi` and installs the matching `cuda-toolkit` (see `.github/actions/install_ubuntu_gpu_deps/action.yml`). The runner host now reports CUDA 13, and CUDA 13 finally removed the long-deprecated `computeMode` and `clockRate` fields from `cudaDeviceProp`. This caused `HelperCuda.h` to fail to compile with:

```
error: 'struct cudaDeviceProp' has no member named 'computeMode'
error: 'struct cudaDeviceProp' has no member named 'clockRate'
```

**Fix:** Replaced the three accesses to `deviceProp.computeMode` and the one access to `deviceProp.clockRate` in `src/esp/gfx_batch/cuda_helpers/HelperCuda.h` with `cudaDeviceGetAttribute` calls (`cudaDevAttrComputeMode` / `cudaDevAttrClockRate`). These attribute queries are supported on **both CUDA 12 and CUDA 13**, so the conda build (CUDA 12.1) and the source build (CUDA 13) both work.

### 2. mypy

`src_python/habitat_sim/utils/datasets_download.py` was missing type annotations on the module-level `data_sources` and `data_groups` dicts:

```
src_python/habitat_sim/utils/datasets_download.py:25: error: Need type annotation for "data_sources"
src_python/habitat_sim/utils/datasets_download.py:26: error: Need type annotation for "data_groups"
```

**Fix:** Added explicit `Dict[str, Dict[str, Any]]` and `Dict[str, List[str]]` annotations.

## Out of scope

The conda upload steps (Linux & macOS) are also failing with `Invalid Username password combination` for the `aihabitat` Anaconda Cloud account. That is a credential/secret rotation issue, not a code issue, and is not addressed here.

## Test plan

Rely on CI: the Linux GPU source build (Conda + source) and the mypy job should now pass.

### DIFF
--- a/src/esp/gfx_batch/cuda_helpers/HelperCuda.h
+++ b/src/esp/gfx_batch/cuda_helpers/HelperCuda.h
@@ -1056,7 +1056,12 @@ inline int gpuDeviceInit(int devID) {
   cudaDeviceProp deviceProp;
   checkCudaErrors(cudaGetDeviceProperties(&deviceProp, devID));
 
-  if (deviceProp.computeMode == cudaComputeModeProhibited) {
+  // NOTE: cudaDeviceProp::computeMode was removed in CUDA 13; query it via
+  // cudaDeviceGetAttribute, which is supported in CUDA 12 and 13.
+  int computeMode = -1;
+  checkCudaErrors(
+      cudaDeviceGetAttribute(&computeMode, cudaDevAttrComputeMode, devID));
+  if (computeMode == cudaComputeModeProhibited) {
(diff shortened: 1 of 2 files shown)


jturner65 (approved):
LGTM

